PR #26, opened 2024-06-20
feat: Added Get Block

a-saksena:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/protos/src/main/protobuf/blockstream.proto
+++ b/protos/src/main/protobuf/blockstream.proto
@@ -46,6 +46,8 @@ service BlockStreamGrpc {
      * message with the id of each block received.
      */
     rpc StreamSource(stream BlockResponse) returns (stream Block) {}
+
+    rpc GetBlock(Block) returns (Block) {}
 }
 
 /**
--- a/protos/src/main/protobuf/blockstream.proto
+++ b/protos/src/main/protobuf/blockstream.proto
@@ -81,3 +81,16 @@ message BlockResponse {
      */
     int64 id = 1;
 }
+
+/**
+ * A block request is a simple message that contains an id.
+ * This specification is a simple example meant to expedite development.
+ * It will be replaced with a PBJ implementation in the future.
+ */
+message BlockRequest {
+    /**
+    * The id of the block which was requested.  Each block id should
+    * correlate with the id of a Block message id.
+    */
+    int64 id = 1;
+}
--- a/server/src/main/java/com/hedera/block/server/BlockStreamService.java
+++ b/server/src/main/java/com/hedera/block/server/BlockStreamService.java
@@ -26,7 +26,6 @@ import com.hedera.block.server.consumer.LiveStreamObserverImpl;
 import com.hedera.block.server.mediator.StreamMediator;
 import com.hedera.block.server.persistence.BlockPersistenceHandler;
 import com.hedera.block.server.producer.ProducerBlockStreamObserver;
-import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
 import io.helidon.webserver.grpc.GrpcService;
 import java.time.Clock;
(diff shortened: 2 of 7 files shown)


protos/src/main/protobuf/blockstream.proto
```diff
@@ -0,0 +1,21 @@
...
+}
+
+message Block {
+  int64 id = 1;
+  string value = 2;
+}
+
+message BlockResponse {
+  int64 id = 1;
+}
```
AlfredoG87, line 17: Why do we need the BlockResponse? is not enough with returning a `Block` ?

server/src/main/java/com/hedera/block/server/BlockStreamService.java
```diff
@@ -0,0 +1,40 @@
...
+import com.hedera.block.protos.BlockStreamServiceGrpcProto;
+import io.grpc.stub.StreamObserver;
+import io.helidon.webserver.grpc.GrpcService;
+
+import java.util.logging.Logger;
+
+import static io.helidon.webserver.grpc.ResponseHelper.complete;
+
+
+public class BlockStreamService implements GrpcService {
```
AlfredoG87, line 13: I think Matt already implemented BlockStream Service, should this be HistoricBlockService?

server/src/main/java/com/hedera/block/server/BlockStreamService.java
```diff
@@ -87,6 +95,7 @@ public String serviceName() {
     public void update(final Routing routing) {
         routing.bidi(CLIENT_STREAMING_METHOD_NAME, this::streamSink);
         routing.bidi(SERVER_STREAMING_METHOD_NAME, this::streamSource);
+        routing.unary("GetBlock", this::getBlock);
```
AlfredoG87, line 98: We should use a Constant as the ones above to avoid having hardcoded values for the method names

server/src/main/java/com/hedera/block/server/BlockStreamService.java
```diff
@@ -129,6 +138,24 @@ private StreamObserver<BlockStreamServiceGrpcProto.BlockResponse> streamSource(f
...
+            LOGGER.log(System.Logger.Level.INFO, "SENDING BLOCK # " + block.getId());
+            complete(responseObserver, responseBlock.get()); // TODO: Should return int and not quoted string
+        } else {
+            LOGGER.log(System.Logger.Level.INFO, "DID NOT FIND YOUR BLOCK");
+            // TODO: Fix below. It could return gRPC equivalent of 404 NOT FOUND
+            responseObserver.onError(Status.NOT_FOUND
+                    .withDescription("DID NOT FIND YOUR BLOCK")
+                    .asRuntimeException()
+            );
+            // complete(responseObserver, BlockStreamServiceGrpcProto.Block.getDefaultInstance());
```
AlfredoG87, line 156: do we need this commented code?

server/src/main/java/com/hedera/block/server/Server.java
```diff
@@ -64,8 +63,11 @@ public static void main(final String[] args) {
 
             // Initialize the block storage, cache, and service
             final BlockStorage<BlockStreamServiceGrpcProto.Block> blockStorage = new FileSystemBlockStorage(BLOCKNODE_STORAGE_ROOT_PATH_KEY, config);
+
+            // TODO: Make timeoutThresholdMillis configurable
```
AlfredoG87, line 67: we can remove this comment, is already configurable, it probably was bringed on the merge

server/src/main/java/com/hedera/block/server/Server.java
```diff
@@ -79,13 +81,18 @@ public static void main(final String[] args) {
                             .bidi(BlockStreamServiceGrpcProto.getDescriptor(),
                                     SERVICE_NAME,
                                     SERVER_STREAMING_METHOD_NAME,
-                                    serverBidiStreamingMethod))
+                                    serverBidiStreamingMethod)
+                            .unary(BlockStreamServiceGrpcProto.getDescriptor(),
+                            "BlockStreamGrpc",
+                            "GetBlock",
```
AlfredoG87, line 87: Same here, we should use Constant values from the constants class

AlfredoG87 (changes requested):
Leaved a few nits

server/src/main/java/com/hedera/block/server/BlockStreamService.java
a-saksena: Keeping this comments for the fix in TODO: in the code.

AlfredoG87 (approved):
LGTM

AlfredoG87 (approved):

Issue #29, opened 2024-06-27
Refactor Live Stream Mediator to decouple producer from consumer


mattp-swirldslabs:
With [this PR](https://github.com/hashgraph/hedera-block-node/pull/23), the live stream of items are broadcast from the producer to all consumers via mediator that all consumers subscribe to.  There are concerns that this solution will not scale well with a larger number of consumers.  This strategy is also at odds with the framework hooks Helidon provides.  The mediator must take greater efforts to make it work.

Given these drawbacks, we should refactor the application to use an intermediate data structure to store the inbound items from the producer.  Consumers will then receive BlockItems from the intermediate data structure and pass them back to their downstream consumer.

PR #40, opened 2024-07-09
feature: design document for streaming blockitems from producer to co…

### LINKED ISSUE #36: Design Document for Live Stream Refactoring
* As a developer, I want to solicit feedback on my approach to Issue #29 
* I'll publish and socialize a design document for the approach

mattp-swirldslabs:
…nsumers

* Design Doc outlining proposed changes to refactor the existing live streaming mechanism.

Fixes #36 

Here is a [Draft PR](https://github.com/hashgraph/hedera-block-node/pull/41) of the Disruptor `prototype` implementation


server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,339 @@
...
+  string value = 2;
+}
+
+/**
+ * A BlockItemResponse is a simple message that contains an id.
+ * The BlockItemResponse is meant to confirm the receipt of a BlockItem.
+ * A future use case may expand on this type to communicate a failure
+ * condition where the BlockItem needs to be resent, etc.
+ */
+message BlockItemResponse {
```
jasperpotts, line 113: I am not sure we need a response to every item. One per block is plenty. It seems a waste of bandwidth and effort to send one per item. If we have 100,000 items a second then that is an extra 100,000 objects that have to be created serialized, sent and garbage collected. 100,000 items is not a crazy number is is only 10 per user transaction which is not far fetched.

jasperpotts: Also the design intent of the feedback was to solve a problem for the consensus node. It has to keep a copy of pre-consensus event stream for events till it is sure those events have been persisted by a block node on disk.  So the response back should be once per block and after it is 100% on disk

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,339 @@
...
+
+**StreamObserver** - StreamObserver is the main interface through which Helidon 4.x.x invokes custom business logic
+to receive and transmit bidirectional BlockItem streams at runtime.
+
+**ProducerBlockItemObserver** - A custom implementation of StreamObserver invoked by Helidon at runtime which is
+responsible for:
+1) Receiving the latest BlockItem from the producer (e.g. Consensus Node).
+2) Returning a response to the producer.
+
+**StreamMediator** - StreamMediator is an implementation of the [Mediator Pattern](https://en.wikipedia.org/wiki/Mediator_pattern)
```
jasperpotts, line 296: I don't understand the value a mediator brings other than adding complexity?

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,339 @@
+# Bi-directional Producer/Consumer Streaming with gRPC
+
+## Purpose
+
+A primary use case of the `hedera-block-node` is to stream live BlockItems (see Terms section) from a producer 
+(e.g. Consensus Node) to N consumers (e.g. Mirror Node) with the lowest possible latency while correctly preserving the 
+order of the BlockItems.  This document outlines several possible strategies to implement this use case and the design
```
rbair23, line 7: nit: one space is preferred (https://www.grammarly.com/blog/spaces-after-period/). I know, it is unimportant bike-shedding, but I must stand on my soap box every once in a while :-D

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,339 @@
...
+
+Helidon provides well-defined APIs and extension points to implement business logic for these services. The main entry 
+point for custom logic is an implementation of `GrpcService`.
+
+---
+
+## Goals
+
+1) Consumers must be able to dynamically subscribe and unsubscribe from the live stream of BlockItems emitted by the
+    producer.
```
rbair23, line 20: Interesting -- if I am a consumer and I subscribe, do I get whatever the next block item is, or do I start from the beginning of a block?

mattp-swirldslabs: This implementation was for the "live stream" so the consumer would get the newest BlockItems in-flight.  It sounds like there are thousands of BlockItems per block.  Is the correct behavior to always start the stream by emitting the first BlockItems from the current Block and then catch up to the "live stream"?  There's a future story where the consumer can specify the block number they'd like to start from and they'd receive continuous stream until they catch up to the "live stream".

jasperpotts: I was expecting we would not start them till the first item of next block. Otherwise sending useless data they are paying for.

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,339 @@
...
+  * BlockItems from the server.  The consumer is expected to respond with a BlockResponse message
+  * with the id of each BlockItem received.
+ */
+service BlockStreamGrpc {
+  /**
+  * StreamSink is a bidirectional streaming method that allows a producer to stream BlockItems 
+  * to the Block Node server.  The server will respond with a BlockResponse message for each
+  * BlockItem received.
+  */
+  rpc StreamSink (stream BlockItem) returns (stream BlockItemResponse) {}
```
jasperpotts, line 80: I don't love the Source and Sink names in API. Saying that open to better ideas `StreamIn` and `StreamOut` are first that come to mind. @rbair23 thoughts?

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: If a new consumer subscribes and just misses the start of a Block, what's the anticipated latency until they'd start seeing BlockItems from the next Block? Is there a danger they think the stream "isn't working" because they're not seeing data immediately flowing by and they retry connections in frustration?

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Is this something the gradle plugins can detect and handle?  Our repo isn't quite up to speed yet on leveraging the plugin capabilities.

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: I can adjust the doc and message type to be `BlockResponse` to communicate that the consumer should only send 1 response per Block received.  Is this a type we need to add to the `hedera-protobufs` or does something like this already exist?

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Should the names reflect additional properties of the methods to leave room for future definitions?  For example, `UnverifiedStreamOut` to leave room for another method called `VerifiedStreamOut`?

server/docs/design/bidi-producer-consumers-streaming.md
Nana-EC: @jsync-swirlds I think this is matching up with [PublishStreamResponse](https://github.com/hashgraph/hedera-protobufs/blob/continue-block-node/block/block_service.proto#L110) do we need to rethink this?

jsync-swirlds: I believe the current service definition accomplishes what Jasper is asking for.  The timing of _when_ the response is sent is more in the protocol realm, and if we want to start by only acknowledging once per block, that's fine.

mattp-swirldslabs: Good point...Joseph and I were just discussing the overlap with the API services defined in [hedera-protobufs](https://github.com/hashgraph/hedera-protobufs/blob/25783427575ded59d06d6bf1ed253fd24ef3c437/block/block_service.proto#L675).  I'll rename these to align with those names and types.

server/docs/design/bidi-producer-consumers-streaming.md
jsync-swirlds: Build plugins won't adjust or modify comments or markdown.
English grammar and typesetting rules are pretty much up to us to follow.

server/docs/design/bidi-producer-consumers-streaming.md
jsync-swirlds: Latency should be a second or less if they miss substantially the entire block.
Current blocks are about one per second, and may be more frequent with future gossip improvements.

server/docs/design/bidi-producer-consumers-streaming.md
jsync-swirlds: I'm not a fan of putting `in` or `out` in rpc verbs.
We currently have the RPC verbs defined as follows for publishing a stream to the block node and subscribing to a stream from the block node:
```protobuf
    rpc publishBlockStream (stream PublishStreamRequest) returns (stream PublishStreamResponse);
    rpc subscribeBlockStream(stream SubscribeStreamRequest) returns (stream SubscribeStreamResponse);
```

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: I removed the API definition and replaced it with a link to the `hedera-protobufs` repo and the branch where to find the correct service definitions and types.

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: @jsync-swirlds Ok, makes sense.  

So, @jasperpotts and @rbair23, if I understand the choices, we have:

1) The system will wait until the next Block to start live streaming BlockItems - consumers will accept the latency

2) The system will immediately stream BlockItems on the current Block to backfill missed items until it catches up to the live stream of BlockItems - consumers will accept some stale BlockItems until the system catches up.

Which one should we go with?

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: I removed the API definition and replaced it with a link to the hedera-protobufs repo and the branch where to find the correct service definitions and types.

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Bummer.  Ok, I changed all the occurrences in this design doc manually.

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: The `StreamMediator` seemed like a reasonable component to encapsulate details around the [Disruptor integration with the StreamObserver](https://github.com/hashgraph/hedera-block-node/blob/fef178deb4f4a22fd4135f370fe87cc74148f108/server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorImpl.java#L77-L105). Without this abstraction, the `RingBuffer` would need to be anchored on the `BlockStreamService` and all subscribe/unsubscribe operations to the `RingBuffer` would be mixed with `GrpcService API` concerns. Do you object to the name? Maybe we could change that to something like `BroadcastStreamController`?

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Given feedback in the demo, I've updated the doc to support option #1

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: I'll keep the `StreamMediator` for the time being unless there's additional feedback.

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -27,20 +28,20 @@ point for custom logic is an implementation of `GrpcService`.
...
+allowing both a client and a server emit a continuous stream of frames without waiting for responses. In this way, gRPC 
+services can be used to efficiently transmit a continuous flow of BlockItem messages while the HTTP/2 connection is open.
 
 **Producer StreamObserver** - The Producer StreamObserver is a custom implementation of the [gRPC StreamObserver 
-interface](https://github.com/grpc/grpc-java/blob/0ff3f8e4ac4c265e91b4a0379a32cf25a0a2b2f7/stub/src/main/java/io/grpc/stub/StreamObserver.java#L45) used by Helidon.  It is initialized by the BlockItemStreamService (see Entities section).  Helidon invokes the Producer
-StreamObserver at runtime when the producer sends a new BlockItem to the `StreamSink` gRPC service. 
+interface](https://github.com/grpc/grpc-java/blob/0ff3f8e4ac4c265e91b4a0379a32cf25a0a2b2f7/stub/src/main/java/io/grpc/stub/StreamObserver.java#L45) used by Helidon. It is initialized by the BlockItemStreamService (see Entities section). Helidon invokes
+the Producer StreamObserver at runtime when the producer sends a new BlockItem to the `StreamSink` gRPC service.
```
jsync-swirlds, line 39: `StreamSink` ==> `publishBlockStream`

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -27,20 +28,20 @@ point for custom logic is an implementation of `GrpcService`.
...
+the Producer StreamObserver at runtime when the producer sends a new BlockItem to the `StreamSink` gRPC service. 
 
 **Consumer StreamObserver** - The Consumer StreamObserver is a custom implementation of the [gRPC StreamObserver 
-interface](https://github.com/grpc/grpc-java/blob/0ff3f8e4ac4c265e91b4a0379a32cf25a0a2b2f7/stub/src/main/java/io/grpc/stub/StreamObserver.java#L45) used by Helidon.  It is initialized by the BlockItemStreamService (see Entities section).  Helidon invokes the Consumer
-StreamObserver at runtime when the downstream consumer of the `StreamSource` gRPC service sends HTTP/2 responses to 
-sent BlockItems. 
+interface](https://github.com/grpc/grpc-java/blob/0ff3f8e4ac4c265e91b4a0379a32cf25a0a2b2f7/stub/src/main/java/io/grpc/stub/StreamObserver.java#L45) used by Helidon. It is initialized by the BlockItemStreamService (see Entities section). Helidon invokes
+the Consumer StreamObserver at runtime when the downstream consumer of the `StreamSource` gRPC service sends HTTP/2
```
jsync-swirlds, line 43: `StreamSource` ==> `subscribeBlockStream`

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -27,20 +28,20 @@ point for custom logic is an implementation of `GrpcService`.
...
 
 **Consumer StreamObserver** - The Consumer StreamObserver is a custom implementation of the [gRPC StreamObserver 
-interface](https://github.com/grpc/grpc-java/blob/0ff3f8e4ac4c265e91b4a0379a32cf25a0a2b2f7/stub/src/main/java/io/grpc/stub/StreamObserver.java#L45) used by Helidon.  It is initialized by the BlockItemStreamService (see Entities section).  Helidon invokes the Consumer
-StreamObserver at runtime when the downstream consumer of the `StreamSource` gRPC service sends HTTP/2 responses to 
-sent BlockItems. 
+interface](https://github.com/grpc/grpc-java/blob/0ff3f8e4ac4c265e91b4a0379a32cf25a0a2b2f7/stub/src/main/java/io/grpc/stub/StreamObserver.java#L45) used by Helidon. It is initialized by the BlockItemStreamService (see Entities section). Helidon invokes
+the Consumer StreamObserver at runtime when the downstream consumer of the `StreamSource` gRPC service sends HTTP/2 
+responses to sent BlockItems. 
 
 **subscribe** - Consumers calling the `StreamSource` gRPC service must be affiliated or subscribed with a producer to
```
jsync-swirlds, line 46: `StreamSource` ==> `subscribeBlockStream`

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -52,118 +53,54 @@ unsubscribed from a producer so that internal objects can be cleaned up and reso
...
 
 All the following approaches require integrating with Helidon 4.x.x gRPC services to implement the bidirectional 
-streaming API methods defined above.  The following objects are used in all approaches:
+streaming API methods defined above. The following objects are used in all approaches:
 
 `BlockItemStreamService` is a custom implementation of the Helidon gRPC `GrpcService`. It is responsible for binding 
 the Helidon routing mechanism to the gRPC streaming methods called by producers and consumers.
 
 `ProducerBlockItemObserver` is a custom implementation of the Helidon gRPC `StreamObserver` interface.  
 `BlockItemStreamService` instantiates a new `ProducerBlockItemObserver` instance when the `StreamSink` gRPC method is
```
jsync-swirlds, line 71: `StreamSink` ==> `publishBlockStream`

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -52,118 +53,54 @@ unsubscribed from a producer so that internal objects can be cleaned up and reso
...
-called by a producer.  Thereafter, Helidon invokes `ProducerBlockItemObserver` methods to receive the latest BlockItem 
+called by a producer. Thereafter, Helidon invokes `ProducerBlockItemObserver` methods to receive the latest BlockItem 
 from the producer and return BlockItemResponses via a bidirectional stream.
 
 `ConsumerBlockItemObserver` is also a custom implementation of the Helidon gRPC `StreamObserver` interface.  
 `BlockItemStreamService` instantiates a new `ConsumerBlockItemObserver` instance when the `StreamSource` gRPC method
```
jsync-swirlds, line 76: `StreamSource` ==> `subscribeBlockStream`

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,276 @@
...
+
+**Consumer StreamObserver** - The Consumer StreamObserver is a custom implementation of the [gRPC StreamObserver 
+interface](https://github.com/grpc/grpc-java/blob/0ff3f8e4ac4c265e91b4a0379a32cf25a0a2b2f7/stub/src/main/java/io/grpc/stub/StreamObserver.java#L45) used by Helidon. It is initialized by the BlockItemStreamService (see Entities section). Helidon invokes
+the Consumer StreamObserver at runtime when the downstream consumer of the `StreamSource` gRPC service sends HTTP/2 
+responses to sent BlockItems. 
+
+**subscribe** - Consumers calling the `StreamSource` gRPC service must be affiliated or subscribed with a producer to 
+receive a live stream of BlockItems from the `hedera-block-node`.
+
+**unsubscribe** - Consumers terminating their connection with the `StreamSource` gRPC service must be unaffiliated or
```
jsync-swirlds, line 49: `StreamSource` ==> `subscribeBlockStream`

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,276 @@
...
+
+When a producer calls the `StreamSink` gRPC method, the `BlockItemStreamService` will create a new 
+`ProducerBlockItemObserver` instance for Helidon to invoke during the lifecycle of the bidirectional connection to the 
+upstream producer. The `ProducerBlockItemObserver` is constructed with a reference to the `StreamMediator` and to
+the `ResponseStreamObserver` managed by Helidon for transmitting BlockItemResponses to the producer.
+See the Producer Registration Flow diagram for more details.
+
+### Consumer Registration Flow
+
+When a consumer calls the `StreamSource` gRPC method, the `BlockItemStreamService` will create a new
```
jsync-swirlds, line 189: `StreamSource` ==> `subscribeBlockStream`

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,276 @@
...
+Given the goals and the proposed approaches, Approach #4 has significant advantages and fewer significant drawbacks.  
+Using the LMAX Disruptor offers low latency and CPU consumption via a well-maintained and tested API. The RingBuffer
+intermediate data structure should serve to decouple the producer bidirectional stream from the consumer bidirectional
+streams. Please see the following Entities section and Diagrams for a visual representation of the design.
+
+### Producer Registration Flow
+
+At boot time, the `BlockItemStreamService` will initialize the `StreamMediator` with the LMAX Disruptor RingBuffer.
+
+When a producer calls the `StreamSink` gRPC method, the `BlockItemStreamService` will create a new
```
jsync-swirlds, line 181: `StreamSink` ==> `publishBlockStream`

jsync-swirlds (commented):
A wording suggestion, and some missed gRPC method names.

AlfredoG87 (commented):
LGTM, leaved some questions only

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Fixed

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Fixed

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Fixed

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Fixed

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Fixed

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Fixed

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Fixed

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Fixed

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,276 @@
+# Bi-directional Producer/Consumer Streaming with gRPC
+
+## Purpose
+
+A primary use case of the `hedera-block-node` is to stream live BlockItems (see Terms section) from a producer 
+(e.g. Consensus Node) to N consumers (e.g. Mirror Node) with the lowest possible latency while correctly preserving the 
+order of the BlockItems. This document outlines several possible strategies to implement this use case and the design 
+of the recommended approach. All strategies rely on the Helidon 4.x.x server implementations of HTTP/2 and gRPC 
+services to ingest BlockItem data from a producer and then to stream the same BlockItems to downstream consumers. It
+does this by defining bidirectional gRPC streaming services based on protobuf definitions.
```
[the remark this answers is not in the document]
mattp-swirldslabs: Good question.  Yes, the consumer-facing connection will be "server streaming" (unidirectional stream to the consumer after the initial request).  I've already implemented that in #29 but I thought I would adjust this design document with that PR so that the changes go together in the same PR.

server/docs/design/bidi-producer-consumers-streaming.md
```diff
@@ -0,0 +1,276 @@
...
+**Consumer StreamObserver** - The Consumer StreamObserver is a custom implementation of the [gRPC StreamObserver 
+interface](https://github.com/grpc/grpc-java/blob/0ff3f8e4ac4c265e91b4a0379a32cf25a0a2b2f7/stub/src/main/java/io/grpc/stub/StreamObserver.java#L45) used by Helidon. It is initialized by the BlockItemStreamService (see Entities section). Helidon invokes
+the Consumer StreamObserver at runtime when the downstream consumer of the `StreamSource` gRPC service sends HTTP/2 
+responses to sent BlockItems. 
+
+**subscribe** - Consumers calling the `StreamSource` gRPC service must be affiliated or subscribed with a producer to 
+receive a live stream of BlockItems from the `hedera-block-node`.
+
+**unsubscribe** - Consumers terminating their connection with the `StreamSource` gRPC service must be unaffiliated or 
+unsubscribed from a producer so that internal objects can be cleaned up and resources released.
```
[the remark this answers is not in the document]
mattp-swirldslabs: Not at this point.  I found a really helpful callback that disconnects a consumer and unsubscribes them as part of Helidon.  I think that should be enough.  We might revisit this scenario when we implement gRPC authentication.

server/docs/design/bidi-producer-consumers-streaming.md
mattp-swirldslabs: Fixed

AlfredoG87 (approved):
LGTM

PR #43, opened 2024-07-11
CI Test Check for Style guidelines using Spotless plugin

### LINKED ISSUE #31: Add spotless gradle plugin support And Verify that is Working as Expected
Add spotless gradle plugin support to enforce SL styleguide

AlfredoG87:
**Description**:
Add a CI Test that checks that the spotless check passes, otherwise it will fail and recommend to run the following command:
```
./gradlew spotlessApply
```

**Related issue(s)**:

Fixes #31 

**Notes for reviewer**:
Raised another PR with the same exact changes, but added another commit with a modification of style so to force the failure of the check.

see here:
https://github.com/hashgraph/hedera-block-node/pull/42
**Checklist**

### DIFF
--- a/.github/workflows/spotless-check.yml
+++ b/.github/workflows/spotless-check.yml
@@ -1,3 +1,19 @@
+##
+# Copyright (C) 2024 Hedera Hashgraph, LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##
+
 name: Spotless Check
 
 on: [pull_request]
--- a/.github/workflows/spotless-check.yml
+++ b/.github/workflows/spotless-check.yml
@@ -33,6 +33,14 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Expand Shallow Clone for Spotless
+        run: |
+          if [ -f .git/shallow ]; then
+            git fetch --unshallow --no-recurse-submodules
+          else
+            echo "Repository is not shallow, no need to unshallow."
+          fi
+
       - name: Set up JDK 21
         uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 # v4.2.1
         with:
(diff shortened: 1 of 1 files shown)


jsync-swirlds:
It might be worth taking a look at [services compile checks](https://github.com/hashgraph/hedera-services/blob/develop/.github/workflows/node-zxc-compile-application-code.yaml), particularly the CGroups setup, to see if we can improve this by using their action code/setup as a model.  We might also want to include module checks (from the same reference), as a way to ensure the module setup remains clean and consistent.

.github/workflows/spotless-check.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Harden Runner
+        uses: step-security/harden-runner@17d0e2bd7d51742c71671bd19fa12bdc9d40a3d6 # v2.8.1
+        with:
+          egress-policy: audit
+
+      - name: Checkout code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          fetch-depth: 0
+
```
jsync-swirlds, line 19: Do we also want to expand the git clone to ensure spotless can apply to all of the code?
```yaml
      - name: Expand Shallow Clone for Spotless
        run: git fetch --unshallow --no-recurse-submodules
```

AlfredoG87: tried to add that step but got the following error, so reverted back to how it was:

```
Run git fetch --unshallow --no-recurse-submodules
  git fetch --unshallow --no-recurse-submodules
  shell: /usr/bin/bash -e {0}
fatal: --unshallow on a complete repository does not make sense
Error: Process completed with exit code 1[2](https://github.com/hashgraph/hedera-block-node/actions/runs/9898941525/job/27346849024?pr=43#step:5:2)8.
```

AlfredoG87:
> It might be worth taking a look at [services compile checks](https://github.com/hashgraph/hedera-services/blob/develop/.github/workflows/node-zxc-compile-application-code.yaml), particularly the CGroups setup, to see if we can improve this by using their action code/setup as a model. We might also want to include module checks (from the same reference), as a way to ensure the module setup remains clean and consistent.


Yes. I did take a look at the services, and that specific workflow as a source of inspiration, however implementing everything in there is probably not necessary right now and out of scope of current task.

But I agree that as the Codebase grows and in time, when we have many CI jobs we can use that model and refactor them all into a single CI WF.

But for now, the objective of this CI Test is to protect the style guidelines.

.github/workflows/spotless-check.yml
AlfredoG87: Upon further research, realized that I can put a conditional to unshallow only if necessary.
image

mattp-swirldslabs (approved):
LGTM

PR #83, opened 2024-08-05
fix: upgraded gradlew to 8.8 to match services and fix 00081

### LINKED ISSUE #81: gradlew tasks command throws errors
As a Block Node Developer I need for `./gradlew tasks` command to complete without
errors, so that additional gradle work will not be impacted by cascading errors.

* After spelunking through git, it appears that this `tasks` command was broken when
we moved to using the `protos` modules, etc.

```
> Task :tasks FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':tasks'.
> Could not create task ':protos:moduleDependencies'.
   > 'org.gradle.api.provider.MapProperty org.gradle.api.provider.MapProperty.unset()'
```

mattp-swirldslabs:
**Description**:
`./gradlew tasks` command was failing because one of the plugins was relying on a MapProperty function in a later version of gradle (at least 8.8)
* I upgraded gradlew from 8.5 to 8.8 and this fixed the issue

**Related issue(s)**:

Fixes #81


AlfredoG87 (approved):
LGTM, nice finding!

PR #103, opened 2024-08-14
fix: fix config exports

### LINKED ISSUE #104: Starting Block Node fails
### Description

Trying to startup block node grpc server fails with following error:

```
Caused by: java.lang.IllegalAccessException: class com.swirlds.config.impl.internal.ConfigDataFactory (in module com.swirlds.config.impl) cannot access class com.hedera.block.server.consumer.ConsumerConfig (in module com.hedera.block.server) because module com.hedera.block.server does not export com.hedera.block.server.consumer to module com.swirlds.config.impl
```

### Steps to reproduce

1. Run `./gradlew run` command

### Additional context

_No response_

### Hedera network

_No response_

### Version

latest

### Operating system

None

AlfredoG87:
**Description**:
On a previous PR #101  we introduced Configuration Classes (Config Holders and owners) using the `Platform SDK` framework, however we removed from modules-info.java file the necessary exports, and this is an issue that is only happening on runtime, so CI checks did not catch it.

This PR fixes that by re-introducing them back.

**Related issue(s)**:

Fixes #104 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/server/src/main/java/module-info.java
+++ b/server/src/main/java/module-info.java
@@ -2,6 +2,11 @@ import com.hedera.block.server.config.BlockNodeConfigExtension;
 
 /** Runtime module of the server. */
 module com.hedera.block.server {
+    exports com.hedera.block.server.consumer to
+            com.swirlds.config.impl;
+    exports com.hedera.block.server.persistence.storage to
+            com.swirlds.config.impl;
+
     requires com.hedera.block.protos;
     requires com.google.protobuf;
     requires com.lmax.disruptor;


georgi-l95 (approved):
LG, I'd suggest that we add ./gradle run as a CI check, maybe after the unit tests in the same workflow, just for sanity check that we didn't brake it after our changes.

mattp-swirldslabs (approved):
Looks good

PR #113, opened 2024-08-15
test: Smoke test

### LINKED ISSUE #109: Smoke test in CI
AS a Block Node Developer
I WANT TO make sure that my PR is not break runtime functionality 
SO THAT PR reviews are safer and quicker.

AlfredoG87:
**Description**:
- Add a new CI that performs a Smoke-Test on the BlockNode Server
- Adds 2 new scripts that are leveraged in order to perform the smoke-test:
  - get-block.sh
  - smoke-test.sh

**Related issue(s)**:

Fixes #109 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/server/src/test/resources/smoke-test.sh
+++ b/server/src/test/resources/smoke-test.sh
@@ -57,15 +57,15 @@ else
     echo "No log file provided, skipping startup pattern check."
 fi
 
-# 2. Start the producer script with parameter 1 and save logs to producer.log
-./producer.sh 1 > producer.log 2>&1 &
-PRODUCER_PID=$!
-echo "Started producer with PID $PRODUCER_PID, logging to producer.log"
-
-# 3. Start the consumer script with parameters 1 1000 and save logs to consumer.log
+# 2. Start the consumer script with parameters 1 1000 and save logs to consumer.log
 ./consumer.sh 1 1000 > consumer.log 2>&1 &
 CONSUMER_PID=$!
 echo "Started consumer with PID $CONSUMER_PID, logging to consumer.log"
+
+# 3. Start the producer script with parameter 1 and save logs to producer.log
+./producer.sh 1 > producer.log 2>&1 &
+PRODUCER_PID=$!
+echo "Started producer with PID $PRODUCER_PID, logging to producer.log"
 
 # Sleep time after starting the consumer
 sleep 5  # Adjust sleep time as needed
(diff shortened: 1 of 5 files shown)


server/src/test/resources/smoke-test.sh
```diff
@@ -0,0 +1,106 @@
...
+    if grep -q "$STARTUP_PATTERN" "$LOG_FILE"; then
+        echo "Startup pattern found in logs."
+    else
+        echo "Startup pattern not found in logs."
+        exit 1
+    fi
+else
+    echo "No log file provided, skipping startup pattern check."
+fi
+
```
mattp-swirldslabs, line 59: Are you seeing output from the `consumer.sh` in this order? The consumers should only start seeing output at the beginning of the next full block (header blockitem). If we're only sending one block, then whether the consumer sees output would be a race condition. If we switch the order of #2 and #3 then the consumer *should* always see the whole block in their output.

mattp-swirldslabs (dismissed):
Looks good! Thanks for adding this. I think we'll see a lot of benefit.

server/src/test/resources/smoke-test.sh
```diff
@@ -0,0 +1,106 @@
...
+        fi
+    fi
+}
+
+# Trap any exit signal to ensure everything is cleaned up
+trap "shutdown; exit 1" ERR SIGINT SIGTERM
+
+# 1. Verify that the logs have the startup pattern (only if log file is provided)
+LOG_FILE=$1
+STARTUP_PATTERN="Block Node Server started at port"
```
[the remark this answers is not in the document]
AlfredoG87: Since we decided that once we have more ports we will update, I leave it the same on the other PR that was already merged.

No need to do any change.

server/src/test/resources/smoke-test.sh
AlfredoG87: Thanks for the tip!

✅  Done!

mattp-swirldslabs (approved):
LGTM

mattp-swirldslabs (dismissed):

mattp-swirldslabs (approved):
LGTM

georgi-l95 (approved):

jeromy-cannon (approved):

rbarkerSL (approved):

PR #216, opened 2024-09-26
build: use common Hiero Gradle config

jjohannes:
**Description**:

Before this can be integrated:

- [x] Publish _hiero-gradle-conventions_ plugins to Gradle Plugin Portal and use from there

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/14255


buildSrc/build.gradle.kts
jsync-swirlds, line 1: With all of these files moved to a single shared conventions plugin, what options exist to modify that behavior for different projects (e.g. this project may not retain identical java version settings to hedera-services indefinitely).

jjohannes: Goal is to move everything that's to be controlled by projects - like the Java version – out of the Gradle convention plugins into separate configuration files.
The Java version will be located here after this change:  [gradle/jdk-version.txt](https://github.com/hashgraph/hedera-block-node/blob/14255-reusable-gradle-config/gradle/jdk-version.txt)

I am fairly confident that we already have a good compromise between what we want to share purposefully to have a alignment between projects and what projects still should/need to control individually. But we will observer this closely after this change and do adjustments as needed.

.github/dependabot.yml
```diff
@@ -11,7 +11,12 @@ updates:
...
+    directory: /hedera-dependency-versions
     schedule:
-      interval: "weekly"
\ No newline at end of file
+      interval: weekly
+
+  - package-ecosystem: gradle
+    directory: /
+    schedule:
+      interval: weekly
```
jjohannes, line 22: With having all versions in the standard format in `hiero-dependency-versions`, Dependbot will start working on this Repo for Gradle-managed dependencies as it does on others repos.

.github/workflows/pr-checks.yaml
```diff
@@ -74,15 +60,15 @@ jobs:
 
       - name: Build
         id: gradle-build
-        run: ${GRADLE_EXEC} build
+        run: ${GRADLE_EXEC} assemble --scan
```
jjohannes, line 63: `build` did everything already and the two other steps below were obsolete.

The split-up in three steps is what we also do in other repos:

- assemple (everything compiles)
- qualityCheck (all checks: spotless, dependency scopes)
- check (run tests)

mishomihov00 (dismissed):
Review applies to `.github/dependabot.yml` and `.github/workflows/pr-checks.yaml`

gradle/wrapper/gradle-wrapper.jar
rbarkerSL, line 1: Should this jar be present? I thought that gradle would pick this up natively. Maybe a miss in the .gitignore?

rbarkerSL (changes requested):
Just one question. Prefer to avoid storing binaries if we can avoid it.

gradle/wrapper/gradle-wrapper.jar
jsync-swirlds: Gradle recommends storing the wrapper jar in the code repo.
The same jar is present in our other gradle managed repos (e.g. services [here](https://github.com/hashgraph/hedera-services/tree/develop/gradle/wrapper)).

jjohannes: Yes that's correct. This Jar contains the "Wrapper Tool" that downloads Gradle itself in the correct version so you do not need to install it manually. And yes, it would be somehow better if it would not be a binary, but it's the only solution available.

On the upside, the Gradle GH action we use validates that this file is the official one. So if if someone would attempt to sneak in a compromised Jar disguised as a Gradle update, the action would detect it.

The Jar typically does not need to be updated. But sometimes there are new features in the Wrapper itself after a Gradle update. Then, when you changed the Gradle version, you can run `./gradlew wrapper` and the Jar is update if a new version exists (which was the case here).

mattp-swirldslabs (commented):
Hi Jendrik, I need to merge in our implementation of the deterministic build and deterministic images before you merge this PR. Is that ok?

jjohannes:
@mattp-swirldslabs sure go ahead.
I am taking some days off for the holidays. I won't get back to this before January anyway.
I will deal with potential conflicts then if there are any.

AlfredoG87 (approved):
LGTM, however needs conflict resolution due to parallel efforts to make the build deterministic.

stream/build.gradle.kts
```diff
@@ -30,28 +31,32 @@ tasks.withType<JavaCompile>().configureEach {
...
+    tasks.register<GitClone>("cloneHederaProtobufs") {
+        url = "https://github.com/hashgraph/hedera-protobufs.git"
+        localCloneDirectory = layout.buildDirectory.dir("hedera-protobufs")
 
-    // uncomment below to use a specific branch
-    // branch = "main"
-}
+        // uncomment below to use a specific tag
+        // tag = "v0.53.0" or a specific commit like "0047255"
+        tag = "1033f10"
```
rbarkerSL, line 41: Assuming this is for testing prior to merge? Will we be back on main when we do merge?

rbarkerSL (dismissed):
Looks okay.

stream/build.gradle.kts
jjohannes: This is the same as on main. The code is only slightly reformatted as the complete task setup is now done here instead of inside a custom plugin. The only real change is replacing `tasks.cloneHederaProtobufs` with  `tasks.register<GitClone>("cloneHederaProtobufs")`.

.github/dependabot.yml
```diff
@@ -11,7 +11,12 @@ updates:
...
+    directory: /hiero-dependency-versions
     schedule:
-      interval: "weekly"
\ No newline at end of file
+      interval: weekly
+
+  - package-ecosystem: gradle
+    directory: /
+    schedule:
+      interval: weekly
```
ata-nas, line 22: File does not end with an newline char.

.github/dependabot.yml
ata-nas, line 1: Should this file also have the SPDX license header?

gradle.properties
ata-nas, line 1: Should this file also have the SPDX license header?

gradle/toolchain-versions.properties
```diff
@@ -0,0 +1 @@
+jdk=21.0.4
```
ata-nas, line 1: Shouldn't this jdk version be `21.0.5`?

gradle/toolchain-versions.properties
ata-nas, line 1: Should this file also have the SPDX license header?

gradle/wrapper/gradle-wrapper.properties
ata-nas, line 1: Should this file also have the SPDX license header?

server/src/main/java/module-info.java
ata-nas, line 1: Should this file also have the SPDX license header?

simulator/src/main/java/module-info.java
ata-nas, line 1: Should this file also have the SPDX license header?

stream/src/main/java/module-info.java
ata-nas, line 1: Should this file also have the SPDX license header?

suites/build.gradle.kts
```diff
@@ -1,27 +1,42 @@
...
+
+// Task to build the Docker image
+tasks.register<Exec>("createDockerImage") {
+    description = "Creates the Docker image of the Block Node Server based on the current version"
+    group = "docker"
+
+    dependsOn(updateDockerEnv, tasks.assemble)
+    workingDir(layout.projectDirectory.dir("../server/docker"))
+    commandLine("./docker-build.sh", project.version, layout.projectDirectory.dir("..").asFile)
+}
```
ata-nas, line 39: Why was this added here?

suites/src/main/java/module-info.java
ata-nas, line 1: Should this file also have the SPDX license header?

version.txt
ata-nas, line 1: Should this file also have the SPDX license header?

ata-nas (commented):
@jjohannes LG. A couple of nits/questions.

Just recently, a migration to SPDX license headers happened. Maybe some headers were missed, I am unsure which or if all files are eligible to have this header. Please feel free to proactively resolve my comments if files, where I have left a comment for the header, should not actually have the header. (My comments are mostly to bring attention in case some headers might be missed)

gradle.properties
jjohannes: In my opinion we do not need that for properties files.

But if, it should also be automated through `spotlessApply`. In this PR, I would prefer to not add it. But we can do it in a follow up, if it is desired.

gradle/toolchain-versions.properties
jjohannes: This is moved from here: https://github.com/hashgraph/hedera-block-node/blob/main/buildSrc/src/main/kotlin/com.hedera.block.conventions.gradle.kts#L20

Updating to Java `21.0.5` should be done in a follow up:  https://github.com/hashgraph/hedera-block-node/issues/453

gradle/toolchain-versions.properties
jjohannes: See my comment on `gradle.properties`.

gradle/wrapper/gradle-wrapper.properties
jjohannes: No. This file is generated autoamtically by Gradle.

version.txt
jjohannes: No. This is only a number and comments are not allowed in this file. (The code we have that parses and automatically updates the file does not support comments atm.)

suites/build.gradle.kts
jjohannes: Good catch! This is a mistake. I don't know how this happened. Will remove.

jjohannes:
Thanks @ata-nas. Addressed the comments and/or answered. Note that the headers I added manually now will also be handled by `spotlessApply` in the near future. Related:

- https://github.com/hiero-ledger/hiero-gradle-conventions/issues/22
- https://github.com/hiero-ledger/hiero-gradle-conventions/issues/30#issuecomment-2580033004

gradle.properties
ata-nas: @AlfredoG87 @mattp-swirldslabs @georgi-l95 @jsync-swirlds @jasperpotts tag for awareness about the headers. I will resolve this comment from my side.

ata-nas (approved):
@jjohannes looks good! Before merging please see also if others have other comments (I remember people asked in the near past to wait with merging this for a bit, since they needed to change things which would conflict with this, I am not sure if everybody is ready) @AlfredoG87 @mattp-swirldslabs @jsync-swirlds @georgi-l95 @jasperpotts tag for awareness.

gradlew
```diff
@@ -15,6 +15,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+# SPDX-License-Identifier: Apache-2.0
+#
```
jsync-swirlds, line 19: This is not our code to relicense, I don't think...
It's also duplicative with the block immediately prior.

gradlew.bat
```diff
@@ -13,6 +13,8 @@
 @rem See the License for the specific language governing permissions and
 @rem limitations under the License.
 @rem
+@rem SPDX-License-Identifier: Apache-2.0
+@rem
```
jsync-swirlds, line 17: Same here, adding SPDX to gradle-owned code.

gradlew
jjohannes: This is not our doing. It is done by Gradle itself. It was recently added: https://github.com/gradle/gradle/pull/28801

It's in this PR because I did a full upgrade by running `./gradlew wrapper` with the latest versions.

gradlew.bat
jjohannes: Same as with `gradlew` script

gradlew
jsync-swirlds: Ahh, I assumed it was our change.
If gradle is making the change, then that's fine.

andrewb1269hg (approved):
Review and approve following files:
`.github/dependabot.yml`
`.github/workflows/pr-checks.yaml`

PR #226, opened 2024-10-02
refactor: simulator entry point to be flexible as test driver

### LINKED ISSUE #190: Modify Simulator entry point to be flexible as test driver
### Problem
Currently starting Simulator is only possible using the main method and without easy way to modify configuration or behaviour on the fly. That will create issues and obstacles, if we want to use it as a test driver for our E2E tests.

### Solution
We'd need to modify the Simulator entry point, so that not only can be started using the main method, but via object instantiation and calling start method. We'd also want to be able to change configuration and more.

georgi-l95:
**Description**:
This PR aims to refactor the simulator entry point and add flexabilty, in terms of how it can be started.
Adding two new constructors and moving the dagger initialization in the applicaiton itself, rather being in the main method.
This allows us to use it easy in the suites module, where we have the E2E tests.

**Related issue(s)**:

Fixes #190 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
@@ -16,6 +16,8 @@
 
 package com.hedera.block.simulator;
 
+import static com.hedera.block.simulator.Constants.CONFIGURATION_FILE;
+
 import com.hedera.block.simulator.config.data.BlockStreamConfig;
 import com.hedera.block.simulator.generator.BlockStreamManager;
 import com.hedera.block.simulator.grpc.PublishStreamGrpcClient;
@@ -28,21 +30,18 @@ import com.swirlds.config.extensions.sources.SystemEnvironmentConfigSource;
 import com.swirlds.config.extensions.sources.SystemPropertiesConfigSource;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
-import javax.inject.Inject;
-import java.io.IOException;
 import java.nio.file.Path;
 import javax.inject.Inject;
 
 /** BlockStream Simulator App */
 public class BlockStreamSimulatorApp {
 
-    private static final System.Logger LOGGER = System.getLogger(BlockStreamSimulatorApp.class.getName());
+    private static final System.Logger LOGGER =
+            System.getLogger(BlockStreamSimulatorApp.class.getName());
 
-    private final Configuration configuration;
-    private BlockStreamManager blockStreamManager;
-    private PublishStreamGrpcClient publishStreamGrpcClient;
-    private BlockStreamConfig blockStreamConfig;
-
+    private final BlockStreamManager blockStreamManager;
+    private final PublishStreamGrpcClient publishStreamGrpcClient;
+    private final BlockStreamConfig blockStreamConfig;
     private final int delayBetweenBlockItems;
 
     private boolean isRunning = false;
@@ -50,75 +49,64 @@ public class BlockStreamSimulatorApp {
     /**
      * Creates a new BlockStreamSimulatorApp instance.
      *
-     * @param configuration           the configuration to be used by the block
-     *                                stream simulator
-     * @param blockStreamManager      the block stream manager to be used by the
-     *                                block stream simulator
-     * @param publishStreamGrpcClient the gRPC client to be used by the block stream
-     *                                simulator
+     * @param configuration the configuration to be used by the block stream simulator
+     * @param blockStreamManager the block stream manager to be used by the block stream simulator
+     * @param publishStreamGrpcClient the gRPC client to be used by the block stream simulator
      */
     @Inject
     public BlockStreamSimulatorApp(
             @NonNull Configuration configuration,
             @NonNull BlockStreamManager blockStreamManager,
             @NonNull PublishStreamGrpcClient publishStreamGrpcClient) {
-        this.configuration = configuration;
         this.blockStreamManager = blockStreamManager;
         this.publishStreamGrpcClient = publishStreamGrpcClient;
 
         blockStreamConfig = configuration.getConfigData(BlockStreamConfig.class);
-
         delayBetweenBlockItems = blockStreamConfig.delayBetweenBlockItems();
     }
 
     /**
-     * Constructs a new {@code BlockStreamSimulatorApp} with the specified
-     * configuration.
+     * Constructor that accepts a custom configuration and initializes dependencies internally.
      *
-     * @param configuration the configuration to use for the simulator app
+     * @param configuration the custom configuration to use
      */
     public BlockStreamSimulatorApp(Configuration configuration) {
-        this.configuration = configuration;
-        initilizeDependencies();
-        initializeConfig();
-
-        this.delayBetweenBlockItems = blockStreamConfig.delayBetweenBlockItems();
+        this(configuration, createComponent(configuration));
     }
 
     /**
-     * Constructs a new {@code BlockStreamSimulatorApp} with the default
-     * configuration.
+     * Private constructor that initializes the class using a configuration and a Dagger component.
      *
-     * @throws IOException if an I/O error occurs while loading the default
-     *                     configuration
+     * @param configuration the configuration to use
+     * @param component the Dagger component for dependency injection
+     */
+    private BlockStreamSimulatorApp(
+            Configuration configuration, BlockStreamSimulatorInjectionComponent component) {
+
+        this(
+                configuration,
+                component.getBlockStreamManager(),
+                component.getPublishStreamGrpcClient());
+    }
+
+    /**
+     * Default constructor that loads the default configuration and initializes dependencies.
+     *
+     * @throws IOException if an I/O error occurs while loading the default configuration
      */
     public BlockStreamSimulatorApp() throws IOException {
-        this.configuration = loadDefaultConfiguration();
-        initilizeDependencies();
-        initializeConfig();
-
-        this.delayBetweenBlockItems = blockStreamConfig.delayBetweenBlockItems();
+        this(loadDefaultConfiguration());
     }
 
     /**
-     * Initializes the configuration by retrieving the {@code BlockStreamConfig}
-     * from the current
-     * configuration.
+     * Helper method to create the Dagger component with the given configuration.
+     *
+     * @param configuration the configuration to use
+     * @return the Dagger component
      */
-    private void initializeConfig() {
-        this.blockStreamConfig = configuration.getConfigData(BlockStreamConfig.class);
-    }
-
-    /**
-     * Initializes the dependencies required by the simulator app using Dagger
-     * dependency injection.
-     */
-    private void initilizeDependencies() {
-        BlockStreamSimulatorInjectionComponent DIComponent = DaggerBlockStreamSimulatorInjectionComponent.factory()
-                .create(configuration);
-
-        this.blockStreamManager = DIComponent.getBlockStreamManager();
-        this.publishStreamGrpcClient = DIComponent.getPublishStreamGrpcClient();
+    private static BlockStreamSimulatorInjectionComponent createComponent(
+            Configuration configuration) {
+        return DaggerBlockStreamSimulatorInjectionComponent.factory().create(configuration);
     }
 
     /**
@@ -127,12 +115,13 @@ public class BlockStreamSimulatorApp {
      * @return the default configuration
      * @throws IOException if an I/O error occurs while loading the configuration
      */
-    private Configuration loadDefaultConfiguration() throws IOException {
-        ConfigurationBuilder configurationBuilder = ConfigurationBuilder.create()
-                .withSource(SystemEnvironmentConfigSource.getInstance())
-                .withSource(SystemPropertiesConfigSource.getInstance())
-                .withSource(new ClasspathFileConfigSource(Path.of("app.properties")))
-                .autoDiscoverExtensions();
+    private static Configuration loadDefaultConfiguration() throws IOException {
+        ConfigurationBuilder configurationBuilder =
+                ConfigurationBuilder.create()
+                        .withSource(SystemEnvironmentConfigSource.getInstance())
+                        .withSource(SystemPropertiesConfigSource.getInstance())
+                        .withSource(new ClasspathFileConfigSource(Path.of(CONFIGURATION_FILE)))
+                        .autoDiscoverExtensions();
         return configurationBuilder.build();
     }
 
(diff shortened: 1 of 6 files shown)


simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -30,14 +36,14 @@ public class BlockStreamSimulatorApp {
...
             System.getLogger(BlockStreamSimulatorApp.class.getName());
 
-    Configuration configuration;
-    BlockStreamManager blockStreamManager;
-    PublishStreamGrpcClient publishStreamGrpcClient;
-    BlockStreamConfig blockStreamConfig;
+    private final Configuration configuration;
+    private BlockStreamManager blockStreamManager;
+    private PublishStreamGrpcClient publishStreamGrpcClient;
+    private BlockStreamConfig blockStreamConfig;
```
ata-nas, line 42: IMO, these should become final. They are only instantiated in constructors, which have the same logic. We could use constructor overloading, which is always a good idea. Also, the final keyword not only makes our code more strict, but actually improves performance by making our field immutable and there are hints to the compiler/JVM which work like magic to make our programs faster. Small things do compound and make a difference.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -30,14 +36,14 @@ public class BlockStreamSimulatorApp {
...
 
     private final int delayBetweenBlockItems;
 
-    boolean isRunning = false;
+    private boolean isRunning = false;
```
ata-nas, line 46: Just a comment for the future, since this has no usage now.
I have a very strong feeling that this field would be used to stop the app, which would very possibly be done from another thread. If that would be the case, this field should become volatile in order to have the visibility (flush the data beyond CPU cache in order for other threads to see it) and the so called 'happens before guarantees'. Without that, we can expect that this flag will not function properly. Also, if we are going to change the data of the flag in a non atomic fashion, then we would need to lock.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -60,6 +66,67 @@ public BlockStreamSimulatorApp(
...
+     *
+     * @return the default configuration
+     * @throws IOException if an I/O error occurs while loading the configuration
+     */
+    private Configuration loadDefaultConfiguration() throws IOException {
+        ConfigurationBuilder configurationBuilder =
+                ConfigurationBuilder.create()
+                        .withSource(SystemEnvironmentConfigSource.getInstance())
+                        .withSource(SystemPropertiesConfigSource.getInstance())
+                        .withSource(new ClasspathFileConfigSource(Path.of("app.properties")))
```
ata-nas, line 125: consider to extract the "app.properties" literal into a class of constants so that we can have a global place for referring to these literals which I have no doubt we will see again.

simulator/src/main/java/module-info.java
```diff
@@ -3,6 +3,7 @@
 /** Runtime module of the simulator. */
 module com.hedera.block.simulator {
     exports com.hedera.block.simulator.config.data;
+    exports com.hedera.block.simulator;
```
ata-nas, line 6: IntelliJ warns me when I use the classes after this export here.

Here some reference: https://stackoverflow.com/questions/54755914/java-10-intellij-warning-class-is-not-exported-accessing-in-same-module

I am not sure how relevant this is, maybe we should check it out?

ata-nas (changes requested):
Not a blocker, but please consider to use the `final` keyword on all places including local variables (inside methods) and method parameters as well. It is best that we are as strict as possible by default. One of the best reasons to use `final` for parameters and local vars is so that we do not accidentally change the reference, which could lead to silent bugs. With `final` if we try to change the ref, we get a compilation error.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
ata-nas, line 42: IMO, these should become final. They are only instantiated in constructors, which have the same logic. We could use constructor overloading, which is always a good idea. Also, the final keyword not only makes our code more strict, but actually improves performance by making our field immutable and there are hints to the compiler/JVM which work like magic to make our programs faster. Small things do compound and make a difference.

I would strongly suggest to overload the constructors and make the fields final.

P.S. with constructor overloading we will also eliminate the code duplication.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
ata-nas, line 46: Just a comment for the future, since this has no usage now.

I have a very strong feeling that this field would be used to stop the app, which would very possibly be done from another thread. If that would be the case, this field should become volatile in order to have the visibility (flush the data beyond CPU cache in order for other threads to see it) and the so called 'happens before guarantees'. Without that, we can expect that this flag will not function properly. Also, if we are going to change the data of the flag in a non atomic fashion, then we would need to lock.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
ata-nas, line 125: Consider to extract the "app.properties" literal into a class of constants so that we can have a global place for referring to these literals which I have no doubt we will see again.

simulator/src/main/java/module-info.java
ata-nas, line 6: IntelliJ warns me when I use the classes after this export here.

Here some reference: https://stackoverflow.com/questions/54755914/java-10-intellij-warning-class-is-not-exported-accessing-in-same-module

I am not sure how relevant this is, maybe we should check it out?

ata-nas (changes requested):
Not a blocker, but please consider to use the final keyword on all places including local variables (inside methods) and method parameters as well. It is best that we are as strict as possible by default. One of the best reasons to use final for parameters and local vars is so that we do not accidentally change the reference, which could lead to silent bugs. With final if we try to change the ref, we get a compilation error.

simulator/src/main/java/module-info.java
georgi-l95: Removed, will be needed in my following PR thou. 👍

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
georgi-l95: Done.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
georgi-l95: True, coudn't agree more. This variable will go through it's fare of changes, however at this point I'm not 100% what role it will play. And also it's not part of this PR scope. So I'm leaving it like this for now, but as you noted, we will change it in the future. 👍

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
georgi-l95: Done.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
ata-nas: Yes, it is completely out of scope, that is why I left this comment for the future. :) Leave like this, resolved for me. Maybe I should have marked it as resolved since the beginning.

ata-nas (approved):
Resolved for me.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -53,15 +58,73 @@ public BlockStreamSimulatorApp(
...
+    public BlockStreamSimulatorApp(Configuration configuration) {
+        this(configuration, createComponent(configuration));
+    }
+
+    /**
+     * Private constructor that initializes the class using a configuration and a Dagger component.
+     *
+     * @param configuration the configuration to use
+     * @param component the Dagger component for dependency injection
+     */
```
mattp-swirldslabs, line 82: No need to javadoc private methods

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -53,15 +58,73 @@ public BlockStreamSimulatorApp(
...
+    }
+
+    /**
+     * Default constructor that loads the default configuration and initializes dependencies.
+     *
+     * @throws IOException if an I/O error occurs while loading the default configuration
+     */
+    public BlockStreamSimulatorApp() throws IOException {
+        this(loadDefaultConfiguration());
+    }
```
mattp-swirldslabs, line 99: This class is using Dagger DI. Shouldn't the callers just @Inject this as well? Why do we need additional constructors?

AlfredoG87 (approved):
LGTM

simulator/src/test/java/com/hedera/block/simulator/BlockStreamSimulatorTest.java
```diff
@@ -92,10 +92,24 @@ void start_exitByBlockNull() throws InterruptedException, ParseException, IOExce
         assertTrue(blockStreamSimulator.isRunning());
     }
 
+    void start_usingConfigurationConstructor()
+            throws InterruptedException, ParseException, IOException {
+        blockStreamSimulator = new BlockStreamSimulatorApp(configuration);
+        blockStreamSimulator.start();
+        assertTrue(blockStreamSimulator.isRunning());
+    }
```
mattp-swirldslabs, line 100: This ^^ method isn't used and it's not annotated with @Test so it's never called

simulator/src/test/java/com/hedera/block/simulator/BlockStreamSimulatorTest.java
```diff
@@ -92,10 +92,24 @@ void start_exitByBlockNull() throws InterruptedException, ParseException, IOExce
         assertTrue(blockStreamSimulator.isRunning());
     }
 
+    void start_usingConfigurationConstructor()
+            throws InterruptedException, ParseException, IOException {
+        blockStreamSimulator = new BlockStreamSimulatorApp(configuration);
```
mattp-swirldslabs, line 97: Perhaps use local variables rather than member variables to ensure deterministic behavior ^^

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -53,15 +58,73 @@ public BlockStreamSimulatorApp(
...
+    /**
+     * Loads the default configuration for the simulator app.
+     *
+     * @return the default configuration
+     * @throws IOException if an I/O error occurs while loading the configuration
+     */
+    private static Configuration loadDefaultConfiguration() throws IOException {
+        ConfigurationBuilder configurationBuilder =
+                ConfigurationBuilder.create()
+                        .withSource(SystemEnvironmentConfigSource.getInstance())
```
mattp-swirldslabs, line 121: This code looks like it's from `Server.java`. We're going to change how we bring in environment variables. Perhaps coordinate with Atanas to make the same changes here

mattp-swirldslabs (commented):
I have some questions about DI versus new constructors

simulator/src/main/java/com/hedera/block/simulator/Constants.java
```diff
@@ -18,8 +18,12 @@
 
 public class Constants {
```
mattp-swirldslabs, line 20: When I execute a build I see these warnings:
```
> Task :simulator:javadoc
/Users/mattpeterson/repos/block-node-secondary/hedera-block-node/simulator/src/main/java/com/hedera/block/simulator/Constants.java:19: warning: no comment
public class Constants {
       ^
/Users/mattpeterson/repos/block-node-secondary/hedera-block-node/simulator/src/main/java/com/hedera/block/simulator/Constants.java:19: warning: use of default constructor, which does not provide a comment
public class Constants {
       ^
2 warnings
```

mattp-swirldslabs: The `Constants` class should be final and have a private default constructor

simulator/src/test/java/com/hedera/block/simulator/BlockStreamSimulatorTest.java
```diff
@@ -92,10 +92,24 @@ void start_exitByBlockNull() throws InterruptedException, ParseException, IOExce
...
+
     private String getAbsoluteFolder(String relativePath) {
         return Paths.get(relativePath).toAbsolutePath().toString();
     }
 
+    @Test
+    void start_usingEmptyConstructor() throws IOException, InterruptedException, ParseException {
+        blockStreamSimulator = new BlockStreamSimulatorApp();
+        blockStreamSimulator.start();
+        assertTrue(blockStreamSimulator.isRunning());
```
mattp-swirldslabs, line 110: Maybe we can expand these tests using mocks, etc to check that get default configuration is actually present and correct, test the `stop()` method...

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
georgi-l95: True, out of habit. Will remove. ✅

simulator/src/test/java/com/hedera/block/simulator/BlockStreamSimulatorTest.java
georgi-l95: Good catch.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
georgi-l95: There can only be one injected constructor. I was thinking about, deleting the other constructors and keeping only the empty one (in terms of passed params). For changing configuration, we could externalize a method to do this at runtime, that would actually aslo benefit as test driver.

mattp-swirldslabs: Ok, so the DI constructor will only be used by the `BlockStreamSimulator` and that's basically using the builder anyway. Maybe we shouldn't use DI for this part?

simulator/src/main/java/com/hedera/block/simulator/Constants.java
ata-nas: @mattp-swirldslabs I will be abstracting all constants and utility classes in the new module based on task #233 which is in progress, so I think it is fine that it will remain like this since it will be changed very soon.

georgi-l95:
Closing in favor of https://github.com/hashgraph/hedera-block-node/pull/242

simulator/src/main/java/com/hedera/block/simulator/Constants.java
mattp-swirldslabs: Ok, but we should be checking to make sure there aren't any "warnings" when we build

PR #232, opened 2024-10-04
refactor: adjust configuration management and smoke test setup to support production config by default

### LINKED ISSUE #185: refactor: adjust configuration management and smoke test setup to support production config by default
As a Block Node administrator
I want production configuration by default
so that I can minimize dealing with app.properties changes

* Currently, `smoke_test.yaml` runs the smoke tests using `gradle run` to verify with `smoke_test.sh`
* It's more difficult to override the default configuration because this does not use a test app.properties file when it starts the Block Node
* Record files like: `MediatorConfig.java` and `NotifierConfig.java` currently must be defaulted to non-production values (e.g. 1024 for the ring buffer rather than a more realistic prod value like 67108864) so that the smoke test does not throw an OutOfMemory error.
* As a result, the production `app.properties` file must contain the prod values instead to override the defaults when running in a container.

Perhaps we should change the CI tests to:
1) Run the smoke tests via docker rather than gradle. Gradle is strictly used for build and development.
2) Set the defaults in the config record files like `MediatorConfig.java` and `NotifierConfig.java` to the production defaults
3) Remove the override values in `app.properties` to minimize the configuration burden.
4) Leverage the test `app.properties` inside the container noted in #1 to override the production defaults with test config.

ata-nas:
**Description**:

- make the default values in config records to be the production ones;
- the app.properties file should not be setting the production values, by default it should be empty, we should only set props there if we want to override the production defaults;
- when executing smoke tests, the config records should be with test values;
- explore options to achieve the result: **(1)** use env variables - does not seem to work as expected, the config dependency gets the vars via `System.getenv()` and it only shows case-sensitive keys, so adding a value like `mediator.ringBufferSize=1024` will result into not being visible in runtime, it seems to me to be inconsistent. **(2)** build the jar for the smoke tests with the `app.properties` from `src/test/resources` - this works, the file gets packaged into the jar, only hurdle currently is to find a good and reliable way to do it with a `gradle` task.
- a single `gradle` task execution should be sufficient to build and run the needed environment for the test to pass

**UPDATE:**
- after some discussions with @mattp-swirldslabs & @AlfredoG87, we have discussed some issues that we have discovered:
    1. environment variables that are set to the docker container are read only if they are all upper case and separated only via `_`.;
    1. the initial idea of making a single `gradle` task (see some previous commits in this PR) that will build the project with correct test props, create the docker image and run it (so we can have an easy single entry point) did not seem to work when I tried it, since when in `gh actions` I execute either a shell script or the `gradle` task that contains a shell invocation of the `docker` executable, `gh actions` failed with exit code 14 (insufficient privileges). After some research, I see that I probably need to use the `docker` executable directly in the `run` part of the step. That worked and the test passed after commiting. So this part is still debatable, can I or can I not use the `docker` executable from a shell script.;

- since this discussions, there are several discoveries:
    1. using the `IntelliJ` IDE, if we modify the run config to include an environment variable with lowercase and `.` separator, it gets discovered. I am still to understand how that works and why, it does not seem to get passed as an CLI arg when starting the jar, so the IDE is doing something else. But that only means that it is indeed possible to discover non-conventional keys for environment variables.;
    1. as an approach, we could use `PropertyFileConfigSource.class` and read vars directly from a file (this is different from `ClasspathFileConfigSource.java` where we technically still load a file `app.properties` but this is packaged into the jar, here we are talking about an arbitrary file outside the `classpath`), this worked and is a potential approach. However, it has the lowest precedence, so if the same key comes from basically any other source, this will get overridden, we need to keep that in mind. Ordinals could be set in the constructor, so we can decide precedence. (see `ConfigSourceOrdinalConstants.java`);

- final approach (or the usage of multiple approaches so more could be tested if decided so) is still under debate;

**UPDATE:**
- after a discussion with @jsync-swirlds, he noted that we will be able to support by default unconventional casing for environment variables (meaning an environment variable like `mediator.ringBufferSize=1024` will be picked up without having to do anything special). Until then, we can use the `MappedConfigSource` in order to support the mapping between uppercase variables and the formate we need them (for example the env variable `MEDIATOR_RING_BUFFER_SIZE=1024` will be available with that key as well as with this one `mediator.ringBufferSize=1024`). For now, we can only pick up uppercase variables with an `_` separator, but with the mapping, we will be able to read the values with the mapped key.

**Related issue(s)**:

Fixes #185 

**Notes for reviewer**:

Tested locally:
- Start the `jar` directly and use `app.properties`
- Start the `jar` directly and use `env` vars
- Start the `docker` container and use `app.properties`
- Start the `docker` container and use `env` vars
- Start the `smoke-test` action locally and use `app.properties`
- Start the `smoke-test` action locally and use `env` vars
- Start `k8s` cluster with the container and use `app.properties`
- Start `k8s` cluster with the container and use `env` vars
- Started all of the above with both `app.properties` and `env` vars, expected that the `env` vars take precedence (as is by default and we have not changed the default precedence)

**Checklist**

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -65,7 +65,6 @@ jobs:
           sudo tar -xzf grpcurl.tar.gz -C /usr/local/bin grpcurl
           rm grpcurl.tar.gz
 
-
       - name: Cache Gradle packages
         uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
         with:
@@ -76,18 +75,12 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
 
-      - name: Build application
-        run: ${{ env.GRADLE_EXEC }} build
+      - name: Build And Run Smoke Tests Container
+        run: ${{ env.GRADLE_EXEC }} buildAndRunSmokeTestsContainer -x test
 
-      - name: Run application in background, capture logs in app.log
-        run: |
-          ${{ env.GRADLE_EXEC }} :server:run 2> server/src/test/resources/app.log < /dev/null &
-          echo "Application started with PID $APP_PID"
-          sleep 10
-
-      - name: Print App Logs
-        run: cat server/src/test/resources/app.log
-
-      - name: Smoke Test
+      - name: Run Smoke Test
         working-directory: server/src/test/resources/
-        run: ./smoke-test.sh app.log
+        run: ./smoke-test.sh
+
+      - name: Stop Smoke Tests Container
+        run: docker compose -p block-node stop
--- a/server/docker/update-env.sh
+++ b/server/docker/update-env.sh
@@ -4,19 +4,33 @@
 # This script is called by gradle and get the current project version as an input param
 
 if [ $# -lt 1 ]; then
-  echo "USAGE: $0 <VERSION> <DEBUG>"
+  echo "USAGE: $0 <VERSION> <DEBUG> <SMOKE_TEST>"
   exit 1
 fi
 
-echo "VERSION=$1" > .env
+project_version=$1
+# determine if we should include debug opts
+[ "$2" = true ] && is_debug=true || is_debug=false
+# determine if we should include smoke test env variables
+[ "$3" = true ] && is_smoke_test=true || is_smoke_test=false
+
+echo "VERSION=$project_version" > .env
 echo "REGISTRY_PREFIX=" >> .env
 # Storage root path, this is temporary until we have a proper .properties file for all configs
 echo "BLOCKNODE_STORAGE_ROOT_PATH=/app/storage" >> .env
 echo "JAVA_OPTS='-Xms8G -Xmx16G'" >> .env
 
-if [ $# -eq 2 ]; then
+if [ true = "$is_debug" ]; then
+  # wait for debugger to attach
   echo "SERVER_OPTS='-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005'" >> .env
 fi
 
-echo "DEBUG $2"
-echo "VERSION/TAG UPDATED TO $1"
+if [ true = "$is_smoke_test" ]; then
+  # add smoke test variables
+  echo "MEDIATOR_RING_BUFFER_SIZE=1024" >> .env
+  echo "NOTIFIER_RING_BUFFER_SIZE=1024" >> .env
+fi
+
+# Output the values
+echo ".env properties:"
+cat .env
--- a/server/src/main/java/com/hedera/block/server/Server.java
+++ b/server/src/main/java/com/hedera/block/server/Server.java
@@ -22,10 +22,10 @@ import static io.helidon.config.ConfigSources.file;
 import static java.lang.System.Logger;
 import static java.lang.System.Logger.Level.INFO;
 
+import com.hedera.block.server.config.ServerMappedConfigSourceInitializer;
 import com.swirlds.config.api.Configuration;
 import com.swirlds.config.api.ConfigurationBuilder;
 import com.swirlds.config.extensions.sources.ClasspathFileConfigSource;
-import com.swirlds.config.extensions.sources.SystemEnvironmentConfigSource;
 import com.swirlds.config.extensions.sources.SystemPropertiesConfigSource;
 import io.helidon.config.Config;
 import java.io.IOException;
@@ -59,7 +59,7 @@ public class Server {
         // Init BlockNode Configuration
         final Configuration configuration =
                 ConfigurationBuilder.create()
-                        .withSource(SystemEnvironmentConfigSource.getInstance())
+                        .withSource(ServerMappedConfigSourceInitializer.getMappedConfigSource())
                         .withSource(SystemPropertiesConfigSource.getInstance())
                         .withSources(new ClasspathFileConfigSource(Path.of(APPLICATION_PROPERTIES)))
                         .autoDiscoverExtensions()
--- a/server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
+++ b/server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.block.server.config;
+
+import com.swirlds.config.extensions.sources.ConfigMapping;
+import com.swirlds.config.extensions.sources.MappedConfigSource;
+import com.swirlds.config.extensions.sources.SystemEnvironmentConfigSource;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.List;
+
+/**
+ * A class that extends {@link MappedConfigSource} ir order to have project-relevant initialization.
+ */
+public final class ServerMappedConfigSourceInitializer {
+    private static final List<ConfigMapping> MAPPINGS =
+            List.of(
+                    new ConfigMapping(
+                            "consumer.timeoutThresholdMillis", "CONSUMER_TIMEOUT_THRESHOLD_MILLIS"),
+                    new ConfigMapping(
+                            "persistence.storage.rootPath", "PERSISTENCE_STORAGE_ROOT_PATH"),
+                    new ConfigMapping("service.delayMillis", "SERVICE_DELAY_MILLIS"),
+                    new ConfigMapping("mediator.ringBufferSize", "MEDIATOR_RING_BUFFER_SIZE"),
+                    new ConfigMapping("notifier.ringBufferSize", "NOTIFIER_RING_BUFFER_SIZE"));
+
+    private ServerMappedConfigSourceInitializer() {}
+
+    /**
+     * This method constructs, initializes and returns a new instance of {@link MappedConfigSource}
+     * which internally uses {@link SystemEnvironmentConfigSource} and maps relevant config keys to
+     * other keys so that they could be used within the application
+     *
+     * @return newly constructed fully initialized {@link MappedConfigSource}
+     */
+    @NonNull
+    public static MappedConfigSource getMappedConfigSource() {
+        final MappedConfigSource config =
+                new MappedConfigSource(SystemEnvironmentConfigSource.getInstance());
+        MAPPINGS.forEach(config::addMapping);
+        return config;
+    }
+}
--- a/server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
+++ b/server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.block.server.config;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.swirlds.config.extensions.sources.ConfigMapping;
+import com.swirlds.config.extensions.sources.MappedConfigSource;
+import java.lang.reflect.Field;
+import java.util.Queue;
+import java.util.function.Predicate;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class ServerMappedConfigSourceInitializerTest {
+    private static final ConfigMapping[] SUPPORTED_MAPPINGS = {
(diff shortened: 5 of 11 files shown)


suites/build.gradle.kts
```diff
@@ -35,6 +35,21 @@ mainModuleInfo {
...
+// what will be run from repo root w/o having qualifiers. For these building tasks are probably not
+// going to mess up some common files between them, it is a concern for the future, since not clearly
+// knowing what is being run could lead to silent bugs that are hard to come by. Not to mention, if
+// we have a task that will run a project, and then a task with the same name that will also run a
+// project, now when executing from root we will run two projects which might be problematic.
+// Having qualifiers is a 'need-to-know' solution which is not preferable. But there is also the case
+// that it would be a good idea to have tasks with the same name that we know for sure run isolated,
+// and we need them to run on all projects easily, so running the task from repo root will execute
+// everywhere it is present. So the question is should these remain with the same name and what
+// should be our approach for the future?
```
ata-nas, line 51: Let's discuss for another refactor. See comment.

.github/workflows/smoke-test.yaml
```diff
@@ -77,17 +76,22 @@ jobs:
...
+        run: |
+          docker buildx build --load -t "block-node-server:0.1.0-SNAPSHOT" --build-context distributions=../build/distributions --build-arg VERSION="0.1.0-SNAPSHOT" . || exit "123"
+          docker compose -p block-node up -d
 
       - name: Smoke Test
         working-directory: server/src/test/resources/
-        run: ./smoke-test.sh app.log
+        run: ./smoke-test.sh
+
+      # TODO Do we need to stop this container? In theory when the job runs, everything gets scrapped
```
mattp-swirldslabs, line 95: I think we should stop the container and also check that the exit code is 0. I've seen a number of defects in the past which surface when shutting down.

ata-nas: Noted.

.github/workflows/smoke-test.yaml
```diff
@@ -76,18 +75,28 @@ jobs:
...
 
-      - name: Print App Logs
-        run: cat server/src/test/resources/app.log
+      # TODO if I use 'docker compose' from a shell script (meaning I call a shell scrip here instead)
+      # I get permission denied and it exits (which makes sense). Locally with the tool act I am able
+      # to reproduce it as well. If I use 'docker-compose' from a shell script however, locally with
+      # act it seems to pass, so maybe will pass in gh actions as well, but when I pushed, it exited
+      # saying 'docker-compose not found'. In theory maybe we can have 'docker-compose' available
+      # in gh actions so we can have a single gradle script that will do all the job and we will
+      # not need all these steps that will use docker directly (which was the initial idea)
```
ata-nas, line 90: Could we address that? See comment and past runs of actions. If I could address that so either the `docker` or `docker-compose` executables work, I could implement the initial idea of centralizing all the logic into a single `gradle` task.

server/src/main/java/com/hedera/block/server/Server.java
```diff
@@ -59,7 +60,9 @@ public static void main(final String[] args) throws IOException {
         // Init BlockNode Configuration
         final Configuration configuration =
                 ConfigurationBuilder.create()
-                        .withSource(SystemEnvironmentConfigSource.getInstance())
+                        .withSource(
+                                new BlockNodeMappedConfigSource(
+                                        SystemEnvironmentConfigSource.getInstance()))
```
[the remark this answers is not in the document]
ata-nas: Noted. 

When I was researching your proposed solution and was trying to understand how the `MappedConfigSource` works, I stumbled on some usages and indeed found what you are saying, to use the class directly [(see here)](https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/internal/ConfigMappings.java), but thought that instead of doing it with a utility class and since the `MappedConfigSource` is not final, I thought that inheritance would package the initialization of the configs in a good way. In any case I think we need to do some maintenance since we need to provide the mappings for what we want to support as env variables that need a different key at runtime. In theory this of course could be done dynamically, but I see no point in doing so, increased complexity and redundant operations to map keys that we would not use anyway. That being said, I think that it is a good idea to somehow do a test that would fail if we introduce a new config class and do not extend the Junit test that would check that all the mappings are correct. 

So let me know what you think, I could switch to a utility class and use the `MappedConfigSource` directly, or I could leave like that. Thanks!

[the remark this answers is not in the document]
ata-nas: After a conversation with @jsync-swirlds, he noted that for the functionality we are going to need to use the `MappedConfigSource`, it is planned to be supported by default. So in reality we are just making a workaround. That being said, I agree completely that making the new class is completely redundant, since we are expecting that it will be removed in the future. Will fall back to using `MappedConfigSource` directly and make some tests.

georgi-l95 (commented):
I know it's still on draft, but so far it looks good and I don't have any questions or nits 👍

georgi-l95 (dismissed):
LGTM

server/docker/update-env.sh
```diff
@@ -4,19 +4,34 @@
...
-  echo "USAGE: $0 <VERSION> <DEBUG>"
+  echo "USAGE: $0 <VERSION> <DEBUG> <SMOKE_TEST>"
   exit 1
 fi
 
-echo "VERSION=$1" > .env
+project_version=$1
+# in cases where no params are supplied for debug or smoke tests, later we still need to log something
+is_debug=$([ "$2" = "true" ] && echo "true" || echo "false")
+is_smoke_test=$([ "$3" = "true" ] && echo "true" || echo "false")
```
mattp-swirldslabs, line 14: Can we set `is_debug` and `is_smoke_test` to boolean values rather than strings?

server/docker/update-env.sh
```diff
@@ -4,19 +4,34 @@
...
+if [ true = "$is_smoke_test" ]; then
+  # add smoke test variables
+  echo "MEDIATOR_RING_BUFFER_SIZE=1024" >> .env
+  echo "NOTIFIER_RING_BUFFER_SIZE=1024" >> .env
+fi
+
+# Output the values
+echo ".env properties:"
+echo "IS DEBUG: $is_debug"
+echo "IS SMOKE_TEST: $is_smoke_test"
```
mattp-swirldslabs, line 36: I think you can remove lines 35 and 36 and add a line to cat the actual values of `.env`

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+
+/**
+ * A class that extends {@link MappedConfigSource} ir order to have project-relevant initialization.
+ */
+public final class ServerMappedConfigSourceInitializer {
+    private static final List<ConfigMapping> MAPPINGS =
+            List.of(
+                    new ConfigMapping("mediator.ringBufferSize", "MEDIATOR_RING_BUFFER_SIZE"),
+                    new ConfigMapping("notifier.ringBufferSize", "NOTIFIER_RING_BUFFER_SIZE"));
```
mattp-swirldslabs, line 31: We should probably add an entries for the other config properties like:
```
consumer.timeoutThresholdMillis
service.delayMillis
persistence.storage.rootPath
```

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private ServerMappedConfigSourceInitializer() {
+        throw new UnsupportedOperationException("Creating instances is not supported!");
+    }
```
mattp-swirldslabs, line 35: You can remove this `UnsupportedOperationException` ^^

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Queue;
+import java.util.function.Predicate;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class ServerMappedConfigSourceInitializerTest {
+    private static final ConfigMapping[] SUPPORTED_MAPPINGS = {
+        new ConfigMapping("mediator.ringBufferSize", "MEDIATOR_RING_BUFFER_SIZE"),
+        new ConfigMapping("notifier.ringBufferSize", "NOTIFIER_RING_BUFFER_SIZE")
```
mattp-swirldslabs, line 34: I think you can also test these 3:
```
consumer.timeoutThresholdMillis
service.delayMillis
persistence.storage.rootPath
```

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
```diff
@@ -0,0 +1,100 @@
...
+     * the developer so we can make sure we are aware of them in order to be sure we require the
+     * change. This test is more to bring attention than to test actual logic. So if this fails, we
+     * either change the {@link #SUPPORTED_MAPPINGS} here or the {@link
+     * ServerMappedConfigSourceInitializer#MAPPINGS} to make this pass.
+     */
+    @Test
+    void test_VerifyAllSupportedMappingsAreAddedToInstance() throws ReflectiveOperationException {
+        final Queue<ConfigMapping> actual = extractConfigMappings();
+
+        Assertions.assertEquals(SUPPORTED_MAPPINGS.length, actual.size());
```
mattp-swirldslabs, line 56: You can use static imports for `Assertions` here ^^

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * This method constructs, initializes and returns a new instance of {@link MappedConfigSource}
+     * which internally uses {@link SystemEnvironmentConfigSource} and maps relevant config keys to
+     * other keys so that they could be used within the application
+     *
+     * @return newly constructed fully initialized {@link MappedConfigSource}
+     */
```
mattp-swirldslabs, line 43: You can add the `@NonNull` annotation to the top of this method.

mattp-swirldslabs (commented):
I added a few comments

rbarkerSL (approved):

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
ata-nas: Do we need any different values for tests currently, or no, we simply need to support the mapping for now?

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
ata-nas: will do

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
ata-nas: will do

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
ata-nas: will do

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
ata-nas: will do. What is the convention of adding this to the method? Is it to state that we will not return a null value? Just to know for the future.

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
```diff
@@ -0,0 +1,100 @@
...
+            final Predicate<ConfigMapping> predicate =
+                    cm ->
+                            current.mappedName().equals(cm.mappedName())
+                                    && current.originalName().equals(cm.originalName());
+            Assertions.assertTrue(
+                    actual.stream().anyMatch(predicate),
+                    "when testing for: ["
+                            + current
+                            + "] it is not contained in mappings of the actual initialized object "
+                            + actual);
```
jsync-swirlds, line 68: nit: It's usually much clearer to use `"template with %s, %d, etc...".formatted(values...)` for messages; it's almost always easier to read the template than to follow a chain of string concatenations.

suites/build.gradle.kts
```diff
@@ -38,14 +38,33 @@ mainModuleInfo {
...
+// knowing what is being run could lead to silent bugs that are hard to come by. Not to mention, if
+// we have a task that will run a project, and then a task with the same name that will also run a
+// project, now when executing from root we will run two projects which might be problematic.
+// Having qualifiers is a 'need-to-know' solution which is not preferable. But there is also the
+// case
+// that it would be a good idea to have tasks with the same name that we know for sure run isolated,
+// and we need them to run on all projects easily, so running the task from repo root will execute
+// everywhere it is present. So the question is should these remain with the same name and what
+// should be our approach for the future?
+
```
jsync-swirlds, line 59: Please open a discussion in github and move this text to that discussion, that way we can comment and discuss without muddying the codebase.

jsync-swirlds (commented):
Looks good, just one nitpick (which can be addressed later), and one block of text that should be a github discussion rather than a comment in code.

suites/build.gradle.kts
ata-nas: moved to discussion #264

server/docker/update-env.sh
ata-nas: done

server/docker/update-env.sh
ata-nas: done

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
ata-nas: done, added only mapping for now assuming that in the `smoke-tests` we will use the default values

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
ata-nas: done

suites/build.gradle.kts
ata-nas: moved to #264

mattp-swirldslabs (approved):
Looks good. Since you changed something in `.github/workflows` you might need approval from someone in RE or DevOps.

jsync-swirlds (approved):

AlfredoG87 (approved):
LGTM

rbarkerSL (approved):
Review applies to 

- `.github/workflows/smoke-test.yaml`
- `README.md`

LGTM

jsync-swirlds:
@ata-nas, it appears that most of your commits on this PR are unsigned.
You can fix this with the following git commands:
```shell
git rebase --gpg-sign -f HEAD~20
git push --force
```

Issue #235, opened 2024-10-08
investigate and fix release automation flow


rbarkerSL:
**Description**:

The block node repository is missing key items that will enable the release automation flow to execute in the same way it does on the Relay and Smart Contracts repositories.

- Need to add a PAT to enable the DCO commit and tag step to function
- Need to ensure permissions have been granted to the workflows
- Need to ensure that the workflow runs with a workflow-dispatch trigger

Issue #229, opened 2024-10-02
Test, Fix and Follow-up on "Automated release process" attempt to release `0.1.0-rc1` and `0.1.0-rc2` versions.


AlfredoG87:
Recently the `release-automation.yaml` was merged into `main` to start automate the release process, however as expected, after merging, is necessary to do some adjustments, fixes and configurations to the repo to be able to iron-out the WF. This is expected since this is a fairly complex workflow with variables that are hard (close to impossible) to replicate locally. 

This tasks is a place-holder for the testing phase.

So far this has come up:
- [x] Change `secrets.HEDERA_BOT_TOKEN` for `secrets.GITHUB_TOKEN` at the checkout step.
- [x] Requested RE team to add the following secrets to the repo: `secrets.GPG_PRIVATE_KEY` and `secrets.GPG_PASSPHRASE`

The conclusion of this ticket is to iron-out the process and verify is working by sucessfully releasing `0.1.0` starting with  attempt to release `0.1.0-rc1` and `0.1.0-rc2` versions.

AlfredoG87:
@rbarkerSL 
https://swirldslabs-help.freshdesk.com/support/tickets/5749

rbarkerSL:
PR #236 will complete this ticket. @AlfredoG87

PR #236, opened 2024-10-08
ci: Update release-automation.yaml to enable completion

rbarkerSL:
**Description**:

Uses the newly added GPG_KEY_CONTENTS, GPG_KEY_PASSPHRASE, and GH_ACCESS_TOKEN github actions secrets to enable the workflow to run.

**Related Issue(s)**:

Fixes: #235
Fixes: #229 

Testing:

See the [v99.99.99-draft workflow run](https://github.com/hashgraph/hedera-block-node/actions/runs/11247622943/job/31272540073) which completes successfully

### DIFF
--- a/.github/workflows/release-automation.yaml
+++ b/.github/workflows/release-automation.yaml
@@ -135,7 +135,7 @@ jobs:
         uses: Decathlon/release-notes-generator-action@98423db7024696a339f3988ac8a2b051c5860741 # v3.1.6
         env:
           FILENAME: ${{ env.RELEASE_NOTES_FILENAME }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
           MILESTONE_NUMBER: ${{ steps.milestone.outputs.milestone_id }}
 
       - name: Commit and Tag
--- a/.github/workflows/release-automation.yaml
+++ b/.github/workflows/release-automation.yaml
@@ -30,9 +30,6 @@ defaults:
   run:
     shell: bash
 
-env:
-  LC_ALL: C.UTF-8
-
 jobs:
   release:
     name: Release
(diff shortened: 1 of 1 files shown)


.github/workflows/release-automation.yaml
```diff
@@ -80,17 +77,18 @@ jobs:
...
           fetch-depth: 0
           ref: main
-          token: ${{ secrets.HEDERA_BOT_TOKEN }}
+          token: ${{ secrets.GH_ACCESS_TOKEN }}
+          persist-credentials: 'true'
 
       - name: Import GPG Key
         id: gpg_importer
-        uses: crazy-max/ghaction-import-gpg@01dd5d3ca463c7f10f7f4f7b4f177225ac661ee4 # v6.1.0
+        uses: step-security/ghaction-import-gpg@6c8fe4d0126a59d57c21f87c9ae5dd3451fa3cca # v6.1.0
```
andrewb1269hg, line 85: Does the comment version `# v6.1.0` need to be updated with the change of commit id?

.github/workflows/release-automation.yaml
```diff
@@ -218,4 +219,4 @@ jobs:
           delete-branch: true
           signoff: true
           title: ${{ needs.release.outputs.pr_title }}
-          token: ${{ secrets.HEDERA_BOT_TOKEN }}
```
andrewb1269hg, line 221: There are 2 locations for `HEDERA_BOT_TOKEN` in this file that have been removed. Can we get rid of the `HEDERA_BOT_TOKEN` all together?

.github/workflows/release-automation.yaml
rbarkerSL: it does not. step-security is now maintaining their initial version with a version matching the source action.

See the step security released version here: https://github.com/step-security/ghaction-import-gpg/releases/tag/v6.1.0

The commit id associated with that version matches what's in this action.

.github/workflows/release-automation.yaml
andrewb1269hg: Roger mentioned this never existed in the first place. Marking resolved.

andrewb1269hg:
Looks good!

andrewb1269hg (approved):

PR #245, opened 2024-10-10
fix: update-env.sh now injects JAVA_OPTS params

### LINKED ISSUE #244: Add docker jvm JAVA_OPTS params to `.env`
### Problem

As a Block Node developer
I want to be able to pass the Block Node JVM parameters into the container
So that Block Node can run with enough memory

### Solution

Augment the `update-env.sh` script to append JAVA_OPTS params to the `.env` file for docker

### Alternatives

_No response_

mattp-swirldslabs:
**Description**:
* Added shell code to append the JVM memory requirements via JAVA_OPTS into the `.env` file

**Related issue(s)**:

Fixes #244


AlfredoG87 (approved):
LGTM

PR #248, opened 2024-10-10
feat: added MILLIS_PER_BLOCK Streaming mode to the simulator

### LINKED ISSUE #249: Add Simulator Streaming Mode to reflect more accurate CN Behaviour
### Problem

Currently the Simulator is streaming block items, and waiting an X amount of time in Nanos in between each block_item.

This is not the actual behaviour by the CN who will attempt to send the block as fast as possible as they become available in a batch of items.

### Solution

We need to add a StreamingMode to the simulator that allows us to attempt to send a block every X amount of time.

### Alternatives

_No response_

AlfredoG87:
**Description**:
Improvements to how the simulator is able to stream to the BlockNode, this commit creates a StreamingMode enum, with 2 values, CONSTANT_RATE and MILLIS_PER_BLOCK.

CONSTANT_RATE is the streaming mode that we have been using, that streams a block item, then waits X amount of NS, then streams another blockItem.

MILLIS_PER_BLOCK, does a lot less thread interruptions, and has a more realistic behaviour, since is going to stream the whole block, as fast as it can, and then going to sleep for X-(time it took to stream) being X the target of block production, by default 1 block is produced each second, so it will attempt to stream 1 block per every 1000 ms, however if the block takes more than 1000 ms to stream, it wont sleep at all, but WARN of a potential issue.

Added documentation for explaining the changes.

**Related issue(s)**:

Fixes #249 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/simulator/README.md
+++ b/simulator/README.md
@@ -38,15 +38,17 @@ There are 2 configuration sets:
 ### BlockStreamConfig
 Uses the prefix `blockStream` so all properties should start with `blockStream.`
 
-| Key                      | Description                                                                                                                                                                | Default Value                   |
-|--------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------|
-| `generationMode`         | The desired generation Mode to use, it can only be `DIR` or `AD_HOC`                                                                                                       | `DIR`                           |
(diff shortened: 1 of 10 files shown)


georgi-l95 (approved):

mattp-swirldslabs (approved):
Looks good. I like all of the tests! :)

PR #267, opened 2024-10-16
docs: Add a protocol document for `publishBlockStream`

### LINKED ISSUE #120: Design handshake between CN and BN to determine what Block to start streaming from.
Design a basic protocol for CN (or simulator) to initiate a stream to a BN that ensures the BN does not have any gaps in the block stream.

#### Block Node Connect
* CN, on connect, sends block header, this contains block number.
   * If next block, no problem, start streaming.
   * If less than last known verified block, respond with "DuplicateBlock"
      * Response includes last known block, so CN can perhaps do its own catch up or reconnect.
      * This REQUIRES CN to check and resend block header, or end the stream and restart.
      * This includes if CN sends a block less than the last known block, that this block
        node, for some reason, does not actually hold.
         * In this case the block node _must_ retrieve the missing block(s) from another
           block node to fill the gaps, but shall always respond to CN with the very
           latest known and verified block.  The streaming API is only for current data,
           not for filling gaps.
   * If greater than next block, we missed block(s)
      * Respond with "Behind"
         * This includes last known block number.
         * CN will send from block after that block, or send "EndOfStream" and retry with exponential backoff.
         * CN will include earliest known block with end of stream, so we have an idea of the range to catch up.
            * This is _advisory_, and will almost certainly change before we finish "catching up".
         * If CN retries before we "catch up", we record the offered block number, and continue trying to "catch up" to that. Response is still "Behind" with last known block number.
            * This allows CN to jump in to "catch us up" directly if we're behind, but close enough.
         * We probably need a failure detection if the required target block doesn't get "closer" with each connection from CN.
      * If CN ends stream, need to catch up from other BN
         * Query BN "status" API, get last available block
            * If greater than or equal to block number CN sent
               * Ask for range, last-known-block+1 to last-available-block.
               * Hopefully catch up before next CN connection.
            * If less than block number CN sent
               * Either ask for stream last-known-block+1 to "infinite" and quit when caught up OR ask another BN, in case all needed blocks available elsewhere.
   * Each CN connect will send a block header, repeat above process until we get a matched block number or CN can finish catching us up.
* Note, we can (re)enter connect _any_ time we get a next block from CN that isn't what we expect. This simplifies logic for working out when to retry or reset a stream.

#### Error Handling
* If CN detects an error at any time
   * Next BlockItem will be an `EndStream` item with an appropriate error code.
   * Block Node will drop any in-progress unproven block from that CN, and, if
     no remaining active incoming streams, notify all subscribers with an
     `EndStream` item specifying "source error".
   * Block Node will continue streaming from other incoming stream sources, if
     any, or await a restarted stream if no other incoming stream sources.
* If a BN detects an error at any time
   * BN will send an `EndStream` response to all incoming streams, with appropriate
     status code.
      * CN, on receiving the end stream, will retry publishing the stream; and will
        use exponential backoff if the BN failure continues.
         * If CN has multiple "downstream" BN options, a CN _may_ connect to an alternate
           BN for reliability and mark the failed BN as a backup.
   * BN will send `EndStream` to all subscribers with appropriate status code.
   * BN will either recover or await manual recovery.

jsync-swirlds:
* Added a new document in `docs/Protocol` for `publishBlockStream`.
* Included a first pass protocol definition.
   * Note this is in general informal language; not Estelle, SDL, LOTOS,
     or other highly formal language.

Fixes #120

### DIFF
--- a/docs/Protocol/publishBlockStream.md
+++ b/docs/Protocol/publishBlockStream.md
@@ -1,25 +1,53 @@
 # Block Node Connect Protocol for `publishBlockStream`
+
+## Abstract
+This protocol describes how a Publisher and Block Node SHALL interact for
+the `publishBlockStream` API. The protocol provides for either side of the
+stream to signal an orderly end of the stream, for the block node to signal
+that it has fallen behind the publisher, for the block node to signal that the
+publisher has fallen behind the block node, and describes basic error
+conditions. The protocol also includes specific provision for how both
+publisher and block node will behave when a block node must spend time to
+"catch up" to a publisher by requesting older blocks from another block node.
+This protocol does not describe behavior related to the similar "pull" API
+`subscribeBlockStream` or to the "catch up" process.
+
+### Reliability note
+This protocol is very careful about error conditions. A consensus node
+publishing to a block node faces a hard requirement that if a block is not
+validated and persisted within a short time, the consensus node must stop
+processing entirely until it does have assurance that its blocks are
+validated and persisted. This occurs because the block stream is the actual
+"block chain" for the network, and if even one block is lost, then the chain
+is broken; with disastrous consequences. Consensus nodes are one of the
+core clients of this protocol as Publisher, so a core assumption of this
+protocol is that if anything occurs to suggest to a publisher that a block
+node might fail to store blocks permanently, the publisher should end the
+stream and retry (either to another block node, or after a short delay).
+
 ## Definitions
 
-Block Node
-: A software system intended to store and process a Block Stream.  The API for
-  a Block Node is defined in HIP 1056, among others.
+<dl>
+<dt>Block Node</dt>
+<dd>A software system intended to store and process a Block Stream.  The API for
+    a Block Node is defined in HIP 1056, among others.</dd>
 
-Block Number
-: A monotonically increasing number assigned by consensus to each block produced
-  by the network.
+<dt>Block Number</dt>
+<dd>A monotonically increasing number assigned by consensus to each block produced
+  by the network.</dd>
 
-Publisher
-: An entity publishing blocks to a Block Node via the `publishBlockStream` API
-  This is typically a Consensus Node or another Block Node.
+<dt>Publisher</dt>
+<dd>An entity publishing blocks to a Block Node via the `publishBlockStream` API
+  This is typically a Consensus Node or another Block Node.</dd>
 
-Subscriber
-: An entity that subscribes to a verified or unverified Block Stream from a
-  Block Node.
+<dt>Subscriber</dt>
+<dd>An entity that subscribes to a verified or unverified Block Stream from a
+  Block Node.</dd>
 
-Verified Block
-: A verified block is a block for which a Block Proof is received and for which
-  the TSS signature of the network ledger ID is valid.
+<dt>Verified Block</dt>
+<dd>A verified block is a block for which a Block Proof is received and for which
+  the TSS signature of the network ledger ID is valid.</dd>
+</dl>
 
 ## Base Protocol
 * Publisher, on connect, sends a block header, this contains a block number.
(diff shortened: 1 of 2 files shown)


docs/Protocol/publishBlockStream.md
```diff
@@ -0,0 +1,164 @@
...
+: An entity publishing blocks to a Block Node via the `publishBlockStream` API
+  This is typically a Consensus Node or another Block Node.
+
+Subscriber
+: An entity that subscribes to a verified or unverified Block Stream from a
+  Block Node.
+
+Verified Block
+: A verified block is a block for which a Block Proof is received and for which
+  the TSS signature of the network ledger ID is valid.
```
georgi-l95, line 22:
```suggestion
`Block Node`: A software system intended to store and process a Block Stream. The API for
  a Block Node is defined in `HIP-1056`, among others.

`Block Number`: A monotonically increasing number assigned by consensus to each block produced
  by the network.

`Publisher`: An entity publishing blocks to a Block Node via the `publishBlockStream` API. This is typically a Consensus Node or another Block Node.

`Subscriber`: An entity that subscribes to a verified or unverified Block Stream from a
  Block Node.

`Verified Block`: A verified block is a block for which a Block Proof is received and for which
  the TSS signature of the network ledger ID is valid.
```

docs/Protocol/publishBlockStream.md
```diff
@@ -0,0 +1,164 @@
...
+            * If it is less than the block number Publisher sent
+               * Either ask for a stream last-known-block+1 to "infinite" and
+                 quit when caught up OR ask another Block Node, in case all
+                 needed blocks available elsewhere.
+   * Each Publisher connect will send a block header, repeat above process until
+     the Block Node gets a matched block number or Publisher can finish
+     catching up that Block Node.
+* Note, a Block Node can (re)enter connect _any_ time that Block Node gets the
+  next block from Publisher with a block number that is not what the Block Node
+  expects. This simplifies logic for working out when to retry or reset a stream.
```
georgi-l95, line 74:
```suggestion
**Note**: Block Node can (re)enter connect _any_ time that Block Node gets the
  next block from Publisher with a block number that is not what the Block Node
  expects. This simplifies logic for working out when to retry or reset a stream.
```

docs/Protocol/publishBlockStream.md
```diff
@@ -0,0 +1,164 @@
...
+
+Subscriber
+: An entity that subscribes to a verified or unverified Block Stream from a
+  Block Node.
+
+Verified Block
+: A verified block is a block for which a Block Proof is received and for which
+  the TSS signature of the network ledger ID is valid.
+
+## Base Protocol
```
georgi-l95, line 24: The format in which is described the logic bugs me a little, as it's gets a little deep and somewhat not very eye friendly. Although I don't have any better suggestions currently, so we can leave it like this. If we come up with something better in the future, this can be easily changed.

docs/Protocol/publishBlockStream.md
```diff
@@ -0,0 +1,164 @@
+# Block Node Connect Protocol for `publishBlockStream`
```
georgi-l95, line 1: We could add overview section, briefly describing what this document is all about. OR we can add a readme in the root folder itself, describing what protocol design documents are stored here.

georgi-l95 (commented):
Overall looks good, I've already reviewed the design, so don't have any comments on it. Left a few formatting suggestions.

docs/Protocol/publishBlockStream.md
jsync-swirlds: This is a definition list (github doesn't support them properly, unfortunately).
The syntax is markdown syntax for a definition list, not a mistake.

Would this read better using HTML definition list syntax?

docs/Protocol/publishBlockStream.md
jsync-swirlds: I agree, I wasn't able to come up with a better layout after trying a few alternatives.
Short of using a formal language (which would have been much less approachable), nothing seemed to improve the structure in a meaningful way.

docs/Protocol/publishBlockStream.md
jsync-swirlds: Good suggestion. I'll add both.

AlfredoG87 (approved):
LGTM 💯

docs/Protocol/publishBlockStream.md
georgi-l95: I think it did, thank you 👍

docs/Protocol/publishBlockStream.md
georgi-l95: Ah, gotcha. I think it looks better now 👍

georgi-l95 (approved):
tyvm, looks good 🚀

PR #269, opened 2024-10-16
refactor: Stream of repeated block items instead of block items

### LINKED ISSUE #261: Change all Helidon-related elements to support `List<BlockItem>`.
As a Consensus Node
I Want to stream Block Items in repeated batches
So That I can improve the channel efficiency and reduce streaming overhead

### Technical Notes
Currently we are using protobuf definitions for singular BlockItem streams.
Supporting batches (repeated in protobuf) requires that all logic related to processing those items needs to support lists of block items instead.
This story will introduce the new data structure and change all related logic to processing blocks in batches.

### LINKED ISSUE #262: Update dependency version for protobufs to bring in new definitions.
Related to https://github.com/hashgraph/hedera-protobufs/pull/421
once the above PR is merged, use main with a `commit` hash reference instead of the branch.

### LINKED ISSUE #268: Update the ring buffer and other "services" in block node to support `List<BlockItem>`.

### LINKED ISSUE #276: Refactor GRPC BlockStream Service to extract BlockAccess as its own Service

### LINKED ISSUE #277: Update smoke tests
As a Block Node
I Want to have smoke tests

Need to do the necessary changes for the smoke tests to pass after the refactor of the grpc services and the repeated block items.

### LINKED ISSUE #278: Update simulator to produce variable length `repeated` block items
as a CN
I want the simulator to stream the block items in batches of X configurable throught a property.

# notes

by default set it at `1000` block items per batch.
block proof can only be at the end of a batch and block header can only be at the beginning of a batch, that way a single batch cannot have more than 1 block and boundaries become simpler.

AlfredoG87:
**Description**:

This PR changes the version of `hedera-protobuf` and refactors all the project due to 2 fundamental changes to the API spec.

1. Service was changed from 1 single service called `BlockStreamService` to 4 different services: `BlockStreamService`, `BlockAccessService`, `BlockNodeService`, `StateService`.
2. `BlockStreamService` --> `publishBlockStream` and `subscribeBlockStream` to use a repeated set of `block_items` per request instead of a single `block_item`.

Changes done to support the change:
1. Updated `stream` branch/commit to use a ~temporal~ temporary branch until is merged to main.
2. Changes to the simulator to steam by batch
3. Changes to Helidon Interfaces and Internal Services
4. Fixing Unit Tests to reflect changes
5. Refactored `BlockStreamService` to extract `BlockStreamAccess` methods.
6. Add `BlockStreamAccesss` Grpc Service to start-up routes.
7. Added new and missing UTs
8. Updated Configuration and Documentation 

**Related issue(s)**:

Fixes #261
Fixes #262 
Fixes #268 
Fixes #276 
Fixes #277 
Fixes #278 

**Notes for reviewer**:
Main refactor is completed, at least is compiling :)

**Checklist**

### DIFF
--- a/common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
+++ b/common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
@@ -16,6 +16,7 @@
 
 package com.hedera.block.common.utils;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -23,8 +24,14 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+/** Utility class for chunking collections. */
 public final class ChunkUtils {
-    public static <T> List<List<T>> chunkify(final Collection<T> collection, final int chunkSize) {
+    /** Chunks a collection into a list of lists of the specified size.
+     * @param collection the collection to chunk, if the collection is empty, an empty list is returned.
+     * @param chunkSize the size of each chunk
+     *  */
+    public static <T> List<List<T>> chunkify(
+            @NonNull final Collection<T> collection, final int chunkSize) {
         Objects.requireNonNull(collection);
         if (chunkSize <= 0) {
             throw new IllegalArgumentException("Chunk size must be greater than 0");
--- a/common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
+++ b/common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.block.common.utils;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public final class ChunkUtils {
+    public static <T> List<List<T>> chunkify(final Collection<T> collection, final int chunkSize) {
+        Objects.requireNonNull(collection);
+        if (chunkSize <= 0) {
+            throw new IllegalArgumentException("Chunk size must be greater than 0");
+        }
+        if (collection.isEmpty()) {
+            return Collections.emptyList(); // or throw, depends on how we want to handle
+        }
+        final List<T> localCollection = List.copyOf(collection);
+        final int localCollectionSize = localCollection.size();
+        return IntStream.iterate(0, i -> i < localCollectionSize, i -> i + chunkSize)
+                .mapToObj(
+                        i ->
+                                localCollection.subList(
+                                        i, Math.min(i + chunkSize, localCollectionSize)))
+                .collect(Collectors.toList());
+    }
+
+    private ChunkUtils() {}
+}
--- a/common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
+++ b/common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
@@ -29,6 +29,8 @@ public final class ChunkUtils {
     /** Chunks a collection into a list of lists of the specified size.
      * @param collection the collection to chunk, if the collection is empty, an empty list is returned.
      * @param chunkSize the size of each chunk
+     * @param <T> the type of the collection
+     * @return a list of lists of the specified size
      *  */
     public static <T> List<List<T>> chunkify(
             @NonNull final Collection<T> collection, final int chunkSize) {
--- a/server/src/main/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserver.java
+++ b/server/src/main/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserver.java
@@ -17,7 +17,6 @@
 package com.hedera.block.server.consumer;
 
 import static com.hedera.block.server.Translator.fromPbj;
-import static com.hedera.block.server.metrics.BlockNodeMetricTypes.Counter.LiveBlockItemsConsumed;
 import static java.lang.System.Logger;
 import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.ERROR;
@@ -27,14 +26,17 @@ import com.hedera.block.server.events.BlockNodeEventHandler;
 import com.hedera.block.server.events.LivenessCalculator;
 import com.hedera.block.server.events.ObjectEvent;
 import com.hedera.block.server.mediator.SubscriptionHandler;
+import com.hedera.block.server.metrics.BlockNodeMetricTypes;
 import com.hedera.block.server.metrics.MetricsService;
 import com.hedera.hapi.block.SubscribeStreamResponse;
 import com.hedera.hapi.block.stream.BlockItem;
 import com.hedera.pbj.runtime.OneOf;
+import com.swirlds.metrics.api.Counter;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import io.grpc.stub.ServerCallStreamObserver;
 import io.grpc.stub.StreamObserver;
 import java.time.InstantSource;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
@@ -55,7 +57,7 @@ public class ConsumerStreamResponseObserver
 
     private final AtomicBoolean isResponsePermitted = new AtomicBoolean(true);
     private final ResponseSender statusResponseSender = new StatusResponseSender();
-    private final ResponseSender blockItemResponseSender = new BlockItemResponseSender();
+    private final ResponseSender blockItemsResponseSender = new BlockItemsResponseSender();
 
     private static final String PROTOCOL_VIOLATION_MESSAGE =
             "Protocol Violation. %s is OneOf type %s but %s is null.\n%s";
@@ -186,7 +188,7 @@ public class ConsumerStreamResponseObserver
                 subscribeStreamResponse.response();
         return switch (oneOfTypeOneOf.kind()) {
             case STATUS -> statusResponseSender;
-            case BLOCK_ITEM -> blockItemResponseSender;
+            case BLOCK_ITEMS -> blockItemsResponseSender;
             default -> throw new IllegalArgumentException(
                     "Unknown response type: " + oneOfTypeOneOf.kind());
         };
@@ -196,35 +198,39 @@ public class ConsumerStreamResponseObserver
         void send(@NonNull final SubscribeStreamResponse subscribeStreamResponse);
     }
 
-    private final class BlockItemResponseSender implements ResponseSender {
+    private final class BlockItemsResponseSender implements ResponseSender {
         private boolean streamStarted = false;
 
         public void send(@NonNull final SubscribeStreamResponse subscribeStreamResponse) {
 
-            // Only start sending BlockItems after we've reached
-            // the beginning of a block.
-            final BlockItem blockItem = subscribeStreamResponse.blockItem();
-            if (blockItem == null) {
+            if (subscribeStreamResponse.blockItems() == null) {
                 final String message =
                         PROTOCOL_VIOLATION_MESSAGE.formatted(
                                 "SubscribeStreamResponse",
-                                "BLOCK_ITEM",
-                                "block_item",
+                                "BLOCK_ITEMS",
+                                "block_items",
                                 subscribeStreamResponse);
                 LOGGER.log(ERROR, message);
                 throw new IllegalArgumentException(message);
-            } else {
-                if (!streamStarted && blockItem.hasBlockHeader()) {
-                    streamStarted = true;
-                }
+            }
 
-                if (streamStarted) {
-                    LOGGER.log(DEBUG, "Sending BlockItem downstream: " + blockItem);
+            final List<BlockItem> blockItems = subscribeStreamResponse.blockItems().blockItems();
+            // Only start sending BlockItems after we've reached
+            // the beginning of a block.
+            if (!streamStarted && blockItems.getFirst().hasBlockHeader()) {
+                streamStarted = true;
+            }
 
-                    // Increment counter
-                    metricsService.get(LiveBlockItemsConsumed).increment();
-                    subscribeStreamResponseObserver.onNext(fromPbj(subscribeStreamResponse));
+            if (streamStarted) {
+                LOGGER.log(DEBUG, "Sending BlockItem Batch downstream");
+
+                final Counter liveBlockItemsConsumed =
+                        metricsService.get(BlockNodeMetricTypes.Counter.LiveBlockItemsConsumed);
+                // Increment counter manually
+                for (int i = 0; i < blockItems.size(); i++) {
+                    liveBlockItemsConsumed.increment();
                 }
+                subscribeStreamResponseObserver.onNext(fromPbj(subscribeStreamResponse));
             }
         }
     }
--- a/server/src/main/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserver.java
+++ b/server/src/main/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserver.java
@@ -31,12 +31,12 @@ import com.hedera.block.server.metrics.MetricsService;
 import com.hedera.hapi.block.SubscribeStreamResponse;
 import com.hedera.hapi.block.stream.BlockItem;
 import com.hedera.pbj.runtime.OneOf;
-import com.swirlds.metrics.api.Counter;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import io.grpc.stub.ServerCallStreamObserver;
 import io.grpc.stub.StreamObserver;
 import java.time.InstantSource;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
@@ -214,22 +214,22 @@ public class ConsumerStreamResponseObserver
                 throw new IllegalArgumentException(message);
             }
 
-            final List<BlockItem> blockItems = subscribeStreamResponse.blockItems().blockItems();
+            final List<BlockItem> blockItems =
+                    Objects.requireNonNull(subscribeStreamResponse.blockItems()).blockItems();
             // Only start sending BlockItems after we've reached
             // the beginning of a block.
             if (!streamStarted && blockItems.getFirst().hasBlockHeader()) {
+                LOGGER.log(
+                        DEBUG,
+                        "Sending BlockItem Batch downstream for block: "
+                                + blockItems.getFirst().blockHeader().number());
                 streamStarted = true;
             }
 
             if (streamStarted) {
-                LOGGER.log(DEBUG, "Sending BlockItem Batch downstream");
-
-                final Counter liveBlockItemsConsumed =
-                        metricsService.get(BlockNodeMetricTypes.Counter.LiveBlockItemsConsumed);
-                // Increment counter manually
-                for (int i = 0; i < blockItems.size(); i++) {
-                    liveBlockItemsConsumed.increment();
-                }
+                metricsService
+                        .get(BlockNodeMetricTypes.Counter.LiveBlockItemsReceived)
+                        .add(blockItems.size());
                 subscribeStreamResponseObserver.onNext(fromPbj(subscribeStreamResponse));
             }
         }
--- a/server/src/main/java/com/hedera/block/server/grpc/BlockAccessService.java
+++ b/server/src/main/java/com/hedera/block/server/grpc/BlockAccessService.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.block.server.grpc;
+
+import static com.hedera.block.server.Constants.SINGLE_BLOCK_METHOD_NAME;
+import static com.hedera.block.server.Translator.fromPbj;
+import static com.hedera.block.server.Translator.toPbj;
+import static com.hedera.block.server.metrics.BlockNodeMetricTypes.Counter.SingleBlocksNotFound;
+import static com.hedera.block.server.metrics.BlockNodeMetricTypes.Counter.SingleBlocksRetrieved;
+import static java.lang.System.Logger.Level.DEBUG;
+import static java.lang.System.Logger.Level.ERROR;
+
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.hedera.block.server.Constants;
+import com.hedera.block.server.metrics.MetricsService;
+import com.hedera.block.server.persistence.storage.read.BlockReader;
+import com.hedera.block.server.service.ServiceStatus;
+import com.hedera.hapi.block.SingleBlockRequest;
+import com.hedera.hapi.block.SingleBlockResponseCode;
+import com.hedera.hapi.block.protoc.BlockService;
+import com.hedera.hapi.block.protoc.SingleBlockResponse;
+import com.hedera.hapi.block.stream.Block;
+import com.hedera.pbj.runtime.ParseException;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.grpc.stub.StreamObserver;
+import io.helidon.webserver.grpc.GrpcService;
+import java.io.IOException;
+import java.util.Optional;
+import javax.inject.Inject;
+
+/**
+ * The BlockAccessService class provides a gRPC service to access blocks.
+ *
+ * <p>This service provides a unary gRPC method to retrieve a single block by block number.
+ */
+public class BlockAccessService implements GrpcService {
+    private final System.Logger LOGGER = System.getLogger(getClass().getName());
+
+    private final ServiceStatus serviceStatus;
+    private final BlockReader<Block> blockReader;
+    private final MetricsService metricsService;
+
+    /**
+     * Constructs a new BlockAccessService instance with the given dependencies.
+     *
+     * @param serviceStatus used to query the service status
+     * @param blockReader used to retrieve blocks
+     * @param metricsService used to observe metrics
+     */
+    @Inject
+    public BlockAccessService(
+            @NonNull ServiceStatus serviceStatus,
+            @NonNull BlockReader<Block> blockReader,
+            @NonNull MetricsService metricsService) {
+        this.serviceStatus = serviceStatus;
+        this.blockReader = blockReader;
+        this.metricsService = metricsService;
+    }
+
+    @Override
+    public Descriptors.FileDescriptor proto() {
+        return BlockService.getDescriptor();
+    }
+
+    @Override
+    public String serviceName() {
+        return Constants.SERVICE_NAME_BLOCK_ACCESS;
+    }
+
+    @Override
+    public void update(Routing routing) {
+        routing.unary(SINGLE_BLOCK_METHOD_NAME, this::protocSingleBlock);
+    }
+
+    void protocSingleBlock(
+            @NonNull final com.hedera.hapi.block.protoc.SingleBlockRequest singleBlockRequest,
+            @NonNull final StreamObserver<SingleBlockResponse> singleBlockResponseStreamObserver) {
+        LOGGER.log(DEBUG, "Executing Unary singleBlock gRPC method");
+
+        try {
+            final SingleBlockRequest pbjSingleBlockRequest =
+                    toPbj(SingleBlockRequest.PROTOBUF, singleBlockRequest.toByteArray());
+            singleBlock(pbjSingleBlockRequest, singleBlockResponseStreamObserver);
+        } catch (ParseException e) {
+            LOGGER.log(ERROR, "Error parsing protoc SingleBlockRequest: {0}", singleBlockRequest);
+            singleBlockResponseStreamObserver.onNext(buildSingleBlockNotAvailableResponse());
+        }
+    }
+
+    private void singleBlock(
+            @NonNull final SingleBlockRequest singleBlockRequest,
+            @NonNull
+                    final StreamObserver<com.hedera.hapi.block.protoc.SingleBlockResponse>
+                            singleBlockResponseStreamObserver) {
+
+        LOGGER.log(DEBUG, "Executing Unary singleBlock gRPC method");
+
+        if (serviceStatus.isRunning()) {
+            final long blockNumber = singleBlockRequest.blockNumber();
+            try {
+                final Optional<Block> blockOpt = blockReader.read(blockNumber);
+                if (blockOpt.isPresent()) {
+                    LOGGER.log(DEBUG, "Successfully returning block number: {0}", blockNumber);
+                    singleBlockResponseStreamObserver.onNext(
+                            fromPbjSingleBlockSuccessResponse(blockOpt.get()));
+
+                    metricsService.get(SingleBlocksRetrieved).increment();
+                } else {
+                    LOGGER.log(DEBUG, "Block number {0} not found", blockNumber);
+                    singleBlockResponseStreamObserver.onNext(buildSingleBlockNotFoundResponse());
+                    metricsService.get(SingleBlocksNotFound).increment();
+                }
+            } catch (IOException e) {
+                LOGGER.log(ERROR, "Error reading block number: {0}", blockNumber);
+                singleBlockResponseStreamObserver.onNext(buildSingleBlockNotAvailableResponse());
+            } catch (ParseException e) {
+                LOGGER.log(ERROR, "Error parsing block number: {0}", blockNumber);
+                singleBlockResponseStreamObserver.onNext(buildSingleBlockNotAvailableResponse());
+            }
+        } else {
+            LOGGER.log(ERROR, "Unary singleBlock gRPC method is not currently running");
+            singleBlockResponseStreamObserver.onNext(buildSingleBlockNotAvailableResponse());
+        }
+
+        // Send the response
+        singleBlockResponseStreamObserver.onCompleted();
+    }
+
+    @NonNull
+    static com.hedera.hapi.block.protoc.SingleBlockResponse buildSingleBlockNotAvailableResponse() {
+        final com.hedera.hapi.block.SingleBlockResponse response =
+                com.hedera.hapi.block.SingleBlockResponse.newBuilder()
+                        .status(SingleBlockResponseCode.READ_BLOCK_NOT_AVAILABLE)
+                        .build();
+
+        return fromPbj(response);
+    }
+
+    @NonNull
+    static com.hedera.hapi.block.protoc.SingleBlockResponse buildSingleBlockNotFoundResponse()
+            throws InvalidProtocolBufferException {
+        final com.hedera.hapi.block.SingleBlockResponse response =
+                com.hedera.hapi.block.SingleBlockResponse.newBuilder()
+                        .status(SingleBlockResponseCode.READ_BLOCK_NOT_FOUND)
+                        .build();
+
+        return fromPbj(response);
+    }
+
+    @NonNull
+    static com.hedera.hapi.block.protoc.SingleBlockResponse fromPbjSingleBlockSuccessResponse(
+            @NonNull final Block block) {
+        final com.hedera.hapi.block.SingleBlockResponse singleBlockResponse =
+                com.hedera.hapi.block.SingleBlockResponse.newBuilder()
+                        .status(SingleBlockResponseCode.READ_BLOCK_SUCCESS)
+                        .block(block)
+                        .build();
+
+        return fromPbj(singleBlockResponse);
+    }
+}
--- a/server/src/main/java/com/hedera/block/server/grpc/BlockAccessService.java
+++ b/server/src/main/java/com/hedera/block/server/grpc/BlockAccessService.java
@@ -16,6 +16,7 @@
 
 package com.hedera.block.server.grpc;
 
+import static com.hedera.block.server.Constants.SERVICE_NAME_BLOCK_ACCESS;
 import static com.hedera.block.server.Constants.SINGLE_BLOCK_METHOD_NAME;
 import static com.hedera.block.server.Translator.fromPbj;
 import static com.hedera.block.server.Translator.toPbj;
@@ -26,7 +27,6 @@ import static java.lang.System.Logger.Level.ERROR;
 
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.InvalidProtocolBufferException;
-import com.hedera.block.server.Constants;
 import com.hedera.block.server.metrics.MetricsService;
 import com.hedera.block.server.persistence.storage.read.BlockReader;
 import com.hedera.block.server.service.ServiceStatus;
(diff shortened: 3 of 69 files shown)


server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorImpl.java
```diff
@@ -87,24 +90,30 @@ class LiveStreamMediatorImpl extends SubscriptionHandlerBase<SubscribeStreamResp
...
+                    SubscribeStreamResponseSet.newBuilder().blockItems(blockItems).build();
             final var subscribeStreamResponse =
-                    SubscribeStreamResponse.newBuilder().blockItem(blockItem).build();
+                    SubscribeStreamResponse.newBuilder().blockItems(blockItemsSet).build();
             ringBuffer.publishEvent((event, sequence) -> event.set(subscribeStreamResponse));
 
             // Increment the block item counter
-            metricsService.get(LiveBlockItems).increment();
+            Counter liveBlockItems = metricsService.get(LiveBlockItems);
+            for (int i = 0; i < blockItems.size(); i++) {
```
georgi-l95, line 114: This seems like a loop inside a hot path. Do you think that it can be avoided somehow ?

georgi-l95 (commented):
LG, some questions

server/src/main/java/com/hedera/block/server/grpc/BlockAccessService.java
```diff
@@ -0,0 +1,176 @@
...
+            LOGGER.log(ERROR, "Unary singleBlock gRPC method is not currently running");
+            singleBlockResponseStreamObserver.onNext(buildSingleBlockNotAvailableResponse());
+        }
+
+        // Send the response
+        singleBlockResponseStreamObserver.onCompleted();
+    }
+
+    @NonNull
+    static com.hedera.hapi.block.protoc.SingleBlockResponse buildSingleBlockNotAvailableResponse() {
```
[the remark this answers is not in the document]
ata-nas: +1

server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorImpl.java
```diff
@@ -87,24 +90,30 @@ class LiveStreamMediatorImpl extends SubscriptionHandlerBase<SubscribeStreamResp
...
+    public void publish(@NonNull final List<BlockItem> blockItems) {
 
         if (serviceStatus.isRunning()) {
 
             // Publish the block for all subscribers to receive
-            LOGGER.log(DEBUG, "Publishing BlockItem: " + blockItem);
+            LOGGER.log(DEBUG, "Publishing BlockItem");
+            final SubscribeStreamResponseSet blockItemsSet =
+                    SubscribeStreamResponseSet.newBuilder().blockItems(blockItems).build();
             final var subscribeStreamResponse =
```
ata-nas, line 108: I would avoid using var anywhere, not a blocker though.

server/src/main/java/com/hedera/block/server/producer/ProducerBlockItemObserver.java
```diff
@@ -157,42 +160,47 @@ public ProducerBlockItemObserver(
...
 
-        try {
-            LOGGER.log(DEBUG, "Received PublishStreamRequest from producer");
-            final BlockItem blockItem =
-                    toPbj(BlockItem.PROTOBUF, publishStreamRequest.getBlockItem().toByteArray());
-            LOGGER.log(DEBUG, "Received block item: " + blockItem);
+        LOGGER.log(DEBUG, "Received PublishStreamRequest from producer");
+        final List<BlockItem> blockItemsPbj = new ArrayList<>();
+        final Counter liveBlockItemsReceived = metricsService.get(LiveBlockItemsReceived);
+        for (final var blockItemProtoc : publishStreamRequest.getBlockItemsList()) {
```
ata-nas, line 166: Personally I would avoid ever using `var`, but not a blocker

server/src/test/java/com/hedera/block/server/grpc/BlockAccessServiceTest.java
```diff
@@ -0,0 +1,288 @@
...
+
+    private static final int testTimeout = 1000;
+
+    private Path testPath;
+    private BlockNodeContext blockNodeContext;
+    private PersistenceStorageConfig config;
+
+    @BeforeEach
+    public void setUp() throws IOException {
+        testPath = Files.createTempDirectory(TEMP_DIR);
```
ata-nas, line 97: Why not use `@TempDir`?

server/src/test/java/com/hedera/block/server/grpc/BlockStreamServiceTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+    private static final int testTimeout = 1000;
+
+    private Path testPath;
+    private BlockNodeContext blockNodeContext;
+    private PersistenceStorageConfig config;
+
+    @BeforeEach
+    public void setUp() throws IOException {
+        testPath = Files.createTempDirectory(TEMP_DIR);
```
ata-nas, line 95: Same here, maybe we can use `@TempDir`?

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
```diff
@@ -136,11 +137,12 @@ public void testRemoveBlockReadPermsRepairFailed() throws IOException, ParseExce
...
 
-        final BlockWriter<BlockItem> blockWriter =
+        final BlockWriter<List<BlockItem>> blockWriter =
                 BlockAsDirWriterBuilder.newBuilder(blockNodeContext).build();
-        for (BlockItem blockItem : blockItems) {
-            blockWriter.write(blockItem);
-        }
+        //        for (BlockItem blockItem : blockItems) {
+        //            blockWriter.write(blockItem);
+        //        }
```
ata-nas, line 144: Commented out code, remove maybe?

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
```diff
@@ -168,11 +170,12 @@ public void testRepairReadPermsFails() throws IOException, ParseException {
...
 
-        final BlockWriter<BlockItem> blockWriter =
+        final BlockWriter<List<BlockItem>> blockWriter =
                 BlockAsDirWriterBuilder.newBuilder(blockNodeContext).build();
-        for (final BlockItem blockItem : blockItems) {
-            blockWriter.write(blockItem);
-        }
+        //        for (final BlockItem blockItem : blockItems) {
+        //            blockWriter.write(blockItem);
+        //        }
```
ata-nas, line 177: Commented out code, remove maybe?

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
```diff
@@ -206,11 +209,12 @@ public void testBlockNodePathReadFails() throws IOException, ParseException {
...
```
ata-nas, line 216: Commented out code, remove maybe?

server/src/test/java/com/hedera/block/server/producer/ProducerBlockItemObserverTest.java
```diff
@@ -149,7 +149,9 @@ public void testBlockItemThrowsParseException() throws IOException {
         fromPbj(PublishStreamResponse.newBuilder().status(endOfStream).build());
 
         // verify the ProducerBlockItemObserver has sent an error response
-        verify(publishStreamResponseObserver, timeout(testTimeout).times(1))
+        verify(
+                        publishStreamResponseObserver,
+                        timeout(testTimeout).atLeast(1)) // It fixes if set it to 2, but why???
```
ata-nas, line 154: Is the comment on line 154 a `todo` or `fixme`?

simulator/README.md
ata-nas, line 1: For the readme and configs in genera, do we need to support configuration with env variables? If yes, then all the key/value pairs need to be mapped inside `com.hedera.block.server.config.ServerMappedConfigSourceInitializer` and the awareness test must be updated. Then if env variables must be supported, I think in the readme this must be pointed out and have the env var keys there that we actually support. If this does not apply here, please ignore.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
```diff
@@ -72,12 +83,28 @@ public boolean streamBlockItem(BlockItem blockItem) {
...
+        List<com.hedera.hapi.block.stream.protoc.BlockItem> blockItemsProtoc = new ArrayList<>();
         for (BlockItem blockItem : block.items()) {
-            streamBlockItem(blockItem);
+            blockItemsProtoc.add(Translator.fromPbj(blockItem));
         }
 
-        // wait for ack on the block
-        // if and when the ack is received return true
+        final int blockItemsNumberOfBatches =
+                block.items().size() / blockStreamConfig.blockItemsBatchSize();
```
ata-nas, line 92: I have a concern here with the way this is being batched, hope I am wrong. This int/int division should result in a whole int that would be the closest to the next int (if result is 2.3 then 2 will be returned, if result is 2.9 then 3 will be returned). Please correct me if I am wrong, but this undeterministic behavior will result in bugs.

Also, lets extract this batching logic in the `common` module, my proposal is something of the sort (or propose similar) this will return the chunks and we can foreach them functionally:
```java
public final class ChunkUtils {
    public static <T> List<List<T>> chunkify(final Collection<T> collection, final int chunkSize) {
        Objects.requireNonNull(collection);
        if (chunkSize <= 0) {
            throw new IllegalArgumentException("Chunk size must be greater than 0");
        }
        return IntStream.iterate(0, i -> i < collection.size(), i -> i + chunkSize)
            .mapToObj(i -> List.copyOf(collection).subList(i, Math.min(i + chunkSize, collection.size())))
            .collect(Collectors.toList());
    }

    private ChunkUtils() {
    }
}
```

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
```diff
@@ -41,6 +41,7 @@ public void onNext(PublishStreamResponse publishStreamResponse) {
     @Override
     public void onError(Throwable throwable) {
         logger.log(Logger.Level.ERROR, "Error: " + throwable.toString());
+        // TODO: Stop the stream, retry, or stop simulation
```
ata-nas, line 44: TODO

server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorImpl.java
```diff
@@ -87,24 +90,30 @@ class LiveStreamMediatorImpl extends SubscriptionHandlerBase<SubscribeStreamResp
...
-                    SubscribeStreamResponse.newBuilder().blockItem(blockItem).build();
+                    SubscribeStreamResponse.newBuilder().blockItems(blockItemsSet).build();
             ringBuffer.publishEvent((event, sequence) -> event.set(subscribeStreamResponse));
 
             // Increment the block item counter
-            metricsService.get(LiveBlockItems).increment();
+            Counter liveBlockItems = metricsService.get(LiveBlockItems);
+            for (int i = 0; i < blockItems.size(); i++) {
+                liveBlockItems.increment();
+            }
```
mattp-swirldslabs, line 116: I think you can use this rather than a loop:
```
metricsService.get(LiveBlockItems).add(blockItems.size())
```

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
```diff
@@ -118,12 +118,14 @@ public void onEvent(ObjectEvent<SubscribeStreamResponse> event, long l, boolean
                             metricsService.get(StreamPersistenceHandlerError).increment();
                             throw new BlockStreamProtocolException(message);
                         } else {
-                            // Persist the BlockItem
-                            Optional<BlockItem> result = blockWriter.write(blockItem);
+                            // Persist the BlockItems
+                            List<BlockItem> blockItems =
+                                    subscribeStreamResponse.blockItems().blockItems();
+                            Optional<List<BlockItem>> result = blockWriter.write(blockItems);
```
mattp-swirldslabs, line 124: We will have to figure out what Optional to return here so the downstream response can be generated. This is probably ok for now.

server/src/test/java/com/hedera/block/server/grpc/BlockAccessServiceTest.java
```diff
@@ -0,0 +1,288 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class BlockAccessServiceTest {
+
+    @Mock private Notifier notifier;
```
mattp-swirldslabs, line 73: Many of these member variable are not used and can be removed

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
```diff
@@ -113,11 +113,12 @@ public void testReadPermsRepairSucceeded() throws IOException, ParseException {
...
 
-        final BlockWriter<BlockItem> blockWriter =
+        final BlockWriter<List<BlockItem>> blockWriter =
                 BlockAsDirWriterBuilder.newBuilder(blockNodeContext).build();
-        for (BlockItem blockItem : blockItems) {
-            blockWriter.write(blockItem);
-        }
+        //        for (BlockItem blockItem : blockItems) {
+        //            blockWriter.write(blockItem);
+        //        }
```
mattp-swirldslabs, line 120: Remove commented code?

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
mattp-swirldslabs, line 144: Remove commented code?

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
mattp-swirldslabs, line 177: Remove commented code?

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
mattp-swirldslabs, line 216: Remove commented code?

mattp-swirldslabs (commented):
Nice work. I added some comments

server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorImpl.java
AlfredoG87: This is another example of PlatformSDK metrics API feature limitations, any other prometheus client would have a `increment(number)` overload that increments more than `1` at the time such counter.

This might be fixed in the future by changes to the SDK or we can decide to change the Metrics, however for this refactor I deemed the best course of action to keep the same Metric intention (block items published) or on a follow-up ticket we can decide to get rid of that metric, and focus on another metric that makes more sense.

But for now I did not found a way to increment the metric other than 1 int at a time.

server/src/main/java/com/hedera/block/server/grpc/BlockAccessService.java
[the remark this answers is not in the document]
mattp-swirldslabs: Not now, these packages are necessary to disambiguate the protoc version from PBJ types

server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorImpl.java
mattp-swirldslabs: This compiles: ` metricsService.get(LiveBlockItems).add(blockItems.size());` but I haven't tested it.

AlfredoG87: I worked!! thanks!! :)

I guess I missed that api due to expecting it as a overload 😅

server/src/test/java/com/hedera/block/server/producer/ProducerBlockItemObserverTest.java
mattp-swirldslabs: Good catch @ata-nas. These test parameters are intentionally very narrow. If you have to loosen them, I think it indicates there might be a bug. Let's collaborate to make sure the code is correct.

AlfredoG87: Good find, this is indeed something that needs more time to determine the cause, I will change it to a TODO, so we don't miss it on a followup.

simulator/README.md
AlfredoG87: I think we should be able to support them, is this something that can be done as part of a separate PR? as a follow-up? I don't think the refactor should have an even bigger scope for now.

My proposal is we create a Bug ticket for fixing the .env variables.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
AlfredoG87: I like your approach, but I rather not use `.forEach` for concurrency and performance concerns.  other than that, happy to implement this improvement, and YES! this is a bug waiting to happen, will use your approah. Thank you!! 🙏

server/src/main/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserver.java
```diff
@@ -196,35 +198,39 @@ private interface ResponseSender {
...
                                 subscribeStreamResponse);
                 LOGGER.log(ERROR, message);
                 throw new IllegalArgumentException(message);
-            } else {
-                if (!streamStarted && blockItem.hasBlockHeader()) {
-                    streamStarted = true;
-                }
+            }
+
+            final List<BlockItem> blockItems = subscribeStreamResponse.blockItems().blockItems();
```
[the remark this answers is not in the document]
AlfredoG87: I think we should never encounter that null, that would mean an empty batch.

server/src/main/java/com/hedera/block/server/producer/ProducerBlockItemObserver.java
jsync-swirlds: I generally agree, but in rare instances it fixes the thoroughly excessive wrapping caused by the current spotless configuration.
When it saves multiple line wraps, and is not ambiguous, I would argue that it is acceptable.

simulator/README.md
jsync-swirlds: Please be careful with `.env`.  In my opinion we should not even use such files, due to the very high incidence of secrets being exposed with `.env` files.
Environment variables (set via appropriate tooling at runtime) is good practice, but storing environment variables in `.env` files is concerning, at best.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
jsync-swirlds: We may want to consider a process for these.  Something like this might work:
```
A "TODO" comment is permitted, but it must include a ticket ID.
We prefer the following format.
`@todo(###) description`.
```
This can also be used for `future` feature notes and other items that should refer to issues.

server/src/main/java/com/hedera/block/server/producer/ProducerBlockItemObserver.java
```diff
@@ -157,42 +160,47 @@ public ProducerBlockItemObserver(
...
+
+                // Stop the server
+                serviceStatus.stopWebServer(getClass().getName());
+            }
+            liveBlockItemsReceived.increment();
+        }
 
-            metricsService.get(LiveBlockItemsReceived).increment();
+        // is this log just too much? see below log for proposed...
+        LOGGER.log(DEBUG, "Received block item batch with {} items.", blockItemsPbj.size());
```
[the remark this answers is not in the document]
jsync-swirlds: We _might_ choose to count message tags in the byte array and record that.
We'll need to be able to split the byte array into separate items without parsing anyway, for verification, so that code will be needed regardless.

server/src/main/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserver.java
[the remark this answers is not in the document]
jsync-swirlds: There is no guarantee we won't receive an empty/null value in the stream.  We'll need to handle it properly.

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
AlfredoG87: should we add it to an epic? Maybe the Persist Story Epic https://github.com/hashgraph/hedera-block-node/issues/194

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
AlfredoG87: ✅ Done!

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
AlfredoG87: ✅ Done!

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
AlfredoG87: ✅ Done!

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
AlfredoG87: ✅ Done!

server/src/test/java/com/hedera/block/server/grpc/BlockAccessServiceTest.java
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorImpl.java
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserver.java
```diff
@@ -196,35 +198,39 @@ private interface ResponseSender {
...
-                    // Increment counter
-                    metricsService.get(LiveBlockItemsConsumed).increment();
-                    subscribeStreamResponseObserver.onNext(fromPbj(subscribeStreamResponse));
+                final Counter liveBlockItemsConsumed =
+                        metricsService.get(BlockNodeMetricTypes.Counter.LiveBlockItemsConsumed);
+                // Increment counter manually
+                for (int i = 0; i < blockItems.size(); i++) {
+                    liveBlockItemsConsumed.increment();
                 }
+                subscribeStreamResponseObserver.onNext(fromPbj(subscribeStreamResponse));
```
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserver.java
```diff
@@ -196,35 +198,39 @@ private interface ResponseSender {
...
+            // Only start sending BlockItems after we've reached
+            // the beginning of a block.
+            if (!streamStarted && blockItems.getFirst().hasBlockHeader()) {
+                streamStarted = true;
+            }
 
-                if (streamStarted) {
-                    LOGGER.log(DEBUG, "Sending BlockItem downstream: " + blockItem);
+            if (streamStarted) {
+                LOGGER.log(DEBUG, "Sending BlockItem Batch downstream");
```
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserver.java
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/grpc/BlockStreamService.java
```diff
@@ -40,26 +35,19 @@
...
-import java.io.IOException;
 import java.time.Clock;
-import java.util.Optional;
 import javax.inject.Inject;
 
 /**
  * The BlockStreamService class defines the gRPC service for the block stream service. It provides
- * the implementation for the bidirectional streaming, server streaming, and unary methods defined
- * in the proto file.
+ * the implementation for the bidirectional streaming, server streaming as defined in the proto file.
```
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/producer/ProducerBlockItemObserver.java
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/grpc/BlockAccessService.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Override
+    public Descriptors.FileDescriptor proto() {
+        return BlockService.getDescriptor();
+    }
+
+    @Override
+    public String serviceName() {
+        return Constants.SERVICE_NAME_BLOCK_ACCESS;
```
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
AlfredoG87: ✅ Done!

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
AlfredoG87: ✅ Done!

simulator/README.md
AlfredoG87: ✅ Done!

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
AlfredoG87: ✅ Done!

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
AlfredoG87: ✅ Done!

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
AlfredoG87: ✅ Done!

server/src/test/java/com/hedera/block/server/grpc/BlockStreamServiceTest.java
AlfredoG87: ✅ Done!

server/src/test/java/com/hedera/block/server/grpc/BlockAccessServiceTest.java
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/producer/ProducerBlockItemObserver.java
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorImpl.java
AlfredoG87: ✅ Done!

server/src/main/java/com/hedera/block/server/grpc/BlockAccessService.java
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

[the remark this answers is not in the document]
AlfredoG87: I believe no need to do this 👍

common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.stream.IntStream;
+
+public final class ChunkUtils {
+    public static <T> List<List<T>> chunkify(final Collection<T> collection, final int chunkSize) {
+        Objects.requireNonNull(collection);
+        if (chunkSize <= 0) {
+            throw new IllegalArgumentException("Chunk size must be greater than 0");
+        }
+        if (collection.isEmpty()) {
+            return Collections.emptyList(); // or throw, depends on how we want to handle
```
ata-nas, line 33: Should we throw here or return an empty collection? What should be the general behavior?

common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public final class ChunkUtils {
+    public static <T> List<List<T>> chunkify(final Collection<T> collection, final int chunkSize) {
```
ata-nas, line 27: Consider to add `@NonNull` to the `collection` parameter

simulator/build.gradle.kts
```diff
@@ -19,6 +19,8 @@ plugins {
     id("com.hedera.block.simulator")
 }
 
+dependencies { implementation(project(":common")) }
+
```
ata-nas, line 23: maybe this was added here by mistake?

server/src/test/java/com/hedera/block/server/grpc/BlockAccessServiceTest.java
```diff
@@ -64,38 +60,29 @@
...
     @Mock private ServiceStatus serviceStatus;
 
-    private final System.Logger LOGGER = System.getLogger(getClass().getName());
-
-    private static final String TEMP_DIR = "block-node-unit-test-dir";
-
     private static final int testTimeout = 1000;
 
-    private Path testPath;
+    @TempDir private Path testPath;
```
ata-nas, line 80: We do not need cleanup in the `@AfterEach` method, by default junit will attempt to recursively delete the `@TempDir` and if it does not succeed it should throw an exception and the test should fail (this should be based on documentation). Consider to take a look if we could remove this teardown logic and if indeed everything cleans up without it by default.

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
```diff
@@ -52,23 +51,19 @@
...
+import org.junit.jupiter.api.io.TempDir;
 
 public class BlockAsDirReaderTest {
 
     private final Logger LOGGER = System.getLogger(getClass().getName());
 
-    private static final String TEMP_DIR = "block-node-unit-test-dir";
-
-    private Path testPath;
+    @TempDir private Path testPath;
```
ata-nas, line 60: We do not need cleanup in the `@AfterEach` method, by default junit will attempt to recursively delete the `@TempDir` and if it does not succeed it should throw an exception and the test should fail (this should be based on documentation). Consider to take a look if we could remove this teardown logic and if indeed everything cleans up without it by default.

ata-nas (commented):
Looking better! A couple comments on latest changes only reviewed.

simulator/README.md
ata-nas: @jsync-swirlds agreed for the .env file, it is dangerous indeed, when I was talking about `env variables` I did indeed mean environment variables and not the .env file, but your point is quite important.
@AlfredoG87 I think yes, we could do it in another commit and should not be a problem if we do so, but I would suggest that a ticket is made so we do not forget about that.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
ata-nas: Noted, glad you like the idea!
For me is resolved, either with the advanced for or the `forEach`, up to you.

I wonder however what concerns do you have about performance and concurrency by using `forEach`? The implementation internally is a for loop again, (with tiny overhead probably against the advanced for which uses the iterator, but I am not sure if it even exists as an overhead)? And about the concurrency what issues could arise? Maybe there is something I do not know, please let me know I would love to learn! Otherwise the advanced for loop and the `forEach` look the same to me.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
ata-nas: @AlfredoG87 noted, I keep this conversation open as a reminder, feel free to resolve it after you open an issue, or if you think that this todo is enough for now. Also @jsync-swirlds has a quite good idea for the convention for the todos, maybe if you open an issue, consider to use his process to tag the todo with the issue number.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriter.java
```diff
@@ -94,43 +95,46 @@ class BlockAsDirWriter implements BlockWriter<BlockItem> {
...
+                        repairPermissions(calculateBlockPath());
+                        LOGGER.log(INFO, "Retrying to write the BlockItem protobuf to a file");
+                    }
                 }
             }
         }
 
-        if (blockItem.hasBlockProof()) {
+        if (blockItems.getLast().hasBlockProof()) {
             metricsService.get(BlocksPersisted).increment();
```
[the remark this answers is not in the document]
jsync-swirlds: This is blocks, not block items, so it only ever increments by one at a time.  We _should not_ ever have multiple blocks in one set of items, nor parts of two different blocks in the same set.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
AlfredoG87: Is my understanding that the `.forEach` is executed in parallel (could be wrong) , also, is less performant (this is true). But ultimately, a regular For gives you more flexibility overall and a cleaner and simpler code.

common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
AlfredoG87: I prefer to return an empty collection, I think it makes it more generic to any use case.
Maybe just add to the javadoc the expected behaviour to keep things simple.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriter.java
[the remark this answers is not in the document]
AlfredoG87: No, this increment refers only to a whole block. 👍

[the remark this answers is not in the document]
ata-nas: Noted, thanks for the clarification! :)

common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
ata-nas: Awesome, sounds good.

common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
ata-nas, line 1: Can you please add some tests for this as well? https://github.com/hashgraph/hedera-block-node/pull/274 is now available and will be even easier! :)

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
AlfredoG87: https://github.com/hashgraph/hedera-block-node/issues/286

simulator/README.md
AlfredoG87: https://github.com/hashgraph/hedera-block-node/issues/285

common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
AlfredoG87: ✅  Done!

common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
AlfredoG87: ✅  Done!

simulator/build.gradle.kts
AlfredoG87: ✅  Removed!

server/src/test/java/com/hedera/block/server/grpc/BlockStreamServiceTest.java
```diff
@@ -82,19 +81,14 @@ public class BlockStreamServiceTest {
 
     private final Logger LOGGER = System.getLogger(getClass().getName());
 
-    private static final String TEMP_DIR = "block-node-unit-test-dir";
-
     private static final int testTimeout = 1000;
 
-    private Path testPath;
+    @TempDir private Path testPath;
```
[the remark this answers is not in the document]
AlfredoG87: ✅  Done!

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
AlfredoG87: ✅  Done!

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
jsync-swirlds: I've opened a [discussion item](https://github.com/hashgraph/hedera-block-node/discussions/287) to discuss the proposed process and format.

common/src/main/java/com/hedera/block/common/utils/ChunkUtils.java
AlfredoG87: ✅  Done!

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
ata-nas: @jsync-swirlds +1 from me for the discussion & resolved by @AlfredoG87

ata-nas (approved):
Looks good to me, all comments I had on everything I managed to find is now resolved. Please see others as well! 👍

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
mattp-swirldslabs: I added this item: "
Review Optional return value from the Writer so downstream code can generate the correct response"

mattp-swirldslabs (approved):
Looks good

server/src/test/java/com/hedera/block/server/consumer/ConsumerStreamResponseObserverTest.java
```diff
@@ -142,8 +145,10 @@ public void testResponseNotPermittedAfterCancel() {
                         testClock, streamMediator, serverCallStreamObserver, testContext);
 
         final List<BlockItem> blockItems = generateBlockItems(1);
+        final SubscribeStreamResponseSet subscribeStreamResponseSet =
```
jsync-swirlds, line 148: Just a note, the latest PR removes this message and replaces it with a more generic `BlockItemSet`, because it turned out we needed a `oneof` in the publish API too.

server/src/test/resources/block_service.proto
jsync-swirlds, line 1: We should plan to remove this soonish.  It was supposed to be temporary, and we are better off using the "real" protos, even for tests.

stream/build.gradle.kts
```diff
@@ -33,10 +33,10 @@ tasks.withType<JavaCompile>().configureEach {
 tasks.cloneHederaProtobufs {
     // uncomment below to use a specific tag
     // tag = "v0.53.0" or a specific commit like "0047255"
-    tag = "d5e6988"
+    // tag = "d5e6988"
 
     // uncomment below to use a specific branch
-    // branch = "main"
+    branch = "block-repeated-fix-tss-message"
```
jsync-swirlds, line 39: Don't forget to reset this before merging.

PR #271, opened 2024-10-17
refactor: use commons module in simulator

### LINKED ISSUE #253: refactor: use commons module in simulator
## Story
**AS A** Block Node Developer
**I WANT** to use already existing logic from the `:common` module that was abstracted there
**SO THAT** we have good architecture and maintainability

## Tech Notes
- refactor `:simulator` to use logic introduced in #233 in the `:common` module

ata-nas:
**Description**:

- removed `Constants.class` from `:simulator`;
- removed `Utils.class` from `:simulator`
- replace usages with `FileUtilities.class` from `:common`

**Related issue(s)**:

Fixes #253

**Notes for reviewer**:

N/A

**Checklist**


common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
```diff
@@ -166,17 +171,38 @@ public static byte[] readGzipFileUnsafe(@NonNull final Path filePath) throws IOE
...
+     *
+     * @return byte array of the content of the file or null if the file extension is not supported
+     *
+     * @throws IOException      if unable to read the file.
+     * @throws OutOfMemoryError if a byte array large enough to contain the file contents cannot be
+     *                          allocated (either because it exceeds MAX_INT bytes or exceeds
+     *                          available heap memory).
      */
     public static byte[] readFileBytesUnsafe(@NonNull final Path filePath) throws IOException {
         return readFileBytesUnsafe(filePath, ".blk", ".gz");
```
AlfredoG87, line 208: should this use the Constants instead of hardcoding the values `.blk` and `.gz`?

AlfredoG87 (approved):
LG, just 1 question/nit.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: @AlfredoG87 good question. I was thinking about that today when reviewing #300, since that issue should IMO use this logic  (see [this](https://github.com/hashgraph/hedera-block-node/pull/300#discussion_r1814782347) comment in the review there). That being said, we used to have both these strings as constants in the beginning, but after a conversation we had with @jsync-swirlds and @mattp-swirldslabs we decided to remove them, since `.blk` is not generic and `.gz` we did not see any point in having, which I agree with both.

 The reason you see that implemented like that is because this is a single use, does not require a constant and is intended to be here only and not visible to the outside world (@jsync-swirlds actually finished that part after I left a comment on that method, I believe this was the thought process, correct me if I am wrong). And this method was extracted from a previously existing one and it was implemented with that if-else check for the extensions. I agree that currently this is maybe not very generic because of the `.blk` extension check.
 
I would propose that this can be utilized now by @jasperpotts for [this](https://github.com/hashgraph/hedera-block-node/pull/300#discussion_r1814782347) comment. and I think that the method should be refactored to **not** check for `.blk` but to be in the same fashion of what @jasperpotts currently has at 6bbb15975b6821a09cc688c252d39866912646fd from my comment, and to also ommit the `null` check. Personally I would propose the following change to the `readFileByteUnsafe`:
 
```java
    public static byte[] readFileBytesUnsafe(
            @NonNull final Path filePath,)
            throws IOException {
        final String filePathAsString = Objects.requireNonNull(filePath).toString();
        Objects.requireNonNull(".gz");
        if (filePathAsString.endsWith(gzipFileExtension)) {
            return readGzipFileUnsafe(filePath);
        } else {
            return Files.readAllBytes(filePath);
        }
    }
```

and of course we remove the overloaded method that already exists since they have now the same parameters.

This seems more generic to me and better fits the `common` module. @jsync-swirlds @mattp-swirldslabs @AlfredoG87  @jasperpotts What do you guys think about that change? I think is better to be done and @jasperpotts  can use the generic logic in his `tools` now. Let me know

jsync-swirlds: None of these extensions is likely to remain in place forever.  The whole point here is to have a method that accepts parameters for file extensions and offer a sane default via an override.  The values are hardcoded in the override just to avoid breaking existing code.  Once existing code changes to call the "parameterized" version, this convenience method can be removed entirely.

Let's _not_ (ever) rewrite things as a single method that hardcodes things like `.gz` extensions.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
```diff
@@ -62,21 +65,22 @@ public final class FileUtilities {
                             PosixFilePermission.OTHERS_EXECUTE));
 
     /**
-     * Log message template used when a path is not created because a file
-     * or folder already exists at the requested path.
+     * Log message template used when a path is not created because a file or folder already exists
+     * at the requested path.
```
jsync-swirlds, line 69: Javadoc should wrap with shorter lines; it's intended to be readable in github which locks down to a narrow viewport for some incomprehensible reason.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
```diff
@@ -62,21 +65,22 @@ public final class FileUtilities {
...
     /**
-     * Create a new path (folder or file) if it does not exist.
-     * Any folders or files created will use default permissions.
+     * Create a new path (folder or file) if it does not exist. Any folders or files created will
+     * use default permissions.
      *
-     * @param toCreate valid, non-null instance of {@link Path} to be created
-     * @param logLevel valid, non-null instance of {@link System.Logger.Level} to use
+     * @param toCreate         valid, non-null instance of {@link Path} to be created
+     * @param logLevel         valid, non-null instance of {@link System.Logger.Level} to use
```
jsync-swirlds, line 79: Please, please, please do _not_ start this horrid pattern.  There is no reason to use aligned indentation here, and it's ridiculously costly as people have to constantly reformat lines for simple variable renames.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
```diff
@@ -62,21 +65,22 @@ public final class FileUtilities {
...
      * @param semanticPathName valid, non-blank {@link String} used for logging that represents the
-     *     desired path semantically
-     * @param createDir {@link Boolean} value if we should create a directory or a file
+     *                         desired path semantically
+     * @param createDir        {@link Boolean} value if we should create a directory or a file
+     *
```
jsync-swirlds, line 83: Same here, do not try to align descriptions.
For Javadoc, we should use one space after the parameter _name_, and 4 space indentation(only) on subsequent lines.

note, if your IDE is doing this, change it to not reformat javadoc this way (IDE formatting doesn't match spotless anyway, so it is just always going to be wrong).

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
```diff
@@ -143,20 +146,22 @@ public static void createPathIfNotExists(
...
+     * @throws IOException      if unable to read the file.
+     * @throws OutOfMemoryError if a byte array large enough to contain the file contents cannot be
+     *                          allocated (either because it exceeds MAX_INT bytes or exceeds
+     *                          available heap memory).
      */
     public static byte[] readGzipFileUnsafe(@NonNull final Path filePath) throws IOException {
         Objects.requireNonNull(filePath);
-        try (final var gzipInputStream = new GZIPInputStream(Files.newInputStream(filePath))) {
+        try (final GZIPInputStream gzipInputStream =
+                new GZIPInputStream(Files.newInputStream(filePath))) {
```
jsync-swirlds, line 164: This was correct as it stood, there is no reason to reformat the line.
The use of `var` was deliberate to avoid an unnecessary line break.
The only reformatting we should be doing is via `./gradlew spotlessApply` so we don't
have random preferences creating back-and-forth warring reformats.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
```diff
@@ -166,17 +171,38 @@ public static byte[] readGzipFileUnsafe(@NonNull final Path filePath) throws IOE
...
+     *
+     * @param file to read bytes from
+     *
+     * @return byte array of the content of the file or null if the file extension is not supported
+     *
+     * @throws IOException      if unable to read the file.
+     * @throws OutOfMemoryError if a byte array large enough to contain the file contents cannot be
+     *                          allocated (either because it exceeds MAX_INT bytes or exceeds
+     *                          available heap memory).
+     */
```
jsync-swirlds, line 185: Don't reformat javadoc like this, please.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
```diff
@@ -166,17 +171,38 @@ public static byte[] readGzipFileUnsafe(@NonNull final Path filePath) throws IOE
...
+    public static byte[] readFileBytesUnsafe(@NonNull final File file) throws IOException {
+        return readFileBytesUnsafe(file.toPath());
+    }
```
jsync-swirlds, line 188: I do not believe this is a good change, generally you had to use a Path to get the File in the first place, and it's better to pass in the Path rather than converting the Path to a File and back to a Path..

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulator.java
```diff
@@ -28,39 +29,45 @@
 import java.lang.System.Logger;
 import java.nio.file.Path;
 
-/** The BlockStreamSimulator class defines the simulator for the block stream. */
+/**
+ * The BlockStreamSimulator class defines the simulator for the block stream.
+ */
```
jsync-swirlds, line 34: We should not expand these.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulator.java
```diff
@@ -28,39 +29,45 @@
...
 public class BlockStreamSimulator {
     private static final Logger LOGGER = System.getLogger(BlockStreamSimulator.class.getName());
 
-    /** This constructor should not be instantiated. */
+    /**
+     * This constructor should not be instantiated.
+     */
```
jsync-swirlds, line 40: There was no reason to expand this.

simulator/src/main/java/com/hedera/block/simulator/Constants.java
jsync-swirlds, line 1: This file should actually be retained.
These constants are specific to the simulator and should be passed to the methods in common, rather than relying on defaults in that module.
We didn't leave constants out of common to _hide_ things, but to encourage modules like simulator to _retain their own_ constants and pass them in as parameters.

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsDirBlockStreamManager.java
```diff
@@ -16,7 +16,7 @@
 
 package com.hedera.block.simulator.generator;
 
-import static com.hedera.block.simulator.generator.Utils.readFileBytes;
+import static com.hedera.block.common.utils.FileUtilities.readFileBytesUnsafe;
```
jsync-swirlds, line 19: Generally, prefer to _not_ static import methods.  It costs little to write `FileUtilities.readFileBytesUnsafe` and is _much_ easier to read and understand without having to simultaneously retain a ton of static imports in your head.

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsDirBlockStreamManager.java
```diff
@@ -106,32 +106,33 @@ public Block getNextBlock() {
...
+                try (final Stream<Path> blockItems =
                         Files.list(blockDirPath).filter(Files::isRegularFile)) {
-                    List<Path> sortedBlockItems =
+                    final List<Path> sortedBlockItems =
                             blockItems
                                     .sorted(
                                             Comparator.comparing(
                                                     BlockAsDirBlockStreamManager
                                                             ::extractNumberFromPath))
                                     .toList();
```
jsync-swirlds, line 126: Ugh, this is just super ugly with the excessive wrapping.
If we can add a temporary variable or two to fix that, it would be good.

jsync-swirlds (commented):
It appears you ran an IDE reformat on these files without disabling a lot of very bad defaults.  It's important to keep in mind that most IDEs do not, and often cannot, correctly format the codebase.  The gradle `spotlessApply` task is much preferred.
Most of these comments are to not apply those reformats (particularly in Javadoc which is using a very resource-expensive, completely unnecessary, and relatively difficult to read formatting pattern).

Beyond that the Constants file should not have been removed (it holds exactly the constants we should be keeping and using from the Simulator).

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: @jsync-swirlds I do share your concern for breaking the existing logic if we change the implementation, I did mention that in my comment, I do not believe that if we change to an `if-else` we will currently break anything (but need to check thoroughly) but yet it is indeed dangerous to change the implementation in such a drastic way. So yes, based on that I do tend to agree that this should remain implemented as is and not get changed. @AlfredoG87 hope that resolves it, from my side it does, the danger of breaking something (even if it might not be visible now) is enough for me to drop the proposal.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: I did this for convenience, since the only place I use that, I need to call `blockFile.getPath()`, and thought that it would be good to simply have this convenience where I do not need to call `getPath()` on a `File` object so that I can read the bytes. If you do really think this is not a good idea at all, I will remove this, but I do believe even in the future we will have places where the local scope only has the `File` object visible and we will need to call `File#getPath()` somewhere to get it.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: ✅ I reversed the formatting on the docs everywhere.

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsDirBlockStreamManager.java
ata-nas: ✅ I improved this, extracted the comparator to be more readable.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulator.java
ata-nas: ✅ I reversed the formatting on the docs everywhere.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulator.java
ata-nas: ✅ I reversed the formatting on the docs everywhere.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: ✅ I reversed the formatting on the docs everywhere.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: ✅ I reversed the formatting on the docs everywhere. It was indeed a problem with the IDE, I do use only spotlessApply, but looks like the IDE formatting also went through and I thought that was spotless that did that, sorry about that, I should not have this problem in the future.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: ✅ Personally I would never use `var`, but since that is the accepted norm and if it saves us the line break, I take a note. This is reversed now. And as for the reformatting, I do only use spotless, but looks like my IDE also applied reformatting and I thought it was from spotless on files I have changed, this should no longer be the case.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: ✅ I reversed the formatting on the docs everywhere. It was indeed a problem with the IDE, I do use only spotlessApply, but looks like the IDE formatting also went through and I thought that was spotless that did that everywhere I had changes, sorry about that, I should not have this problem in the future.

simulator/src/main/java/com/hedera/block/simulator/Constants.java
ata-nas: I agree that the file should generally be retained, I removed it since all the usages it had are no longer used, and we would have had a class with either no fields (if I deleted the unused, in our case these are all fields), or only with fields that have 0 usages.

If you think I should bring that back now, let me know, and also if I bring it back, should I exclude the fields that were there, since they have 0 usages and we will be left with an empty placeholder file, or should I return it as it was and it will be a file with all fields unused?

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsDirBlockStreamManager.java
ata-nas: ✅ Done.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
jsync-swirlds: IDE defaults are a pain sometimes...

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
jsync-swirlds: I generally agree on `var`, except when it makes a major improvement in readability, as here.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
jsync-swirlds: `File` is an older API that should be replaced by `Path` in nearly all cases.
In most cases the construction of a File is either in unit tests, or a leftover from older API assumptions.

Note, the one place (in non-test code) I see you using this, the File is constructed from Path and filename.  A Path should be `resolve`d _instead_ and passed in, avoiding _any_ construction of the `File` object (and making the caller more efficient in general).

Example:
We have (in relevant part)
```java
        final File blockFile = new File(blockstreamPath, nextBlockFileName);

        if (!blockFile.exists()) {
            return null;
        }

        try {
            final byte[] blockBytes = FileUtilities.readFileBytesUnsafe(blockFile);
```
Should be
```java
        final Path blockFile = blockstreamPath.resolve(nextBlockFileName);
        if (!Files.isReadable(blockFile)) {
            return null;
        }
        try {
            final byte[] blockBytes = FileUtilities.readFileBytesUnsafe(blockFile);
```

simulator/src/main/java/com/hedera/block/simulator/Constants.java
jsync-swirlds: It should be restored and the constants _used_ in the appropriate calls.  The reason you don't see usage is that you removed that usage.  In reality we should be using these constants as parameters provided to methods in the `common` library when called from this module.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: @jsync-swirlds noted, the context that you give changes everything, and that context was missing for me, hence the misunderstanding on my side as well as for the implementation details in #272 :)

That being said, I agree with you and the read method should have by default the ability to provide as arguments the suffix.

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsFileLargeDataSets.java
```diff
@@ -78,22 +81,24 @@ public BlockItem getNextBlockItem() throws IOException, BlockSimulatorParsingExc
...
+            final byte[] blockBytes =
+                    FileUtilities.readFileBytesUnsafe(
+                            localBlockStreamPath, RECORD_EXTENSION, GZ_EXTENSION);
 
-            LOGGER.log(INFO, "Loading block: " + blockFile.getName());
+            LOGGER.log(INFO, "Loading block: " + localBlockStreamPath.getFileName());
 
-            Block block = Block.PROTOBUF.parse(Bytes.wrap(blockBytes));
+            // todo blockBytes could be null, should we hande in some way or we need this method to
+            // fail here?
```
ata-nas, line 97: TODO, does anyone have any opinion on this?

@jsync-swirlds @mattp-swirldslabs @georgi-l95

simulator/src/main/java/com/hedera/block/simulator/Constants.java
ata-nas: True, I did remove the usages, because I used the overloaded method in the `FileUtilities` that existed for reading bytes. I have however brought back the file and use the extensions. Also it is my belief that the `blk` extension should have a `.` before it, so I changed it to `.blk`, but please correct me if that is wrong.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: I updated my example as you have proposed, have left a todo also in case someone wants to comment on the possibility of a null value now returned by our `common` lib method usage, let me know also your opinion, the todo is [here](https://github.com/hashgraph/hedera-block-node/pull/271/files#r1818628609).

common/src/test/java/com/hedera/block/common/utils/FileUtilitiesTest.java
```diff
@@ -32,9 +32,10 @@
 import org.junit.jupiter.params.provider.MethodSource;
 
 class FileUtilitiesTest {
+    private static final String FILE_WITH_UNRECOGNIZED_EXTENSION = "src/test/resources/nonexistent.unrecognized";
+
     @Test
-    void test_createPathIfNotExists_CreatesDirIfDoesNotExist(@TempDir final Path tempDir)
-            throws IOException {
+    void test_createPathIfNotExists_CreatesDirIfDoesNotExist(@TempDir final Path tempDir) throws IOException {
```
jsync-swirlds, line 38: Minor style note, we prefer to _not_ use mixed_SnakeCase like this for test names, and we also prefer to _not_ describe the test case in the method name with such detail.  It creates a lot of unnecessary maintenance with test method names changing for every test detail change.
Less detailed descriptions also work better with _parameterized_ tests (which are highly recommended).

Additional note:
This, and several following tests, could be a single parameterized `testCreatePathIfNotExists` and all the cases handled via parameter sets (likely from a `methodSource`).

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsFileLargeDataSets.java
jsync-swirlds: In the _simulator_, I would recommend failing quickly.  The simulator should not attempt to continue if input could not be read.
We would need to log clearly what happened, of course.

mattp-swirldslabs: Yeah, maybe add a null check and a fail the simulator quickly with a message.

ata-nas: Noted, I was thinking the same, I will throw here if null with a message (in the manner of "unable to read block file, most likely not found with given extensions", since we expect null in the else clause of the method, if the file is empty we should be getting a non-null empty array, the other option is an exception, which will happen before my check here, so is not within this scope) to the exception, hope that should be sufficient.

mattp-swirldslabs (dismissed):
Joseph has a few comments but otherwise looks good

common/src/test/java/com/hedera/block/common/utils/FileUtilitiesTest.java
ata-nas: Noted, can I address this comment in #272 since both are related and in any case there are merge conflicts between them, no matter which one goes in first? I suggest that this one goes in first since let's say it is more or less resolved and is a lot simpler than #272 and in #272 I will resolve the merge conflicts and address this. Also in #272 there are additional changes to these tests and it would be more difficult to resolve merge conflicts if I change now all the names here as well. Sounds good?

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsFileLargeDataSets.java
ata-nas: ✅ Done.

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsFileLargeDataSets.java
```diff
@@ -78,22 +82,27 @@ public BlockItem getNextBlockItem() throws IOException, BlockSimulatorParsingExc
...
-            LOGGER.log(INFO, "Loading block: " + blockFile.getName());
+            LOGGER.log(INFO, "Loading block: " + localBlockStreamPath.getFileName());
 
-            Block block = Block.PROTOBUF.parse(Bytes.wrap(blockBytes));
+            final Block block = Block.PROTOBUF.parse(Bytes.wrap(blockBytes));
             LOGGER.log(INFO, "block loaded with items size= " + block.items().size());
             return block;
-        } catch (ParseException e) {
+        } catch (final ParseException e) {
             throw new BlockSimulatorParsingException(e.getMessage());
```
jsync-swirlds, line 106: Quick note (that can be resolved in a later PR):
Never pass a message to a chained exception, always pass the whole exception.
If our custom exception does not have a chained exception constructor, add it.

jsync-swirlds (approved):

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsFileLargeDataSets.java
ata-nas: Noted, this was from the past, I will address it in #272.

ata-nas: ✅ Done.

common/src/test/java/com/hedera/block/common/utils/FileUtilitiesTest.java
ata-nas: ✅ Done. In addition, made minor changes to `StringUtilitiesTest` as well since I used the same naming conventions there.

PR #272, opened 2024-10-17
refactor: use commons module in server

### LINKED ISSUE #252: refactor: use commons module in server
## Story
**AS A** Block Node Developer
**I WANT** to use already existing logic from the `:common` module that was abstracted there
**SO THAT** we have good architecture and maintainability

## Tech Notes
- refactor `:server` to use logic introduced in #233 in the `:common` module

ata-nas:
**Description**:
- refactor `:server` to start using logic from `:common`

**Related issue(s)**:
Fixes #252 

**Notes for reviewer**:
- in the places where `com.hedera.block.common.utils.FileUtilities#createPathIfNotExists` is used, the previous alternative used `Files#createDirectory`, where the new version uses `Files#createDirectories` which acts quite differently, in fact, one unit test, namely `testProvidesBlockWriter_IOException`, which expects an exception to be thrown was not able to run successfully since it now creates the path using the new implementation with `Files#createDirectories` and previously it was not able to (the test is now modified). We should be careful about this since it is possible that the new usage could lead to silent bugs if unit tests do not already cover the needed edge cases.
- have I missed any places to refactor?

**Checklist**
- [x] refactor `com.hedera.block.server.Constants` remove `app.properties` and `logging.properties` and use them from `com.hedera.block.common.constants.StringsConstants`;
- [x] remove `com.hedera.block.server.persistence.storage.FileUtils` and start using `com.hedera.block.common.utils.FileUtilities` instead;
---
- [x] Tested (unit, integration, etc.) - **N/A**


AlfredoG87 (dismissed):
LGTM 👍

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
```diff
@@ -16,37 +16,40 @@
...
  * Use builder methods to create a {@link BlockReader} to read blocks from storage.
  *
  * <p>When a block reader is created, it will provide access to read blocks from storage.
  */
-public class BlockAsDirReaderBuilder {
-
-    private FileAttribute<Set<PosixFilePermission>> filePerms = FileUtils.defaultPerms;
+public final class BlockAsDirReaderBuilder {
     private final PersistenceStorageConfig config;
+    private FileAttribute<Set<PosixFilePermission>> filePerms = DEFAULT_DIR_PERMISSIONS;
```
jsync-swirlds, line 36: This is bad practice.
1. Copying a constant to a class attribute is an anti-pattern
1. Using a remote default constant when an overload exists to apply defaults _internally_ to the library is an anti-pattern.

As far as the way this builder is setup, the _default_ should be `null`.
That way, the BlockAsDirReader it creates _either_ calls the commons overload that takes no permissions, _or_ calls the overload _with_ permissions (both file _and_ folder, which _must be different_) and has a configured value for that.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
```diff
@@ -16,37 +16,40 @@
...
+        this.config = Objects.requireNonNull(config);
     }
 
     /**
      * Creates a new block reader builder using the minimum required parameters.
      *
      * @param config is required to supply pertinent configuration info for the block reader to
-     *     access storage.
+     *               access storage.
+     *
```
jsync-swirlds, line 47: This is unnecessary.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
```diff
@@ -55,16 +58,17 @@ public static BlockAsDirReaderBuilder newBuilder(@NonNull PersistenceStorageConf
      * and directories.
      *
      * <p>By default, the block reader will use the permissions defined in {@link
-     * FileUtils#defaultPerms}. This method is primarily used for testing purposes. Default values
-     * should be sufficient for production use.
+     * com.hedera.block.common.utils.FileUtilities#DEFAULT_DIR_PERMISSIONS}. This method is
+     * primarily used for testing purposes. Default values should be sufficient for production use.
```
jsync-swirlds, line 62: Just remove the paragraph, and leave the default permissions as an internal detail of the commons module.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
```diff
@@ -55,16 +58,17 @@ public static BlockAsDirReaderBuilder newBuilder(@NonNull PersistenceStorageConf
...
      *
      * @param filePerms the file permissions to use when managing block files and directories.
+     *
```
jsync-swirlds, line 65: unnecessary.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
```diff
@@ -111,22 +109,20 @@ public static void createPathIfNotExists(
     public static void createPathIfNotExists(
             @NonNull final Path toCreate,
             @NonNull final System.Logger.Level logLevel,
-            @NonNull final FileAttribute<Set<PosixFilePermission>> filePermissions,
-            @NonNull final FileAttribute<Set<PosixFilePermission>> folderPermissions,
+            @NonNull final FileAttribute<Set<PosixFilePermission>> permissions,
```
jsync-swirlds, line 112: This is wrong.  Files and folders MUST NOT have the same permissions.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
jsync-swirlds, line 1: All changes to this file are incorrect.
You are re-introducing a severe design flaw by merging the permissions types, and the constants are not supposed to be public _for good reason_.

jsync-swirlds (changes requested):
Still reviewing, but these changes are not correct.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriter.java
```diff
@@ -169,7 +171,12 @@ private void resetState(@NonNull final BlockItem blockItem) throws IOException {
         repairPermissions(blockNodeRootPath);
 
         // Construct the path to the block directory
-        FileUtils.createPathIfNotExists(calculateBlockPath(), DEBUG, filePerms);
+        createPathIfNotExists(
+                calculateBlockPath(),
+                DEBUG,
+                filePerms,
+                BLOCK_NODE_ROOT_DIRECTORY_SEMANTIC_NAME,
+                true);
```
jsync-swirlds, line 179: Just use the override with defaults for this class; it's not a production class.
In the production classes we'll _always_ set file and folder permissions, which should _not_ have a default outside of the configuration records.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
```diff
@@ -16,8 +16,9 @@
 
 package com.hedera.block.server.persistence.storage.write;
 
+import static com.hedera.block.common.utils.FileUtilities.DEFAULT_DIR_PERMISSIONS;
```
jsync-swirlds, line 19: We shouldn't be digging into commons for constants, any constants we need belong in this module.
That said, there should not be defaults for file or folder permissions.  Those need to be in configuration or nowhere.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
```diff
@@ -28,38 +29,37 @@
...
  * Use builder methods to create a {@link BlockWriter} to write blocks to storage.
  *
  * <p>When a block writer is created, it will provide access to write blocks to storage.
  */
-public class BlockAsDirWriterBuilder {
-
+public final class BlockAsDirWriterBuilder {
     private final BlockNodeContext blockNodeContext;
-    private FileAttribute<Set<PosixFilePermission>> filePerms = FileUtils.defaultPerms;
+    private FileAttribute<Set<PosixFilePermission>> filePerms = DEFAULT_DIR_PERMISSIONS;
```
jsync-swirlds, line 42: As noted elsewhere, it's never correct to copy a constant like this into a class attribute. Either the caller supplies the value, or it's null.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
```diff
@@ -28,38 +29,37 @@
...
     private BlockAsDirWriterBuilder(@NonNull final BlockNodeContext blockNodeContext) {
-        this.blockNodeContext = blockNodeContext;
-        PersistenceStorageConfig config =
+        this.blockNodeContext = Objects.requireNonNull(blockNodeContext);
+        final PersistenceStorageConfig config =
                 blockNodeContext.configuration().getConfigData(PersistenceStorageConfig.class);
-
         this.blockRemover =
-                new BlockAsDirRemover(Path.of(config.rootPath()), FileUtils.defaultPerms);
+                new BlockAsDirRemover(Path.of(config.rootPath()), DEFAULT_DIR_PERMISSIONS);
```
jsync-swirlds, line 50: Why are we setting permissions for removing folders?  That seems broken and incorrect.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
```diff
@@ -68,16 +68,17 @@ public static BlockAsDirWriterBuilder newBuilder(
      * and directories.
      *
      * <p>By default, the block writer will use the permissions defined in {@link
-     * FileUtils#defaultPerms}. This method is primarily used for testing purposes. Default values
-     * should be sufficient for production use.
+     * com.hedera.block.common.utils.FileUtilities#DEFAULT_DIR_PERMISSIONS}. This method is
+     * primarily used for testing purposes. Default values should be sufficient for production use.
```
jsync-swirlds, line 72: Just remove this paragraph, it is describing internal details of a different library.
Also, defaults are never acceptable for production in this case; defaults are for testing and development; configuration is for production.

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
```diff
@@ -241,7 +241,7 @@ private void removeBlockItemReadPerms(
     // IOException while allowing the real setPerm() method to remain protected.
     private static final class TestBlockAsDirReader extends BlockAsDirReader {
         public TestBlockAsDirReader(PersistenceStorageConfig config) {
-            super(config, FileUtils.defaultPerms);
+            super(config, DEFAULT_DIR_PERMISSIONS);
```
jsync-swirlds, line 244: Either accept _private_ defaults, or set explicit values in the test.  Having a test explicitly set a value it doesn't control is asking for a brittle test failure.

server/src/test/java/com/hedera/block/server/persistence/storage/remove/BlockAsDirRemoverTest.java
```diff
@@ -76,7 +76,7 @@ public void testRemoveNonExistentBlock() throws IOException, ParseException {
         }
 
         // Remove a block that does not exist
-        final BlockRemover blockRemover = new BlockAsDirRemover(testPath, FileUtils.defaultPerms);
+        final BlockRemover blockRemover = new BlockAsDirRemover(testPath, DEFAULT_DIR_PERMISSIONS);
```
jsync-swirlds, line 79: Same comments as above.

jsync-swirlds (changes requested):
That's the rest reviewed.
Let's chat about this code, as there are obviously some gaps in terms of understanding why some things are written the way they were.

server/src/main/java/com/hedera/block/server/Constants.java
```diff
@@ -16,36 +16,29 @@
...
-    @NonNull public static final String SERVER_STREAMING_METHOD_NAME = "subscribeBlockStream";
+    public static final String SERVER_STREAMING_METHOD_NAME = "subscribeBlockStream";
 
     /** Constant mapped to the singleBlock service method name in the .proto file */
-    @NonNull public static final String SINGLE_BLOCK_METHOD_NAME = "singleBlock";
+    public static final String SINGLE_BLOCK_METHOD_NAME = "singleBlock";
 
     /** Constant defining the block file extension */
-    @NonNull public static final String BLOCK_FILE_EXTENSION = ".blk";
+    public static final String BLOCK_FILE_EXTENSION = ".blk";
```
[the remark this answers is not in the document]
ata-nas: :white_check_mark: Done.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -16,21 +16,22 @@
 
 package com.hedera.block.server.persistence.storage;
 
+import static com.hedera.block.common.utils.FileUtilities.createPathIfNotExists;
```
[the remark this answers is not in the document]
ata-nas: :white_check_mark: Done.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -16,21 +16,22 @@
...
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
 /**
  * Use this configuration across the persistent storage package
  *
  * @param rootPath provides the root path for saving block data, if you want to override it need to
- *     set it as persistence.storage.rootPath
+ *                 set it as persistence.storage.rootPath
```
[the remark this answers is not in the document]
ata-nas: :white_check_mark: Done.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
ata-nas: :white_check_mark: Done.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
ata-nas: :white_check_mark: Done.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
ata-nas: :white_check_mark: Done.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
ata-nas: :white_check_mark: Done.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: I seem to be confused about this.

Technically we will always either create a directory or a file. We will never need both `filePermissions` and `folderPermissions`, we technically need only one set of permissions and to have the boolean to either create a directory or a folder using the permissions we pass. To me, if I will supply file and folder permissions and the boolean `createDir` is true, I will always use only the `folderPermissions` and if the boolean `createDir` is false, then I will always use only the `filePermissions`, so if I will always use only one set of permissions, why should I be supplying both?

I do understand of course that files and directories must not have the same permissions, that is obvious. But why I must be supplying both, when only one will ever be used based on the boolean? What is the design flaw in supplying only one set of permissions? Is it because it is error prone in the sense that I can supply mistakenly directory permissions and have the boolean `createDir` set to false and vice versa? If that is the reason, then I would argue that forcing the caller to supply both file and folder permissions is the same in terms of being error prone, if not even more error prone since we need to now consider to not mistakenly put the file permissions in the place of the folder permissions as well?

But I believe I am missing something here and I fail to understand the design flaw of having to supply only one set of permissions and having the boolean instead of both permissions and having the boolean as well.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: I agree for the constants of the file permissions that should be private, it does make sense for them to be only part of the `FileUtilities` _internally_.

I made them public, because I only wanted to make changes for this PR to `reroute` the already existing places where usages of the now deleted file utility class that was part of the server. I do however agree that some things need to be changed in general and we should have a bit of a different approach for the permission sets as you have suggested in previous comments, for example in the builders and tests and other places you have mentioned.

As for the re-introduction of a severe design flaw by merging the permission types into one parameter, I do fail to understand that part, explained in the [comment above](https://github.com/hashgraph/hedera-block-node/pull/272#discussion_r1816222919).

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
```diff
@@ -16,37 +16,40 @@
 
 package com.hedera.block.server.persistence.storage.read;
 
-import com.hedera.block.server.persistence.storage.FileUtils;
+import static com.hedera.block.common.utils.FileUtilities.DEFAULT_DIR_PERMISSIONS;
```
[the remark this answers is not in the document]
ata-nas: Agreed about that.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriter.java
ata-nas: Agreed about that, but since this class is not a production class, do you envision that we will ever need anything else than the defaults, meaning to use the overloaded method? If not, can we then consider to remove the `filePerms` field, as well as in the builder for this class and also to do the same in the `BlockAsDirReader` and it's builder, since it is the same thing there.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
ata-nas: Please see my response for this in the above comments, for the `BlockAsDirWriterBuilder` and `BlockAsDirWriterBuilder`, it applies here as well.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
ata-nas: Please see my response for this in the above comments, for the `BlockAsDirWriterBuilder` and `BlockAsDirWriterBuilder`, it applies here as well.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
ata-nas: Good question, I am not sure as well. The file perms are used in a method inside the remover `remove(long)`. Based on the comments inside and the single usage it has currently, I believe the method is intended as a best-effort attempt to delete side effects caused by exceptions that could occur during the writing of a `BlockItem` to a file. There is a retry for the write which will first attempt to fix the file permissions and if it fails again then we go into the `remove(long)` method of the remover. Inside the remove method, the file perms that are passed in the constructor are used as `Files.setPosixFilePermissions(blockPath, filePerms.value());` before attempting to do the actual removal of the file. Maybe @mattp-swirldslabs could let us know and give us more context as to why this should be done like so?

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderTest.java
ata-nas: Agreed about that. Based on comments above and on what we shall agree on, if to remove or not the `filePerms` field in the super, this would either become `null` and the actual reader will use the overloaded method internally, or I will supply them by locally having them declared here and not use from the `common` module

server/src/test/java/com/hedera/block/server/persistence/storage/remove/BlockAsDirRemoverTest.java
ata-nas: Also, my comment is the same as my above response, considered only for the tests isolated.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
ata-nas: Same as above comment

ata-nas:
@jsync-swirlds thanks for the reviews.

Keep in mind that when I was doing the changes, my aim was to only `redirect` all current usages of the utilities and constants to the new `common` module that we introduced. That is why I have left everything `as is` and have not made any significant changes, but took the approach to modify a bit the `common` module in order to fit the `redirection`, which I do agree with you is not a good thing to do. 

I have fixed the obvious and commented on the rest, but I agree, lets have a conversation in general about the topics here and others as well, since this would help me to gain more insight into the Hedera philosophy and the team's specific views on what we should aim for in terms of implementation and thought process. This will help avoid such big gaps and so many questions in the future. Thanks! :)

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriter.java
jsync-swirlds: The point of the overload is to enable a class like this to accept defaults without having to manage that default value.

It's fine to remove the class attribute, but it also makes sense to _allow_ for override in a test/development class like this.  If the value is not set (it is `null`) then it was not overridden and the code will call the overload without a permissions.  If the value _is_ set (it is not `null`), then the code calls the version that requires permissions.  The tricky part is that file and folder _must not_ have the same permissions.
For now, just removing the permissions support in this class and calling the overload without the permissions parameters is acceptable.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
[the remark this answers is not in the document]
jsync-swirlds: 1. Yes, both reader and writer are not expected to be used in production, but we mustn't design them as test code, because any operator _might_ choose to use them as the production storage for whatever reason (perhaps their node stores a small subset of block data and they want other processes to have easy access).
1. Always try to design software well, and make the formal design documents (whether Java, C++, Rust, etc...) as readable and clear as you can within constraints.
   * This includes good practice of configurable classes with reasonable defaults managed as private data within the class.
1. It is reasonable to remove the filePerms field, but might be _better_ to leave it in with a default value of `null`.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
jsync-swirlds: As we discussed, it comes down to file and folder permissions and the need to keep them separate.
Also, it seems the commons method should have just been for creating folders, not both files and folders  (and making it more specific clears up the design flaws too).

If we need a similar method for files **in the future**, we're probably better off having two methods, the current method for folders, and a new method for files.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
jsync-swirlds: I would recommend just removing these permissions from the remover.
If we cannot delete a file, we also cannot change permissions (at least in any POSIX compliant system).

jsync-swirlds:
> @jsync-swirlds thanks for the reviews.
> 
> Keep in mind that when I was doing these changes (as well as the ones in the simulator in the other PR), my aim was to only kind of 'redirect' all current usages of the utilities and constants to the new `common` module that we introduced. That is why I have left everything `as is` and have not made any significant changes, but took the approach to modify a bit the `common` module in order to fit the 'redirection', which I do agree with you is not a good thing to do.
> 
> I have fixed the obvious and commented on the rest, but I agree, lets have a conversation in general about the topics here and others as well, since this would help me to gain more insight into the Hedera philosophy and the team's specific views on what we should aim for in terms of implementation and thought process. This will help avoid such big gaps and so many questions in the future. Thanks! :)

I understand the intent.  I think (as we discussed) that you would be better off (and the result will be better designed) to keep the shared library as clean as practical, and make whatever changes are needed in the server to correctly use that library.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -52,12 +53,10 @@ public record PersistenceStorageConfig(@ConfigProperty(defaultValue = "") String
...
-        if (Files.notExists(path)) {
-            try {
-                FileUtils.createPathIfNotExists(path, ERROR, FileUtils.defaultPerms);
-            } catch (IOException e) {
-                throw new RuntimeException(e);
-            }
+        try {
+            createPathIfNotExists(path, ERROR, BLOCK_NODE_ROOT_DIRECTORY_SEMANTIC_NAME, true);
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
ata-nas: ✅ Done.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: @jsync-swirlds agreed on that, thanks for the clarification, it is best currently to create only folders and if we need a file creation method, we will introduce it in the future.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
ata-nas: ✅ Done.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: ✅ Done.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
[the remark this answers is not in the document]
ata-nas: ✅ Done.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReaderBuilder.java
ata-nas: ✅ Done.

common/src/main/java/com/hedera/block/common/utils/FileUtilities.java
ata-nas: Noted, thanks for the explanation, looks like I was on the right track as far as the reasoning behind the design flaw. Since it seems that currently the only usage for the create path utility method is to create directories, I removed support for the file path creation functionality, as it really is the best in terms of design. In the future if it is needed, we could always add a new method that will do the creation of file paths.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
ata-nas: ✅ Done.

server/src/test/java/com/hedera/block/server/persistence/storage/remove/BlockAsDirRemoverTest.java
ata-nas: ✅ Done.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriter.java
ata-nas: ✅ Done. Keep in mind that on a second inspection, the perms were used not only for the create path utility, but also for other reasons both in `BlockAsDirWriter` and `BlockAsDirReader`. That being said, if I left the field `null` it would have been quite a pain again, since I do need to provide permissions for the other usages of the `filePerms` field OR I must have not executed the logic that needed the non-null `filePerms`. I do not think the latter must happen, so instead of the possibility of the field to be `null`, I provide the field sane defaults (local to the respective class in order to fit the philosophy) in case the `filePerms` argument of the constructor is `null`.

jsync-swirlds: file permissions is one of those things that defaults are a bit risky.
I still recommend changing the code to handle a null value (and not have a local default value); that can be done separately, however.

ata-nas:
Handle [this comment](https://github.com/hashgraph/hedera-block-node/pull/271#discussion_r1819364675) in this PR here.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriterBuilder.java
jsync-swirlds: I cannot say why the original set permissions for delete, but there is no possible scenario where it makes sense. We cannot set permissions in a way that changes whether we can remove the file.
There is no reason to be nervous at this stage, as we aren't deploying anything live and can always add things back if we later determine it is necessary.

I would strongly recommend removing that test (which was almost certainly added only to achieve a high coverage metric, rather than implementing a requirement or constraint) and removing the unnecessary value.

Side note, this is why sometimes less tests are better, we have code that someone is afraid to remove because it has a test, but both code and test are unequivocally unnecessary, as the value in question cannot be useful in any likely scenario.

This also highlights the value of a practice I've long advocated.  For every unit test scenario, document the test (or test case) with a clear description of the _requirement_ or _constraint_ being verified.  If a test does not verify a _requirement_ or _constraint_, then it should not exist.

ata-nas: @jsync-swirlds thanks for the suggestion, I will remove it, to me also the functionality made no sense especially given the context that you provide in this thread. But I wanted to be absolutely sure nevertheless before doing it. :)

ata-nas: ✅ Done.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReader.java
```diff
@@ -57,12 +57,10 @@ class BlockAsDirReader implements BlockReader<Block> {
...
      *
      * @param config the configuration to retrieve the block node root path
-     * @param filePerms the file permissions to set on the block node root path
+     * @param filePerms the file permissions to set on the block node root path, default  will be used if null provided
      */
     BlockAsDirReader(
-            @NonNull final PersistenceStorageConfig config,
-            @NonNull final FileAttribute<Set<PosixFilePermission>> filePerms) {
-
+            @NonNull final PersistenceStorageConfig config, final FileAttribute<Set<PosixFilePermission>> filePerms) {
```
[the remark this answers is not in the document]
ata-nas: I forgot to mention that in the threads, was thinking about that today as well, I was thinking about just `permissions`, but `folderPermissions` is more descriptive. Since #271 will be first, I will change the names to your suggestions when I resolve the merge conflicts and resolve your other comment about the tests from #271.

ata-nas:
address [this comment](https://github.com/hashgraph/hedera-block-node/pull/271#discussion_r1819543073) as well

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriter.java
jsync-swirlds: I am seeing a number of possible errors as I look at the full _usage_ of this permissions variable.  It seems everything in this class is assuming identical permissions for both files and folders.
This class probably needs some added overhaul to handle all this correctly, so perhaps a separate issue is appropriate?

Additional Discussion:
This is _exactly_ why having a single permissions value is a problem.
I recommend, if we continue to set permissions at all (see below), that _this class_ split file and folder permissions to make them expressly different, and ensure that files _do not_ have the execute bit set at any point.  I also recommend that group and other permissions be limited to, at most, read for files and read+execute for folders.

Ideally just _not setting permissions_ for files is better. The server should be relying on system umask (set in the launch script for the server) to correctly set permissions for new files and folders (something like umask 033 would be appropriate). This would place system file permissions _properly_ in the control of system operations.

jsync-swirlds (approved):
There are a couple items noted that should be addressed as a separate issue.
Please file issues for those prior to merging this PR.

server/src/main/java/com/hedera/block/server/persistence/storage/write/BlockAsDirWriter.java
ata-nas: @jsync-swirlds I opened an issue #319 for this thread of comments here.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsDirReader.java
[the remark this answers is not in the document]
ata-nas: ✅ Done.

mattp-swirldslabs (approved):
Looks good

jsync-swirlds (approved):

PR #284, opened 2024-10-21
feat: Move out concrete working implementation and add working modes

### LINKED ISSUE #239: Refactor start and stop functionality of the simulator
### Problem

Currently stop method does nothing else, apart from setting a variable to false.
Start method Only initiates a while loop and start streaming already loaded data.

### Solution

- [ ] Stop method should stop the streaming or consuming
- [ ] Start method should start streaming or consuming (depending on configuration
  - [ ] Extract current streaming logic in outside method
  - [ ] Extract future consumer logic in outside method
 - [ ] Start method should actually, depending on configuration, setup the channel, otherwise it produces errors.

### Alternatives

_No response_

georgi-l95:
**Description**:
WIP -> missing unit tests

This PR aims mainly to make the maintanence and using of simulator (as test driver) easier. It moves away the implementation of the working modes and exposes only abstractions. So that all concrete implementations can be refactored (if needed), without the main application, knowing or caring about it. 
Main simulator app should only be responsible for starting the processes and nothing more. Wherther the simulator will work in publisher or consumer mode, or both is not responsibility of the main class.

Also adds E2E test for data persistence.

**Related issue(s)**:

Fixes #239 

**Notes for reviewer**:

**Checklist**


georgi-l95:
Superseeded by https://github.com/hashgraph/hedera-block-node/pull/307

PR #327, opened 2024-10-30
feat: add error handling in publisher

### LINKED ISSUE #315: Add error handling mechanism to the Simulator to stop sending data
As a simulator user
I want the simulator to stop sending items when the simulator receives a message through `PublishStreamObserver.onError()` method
So troubleshooting is easier since I know when an error occurred

Background:
* I noticed while testing #178 that the simulator continued to send data even though the server was notifying the simulator through the `PublishStreamObserver.onError()` method to stop sending items.

Suggested approach:
* Error messages from the server are received through `PublishStreamObserver.onError()`
* With the current design, we must signal to components outside this object that they should no longer send data to the server. Here's how I did it:

```
public class PublishStreamObserver implements StreamObserver<PublishStreamResponse> {

    private final Logger logger = System.getLogger(getClass().getName());
    private final AtomicBoolean allowNext;

    /** Creates a new PublishStreamObserver instance. */
    public PublishStreamObserver(final AtomicBoolean allowNext) {
        this.allowNext = allowNext;
    }

    /** what will the stream observer do with the response from the server */
    @Override
    public void onNext(PublishStreamResponse publishStreamResponse) {
        logger.log(Logger.Level.INFO, "Received Response: " + publishStreamResponse.toString());
    }

    /** what will the stream observer do when an error occurs */
    @Override
        logger.log(Logger.Level.ERROR, "Error: " + throwable.toString());
        allowNext.set(false);
    }

    /** what will the stream observer do when the stream is completed */
    @Override
    public void onCompleted() {
        logger.log(Logger.Level.DEBUG, "Completed");
    }
}
```
```
public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {

    private final System.Logger LOGGER = System.getLogger(getClass().getName());

    private final BlockStreamServiceGrpc.BlockStreamServiceStub stub;
    private final StreamObserver<PublishStreamRequest> requestStreamObserver;
    private final AtomicBoolean allowNext = new AtomicBoolean(true);

    /**
     * Creates a new PublishStreamGrpcClientImpl instance.
     *
     * @param grpcConfig the gRPC configuration
     */
    @Inject
    public PublishStreamGrpcClientImpl(
            @NonNull GrpcConfig grpcConfig, @NonNull MetricsService metricsService) {
        ManagedChannel channel =
                ManagedChannelBuilder.forAddress(grpcConfig.serverAddress(), grpcConfig.port())
                        .usePlaintext()
                        .build();
        stub = BlockStreamServiceGrpc.newStub(channel);

        PublishStreamObserver publishStreamObserver = new PublishStreamObserver(allowNext);
        requestStreamObserver = stub.publishBlockStream(publishStreamObserver);
        this.metricsService = metricsService;
    }

    /**
     * The PublishStreamObserver class implements the StreamObserver interface to observe the
     * stream.
     */
    @Override
    public boolean streamBlockItem(BlockItem blockItem) {
        if (allowNext.get()) {
            requestStreamObserver.onNext(
                    PublishStreamRequest.newBuilder()
                            .setBlockItem(Translator.fromPbj(blockItem))
                            .build());
            metricsService.get(LiveBlockItemsSent).increment();

        } else {
            LOGGER.log(ERROR, "Not allowed to send next block item");
        }

        return allowNext.get();
    }

@Override
    public boolean streamBlock(Block block) {
        for (int count = 0; count < 2; count++) {
            if (!streamBlockItem(block.items().get(count))) {
                LOGGER.log(ERROR, "Count was: " + count);

                                if (count == 0) {
                                    LOGGER.log(ERROR, "First block item: " +
                 block.items().get(count));
                                } else {
                                    LOGGER.log(ERROR, "Previous block item: " +
                 block.items().get(count - 1));
                                }
                return false;
            }
        }

        return true;
    }

}
```

```
BlockStreamSimulatorApp calling streamBlockItem()

 private void constantRateStreaming()
            throws InterruptedException, IOException, BlockSimulatorParsingException {
        int delayMSBetweenBlockItems = delayBetweenBlockItems / 1_000_000;
        int delayNSBetweenBlockItems = delayBetweenBlockItems % 1_000_000;
        boolean streamBlockItem = true;
        int blockItemsStreamed = 0;

        while (streamBlockItem) {
            // get block item
            BlockItem blockItem = blockStreamManager.getNextBlockItem();

            if (blockItem == null) {
                LOGGER.log(INFO, "Block Stream Simulator has reached the end of the block items");
                break;
            }

            if (!publishStreamGrpcClient.streamBlockItem(blockItem)) {
                LOGGER.log(INFO, "Exiting");
                break;
            }

            blockItemsStreamed++;

            Thread.sleep(delayMSBetweenBlockItems, delayNSBetweenBlockItems);

            if (blockItemsStreamed >= blockStreamConfig.maxBlockItemsToStream()) {
                LOGGER.log(
                        INFO,
                        "Block Stream Simulator has reached the maximum number of block items to"
                                + " stream");
                streamBlockItem = false;
            }
        }
    }
```

* These objects share an `AtomicBoolean` flag to check before sending items

georgi-l95:
**Description**:
This PR adds simple error handling mechanism in the publisher, when onError is invoked.

**Related issue(s)**:

Fixes #315 

**Notes for reviewer**:
This change is probably a very first version of the error handling mechanism that we want to have. We'd probably want to add checks and do different things, depending on the error. But as we are still in the early stage, this will do for now.

**Checklist**

### DIFF
--- a/simulator/src/main/java/com/hedera/block/simulator/grpc/GrpcInjectionModule.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/grpc/GrpcInjectionModule.java
@@ -18,6 +18,8 @@ package com.hedera.block.simulator.grpc;
 
 import dagger.Binds;
 import dagger.Module;
+import dagger.Provides;
+import java.util.concurrent.atomic.AtomicBoolean;
 import javax.inject.Singleton;
 
 /** The module used to inject the gRPC client. */
@@ -32,6 +34,16 @@ public interface GrpcInjectionModule {
      */
     @Singleton
     @Binds
-    PublishStreamGrpcClient bindPublishStreamGrpcClient(
-            PublishStreamGrpcClientImpl publishStreamGrpcClient);
+    PublishStreamGrpcClient bindPublishStreamGrpcClient(PublishStreamGrpcClientImpl publishStreamGrpcClient);
+
+    /**
+     * Provides the stream enabled flag
+     *
+     * @return the stream enabled flag
+     */
+    @Singleton
+    @Provides
+    static AtomicBoolean provideStreamEnabledFlag() {
+        return new AtomicBoolean(true);
+    }
 }
--- a/simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
@@ -35,6 +35,7 @@ import io.grpc.ManagedChannelBuilder;
 import io.grpc.stub.StreamObserver;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import javax.inject.Inject;
 
 /**
@@ -47,6 +48,7 @@ public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
     private StreamObserver<PublishStreamRequest> requestStreamObserver;
     private final BlockStreamConfig blockStreamConfig;
     private final GrpcConfig grpcConfig;
+    private final AtomicBoolean allowNext = new AtomicBoolean(true);
     private ManagedChannel channel;
     private final MetricsService metricsService;
 
@@ -76,7 +78,7 @@ public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
                 .usePlaintext()
                 .build();
         BlockStreamServiceGrpc.BlockStreamServiceStub stub = BlockStreamServiceGrpc.newStub(channel);
-        PublishStreamObserver publishStreamObserver = new PublishStreamObserver();
+        PublishStreamObserver publishStreamObserver = new PublishStreamObserver(allowNext);
         requestStreamObserver = stub.publishBlockStream(publishStreamObserver);
     }
 
@@ -110,6 +112,9 @@ public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
      */
     @Override
     public boolean streamBlock(Block block) {
+        if (!allowNext.get()) {
+            return false;
+        }
         List<com.hedera.hapi.block.stream.protoc.BlockItem> blockItemsProtoc = new ArrayList<>();
         for (BlockItem blockItem : block.items()) {
             blockItemsProtoc.add(Translator.fromPbj(blockItem));
--- a/simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
@@ -48,7 +48,7 @@ public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
     private StreamObserver<PublishStreamRequest> requestStreamObserver;
     private final BlockStreamConfig blockStreamConfig;
     private final GrpcConfig grpcConfig;
-    private final AtomicBoolean allowNext = new AtomicBoolean(true);
+    private final AtomicBoolean streamEnabled;
     private ManagedChannel channel;
     private final MetricsService metricsService;
 
@@ -58,15 +58,18 @@ public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
      * @param grpcConfig the gRPC configuration
      * @param blockStreamConfig the block stream configuration
      * @param metricsService the metrics service
+     * @param streamEnabled the flag responsible for enabling and disabling of the streaming
      */
     @Inject
     public PublishStreamGrpcClientImpl(
             @NonNull final GrpcConfig grpcConfig,
             @NonNull final BlockStreamConfig blockStreamConfig,
-            @NonNull final MetricsService metricsService) {
+            @NonNull final MetricsService metricsService,
+            @NonNull final AtomicBoolean streamEnabled) {
         this.grpcConfig = requireNonNull(grpcConfig);
         this.blockStreamConfig = requireNonNull(blockStreamConfig);
         this.metricsService = requireNonNull(metricsService);
+        this.streamEnabled = requireNonNull(streamEnabled);
     }
 
     /**
@@ -78,7 +81,7 @@ public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
                 .usePlaintext()
                 .build();
         BlockStreamServiceGrpc.BlockStreamServiceStub stub = BlockStreamServiceGrpc.newStub(channel);
-        PublishStreamObserver publishStreamObserver = new PublishStreamObserver(allowNext);
+        PublishStreamObserver publishStreamObserver = new PublishStreamObserver(streamEnabled);
         requestStreamObserver = stub.publishBlockStream(publishStreamObserver);
     }
 
@@ -112,7 +115,7 @@ public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
      */
     @Override
     public boolean streamBlock(Block block) {
-        if (!allowNext.get()) {
+        if (!streamEnabled.get()) {
             return false;
         }
         List<com.hedera.hapi.block.stream.protoc.BlockItem> blockItemsProtoc = new ArrayList<>();
--- a/simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
@@ -17,6 +17,7 @@
 package com.hedera.block.simulator.grpc;
 
 import com.hedera.hapi.block.protoc.PublishStreamResponse;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
 import java.lang.System.Logger;
@@ -29,11 +30,14 @@ import java.util.concurrent.atomic.AtomicBoolean;
 public class PublishStreamObserver implements StreamObserver<PublishStreamResponse> {
 
     private final Logger logger = System.getLogger(getClass().getName());
-    private final AtomicBoolean allowNext;
+    private final AtomicBoolean streamEnabled;
 
-    /** Creates a new PublishStreamObserver instance. */
-    public PublishStreamObserver(final AtomicBoolean allowNext) {
-        this.allowNext = allowNext;
+    /** Creates a new PublishStreamObserver instance.
+     *
+     * @param streamEnabled is responsible for signaling, whether streaming should continue
+     */
+    public PublishStreamObserver(final AtomicBoolean streamEnabled) {
+        this.streamEnabled = streamEnabled;
     }
 
     /** what will the stream observer do with the response from the server */
@@ -44,10 +48,13 @@ public class PublishStreamObserver implements StreamObserver<PublishStreamRespon
 
     /** Responsible for stream observer behaviour, in case of error. For now, we will stop the stream for every error. In the future we'd want to have a retry mechanism depending on the error. */
     @Override
-    public void onError(Throwable throwable) {
-        Status status = Status.fromThrowable(throwable);
-        logger.log(Logger.Level.ERROR, "Error: " + throwable.toString() + " with status code: " + status.getCode());
-        allowNext.set(false);
+    public void onError(@NonNull final Throwable streamError) {
+        streamEnabled.set(false);
+        Status status = Status.fromThrowable(streamError);
+        logger.log(
+                Logger.Level.ERROR,
+                "Error %s with status code %s.".formatted(status.getCode(), streamError),
+                streamError);
     }
 
     /** what will the stream observer do when the stream is completed */
--- a/simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
@@ -17,8 +17,10 @@
 package com.hedera.block.simulator.grpc;
 
 import com.hedera.hapi.block.protoc.PublishStreamResponse;
+import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
 import java.lang.System.Logger;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * The PublishStreamObserver class provides the methods to observe the stream of the published
@@ -27,9 +29,12 @@ import java.lang.System.Logger;
 public class PublishStreamObserver implements StreamObserver<PublishStreamResponse> {
 
     private final Logger logger = System.getLogger(getClass().getName());
+    private final AtomicBoolean allowNext;
 
     /** Creates a new PublishStreamObserver instance. */
-    public PublishStreamObserver() {}
+    public PublishStreamObserver(final AtomicBoolean allowNext) {
+        this.allowNext = allowNext;
+    }
 
     /** what will the stream observer do with the response from the server */
     @Override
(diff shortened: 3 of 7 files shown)


simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
```diff
@@ -27,21 +29,28 @@
...
 
-    /** what will the stream observer do when an error occurs */
+    /** Responsible for stream observer behaviour, in case of error. For now, we will stop the stream for every error. In the future we'd want to have a retry mechanism depending on the error. */
     @Override
     public void onError(Throwable throwable) {
-        logger.log(Logger.Level.ERROR, "Error: " + throwable.toString());
-        // @todo(286) - handle the error
+        Status status = Status.fromThrowable(throwable);
+        logger.log(Logger.Level.ERROR, "Error: " + throwable.toString() + " with status code: " + status.getCode());
+        allowNext.set(false);
```
mattp-swirldslabs, line 53: Maybe put `allowNext.set(false)` as the first call in this method. We want to shut it down as quickly as possible.

simulator/src/main/java/com/hedera/block/simulator/mode/PublisherModeHandler.java
```diff
@@ -80,13 +80,17 @@ public void start() throws BlockSimulatorParsingException, IOException, Interrup
...
 
         Block nextBlock = blockStreamManager.getNextBlock();
         while (nextBlock != null) {
             long startTime = System.nanoTime();
-            publishStreamGrpcClient.streamBlock(nextBlock);
+            final boolean streamSuccess = publishStreamGrpcClient.streamBlock(nextBlock);
+            if (!streamSuccess) {
+                LOGGER.log(System.Logger.Level.INFO, "Block Stream Simulator stopped streaming due to errors.");
+                break;
+            }
```
mattp-swirldslabs, line 92: I think you can skip the variable assignment and just have:
```
if (!publishStreamGrpcClient.streamBlock(nextBlock)) {
   // log message
  break;
}
```

simulator/src/main/java/com/hedera/block/simulator/mode/PublisherModeHandler.java
```diff
@@ -120,7 +123,11 @@ private void constantRateStreaming() throws InterruptedException, IOException, B
                 break;
             }
 
-            publishStreamGrpcClient.streamBlock(block);
+            final boolean streamSuccess = publishStreamGrpcClient.streamBlock(block);
+            if (!streamSuccess) {
+                LOGGER.log(System.Logger.Level.INFO, "Block Stream Simulator stopped streaming due to errors.");
+                break;
+            }
```
mattp-swirldslabs, line 130: I think you can skip the boolean variable assignment (see comment above)

mattp-swirldslabs (dismissed):
Looks good

jsync-swirlds (commented):
The need to reach through reflection in unit tests (always a code smell) reveals an opportunity for improvement.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
georgi-l95: Done ✅

simulator/src/main/java/com/hedera/block/simulator/mode/PublisherModeHandler.java
georgi-l95: Changed, although I usually like to assign, just for better readability.

simulator/src/main/java/com/hedera/block/simulator/mode/PublisherModeHandler.java
georgi-l95: Changed, although I usually like to assign, just for better readability.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
```diff
@@ -27,21 +29,28 @@
...
     }
 
-    /** what will the stream observer do when an error occurs */
+    /** Responsible for stream observer behaviour, in case of error. For now, we will stop the stream for every error. In the future we'd want to have a retry mechanism depending on the error. */
     @Override
     public void onError(Throwable throwable) {
-        logger.log(Logger.Level.ERROR, "Error: " + throwable.toString());
-        // @todo(286) - handle the error
+        Status status = Status.fromThrowable(throwable);
+        logger.log(Logger.Level.ERROR, "Error: " + throwable.toString() + " with status code: " + status.getCode());
```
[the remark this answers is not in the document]
georgi-l95: Agree, changed. ✅

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
```diff
@@ -27,21 +29,28 @@
...
     /** what will the stream observer do with the response from the server */
     @Override
     public void onNext(PublishStreamResponse publishStreamResponse) {
         logger.log(Logger.Level.INFO, "Received Response: " + publishStreamResponse.toString());
     }
 
-    /** what will the stream observer do when an error occurs */
+    /** Responsible for stream observer behaviour, in case of error. For now, we will stop the stream for every error. In the future we'd want to have a retry mechanism depending on the error. */
     @Override
     public void onError(Throwable throwable) {
```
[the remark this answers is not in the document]
georgi-l95: True, changed ✅

simulator/src/test/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImplTest.java
```diff
@@ -77,6 +79,28 @@ void streamBlock() {
...
+            allowNextField.setAccessible(true);
+            AtomicBoolean allowNext = (AtomicBoolean) allowNextField.get(publishStreamGrpcClient);
+            allowNext.set(false);
+        } finally {
+            allowNextField.setAccessible(false);
+        }
+
+        boolean result = publishStreamGrpcClient.streamBlock(block);
+        assertFalse(result);
+    }
```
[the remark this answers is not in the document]
georgi-l95: About the name of the variable, i agree, changed it.
I've also refactored, by adding it in the constructor, alothough I really dislike it, because it feels like I'm altering and introducing coupling in production code, just for the sake of easier testing. Tried different strategies and none feels right for testing, therefore injected in constructor is the way for now.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
```diff
@@ -42,6 +43,7 @@ public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
     private StreamObserver<PublishStreamRequest> requestStreamObserver;
     private final BlockStreamConfig blockStreamConfig;
     private final GrpcConfig grpcConfig;
+    private final AtomicBoolean allowNext = new AtomicBoolean(true);
```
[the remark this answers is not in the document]
georgi-l95: Answer in the previous comment is valid for here also 👍

simulator/src/main/java/com/hedera/block/simulator/grpc/GrpcInjectionModule.java
```diff
@@ -32,6 +34,16 @@ public interface GrpcInjectionModule {
...
+    /**
+     * Provides the stream enabled flag
+     *
+     * @return the stream enabled flag
+     */
+    @Singleton
+    @Provides
+    static AtomicBoolean provideStreamEnabledFlag() {
+        return new AtomicBoolean(true);
+    }
```
ata-nas, line 48: I am not quite sure if we should ever declare and DI core java classes like that. Maybe a better approach would be to do `@Provides` at the place where this `AtomicBoolean` must be injected (I believe this is the same file at line 37) and instantiate it there, while getting the rest of the `dagger` DI enabled dependencies as method arguments.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
```diff
@@ -66,7 +72,7 @@ public void init() {
                 .usePlaintext()
                 .build();
         BlockStreamServiceGrpc.BlockStreamServiceStub stub = BlockStreamServiceGrpc.newStub(channel);
-        PublishStreamObserver publishStreamObserver = new PublishStreamObserver();
+        PublishStreamObserver publishStreamObserver = new PublishStreamObserver(streamEnabled);
```
ata-nas, line 75: I do not think sharing this `streamEnabled` instance like that between `this` and the `publishStreamObserver` is a good idea. Yes, the state change will be visible, but think about how error prone could that be. Without the context as to what this variable intends to do, in the future it will be misused. Private members _must_ remain so and interaction with the outside world should not happen like that. This variable is not updated anywhere in this class, it is only used to read from, but updated in another place. That being said, this should really be a property of the place that is responsible for updating it.

It would be better if the `streamEnabled` is a private member of the `PublishStreamObserver` and to be read via a method call from that object like `publisStreamObserver.isStreamEnabled()` rather than shared between two places and read directly from the instance on line 104 of this class.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
```diff
@@ -95,6 +101,9 @@ public boolean streamBlockItem(List<BlockItem> blockItems) {
      */
     @Override
     public boolean streamBlock(Block block) {
+        if (!streamEnabled.get()) {
+            return false;
+        }
```
ata-nas, line 106: See above comment for `streamEnabled` shared instance.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
```diff
@@ -27,21 +30,31 @@
...
+    private final AtomicBoolean streamEnabled;
 
-    /** Creates a new PublishStreamObserver instance. */
-    public PublishStreamObserver() {}
+    /** Creates a new PublishStreamObserver instance.
+     *
+     * @param streamEnabled is responsible for signaling, whether streaming should continue
+     */
+    public PublishStreamObserver(final AtomicBoolean streamEnabled) {
+        this.streamEnabled = streamEnabled;
```
ata-nas, line 40: Maybe a `@NonNull` and a null check would be good here?

simulator/src/test/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImplTest.java
```diff
@@ -38,13 +40,15 @@ class PublishStreamGrpcClientImplTest {
...
     BlockStreamConfig blockStreamConfig;
+    AtomicBoolean streamEnabled;
 
     @BeforeEach
     void setUp() throws IOException {
 
         grpcConfig = TestUtils.getTestConfiguration().getConfigData(GrpcConfig.class);
         blockStreamConfig = TestUtils.getTestConfiguration(Map.of("blockStream.blockItemsBatchSize", "2"))
                 .getConfigData(BlockStreamConfig.class);
+        streamEnabled = new AtomicBoolean(true);
```
ata-nas, line 51: Based on [this comment](https://github.com/hashgraph/hedera-block-node/pull/327#discussion_r1824545774) if you chose to remove the boolean from the constructor of the grpc client and to DI the `PublishStreamObserver`, you can now mock the stream observer and tell it what to do, in that way you will have control over the internal boolean with mockito.

simulator/src/test/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImplTest.java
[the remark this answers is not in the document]
jsync-swirlds: Injection is the preferred mechanism for things that need to modified when unit testing, but yes, injecting an atomic boolean is not ideal.
It's most likely just a short-term quirk, however. I suspect that we'll eventually replace this single value with something like a stream control object in future work.

simulator/src/main/java/com/hedera/block/simulator/grpc/GrpcInjectionModule.java
jsync-swirlds: I believe we are just seeing the _incomplete_ nature of the system at the moment, here.  This shouldn't be a single `AtomicBoolean` long term, that's just a temporary expedient.  Longer term we'll almost certainly have some sort of "stream control" object that includes a number of different control attributes and we'll replace this single value with that more complete object.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
jsync-swirlds: I think we're in an in-between state at the moment where all DI options will look a bit ugly.
Injecting the observer requires a lot of extra setup that shouldn't be externalized.
Injecting the boolean looks a bit ugly, but _should_ be a temporary expedient.

Longer term I hope we'll collect this and several other items into some sort of stream control object that makes more sense and looks much cleaner when injected.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
```diff
@@ -27,21 +30,31 @@
...
     @Override
-    public void onError(Throwable throwable) {
-        logger.log(Logger.Level.ERROR, "Error: " + throwable.toString());
-        // @todo(286) - handle the error
+    public void onError(@NonNull final Throwable streamError) {
+        streamEnabled.set(false);
+        Status status = Status.fromThrowable(streamError);
+        logger.log(
+                Logger.Level.ERROR,
+                "Error %s with status code %s.".formatted(status.getCode(), streamError),
```
jsync-swirlds, line 56: I think my suggestion had the values reversed, my apologies.
```suggestion
                "Error %s with status code %s.".formatted(streamError, status),
```

jsync-swirlds (commented):
Much improved, thank you.

simulator/src/main/java/com/hedera/block/simulator/grpc/GrpcInjectionModule.java
ata-nas: @jsync-swirlds I do agree that the ticket aims to be a temporary fix and not the final solution. That being said, I did also aim to make my comments based on that.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
ata-nas: @jsync-swirlds I agree with the in-between state, we also discussed that with @georgi-l95 as well. For me his changes are ok based on that fact, my proposal aims to clean up a bit and make easier to test, and I am aware of the fact that this is absolutely not a more permanent production ready solution. Also, yes, lets discuss in the future what the approach should be, not only about the control flow, but error handling approach in general.

@georgi-l95 as we spoke, the proposed is up to you if you see fit to implement, I agree with both either way, since with the proposed we achieve _not_ much more than maybe a bit better testing for now, but maybe you can consider only to not share the atomic boolean but poll it via a method as shown in [this comment](https://github.com/hashgraph/hedera-block-node/pull/327#discussion_r1824547888) since this change is very small, you simply delete the field from the client and call a method instead (the atomic boolean remains as a constructor parameter for the client to be able to pass the tests).

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
georgi-l95: You're right, thanks for noticing 👍

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamObserver.java
georgi-l95: Added. ✅

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
georgi-l95: Let's go atm with the approach where we introduce this variable via injection in constructor. It's a hack, but will do the job for now. We will 100% revisit and refactor this whole flow in the future.

ata-nas: @georgi-l95 agreed on my side.

ata-nas (approved):
From my side, the comments are resolved. Agreed that this is a temporary change that aims to immediately solve us the problem and we need to revisit this in the future.

mattp-swirldslabs (approved):

PR #330, opened 2024-11-07
test: Add E2E test and enhance simulator flexibility

### LINKED ISSUE #186: E2E Tests: Endpoint Accessibility
**Objective:** Ensure all gRPC endpoints (services) are available and responding.

| Test Case ID | Test Name | Test Description | Expected Outcome | Negative Scenario
|--------|--------|--------|--------|--------|
| TC_GRPC_003 | `verifyPublishBlockStreamEndpoint` | Test the `publishBlockStream` RPC by publishing a valid block stream. | Server acknowledges each block with `BlockAcknowledgement`. | Send `BlockItems` out of order.

georgi-l95:
**Description**:
This PR adds new functionality aimed at making the simulator a better and more flexible test driver for E2E tests. It achieves this by:

- Adding a new StreamStatus data object, which collects and stores information about the recently published/consumed blocks, as well as the latest statuses.
- Exposing the necessary methods to make this usable in other modules, primarily the E2E Tests module.
- Adding a new E2E test that covers the positive scenario where the simulator sends blocks to the Block Node, which responds with the correct response.
- Adds a new CustomThreadPoolExecutor to handle simulator threads, for now I'm just handling thrown errors, but in the future we might extend, depending on our needs.

**Related issue(s)**:

Fixes #186 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,8 @@ server/src/test/resources/test_output/
 
 # simulator files, this files are un-tarred before build
 /simulator/src/main/resources/block-0.0.3/
+/common/bin
+/server/bin
+/simulator/bin
+/stream/bin
+/suites/bin
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 
 # Mobile Tools for Java (J2ME)
 .mtj.tmp/
+bin/
 
 # Package Files #
 *.jar
@@ -58,8 +59,3 @@ server/src/test/resources/test_output/
 
 # simulator files, this files are un-tarred before build
 /simulator/src/main/resources/block-0.0.3/
-/common/bin
-/server/bin
-/simulator/bin
-/stream/bin
-/suites/bin
--- a/simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
@@ -19,6 +19,8 @@ package com.hedera.block.simulator.config.data;
 import java.util.ArrayList;
 import java.util.List;
 
+import static java.util.Objects.requireNonNull;
+
 /**
  * Represents the status of the stream.
  *
@@ -87,7 +89,8 @@ public record StreamStatus(
          * @return the builder instance
          */
         public Builder lastKnownPublisherStatuses(List<String> lastKnownPublisherStatuses) {
-            this.lastKnownPublisherStatuses = lastKnownPublisherStatuses;
+            requireNonNull(lastKnownPublisherStatuses);
+            this.lastKnownPublisherStatuses = new ArrayList<>(lastKnownPublisherStatuses);
             return this;
         }
 
@@ -98,7 +101,8 @@ public record StreamStatus(
          * @return the builder instance
          */
         public Builder lastKnownConsumersStatuses(List<String> lastKnownConsumersStatuses) {
-            this.lastKnownConsumersStatuses = lastKnownConsumersStatuses;
+            requireNonNull(lastKnownConsumersStatuses);
+            this.lastKnownConsumersStatuses = new ArrayList<>(lastKnownConsumersStatuses);
             return this;
         }
 
--- a/simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.block.simulator.config.data;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents the status of the stream.
+ *
+ * @param publishedBlocks the number of published blocks
+ * @param consumedBlocks the number of consumed blocks
+ * @param lastKnownPublisherStatuses the last known publisher statuses
+ * @param lastKnownConsumersStatuses the last known consumers statuses
+ */
+public record StreamStatus(
+        int publishedBlocks,
+        int consumedBlocks,
+        List<String> lastKnownPublisherStatuses,
+        List<String> lastKnownConsumersStatuses) {
+
+    /**
+     * Creates a new {@link Builder} instance for constructing a {@code StreamStatus}.
+     *
+     * @return a new {@code Builder}
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * A builder for creating instances of {@link StreamStatus}.
+     */
+    public static class Builder {
+        private int publishedBlocks = 0;
+        private int consumedBlocks = 0;
+        private List<String> lastKnownPublisherStatuses = new ArrayList<>();
+        private List<String> lastKnownConsumersStatuses = new ArrayList<>();
+
+        /**
+         * Creates a new instance of the {@code Builder} class with default configuration values.
+         */
+        public Builder() {
+            // Default constructor
+        }
+
+        /**
+         * Sets the number of published blocks.
+         *
+         * @param publishedBlocks the number of published blocks
+         * @return the builder instance
+         */
+        public Builder publishedBlocks(int publishedBlocks) {
+            this.publishedBlocks = publishedBlocks;
+            return this;
+        }
+
+        /**
+         * Sets the number of consumed blocks.
+         *
+         * @param consumedBlocks the number of consumed blocks
+         * @return the builder instance
+         */
+        public Builder consumedBlocks(int consumedBlocks) {
+            this.consumedBlocks = consumedBlocks;
+            return this;
+        }
+
+        /**
+         * Sets the last known publisher statuses.
+         *
+         * @param lastKnownPublisherStatuses the last known publisher statuses
+         * @return the builder instance
+         */
+        public Builder lastKnownPublisherStatuses(List<String> lastKnownPublisherStatuses) {
+            this.lastKnownPublisherStatuses = lastKnownPublisherStatuses;
+            return this;
+        }
+
+        /**
+         * Sets the last known consumers statuses.
+         *
+         * @param lastKnownConsumersStatuses the last known consumers statuses
+         * @return the builder instance
+         */
+        public Builder lastKnownConsumersStatuses(List<String> lastKnownConsumersStatuses) {
+            this.lastKnownConsumersStatuses = lastKnownConsumersStatuses;
+            return this;
+        }
+
+        /**
+         * Builds a new {@link StreamStatus} instance.
+         *
+         * @return a new {@link StreamStatus} instance
+         */
+        public StreamStatus build() {
+            return new StreamStatus(
+                    publishedBlocks, consumedBlocks, lastKnownPublisherStatuses, lastKnownConsumersStatuses);
+        }
+    }
+}
--- a/simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
@@ -16,13 +16,11 @@
 
 package com.hedera.block.simulator.config.data;
 
-import com.hedera.block.common.utils.Preconditions;
+import static com.hedera.block.common.utils.Preconditions.requirePositive;
+import static java.util.Objects.requireNonNull;
 
 import java.util.ArrayList;
 import java.util.List;
-
-import static com.hedera.block.common.utils.Preconditions.requirePositive;
-import static java.util.Objects.requireNonNull;
 
 /**
  * Represents the status of the stream.
--- a/simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
@@ -16,7 +16,7 @@
 
 package com.hedera.block.simulator.config.data;
 
-import static com.hedera.block.common.utils.Preconditions.requirePositive;
+import static com.hedera.block.common.utils.Preconditions.requireWhole;
 import static java.util.Objects.requireNonNull;
 
 import java.util.ArrayList;
@@ -31,8 +31,8 @@ import java.util.List;
  * @param lastKnownConsumersStatuses the last known consumers statuses
  */
 public record StreamStatus(
-        int publishedBlocks,
-        int consumedBlocks,
+        long publishedBlocks,
+        long consumedBlocks,
         List<String> lastKnownPublisherStatuses,
         List<String> lastKnownConsumersStatuses) {
 
@@ -49,8 +49,8 @@ public record StreamStatus(
      * A builder for creating instances of {@link StreamStatus}.
      */
     public static class Builder {
-        private int publishedBlocks = 0;
-        private int consumedBlocks = 0;
+        private long publishedBlocks = 0;
+        private long consumedBlocks = 0;
         private List<String> lastKnownPublisherStatuses = new ArrayList<>();
         private List<String> lastKnownConsumersStatuses = new ArrayList<>();
 
@@ -67,8 +67,8 @@ public record StreamStatus(
          * @param publishedBlocks the number of published blocks
          * @return the builder instance
          */
-        public Builder publishedBlocks(int publishedBlocks) {
-            requirePositive(publishedBlocks);
+        public Builder publishedBlocks(long publishedBlocks) {
+            requireWhole(publishedBlocks);
             this.publishedBlocks = publishedBlocks;
             return this;
         }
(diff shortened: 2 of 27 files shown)


suites/src/main/java/com/hedera/block/suites/BaseSuite.java
```diff
@@ -121,6 +130,34 @@ protected static GenericContainer<?> createContainer() {
...
+     * @return a {@link Future} representing the asynchronous execution of the block stream simulator
+     */
+    protected Future<?> startSimulatorInThread(BlockStreamSimulatorApp blockStreamSimulatorAppInstance) {
+        return executorService.submit(() -> {
+            try {
+                blockStreamSimulatorAppInstance.start();
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        });
```
ata-nas, line 146: Using the thread pool like that will silence exceptions, we will not know in any way if an exception occurs in the submitted `Future`. If something is happening here, we will never know.

In order to not silence exceptions that can occur in the runnable, we need to override the `ThreadPoolExecutor#afterExecute(Runnable, Throwable)` method. In the javadoc of that method, there is an example override that shows how we can not silence exceptions, but here I provide also some very simple code to play around and test that (taken from the javadoc example):

```java
public class Main {
    public static void main(String[] args) {
        silencing();
//        nonSilencing();
    }

    private static void silencing() {
        final ExecutorService executorService = Executors.newFixedThreadPool(8);

        final Future<Object> future = executorService.submit(() -> {
            throw new RuntimeException("This is a test");
        });

        while (!future.isDone()) {
        }

        executorService.shutdownNow();
    }

    private static void nonSilencing() {
        final ExecutorService executorService = new ThreadPoolExecutor(8, 8, 0L, TimeUnit.MILLISECONDS,
            new LinkedBlockingQueue<>()) {
            @Override
            protected void afterExecute(final Runnable r, final Throwable t) {

                Throwable localThrowable = t;

                if (localThrowable == null && r instanceof Future<?> && ((Future<?>) r).isDone()) {
                    try {
                        final Object result = ((Future<?>) r).get();
                    } catch (CancellationException ce) {
                        localThrowable = ce;
                    } catch (ExecutionException ee) {
                        localThrowable = ee.getCause();
                    } catch (
                        InterruptedException ie) {// ignore/ reset
                    }
                }
                if (localThrowable != null) {
                    localThrowable.printStackTrace(); // here you will see the exception printed, we handle as we wish
                }
            }
        };

        final Future<Object> future = executorService.submit(() -> {
            throw new RuntimeException("This is a test");
        });

        while (!future.isDone()) {
        }

        executorService.shutdownNow();
    }
}
```

I believe we should have a similar implementation, of course modified to fit our needs, otherwise we will not be able to catch problems that can occur which could lead to hours of investigation. This should be noted everywhere we use thread pools!

simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
```diff
@@ -0,0 +1,115 @@
...
+        /**
+         * Sets the number of consumed blocks.
+         *
+         * @param consumedBlocks the number of consumed blocks
+         * @return the builder instance
+         */
+        public Builder consumedBlocks(int consumedBlocks) {
+            this.consumedBlocks = consumedBlocks;
+            return this;
+        }
```
ata-nas, line 81: Should we not add some logic that would test for preconditions, like I presume both ints must be positive? If we fail early is best, especially in object creation logic.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
```diff
@@ -138,10 +148,45 @@ public boolean streamBlock(Block block) {
...
+    /**
+     * Sends a onCompleted message to the server and waits for a short period of
+     * time to ensure the message is sent.
+     *
+     * @throws InterruptedException if the thread is interrupted
+     */
+    @Override
+    public void completeStreaming() throws InterruptedException {
+        requestStreamObserver.onCompleted();
+        Thread.sleep(100);
```
ata-nas, line 164: Do we really need to sleep here? Is the `onCompleted()` method not blocking or am I mistaken?

simulator/src/test/java/com/hedera/block/simulator/BlockStreamSimulatorTest.java
```diff
@@ -75,8 +80,12 @@ void setUp() throws IOException {
     }
 
     @AfterEach
-    void tearDown() {
-        blockStreamSimulator.stop();
+    void tearDown() throws InterruptedException {
+        try {
+            blockStreamSimulator.stop();
+        } catch (UnsupportedOperationException e) {
+            // ignore for now, temporary until consumer mode in simulator is added
```
ata-nas, line 87: Maybe a good idea is to add a `@todo` with an issue number (create a new issue for that test) so we do not forget this. Based on #287.

suites/src/main/java/com/hedera/block/suites/persistence/positive/PositiveDataPersistenceTests.java
```diff
@@ -70,8 +66,8 @@ public void verifyBlockDataSavedInCorrectDirectory() throws InterruptedException
         String savedBlocksFolderBefore = getContainerCommandResult(GET_BLOCKS_COMMAND);
         int savedBlocksCountBefore = getSavedBlocksCount(savedBlocksFolderBefore);
 
-        BlockStreamSimulatorApp blockStreamSimulatorApp = createBlockSimulator();
-        startSimulatorThread(blockStreamSimulatorApp);
+        blockStreamSimulatorApp = createBlockSimulator();
+        simulatorThread = startSimulatorInThread(blockStreamSimulatorApp);
         Thread.sleep(5000);
```
ata-nas, line 71: I do not think it is a good idea to `sleep` in tests, this makes them fragile and potentially flaky based on the logic they test. Today they may pass, but tomorrow the sleep time might not be enough. Another downside is that we may `sleep` for a lot longer than needed thus slowing down the tests if we `sleep` a lot in a lot of places. If the sleep here serves only the purpose to wait for the `simulatorThread` `Future` (which I thing should also be renamed to `simulatorFuture` to better represent the object, since the `Future` is not the actual thread), then there are better ways to wait for the future to complete and to not make the test fragile.

Consider this comment for all similar places.

suites/src/main/java/com/hedera/block/suites/grpc/positive/PositiveEndpointBehaviourTests.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+    /** Default constructor for the {@link PositiveEndpointBehaviourTests} class. */
+    public PositiveEndpointBehaviourTests() {}
+
+    @Test
+    void verifyPublishBlockStreamEndpoint() throws IOException, InterruptedException {
+        blockStreamSimulatorApp = createBlockSimulator();
+        simulatorThread = startSimulatorInThread(blockStreamSimulatorApp);
+        Thread.sleep(5000);
```
ata-nas, line 59: See comment below.

ata-nas (commented):
This is my first review, I will need to do another one for sure. Generally looks good, but I have a couple of questions and a couple of concerns, also a few nit picks.

**General comment:**
Can you please consider to add javadocs to the tests describing what they aim to achieve? Will make them a lot easier to review and for us to know their purpose in order to validate them?

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
```diff
@@ -52,14 +54,17 @@ public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
     private final AtomicBoolean streamEnabled;
     private ManagedChannel channel;
     private final MetricsService metricsService;
+    private final List<String> lastKnownStatuses = new ArrayList<>();
+    private int publishedBlocks;
```
mattp-swirldslabs, line 58: Rather than use a new variable (publishedBlocks), could we use a new block counter metric in `SimulatorMetricMetadata` like `LiveBlocksSent("live_blocks_sent", "Live Blocks Sent")`? You can query it like:
```
@Override
    public long getPublishedBlocks() {
        return metricsService.get(LiveBlockItemsSent).get();
    }
```
Later, we'll be able to see it in a dashboard as well.

mattp-swirldslabs (dismissed):
Nice work, I like seeing all this new test coverage! I added one suggestion.

suites/src/main/java/com/hedera/block/suites/persistence/positive/PositiveDataPersistenceTests.java
georgi-l95: Usually I would agree, and you are correct that this pattern is not good. However there are two reasons for the use of sleep here: 
- This was the easiest way, without any overengineering to achive the needed results and test the behaviour I want to test. Other way would be too add pooling mechanism, but wasn't in the scope of the tasks.
- Sleep here is a benefit and needed in our test cases. Here is why: We want to start and stream data to the block-node, and becuase block-node performance is such an integral part of the application, we want to make sure that this data was received and procesed very fast. Therefore waiting a little (5seconds, but we can lower it in the future) and checking, whether the data that we've send was processed will tell us, if block-node is working is expected. And if this fails for some reason, it will signal us that there is something wrong with the performance. Therefore in this case, if this test and the other ones that we us similar sleep fails, it won't be because it's flaky, but because we have a problem with block-node performance, which is good, because we can start investigating.

simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
georgi-l95: Good point, added. ✅

simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
```diff
@@ -0,0 +1,115 @@
...
+        /**
+         * Sets the last known consumers statuses.
+         *
+         * @param lastKnownConsumersStatuses the last known consumers statuses
+         * @return the builder instance
+         */
+        public Builder lastKnownConsumersStatuses(List<String> lastKnownConsumersStatuses) {
+            this.lastKnownConsumersStatuses = lastKnownConsumersStatuses;
+            return this;
+        }
```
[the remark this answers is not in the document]
georgi-l95: Good point, added. ✅

.gitignore
```diff
@@ -58,3 +58,8 @@ server/src/test/resources/test_output/
 
 # simulator files, this files are un-tarred before build
 /simulator/src/main/resources/block-0.0.3/
+/common/bin
+/server/bin
+/simulator/bin
+/stream/bin
+/suites/bin
```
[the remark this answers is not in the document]
georgi-l95: You're right. Changed. ✅

simulator/src/test/java/com/hedera/block/simulator/BlockStreamSimulatorTest.java
georgi-l95: Changed 👍

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
```diff
@@ -138,10 +148,45 @@ public boolean streamBlock(Block block) {
...
+    }
+
+    /**
+     * Gets the last known statuses.
+     *
+     * @return the last known statuses
+     */
+    @Override
+    public List<String> getLastKnownStatuses() {
+        return lastKnownStatuses;
```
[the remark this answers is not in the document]
georgi-l95: True, changed. ✅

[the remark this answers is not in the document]
ata-nas: Returning a copy is also an option as we do not expose the internal ref 👍

simulator/src/main/java/com/hedera/block/simulator/config/data/StreamStatus.java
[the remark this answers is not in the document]
ata-nas: Maybe you forgot to add this change?

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
ata-nas: I see, it would be good to investigate this in the future, do you agree to make an issue for the future and tagging it with our `@todo(issueNumber)` convention so we do not forget?

suites/src/main/java/com/hedera/block/suites/persistence/positive/PositiveDataPersistenceTests.java
ata-nas: I do agree with you that sleeping is the easiest way to achieve this, for this test even a while not done with an `@Timeout` on the test method for security of the test not going on forever would do, that is IF the Future shall complete soon (in your case 5 secs), otherwise if we only need to start up the Future and let it then go on, that is another case. Yet I do think it is good to take a look at this in an upcoming issue. Would you agree to make an issue and tag it with our `@todo(issueNumber)` convention?

suites/src/main/java/com/hedera/block/suites/grpc/positive/PositiveEndpointBehaviourTests.java
ata-nas: [Answered](https://github.com/hashgraph/hedera-block-node/pull/330#discussion_r1846513158).

suites/src/main/java/com/hedera/block/suites/persistence/positive/PositiveDataPersistenceTests.java
georgi-l95: `@Timeout` is not suitable here. Sleep is intentional here, this future, in our case simulator can go on forever. We just need it run a little and then check if it's done it's job.

simulator/src/main/java/com/hedera/block/simulator/grpc/PublishStreamGrpcClientImpl.java
georgi-l95: Done, thanks for this. ✅

suites/src/main/java/com/hedera/block/suites/BaseSuite.java
georgi-l95: Good point, I've added a custom thread executor, with simple funtionality to log the errors for now. We can extend as needed in the future..

mattp-swirldslabs (approved):

mattp-swirldslabs (dismissed):

suites/src/main/java/com/hedera/block/suites/CustomThreadPoolExecutor.java
```diff
@@ -0,0 +1,76 @@
...
+ *   <li>Core pool size: 8 threads
+ *   <li>Maximum pool size: 8 threads
+ *   <li>Keep-alive time: 10 seconds
+ *   <li>Work queue: {@link LinkedBlockingQueue} with default capacity
+ * </ul>
+ *
+ * <p>The executor overrides the {@link #afterExecute(Runnable, Throwable)} method to log any
+ * exceptions thrown by tasks after their execution, aiding in debugging and error tracking.
+ */
+public class CustomThreadPoolExecutor extends ThreadPoolExecutor {
```
ata-nas, line 40: I suggest that we rename this to `SimpleErrorLoggingExecutor` or similar, since the name `Custom` is really not descriptive.

suites/src/main/java/com/hedera/block/suites/CustomThreadPoolExecutor.java
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * @param r the runnable task that has completed execution
+     * @param t the exception that caused termination, or {@code null} if execution completed normally
+     */
+    @Override
+    protected void afterExecute(Runnable r, Throwable t) {
+        super.afterExecute(r, t);
+        if (t != null) {
+            LOGGER.log(ERROR, "Task encountered an error: " + t.getMessage());
+        }
```
ata-nas, line 74: Unfortunately this will not work again, as seen in the thread pool docs, most of the time we expect the`Throwable t` to be null, effectively we will be silencing exceptions again. Also, `super.afterExecute` in this case does nothing, it has empty body. Also, we should pass the exception as an object instead to get the message for the logger.

The correct implementation would be:

```suggestion
        Throwable localThrowable = t;
        if (localThrowable == null && r instanceof Future<?> && ((Future<?>) r).isDone()) {
            try {
                final Object result = ((Future<?>) r).get();
            } catch (CancellationException ce) {
                localThrowable = ce;
            } catch (ExecutionException ee) {
                localThrowable = ee.getCause();
            } catch (
                InterruptedException ie) {// ignore/ reset
                Thread.currentThread().interrupt();
            }
        }
        if (localThrowable != null) {
            LOGGER.log(ERROR, "Task encountered an error: ", localThrowable);
        }
```

Most of the times, we expect the throwable to be null and then when we get the runnable, we expect an `ExecutionException` to be thrown.

ata-nas (changes requested):
@georgi-l95 looks good, but the thread pool requires some changes as the current implementation will still silence exceptions.

suites/src/main/java/com/hedera/block/suites/BaseSuite.java
ata-nas: Agreed that for now we simply need to log, it is enough just to catch exceptions so we can debug if something is wrong inside these futures.

suites/src/main/java/com/hedera/block/suites/CustomThreadPoolExecutor.java
georgi-l95: Changed to `ErrorLoggingExecutor`

suites/src/main/java/com/hedera/block/suites/CustomThreadPoolExecutor.java
georgi-l95: Yes, correct. I've left it more as a placeholder to be address in future PRs, as we don't heavly rely on the E2E tests currently. But I've added the suggested changes from you, so we don't silence exceptions. Thank you 👍

ata-nas (approved):
Looks great! Good job @georgi-l95!

rbarkerSL (approved):
Approval applies to `.gitignore`

PR #337, opened 2024-11-12
fix: gradle tasks from different subproject are not isolated

### LINKED ISSUE #299: bug: gradle tasks from different subproject are not isolated
### Description

Due to some concerns that arose in #264, there are repeating task names `createDockerImage` (outside the scope of the issue, but inside the scope of the discussion #264) which do touch on a common file that gets created via a shell script `update-env` and in the asynchronous nature of the running of the tasks, it could happen that a project gets build with the wrong values that come from the common `.env` file.

### Steps to reproduce

N/A

### Additional context

_No response_

### Hedera network

_No response_

### Version

0.2.0-SNAPSHOT

### Operating system

None

ata-nas:
**Description**:

Isolate `gradle` tasks to not go outside of the respective project's build root.

**Related issue(s)**:

Fixes #299

**Notes for reviewer**:

- am I missing something?

**Checklist**

### DIFF
--- a/server/build.gradle.kts
+++ b/server/build.gradle.kts
@@ -84,14 +84,16 @@ val updateDockerEnv: TaskProvider<Exec> =
         )
     }
 
-val createDockerImage: TaskProvider<Exec> = tasks.register<Exec>("createDockerImage") {
-    description = "Creates the docker image of the Block Node Server based on the current version"
-    group = "docker"
+val createDockerImage: TaskProvider<Exec> =
+    tasks.register<Exec>("createDockerImage") {
+        description =
+            "Creates the docker image of the Block Node Server based on the current version"
+        group = "docker"
 
-    dependsOn(updateDockerEnv, tasks.assemble)
-    workingDir(dockerRootDirectory)
-    commandLine("./docker-build.sh", project.version, layout.projectDirectory.dir("..").asFile)
-}
+        dependsOn(updateDockerEnv, tasks.assemble)
+        workingDir(dockerRootDirectory)
+        commandLine("./docker-build.sh", project.version, layout.projectDirectory.dir("..").asFile)
+    }
 
 tasks.register<Exec>("startDockerContainer") {
     description = "Starts the docker container of the Block Node Server of the current version"
@@ -99,7 +101,11 @@ tasks.register<Exec>("startDockerContainer") {
 
     dependsOn(createDockerImage)
     workingDir(dockerRootDirectory)
-    commandLine("sh", "-c", "docker compose --env-file $buildRootEnvFileAbsolutePath -p block-node up -d")
+    commandLine(
+        "sh",
+        "-c",
+        "docker compose --env-file $buildRootEnvFileAbsolutePath -p block-node up -d"
+    )
 }
 
 tasks.register<Exec>("startDockerDebugContainer") {
--- a/server/build.gradle.kts
+++ b/server/build.gradle.kts
@@ -77,8 +77,9 @@ val copyDockerFolder: TaskProvider<Copy> =
         into(dockerBuildRootDirectory)
     }
 
-// createProductionDotEnv is temporary and used by the suites, once the suites no longer rely on the
-// .env file from the build root we should remove this task
+// @todo(#343) - createProductionDotEnv is temporary and used by the suites,
+//  once the suites no longer rely on the .env file from the build root we
+//  should remove this task
 val createProductionDotEnv: TaskProvider<Exec> =
     tasks.register<Exec>("createProductionDotEnv") {
         description = "Creates the default dotenv file for the Block Node Server"
--- a/server/build.gradle.kts
+++ b/server/build.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -65,14 +65,20 @@ tasks.register("bumpVersion") {
 
 // Docker related tasks
 
-var updateDockerEnv =
+val buildRootAbsolutePath: String = layout.buildDirectory.get().asFile.toPath().toAbsolutePath().toString()
+
+val updateDockerEnv: TaskProvider<Exec> =
     tasks.register<Exec>("updateDockerEnv") {
         description =
             "Creates the .env file in the docker folder that contains environment variables for docker"
         group = "docker"
 
         workingDir(layout.projectDirectory.dir("docker"))
-        commandLine("sh", "-c", "./update-env.sh ${project.version} false false")
+        commandLine(
+            "sh",
+            "-c",
+            "./update-env.sh $buildRootAbsolutePath ${project.version} false false"
+        )
     }
 
 tasks.register<Exec>("createDockerImage") {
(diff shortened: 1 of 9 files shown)


server/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
```
jsync-swirlds, line 2: In most cases we do not want to remove starting year from copyright.
My understanding is that even if we copied then modified another file, the result is a modification, and would generally maintain the chain of copyright claims.
It _might_ be OK in this case, but it's probably safer to always keep the first year and only ever update the second (aside from brand new files, of course).

server/docker/docker-compose.yml
```diff
@@ -1,11 +1,9 @@
-version: '3.8'
-
 services:
   block-node-server:
     container_name: block-node-server
     image: block-node-server:${VERSION}
     env_file:
-      - .env
+      - ../build/.env
```
jsync-swirlds, line 6: If you need `..` to get to `build`, then either the working folder is wrong (escapes build), or we have a bug somewhere.
Docker build, in particular, should never be run from outside the `build` folder, as everything intended to be distributed _must_ be in build.
(even files that are not compiler-generated, the gradle build process should be copying them into the build folder; not a docker process).

server/docker/update-env.sh
```diff
@@ -3,16 +3,21 @@
 # This scripts create a '.env' file that is used for docker & docker-compose as an input of environment variables.
 # This script is called by gradle and get the current project version as an input param
 
-if [ $# -lt 1 ]; then
-  echo "USAGE: $0 <VERSION> <DEBUG> <SMOKE_TEST>"
+if [ $# -lt 2 ]; then
+  # <PROJECT_BUILD_ROOT> and <VERSION> are required!
+  echo "USAGE: $0 <PROJECT_BUILD_ROOT> <VERSION> [DEBUG] [SMOKE_TEST]"
```
jsync-swirlds, line 8: I wonder why we need to specify location.
The gradle build should be copying any "source" `.env` file into the build folder, and this file should only _ever_ modify a `.env` file that is _within_ the build folder...
If the "working" folder for all processes is always the gradle build folder (it should be), then we wouldn't need to specify the build root here...

suites/src/main/java/com/hedera/block/suites/BaseSuite.java
```diff
@@ -47,6 +47,12 @@
  * <p>The Block Node Application version is retrieved dynamically from an environment file (.env).
  */
 public abstract class BaseSuite {
+    /**
+     * Dotenv instance to load the environment variables from the .env file that
+     * is inside the build root of the :server.
+     */
+    protected static final Dotenv DOTENV =
+            Dotenv.configure().directory("../server/build").filename(".env").load();
```
jsync-swirlds, line 55: This is not ideal, as it bypasses project isolation and ties this suite to a specific _other project_ build output.  Whatever runs _this suite_ should generate a _local_ `.env` and use that (if needed).  Otherwise the suite is tied to the server build via a file that's not part of the server outputs, and that's not a good pattern.

server/build.gradle.kts
jsync-swirlds: If spotless is removing earlier years, then we need to fix the copyright check in spotless (which we do control, it's a block in a gradle file).

ata-nas: ✅ Done. Reverted.

server/docker/docker-compose.yml
ata-nas: Thanks for that insight, initially, I did not see it in that light, but copying the `docker` dir into the project `build` dir and then containing all shell executions from `build` as a working directory is a lot safer and cleaner. I have made some changes.

server/docker/update-env.sh
ata-nas: Now when we copy the `docker` dir into the project `build` dir, and executing the shell scripts from there, this is no longer needed.

suites/src/main/java/com/hedera/block/suites/BaseSuite.java
ata-nas: Agreed, it is not ideal indeed. This was introduced as a temporary solution in the past. In theory, the test suites require the server docker image to be built, they are "coupled" in that sense, or rather the suites are coupled to the server, not the other way around. If we introduce a new `.env` file that is local to the suites, it will be an improvement, but yet again, the server version must still "come" from the server (granted there are other ways to handle that). I propose to make an issue to describe this and we will handle it in the future.

server/docker/docker-compose.yml
```diff
@@ -1,5 +1,3 @@
-version: '3.8'
-
```
ata-nas, line 2: `version` is removed, was getting warnings that is deprecated and needs to be removed, please verify

suites/src/main/java/com/hedera/block/suites/BaseSuite.java
jsync-swirlds: If we have the suites build _depend_ on the server build, then the server artifacts (jar files, mostly) will be copied _by gradle_ into the build folder (I'd have to look up where, usually it's a "libs" folder).
We can (should?) get all of the server information from the jar if we've designed this correctly.

It's fine to create a separate issue to resolve this.

server/docker/docker-compose.yml
mishomihov00: I think this is ok.

mishomihov00 (dismissed):
Reviewed and approved only these files:
`.github/workflows/e2e-tests.yaml`
`.github/workflows/smoke-test.yaml`
`server/docker/docker-build.sh`
`server/docker/docker-compose.yml`
`server/docker/update-env.sh`

suites/src/main/java/com/hedera/block/suites/BaseSuite.java
ata-nas: #343 is now open and properly tagged within this PR's changes.

mishomihov00 (approved):
Reviewed and approved only these files:
`.github/workflows/e2e-tests.yaml`
`.github/workflows/smoke-test.yaml`
`server/docker/docker-build.sh`
`server/docker/docker-compose.yml`
`server/docker/update-env.sh`

georgi-l95 (dismissed):
👁️ 👍 👉 👨‍💻

jeromy-cannon (dismissed):

server/build.gradle.kts
```diff
@@ -63,79 +63,91 @@ tasks.register("bumpVersion") {
...
-}
+        dependsOn(copyDockerFolder, tasks.assemble)
+        workingDir(dockerBuildRootDirectory)
+        commandLine("sh", "-c", "./docker-build.sh ${project.version}")
+    }
 
 tasks.register<Exec>("startDockerContainer") {
-    description = "Starts the docker container of the Block Node Server of the current version"
+    description =
+        "Starts the docker productiongit a container of the Block Node Server for the current version"
```
rbarkerSL, line 106:
```suggestion
        "Starts the docker production container of the Block Node Server for the current version"
```

rbarkerSL (changes requested):
One small nit

server/build.gradle.kts
ata-nas: ✅ @rbarkerSL thanks, I have missed that typo. This is pushed.

jeromy-cannon (approved):

rbarkerSL (approved):
lgtm

jeromy-cannon (approved):

georgi-l95 (approved):

PR #360, opened 2024-11-22
feat: Implement consumer handler in the Simulator

### LINKED ISSUE #121: Implement consumer logic in the Simulator
As a block node developer
I want to be able to test streaming data to a consumer
So we need to add consumer side setup on the simulator side to test out this capability.

---
### Additional Information

Currently the simulator does not have any working implementation for consuming data. There is already separtion in the code, therefore It would be easy to add this in the `consumerModeHandler`.

georgi-l95:
**Description**:
This pull request aims to make use of the consumer mode in the simulator, by adding needed implementations.
We add:
- gRPC client and observer for consuming blocks, which utilizes `SubscribeStreamRequest` and `SubscribeStreamResponse`, which are part of the `subscribeBlockStream` rpc.
- Makes the neccessery changes to initilize managers and clients, depending on mode.
- Adds neccessery unit and integration tests for sufficient code coverage, by making use of as less mocks as possible.

**Related issue(s)**:

Fixes #121 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/common/src/main/java/com/hedera/block/common/utils/Preconditions.java
+++ b/common/src/main/java/com/hedera/block/common/utils/Preconditions.java
@@ -55,8 +55,8 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      */
     public static String requireNotBlank(final String toCheck, final String errorMessage) {
         if (StringUtilities.isBlank(toCheck)) {
-            final String message = Objects.isNull(errorMessage) ? "The input String is required to be non-blank."
-                    : errorMessage;
+            final String message =
+                    Objects.isNull(errorMessage) ? "The input String is required to be non-blank." : errorMessage;
             throw new IllegalArgumentException(message);
         } else {
             return toCheck;
@@ -257,6 +257,5 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
         }
     }
 
-    private Preconditions() {
-    }
+    private Preconditions() {}
 }
--- a/common/src/main/java/com/hedera/block/common/utils/Preconditions.java
+++ b/common/src/main/java/com/hedera/block/common/utils/Preconditions.java
@@ -27,11 +27,11 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * we return it, else we throw {@link IllegalArgumentException}.
      *
      * @param toCheck a {@link String} to be checked if is blank as defined
-     * above
+     *                above
      * @return the {@link String} to be checked if it is not blank as defined
-     * above
+     *         above
      * @throws IllegalArgumentException if the input {@link String} to be
-     * checked is blank
+     *                                  checked is blank
      */
     public static String requireNotBlank(final String toCheck) {
         return requireNotBlank(toCheck, null);
@@ -43,19 +43,20 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * {@link String#isBlank()}. If the given {@link String} is not blank, then
      * we return it, else we throw {@link IllegalArgumentException}.
      *
-     * @param toCheck a {@link String} to be checked if is blank as defined
-     * above
+     * @param toCheck      a {@link String} to be checked if is blank as defined
+     *                     above
      * @param errorMessage the error message to be used in the exception if the
-     * input {@link String} to be checked is blank, if null, a default message
+     *                     input {@link String} to be checked is blank, if null, a
+     *                     default message
      * @return the {@link String} to be checked if it is not blank as defined
-     * above
+     *         above
      * @throws IllegalArgumentException if the input {@link String} to be
-     * checked is blank
+     *                                  checked is blank
      */
     public static String requireNotBlank(final String toCheck, final String errorMessage) {
         if (StringUtilities.isBlank(toCheck)) {
-            final String message =
-                    Objects.isNull(errorMessage) ? "The input String is required to be non-blank." : errorMessage;
+            final String message = Objects.isNull(errorMessage) ? "The input String is required to be non-blank."
+                    : errorMessage;
             throw new IllegalArgumentException(message);
         } else {
             return toCheck;
@@ -69,7 +70,7 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * @param toCheck the number to check if it is a positive power of two
      * @return the number to check if it is positive
      * @throws IllegalArgumentException if the input number to check is not
-     * positive
+     *                                  positive
      */
     public static int requirePositive(final int toCheck) {
         return requirePositive(toCheck, null);
@@ -79,13 +80,14 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * This method asserts a given integer is a positive. An integer is positive
      * if it is NOT equal to zero and is greater than zero.
      *
-     * @param toCheck the integer to check if it is a positive power of two
+     * @param toCheck      the integer to check if it is a positive power of two
      * @param errorMessage the error message to be used in the exception if the
-     * input integer to check is not positive, if null, a default message will
-     * be used
+     *                     input integer to check is not positive, if null, a
+     *                     default message will
+     *                     be used
      * @return the number to check if it is positive
      * @throws IllegalArgumentException if the input number to check is not
-     * positive
+     *                                  positive
      */
     public static int requirePositive(final int toCheck, final String errorMessage) {
         if (0 >= toCheck) {
@@ -105,7 +107,7 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * @param toCheck the long to check if it is a positive power of two
      * @return the long to check if it is positive
      * @throws IllegalArgumentException if the input long to check is not
-     * positive
+     *                                  positive
      */
     public static long requirePositive(final long toCheck) {
         return requirePositive(toCheck, null);
@@ -115,13 +117,14 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * This method asserts a given long is a positive. A long is positive
      * if it is NOT equal to zero and is greater than zero.
      *
-     * @param toCheck the long to check if it is a positive power of two
+     * @param toCheck      the long to check if it is a positive power of two
      * @param errorMessage the error message to be used in the exception if the
-     * input long to check is not positive, if null, a default message will
-     * be used
+     *                     input long to check is not positive, if null, a default
+     *                     message will
+     *                     be used
      * @return the long to check if it is positive
      * @throws IllegalArgumentException if the input long to check is not
-     * positive
+     *                                  positive
      */
     public static long requirePositive(final long toCheck, final String errorMessage) {
         if (0L >= toCheck) {
@@ -135,13 +138,61 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
     }
 
     /**
+     * Ensures that a given long value is greater than or equal to a specified base
+     * value.
+     * If the value does not meet the requirement, an
+     * {@link IllegalArgumentException} is thrown.
+     *
+     * <p>
+     * This method delegates the validation to
+     * {@link #requireGreaterOrEqual(long, long, String)},
+     * using a default error message if the check fails.
+     * </p>
+     *
+     * @param toTest the long value to test
+     * @param base   the base value to compare against
+     * @return the input {@code toTest} if it is greater than or equal to
+     *         {@code base}
+     * @throws IllegalArgumentException if {@code toTest} is less than {@code base}
+     */
+    public static long requireGreaterOrEqual(final long toTest, final long base) {
+        return requireGreaterOrEqual(toTest, base, null);
+    }
+
+    /**
+     * Ensures that a given long value is greater than or equal to a specified base
+     * value.
+     * If the value does not meet the requirement, an
+     * {@link IllegalArgumentException} is thrown.
+     *
+     * @param toTest       the long value to test
+     * @param base         the base value to compare against
+     * @param errorMessage the error message to include in the exception if the
+     *                     check fails;
+     *                     if {@code null}, a default message is used
+     * @return the input {@code toTest} if it is greater than or equal to
+     *         {@code base}
+     * @throws IllegalArgumentException if {@code toTest} is less than {@code base}
+     */
+    public static long requireGreaterOrEqual(final long toTest, final long base, final String errorMessage) {
+        if (toTest >= base) {
+            return toTest;
+        }
+
+        final String message = Objects.isNull(errorMessage)
+                ? "The input integer [%d] is required be greater or equal than [%d].".formatted(toTest, base)
+                : errorMessage;
+        throw new IllegalArgumentException(message);
+    }
+
+    /**
      * This method asserts a given long is a whole number. A long is whole
      * if it is greater or equal to zero.
      *
      * @param toCheck the long to check if it is a whole number
      * @return the number to check if it is whole number
      * @throws IllegalArgumentException if the input number to check is not
-     * positive
+     *                                  positive
      */
     public static long requireWhole(final long toCheck) {
         return requireWhole(toCheck, null);
@@ -151,13 +202,14 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * This method asserts a given long is a whole number. A long is whole
      * if it is greater or equal to zero.
      *
-     * @param toCheck the long to check if it is a whole number
+     * @param toCheck      the long to check if it is a whole number
      * @param errorMessage the error message to be used in the exception if the
-     * input long to check is not a whole number, if null, a default message will
-     * be used
+     *                     input long to check is not a whole number, if null, a
+     *                     default message will
+     *                     be used
      * @return the number to check if it is whole number
      * @throws IllegalArgumentException if the input number to check is not
-     * positive
+     *                                  positive
      */
     public static long requireWhole(final long toCheck, final String errorMessage) {
         if (toCheck >= 0) {
@@ -176,7 +228,7 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * @param toCheck the number to check if it is a power of two
      * @return the number to check if it is a power of two
      * @throws IllegalArgumentException if the input number to check is not a
-     * power of two
+     *                                  power of two
      */
     public static int requirePowerOfTwo(final int toCheck) {
         return requirePowerOfTwo(toCheck, null);
@@ -185,13 +237,14 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
     /**
      * This method asserts a given integer is a power of two.
      *
-     * @param toCheck the number to check if it is a power of two
+     * @param toCheck      the number to check if it is a power of two
      * @param errorMessage the error message to be used in the exception if the
-     * input integer to check is not a power of two, if null, a default message
-     * will be used
+     *                     input integer to check is not a power of two, if null, a
+     *                     default message
+     *                     will be used
      * @return the number to check if it is a power of two
      * @throws IllegalArgumentException if the input number to check is not a
-     * power of two
+     *                                  power of two
      */
     public static int requirePowerOfTwo(final int toCheck, final String errorMessage) {
         if (!MathUtilities.isPowerOfTwo(toCheck)) {
@@ -204,5 +257,6 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
         }
     }
 
-    private Preconditions() {}
+    private Preconditions() {
+    }
 }
--- a/simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
@@ -57,17 +57,15 @@ import javax.inject.Inject;
  * execution, and shutdown of streaming operations based on the configured mode.
  */
 public class BlockStreamSimulatorApp {
-    /** Logger for this class */
     private final System.Logger LOGGER = System.getLogger(getClass().getName());
 
     // Service dependencies
     private final PublishStreamGrpcClient publishStreamGrpcClient;
     private final ConsumerStreamGrpcClient consumerStreamGrpcClient;
     private final SimulatorModeHandler simulatorModeHandler;
-    private final MetricsService metricsService;
 
     // State
-    private final AtomicBoolean isRunning;
+    private final AtomicBoolean isRunning = new AtomicBoolean(false);
 
     /**
      * Creates a new BlockStreamSimulatorApp instance with the specified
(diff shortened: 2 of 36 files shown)


simulator/src/main/java/com/hedera/block/simulator/mode/ConsumerModeHandler.java
```diff
@@ -34,31 +36,58 @@
  * {@link UnsupportedOperationException}.
  */
 public class ConsumerModeHandler implements SimulatorModeHandler {
+    /** Logger for this class */
+    private final System.Logger LOGGER = System.getLogger(getClass().getName());
 
+    // Configuration
     private final BlockStreamConfig blockStreamConfig;
```
mattp-swirldslabs, line 43: `BlockStreamConfig` is never read. Should it be removed?

simulator/src/main/java/com/hedera/block/simulator/generator/BlockStreamManager.java
```diff
@@ -25,6 +25,11 @@
 /** The block stream manager interface. */
 public interface BlockStreamManager {
 
+    /**
+     * Initialize the block stream manager and load blocks into memory.
+     */
+    void init();
```
mattp-swirldslabs, line 31: Maybe change this to a default interface method like: `default void init() {}` so you don't have to override with a no-op comment in `BlockAsFileLargeDataSets`

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsFileLargeDataSets.java
```diff
@@ -65,6 +65,14 @@ public BlockAsFileLargeDataSets(@NonNull BlockGeneratorConfig config) {
         this.formatString = "%0" + config.paddedLength() + "d" + config.fileExtension();
     }
 
+    /**
+     * Initialize the block stream manager and load blocks into memory.
+     */
+    @Override
+    public void init() {
+        // Do nothing, because we don't have real initializing and loading blocks into memory for this implementation.
+    }
```
mattp-swirldslabs, line 74: You should be able to remove this `init()` method. See my comment in `BlockStreamManager`

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamGrpcClientImpl.java
```diff
@@ -0,0 +1,121 @@
...
+        stub.subscribeBlockStream(request, consumerStreamObserver);
+
+        streamLatch.await();
+    }
+
+    @Override
+    public void completeStreaming() throws InterruptedException {
+        consumerStreamObserver.onCompleted();
+        // todo(352) Find a suitable solution for removing the sleep
+        Thread.sleep(100);
```
mattp-swirldslabs, line 104: I'm not 100% sure but I think Netty will continue to send BlockItems to onNext() as long as the channel is open. Right now, I think `onCompleted()` will only release the starting thread waiting at line 97. You might need to close the channel in this method and then release the starting thread.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public void onCompleted() {
+        LOGGER.log(INFO, "Subscribe request completed.");
+        streamLatch.countDown();
+    }
+
+    private void processBlockItems(List<BlockItem> blockItems) {
+        blockItems.stream()
+                .filter(BlockItem::hasBlockProof)
+                .forEach(__ -> metricsService.get(LiveBlocksConsumed).increment());
```
mattp-swirldslabs, line 110: If you start the consumer up and start streaming blockitems via the producer to the BN, it looks like nothing is happening on the consumer side. Consider changing this to something like:
```
blockItems.stream()
                .filter(BlockItem::hasBlockProof)
                .forEach(blockItem -> {
                    metricsService.get(LiveBlocksConsumed).increment();
                    LOGGER.log(INFO, "Received block number: " + blockItem.getBlockProof().getBlock());
                });
```
That should print the last block received without overwhelming the terminal. This gives the consumer parity with the producer which prints the response hashes right now.

mattp-swirldslabs (approved):
I left a few suggestions

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -36,96 +37,119 @@
...
      */
     public void start() throws InterruptedException, BlockSimulatorParsingException, IOException {
-        LOGGER.log(System.Logger.Level.INFO, "Block Stream Simulator started initializing components...");
-        publishStreamGrpcClient.init();
+        LOGGER.log(INFO, "Block Stream Simulator started initializing components...");
+        simulatorModeHandler.init();
+
         isRunning.set(true);
 
         simulatorModeHandler.start();
```
ata-nas, line 115: Shouldn't the `isRunning` field be set to `true` after the starting of the `simulatorModeHandler`?

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -36,96 +37,119 @@
...
      *
-     * @throws InterruptedException if the thread is interrupted
+     * @throws InterruptedException if the shutdown process is interrupted
      */
     public void stop() throws InterruptedException {
         simulatorModeHandler.stop();
-        publishStreamGrpcClient.completeStreaming();
-
-        publishStreamGrpcClient.shutdown();
         isRunning.set(false);
```
ata-nas, line 135: I believe we have made in the past an issue to add different statuses for the simulator in order to have more granularity in determining if the app is shutting down or is shut down or is initializing or running or paused etc. etc.
If that is so maybe we could tag this `stop` method with the `@todo(issueNumber)` convention?

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamGrpcClientImpl.java
```diff
@@ -0,0 +1,121 @@
...
+                .usePlaintext()
+                .build();
+        stub = BlockStreamServiceGrpc.newStub(channel);
+        lastKnownStatuses.clear();
+    }
+
+    @Override
+    public void requestBlocks(long startBlock, long endBlock) throws InterruptedException {
+        Preconditions.requireWhole(startBlock);
+        Preconditions.requireWhole(endBlock);
```
ata-nas, line 85: Maybe here would be a good idea to have a precondition check to verify that the `endBlock` >= `startBlock`, seems logical to me. If that is so, this is a very good candidate for a new `Preconditions` method, named for example `Preconditions#requireGreaterOrEqual(final long longToTest, final long longBase)` or something similar.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamGrpcClientImpl.java
```diff
@@ -42,29 +43,37 @@
...
 /**
- * The PublishStreamGrpcClientImpl class provides the methods to stream the
- * block and block item.
+ * Implementation of {@link PublishStreamGrpcClient} that handles the publication of blocks
+ * via gRPC streaming. This implementation manages the connection to the server, handles
+ * block chunking, and tracks metrics related to block publication.
  */
 public class PublishStreamGrpcClientImpl implements PublishStreamGrpcClient {
-
+    /** Logger for this class */
```
ata-nas, line 51: Doc for logger could be omitted.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamObserver.java
```diff
@@ -14,66 +14,76 @@
...
- * The PublishStreamObserver class provides the methods to observe the stream of
- * the published
- * stream.
+ * Implementation of StreamObserver that handles responses from the block publishing stream.
+ * This class processes server responses and manages the stream state based on server feedback.
  */
 public class PublishStreamObserver implements StreamObserver<PublishStreamResponse> {
 
-    private final Logger logger = System.getLogger(getClass().getName());
+    /** Logger for this class */
```
ata-nas, line 36: Docs for logger could be omitted.

simulator/src/main/java/com/hedera/block/simulator/mode/ConsumerModeHandler.java
```diff
@@ -34,31 +36,58 @@
  * {@link UnsupportedOperationException}.
  */
 public class ConsumerModeHandler implements SimulatorModeHandler {
+    /** Logger for this class */
```
ata-nas, line 39: Docs for logger could be omitted.

simulator/src/main/java/com/hedera/block/simulator/mode/CombinedModeHandler.java
```diff
@@ -37,28 +37,39 @@ public class CombinedModeHandler implements SimulatorModeHandler {
...
     /**
-     * Constructs a new {@code CombinedModeHandler} with the specified block stream configuration.
+     * Constructs a new {@code CombinedModeHandler} with the specified configuration.
      *
-     * @param blockStreamConfig the configuration data for managing block streams
+     * @param blockStreamConfig The configuration for block streaming parameters
+     * @throws NullPointerException if blockStreamConfig is null
      */
     public CombinedModeHandler(@NonNull final BlockStreamConfig blockStreamConfig) {
         this.blockStreamConfig = requireNonNull(blockStreamConfig);
```
ata-nas, line 46: The `blockStreamConfig` field seems to be unused, we could remove?

simulator/src/main/java/com/hedera/block/simulator/mode/ConsumerModeHandler.java
ata-nas: +1

simulator/src/main/java/com/hedera/block/simulator/mode/ConsumerModeHandler.java
```diff
@@ -34,31 +36,58 @@
...
+     * Gracefully stops block consumption and shuts down the gRPC client.
+     *
+     * @throws InterruptedException if the shutdown process is interrupted
      */
     @Override
-    public void stop() {
-        throw new UnsupportedOperationException();
+    public void stop() throws InterruptedException {
+        consumerStreamGrpcClient.completeStreaming();
+        consumerStreamGrpcClient.shutdown();
```
ata-nas, line 91: Generally it is a good idea to have a single 'teardown' type method because it is very error prone when when we need to call multiple things to shutdown something. It is a need to know solution and when used by someone else that someone else could not use the teardown logic properly as it was intended by the creator. I would suggest to streamline this, as well as in the other implementations of the handlers and clients in order to be clear how to shutdown (via interface) and then in the single shutdown method we should have all cleanup logic based on the specific implementation.

simulator/src/main/java/com/hedera/block/simulator/mode/PublisherModeHandler.java
```diff
@@ -170,7 +181,9 @@ private void constantRateStreaming() throws InterruptedException, IOException, B
      * Stops the handler and manager from streaming.
      */
     @Override
-    public void stop() {
+    public void stop() throws InterruptedException {
         shouldPublish.set(false);
+        publishStreamGrpcClient.completeStreaming();
+        publishStreamGrpcClient.shutdown();
```
ata-nas, line 187: See comment from `ConsumerModeHandler` for the same teardown logic.

simulator/src/main/java/com/hedera/block/simulator/mode/PublisherModeHandler.java
```diff
@@ -46,27 +46,31 @@
  * the simulator needs to handle publication of blocks.
  */
 public class PublisherModeHandler implements SimulatorModeHandler {
+    /** Logger for this class */
```
ata-nas, line 49: Docs for logger could be omitted.

simulator/src/test/java/com/hedera/block/simulator/mode/ConsumerModeHandlerTest.java
```diff
@@ -17,24 +17,71 @@
 package com.hedera.block.simulator.mode;
 
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.*;
```
ata-nas, line 20: Consider to remove this kind of wildcard imports.

ata-nas (commented):
@georgi-l95 generally looks good on my first pass! It is nice to see a clear distinction of different components.

I do have some comments left here and there.

Also, please consider to write javadocs for the tests, since it is not possible to see what is the intent of the test and it is very hard to review them.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -36,96 +37,119 @@
...
-
+    /** Logger for this class */
     private final System.Logger LOGGER = System.getLogger(getClass().getName());
 
+    // Service dependencies
     private final PublishStreamGrpcClient publishStreamGrpcClient;
+    private final ConsumerStreamGrpcClient consumerStreamGrpcClient;
     private final SimulatorModeHandler simulatorModeHandler;
-    private final AtomicBoolean isRunning = new AtomicBoolean(false);
     private final MetricsService metricsService;
```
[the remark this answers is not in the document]
mattp-swirldslabs: I think I did that. I was anticipating using metrics at the top level as well (and I think it was passing the metrics object down before the refactor? I'm not sure.). We can remove the reference for now if we're not going to tally top level metrics.

simulator/src/test/java/com/hedera/block/simulator/mode/ConsumerModeHandlerTest.java
mattp-swirldslabs: If you're using IntelliJ, you can change the default behavior to only use wildcard imports when the number of classes exceeds 999. That's how I was able to "turn off" wildcard imports:
image

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -36,96 +37,119 @@
...
+ * The BlockStream Simulator Application manages the lifecycle and coordination of block streaming
+ * operations. It supports different modes of operation including publishing, consuming, or both
+ * simultaneously.
+ *
+ * <p>This class serves as the main entry point for the simulator, handling initialization,
+ * execution, and shutdown of streaming operations based on the configured mode.
+ */
 public class BlockStreamSimulatorApp {
-
+    /** Logger for this class */
```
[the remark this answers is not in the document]
georgi-l95: Yeah, you're right, bad habit from my side. Removed.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -36,96 +37,119 @@
...
+            @NonNull ConsumerStreamGrpcClient consumerStreamGrpcClient,
             @NonNull MetricsService metricsService) {
 
         requireNonNull(configuration);
         requireNonNull(blockStreamManager);
+
         this.metricsService = requireNonNull(metricsService);
         this.publishStreamGrpcClient = requireNonNull(publishStreamGrpcClient);
+        this.consumerStreamGrpcClient = requireNonNull(consumerStreamGrpcClient);
+        this.isRunning = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
georgi-l95: I was moving some stuff around and forgot to move this back. Fixed now, but it should work as expected both ways.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -36,96 +37,119 @@
...
-            case PUBLISHER -> simulatorModeHandler = new PublisherModeHandler(
+        this.simulatorModeHandler = switch (simulatorMode) {
+            case PUBLISHER -> new PublisherModeHandler(
                     blockStreamConfig, publishStreamGrpcClient, blockStreamManager, metricsService);
-            case CONSUMER -> simulatorModeHandler = new ConsumerModeHandler(blockStreamConfig);
-            case BOTH -> simulatorModeHandler = new CombinedModeHandler(blockStreamConfig);
+            case CONSUMER -> new ConsumerModeHandler(blockStreamConfig, consumerStreamGrpcClient);
+            case BOTH -> new CombinedModeHandler(blockStreamConfig);
             default -> throw new IllegalArgumentException("Unknown SimulatorMode: " + simulatorMode);
-        }
```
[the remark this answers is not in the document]
georgi-l95: You're right. Removed.

simulator/src/main/java/com/hedera/block/simulator/mode/ConsumerModeHandler.java
georgi-l95: Removed.

simulator/src/main/java/com/hedera/block/simulator/generator/BlockStreamManager.java
georgi-l95: Thanks, good point. fixed

simulator/src/main/java/com/hedera/block/simulator/generator/BlockAsFileLargeDataSets.java
georgi-l95: Fixed 👍

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamObserver.java
georgi-l95: True, added. Thank you 👍

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -36,96 +37,119 @@
...
-        switch (simulatorMode) {
-            case PUBLISHER -> simulatorModeHandler = new PublisherModeHandler(
+        this.simulatorModeHandler = switch (simulatorMode) {
+            case PUBLISHER -> new PublisherModeHandler(
                     blockStreamConfig, publishStreamGrpcClient, blockStreamManager, metricsService);
-            case CONSUMER -> simulatorModeHandler = new ConsumerModeHandler(blockStreamConfig);
-            case BOTH -> simulatorModeHandler = new CombinedModeHandler(blockStreamConfig);
+            case CONSUMER -> new ConsumerModeHandler(blockStreamConfig, consumerStreamGrpcClient);
+            case BOTH -> new CombinedModeHandler(blockStreamConfig);
             default -> throw new IllegalArgumentException("Unknown SimulatorMode: " + simulatorMode);
```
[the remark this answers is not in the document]
georgi-l95: Removed, thanks ✅

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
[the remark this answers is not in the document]
georgi-l95: Removed ✅

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
georgi-l95: No because start is a synchronous process, which will end when esentially the simulator is no longer running, therefore we set it to true just before we enter.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamGrpcClientImpl.java
georgi-l95: Removed. ✅

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamObserver.java
georgi-l95: Removed. ✅

simulator/src/main/java/com/hedera/block/simulator/mode/ConsumerModeHandler.java
georgi-l95: Removed. ✅

simulator/src/main/java/com/hedera/block/simulator/mode/CombinedModeHandler.java
georgi-l95: Removed ✅

simulator/src/main/java/com/hedera/block/simulator/mode/PublisherModeHandler.java
georgi-l95: Removed ✅

simulator/src/test/java/com/hedera/block/simulator/mode/ConsumerModeHandlerTest.java
georgi-l95: Thanks, I missed this one. fixed ✅

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamGrpcClientImpl.java
georgi-l95: Hm, you're right. Fixed it.

simulator/src/main/java/com/hedera/block/simulator/mode/ConsumerModeHandler.java
georgi-l95: I've streamlined it here.

simulator/src/main/java/com/hedera/block/simulator/mode/PublisherModeHandler.java
georgi-l95: Fixed.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
georgi-l95: Added this to this method, no need to spam all the places, issue is in the backlog and is visible https://github.com/hashgraph/hedera-block-node/issues/322

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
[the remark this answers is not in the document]
georgi-l95: Thanks, created new issue to handle this and added as a comment. #386

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamGrpcClientImpl.java
```diff
@@ -99,9 +100,8 @@ public void requestBlocks(long startBlock, long endBlock) throws InterruptedExce
 
     @Override
     public void completeStreaming() throws InterruptedException {
-        consumerStreamObserver.onCompleted();
-        // todo(352) Find a suitable solution for removing the sleep
```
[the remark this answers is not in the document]
georgi-l95: Unfortunately, it’s still the case with the publisher, but I’ll think about it another time.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamGrpcClientImpl.java
georgi-l95: Done 🤝

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
```diff
@@ -27,11 +27,11 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * we return it, else we throw {@link IllegalArgumentException}.
      *
      * @param toCheck a {@link String} to be checked if is blank as defined
-     * above
+     *                above
```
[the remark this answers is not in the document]
mattp-swirldslabs: If you're using `spotlessApply` we should see these type of reformatting changes on files we touch since Jasper and Joseph changed the formatting profile

mattp-swirldslabs (approved):
Looks good

PR #361, opened 2024-11-22
build: update helm publish registry to `ghcr.io`

### LINKED ISSUE #256: Move Publish of Helm Charts out of GH Pages into JFrog or GHCR
Adding below answer from Nathan from Release Engineering to how to set-up the GHCR.io as a repo for helm charts :) 

> Hi Freddy,
> 
> I apologize for the delay in getting back to you with the needed examples. Please see the hashgraph/solo-charts repository for an example of using ghcr.io to publish your charts. 
> 
> I have included links below to the relevant files. This repository showcases using semantic-release + helm + ghcr.io; however, the commands in the .releaserc file can be moved directly to the CI workflow in the event you are not using semantic-release.
> 
> CI Workflow: https://github.com/hashgraph/solo-charts/blob/main/.github/workflows/flow-deploy-release-artifact.yaml
> Semantic Release Configuration: https://github.com/hashgraph/solo-charts/blob/main/.releaserc
> 
> Sincerely,
> 
> Release Engineering

AlfredoG87:
**Description**:

- updated chart name to hedera-block-node-chart, in order to differentiate it from the image name hedera-block-node.
- Changed the workflow that publishes the helm chart to use ghcr.io instead of a gh-pages branch.

**Related issue(s)**:

Fixes #256 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/release-push-image.yaml
+++ b/.github/workflows/release-push-image.yaml
@@ -118,10 +118,17 @@ jobs:
       - name: Install Helm
         uses: azure/setup-helm@fe7b79cd5ee1e45176fcad797de68ecaf3ca4814 # v4.2.0
 
-      - name: Publish helm chart
-        uses: step-security/helm-gh-pages@6a390e89293c1ec8bc5120f6692f3b8a313a9a3d # v1.7.0
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@0d4c9c5ea7693da7b068278f7b52bda2a190a446 # v3.2.0
         with:
-          target_dir: charts
-          token: ${{ secrets.GITHUB_TOKEN }}
-          branch: gh-pages
-          app_version: ${{ env.VERSION }}
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Package helm chart
+        run: |
+          helm package charts/hedera-block-node
+
+      - name: Push helm chart
+        run: |
+          helm push hedera-block-node-chart-${{ env.VERSION }}.tgz oci://ghcr.io/hashgraph/hedera-block-node
--- a/.github/workflows/release-push-image.yaml
+++ b/.github/workflows/release-push-image.yaml
@@ -125,6 +125,12 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Extract version
+        id: extract_version
+        run: |
+          VERSION=$(grep 'version=' gradle.properties | cut -d '=' -f2)
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+
       - name: Package helm chart
         run: |
           helm package charts/hedera-block-node
--- a/.github/workflows/release-push-image.yaml
+++ b/.github/workflows/release-push-image.yaml
@@ -138,4 +138,4 @@ jobs:
 
       - name: Push helm chart
         run: |
-          helm push hedera-block-node-chart-${{ env.VERSION }}.tgz oci://ghcr.io/hashgraph/hedera-block-node
+          helm push block-node-helm-chart-${{ env.VERSION }}.tgz oci://ghcr.io/hashgraph/hedera-block-node
--- a/.github/workflows/release-push-image.yaml
+++ b/.github/workflows/release-push-image.yaml
@@ -133,6 +133,7 @@ jobs:
 
       - name: Package helm chart
         run: |
+          helm dependency update charts/hedera-block-node
           helm package charts/hedera-block-node
 
       - name: Push helm chart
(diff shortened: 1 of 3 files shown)


AlfredoG87:
Waiting for this ticket to be resolved before being able to finish this ticket:
https://swirldslabs.atlassian.net/servicedesk/customer/portal/4/ITSD-133

mattp-swirldslabs (dismissed):
👍

georgi-l95 (approved):

charts/hedera-block-node/README.md
```diff
@@ -11,23 +11,35 @@ minikube delete && minikube start --kubernetes-version=v1.23.0 --memory=8g --boo
 ```
 
 Set environment variables that will be used for the remainder of the document:
+Replacing the values with the appropriate values for your environment: bn1 is short for block-node-1, but you can name you release as you wish. And use the version that you want to install.
 ```bash
 export RELEASE="bn1"
```
jsync-swirlds, line 16: The name `bn1` is not great, even as a "please replace this" default.  I would suggest something more useful, such as `Hedera Block Node`.

charts/hedera-block-node/README.md
```diff
@@ -11,23 +11,35 @@ minikube delete && minikube start --kubernetes-version=v1.23.0 --memory=8g --boo
...
 ## Template
-To generate the K8 manifest files without installing the chart:
+To generate the K8 manifest files without installing the chart, you need to clone this repo and navigate to `/charts` folder.
 ```bash
 helm template --name-template my-bn hedera-block-node/ --dry-run --output-dir out
 ```
 
-## Install
-To install the chart:
+## Install using released chart
```
jsync-swirlds, line 26:
```suggestion
## Install using a published chart
```

charts/hedera-block-node/README.md
```diff
@@ -11,23 +11,35 @@ minikube delete && minikube start --kubernetes-version=v1.23.0 --memory=8g --boo
...
+
+To pull the packaged chart from public repo:
```
jsync-swirlds, line 28:
```suggestion
To pull the packaged chart from the public repo:
```

jsync-swirlds (commented):
Just some small typos and a _future_ suggestion.

.github/workflows/release-push-image.yaml
```diff
@@ -118,10 +116,24 @@ jobs:
...
-          app_version: ${{ env.VERSION }}
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract version
+        id: extract_version
+        run: |
+          VERSION=$(grep 'version=' gradle.properties | cut -d '=' -f2)
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
AlfredoG87: We need to get the version from the file, since the version of the helm to publish is in lockstep with the version of the tagged release and the version of the block-node release.

charts/hedera-block-node/README.md
AlfredoG87: Thank you, I agree with the suggested wording!
✅  Done!

charts/hedera-block-node/README.md
AlfredoG87: Thank you, I agree with the suggested wording!
✅  Done!

charts/hedera-block-node/README.md
jsync-swirlds: Hmmm, `bn1` is still not great, `bn` is OK, and `blkNod` would be better.
Adding numbers tends to create issues for managing clusters more than help.

.github/workflows/release-push-image.yaml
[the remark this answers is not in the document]
jsync-swirlds: Just because they do this doesn't mean it's ideal; both of those might just be a legacy or gradle-ism.
Why should the helm chart version _necessarily_ be identical to the tagged release and block-node release?  I would think a helm chart can have different release and update cycles than the software it deploys without breaking...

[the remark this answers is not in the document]
AlfredoG87: I agree with you that the helm version and the tagged release version of the block-node don't need to be in lockstep, however is the convention we've been using and currently is the same workflow for both artifacts... for simplicity sake and since we are using the same repo, I would argue against dividing the workflows and separating the versions unless there is a clear need that compels us to do it.

charts/hedera-block-node/README.md
AlfredoG87: Yes, Agreed, will update it to `blkNod` since even when is only an example, most people might just use it as is 😓

AlfredoG87: ✅ Done!

.github/workflows/release-push-image.yaml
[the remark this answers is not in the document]
jsync-swirlds: That's reasonable.  This will work fine for now; though we may want to revisit in the future, as we further mature the product and processes.

jsync-swirlds (approved):

georgi-l95 (approved):

mishomihov00 (approved):
LGTM

ata-nas (approved):
LG

PR #365, opened 2024-11-27
feat: server properties configurable

### LINKED ISSUE #344: feat: Make maxMessageSizeBytes configurable via a property in app.properties
As an engineer running a Block Node
I want to be able to configure the `maxMessageSizeBytes` for the Helidon PBJ Plugin
So that I can tune this parameter for performance

## Tech Notes
* This property is currently exposed in `BlockNodeApp.start()`:

```
// Override the default message size
final PbjConfig pbjConfig = PbjConfig.builder()
                .name(PBJ_PROTOCOL_PROVIDER_CONFIG_NAME)
                .maxMessageSizeBytes(1024 * 4096)
                .build();
```
* We already have an established pattern for ingesting parameters found in app.properties in the Server.java class. This feature should follow the existing pattern.
* The property will need to be an integer with a min value of 10240 bytes (PBJ default) and a max possibly determined by the HTTP/2 spec for frame size, etc.
* The current default should be 4,194,304 (1024 * 4096)
* The property might be named `server.maxMessageSizeBytes`

### LINKED ISSUE #345: feat: Make server port configurable via a property in app.properties
As an engineer running a Block Node
I want to be able to configure the `port` for the Block Node
So that I can run the server on a different port

## Tech Notes
* This property is currently exposed in `BlockNodeApp.start()`:

```
 // Build the web server
        // TODO: make port server a configurable value.
        final WebServer webServer = webServerBuilder
                .port(8080)
                .addProtocol(pbjConfig)
                .addRouting(pbjRouting)
                .addRouting(httpRouting)
                .build();
```
* We already have an established pattern for ingesting parameters found in app.properties in the Server.java class. This feature should follow the existing pattern.
* The property will need to be sanity checked as a positive integer with a min value of 1 and a reasonable max port value
* The current default 8080
* The property might be named `server.port`

AlfredoG87:
**Description**:

- Adding a new Config Class called ServerConfig for Server specific properties.
- Adding 2 new properties to the newly created ServerConfig, and mapping them for ENV override using the ServerMappedConfigSourceInitializer. maxMessageSizeBytes and port.
- Replacing the hardcoded values on the codebase with the new configurable properties from ServerConfig.
- Added the new configuration to the ConfigInjectionModule
- Created UT for testing the Config validations and restrictions.
- Added Test case for ConfigInjectionModule to retrieve the newly config object.

**Related issue(s)**:

Fixes #344 
Fixes #345 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/server/src/main/java/com/hedera/block/server/ServerConfig.java
+++ b/server/src/main/java/com/hedera/block/server/ServerConfig.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.block.server;
+
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import com.swirlds.config.api.validation.annotation.Max;
+import com.swirlds.config.api.validation.annotation.Min;
+
+/**
+ * Use this configuration across the server features
+ *
+ * <p>ServerConfig will have settings for the server.
+ *
+ * @param maxMessageSizeBytes the http2 max message/frame size in bytes
+ * @param port the port the server will listen on
+ */
+@ConfigData("server")
+public record ServerConfig(
+        @ConfigProperty(defaultValue = "4_194_304") @Min(10_240) @Max(16_777_215) int maxMessageSizeBytes,
+        @ConfigProperty(defaultValue = "8080") @Min(1024) @Max(65_535) int port) {
+    private static final System.Logger LOGGER = System.getLogger(ServerConfig.class.getName());
+
+    /**
+     * Validate the configuration.
+     *
+     * @throws IllegalArgumentException if the configuration is invalid
+     */
+    public ServerConfig {
+
+        validateMaxMessageSizeBytes(maxMessageSizeBytes);
+        validatePort(port);
+
+        LOGGER.log(System.Logger.Level.INFO, "Server configuration server.maxMessageSizeBytes: " + maxMessageSizeBytes);
+        LOGGER.log(System.Logger.Level.INFO, "Server configuration server.port: " + port);
+    }
+
+    private void validatePort(int port) {
+        if (port < 1024) {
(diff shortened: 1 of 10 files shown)


server/src/main/java/com/hedera/block/server/ServerConfig.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * <p>ServerConfig will have settings for the server.
+ *
+ * @param maxMessageSizeBytes the http2 max message/frame size in bytes
+ * @param port the port the server will listen on
+ */
+@ConfigData("server")
+public record ServerConfig(
+        @ConfigProperty(defaultValue = "4_194_304") @Min(10_240) @Max(16_777_215) int maxMessageSizeBytes,
+        @ConfigProperty(defaultValue = "8080") @Min(1024) @Max(65_535) int port) {
```
ata-nas, line 35: You could extract these numbers in a private static final field. They are used in a couple of places, would be great to be managed from a single place to not be error prone in case of change.

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
```diff
@@ -26,15 +26,14 @@
...
-                    new ConfigMapping("mediator.ringBufferSize", "MEDIATOR_RING_BUFFER_SIZE"),
-                    new ConfigMapping("notifier.ringBufferSize", "NOTIFIER_RING_BUFFER_SIZE"));
+    private static final List<ConfigMapping> MAPPINGS = List.of(
+            new ConfigMapping("consumer.timeoutThresholdMillis", "CONSUMER_TIMEOUT_THRESHOLD_MILLIS"),
+            new ConfigMapping("persistence.storage.rootPath", "PERSISTENCE_STORAGE_ROOT_PATH"),
+            new ConfigMapping("service.delayMillis", "SERVICE_DELAY_MILLIS"),
+            new ConfigMapping("mediator.ringBufferSize", "MEDIATOR_RING_BUFFER_SIZE"),
+            new ConfigMapping("notifier.ringBufferSize", "NOTIFIER_RING_BUFFER_SIZE"),
+            new ConfigMapping("server.maxMessageSizeBytes", "SERVER_MAX_MESSAGE_SIZE_BYTES"),
+            new ConfigMapping("server.port", "SERVER_PORT"));
```
ata-nas, line 36: A suggestion: let's update the documentation as well. There is a `configuration.md`.

ata-nas (commented):
@AlfredoG87 looks good! A couple of nits.

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
AlfredoG87: ✅  Done!

server/src/main/java/com/hedera/block/server/ServerConfig.java
AlfredoG87: Thanks, this is indeed a great suggestion indeed :)

✅ Done!

ata-nas (approved):
@AlfredoG87 looks great!

Issue #281, opened 2024-10-21
feat: new BlockWriter implementation for block-as-file


AlfredoG87:
## Story
**AS A** Block Node Operator
**I WANT** to reduce the number of system calls needed to persist a single block
**SO THAT** I can minimize operating costs.

This task is to change block persistence to persist the blocks as whole files rather than directories of files per item (`BlockAsDirWriter`).

## Tech Notes:

- Utilize the directory structure from #125. Blocks will have a number (long) which will be also the block file's name. The trie we will have to facilitate these files will be a single digit per dir (node) for a depth of 18 (second to last possible digit 
position for a `long`, which has 19 digits) and the final digit will be part of the block file's name.

- For this issue's scope we will only need to support writing the block files to disk and we can disregard the rest of the notes of #125 which will be part of subsequent issues, thus being said, we are mainly interested in:
   - When writing, initially we only care to write a new block file to the fs in the correct place, based on it's block number.
   - When writing a new block file, we need to check if the path to the file we want to write already exists, if it does, there is some problem, we should handle this case directly in our service.
   - When writing, we need to make sure that we will be making available the needed path for the trie ad hoc.

- Currently we have a `interface BlockWriter<V>` that is writing block items as files inside a directory with the number of the block, this causes multiple (potentially) thousands of system calls per block, slowing down the whole system (does it? Are system calls actually expensive or are we just assuming they are?).

- Current and only implementation `BlockAsDirWriter` was meant to support development and help on debugging by making it easy to see the block items as they are arriving, however due to the high TPS needed for the system, we want to create an _additional_ `BlockAsFileWriter` implementation of `BlockWriter<List<BlockItem>>` to provide more performant support for production purposes.

- **Some notes on permissions:**
   - Do not pre-emptively attempt to `fix` file permissions. (this only creates more system calls that are completely not needed), instead if there is a IO issue, attempt to fix permissions (in case is that) before retrying and continuing.
   - Don't set file permissions too permissive, however, it's bad practice and dangerous to have folders or files writable by other processes.  Set appropriate umask up front and ensure files are created with appropriate permissions.

- **Some notes on memory usage:**
   - Keep block items in memory until you receive a Block Proof, then persist to disk (perhaps after verification, perhaps before).
   - Holding items in memory is less reliable, so we should be careful to ensure, data is reliable, that the storage of in-memory items is not in the wrong component, and data is not duplicated unnecessarily.
[...]

PR #385, opened 2024-12-06
feat: new BlockReader implementation for block-as-file

### LINKED ISSUE #349: feat: new BlockReader implementation for block-as-file
## Story

**AS A** Block Node Operator/User
**I WANT** to be able to retrieve blocks written as files
**SO THAT** I will be able to use the blocks

## Tech Notes

- **DEPENDS ON** #281

- See #125 where we have deigned what the functionality of the `block-as-file` format should be.

- The scope of this issue is to be able to accurately, performantly and reliably read files from the file system, that represent any given `block-as-file`.

- Initially, we should be able to read only the raw files that are written to the file system, then, as the file writing logic gets extended where it is being compressed optionally and then also archived, we should extend the reading functionality to support that as well.

ata-nas:
**Description**:

- adding reading logic to complement #281 

**Related issue(s)**:

Fixes #349 

**Notes for reviewer**:

- what would be the best way to parse the block file, having a file input stream and calling readAllBytes does not seem like the best option, there are other (based on the api), but which one would be preferable?

**Checklist**

### DIFF
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
@@ -66,12 +66,15 @@ public final class BlockAsLocalFileReader implements LocalBlockReader<BlockUnpar
         if (Files.exists(resolvedBlockPath)) {
             return Optional.of(doRead(resolvedBlockPath));
         } else {
+            // todo we generally expect the block to be there, so we
+            // return an empty optional here, or do something else?
+            // of course sometimes the block may not even exist yet.
             return Optional.empty();
         }
     }
 
     private BlockUnparsed doRead(final Path resolvedBlockPath) throws IOException, ParseException {
-        // todo there are other ways of parsing, for example we can pass ReadableSequentialData instead of the fos
+        // todo there are other ways of parsing, for example we can pass ReadableSequentialData instead of the fis
         // because the readAllBytes does not seem like the best idea, what is the correct (or better) way to parse
         // the block file?
         try (final FileInputStream fis = new FileInputStream(resolvedBlockPath.toFile())) {
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
@@ -17,34 +17,65 @@
 package com.hedera.block.server.persistence.storage.read;
 
 import com.hedera.block.common.utils.Preconditions;
+import com.hedera.block.server.persistence.storage.path.BlockPathResolver;
 import com.hedera.hapi.block.BlockUnparsed;
+import com.hedera.pbj.runtime.ParseException;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Objects;
 import java.util.Optional;
 
 /**
  * A Block reader that reads block-as-file.
  */
 public final class BlockAsLocalFileReader implements LocalBlockReader<BlockUnparsed> {
+    private final BlockPathResolver pathResolver;
+
     /**
      * Constructor.
+     *
+     * @param pathResolver valid, {@code non-null} instance of
+     * {@link BlockPathResolver} used to resolve paths to block files
      */
-    private BlockAsLocalFileReader() {}
+    private BlockAsLocalFileReader(@NonNull final BlockPathResolver pathResolver) {
+        this.pathResolver = Objects.requireNonNull(pathResolver);
+    }
 
     /**
      * This method creates and returns a new instance of
      * {@link BlockAsLocalFileReader}.
      *
+     * @param pathResolver valid, {@code non-null} instance of
+     * {@link BlockPathResolver} used to resolve paths to block files
      * @return a new, fully initialized instance of
      * {@link BlockAsLocalFileReader}
      */
-    public static BlockAsLocalFileReader newInstance() {
-        return new BlockAsLocalFileReader();
+    public static BlockAsLocalFileReader of(@NonNull final BlockPathResolver pathResolver) {
+        return new BlockAsLocalFileReader(pathResolver);
     }
 
     @NonNull
     @Override
-    public Optional<BlockUnparsed> read(final long blockNumber) {
+    public Optional<BlockUnparsed> read(final long blockNumber) throws IOException, ParseException {
         Preconditions.requireWhole(blockNumber);
-        throw new UnsupportedOperationException("Not implemented yet");
+        final Path resolvedBlockPath = pathResolver.resolvePathToBlock(blockNumber);
+        if (Files.exists(resolvedBlockPath)) {
+            return Optional.of(doRead(resolvedBlockPath));
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    private BlockUnparsed doRead(final Path resolvedBlockPath) throws IOException, ParseException {
+        // todo there are other ways of parsing, for example we can pass ReadableSequentialData instead of the fos
+        // because the readAllBytes does not seem like the best idea, what is the correct (or better) way to parse
+        // the block file?
+        try (final FileInputStream fis = new FileInputStream(resolvedBlockPath.toFile())) {
+            return BlockUnparsed.PROTOBUF.parse(Bytes.wrap(fis.readAllBytes()));
+        }
     }
 }
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
@@ -20,9 +20,8 @@ import com.hedera.block.common.utils.Preconditions;
 import com.hedera.block.server.persistence.storage.path.BlockPathResolver;
 import com.hedera.hapi.block.BlockUnparsed;
 import com.hedera.pbj.runtime.ParseException;
-import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.hedera.pbj.runtime.io.stream.ReadableStreamingData;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -66,19 +65,13 @@ public final class BlockAsLocalFileReader implements LocalBlockReader<BlockUnpar
         if (Files.exists(resolvedBlockPath)) {
             return Optional.of(doRead(resolvedBlockPath));
         } else {
-            // todo we generally expect the block to be there, so we
-            // return an empty optional here, or do something else?
-            // of course sometimes the block may not even exist yet.
             return Optional.empty();
         }
     }
 
     private BlockUnparsed doRead(final Path resolvedBlockPath) throws IOException, ParseException {
-        // todo there are other ways of parsing, for example we can pass ReadableSequentialData instead of the fis
-        // because the readAllBytes does not seem like the best idea, what is the correct (or better) way to parse
-        // the block file?
-        try (final FileInputStream fis = new FileInputStream(resolvedBlockPath.toFile())) {
-            return BlockUnparsed.PROTOBUF.parse(Bytes.wrap(fis.readAllBytes()));
+        try (final ReadableStreamingData data = new ReadableStreamingData(resolvedBlockPath)) {
+            return BlockUnparsed.PROTOBUF.parse(data);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
```diff
@@ -17,34 +17,65 @@
...
+            return Optional.of(doRead(resolvedBlockPath));
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    private BlockUnparsed doRead(final Path resolvedBlockPath) throws IOException, ParseException {
+        // todo there are other ways of parsing, for example we can pass ReadableSequentialData instead of the fos
+        // because the readAllBytes does not seem like the best idea, what is the correct (or better) way to parse
+        // the block file?
```
ata-nas, line 76: Opinions?

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
```diff
@@ -17,34 +17,68 @@
...
+    public Optional<BlockUnparsed> read(final long blockNumber) throws IOException, ParseException {
         Preconditions.requireWhole(blockNumber);
-        throw new UnsupportedOperationException("Not implemented yet");
+        final Path resolvedBlockPath = pathResolver.resolvePathToBlock(blockNumber);
+        if (Files.exists(resolvedBlockPath)) {
+            return Optional.of(doRead(resolvedBlockPath));
+        } else {
+            // todo we generally expect the block to be there, so we
+            // return an empty optional here, or do something else?
+            // of course sometimes the block may not even exist yet.
```
ata-nas, line 71: Opinions?

AlfredoG87: I agree that returning an empty optional is correct.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
```diff
@@ -17,34 +17,68 @@
...
+            return Optional.empty();
+        }
+    }
+
+    private BlockUnparsed doRead(final Path resolvedBlockPath) throws IOException, ParseException {
+        // todo there are other ways of parsing, for example we can pass ReadableSequentialData instead of the fis
+        // because the readAllBytes does not seem like the best idea, what is the correct (or better) way to parse
+        // the block file?
+        try (final FileInputStream fis = new FileInputStream(resolvedBlockPath.toFile())) {
+            return BlockUnparsed.PROTOBUF.parse(Bytes.wrap(fis.readAllBytes()));
```
AlfredoG87, line 81: I think that the biggest block files will be around the 10Mb, so I would say we don't need to bother with a Buffered approach for now. But might be good to create a ticket with future tag to keep it in the radar 👍

AlfredoG87 (approved):
LGTM

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
jsync-swirlds: I am assuming, here, that you intend to just wrap the input stream in the readable sequential data and pass that to the parse method instead of calling `readAllBytes`.

I believe that constructing a `ReadableSequentialData` here, instead of reading the full file as bytes before parsing, will save a byte array copy, and should improve memory efficiency for minimal change in complexity.

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReaderTest.java
```diff
@@ -29,25 +51,64 @@
...
      * a given block number.
-     *
-     * @param toRead parameterized, block number
      */
-    @ParameterizedTest
-    @MethodSource("validBlockNumbers")
-    void testSuccessfulBlockRead(final long toRead) {
-        // todo currently throws UnsupportedOperationException
-        assertThatExceptionOfType(UnsupportedOperationException.class).isThrownBy(() -> toTest.read(toRead));
+    @Test
```
jsync-swirlds, line 82: Why remove the parameterized test case source?  This runs fewer scenarios and reduces the number of requirements validated...

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
jsync-swirlds: Already removed, but I would agree that a missing block file is _normal_ and should be handled without exceptions.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
jsync-swirlds: We found in CN that we need to buffer reads because otherwise the parse is _very_ costly as it reads one byte at a time (which causes excessive lock contention in the java internals).

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
ata-nas: ✅ Indeed, wrapping the input stream is what I meant. I've addressed that, thanks!

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
ata-nas: ✅ Thanks for the clarification @AlfredoG87 @jsync-swirlds

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
ata-nas: ✅ Thanks, I did migrate to the readable sequential data.

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReaderTest.java
ata-nas: The proper setup if I have kept the parameterized test was too difficult to do. I do actually write to the fs here, so it requires writing the test data first and then reading it. I'll see how to improve this with more test cases. Also, I will make another test to assert that an empty optional is returned if the block file does not exist (but block number is valid)

mattp-swirldslabs (approved):
Looks good

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReaderTest.java
jsync-swirlds: Should have been relatively straightforward, mostly the test you have, but the block number is a parameter (and that determines the file name).
It will write a file, but it should also remove that file at the end of the test method, and in between it is basically unchanged...

##### General note
The use of parameterized tests is encouraged for _all_ unit tests; if the unit test is hard to parameterize, it probably needs to be rethought (not always, but in most cases, it can be parameterized with less change than you might expect).

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReaderTest.java
```diff
@@ -29,25 +51,64 @@
...
+                .isExactlyInstanceOf(BlockUnparsed.class)
+                .returns(1L, from(blockUnparsed -> {
+                    try {
+                        return BlockHeader.PROTOBUF
+                                .parse(Objects.requireNonNull(
+                                        blockUnparsed.blockItems().getFirst().blockHeader()))
+                                .number();
+                    } catch (final ParseException e) {
+                        throw new RuntimeException(e);
+                    }
```
jsync-swirlds, line 104: FYI, unit tests generally should not catch exceptions; let the exception be thrown from the test method and the test runner will give you better information in the test report.

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReaderTest.java
```diff
@@ -29,25 +51,64 @@
...
+                }))
+                .extracting(BlockUnparsed::blockItems)
+                .asList()
+                .isNotNull()
+                .isNotEmpty()
+                .hasSize(blockItemUnparsed.size())
+                .containsExactlyElementsOf(blockItemUnparsed);
```
jsync-swirlds, line 111: These two items are almost certainly overly specific and make the test brittle.
Verifying that what is read is _exactly_ what was written is better done in a separate test _by itself_ with its own input to reduce the probability of tests broken in ways that have nothing to do with whether the design is correct.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
```diff
@@ -17,34 +17,61 @@
...
-    public Optional<BlockUnparsed> read(final long blockNumber) {
+    public Optional<BlockUnparsed> read(final long blockNumber) throws IOException, ParseException {
         Preconditions.requireWhole(blockNumber);
-        throw new UnsupportedOperationException("Not implemented yet");
+        final Path resolvedBlockPath = pathResolver.resolvePathToBlock(blockNumber);
+        if (Files.exists(resolvedBlockPath)) {
+            return Optional.of(doRead(resolvedBlockPath));
+        } else {
+            return Optional.empty();
+        }
```
jsync-swirlds, line 69: Small note:
A future refactor should push the "find this file" into the resolver, so if the `Path` is not "empty" from the resolver (which probably should return an `Optional`), we are guaranteed that the file exists.  That both supports the resolver testing multiple paths (e.g. for zip archives) and also simplifies this method.

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReaderTest.java
ata-nas: ✅ I have streamlined this, the parameterized tests are back and I have added another helper method to generate block items for a block with specific number to facilitate this.

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReaderTest.java
ata-nas: I agree, however the `returns` test requires a `Function` which unfortunately does not throw, and the `parse` method throws an unchecked exception. I do re-throw it immediately however so we do not silence the exception (as it should be the case).

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
ata-nas: That is a really good point! I will make sure to add that in the upcoming PRs where we will be also using the archive root (we will need this logic there as well) and will migrate such uses of the `if-else` where we ask if the path to a block resolved exists.

server/src/test/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReaderTest.java
ata-nas: ✅ That is a good point, I split the tests. I do agree that the tests should be testing specific aspects of a given method and when there are such clearly defined and visible aspects of a method, like the method relies on something to be present on the filesystem and only then it can read the contents, then these are indeed two different tests, because if one breaks, this would very easily indicate which aspect of the intended functionality was broken.

However for the brittleness, while I do agree with you that it is good to have locally generated data, I have added another utility method to generate the test data so it could be reused. I think that if there is another case, where we need test data generated differently, we should add another method and not tamper with the logic of the already existing utility method that generates our data. For such a complex data generation logic, I do actually think it is better to indeed be in a publicly accessible utility, because otherwise if something changes with the data model, it will be a very big hustle to support.

AlfredoG87 (approved):
a few nits, looks good!

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
AlfredoG87: remove the comments?

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
```diff
@@ -17,34 +17,68 @@
...
+            // return an empty optional here, or do something else?
+            // of course sometimes the block may not even exist yet.
+            return Optional.empty();
+        }
+    }
+
+    private BlockUnparsed doRead(final Path resolvedBlockPath) throws IOException, ParseException {
+        // todo there are other ways of parsing, for example we can pass ReadableSequentialData instead of the fis
+        // because the readAllBytes does not seem like the best idea, what is the correct (or better) way to parse
+        // the block file?
```
AlfredoG87: remove the comments?

ata-nas: @AlfredoG87 you seem to be looking at an old version of this file, changes were already pushed before you have submitted your review.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
ata-nas: @AlfredoG87 you seem to be looking at an old version of this file, changes were already pushed before you have submitted your review.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
ata-nas: @AlfredoG87 you seem to be looking at an old version of this file.

AlfredoG87 (approved):
LGTM

Issue #309, opened 2024-10-24
EPIC: Essential Local Block Storage & Retrieval Support


jsync-swirlds:
## Epic Goal
Design/implementation for essential block storage and retrieval support to enable the Phase 2 block node features (pre-release).

In this first iteration of the persistence layer, it is important to lay the ground work, to streamline the process and to see it work. Once everything is wired up and running, we will be able to iterate over this base in later iterations to introduce improvements to the persistence layer and clear out any uncertainties.

See #125 where we have defined the design of the functionality that now we need to implement

```[tasklist]
### Tasks (in recommended order of implementation)
- [ ] https://github.com/hashgraph/hedera-block-node/issues/281
- [ ] https://github.com/hashgraph/hedera-block-node/issues/319
- [ ] https://github.com/hashgraph/hedera-block-node/issues/348
- [ ] https://github.com/hashgraph/hedera-block-node/issues/349
- [ ] https://github.com/hashgraph/hedera-block-node/issues/282
- [ ] https://github.com/hashgraph/hedera-block-node/issues/350
- [ ] https://github.com/hashgraph/hedera-block-node/issues/339
- [ ] https://github.com/hashgraph/hedera-block-node/issues/351
- [ ] https://github.com/hashgraph/hedera-block-node/issues/370
- [ ] https://github.com/hashgraph/hedera-block-node/issues/381
```

PR #387, opened 2024-12-09
feat: add compression mode for BlockAsFileWriter

### LINKED ISSUE #282: feat: add compression mode for BlockAsFileWriter
## Story
**AS A** Block Node Operator
**I WANT** to persist blocks in compressed form
**SO THAT** I can minimize disk storage cost

## Tech Notes

- **DEPENDS ON #281**

- See the work done by @jasperpotts on compression to use that Algorithm. see [here](https://www.notion.so/swirldslabs/On-Disk-Block-Archive-Format-11c7c9ab259180319698cd5dad76a75f)

- Enable or Disable compression using a configurable property, be default should be enabled.

- This compression should only be added in the _new_ implementation class designed for #281 and should follow that ticket.

- Conditionally a NoOp implementation could be injected if we want to disable the compression.

ata-nas:
**Description**:

- add compression support for the block-as-file writer

**Related issue(s)**:

Fixes #282 

**Notes for reviewer**:

TBD

**Checklist**

---

- [x] add a new compression component
- [x] add configuration for compression
- [x] ? (maybe next ticket part of the #309 epic, reading support for compressions will require it, lets not block the follow up PRs) extend the path resolver to be able to "find" a block, now we have multiple file extensions, so all must be supported


server/src/main/java/com/hedera/block/server/persistence/storage/compression/ZstdCompression.java
```diff
@@ -0,0 +1,38 @@
...
+ * TODO: add documentation
+ */
+public class ZstdCompression implements Compression {
+    private static final String ZSTD_EXTENSION = ".zstd";
+
+    @NonNull
+    @Override
+    public OutputStream newCompressingOutputStream(@NonNull final Path pathToFile) throws IOException {
+        final Path localPathToFile = pathToFile.resolveSibling(pathToFile.getFileName() + ZSTD_EXTENSION);
+        return new ZstdOutputStream(Files.newOutputStream(localPathToFile));
```
ata-nas, line 36: The `zstd` lib allows us to set a custom `compression level` for the out stream, do we need anything other than their default?

server/src/main/java/com/hedera/block/server/persistence/storage/compression/Compression.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public interface Compression {
+    /**
+     * This method aims to create a new {@link OutputStream} instance that will
+     * run the input data through a compression algorithm (algorithm is based on
+     * specific implementation) before writing it to the input path. The input
+     * path must be a file and not a directory. This method and all it`s
+     * implementations are NOT responsible for creating the path to
+     * the parent directory of the input file path provided, but WILL create the
+     * actual file if it does not exist.
```
ata-nas, line 36: I am still a bit uncertain about this method api contract about not being responsible for creating the path to the parent of the desired file where we want to write the compressed data. Internally, both implementations use `Files#newOutputStream(Path)` which will create the actual file we want to write to if it does not exist, but it will throw an exception if the path to the parent folder of our target path does not exist. Should it be part of the api that the new output stream shall ensure also the creation of the parent of the desired target? My point here is if we can ensure that all possible other implementations can (and should) enforce that logic, since this is the highest level abstraction and all specific implementations **must** comply with this contract. Opinions?

server/src/main/java/com/hedera/block/server/persistence/storage/compression/AbstractCompression.java
```diff
@@ -0,0 +1,51 @@
...
+        final String errorMessage;
+        if (Files.isDirectory(pathToFile)) {
+            errorMessage = "The input path [%s] must not be a directory!".formatted(pathToFile);
+        } else if (Files.notExists(pathToFile.getParent())) {
+            errorMessage = "The path to the parent directory of the input path [%s] must exist!".formatted(pathToFile);
+        } else {
+            return createNewCompressingOutputStream(pathToFile);
+        }
+        throw new IllegalArgumentException(errorMessage);
+    }
```
ata-nas, line 50: My whole idea for this abstract class is to ensure that preconditions that are part of the highest level api are always enforced. I want to make sure that it is not error prone to enforce preconditions, however I have a bit of a strange `code smell` feeling here. Opinions?

server/src/main/java/com/hedera/block/server/persistence/storage/compression/ZstdCompression.java
jsync-swirlds: I believe we wanted this to be configurable, with default of `6`.  Min is `0` and Max is `9`, if I recall correctly.

server/src/main/java/com/hedera/block/server/persistence/storage/compression/Compression.java
jsync-swirlds: Why is Compression creating files?  I would expect the Resolver to have a method for creating a file...
My reasoning is that not all implementations will create files, and having the Resolver do so (optionally, via a method call) seems the most logical place to add that function.

server/src/main/java/com/hedera/block/server/persistence/storage/compression/ZstdCompression.java
ata-nas: Thanks, that makes sense, I will look into that.

server/src/main/java/com/hedera/block/server/persistence/storage/compression/AbstractCompression.java
jsync-swirlds: I would expect `CompressingOutputStream` to wrap an `OutputStream`, not a `Path`.  This is assuming that all compressed streams are files; which is not always true.
At minimum we already know we must support S3 style bucket outputs; and we might need or want compression for quite a few other non-file options in the future.

server/src/main/java/com/hedera/block/server/persistence/storage/compression/Compression.java
ata-nas: Yep. My point exaclty. I do not think that file creation should be the responsibility of the `Compression` abstraction. In fact that was my initial implementation. I think I will revert to that... It also does not seem _appropriate_ that all specific implementations would fit this contract.

server/src/main/java/com/hedera/block/server/persistence/storage/compression/AbstractCompression.java
ata-nas: Hmm... yes, in addition to the comment above, I think this was the missing link on my side, assuming that the contract `Compression` is appropriate to generate streams based on a `Path`. In fact, now that I think of it, instead of the method to be called `newCompressingOutputStream`, I think it would be a better fit to name the method as `wrap` or similar, and as input, to accept an `OutputStream`. In that way I also decouple myself from the responsibility and the questionable contract behavior to _partially_ create the path to the destination (that is if the specific implementation will actually target a file on the filesystem...). Thanks for that! :)

server/src/main/java/com/hedera/block/server/persistence/storage/compression/ZstdCompression.java
ata-nas: @jsync-swirlds the java zstd lib does not provide the best documentation, but after running the following:

```java
public static void main(String[] args) {
    System.out.println(Zstd.minCompressionLevel());
    System.out.println(Zstd.defaultCompressionLevel());
    System.out.println(Zstd.maxCompressionLevel());
}
```

I get this as a result:

```text
-131072
3
22
```

I suppose I _should_ trust these values... document the range and have a precondition check for the provided value externally, this best fails before we have a running app...

AlfredoG87: I think we should add the link to the documentation on the comments of the configurable property and try out with 6.

AlfredoG87 (changes requested):
I agree that the Compression class should receive a stream and return a stream. otherwise looks good.

server/build.gradle.kts
```diff
@@ -24,6 +24,9 @@ application {
     mainClass = "com.hedera.block.server.Server"
 }
 
+// todo what is the correct way to add this dependency?
+dependencies { implementation("com.github.luben:zstd-jni:1.5.6-8") }
+
```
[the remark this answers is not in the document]
ata-nas: Thanks!

server/src/main/java/com/hedera/block/server/persistence/storage/compression/ZstdCompression.java
ata-nas: @jsync-swirlds @AlfredoG87 I added the configuration for the compression level, left a `todo` there since I am not certain what should be the final version for the default (what it should be based on) and if we have defined a range internally that should be used, otherwise maybe we should use the lib's ranges by invoking it's methods.

jsync-swirlds: My recollection of min/max is very likely faulty.  We should use what is documented for the library.
We'll have to test and optimize a bit regardless, so it doesn't matter much what we start with as long as it is not at the extremes.

ata-nas: ✅ Done. I am using the lib's defaults for now.

server/src/main/java/com/hedera/block/server/persistence/storage/compression/Compression.java
ata-nas: ✅ Done. The compression now wraps a stream instead.

server/src/main/java/com/hedera/block/server/persistence/storage/compression/AbstractCompression.java
ata-nas: ✅ Done. We have moved to another implementation.

server/src/main/java/com/hedera/block/server/persistence/PersistenceInjectionModule.java
```diff
@@ -145,6 +150,23 @@ static BlockPathResolver providesPathResolver(@NonNull final PersistenceStorageC
...
+     * compression
+     * @return a compression singleton
+     */
+    @Provides
+    @Singleton
+    static Compression providesCompression(@NonNull final PersistenceStorageConfig config) {
+        final CompressionType compressionType = config.compression();
+        return switch (compressionType) {
+            case ZSTD -> new ZstdCompression();
+            case NONE -> new NoOpCompression();
```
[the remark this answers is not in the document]
ata-nas: Think of this from the perspective of the user/operator. `NO_OP` would imply that compression **must** be present and the app cannot run without it. `NO_OP` means there **is** a compression going on that is of type `NO_OP`. That is not always the case. Compression is not a part of the application in the same sense as persistence/retrieval is. Persistence/retrieval (whatever the implementation is, even `NO_OP`) is a composing part of the application. The application cannot exist without it. Whereas with compression, the app **can** exist perfectly fine. So when I say 'persistence.storage.type=NO_OP` I mean that the absolutely necessary part of the app (that is persistence type) is of type `NO_OP`. Whereas when I say `persistence.storage.compression=NONE` means from the user's perspective, the app will run with **no** compression capability whatsoever. That is quite different and semantically feels a lot better. I am pro keeping this. But if others also disagree with me, I will change it. Opinions?

server/src/main/java/com/hedera/block/server/persistence/PersistenceInjectionModule.java
```diff
@@ -145,6 +150,23 @@ static BlockPathResolver providesPathResolver(@NonNull final PersistenceStorageC
         };
     }
 
+    /**
+     * Provides a compression singleton using the compression config.
```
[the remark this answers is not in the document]
ata-nas: ✅ Done. Typo on my side, must be `persistence config`.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -43,7 +44,8 @@ public record PersistenceStorageConfig(
         @ConfigProperty(defaultValue = "") String liveRootPath,
         // @todo(#371) - the default life/archive root path must be absolute starting from /opt
         @ConfigProperty(defaultValue = "") String archiveRootPath,
-        @ConfigProperty(defaultValue = "BLOCK_AS_LOCAL_FILE") StorageType type) {
+        @ConfigProperty(defaultValue = "BLOCK_AS_LOCAL_FILE") StorageType type,
+        @ConfigProperty(defaultValue = "ZSTD") CompressionType compression) {
```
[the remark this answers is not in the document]
ata-nas: Thinking from the perspective of a user/operator, `compressionType` would imply that compression is an integral composing part of the app, which is not the case. See [this comment](https://github.com/hashgraph/hedera-block-node/pull/387#discussion_r1882165302) for more detailed info. If others disagree I could migrate both. For semantics, I think simply `compression` fits better.

server/src/main/java/com/hedera/block/server/persistence/PersistenceInjectionModule.java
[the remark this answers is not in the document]
jsync-swirlds: I think None makes sense here for the enumeration.
I might suggest renaming the `NoOpCompression` class to `NoneCompression` to be consistent, but that is a minor consideration.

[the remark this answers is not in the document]
ata-nas: @jsync-swirlds I would go with `NoOpCompression`, on our side as developers that sounds better IMO.

AlfredoG87 (dismissed):
LGTM

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
```diff
@@ -185,6 +185,50 @@ public static long requireGreaterOrEqual(final long toTest, final long base, fin
...
+            return toCheck;
+        } else {
+            final String message = Objects.isNull(errorMessage)
+                    ? "The input int [%d] is required to be in the range [%d, %d] boundaries included."
+                            .formatted(toCheck, lowerBoundary, upperBoundary)
+                    : errorMessage;
+            throw new IllegalArgumentException(message);
+        }
+    }
+
```
jsync-swirlds, line 231: Just FYI as an example.  Something to work on in _future_ PRs.
Here is how we could make `errorMessage` `@NonNull` and still be efficient:
```suggestion
    private static final String OUT_OF_RANGE_DEFAULT_MESSAGE =
            "The input value [%d] is required to be in the range [%d, %d], inclusive.";
    /**
     * This method asserts a given int is within a range (boundaries
     * included). If the given int is within the range, then we return it,
     * else, an {@link IllegalArgumentException} is thrown.
     *
     * @param toCheck the int value to test
     * @param lowerBoundary the lower boundary
     * @param upperBoundary the upper boundary
     * @return the input {@code toCheck} if it is within the range (boundaries
     * included)
     * @throws IllegalArgumentException if the input int does not pass the test
     */
    public static int requireInRange(final int toCheck, final int lowerBoundary, final int upperBoundary) {
        return requireInRange(toCheck, lowerBoundary, upperBoundary, OUT_OF_RANGE_DEFAULT_MESSAGE);
    }

    /**
     * This method asserts a given int is within a range (boundaries
     * included). If the given int is within the range, then we return it,
     * else, an {@link IllegalArgumentException} is thrown.
     *
     * @param toCheck the int value to check
     * @param lowerBoundary the lower boundary
     * @param upperBoundary the upper boundary
     * @param errorMessage A format string {@see Formatter#syntax} with three
     *     decimal parameters for {@code toCheck}, {@code lowerBoundary},
     *     and {@code upperBoundary}.<br/>
     *     Example: {@code "The input (%d) is out of range (%d, %d)"}
     * @return the input {@code toCheck} if it is within the range (boundaries
     * included)
     * @throws IllegalArgumentException if the input int does not pass the test
     */
    public static int requireInRange(
            final int toCheck, final int lowerBoundary, final int upperBoundary, final String errorMessage) {
        if (toCheck >= lowerBoundary && toCheck <= upperBoundary) {
            return toCheck;
        } else {
            final String message = errorMessage.formatted(toCheck, lowerBoundary, upperBoundary);
            throw new IllegalArgumentException(message);
        }
    }

```

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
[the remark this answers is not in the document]
jsync-swirlds: FYI, I actually agree that "compression" is not ideal.
I wonder if "compressor" (used everywhere we currently have "compression" by itself) would be better.  I think it reads better, at least.

Not a major item; we can leave it as-is and think about possibly changing it later if you prefer that.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -43,7 +47,9 @@ public record PersistenceStorageConfig(
         @ConfigProperty(defaultValue = "") String liveRootPath,
         // @todo(#371) - the default life/archive root path must be absolute starting from /opt
         @ConfigProperty(defaultValue = "") String archiveRootPath,
-        @ConfigProperty(defaultValue = "BLOCK_AS_LOCAL_FILE") StorageType type) {
+        @ConfigProperty(defaultValue = "BLOCK_AS_LOCAL_FILE") StorageType type,
+        @ConfigProperty(defaultValue = "ZSTD") CompressionType compression,
+        @ConfigProperty(defaultValue = "3") int compressionLevel) {
```
jsync-swirlds, line 52: I believe we should set bounds for this config.
Also, while zstd allows for negative value, I believe we should constrain to strictly greater than or equal to `0`.
```suggestion
        @ConfigProperty(defaultValue = "3") @Min(0) @Max(20) int compressionLevel) {
```

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -56,11 +62,28 @@ public record PersistenceStorageConfig(
...
+     *
+     * @param compressionType the compression type
+     * @param compressionLevelToCheck the compression level to check
+     */
+    private void verifyCompressionLevel(final CompressionType compressionType, final int compressionLevelToCheck) {
+        switch (compressionType) {
+            case ZSTD -> Preconditions.requireInRange(
+                    compressionLevelToCheck, Zstd.minCompressionLevel(), Zstd.maxCompressionLevel());
+            case NONE -> Preconditions.requireInRange(compressionLevelToCheck, Integer.MIN_VALUE, Integer.MAX_VALUE);
+        }
```
jsync-swirlds, line 84: Instead of switch and case, why not just have the CompressionType interface expose the min/max compression level property, and then this method is much simpler:
```suggestion
        Preconditions.requireInRange(compressionLevelToCheck, compressionType.minCompressionLevel(), compressionType.maxCompressionLevel());
```

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -146,4 +169,20 @@ public enum StorageType {
...
+        /**
+         * This type of compression is used to compress the blocks using the
+         * `Zstandard` algorithm.
+         */
+        ZSTD,
+        /**
+         * This type means no compression will be done.
+         */
+        NONE
+    }
```
jsync-swirlds, line 187: To go along with the previous suggestion:
```suggestion
    public enum CompressionType {
        /**
         * This type of compression is used to compress the blocks using the
         * `Zstandard` algorithm.
         */
        ZSTD(0, 20),

        /**
         * This type means no compression will be done.
         */
        NONE(Integer.MIN_VALUE, Integer.MAX_VALUE);

        private final int min;
        private final int max;

        private CompressionType(final int minValue, final int maxValue) {
            min = minValue;
            max = maxValue;
        }

        public void verifyCompressionLevel(final int levelToCheck) {
            Preconditions.requireInRange(levelToCheck, min, max);
        }
    }
```

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -56,11 +62,28 @@ public record PersistenceStorageConfig(
      */
     public PersistenceStorageConfig {
         Objects.requireNonNull(type);
+        Objects.requireNonNull(compression);
+        verifyCompressionLevel(compression, compressionLevel);
```
jsync-swirlds, line 66: Simplification:
```suggestion
        verifyCompressionLevel(Objects.requireNonNull(compression), compressionLevel);
```
OR (see below)
```suggestion
        compression.verifyCompressionLevel(compressionLevel);
```

server/src/main/resources/app.properties
```diff
@@ -14,3 +14,4 @@ prometheus.endpointPortNumber=9999
 #persistence.storage.archiveRootPath=
 # @todo(#372) - default persistence type should be BLOCK_AS_LOCAL_FILE
 persistence.storage.type=BLOCK_AS_LOCAL_DIRECTORY
+#persistence.storage.compression=NONE
```
jsync-swirlds, line 17: Missing newline at EOF.
image

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
ata-nas: @jsync-swirlds that is indeed a good proposal. One thing though, having a pre-defined template to resolve is quite restrictive (we cannot provide anything outside the pre-defined template), so I think a better api and implementation would be the following:

```java
    public static int requireInRange(final int toCheck, final int lowerBoundary, final int upperBoundary) {
        return requireInRange(
                toCheck,
                lowerBoundary,
                upperBoundary,
                "The input value [%d] is required to be in the range [%d, %d], inclusive.",
                toCheck,
                lowerBoundary,
                upperBoundary);
    }

    public static int requireInRange(
            final int toCheck,
            final int lowerBoundary,
            final int upperBoundary,
            @NonNull final String errorMessage,
            final Object... errorMessageArgs) {
        if (toCheck >= lowerBoundary && toCheck <= upperBoundary) {
            return toCheck;
        } else {
            final String message = errorMessage.formatted(errorMessageArgs);
        }
    }
```

This gives us complete freedom over what the message would be, while we keep the lazy formatting for the message. I think this is the best way to go, since we can have any template we chose or no template at all. So let me know also on your opinion, but I think this is the way to go.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
ata-nas: ✅ Done.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
ata-nas: ✅ Done.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
ata-nas: ✅ Done.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
ata-nas: ✅ Done.

server/src/main/resources/app.properties
ata-nas: ✅ Done.

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
```diff
@@ -185,6 +185,50 @@ public static long requireGreaterOrEqual(final long toTest, final long base, fin
...
+     *
+     * @param toCheck the int value to test
+     * @param lowerBoundary the lower boundary
+     * @param upperBoundary the upper boundary
+     * @return the input {@code toCheck} if it is within the range (boundaries
+     * included)
+     * @throws IllegalArgumentException if the input int does not pass the test
+     */
+    public static int requireInRange(final int toCheck, final int lowerBoundary, final int upperBoundary) {
+        return requireInRange(toCheck, lowerBoundary, upperBoundary, null);
```
mattp-swirldslabs, line 201: I agree with the idea of replacing the `null` parameter with a default value.

mattp-swirldslabs (dismissed):
Looks good, thanks!

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
ata-nas: Will be addressed in a chore a bit later.

server/src/main/java/com/hedera/block/server/persistence/storage/compression/NoOpCompression.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * An implementation of {@link Compression} that does not compress the data. It
+ * does not use any algorithm, but simply generates a stream that writes the
+ * data to it`s destination, as it is received.
+ */
+public class NoOpCompression implements Compression {
+    /**
+     * Constructor.
+     */
+    private NoOpCompression() {}
```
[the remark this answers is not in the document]
ata-nas: If I am not mistaken, I get a build warning from gradle that I use a constructor without documentation...

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
jsync-swirlds: I don't expect to handle this in this issue; the comment was just an example for the future chore.

What I wrote doesn't set a single pre-defined template, it has a _default_ template and the option to replace that template with anything the caller prefers.  The only pre-defined element is what values are filled in.  Given that the only thing a caller cannot pre-populate before calling is the parameters; what you propose is overkill (not to mention _more_ costly as you're passing the same parameters multiple times and wrapping primitives in most cases).

It's vastly simpler, and no less flexible, to stick with the approach I suggested.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -56,6 +63,7 @@ public record PersistenceStorageConfig(
      */
     public PersistenceStorageConfig {
         Objects.requireNonNull(type);
+        Objects.requireNonNull(compression).verifyCompressionLevel(compressionLevel);
```
jsync-swirlds, line 66: We do not need the first requireNonNull; the call will throw without it.
```suggestion
        compression.verifyCompressionLevel(compressionLevel);
```
General principle: Use `requireNonNull` if we would not otherwise throw a NPE within a constructor, or within the first few lines of any other method.

Letting the "organic" NPE throw is preferable because the JVM will usually fill in better details than `Objects.requireNonNull` will.

ata-nas: ✅ Done.

mattp-swirldslabs (approved):

PR #389, opened 2024-12-10
feat: Tool for converting Record Files to Block Stream

### LINKED ISSUE #359: Create a tool to convert existing record files to block files

jasperpotts:
Command line tool adding extra command for converting Record Files to Block Stream.
Resolves #359


jasperpotts:
Not 100% done yet but made ready for review so the build actions are executed to make sure build works on sever because it as been such a pain to get to build locally.

settings.gradle.kts
```diff
@@ -68,6 +68,9 @@ dependencyResolutionManagement {
             version("com.google.auto.service", "1.1.1")
             version("org.hyperledger.besu.nativelib.secp256k1", "0.8.2")
             version("info.picocli", "4.7.6")
+            version("com.github.luben.zstd_jni", "1.5.6-6")
```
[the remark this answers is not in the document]
ata-nas: @AlfredoG87 yep.
@jasperpotts FYI, there is a newer version (I suppose we should use that, unless there is a reason to use this one specifically)
`version("com.github.luben.zstd_jni","1.5.6-8")`

tools/src/main/java/com/hedera/block/tools/commands/record2blocks/model/SignatureFile.java
```diff
@@ -0,0 +1,108 @@
...
+ * <td>byte[]</td>
+ * <td>Signature bytes</td>
+ * </tr>
+ * </tbody>
+ * </table>
+ *
+ * @param fileHash SHA384 hash of corresponding *.rcd file
+ * @param signature Signature bytes or RSA signature of the file hash, signed by the node's private key
+ */
+public record SignatureFile(byte[] fileHash, byte[] signature) {
```
jsync-swirlds, line 66: Xin brought up a good point in the HIP review.
We probably need to track _which_ node this signature is for, and include that in the bytes stored in the record file item.  Otherwise we'll have no way to verify signatures on those items in the block stream because we won't know which node's public key to use.

Also, we'll need some way to populate what address book to use (unless we have some other source available for that).

settings.gradle.kts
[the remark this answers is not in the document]
jasperpotts: Fixed guess newer than when I started project

tools/README.md
```diff
@@ -58,6 +67,119 @@ Prints info for block files
...
+- `--block-times=<blockTimesFile>`
+   - Path to the block times ".bin" file.
+   - Default: "data/block_times.bin"
+
+### The `fetchRecordsCsv` Subcommand
+
+Download mirror node record table CSV dump from GCP bucket. The records table on mirror node has a row for every block 
+mirror node knows about. The CSV file is huge 11GB+ in November 2024. This data is important for records to blocks 
+conversion as we have to make sure the block number assigned for a record file matches what mirror node says as the 
+source of truth.
```
jsync-swirlds, line 115: Question here, are we saying that the block number in mirror node doesn't match the block number in the record file?

tools/build.gradle.kts
```diff
@@ -17,19 +19,58 @@
...
+    }
+}
+
+// Switch compilation from modules back to classpath because 3rd party libraries are not modularized
+tasks.compileJava {
+    // Do not use '--module-path' despite `module-info.java`
+    modularity.inferModulePath = false
+    // Do not compile module-info, we use it only to extract dependencies for now
+    exclude("module-info.java")
+}
```
jjohannes, line 48:
```suggestion
```

As you removed moule-info.java completely (which is good for consistency) you do not need this block. Gradle automatically goes into "classpath mode" if the file does not exist.

tools/build.gradle.kts
```diff
@@ -17,19 +19,58 @@
...
 
 description = "Hedera Block Stream Tools"
 
 application {
-    mainModule = "com.hedera.block.tools"
     mainClass = "com.hedera.block.tools.BlockStreamTool"
 }
 
+// Generate Manifest with Main-Class and Implementation-Title
+tasks.withType<Jar> {
```
jjohannes, line 32:
```suggestion
tasks.withType<Jar>().configureEach {
```

Minor tweak to make Gradel do things more lazily
(https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#eager_apis_to_avoid)

tools/build.gradle.kts
```diff
@@ -17,19 +19,58 @@
...
+// Allow non-module Jar
+extraJavaModuleInfo {
+    failOnMissingModuleInfo = false
+    failOnAutomaticModules = false
+}
+
+// Disable module directives scope check as we are not using modules
+tasks.withType<ModuleDirectivesScopeCheck>().configureEach {
+    enabled = false
+}
```
jjohannes, line 59: Note: after updating to the new Gradle configuration (#216) I will take another look at how we could improve this to have the check configured without modules.

settings.gradle.kts
```diff
@@ -68,6 +68,9 @@ dependencyResolutionManagement {
             version("com.google.auto.service", "1.1.1")
             version("org.hyperledger.besu.nativelib.secp256k1", "0.8.2")
             version("info.picocli", "4.7.6")
+            version("com.github.luben.zstd_jni", "1.5.6-8")
+            version("com.google.cloud.core", "2.48.0")
+            version("com.google.cloud.storage", "2.45.0")
```
jjohannes, line 73:
```suggestion
```

AS you have the versions in `build.gradle.kts` directly, you do not need them here.

Note: with #216, we can move these versions together witth the others to `hiero-dependency-versions/build.grlad.kts`. I will do that then.

jjohannes (dismissed):
Review for the Gradle configuration part

jjohannes:
Comment on Gradle setup:
I will revisit the "classpath-based" setup of the "tools" project this PR introduces as part of https://github.com/hiero-ledger/hiero-gradle-conventions/issues/53

settings.gradle.kts
ata-nas: Note, `version("com.github.luben.zstd_jni", "1.5.6-8")` is being used elsewhere as well, it should remain here I think.

ata-nas (approved):
Code looks good. Some comments from @jjohannes about gradle.

tools/README.md
jasperpotts: Older record files did not have block numbers in them. They were only added maybe 18 mths ago. For a long time block numbers were assigned on mirror node and only existed there.

ata-nas (approved):
🔥

jsync-swirlds (approved):

PR #406, opened 2024-12-13
chore: Fix Simulator unbounded memory growth

### LINKED ISSUE #404: feat: Update Consumer Simulator to be more memory efficient
As a Consumer Simulator User
I want to be able to consume the 10k TPS recording of real data with the setting `blockStream.millisecondsPerBlock=500` without running out of memory in the simulator.
So that I can confirm end-to-end BN streaming.

* While testing [PBJ Issue #329](https://github.com/hashgraph/pbj/issues/329), I found that the Consumer Simulator would run out of memory (16GB) before consuming all 4058 blocks of the 10k TPS recording
* On a local branch, I was able to keep the Consumer Simulator at around 1 GB of memory while consuming the entire 4058 blocks of the recording to test [#329](https://github.com/hashgraph/pbj/issues/329). I was able to fix the memory issue by using the "Unparsed" BlockItem strategy we're using on the BN server. However, after a stand up discussion, we determined that we should NOT use the "Unparsed" strategy for the Consumer Simulator.
* Instead of "Unparsed" types, we need to find ways to conserve memory without changing the existing `protoc` types if possible.

## Tech Notes

* Getting to the point where you can see the OutOfMemoryError on the Consumer Simulator may require the Helidon fixes coming in `4.2.0-SNAPSHOT`. Here's how you can get these HTTP/2 fixes to run the 10k TPS recording successfully from the producer -> BN -> consumer:

0) Prerequisite: install Maven. I used version 3.8.7
1) On a terminal cd to a directory to clone the Helidon repo
2) `git clone git@github.com:helidon-io/helidon.git && cd helidon && mvn install -DskipTests`
3) In `settings.gradle.kts` change `val helidonVersion = "4.1.1"` to `val helidonVersion = "4.2.0-SNAPSHOT"
4) Build and deploy the BN container as you usually would

georgi-l95:
**Description**:
This PR aims to resolve unbounded memory growth within the simulator, when running in consumer mode. It's was caused by the uncontrollable grow of the lastKnownStatuses list, when running in "high-speed" mode.
- Changed to ArrayDeque with predefined capacity.
- Added new blockStreamConfig called lastKnownStatusesCapacity, set to be 10 by default

**Related issue(s)**:

Fixes #404 

**Notes for reviewer**:
Here is image of the heap size before the fix:
image
Here is image of the heap size after the fix:
image

**Checklist**

### DIFF
--- a/simulator/src/main/java/com/hedera/block/simulator/config/data/BlockStreamConfig.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/config/data/BlockStreamConfig.java
@@ -34,6 +34,7 @@ import com.swirlds.config.api.ConfigProperty;
 @ConfigData("blockStream")
 public record BlockStreamConfig(
         @ConfigProperty(defaultValue = "PUBLISHER") SimulatorMode simulatorMode,
+        @ConfigProperty(defaultValue = "10") int lastKnownStatusesCapacity,
         @ConfigProperty(defaultValue = "1_500_000") int delayBetweenBlockItems,
         @ConfigProperty(defaultValue = "100_000") int maxBlockItemsToStream,
         @ConfigProperty(defaultValue = "MILLIS_PER_BLOCK") StreamingMode streamingMode,
@@ -54,6 +55,7 @@ public record BlockStreamConfig(
      */
     public static class Builder {
         private SimulatorMode simulatorMode = SimulatorMode.PUBLISHER;
+        private int lastKnownStatusesCapacity = 10;
         private int delayBetweenBlockItems = 1_500_000;
         private int maxBlockItemsToStream = 10_000;
         private StreamingMode streamingMode = StreamingMode.MILLIS_PER_BLOCK;
@@ -75,6 +77,17 @@ public record BlockStreamConfig(
          */
         public Builder simulatorMode(SimulatorMode simulatorMode) {
             this.simulatorMode = simulatorMode;
+            return this;
+        }
+
+        /**
+         * Sets the capacity of the last known statuses.
+         *
+         * @param lastKnownStatusesCapacity the capacity
+         * @return this {@code Builder} instance
+         */
+        public Builder lastKnownStatusesCapacity(int lastKnownStatusesCapacity) {
+            this.lastKnownStatusesCapacity = lastKnownStatusesCapacity;
             return this;
         }
 
@@ -141,6 +154,7 @@ public record BlockStreamConfig(
         public BlockStreamConfig build() {
             return new BlockStreamConfig(
                     simulatorMode,
+                    lastKnownStatusesCapacity,
                     delayBetweenBlockItems,
                     maxBlockItemsToStream,
                     streamingMode,
--- a/simulator/src/main/java/com/hedera/block/simulator/config/data/BlockStreamConfig.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/config/data/BlockStreamConfig.java
@@ -25,6 +25,7 @@ import com.swirlds.config.api.ConfigProperty;
  * Defines the configuration data for the block stream in the Hedera Block Simulator.
  *
  * @param simulatorMode the mode of the simulator, in terms of publishing, consuming or both
+ * @param lastKnownStatusesCapacity the capacity of the last known statuses
  * @param delayBetweenBlockItems the delay in microseconds between streaming each block item
  * @param maxBlockItemsToStream the maximum number of block items to stream before stopping
  * @param streamingMode the mode of streaming for the block stream (e.g., time-based, count-based)
@@ -38,7 +39,7 @@ public record BlockStreamConfig(
         @ConfigProperty(defaultValue = "1_500_000") int delayBetweenBlockItems,
         @ConfigProperty(defaultValue = "100_000") int maxBlockItemsToStream,
         @ConfigProperty(defaultValue = "MILLIS_PER_BLOCK") StreamingMode streamingMode,
-        @ConfigProperty(defaultValue = "10") int millisecondsPerBlock,
+        @ConfigProperty(defaultValue = "1000") int millisecondsPerBlock,
         @ConfigProperty(defaultValue = "1000") int blockItemsBatchSize) {
 
     /**
--- a/simulator/src/main/java/com/hedera/block/simulator/config/data/BlockStreamConfig.java
+++ b/simulator/src/main/java/com/hedera/block/simulator/config/data/BlockStreamConfig.java
@@ -38,7 +38,7 @@ public record BlockStreamConfig(
         @ConfigProperty(defaultValue = "1_500_000") int delayBetweenBlockItems,
         @ConfigProperty(defaultValue = "100_000") int maxBlockItemsToStream,
         @ConfigProperty(defaultValue = "MILLIS_PER_BLOCK") StreamingMode streamingMode,
-        @ConfigProperty(defaultValue = "1000") int millisecondsPerBlock,
+        @ConfigProperty(defaultValue = "10") int millisecondsPerBlock,
         @ConfigProperty(defaultValue = "1000") int blockItemsBatchSize) {
 
     /**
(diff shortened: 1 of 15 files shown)


simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamGrpcClientImpl.java
```diff
@@ -53,7 +55,8 @@ public class ConsumerStreamGrpcClientImpl implements ConsumerStreamGrpcClient {
     private StreamObserver<SubscribeStreamResponse> consumerStreamObserver;
 
     // State
-    private final List<String> lastKnownStatuses;
+    private final int lastKnownStatusesCapacity;
+    private final ArrayDeque<String> lastKnownStatuses;
```
mattp-swirldslabs, line 59: `ArrayDeque<String>` => `Deque<String>` here ^^

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -70,6 +74,9 @@ public ConsumerStreamObserver(
     @Override
     public void onNext(SubscribeStreamResponse subscribeStreamResponse) {
         final SubscribeStreamResponse.ResponseCase responseType = subscribeStreamResponse.getResponseCase();
+        if (lastKnownStatuses.size() == lastKnownStatusesCapacity) {
+            lastKnownStatuses.removeFirst();
```
mattp-swirldslabs, line 78: The javadoc for `removeFirst()` indicates that it will throw an exception if the Deque is empty. `pollFirst()` has the same behavior but it will just return `null` instead of throwing an exception. I'm not sure of the behavior we want for the simulator. What do you think?
image

mattp-swirldslabs (approved):
Looks good. I added a few comments and questions. Also, I'm seeing a few javadoc gaps when I build it:
image

simulator/src/main/java/com/hedera/block/simulator/config/data/BlockStreamConfig.java
```diff
@@ -34,10 +34,11 @@
 @ConfigData("blockStream")
 public record BlockStreamConfig(
         @ConfigProperty(defaultValue = "PUBLISHER") SimulatorMode simulatorMode,
+        @ConfigProperty(defaultValue = "10") int lastKnownStatusesCapacity,
```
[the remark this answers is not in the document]
georgi-l95: I added it like this for two reasons, tell me what you think:
1. One central place to control it for both clients.
2. We can control it via env. variable from docker or something similar.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -55,10 +57,12 @@ public class ConsumerStreamObserver implements StreamObserver<SubscribeStreamRes
     public ConsumerStreamObserver(
             @NonNull final MetricsService metricsService,
             @NonNull final CountDownLatch streamLatch,
-            @NonNull final List<String> lastKnownStatuses) {
+            @NonNull final ArrayDeque<String> lastKnownStatuses,
+            @NonNull final int lastKnownStatusesCapacity) {
```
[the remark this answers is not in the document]
georgi-l95: Ah, yes thank you, removed.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamObserver.java
```diff
@@ -46,9 +47,12 @@ public class PublishStreamObserver implements StreamObserver<PublishStreamRespon
      * @throws NullPointerException if any parameter is null
      */
     public PublishStreamObserver(
-            @NonNull final AtomicBoolean streamEnabled, @NonNull final List<String> lastKnownStatuses) {
+            @NonNull final AtomicBoolean streamEnabled,
+            @NonNull final ArrayDeque<String> lastKnownStatuses,
+            @NonNull final int lastKnownStatusesCapacity) {
```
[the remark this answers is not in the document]
georgi-l95: Ah, yes thank you, removed.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamObserver.java
```diff
@@ -46,9 +47,12 @@ public class PublishStreamObserver implements StreamObserver<PublishStreamRespon
      * @throws NullPointerException if any parameter is null
      */
     public PublishStreamObserver(
-            @NonNull final AtomicBoolean streamEnabled, @NonNull final List<String> lastKnownStatuses) {
+            @NonNull final AtomicBoolean streamEnabled,
+            @NonNull final ArrayDeque<String> lastKnownStatuses,
```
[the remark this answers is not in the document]
georgi-l95: You're right, changed.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -55,10 +57,12 @@ public class ConsumerStreamObserver implements StreamObserver<SubscribeStreamRes
     public ConsumerStreamObserver(
             @NonNull final MetricsService metricsService,
             @NonNull final CountDownLatch streamLatch,
-            @NonNull final List<String> lastKnownStatuses) {
+            @NonNull final ArrayDeque<String> lastKnownStatuses,
```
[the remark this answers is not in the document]
georgi-l95: You're right, changed.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamGrpcClientImpl.java
georgi-l95: Changed 👍

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -70,6 +75,9 @@ public ConsumerStreamObserver(
     @Override
     public void onNext(SubscribeStreamResponse subscribeStreamResponse) {
         final SubscribeStreamResponse.ResponseCase responseType = subscribeStreamResponse.getResponseCase();
+        if (lastKnownStatuses.size() == lastKnownStatusesCapacity) {
```
ata-nas, line 78: I think better do >= on line 78, since it is way less error prone. There could be a case where we could go above the capacity before we make this check, which means we will never enter the `if`. I do not think we need to be _exact_ here for the simulator, but rather clean the collection as long as we are overboard. Also, that being said, if we are ever above the capacity, maybe we should poll the difference, as big as it is, in this way we ensure we always keep proper capacity.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamObserver.java
```diff
@@ -58,6 +63,9 @@ public PublishStreamObserver(
      */
     @Override
     public void onNext(PublishStreamResponse publishStreamResponse) {
+        if (lastKnownStatuses.size() == lastKnownStatusesCapacity) {
```
ata-nas, line 66: Same as in the comment in the other observer.

ata-nas (approved):
Looks good @georgi-l95, just a couple of nits to consider.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamObserver.java
georgi-l95: Very good point, thanks 👍

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/ConsumerStreamObserver.java
georgi-l95: Very good point, thanks 👍

ata-nas (approved):
LGTM

PR #415, opened 2024-12-17
chore: migrate preconditions to use non-null error messages

### LINKED ISSUE #381: chore: migrate preconditions to use non-null error messages
## Description

Migrate the preconditions class to use non-null error messages as seen in [this comment](https://github.com/hashgraph/hedera-block-node/pull/387#discussion_r1884561268)

e.g.
```java
    private static final String OUT_OF_RANGE_DEFAULT_MESSAGE =
            "The input value [%d] is required to be in the range [%d, %d], inclusive.";
    /**
     * This method asserts a given int is within a range (boundaries
     * included). If the given int is within the range, then we return it,
     * else, an {@link IllegalArgumentException} is thrown.
     *
     * @param toCheck the int value to test
     * @param lowerBoundary the lower boundary
     * @param upperBoundary the upper boundary
     * @return the input {@code toCheck} if it is within the range (boundaries
     * included)
     * @throws IllegalArgumentException if the input int does not pass the test
     */
    public static int requireInRange(final int toCheck, final int lowerBoundary, final int upperBoundary) {
        return requireInRange(toCheck, lowerBoundary, upperBoundary, OUT_OF_RANGE_DEFAULT_MESSAGE);
    }
    /**
     * This method asserts a given int is within a range (boundaries
     * included). If the given int is within the range, then we return it,
     * else, an {@link IllegalArgumentException} is thrown.
     *
     * @param toCheck the int value to check
     * @param lowerBoundary the lower boundary
     * @param upperBoundary the upper boundary
     * @param errorMessage A format string {@see Formatter#syntax} with three
     *     decimal parameters for {@code toCheck}, {@code lowerBoundary},
     *     and {@code upperBoundary}.<br/>
     *     Example: {@code "The input (%d) is out of range (%d, %d)"}
     * @return the input {@code toCheck} if it is within the range (boundaries
     * included)
     * @throws IllegalArgumentException if the input int does not pass the test
     */
    public static int requireInRange(
            final int toCheck, final int lowerBoundary, final int upperBoundary, @NonNull final String errorMessage) {
        if (toCheck >= lowerBoundary && toCheck <= upperBoundary) {
            return toCheck;
        } else {
            final String message = errorMessage.formatted(toCheck, lowerBoundary, upperBoundary);
        }
    }
```

ata-nas:
**Description**

- migrate the preconditions class to use non-null error messages as seen in this comment

**Related issue(s)**:

Fixes #381 

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/common/src/main/java/com/hedera/block/common/utils/Preconditions.java
+++ b/common/src/main/java/com/hedera/block/common/utils/Preconditions.java
@@ -16,10 +16,14 @@
 
 package com.hedera.block.common.utils;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Objects;
 
 /** A utility class used to assert various preconditions. */
-public final class Preconditions { // @todo(381) change the APIs to accept non-null error messages
+public final class Preconditions {
+    private static final String DEFAULT_NOT_BLANK_MESSAGE = "The input String is required to be non-blank.";
+    private static final String DEFAULT_REQUIRED_POSITIVE_MESSAGE = "The input number [%d] is required to be positive.";
+
     /**
      * This method asserts a given {@link String} is not blank, meaning it is
      * not {@code null} or does not contain only whitespaces as defined by
@@ -27,14 +31,14 @@ public final class Preconditions { // @todo(381) change the APIs to accept non-n
      * we return it, else we throw {@link IllegalArgumentException}.
      *
      * @param toCheck a {@link String} to be checked if is blank as defined
-     *                above
+     * above
      * @return the {@link String} to be checked if it is not blank as defined
-     *         above
+     * above
      * @throws IllegalArgumentException if the input {@link String} to be
-     *                                  checked is blank
+     * checked is blank
      */
     public static String requireNotBlank(final String toCheck) {
-        return requireNotBlank(toCheck, null);
+        return requireNotBlank(toCheck, DEFAULT_NOT_BLANK_MESSAGE);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


georgi-l95 (dismissed):

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
```diff
@@ -16,25 +16,36 @@
...
      *
      * @param toCheck a {@link String} to be checked if is blank as defined
-     *                above
+     * above
      * @return the {@link String} to be checked if it is not blank as defined
-     *         above
+     * above
      * @throws IllegalArgumentException if the input {@link String} to be
-     *                                  checked is blank
+     * checked is blank
```
jsync-swirlds, line 45: A bit of a nitpick, but javadoc should follow 4-space indentation, and it's OK to extend lines a little if it avoids orphans.
```suggestion
     * @param toCheck a {@link String} to be checked if is blank as defined above.
     * @return the {@link String} to be checked if it is not blank as defined above.
     * @throws IllegalArgumentException if the input {@link String} to be
     *     checked is blank.
```

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
```diff
@@ -16,25 +16,36 @@
...
+    private static final String DEFAULT_REQUIRE_WHOLE_MESSAGE =
+            "The input number [%d] is required to be a whole number.";
+    private static final String DEFAULT_REQUIRE_POWER_OF_TWO_MESSAGE =
+            "The input number [%d] is required to be a power of two.";
+
     /**
      * This method asserts a given {@link String} is not blank, meaning it is
      * not {@code null} or does not contain only whitespaces as defined by
      * {@link String#isBlank()}. If the given {@link String} is not blank, then
      * we return it, else we throw {@link IllegalArgumentException}.
```
jsync-swirlds, line 38: Another slight nitpick.  Javadoc treats the first "sentence" specially (defined as everything up to a `.` followed by a newline), and that should be a short "summary" line for a method.
```suggestion
     * This method asserts a given {@link String} is not blank.
     * A blank string is {@code null} or contains only whitespaces as defined by
     * {@link String#isBlank()}.<br/>If the given {@link String} is not blank,
     * then we return it, otherwise we throw {@link IllegalArgumentException}.
```

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
```diff
@@ -107,39 +112,36 @@ public static int requirePositive(final int toCheck, final String errorMessage)
...
-     * @param toCheck      the long to check if it is a positive power of two
-     * @param errorMessage the error message to be used in the exception if the
-     *                     input long to check is not positive, if null, a default
-     *                     message will
-     *                     be used
-     * @return the long to check if it is positive
+     * @param toCheck the long to check if it is a positive power of two
+     * @param errorMessage a formatted string with one decimal parameters for
+     * {@code toCheck}, must not be {@code null}.<br/>
+     * Example error message: {@code "The input number (%d) must be positive."}
```
jsync-swirlds, line 128: These example items might be easier if you use `@value` tags.
```suggestion
     * Example error message: {@value #DEFAULT_REQUIRE_POSITIVE_MESSAGE}
```

jsync-swirlds (approved):
Looks good, a few opportunities to make Javadoc a little cleaner, which could be done in a future PR.

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
ata-nas: ✅ Indeed, it makes the javadoc more readable in raw format. I've made this improvement.

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
ata-nas: ✅ Thanks for that, I've made this improvement.

common/src/main/java/com/hedera/block/common/utils/Preconditions.java
ata-nas: ✅ Thanks for that, indeed this makes the javadoc better and if ever the default message changes, the javadoc will reflect that. I've made this improvement as well!

georgi-l95 (approved):

jsync-swirlds (approved):

PR #416, opened 2024-12-17
feat: add docker configuration for the simulator

### LINKED ISSUE #412: Add Docker configuration for the Simulator
As a block stream simulator operator
I want to be able to start in both modes using only one task
To be able start in a isolated enviorment easy and fast.

georgi-l95:
**Description**:

This PR simplifies the simulator development process by introducing Docker containerization and improves documentation clarity.

- Add Docker setup for simulator components
- Add new Gradle task startSimulatorDockerContainers
- Reorganize configuration documentation with consistent table format
- Simplify metrics viewing process in quickstart guide

The changes streamline the development process by replacing manual configuration steps with automated Docker container setup, making it easier for users to run and monitor the simulator in both Publisher and Consumer modes.

**Related issue(s)**:

Fixes #412 

**Notes for reviewer**:

**Checklist**


simulator/build.gradle.kts
```diff
@@ -91,3 +91,29 @@ tasks.register<Copy>("untarTestBlockStream") {
 tasks.named("processResources") { dependsOn(tasks.named("untarTestBlockStream")) }
 
 tasks.named("sourcesJar") { dependsOn(tasks.named("untarTestBlockStream")) }
+
+// Task to prepare and run Docker container
+tasks.register("startSimulatorDockerContainers") {
+    description = "Prepare and run the simulator containers in publisher and consumer mode"
+    group = "docker"
+
+    doLast {
```
ata-nas, line 100: why is there `doLast` here?

simulator/build.gradle.kts
```diff
@@ -91,3 +91,29 @@ tasks.register<Copy>("untarTestBlockStream") {
 tasks.named("processResources") { dependsOn(tasks.named("untarTestBlockStream")) }
 
 tasks.named("sourcesJar") { dependsOn(tasks.named("untarTestBlockStream")) }
+
+// Task to prepare and run Docker container
+tasks.register("startSimulatorDockerContainers") {
```
ata-nas, line 96: The name of the task should be matching with the one in the server `startDockerContainer` as is per the [agreed upon convention](https://github.com/hashgraph/hedera-block-node/discussions/264#discussioncomment-11272142).

ata-nas (changes requested):
I believe changes must be made to match the [agreed upon team convention](https://github.com/hashgraph/hedera-block-node/discussions/264#discussioncomment-11272142) for making gradle tasks.

simulator/build.gradle.kts
georgi-l95: I want to make sure that those steps are the last thing we do in the whole lifecycle. more info here https://docs.gradle.org/current/userguide/build_lifecycle.html

simulator/build.gradle.kts
```diff
@@ -91,3 +91,29 @@ tasks.register<Copy>("untarTestBlockStream") {
...
+        exec { commandLine("./docker/prepare-docker.sh") }
+
+        // Change to docker directory and run docker-compose
+        exec {
+            workingDir("docker")
+            commandLine("sh", "-c", "docker compose -p simulator up --build -d")
+        }
+
+        // Run cleanup script
+        exec { commandLine("./docker/cleanup-docker.sh") }
```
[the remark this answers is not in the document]
georgi-l95: Why ? I don't want to manually delete some files everytime.

server/docs/configuration.md
```diff
@@ -1,22 +1,78 @@
...
-| MEDIATOR_RING_BUFFER_SIZE | Size of the ring buffer used by the mediator (must be a power of 2) | 67108864 |
-| NOTIFIER_RING_BUFFER_SIZE | Size of the ring buffer used by the notifier (must be a power of 2) | 2048 |
-| SERVER_PORT | The port the server will listen on | 8080 |
-| SERVER_MAX_MESSAGE_SIZE_BYTES | The maximum size of a message frame in bytes | 1048576 |
+## ConsumerConfig
+
+Uses the prefix `consumer` so all properties should start with `consumer.`
+
+| Key                      | Environment Variable                | Description                                                       | Default Value |
+| :----------------------- | :---------------------------------- | :---------------------------------------------------------------- | ------------: |
```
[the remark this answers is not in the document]
georgi-l95: I disagree here; maintenance will be the same. However, it looks far more organized now.

simulator/build.gradle.kts
```diff
@@ -91,3 +91,29 @@ tasks.register<Copy>("untarTestBlockStream") {
...
 tasks.named("sourcesJar") { dependsOn(tasks.named("untarTestBlockStream")) }
+
+// Task to prepare and run Docker container
+tasks.register("startSimulatorDockerContainers") {
+    description = "Prepare and run the simulator containers in publisher and consumer mode"
+    group = "docker"
+
+    doLast {
+        // Make the prepare and cleanup scripts executable
+        exec { commandLine("chmod", "+x", "docker/prepare-docker.sh", "docker/cleanup-docker.sh") }
```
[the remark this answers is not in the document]
georgi-l95: Yes, thanks, leftover that I forgot to delete.

simulator/docker/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+version: "3.8"
```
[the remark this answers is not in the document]
georgi-l95: True, will remove.

simulator/docker/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      context: .
+      dockerfile: Dockerfile
+    image: hedera-block-simulator:latest
+    networks:
+      - block-node_default
+    environment:
+      - GRPC_SERVER_ADDRESS=block-node-server
+      - BLOCK_STREAM_SIMULATOR_MODE=PUBLISHER
+      - PROMETHEUS_ENDPOINT_ENABLED=true
+      - PROMETHEUS_ENDPOINT_PORT_NUMBER=9998
```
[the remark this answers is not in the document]
georgi-l95: Make sense, will move them in a .env.

simulator/docker/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+      context: .
+      dockerfile: Dockerfile
+    image: hedera-block-simulator:latest
+    networks:
+      - block-node_default
+    environment:
+      - GRPC_SERVER_ADDRESS=block-node-server
+      - BLOCK_STREAM_SIMULATOR_MODE=CONSUMER
+      - PROMETHEUS_ENDPOINT_ENABLED=true
+      - PROMETHEUS_ENDPOINT_PORT_NUMBER=9997
```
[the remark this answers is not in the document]
georgi-l95: Make sense, will move them in a .env.

simulator/docs/configuration.md
```diff
@@ -9,32 +10,43 @@ There are 3 configuration sets:
...
-| `delayBetweenBlockItems` | The delay between each block item in nanoseconds, only applicable when streamingMode=CONSTANT_RATE | `1_500_000` |
-| `maxBlockItemsToStream` | exit condition for the simulator and the circular implementations such as `BlockAsDir` or `BlockAsFile` implementations | `10_000` |
-| `streamingMode` | can either be `CONSTANT_RATE` or `MILLIS_PER_BLOCK` | `CONSTANT_RATE` |
-| `millisecondsPerBlock` | if streamingMode is `MILLIS_PER_BLOCK` this will be the time to wait between blocks in milliseconds | `1_000` |
-| `blockItemsBatchSize` | the number of block items to send in a single batch, however if a block has less block items, it will send all the items in a block | `1_000` |
+| Key                      | Environment Variable                     | Description                                                                                                                         |   Default Value |
+| :----------------------- | :--------------------------------------- | :---------------------------------------------------------------------------------------------------------------------------------- | --------------: |
```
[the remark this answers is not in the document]
georgi-l95: I don't see how this will be harder to support. But I see how it's adding to readability.

simulator/build.gradle.kts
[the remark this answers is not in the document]
ata-nas: If we never leave the volatile build root and we generate everything that is temporary and needs to be scrapped there, then we really do not need to delete any files, as the build root is not under the management of the source control and with the next run of gradle's clean task, all files will be scrapped.

simulator/build.gradle.kts
georgi-l95: Yes, completly forgot about it, will change it to match 👍

simulator/build.gradle.kts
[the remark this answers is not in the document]
georgi-l95: Yeah, but i'm currently using them from docker folder, will modify to try to take them without copying.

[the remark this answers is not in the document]
ata-nas: The point here is to copy the docker folder into the build root of the subproject. The docker folder must contain files which when used with the build root as a working directory, they will not need to go out of it to supply themselves with anything that is external to the build root. In this way we ensure that the builds are only restricted to the build root (which should be the case).

server/docs/configuration.md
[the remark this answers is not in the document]
jsync-swirlds: Probably something we should discuss as a team after the break.
It's a _lot_ easier to manage tables without the extra alignment, and the output is identical.  It's not as "pretty" when viewed as raw text, however.

Ideally we'll come to agreement in the team and have a _consistent_ approach from there.

simulator/build.gradle.kts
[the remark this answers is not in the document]
jsync-swirlds: +1 @ata-nas The build is creating/copying files already, so it's entirely reasonable to copy docker definitions into the build folder and write any changes there only, so that the build folder is the only thing modified during a build.

This makes system audits easier and also makes it easier to create repeatable builds because we always start from a clean base.

simulator/docs/configuration.md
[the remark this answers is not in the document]
jsync-swirlds: As above, we should discuss as a team.  Let's get a single agreed consistent pattern and all follow that so we don't have this churning back and forth.

jsync-swirlds (commented):
Just a couple minor (non-blocking) items.

server/docker/metrics/prometheus.yml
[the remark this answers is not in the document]
georgi-l95: Good question, this I think can be done, but might require some experimentation, I was thinking of something similar, because it bugging me out that I have simulator stuff in the server.
Prometheus support file based discovery, where I should be able to place dynamically configs, about scrape endpoints. And reload configuration in Prometheus simply by calling endpoint. That's on theory, will open an issue and try it out, when this is merged. 👍

simulator/docker/prepare-docker.sh
```diff
@@ -0,0 +1,32 @@
...
+# Create docker directory if it doesn't exist
+mkdir -p "$SCRIPT_DIR"
+
+# Copy simulator distribution
+echo "📋 Copying simulator distribution..."
+cp simulator/build/distributions/simulator-*.tar "$SCRIPT_DIR/"
+
+# Copy block data
+echo "📋 Copying block data..."
+cp simulator/src/main/resources/block-0.0.3.tar.gz "$SCRIPT_DIR/"
```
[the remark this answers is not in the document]
georgi-l95: Agree with you, will open an issue and we can think of an approach.

server/docs/configuration.md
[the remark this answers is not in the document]
georgi-l95: Okay, reverting, we can discuss, when the time permits, it's not high priority at all.

simulator/docs/configuration.md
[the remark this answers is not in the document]
georgi-l95: Okay, reverting, we can discuss, when the time permits, it's not high priority at all.

simulator/build.gradle.kts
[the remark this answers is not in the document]
georgi-l95: Changed to copy and work in build root.

simulator/docker/prepare-docker.sh
```diff
@@ -0,0 +1,32 @@
...
+echo "🚀 Preparing Docker build environment..."
+
+# Get the script's directory
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+PROJECT_ROOT="$SCRIPT_DIR/../.."
+
+# Build new distribution
+echo "📦 Building simulator distribution..."
+cd "$PROJECT_ROOT"
+./gradlew :simulator:assemble
```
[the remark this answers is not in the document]
georgi-l95: Removed from there. 👍

simulator/docker/prepare-docker.sh
```diff
@@ -7,26 +7,27 @@ echo "🚀 Preparing Docker build environment..."
...
 
 # Copy block data
 echo "📋 Copying block data..."
-cp simulator/src/main/resources/block-0.0.3.tar.gz "$SCRIPT_DIR/"
+cp ../../src/main/resources/block-0.0.3.tar.gz "$SCRIPT_DIR/"
 
 # Copy logging properties
 echo "📋 Copying logging properties..."
-cp simulator/src/main/resources/logging.properties "$SCRIPT_DIR/logging.properties"
+cp ../../src/main/resources/logging.properties "$SCRIPT_DIR/logging.properties"
```
ata-nas, line 25: Let's move these copies to be done by the `copyDockerFolder` task, which could then be renamed to `copyExternalBuildDependencies` or similar, since it would no longer copy the docker folder only. In this way, we will also not need to do the logic before these `cp`s where we are resolving and making dirs.

simulator/docker/prepare-docker.sh
```diff
@@ -7,26 +7,27 @@ echo "🚀 Preparing Docker build environment..."
...
+
+sh -c ./update-env.sh
 
 echo "✅ Docker build environment prepared successfully!"
+
+docker compose -p simulator up --build -d
```
ata-nas, line 31: IMO, we better execute compose from the gradle task.

simulator/docker/prepare-docker.sh
```diff
@@ -7,26 +7,27 @@ echo "🚀 Preparing Docker build environment..."
 
 # Get the script's directory
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-PROJECT_ROOT="$SCRIPT_DIR/../.."
-
-# Build new distribution
-echo "📦 Building simulator distribution..."
-cd "$PROJECT_ROOT"
-./gradlew :simulator:assemble
+PROJECT_ROOT="$SCRIPT_DIR/../../.."
```
ata-nas, line 10: Again, here we are going out of the build root `after` we have done our initial setup of copying all needed files. The execution itself also must not go out of the build root, everything that is needed for the execution must be present and accessible inside the build root.

simulator/docker/docker-compose.yml
```diff
@@ -9,11 +7,13 @@ services:
...
+      - .env
     environment:
-      - GRPC_SERVER_ADDRESS=block-node-server
-      - BLOCK_STREAM_SIMULATOR_MODE=PUBLISHER
-      - PROMETHEUS_ENDPOINT_ENABLED=true
-      - PROMETHEUS_ENDPOINT_PORT_NUMBER=9998
+      - GRPC_SERVER_ADDRESS=${GRPC_SERVER_ADDRESS}
+      - BLOCK_STREAM_SIMULATOR_MODE=${PUBLISHER_BLOCK_STREAM_SIMULATOR_MODE}
+      - PROMETHEUS_ENDPOINT_ENABLED=${PROMETHEUS_ENDPOINT_ENABLED}
+      - PROMETHEUS_ENDPOINT_PORT_NUMBER=${PUBLISHER_PROMETHEUS_ENDPOINT_PORT_NUMBER}
```
ata-nas, line 16: Do we need the `environment` since now we have the `env_file`? One will take precedence over the other, I do not think we should be using both.

simulator/docker/docker-compose.yml
```diff
@@ -28,11 +28,13 @@ services:
...
+      - .env
     environment:
-      - GRPC_SERVER_ADDRESS=block-node-server
-      - BLOCK_STREAM_SIMULATOR_MODE=CONSUMER
-      - PROMETHEUS_ENDPOINT_ENABLED=true
-      - PROMETHEUS_ENDPOINT_PORT_NUMBER=9997
+      - GRPC_SERVER_ADDRESS=${GRPC_SERVER_ADDRESS}
+      - BLOCK_STREAM_SIMULATOR_MODE=${CONSUMER_BLOCK_STREAM_SIMULATOR_MODE}
+      - PROMETHEUS_ENDPOINT_ENABLED=${PROMETHEUS_ENDPOINT_ENABLED}
+      - PROMETHEUS_ENDPOINT_PORT_NUMBER=${CONSUMER_PROMETHEUS_ENDPOINT_PORT_NUMBER}
```
ata-nas, line 37: Do we need the `environment` since now we have the `env_file`? One will take precedence over the other, I do not think we should be using both.

simulator/docker/update-env.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# For publisher service
+PUBLISHER_BLOCK_STREAM_SIMULATOR_MODE=PUBLISHER
+PUBLISHER_PROMETHEUS_ENDPOINT_PORT_NUMBER=9998
+
+# For consumer service
+CONSUMER_BLOCK_STREAM_SIMULATOR_MODE=CONSUMER
+CONSUMER_PROMETHEUS_ENDPOINT_PORT_NUMBER=9997
+EOL
+
```
ata-nas, line 21: here we could add a conditional:

```shell
if [ true = "$is_debug" ]; then
  # The server will wait for the debugger to attach on port 5005
  # JProfiler can attach on port 8849
  echo "SERVER_OPTS='-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005' -agentpath:/path/to/libjprofilerti.so=port=8849 " >> .env
fi
```

In this way, if you update the env conditionally directly from gradle task (like in the server) we can very easily enable debugging. And you can simply create the `startDockerContainerDebug` task also in the simulator so we can bind to it remotely. I would not reuse the `5005` port though, the agentlib path to jprofiler could be dropped as well I think.

simulator/docker/prepare-docker.sh
ata-nas, line 1: I think this file should be deleted. We do not need it as it simply copies resources which must be done with gradle task before the actual start of the container starting task. You can take inspiration from the server where in the startDockerContainer task we directly update the .env file by executing the `update-env.sh`. And then we can simply call the docker compose command. In this way it is a lot cleaner, we have one less file and we can conditionally put options inside the .env file very easily from the gradle task (as is in server) and enable debug for example

simulator/build.gradle.kts
```diff
@@ -94,22 +94,26 @@ tasks.named("processResources") { dependsOn(tasks.named("untarTestBlockStream"))
...
+    tasks.register<Exec>("startDockerContainer") {
+        description = "Creates and starts the docker image of the Block Stream Simulator"
+        group = "docker"
 
-        println("✅ Docker simulator is now running!")
-        println("🧹 Build artifacts have been cleaned up")
+        dependsOn(copyDockerFolder, tasks.assemble)
+        workingDir(dockerBuildRootDirectory)
+        commandLine("sh", "-c", "./prepare-docker.sh")
     }
```
ata-nas, line 119: Let's also add the `startDockerContainerDebug` and `stopDockerContainer` tasks as it is in the server.

ata-nas (commented):
Much improved @georgi-l95. 
A few more things I think we should do, described below, but here the gist of it:

1. Add `startDockerContainerDebug` and `stopDockerContainer`
1. We can remove the `prepare-docker.sh` and simplify a lot the process
1. The copying of resources in the `prepare-docker.sh` must be extracted and done in the copy task in gradle which you created.
1. We can very easily add some conditional logic to the `update-env.sh` to enable debugging for example.

The inspiration is coming from server.

simulator/docker/docker-compose.yml
georgi-l95: Usually no. But in our case that's the case. Meaning we need them, but not all. Reason is because we are starting the same container in twice with different configuration, so it needs to now, which variable to map. Pushed a commit where I remove the unneded.

simulator/docker/docker-compose.yml
georgi-l95: Usually no. But in our case that's the case. Meaning we need them, but not all. Reason is because we are starting the same container in twice with different configuration, so it needs to now, which variable to map. Pushed a commit where I remove the unneded.

simulator/build.gradle.kts
georgi-l95: Added `stopDockerContainer`, but skipped `startDockerContainerDebug` for two reasons:
1. Currently we are starting two instances of the same application, therefore have to take care of mapping the ports correcty.
2. There is no request for this now, don't want to add unnecessary things for them to never be used. When there is a need, we can add it easily.

simulator/docker/prepare-docker.sh
georgi-l95: Deleted, because everything is moved in gradle tasks. 👍

simulator/docker/update-env.sh
georgi-l95: Answered here: https://github.com/hashgraph/hedera-block-node/pull/416#discussion_r1893164374

simulator/docker/prepare-docker.sh
georgi-l95: Removed this script.

simulator/docker/prepare-docker.sh
georgi-l95: Moved 👍

simulator/docker/prepare-docker.sh
georgi-l95: Moved in gradle task 👍

server/build.gradle.kts
```diff
@@ -79,11 +79,11 @@ val copyDockerFolder: TaskProvider<Copy> =
         into(dockerBuildRootDirectory)
     }
 
-// @todo(#343) - createProductionDotEnv is temporary and used by the suites,
+// @todo(#343) - createDotEnv is temporary and used by the suites,
 //  once the suites no longer rely on the .env file from the build root we
 //  should remove this task
-val createProductionDotEnv: TaskProvider<Exec> =
-    tasks.register<Exec>("createProductionDotEnv") {
+val createDotEnv: TaskProvider<Exec> =
```
georgi-l95, line 85: Changed to createDotEnv for several reasons:
1. We have similar task in simulator and they essentially do the same thing (create env file). So i wanted to match the names.
2. This is temporary here.
3. Those are default, not production values, but moreover it does not make sense at all to name the simulator task `createProductionDotEnv`.

simulator/docker/prepare-docker.sh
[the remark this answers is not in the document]
georgi-l95: Opened here https://github.com/hashgraph/hedera-block-node/issues/429, we can disscuss it more on some of the ceremonies.

server/build.gradle.kts
ata-nas, line 85: This is planned to be removed. I do not think we need to change anything here, The update shell could be called by the command line from gradle, IMO this change is unnecessary and would entail additional maintenance.

simulator/build.gradle.kts
```diff
@@ -93,3 +93,71 @@ tasks.register<Copy>("untarTestBlockStream") {
...
+
+val startDockerContainer: TaskProvider<Exec> =
+    tasks.register<Exec>("startDockerContainer") {
+        description = "Creates and starts the docker image of the Block Stream Simulator"
+        group = "docker"
+
+        dependsOn(createDotEnv, tasks.assemble)
+        workingDir(dockerBuildRootDirectory)
+
+        commandLine("sh", "-c", "docker compose -p simulator up -d")
```
ata-nas, line 153: here you can simply do:

```suggestion
        dependsOn(tasks.assemble)
        workingDir(dockerBuildRootDirectory)

        commandLine("sh", "-c", "./update-env.sh && docker compose -p simulator up -d")
```
we are in the correct working directory (same as in the `createDotEnv task`)

jsync-swirlds:
Quick note. Consensus teams were looking to use the simulator for their own testing, and found the readme and quickstart woefully outdated.
Would you mind creating an issue to update those documents after this PR merges to reflect all of the recent changes?

suites/build.gradle.kts
```diff
@@ -43,8 +43,8 @@ tasks.register<Test>("runSuites") {
     group = "suites"
     modularity.inferModulePath = false
 
-    // @todo(#343) - :server:createProductionDotEnv should disappear
-    dependsOn(":server:createDockerImage", ":server:createProductionDotEnv")
+    // @todo(#343) - :server:createDotEnv should disappear
+    dependsOn(":server:createDockerImage", ":server:createDotEnv")
```
ata-nas, line 47: Let's revert this, it is planned to be removed and is a temporary workaround.

ata-nas (commented):
@georgi-l95 this looks amazing!

However I would suggest to revert the `createDotEnv` chagnes and simply call the `./update-env.sh` directly at the `startDockerContainer` task. This simplifies

server/build.gradle.kts
ata-nas: This was added as a workaround because the suites rely on a built .env of the server to run (they need the version from there to build the correct test container). The only way to do that (as a workaround at that time) was to create this simple task in order to do `dependsOn` that task. I suggest we do not touch that and we simply call the local to the simulator `update-env.sh` directly form the `commandline` of gradle inside the `startDockerContainer` task. In this way we will not have these changes here and we will have one less task in the gradle file of the simulator and we are completely isolated.

simulator/build.gradle.kts
georgi-l95: Added, good suggestion

georgi-l95:
> Quick note. Consensus teams were looking to use the simulator for their own testing, and found the readme and quickstart woefully outdated. Would you mind creating an issue to update those documents after this PR merges to reflect all of the recent changes?

Yes, I've noticed that too, mostly about configuration and opened a issue couple of days ago. https://github.com/hashgraph/hedera-block-node/issues/407

suites/build.gradle.kts
```diff
@@ -43,7 +43,7 @@ tasks.register<Test>("runSuites") {
     group = "suites"
     modularity.inferModulePath = false
 
-    // @todo(#343) - :server:createProductionDotEnv should disappear
+    // @todo(#343) - :server:createDotEnv should disappear
```
ata-nas, line 46:
```suggestion
    // @todo(#343) - :server:createProductionDotEnv should disappear
```

ata-nas (dismissed):
@georgi-l95 looks great!

suites/build.gradle.kts
ata-nas: @georgi-l95 feel free to close this comment, it is non blocking, up to you.

georgi-l95: Thanks for catching it, renamed.

ata-nas (approved):
Looks good! 🔥

jsync-swirlds (approved):

AlfredoG87 (commented):
LGTM

AlfredoG87 (approved):
LGTM

PR #419, opened 2024-12-17
feat: deterministic docker changes part 1

### LINKED ISSUE #409: Configure the Block Node Dockerfile to build deterministic images
As a Block Node engineer
I want to configure the Dockerfile to build deterministically
So that I can ensure our build image aligns with our deployment goals of determinism.

* This is the second step in achieving a deterministic software deployment as outlined by the `hedera-services` build.
* Integrate deterministic Dockerfile directives to produce a deterministic image

mattp-swirldslabs:
**Description**:
* Added Dockerfile determinism directives
* Added workflow entry point... another PR to come

**Related issue(s)**:

Fixes #409 

**Notes for reviewer**:

**Checklist**


AlfredoG87 (dismissed):
LGTM

.github/workflows/zxc-verify-docker-build-determinism.yaml
```diff
@@ -0,0 +1,81 @@
...
+      java-distribution:
+        description: "Java JDK Distribution:"
+        type: string
+        required: false
+        default: "temurin"
+      java-version:
+        description: "Java JDK Version:"
+        type: string
+        required: false
+        default: "21.0.4"
```
jsync-swirlds, line 34: For OpenJDK or Temurin builds we should be using `21.0.5` (the latest security patch version).

.github/workflows/zxc-verify-gradle-build-determinism.yaml
```diff
@@ -53,10 +52,171 @@ permissions:
...
+    needs:
+      - generate-baseline
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-22.04
+          - ubuntu-20.04
+          - windows-2022
+          - windows-2019
```
jsync-swirlds, line 134: There is zero reason, in my opinion, to support Block Node builds on Windows.  The target should always be Kubernetes, and Kubernetes does not run natively on Windows.

We can reduce cost (and build time) by not running Windows builds here, and should do so, unless there is a Solo requirement (even then we should discuss with Solo team how that requirement can be relaxed in the future).

gradlew.bat
```diff
@@ -43,11 +43,11 @@ set JAVA_EXE=java.exe
...
-echo.
-echo ERROR: JAVA_HOME is not set and no 'java' command could be found in your PATH.
-echo.
-echo Please set the JAVA_HOME variable in your environment to match the
-echo location of your Java installation.
+echo. 1>&2
+echo ERROR: JAVA_HOME is not set and no 'java' command could be found in your PATH. 1>&2
+echo. 1>&2
+echo Please set the JAVA_HOME variable in your environment to match the 1>&2
+echo location of your Java installation. 1>&2
```
jsync-swirlds, line 50: These kinds of "redirected" error blocks often read better as a quick function:
```suggestion
write_java_home_error() {
  echo
  echo ERROR: JAVA_HOME is not set and no 'java' command could be found in your PATH.
  echo
  echo Please set the JAVA_HOME variable in your environment to match the
  echo location of your Java installation.
}
write_java_home_error 1>&2
```

server/docker/Dockerfile
```diff
@@ -1,5 +1,84 @@
...
+             ;; \
+           ppc64el|powerpc:common64) \
+             ESUM='9574828ef3d735a25404ced82e09bf20e1614f7d6403956002de9cfbfcb8638f'; \
+             BINARY_URL='https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.1%2B12/OpenJDK21U-jdk_ppc64le_linux_hotspot_21.0.1_12.tar.gz'; \
+             ;; \
+           *) \
+             echo "Unsupported arch: ${ARCH}"; \
+             exit 1; \
+             ;; \
+        esac; \
```
jsync-swirlds, line 46: These versions are well behind; and we _should not_ be using them.  We probably need to update the checksum and URL for the latest 21.0.5 build.
Also; is there a reason we don't use the `apt` or `dpkg` installs?

server/docker/Dockerfile
```diff
@@ -1,5 +1,84 @@
...
+COPY --from=openjdk-builder /usr/local/java/ /usr/local/java/
+
+# Install Basic OS Requirements
+RUN apt-get update && \
+	apt-get install --yes --no-install-recommends tar gzip openssl curl && \
+	apt-get autoremove --yes && \
+	apt-get autoclean --yes && \
+	apt-get clean all --yes && \
+    rm -rf /var/lib/{apt,dpkg,cache,log}/
+
```
jsync-swirlds, line 81: We should probably work with release engineering to have a baseline image (with the JDK and tool installs added) published to our registry on a regular basis instead of doing all these updates in every single build.
It would remove a surprising amount of time and cost from the thousands of these builds (across multiple projects) that run each month...

jsync-swirlds (approved):
Nothing blocking, but we should upgrade the Java version (both in scripts and in the dockerfile); and I have a few questions and suggestions for future improvements (e.g. part 2).

.github/workflows/release-push-image.yaml
```diff
@@ -43,6 +43,22 @@ env:
...
+#      java-distribution: ${{ inputs.java-distribution || 'temurin' }}
+#      java-version: ${{ inputs.java-version || '21.0.4' }}
+
+#  check-docker:
+#    name: Docker
+#    uses: ./.github/workflows/zxc-verify-docker-build-determinism.yaml
+#    with:
+#      ref: ${{ github.event.inputs.ref || '' }}
+#      java-distribution: ${{ inputs.java-distribution || 'temurin' }}
+#      java-version: ${{ inputs.java-version || '21.0.4' }}
```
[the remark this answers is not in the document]
mattp-swirldslabs: Yes, I need to get the yamls into `main` before they're recognized by GHA so I can start testing on my branch.

.github/workflows/zxc-verify-gradle-build-determinism.yaml
mattp-swirldslabs: Copy/paste from Nathan's process

server/docker/Dockerfile
mattp-swirldslabs: Copy/paste from Nathan's code

.github/workflows/zxc-verify-gradle-build-determinism.yaml
jsync-swirlds: That process is for Consensus Node, which may be inappropriate for our needs here.
It can/should be future PR, but we shouldn't blindly follow the existing approach as a cookbook either.

gradlew.bat
jsync-swirlds: This isn't optimization, it's clarity and error reduction.
Also, perhaps don't push a PR if the response to everything will be "I'm not changing it"...

jsync-swirlds: It's not optimization, just readability and less error-prone when editing...

AlfredoG87 (approved):

rbarkerSL (approved):

PR #446, opened 2025-01-04
refactor: Logs improvements

### LINKED ISSUE #443: Improve Logs Format
Current Log Output of the Block-Node app is as follows:

```
Jan 04, 2025 12:34:43 AM com.hedera.block.server.Server main
INFO: Starting BlockNode Server
Jan 04, 2025 12:34:43 AM com.hedera.block.server.ServerConfig <init>
INFO: Server configuration server.maxMessageSizeBytes: 4194304
Jan 04, 2025 12:34:43 AM com.hedera.block.server.ServerConfig <init>
INFO: Server configuration server.port: 8080
Jan 04, 2025 12:34:43 AM com.hedera.block.server.notifier.NotifierConfig <init>
INFO: Notifier configuration notifier.ringBufferSize: 1024
Jan 04, 2025 12:34:43 AM com.hedera.block.server.consumer.ConsumerConfig <init>
INFO: Consumer configuration timeoutThresholdMillis: 1500
Jan 04, 2025 12:34:43 AM com.hedera.block.server.service.ServiceConfig <init>
INFO: Service configuration service.delayMillis: 500
Jan 04, 2025 12:34:43 AM com.hedera.block.server.mediator.MediatorConfig <init>
INFO: Mediator configuration mediator.ringBufferSize: 4194304
Jan 04, 2025 12:34:43 AM com.hedera.block.server.mediator.MediatorConfig <init>
INFO: Mediator configuration mediator.type: PRODUCTION
Jan 04, 2025 12:34:43 AM com.hedera.block.server.verification.VerificationConfig <init>
INFO: Verification configuration enabled: true
Jan 04, 2025 12:34:43 AM com.hedera.block.server.verification.VerificationConfig <init>
INFO: Verification configuration sessionType: ASYNC
Jan 04, 2025 12:34:43 AM com.hedera.block.server.verification.VerificationConfig <init>
INFO: Verification configuration hashCombineBatchSize: 32
Jan 04, 2025 12:34:43 AM com.hedera.block.server.producer.ProducerConfig <init>
INFO: Producer configuration producer.type: PRODUCTION
Jan 04, 2025 12:34:44 AM com.hedera.block.server.persistence.storage.write.BlockAsLocalDirWriter <init>
INFO: Initializing com.hedera.block.server.persistence.storage.write.BlockAsLocalDirWriter...
Jan 04, 2025 12:34:44 AM com.hedera.block.common.utils.FileUtilities createFolderPathIfNotExists
INFO: Requested [Block Node Live Root Directory] not created because the directory already exists at '/app/hashgraph/blocknode/data/live'
Jan 04, 2025 12:34:44 AM com.hedera.block.server.persistence.storage.read.BlockAsLocalDirReader <init>
INFO: Initializing com.hedera.block.server.persistence.storage.read.BlockAsLocalDirReader...
Jan 04, 2025 12:34:44 AM io.helidon.common.features.HelidonFeatures features
INFO: Helidon SE 4.1.1 features: [Config, Encoding, Media, Metrics, WebServer]
Jan 04, 2025 12:34:44 AM io.helidon.webserver.ServerListener start
```

The above seems like is not well formatted, appears to be `multiline`, also it seems like Helidon is using a separate logging format.

- [x] Improves the Log Format to `<Date> <Time> <Level> [<class> <method>] <Message> <throwable>`
- [x] Uses a single format that gets loaded from the very start.
- [x] The block-node should use UTC timezone, and so the logs

### LINKED ISSUE #444: Add File Handler to Log Output
As a BlockNode Operator and Developer
I want to have a failback or backup option in case my observability stack malfunctions or a log is lost in the process.
For that the block-node app should write logs to both `console` and a given file. 
Currently only writes to console and the observability stack using `promtail` via that handler.

### LINKED ISSUE #445: Make Logs properties configurable

AlfredoG87:
**Description**:
Improves logs by doing the several changes:

Docker Image
- [x] Java uses UTC Timezone explicitly, so logs always have the UTC date time values.
- [x] Java loads the `logging.properties' from the very beginning.
- [x] Better Log Format is defined on `logging.properties` (mean to be single line)
- [x] FileHandler is added to the logs handlers, creating a log file backup on file
- [x] Removed the need to Set Helidon log config separately and manually on Server initialization.

Helm Chart - K8 (Cloud Env)
- [x] Defines a new values config with the default logging.properties content, and overridable.
- [x] Creates a ConfigMap that generates the `logging.properties` that the block-node server deployment will use.

Docker-Compose (Local Dev Env)
- [x] Mounts the `logging.properties` on `build/docker` where the docker stack is created to the server instance, so it can be easily configurable and properties changed.

**Related issue(s)**:

Fixes #445 
Fixes #444 
Fixes #443 

**Notes for reviewer**:
Example of how logs are looking now:

```
Picked up JAVA_TOOL_OPTIONS: -Djava.util.logging.config.file=/app/logs/config/logging.properties -Duser.timezone=UTC
2025-01-05 06:46:12 INFO    [com.hedera.block.server.Server main] Starting BlockNode Server 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.producer.ProducerConfig <init>] Producer configuration producer.type: PRODUCTION 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.ServerConfig <init>] Server configuration server.maxMessageSizeBytes: 4194304 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.ServerConfig <init>] Server configuration server.port: 8080 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.service.ServiceConfig <init>] Service configuration service.delayMillis: 500 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.mediator.MediatorConfig <init>] Mediator configuration mediator.ringBufferSize: 4194304 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.mediator.MediatorConfig <init>] Mediator configuration mediator.type: PRODUCTION 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.notifier.NotifierConfig <init>] Notifier configuration notifier.ringBufferSize: 1024 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.consumer.ConsumerConfig <init>] Consumer configuration timeoutThresholdMillis: 1500 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.persistence.storage.write.BlockAsLocalDirWriter <init>] Initializing com.hedera.block.server.persistence.storage.write.BlockAsLocalDirWriter... 
2025-01-05 06:46:12 INFO    [com.hedera.block.common.utils.FileUtilities createFolderPathIfNotExists] Requested [Block Node Live Root Directory] not created because the directory already exists at '/app/hashgraph/blocknode/data/live' 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.persistence.storage.read.BlockAsLocalDirReader <init>] Initializing com.hedera.block.server.persistence.storage.read.BlockAsLocalDirReader... 
2025-01-05 06:46:12 INFO    [io.helidon.webserver.ServerListener start] [0x6cc709a3] http://0.0.0.0:8080 bound for socket '@default' 
2025-01-05 06:46:12 INFO    [io.helidon.common.features.HelidonFeatures features] Helidon SE 4.1.1 features: [Config, Encoding, Media, Metrics, WebServer] 
2025-01-05 06:46:12 INFO    [io.helidon.logging.jul.JulProvider doConfigureLogging] Logging at runtime configured using /app/logs/config/logging.properties 
2025-01-05 06:46:12 INFO    [io.helidon.webserver.LoomServer startIt] Started all channels in 10 milliseconds. 914 milliseconds since JVM startup. Java 21.0.1+12-LTS 
2025-01-05 06:46:12 INFO    [com.hedera.block.server.BlockNodeApp start] Block Node Server started at port: 8080 
```

**Checklist**

### DIFF
--- a/charts/hedera-block-node/values.yaml
+++ b/charts/hedera-block-node/values.yaml
@@ -133,7 +133,7 @@ blockNode:
       java.util.logging.SimpleFormatter.format: "%1$tF %1$tT %4$-7s [%2$s] %5$s %n"
 
 kubepromstack:
-  enabled: false
+  enabled: true
   prometheusOperator:
     namespaces:
       releaseNamespace: true
--- a/charts/hedera-block-node/values.yaml
+++ b/charts/hedera-block-node/values.yaml
@@ -101,50 +101,36 @@ blockNode:
     # Available Levels are (from most verbose to least verbose):
     # ALL FINEST FINER FINE CONFIG INFO WARNING SEVERE OFF
     level: "INFO"
-    ################################################################################
-    # SimpleFormatter single-line format configuration
-    ################################################################################
-    # The format syntax uses java.util.Formatter.
-    # The parameters are:
-    #   %1$ - date/time (java.util.Date)
-    #   %2$ - source (usually class and method)
-    #   %3$ - logger?s name
-    #   %4$ - log level
-    #   %5$ - log message
-    #   %6$ - throwable trace
-    #
-    # Example to produce a line such as:
-    # 2025-01-04 00:34:43 INFO [com.hedera.block.server.Server main] Starting BlockNode Server
-    #
-    format: "%1$tF %1$tT %4$-7s [%2$s] %5$s %n"
     loggingProperties:
-      # Add any additional logging configuration properties here
-      - name: "io.helidon.webserver.level"
-        value: "INFO"
-      - name: "io.helidon.config.level"
-        value: "SEVERE"
-      - name: "io.helidon.security.level"
-        value: "INFO"
-      - name: "io.helidon.common.level"
-        value: "INFO"
-      - name: "handlers"
-        value: "java.util.logging.ConsoleHandler, java.util.logging.FileHandler"
-      - name: "java.util.logging.ConsoleHandler.level"
-        value: "FINE"
-      - name: "java.util.logging.ConsoleHandler.formatter"
-        value: "java.util.logging.SimpleFormatter"
-      - name: "java.util.logging.FileHandler.pattern"
-        value: "/app/logs/blocknode-%g.log"
-      - name: "java.util.logging.FileHandler.append"
-        value: "true"
-      - name: "java.util.logging.FileHandler.limit"
-        value: "5000000"
-      - name: "java.util.logging.FileHandler.count"
-        value: "5"
-      - name: "java.util.logging.FileHandler.level"
-        value: "FINE"
-      - name: "java.util.logging.FileHandler.formatter"
-        value: "java.util.logging.SimpleFormatter"
+      io.helidon.webserver.level: "INFO"
+      io.helidon.webserver.access.level: "INFO"
+      io.helidon.config.level: "SEVERE"
+      io.helidon.security.level: "INFO"
+      io.helidon.common.level: "INFO"
+      handlers: "java.util.logging.ConsoleHandler, java.util.logging.FileHandler"
+      java.util.logging.ConsoleHandler.level: "FINE"
+      java.util.logging.ConsoleHandler.formatter: "java.util.logging.SimpleFormatter"
+      java.util.logging.FileHandler.pattern: "/app/logs/blocknode-%g.log"
+      java.util.logging.FileHandler.append: "true"
+      java.util.logging.FileHandler.limit: "5_000_000"
+      java.util.logging.FileHandler.count: "5"
+      java.util.logging.FileHandler.level: "FINE"
+      java.util.logging.FileHandler.formatter: "java.util.logging.SimpleFormatter"
+      ################################################################################
+      # SimpleFormatter single-line format configuration
+      ################################################################################
+      # The format syntax uses java.util.Formatter.
+      # The parameters are:
+      #   %1$ - date/time (java.util.Date)
+      #   %2$ - source (usually class and method)
+      #   %3$ - logger?s name
+      #   %4$ - log level
+      #   %5$ - log message
+      #   %6$ - throwable trace
+      #
+      # Example to produce a line such as:
+      # 2025-01-04 00:34:43 INFO [com.hedera.block.server.Server main] Starting BlockNode Server
+      java.util.logging.SimpleFormatter.format: "%1$tF %1$tT %4$-7s [%2$s] %5$s %n"
 
 kubepromstack:
   enabled: false
--- a/charts/hedera-block-node/values.yaml
+++ b/charts/hedera-block-node/values.yaml
@@ -79,6 +79,7 @@ blockNode:
     # Add any additional env configuration here
     # key: value
     BLOCKNODE_STORAGE_ROOT_PATH: "/app/storage"
+    JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties -Duser.timezone=UTC"
   secret:
     # if blank will use same as AppVersion of chart.
     PRIVATE_KEY: "fake_private_key"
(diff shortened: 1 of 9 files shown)


server/docker/update-env.sh
```diff
@@ -38,7 +38,11 @@ else
   echo "JAVA_OPTS='-Xms16G -Xmx16G'" >> .env
 fi
 
+# Set the timezone to UTC and the logging properties file
+# file is mounted in the docker-compose.yml, changes to the file will be reflected in the container by simply restarting it
+echo "JAVA_TOOL_OPTIONS='-Djava.util.logging.config.file=/app/logs/config/logging.properties -Duser.timezone=UTC'" >> .env
```
ata-nas, line 43: Is it a good idea to change the jvm timezone (whatever it may be) to UTC always?

server/src/main/java/com/hedera/block/server/Server.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024-2025 Hedera Hashgraph, LLC
```
ata-nas, line 2: Was this header updated automatically (spotlessApply)? If these should be updated, this should come from spotless I think.

charts/hedera-block-node/values.yaml
```diff
@@ -96,6 +97,40 @@ blockNode:
...
+  logs:
+    # Available Levels are (from most verbose to least verbose):
+    # ALL FINEST FINER FINE CONFIG INFO WARNING SEVERE OFF
+    level: "INFO"
+    loggingProperties:
+      io.helidon.webserver.level: "INFO"
+      io.helidon.webserver.access.level: "INFO"
+      io.helidon.config.level: "SEVERE"
+      io.helidon.security.level: "INFO"
+      io.helidon.common.level: "INFO"
```
ata-nas, line 109: Are these logging levels the best? Info for server seems way too loose IMO. INFO logs would get quite cluttered for things we expect to happen (this is a wild guess, most servers log a lot of "debug" or "trace" lines). Maybe we should consider another default? I wonder what the team thinks about this.

ata-nas (commented):
LGTM, couple of questions/considerations.

server/docker/update-env.sh
AlfredoG87: Good question, that is one discussion we can have as a team, however I think that the alternative is that the block-node will take the time that is set at the Node (OS) level and it will be harder to troubleshoot, having a standard time makes it easy to convert it to your local time on Grafana at the visualization moment.

server/src/main/java/com/hedera/block/server/Server.java
AlfredoG87: Yes, all of these are required by Spotless indeed. and the changes made by running `spotless apply`

charts/hedera-block-node/values.yaml
AlfredoG87: I agree with you that we need to discuss this, for now, I just copy what we already have on the `logging.properties` file to helm in  the form of values, so those are easily configurable by the operator.

server/docker/update-env.sh
ata-nas: System time should always be UTC I think (System.now). User time is different however. I think we are mixing the two. But please correct me if I am wrong. (Logs should be using System.time which should be a unix timestamp).

jsync-swirlds: User and user time are set in the container configuration, and we should probably set the user time to UTC at the container level (so we do not need to override in JVM command line).
I second the suggestion to discuss across the team.

server/src/main/java/com/hedera/block/server/Server.java
jsync-swirlds: I, for one, am looking forward to changing the repo over to SPDX so we have much simpler single-line assertions.

server/src/main/java/com/hedera/block/server/Server.java
```diff
@@ -49,14 +44,6 @@ private Server() {}
     public static void main(final String[] args) throws IOException {
         LOGGER.log(INFO, "Starting BlockNode Server");
 
-        // Set the global configuration
-        final Config config = Config.builder()
-                .sources(file(Paths.get("/app", LOGGING_PROPERTIES)).optional())
-                .sources(classpath("helidon.properties"))
```
jsync-swirlds, line 55: This is removing the _helidon_ configuration, which isn't just logging, as I understand it.  Are we certain that doesn't have unexpected side effects?

server/docker/update-env.sh
AlfredoG87: I like the idea of setting it at the dockerfile (container level) 👍

server/src/main/java/com/hedera/block/server/Server.java
AlfredoG87: Helidon by default  already loading those config properties files, no need to be explicit in my understanding.

That said, this was added originally by @mattp-swirldslabs and I believe it was due to the logs, but we can confirm, since I had the same concern. 👍

server/src/main/java/com/hedera/block/server/Server.java
ata-nas: Moved to SPDX.

server/docker/update-env.sh
jsync-swirlds: Do we have a story to make that change in the future, or will that change be made in this PR?

ata-nas: @jsync-swirlds was going to ask the same, should we continue with this as-is and discuss in the futre what the levels should be, or should this be done now?

charts/hedera-block-node/values.yaml
ata-nas: @jsync-swirlds @AlfredoG87 should we agree to go with the levels as they are defined for now and make a story for the future to discuss what the final levels should be, or should we do it now?

jsync-swirlds: I believe we are better off keeping the levels as defined and adding a story to revisit.

server/docker/update-env.sh
AlfredoG87: We should do it in a follow up PR, is not trivial (we need to install tzdata package) and it could affect the other PR in progress the deterministic builds. so for now, we should move forward and leave out of this PR.

This is the created [ticket](https://github.com/hashgraph/hedera-block-node/issues/450)

charts/hedera-block-node/values.yaml
AlfredoG87: @ata-nas  if nothing further I would appreciate Resolving the conversation.

server/docker/update-env.sh
ata-nas: Sounds good, I agree with you both, see no reason to block the progress here now. Resolved.

charts/hedera-block-node/values.yaml
ata-nas: I agree we should not block the progress here. Resolved.

mattp-swirldslabs (approved):
Looks good. Thanks for the changes.

PR #447, opened 2025-01-06
chore: Switch license declarations to SPDX

jsync-swirlds:
Change all license declarations to SPDX format and change spotless configuration to require the same.

### DIFF
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -1,18 +1,4 @@
-##
-# Copyright (C) 2024 Hedera Hashgraph, LLC
-#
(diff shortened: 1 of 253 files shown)


andrewb1269hg (approved):
Approve the Devops-CI assigned files.

AlfredoG87 (approved):
LGTM

mattp-swirldslabs (approved):
Looks good

ata-nas (approved):
Looks good! As a side effect, we have also had the chance to run spotless on all files now with palantir, which is nice, some code was formatted to match that.

PR #449, opened 2025-01-07
feat: add compression mode support for BlockAsFileReader

### LINKED ISSUE #350: feat: add compression mode support for BlockAsFileReader
## Story

**AS A** Block Node Operator/User
**I WANT** to be able to retrieve compressed block files that are written to the file system
**SO THAT** I will be able to use them

## Tech Notes

- **DEPENDS ON #282**

- See #125 where we have defined the core functionality of the `block-as-file`.

- After the implementation of #282 we should support the corresponding reading logic.

- The scope of this issue is to be able to read compressed block files.

ata-nas:
**Description**:

- add compression support for the reader

**Related issue(s)**:

Fixes #350 

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
@@ -149,22 +149,28 @@ public record PersistenceStorageConfig(
          * This type of compression is used to compress the blocks using the
          * `Zstandard` algorithm.
          */
-        ZSTD(0, 20),
+        ZSTD(0, 20, ".zstd"),
         /**
          * This type means no compression will be done.
          */
-        NONE(Integer.MIN_VALUE, Integer.MAX_VALUE);
+        NONE(Integer.MIN_VALUE, Integer.MAX_VALUE, "");
 
         private final int minCompressionLevel;
         private final int maxCompressionLevel;
+        private final String fileExtension;
 
-        CompressionType(final int minCompressionLevel, final int maxCompressionLevel) {
+        CompressionType(final int minCompressionLevel, final int maxCompressionLevel, final String fileExtension) {
             this.minCompressionLevel = minCompressionLevel;
             this.maxCompressionLevel = maxCompressionLevel;
+            this.fileExtension = fileExtension;
         }
 
         public void verifyCompressionLevel(final int levelToCheck) {
             Preconditions.requireInRange(levelToCheck, minCompressionLevel, maxCompressionLevel);
+        }
+
+        public String getFileExtension() {
+            return fileExtension;
         }
     }
 }
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
@@ -62,7 +62,7 @@ public final class BlockAsLocalFilePathResolver implements BlockPathResolver {
 
     @NonNull
     @Override
-    public Optional<Path> findBlock(final long blockNumber) { // todo add archive root search & abstract if common
+    public Optional<Path> findBlock(final long blockNumber) {
         Preconditions.requireWhole(blockNumber);
         final Path liveRawRootPath = resolveLiveRawPathToBlock(blockNumber);
         final Path compressionExtendedLiveRawRootPath =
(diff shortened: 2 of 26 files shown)


server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -149,22 +149,28 @@ public enum CompressionType {
          * This type of compression is used to compress the blocks using the
          * `Zstandard` algorithm.
          */
-        ZSTD(0, 20),
+        ZSTD(0, 20, ".zstd"),
```
mattp-swirldslabs, line 152: I think it would be helpful to log at INFO the current configuration at startup like we do here: https://github.com/hashgraph/hedera-block-node/blob/c303fa0f19f2dc10406de9fd8d0d2c9562bf7e0d/server/src/main/java/com/hedera/block/server/producer/ProducerConfig.java#L17
That way, it's easy to confirm if you're using compression, which persistence type you have configured, etc. I think the other Config classes do this so we'll be able to confirm the server configuration by looking at the log output in Grafana rather than logging onto the container, etc.

mattp-swirldslabs (approved):
Nice work! One other thing to adjust would be this logger: https://github.com/hashgraph/hedera-block-node/blob/c303fa0f19f2dc10406de9fd8d0d2c9562bf7e0d/server/docker/logging.properties#L27. Please confirm you're getting output from somewhere in the persistence layer so we can uncomment a line in the logging.properties to troubleshoot issues.

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
```diff
@@ -1,50 +1,79 @@
...
+                FileUtilities.appendExtension(liveRawRootPath, compressionType.getFileExtension());
+        if (Files.exists(compressionExtendedLiveRawRootPath)) {
+            return Optional.of(compressionExtendedLiveRawRootPath);
+        } else if (Files.exists(liveRawRootPath)) {
+            return Optional.of(liveRawRootPath);
+        } else {
+            return Optional.empty();
+        }
+    } // todo consider to add additional handling here, like test for other compression types (currently not existing)
+    // and also look for archived blocks (will be implemented in a follow-up PR)
```
AlfredoG87, line 78: are we using the todo and other labels defined for the team with an explicit link to a ticket?

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
```diff
@@ -18,45 +19,50 @@
...
     }
 
     private BlockUnparsed doRead(final Path resolvedBlockPath) throws IOException, ParseException {
-        try (final ReadableStreamingData data = new ReadableStreamingData(resolvedBlockPath)) {
+        try (final ReadableStreamingData data =
+                new ReadableStreamingData(compression.wrap(Files.newInputStream(resolvedBlockPath)))) {
             return BlockUnparsed.PROTOBUF.parse(data);
-        }
+        } // todo this method must be extended to try to read uncompressed data if compressed data is not found
+        // need some mechanism that would be convenient to use not only here but also in other places
```
AlfredoG87, line 66: same as previous one?

AlfredoG87 (approved):
Looks good, just 2 nit questions.

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
ata-nas: In general yes, we must follow the convention, but since I am immediately changing this with the next step of #309, I did not add anything, this todo is more for me to mark for myself where I need to introduce my immediate changes, not for the team for the future. I will resolve this comment so that it is non-blocking for the progress of #309.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
ata-nas: Same as above.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
ata-nas: I am not opposed to that generally. Personally, I would prefer to rely on configuration (which you can just as easily see) that is testable and verifiable rather than a simple log. I propose to make a convention, lets vote on it, make a discussion and decide on what we should follow. I will resolve the comment so that it is non-blocking for the progress of the #309 epic. (discussion to follow)

PR #485, opened 2025-01-13
feat: add block archiving support

### LINKED ISSUE #339: feat: add block archiving support
## Story

**AS A** Block Node User
**I WANT** to have my block files archived
**SO THAT** I will have a reduced storage cost

## Tech Notes:

- **DEPENDS ON #282**

- See #125 where we have defined the design of the new `block-as-file`.

- The scope of this issue is to add archiving support to the block files that will be written to disk.

- We need to implement a process/thread, that will be going to the fs and will be archiving block files that are already written to disk.

- The parameters of this process/thread _should probably_ be configurable externally.

---

**The Archiving Process**:

- we have a _live_ root and an _archive_ root
- has configurable amount of decimal order of magnitude for amount
  of items to be archived at a given time
- we always want to be one step behind the live root (writing)
  e.g.
  we are archiving per 1000, the live root will be writing 2000-2999,
  1000-1999 is already written, and we are archiving at 0-999,
  so in that way we have the less than 1000 most recent and also the
  1000 before the most recent
- the "gap" should not be less than 1 archive **unit**. It could be more,
  but must not be less. Reasons for doing this:
    - we want to make sure the most recent x amount are rapidly available
    - we want to make sure that we do not accidentally try to archive an active block

**The Algorithm**:

1. Writing the blocks:
   1. We first write the block where it must be located in the live root
   1. Then, we have a process that will archive a given amount of blocks
      which will be an order of magnitude of 10
      - the archiving must be _behind_ the live root with at least 1 full order
        of magnitude as configured
      - the archiving must not go below 1 order of magnitude of difference
        between the live root and the archive root in order to ensure that
        no partially written block will be archived, but also to have the
        latest blocks present immediately which will be always more performant
        for fetching them
      - visualization:
        > written 2000
        > archived 0-999
        > **time passes...**
        > written 3000
        > archived 1000-1999
   1. After the archiving is done, there needs to be a symlink made in the live root
      - after we have archived, now we need to symlink the zip in the live root
      - the symlink will point to the archive root where the actual blocks now reside
      - the blocks now can again be resolved from the live root, but now they need to
        be searched for under the symlink
      - visualization:
        > let's archive all between 7..0001 and 7..1000:
        > ...
        > live/0/0/0/7/0/0/0001.blk
        > ...
        > live/0/0/0/7/0/0/1000.blk
        > - first of all, we zip all between 7..0001 and 7..1000
        > - the dest. of the zip will be at archive/0/0/0/7.zip/...
        > - archiving is now done completely
        > - now we need to symlink live/0/0/0/7.zip -> archive/0/0/0/7.zip
   1. After archiving is done completely and the symlink is made, we need to delete
      what is in the live root
      - visualization:
        > utilizing the previous example to archive all between 7..0001 and 7..1000:
        > - we are certain the archiving is entirely done and symlink is made
        > - we now delete live/0/0/0/7/...
        > - blocks between 7..0001 and 7..1000 can now be found through the symlink

1. Searching for a block:
   1. We first search for the block under the live root (currently implemented)
      - if found, we return it
   1. If not found, then we need to search the archive, again under the live root,
      since there will be a symlink there that points to the actual place in the
      archive root, but now we need to search for the symlink instead of where
      the block would actually be
      - if found there, we return it

ata-nas:
**Description**:

- add block archiving support

**Related issue(s)**:

Fixes #339 #351

**Notes for reviewer**:

- see #339 for the description of the archiving algorithm
- this is v1 of archiving, it is made to wire the logic and explore setbacks and considerations which need to be taken into account
- while developing this, there are multiple considerations already captured, mainly we will switch to a task based implementation. In fact that was the initial idea, however there were way to many unknowns due to the complexity of the archiving algorithm in general and how to handle edge cases, like what would happen if we have task based implementation and some task fails, it will be stateless, so the archive will be missed, also what would happen if group size changes for a root that is already created and worked with x group size. Doing this simpler and somewhat naive implementation allows for the algorithm to be seen in action and to be able to recognize how it can be made task based and how the considerations that come up can be solved. In a way this v1 can be considered as a prototype (predecessor) to the actual archiving component, rather a real v1.
- all pending items are captured with `todo`s, the implementation will be immediately addressed in #517 
- there are not tests because implementation will change with #517 which would require to change tests also

### DIFF
--- a/common/src/main/java/com/hedera/block/common/utils/MathUtilities.java
+++ b/common/src/main/java/com/hedera/block/common/utils/MathUtilities.java
@@ -1,10 +1,37 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.hedera.block.common.utils;
 
+import java.util.Arrays;
+
 /**
  * A utility class that deals with logic related to Mathematics.
  */
 public final class MathUtilities {
+    /**
+     * This array contains the powers of ten from 10^0 to 10^18 (long).
+     */
+    private static final long[] POSITIVE_POWERS_OF_TEN = {
+        1L,
+        10L,
+        100L,
+        1_000L,
+        10_000L,
+        100_000L,
+        1_000_000L,
+        10_000_000L,
+        100_000_000L,
+        1_000_000_000L,
+        10_000_000_000L,
+        100_000_000_000L,
+        1_000_000_000_000L,
+        10_000_000_000_000L,
+        100_000_000_000_000L,
+        1_000_000_000_000_000L,
+        10_000_000_000_000_000L,
+        100_000_000_000_000_000L,
+        1_000_000_000_000_000_000L
+    };
+
     /**
      * This method checks if the given number is a power of two.
      *
@@ -27,6 +54,30 @@ public final class MathUtilities {
      */
     public static boolean isEven(final int toCheck) {
         return (toCheck & 1) == 0;
+    }
+
+    /**
+     * This method checks if a given long is a positive power of 10.
+     * E.G.
+     * <pre>
+     *     isPowerOf10(1) = true
+     *     isPowerOf10(10) = true
+     *     isPowerOf10(100) = true
+     *     isPowerOf10(1000) = true
+     *     isPowerOf10(11) = false
+     *     isPowerOf10(50) = false
+     *     isPowerOf10(99) = false
+     *     isPowerOf10(101) = false
+     * </pre>
+     * @param toCheck the long to check
+     * @return {@code true} if the given long is a power of 10 {@code false} otherwise
+     */
+    public static boolean isPositivePowerOf10(final long toCheck) {
+        if (toCheck <= 0) {
+            return false;
+        } else {
+            return Arrays.binarySearch(POSITIVE_POWERS_OF_TEN, toCheck) >= 0;
+        }
     }
 
     private MathUtilities() {}
--- a/common/src/test/java/com/hedera/block/common/CommonsTestUtility.java
+++ b/common/src/test/java/com/hedera/block/common/CommonsTestUtility.java
@@ -253,5 +253,90 @@ public final class CommonsTestUtility {
         return Stream.concat(Stream.of(Arguments.of(0)), negativeIntegers());
     }
 
+    public static Stream<Arguments> positivePowersOf10() {
+        return Stream.of(
+                Arguments.of(1L),
+                Arguments.of(10L),
+                Arguments.of(100L),
+                Arguments.of(1_000L),
+                Arguments.of(10_000L),
+                Arguments.of(100_000L),
+                Arguments.of(1_000_000L),
+                Arguments.of(10_000_000L),
+                Arguments.of(100_000_000L),
+                Arguments.of(1_000_000_000L),
+                Arguments.of(10_000_000_000L),
+                Arguments.of(100_000_000_000L),
+                Arguments.of(1_000_000_000_000L),
+                Arguments.of(10_000_000_000_000L),
+                Arguments.of(100_000_000_000_000L),
+                Arguments.of(1_000_000_000_000_000L),
+                Arguments.of(10_000_000_000_000_000L),
+                Arguments.of(100_000_000_000_000_000L),
+                Arguments.of(1_000_000_000_000_000_000L));
+    }
+
+    public static Stream<Arguments> negativePowersOf10() {
+        return Stream.of(
+                Arguments.of(-1L),
+                Arguments.of(-10L),
+                Arguments.of(-100L),
+                Arguments.of(-1_000L),
+                Arguments.of(-10_000L),
+                Arguments.of(-100_000L),
+                Arguments.of(-1_000_000L),
+                Arguments.of(-10_000_000L),
+                Arguments.of(-100_000_000L),
+                Arguments.of(-1_000_000_000L),
+                Arguments.of(-10_000_000_000L),
+                Arguments.of(-100_000_000_000L),
+                Arguments.of(-1_000_000_000_000L),
+                Arguments.of(-10_000_000_000_000L),
+                Arguments.of(-100_000_000_000_000L),
+                Arguments.of(-1_000_000_000_000_000L),
+                Arguments.of(-10_000_000_000_000_000L),
+                Arguments.of(-100_000_000_000_000_000L),
+                Arguments.of(-1_000_000_000_000_000_000L));
+    }
+
+    public static Stream<Arguments> positiveIntPowersOf10() {
+        return Stream.of(
+                Arguments.of(1),
+                Arguments.of(10),
+                Arguments.of(100),
+                Arguments.of(1_000),
+                Arguments.of(10_000),
+                Arguments.of(100_000),
+                Arguments.of(1_000_000),
+                Arguments.of(10_000_000),
+                Arguments.of(100_000_000),
+                Arguments.of(1_000_000_000));
+    }
+
+    public static Stream<Arguments> negativeIntPowersOf10() {
+        return Stream.of(
+                Arguments.of(-1),
+                Arguments.of(-10),
+                Arguments.of(-100),
+                Arguments.of(-1_000),
+                Arguments.of(-10_000),
+                Arguments.of(-100_000),
+                Arguments.of(-1_000_000),
+                Arguments.of(-10_000_000),
+                Arguments.of(-100_000_000),
+                Arguments.of(-1_000_000_000));
+    }
+
+    // @todo(517) add 0 and MIN MAX values here as well, also, make the same logic to test for longs as well
+    public static Stream<Arguments> nonPowersOf10() {
+        return Stream.of(
+                Arguments.of(2),
+                Arguments.of(11),
+                Arguments.of(20),
+                Arguments.of(50),
+                Arguments.of(101),
+                Arguments.of(10_100));
+    }
+
     private CommonsTestUtility() {}
 }
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
@@ -1,0 +1,313 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.block.server.persistence.storage.archive;
+
+import com.hedera.block.common.utils.FileUtilities;
+import com.hedera.block.server.persistence.storage.PersistenceStorageConfig;
+import com.hedera.block.server.persistence.storage.path.BlockPathResolver;
+import com.hedera.block.server.persistence.storage.path.LiveBlockPath;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.lang.System.Logger.Level;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.TreeMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+// @todo(517) add documentation once we have the final implementation
+public final class BlockAsLocalFileArchiver implements LocalBlockArchiver {
+    private final BlockArchiverRunnable archiverRunnable;
+    private final ExecutorService executor;
+
+    private BlockAsLocalFileArchiver(
+            @NonNull final PersistenceStorageConfig config, @NonNull final BlockPathResolver blockPathResolver) {
+        this.archiverRunnable =
+                new BlockArchiverRunnable(Objects.requireNonNull(config), Objects.requireNonNull(blockPathResolver));
+        this.executor = Executors.newSingleThreadExecutor();
+        this.executor.submit(archiverRunnable);
+    }
+
+    public static BlockArchiver of(
+            @NonNull final PersistenceStorageConfig config, @NonNull final BlockPathResolver blockPathResolver) {
+        return new BlockAsLocalFileArchiver(config, blockPathResolver);
+    }
+
+    @Override
+    public void signalBlockWritten(final long latestBlockNumber) {
+        this.archiverRunnable.signalBlockWritten(latestBlockNumber);
+    }
+
+    @Override
+    public void stop() throws InterruptedException {
+        // @todo(517) this will change a bit once we move to task based solution
+        archiverRunnable.stop();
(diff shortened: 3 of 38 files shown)


common/src/main/java/com/hedera/block/common/utils/MathUtilities.java
```diff
@@ -1,10 +1,37 @@
...
 package com.hedera.block.common.utils;
 
+import java.util.Arrays;
+
 /**
  * A utility class that deals with logic related to Mathematics.
  */
 public final class MathUtilities {
+    /**
+     * This array contains the powers of ten from 10^0 to 10^18 (long).
```
jsync-swirlds, line 11: 10_000_000_000_000_000_000 is also valid for a long.

common/src/test/java/com/hedera/block/common/CommonsTestUtility.java
```diff
@@ -253,5 +253,89 @@ public static Stream<Arguments> zeroAndNegativeIntegers() {
...
+                Arguments.of(-1_000),
+                Arguments.of(-10_000),
+                Arguments.of(-100_000),
+                Arguments.of(-1_000_000),
+                Arguments.of(-10_000_000),
+                Arguments.of(-100_000_000),
+                Arguments.of(-1_000_000_000));
+    }
+
+    public static Stream<Arguments> nonPowersOf10() {
```
jsync-swirlds, line 330: Might be worth tossing in the extreme edges here
```
                Arguments.of(Long.MAX_VALUE),
                Arguments.of(Long.MIN_VALUE),
                Arguments.of(0),
```

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    @Override
+    public void signalBlockWritten(final long latestBlockNumber) {
+        this.archiverRunnable.signalBlockWritten(latestBlockNumber);
+    }
+
+    @Override
+    public void stop() throws InterruptedException {
+        // todo(517) this will change a bit once we move to task based solution
```
jsync-swirlds, line 58: For consistency, all of the `todo` items should be `@todo`.

It is possible to change what is considered "todo" in the IDE settings so only the properly formatted values highlight correctly; that can help when writing.

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockPathResolver.java
```diff
@@ -35,59 +35,7 @@ public interface BlockPathResolver {
...
-     * E.G. (illustrative example, actual path may vary):
-     * <pre>
-     *     If the blockNumber is 10, the resolved path will be:
-     *     <b>/path/to/archive/block/storage.zip/0000000000000000010.blk</b>
-     * </pre>
-     *
-     * @param blockNumber to be resolved the path for
-     * @return the resolved path to the given Block by a number
-     * @throws IllegalArgumentException if the blockNumber IS NOT a whole number
-     */
```
jsync-swirlds, line 59: Why was the documentation removed?
At a minimum, the next PR should definitely include API specifications at least as thorough as what was removed.

jsync-swirlds (dismissed):
Approved as an _interim_ PR only (as requested).
Lots of things (already noted via `@todo` in most case) that need considerable work in the next PR.

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
```diff
@@ -0,0 +1,317 @@
...
+        // @todo(517) this will be removed, if we are in a task based solution, we will not need this probably
+        public void stop() {
+            running = false;
+            threadSignalCarrier.doNotify();
+        }
+    }
+
+    /**
+     * todo(517) remove this, this is a temporary solution to v1 of archiving, which will change to task based implementation
+     */
```
AlfredoG87, line 264: This temporary solution seems way more convoluted to me than a task based impl. waiting to see how much simpler will be :)

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
```diff
@@ -0,0 +1,317 @@
...
+            }
+        }
+
+        // @todo(517) this will be improved
+        public Path resolveArchivePathForZipOfBlockNumber(long blockNumber, int batchSize, Path archiveRootPath) {
+            // Calculate the batch start number
+            long batchStartNumber = (blockNumber / batchSize) * batchSize;
+
+            // Format the batch start number to a 19-digit string
+            String formattedBatchStartNumber = String.format("%0" + 19 + "d", batchStartNumber);
```
AlfredoG87, line 231: are the number of digits for the file `mask` configurable somewhere? or at least a Constant?

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
```diff
@@ -0,0 +1,317 @@
...
+    public void signalBlockWritten(final long latestBlockNumber) {
+        this.archiverRunnable.signalBlockWritten(latestBlockNumber);
+    }
+
+    @Override
+    public void stop() throws InterruptedException {
+        // todo(517) this will change a bit once we move to task based solution
+        archiverRunnable.stop();
+        executor.shutdown();
+        final boolean terminatedGracefully = this.executor.awaitTermination(60, TimeUnit.SECONDS);
```
AlfredoG87, line 61: why `60` Seconds?? 

this might need to be a configurable property.

AlfredoG87 (dismissed):
Can't wait to see the task based approach!
This is a good start, but does indeed need a refactor to reduce complexity on Task based approach.  Approving for now and leaved some comments of changes that can be considered on the next PR.

common/src/test/java/com/hedera/block/common/CommonsTestUtility.java
ata-nas: @jsync-swirlds thanks, will make sure to add to the follow up pr.

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockPathResolver.java
ata-nas: The method, associated with this documentation was removed, hence the documentation also.

server/src/main/java/com/hedera/block/server/persistence/storage/archive/LocalBlockArchiver.java
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.block.server.persistence.storage.archive;
+
+/**
+ *
+ * todo(517) maybe we will not need this interface
```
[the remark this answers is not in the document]
ata-nas: @AlfredoG87 as we already spoken, there were a lot of unknowns, hence we have this v1 prototype of implementation that covers use cases that came up during the design and development of the initial archiving logic. This being said, this interface was initially added to be consistent with other abstractions present in the persistence, but it might as well be unnecessary, will see in the follow up pr.

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
```diff
@@ -0,0 +1,317 @@
...
+                        }
+                    }
+                }
+
+                for (final Path folder : pathsToArchive.keySet()) {
+                    final Path livePathSymlink = Path.of(folder.toString().replace("archive", "live"));
+                    Files.createSymbolicLink(livePathSymlink, folder);
+                    LOGGER.log(Level.INFO, "Created symlink: {0}", livePathSymlink);
+                    final Path toDelete = Path.of(livePathSymlink.toString().replace(".zip", ""));
+                    LOGGER.log(Level.INFO, "Deleting: {0}", toDelete);
```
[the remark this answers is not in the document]
ata-nas: @AlfredoG87 these will be removed, I added a lot of logging in order to showcase the algorithm and to see the archiver working, these I forgot to remove, but anyhow, logging will be different with the task based implementation.

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
ata-nas: @AlfredoG87 indeed, but this naive approach for example solves very easily some use cases. For example, what happens if an archive round fails with task based? Current approach picks up the last archived (for brevity reasons, this is done in memory now, just to showcase it, but in previous commit, there is also the logic that goes to the archive root and searches last archived, this is super performant due to the trie structure, we simply DFS into the highest possible dir number until we hit the first zip and then we can see which is the highest entry number inside, worst case is a for loop with 18 iterations - max depth, the more files we have the faster it is, same applies for searching the live root if we so choose so for the naive approach, but we have no zip there) and goes from there, where the task based will be given a block number (threshold) and it will find the dir where the given block number -1 resides and will archive everything under that root. So handling missed archives needs to be done differently, most likely with another process that will be run periodically and will be detecting missed archives and then attempting them again.

There are more considerations, I simply name one just to showcase some of the things that came up during the development of the v1, things which cannot really be foreseen before we have some code.

Also consider, that recent developments now will require a different approach, where logic that denotes blocks incoming will be done by the persistence handler (block writers will change also) and the persistence handler will be the decision maker as to what shall be done (persist a block, archive a group etc. etc.). This also gives us the opportunity to have a different approach to pass the current passed threshold number to the new task based archiver.

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
ata-nas: @AlfredoG87 an arbitrary number, that being said, we should probably discuss the shutdown logic for the task based implementation, it is not used now.

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
ata-nas: @AlfredoG87 this logic will probably not be present here, since the archive will no longer need to do that, it will receive a block number at passed threshold and it will archive everything under that directory.

common/src/test/java/com/hedera/block/common/CommonsTestUtility.java
ata-nas: Thanks for this, I have added todo since if I add these longs now the parameterized test will fail (it requires ints). I will extend this logic to also have the checks for long and will add these there.

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
ata-nas: ✅ Done.

ata-nas:
@jsync-swirlds @AlfredoG87 thanks for your reviews. I have captured your findings, will be addressed in #517 together with the change to task based archiving. I hope the comments are resolved on your side.

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    @Override
+    public void stop() throws InterruptedException {
+        // @todo(517) this will change a bit once we move to task based solution
+        archiverRunnable.stop();
+        executor.shutdown();
+        final boolean terminatedGracefully = this.executor.awaitTermination(60, TimeUnit.SECONDS);
+        if (!terminatedGracefully) {
+            // todo log or do something
```
mattp-swirldslabs, line 61: I've seen other comments about using `@todo`. Maybe here too ^^

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
```diff
@@ -0,0 +1,313 @@
...
+                }
+
+                pathsToArchive = pathsToArchive.entrySet().stream()
+                        .filter(entry -> entry.getValue().size() == archiveBatchSize
+                                || entry.getKey()
+                                        .toString()
+                                        .endsWith("/0".repeat(19 - (int) Math.log10(archiveBatchSize)) + ".zip"))
+                        .collect(
+                                Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, TreeMap::new));
+
```
mattp-swirldslabs, line 178: Will the task-based solution separate the logic of what's to be archived from how it is archived? This zip logic could be behind another interface that this `archive()` method delegates to.

mattp-swirldslabs (approved):
Looking forward to seeing the v2.

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockPathResolver.java
jsync-swirlds: We need to create equivalent documentation when we change methods; it looks like the replacement method has very little API specification.

Not for this PR, but a note of recommendation for the next one.

jsync-swirlds (dismissed):

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
ata-nas: @mattp-swirldslabs indeed, the archiver will make no decision as to what will be archived, this will be supplied to the task, current thought is, the block number of the latest block written that passes the configured threshold. From there, since we have the trie structure, we can determine which dir we should archive by doing a simple integer division. Everything under that directory (whatever may be in there) will be archived. The trie structure is perfect for avoiding any conflicts. This also opens up the possibility to run multiple tasks in parallel w/o worrying about conflicts, in theory let's say we will archive 1000s and 2000s, the folders are in the same level in the trie, so everything inside the 1000s folder is 0-999 and in the 2000s folder is 1000-1999. Let's see how it will come up :)

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
ata-nas: Seem to have missed that, I would keep this one like so, otherwise re-approvals are needed. Hope it is resolved for you.

server/src/main/java/com/hedera/block/server/persistence/storage/archive/BlockAsLocalFileArchiver.java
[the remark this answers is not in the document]
mattp-swirldslabs: If the logging is useful during development, consider changing it to DEBUG (and maybe wrapping in `LOGGER.isLoggable(DEBUG)` if it's compute-intensive) and leaving it off in `logging.properties`. When we start going into production, we may need this kind of info to troubleshoot such a new system.

AlfredoG87 (approved):

PR #497, opened 2025-01-14
chore: Bump version of apps to 0.4.0-SNAPSHOT

AlfredoG87:
**Description**:
Manual PR to bump version of Main to `0.4.0-SNAPSHOT` since we are already releasing `0.3.0`

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


gradle.properties
```diff
@@ -1,3 +1,6 @@
+# Version number
+version=0.4.0-SNAPSHOT
+
```
jjohannes, line 3: The version is now defined in `version.txt`. Please update it there. 

`gradle.properties` is now only used for technical Gradle settings and usually does not need to be changed anymore.

A full overview of the build related files can be found here: https://github.com/hiero-ledger/hiero-gradle-conventions?tab=readme-ov-file#project-structure

AlfredoG87: ✅ Done!

Thank you for the prompt revision and explanation, I will do a follow up PR to fix the Release Process that is currently broken due to this change. But this PR is just for updating the Snapshot version.

jjohannes: Thanks. I was not aware of this part of the release process here. I will take a look at the other PR. We may be able to simplify this a bit.

jjohannes (approved):

jjohannes (changes requested):
There are two more `version: 0.3.0-SNAPSHOT` entries in the YAML files that should be updated as well.

AlfredoG87:
> There are two more version: 0.3.0-SNAPSHOT entries in the YAML files that should be updated as well.

@jjohannes Thank you for noticing!! 😅 

I already have another PR to fix the automated task using gradle, so we don't have to worry too much on doing it manually and forgetting something.

jjohannes (approved):

mattp-swirldslabs (approved):

PR #501, opened 2025-01-15
feat: centralize configuration logging

### LINKED ISSUE #471: Centralize Record Configuration Logging

### LINKED ISSUE #474: Implement Log Safety Mechanism

mattp-swirldslabs:
**Description**:

**Related issue(s)**:

Fixes #471 
Fixes #474 

**Notes for reviewer**:

**Checklist**


AlfredoG87 (approved):
LGTM, nice work.

AlfredoG87:
@mattp-swirldslabs, this should be part of Milestone `0.4.0` already moved it.

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
ata-nas, line 1: Some thoughts on the config logger sanity test:

Back when the sanity test for the environment variable mappings was added, there were a couple of considerations that came up and had to be addressed:
1. There are some config classes that are under our control and some that are not.
1. By default, the sanity test should fail, unless the proper conditions are met OR we have explicitly filtered out something (be it under our control or not).
1. The sanity test should catch changes that are made to the source, no matter if we have changed a config class that we control, or someone else has changed a config class that we do not control, be that the change is addition, removal or modification of any field/fields. We should be able to immediately see a failing test.

Having this philosophy, the sanity test "mimics" the internal config of the production class (that builds the env var mapping). The reason is to be explicit in declaring, hey, I have added/removed/modified a field, yes I am absolutely sure I want to do that, this is why I am changing the sanity test to reflect that modification and it now no longer fails. And of course, if the diff is not generated by us, the test will fail, notifying us of a change someone else has made of a config we use.

I think that the new addition about the logging should do the same, by default, we should always fail, until we explicitly confirm that we want to log something, or to not log it. Moreover, I think we should be able to filter loggable/not loggable fields, not only classes. This gives us greater versatility.

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
```diff
@@ -110,29 +113,50 @@ void testVerifyAllFieldsInRecordsAreMapped(
...
+                    .isPresent();
+        }
+    }
+
+    private void verifyLoggableAnnotation(
+            final RecordComponent recordComponent,
+            final String fieldName,
+            final String configClassName,
+            final String prefix) {
+        if (!recordComponent.isAnnotationPresent(Loggable.class) && !LOGGABLE_PACKAGES.contains(prefix)) {
```
ata-nas, line 153: Not checking or disregarding a whole config class is not the same as checking fields. I think these two are mixed here.

server/src/main/java/com/hedera/block/server/config/logging/ConfigurationLoggingImpl.java
```diff
@@ -0,0 +1,113 @@
...
+    private final Set<String> loggablePackages = new HashSet<>();
+
+    @Inject
+    public ConfigurationLoggingImpl(@NonNull final Configuration configuration) {
+        this.configuration = requireNonNull(configuration);
+
+        // The configuration properties in these packages
+        // are out of our control. So allow them to be logged.
+        loggablePackages.add("metrics");
+        loggablePackages.add("prometheus");
```
ata-nas, line 33: I believe the loggable packages should be a `Map<Class<? extends Record>, List<String>>` which would be much more reliable, since if a config class is changed (renamed/removed) we will have a compile error, which would be nice. Also if the string value of the `@ConfigData` can change and we will not detect this immediately since the code will compile. I think the loggable packages are made to say "hey, please log that class which I do not support myself, but use from someone else's library". Since we have no way to put the `@Loggable` annotation on these, we can emulate the same with the list which would be denoted by the class in the map, where we can put the field name, which should be logged. In this way we are just as explicit as with the `@Loggable` annotation. And we can then make sure that the sanity test will be able to do it's function which should be an explicit confirmation of, I want/don't want to log something, failing by default.

server/src/test/java/com/hedera/block/server/mediator/MediatorConfigTest.java
```diff
@@ -29,14 +29,14 @@ public void testMediatorConfig_powerOf2Values() {
...
+            MediatorConfig mediatorConfig = new MediatorConfig(powerOf2Value, null);
             assertEquals(powerOf2Value, mediatorConfig.ringBufferSize());
         }
 
         // Test the non-power of 2 values
         for (int powerOf2Value : powerOf2Values) {
             IllegalArgumentException exception =
-                    assertThrows(IllegalArgumentException.class, () -> new MediatorConfig(powerOf2Value + 1, ""));
+                    assertThrows(IllegalArgumentException.class, () -> new MediatorConfig(powerOf2Value + 1, null));
             assertEquals("Mediator Ring Buffer Size must be a power of 2", exception.getMessage());
```
ata-nas, line 40: These should really be parameterized tests and be two separate tests, we should not be mixing passing/failing conditions of a test in a single method, I would make exceptions if we have a method override which we expect the core logic to be the same, but something like a log statement changes. (outside of the scope of the issue, but just pointing out, we could cleanup in the future, we could make some issues to not forget these).

server/src/test/java/com/hedera/block/server/verification/VerificationInjectionModuleTest.java
```diff
@@ -48,7 +50,7 @@ void testProvideBlockVerificationService_disabled() throws IOException {
     @Test
     void testProvidesBlockVerificationSessionFactory() {
         // given
-        VerificationConfig verificationConfig = new VerificationConfig(true, BlockVerificationSessionType.ASYNC, 32);
+        VerificationConfig verificationConfig = new VerificationConfig(null, ASYNC, 32);
```
ata-nas, line 53: Why null for type? should we not check non null values for the enums?

server/src/test/java/com/hedera/block/server/verification/session/BlockVerificationSessionFactoryTest.java
```diff
@@ -55,7 +55,7 @@ void createSession_whenSessionTypeIsAsync_returnsBlockVerificationSessionAsync()
     @Test
     void createSession_whenSessionTypeIsSync_returnsBlockVerificationSessionSync() {
         // Given
-        VerificationConfig config = new VerificationConfig(true, BlockVerificationSessionType.SYNC, 32);
+        VerificationConfig config = new VerificationConfig(null, BlockVerificationSessionType.SYNC, 32);
```
ata-nas, line 58: Same as above comment for null check of enums.

ata-nas (commented):
Good job @mattp-swirldslabs! the logging looks really good!
I do think however that some changes are due in the sanity test and in the way we handle config classes we do not own. Also some optional nits and cleanup suggested.

server/src/main/java/com/hedera/block/server/config/logging/ConfigurationLoggingImpl.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public void log() {
+
+        if (LOGGER.isLoggable(INFO)) {
```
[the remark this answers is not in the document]
mattp-swirldslabs: "Added a commented out property in logging.properties showing how to switch OFF the output"

server/src/main/java/com/hedera/block/server/consumer/ConsumerConfig.java
```diff
@@ -23,8 +23,5 @@ public record ConsumerConfig(@ConfigProperty(defaultValue = "1500") long timeout
         if (timeoutThresholdMillis <= 0) {
```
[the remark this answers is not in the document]
mattp-swirldslabs: Yeah, the Preconditions didn't exist when this was first written. I replaced the check here.

server/src/main/java/com/hedera/block/server/config/logging/Loggable.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.RECORD_COMPONENT})
+public @interface Loggable {
+    String value() default "";
```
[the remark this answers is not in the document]
mattp-swirldslabs: Good catch. Fixed

server/src/main/java/com/hedera/block/server/mediator/MediatorConfig.java
```diff
@@ -19,9 +18,8 @@
 // 131072 works but not with persistence
 @ConfigData("mediator")
 public record MediatorConfig(
-        @ConfigProperty(defaultValue = "4194304") int ringBufferSize,
-        @ConfigProperty(defaultValue = "PRODUCTION") String type) {
-    private static final System.Logger LOGGER = System.getLogger(MediatorConfig.class.getName());
+        @Loggable @ConfigProperty(defaultValue = "4194304") int ringBufferSize,
```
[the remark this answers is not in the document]
mattp-swirldslabs: Changed to an underscore embedded numeric literal

server/src/main/java/com/hedera/block/server/config/logging/ConfigurationLoggingImpl.java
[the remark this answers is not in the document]
ata-nas: @mattp-swirldslabs can we not use `Logger.Level.OFF` (I think that was how it was called?

server/src/test/java/com/hedera/block/server/verification/VerificationInjectionModuleTest.java
ata-nas: We should not be having a null by default I think. That being said, it is ok to do this another time.

server/src/main/java/com/hedera/block/server/config/logging/ConfigurationLoggingImpl.java
mattp-swirldslabs: Good suggestion. I think the type-safety here is worth it. However, I'm not sure I understand the usage of a List<String> in this context. Take a look at the change I made. By matching on the record type, all fields should be allowed to log.

ata-nas: So, first off, you could just use for example `PrometheusConfig.class` directly instead of getting the instance from the configuration, you only need the class really. Then the reason for the list is to be able to emulate adding the `@Loggable` annotation to classes we are managing. By explicitly adding field names (strings) to the list bound to a config class we do not controll, it would be the same as adding the `@Loggable` annotation to a config class we are controlling. And so when you iterate over the map, and let's say you are currently on the entry that will give you the list of fieldnames to log for `PrometheusConfig`, you can then take the list and log exactly what we have declared to log, cause imagine there could be things that should not be logged in config classes not managed by us. That's my thought process.

server/src/main/java/com/hedera/block/server/service/ServiceConfig.java
```diff
@@ -9,22 +10,19 @@
...
     /**
      * Validate the configuration.
      *
      * @throws IllegalArgumentException if the configuration is invalid
      */
     public ServiceConfig {
-        if (delayMillis <= 0) {
+        if (shutdownDelayMillis <= 0) {
             throw new IllegalArgumentException("Delay milliseconds must be greater than 0");
         }
```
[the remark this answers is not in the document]
mattp-swirldslabs: done

server/src/main/java/com/hedera/block/server/config/logging/ConfigurationLoggingImpl.java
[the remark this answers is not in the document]
mattp-swirldslabs: I thought about this more and added checks for DEBUG and TRACE since those are akin to INFO

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
mattp-swirldslabs: My assumptions are: 1) 3rd party config will never contain a sensitive value 2) we want to see all available properties from 3rd party config.

ata-nas: I would agree to that, but I personally cannot trust that such a thing would not happen. If something could happen, it probably will, streched in the long run. But I think the team could also give an opinion on that. I do agree that we should log all properties of the 3rd party (except sensitive, if there are any).
@AlfredoG87 @georgi-l95 @jasperpotts @jsync-swirlds what do you guys think on that point?

jsync-swirlds: Plugins should be completely responsible for their own configuration.  Plugins _we write_ may use the platform library configuration (and should annotate sensitive values), but I would assume that third party plugins will _not_ do so, and probably should not even have the _option_ to do so.

The more "severable" the plugins are (in every respect, including logging, configuration, etc...), the closer we get to the stated requirements.  Obviously we cannot afford the time to reach a "perfect" state, but getting closer within reasonable time limits is a good goal.

Note also, I would recommend that we try to have _just enough_ configuration, and no more.  Part of what is driving this desire to print out the whole config is that we probably have _too many_ configurable options (certainly using proper service loading will reduce that quite a bit).

server/src/test/java/com/hedera/block/server/mediator/MediatorConfigTest.java
mattp-swirldslabs: Might be fixed in another PR

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
```diff
@@ -12,24 +12,25 @@
...
             new ConfigMapping("prometheus.endpointPortNumber", "PROMETHEUS_ENDPOINT_PORT_NUMBER"),
-            new ConfigMapping("verification.enabled", "VERIFICATION_ENABLED"),
-            new ConfigMapping("verification.sessionType", "VERIFICATION_SESSION_TYPE"),
-            new ConfigMapping("verification.hashCombineBatchSize", "VERIFICATION_HASH_COMBINE_BATCH_SIZE"));
+            new ConfigMapping("server.maxMessageSizeBytes", "SERVER_MAX_MESSAGE_SIZE_BYTES"),
+            new ConfigMapping("server.port", "SERVER_PORT"),
+            new ConfigMapping("service.shutdownDelayMillis", "SERVICE_SHUTDOWN_DELAY_MILLIS"),
+            new ConfigMapping("verification.type", "VERIFICATION_TYPE"),
+            new ConfigMapping("verification.hashCombineBatchSize", "VERIFICATION_HASH_COMBINE_BATCH_SIZE"),
+            new ConfigMapping("verification.sessionType", "VERIFICATION_SESSION_TYPE"));
```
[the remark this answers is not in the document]
mattp-swirldslabs: I changed the ordering in the sanity test to be alphabetical.

server/src/main/java/com/hedera/block/server/config/logging/ConfigurationLoggingImpl.java
```diff
@@ -0,0 +1,127 @@
...
+        loggablePackages.add(configuration.getConfigData(PrometheusConfig.class));
+    }
+
+    /**
+     * Log the configuration data.
+     */
+    @Override
+    public void log() {
+
+        if (LOGGER.isLoggable(INFO) || LOGGER.isLoggable(DEBUG) || LOGGER.isLoggable(TRACE)) {
```
jsync-swirlds, line 54: Why include DEBUG and TRACE here? If a DEBUG or TRACE message is loggable, so is INFO.
In essence if any of the three is true, the first is true.

mattp-swirldslabs: I made the change to INFO checking only based on my conversation with DevOps. They're expecting to see configuration at INFO level as is the precedent with the CN

AlfredoG87 (approved):
LGTM.

@mattp-swirldslabs @ata-nas @jsync-swirlds let's try to capture nits, improvements and suggestions on a follow-up ticket, and move forward with this, we cannot afford to spend any more time on some of the nits am reading here. This is way better than what we used to have (no logging or visibility of configuration loaded by the system).

My suggestions is to merge as currently is this PR.

ata-nas:
> LGTM.
> 
> @mattp-swirldslabs @ata-nas @jsync-swirlds let's try to capture nits, improvements and suggestions on a follow-up ticket, and move forward with this, we cannot afford to spend any more time on some of the nits am reading here. This is way better than what we used to have (no logging or visibility of configuration loaded by the system).
> 
> My suggestions is to merge as currently is this PR.

I would agree to that, @mattp-swirldslabs please capture/acknowledge nits/considerations (which are not even final, could be or not be implemented in the long run, mind you there are already planned changes to the configuration with the query endpoint) and we can proceed.

server/src/test/java/com/hedera/block/server/verification/VerificationInjectionModuleTest.java
mattp-swirldslabs: Ok, we can take a look at this in the future.

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
mattp-swirldslabs: I'm unclear if the team is asking for a change here or if I'm clear to resolve this. Is it ok to resolve this?

server/src/main/java/com/hedera/block/server/config/logging/ConfigurationLoggingImpl.java
mattp-swirldslabs: Would it be ok to close this now and bring up another ticket if we need finer-grained controls around what gets emitted from 3rd party configs?

ata-nas: @mattp-swirldslabs agreed on that. Let's capture this in a follow up. Resolved for me.

server/src/test/java/com/hedera/block/server/verification/VerificationInjectionModuleTest.java
ata-nas: Agreed on that.

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
ata-nas: I think we can capture in the follow up. For me it is resolved, but let's see if the follow up shall need this change or not as well.

jsync-swirlds: Perhaps just note the concerns raised (and options offered) in a "for future work" issue, and finish off this PR as-is?

ata-nas: Resolved for me.

ata-nas (approved):
🔥

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
ata-nas: @mattp-swirldslabs please note this comment for the follow up prs.

PR #548, opened 2025-01-28
feat: add support for `PublishBlockStream` to simulator

### LINKED ISSUE #430: Add support for PublishBlockStream to simulator
As a Consensus Node Engineer
I want to use the Block Node Simulator to test my Consensus Node
So That I can continuously validate the communication protocol is correct.

### Notes
Teams working on the Consensus Node software have asked if the simulator can be used to simulate a Block Node for use in testing the Consensus Node.  This is a good idea, and we should ensure that they can do so.  We know there is already a small disconnect in protocol due to differences in the protobuf definitions (partly caused by having 3 different "authoritative" sources, plus semi-authoritative sources), and using the block node simulator could help ensure those issues are caught and handled early.

georgi-l95:
**Description**:
This PR adds support for `PublishBlockStream` in the simulator, by introducing new wokring mode, called PublisherServerMode, where the application acts a server, which listents and awaits stream of block data. With the effort here I aim mainly to introduce the feature, where in future ones we will refine it, according to needs to be able to use it as testing driver.

**Related issue(s)**:

Fixes #430 

**Notes for reviewer**:

**Checklist**


simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamGrpcServerImpl.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.List;
+import javax.inject.Inject;
+
+/**
+ * Implementation of {@link PublishStreamGrpcServer} that handles incoming block stream publications
+ * via gRPC streaming. This implementation manages the server setup, handles incoming block streams,
+ * tracks processed blocks, and maintains a history of stream statuses. It provides functionality
+ * to start, monitor, and shutdown the gRPC server.
+ */
+public class PublishStreamGrpcServerImpl implements PublishStreamGrpcServer {
```
georgi-l95, line 31: All implementations could be unified under one interface, that means that we close the idea of starting couple of implementations under one instance. Which is okay, judging by the current and future needs.

simulator/src/main/java/com/hedera/block/simulator/grpc/impl/PublishStreamServerObserver.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Processes incoming publish stream requests, maintaining a history of requests and handling block acknowledgements.
+     * If the request contains block items with a block proof, generates and sends an acknowledgement response.
+     *
+     * @param publishStreamRequest The incoming stream request to process
+     */
+    @Override
+    public void onNext(PublishStreamRequest publishStreamRequest) {
```
georgi-l95, line 68: In future PRs we will add handlers here to respond, depending on predefined configuration or depending on controller (don't exist yet). For now we have only one handler, which returns block ack.

simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulatorApp.java
```diff
@@ -148,8 +134,10 @@ public void stop() throws InterruptedException {
     public StreamStatus getStreamStatus() {
         return StreamStatus.builder()
                 .publishedBlocks(publishStreamGrpcClient.getPublishedBlocks())
+                .processedBlocks(publishStreamGrpcServer.getProcessedBlocks())
```
georgi-l95, line 137: If all grpc implementations are unifed under one interface, this could be improved and reduced.

ata-nas (approved):
LGTM. Nice Job!

PR #555, opened 2025-01-29
feat: added async handling for live consumers

### LINKED ISSUE #520: Decouple Consumer Streaming Components from RingBuffer Threads
As a Block Node Operator
I want the Consumer streaming connections to be decoupled from the Mediator RingBuffer threads
So that slow consumption by one streaming connection does not affect all other consumers

The mediator RingBuffer implementation leverages a pool of worker threads. When a consumer connects to the BN server, they're subscribed to the RingBuffer. As batches of blockitems arrive from a producer, they are broadcast to all subscribers. At a low level, this means the reference to each batch is put into the RB data structure and RB worker threads call the `onEvent()` with each batch reference. In doing so, the worker thread executes the code inside each of the consumer observers, the Helidon plugin and, ultimately the Helidon code to put each batch on the wire. In the worst case, if the RingBuffer worker thread is captured by Helidon, you can end up with [issues like this](https://github.com/hashgraph/pbj/issues/329). As the number of consumers scales up, all consumers will see degraded performance as the RB workers struggle to push blockitems out.

Instead, each subscriber should leverage it's own, independent thread. In this way, a RB worker thread simply publishes the blockitem reference to the consumer and then returns to the RB to continue work. A dedicated async worker thread on each consumer then handles picking up blockitems from it's queue and executing all the code to get the blockitems on the wire. Consequently, each consumer will operate independently. One slow consumer will not incur drag on all other consumers.

### LINKED ISSUE #549: StreamPersistenceHandlerImpl ERROR log does not print stacktrace
### Description

See the reproduction steps below

### Steps to reproduce

Steps to Reproduce:
1) Start the BN container with docker compose
2) Start the Producer simulator running the 10k TPS recording
3) Verify some block were persisted via the dashboard
4) Stop the simulator
5) Without recycling the container, restart the simulator to resend blocks that were already saved

The container stopped as an exception was thrown. There was an ERROR log:

```
2025-01-28 14:14:19 SEVERE  [com.hedera.block.server.persistence.StreamPersistenceHandlerImpl onEvent] Failed to persist BlockItems:
```
This message did not contain the stacktrace to find where the exception was thrown.

` LOGGER.log(ERROR, "Failed to persist BlockItems: ", e);`

Here's the line in the code: https://github.com/hashgraph/hedera-block-node/blob/e1f7245c5948d8f5662bb1a901cd055198a79b1f/server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java#L123

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.4.0

### Operating system

None

mattp-swirldslabs:
**Description**:

* Added `AsyncConsumerStreamResponseObserver` injected with an `ExecutorService` to run the Callable `Functions.Task.call()` for each consumer created in `PbjBlockStreamServiceProxy`
* The current implementation of the `ExecutorService` is `Executors.newVirtualThreadPerTaskExecutor()` since it was the highest performing pooled Executor.
* Added new dashboard row called "Live Block Latency" to track how many blocks the live consumers are behind
* Added 2 new socket config properties to ServerConfig for experimental perf tuning later
* Refactored ServerConfig to leverage Preconditions
* Small change to `NoOpBlockVerificationService` to remain quiet when running in NO_OP mode
* Reduced the `mediator` RingBuffer default size. The `AsyncConsumerStreamResponseObserver` now buffer block items so we should not need as much RB space.

**Related issue(s)**:

Fixes #520 
Fixes #549 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/server/src/main/java/com/hedera/block/server/BlockNodeApp.java
+++ b/server/src/main/java/com/hedera/block/server/BlockNodeApp.java
@@ -14,6 +14,7 @@ import com.hedera.block.server.service.ServiceStatus;
 import com.hedera.pbj.grpc.helidon.PbjRouting;
 import com.hedera.pbj.grpc.helidon.config.PbjConfig;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import io.helidon.webserver.ConnectionConfig;
 import io.helidon.webserver.WebServer;
 import io.helidon.webserver.WebServerConfig;
 import io.helidon.webserver.http.HttpRouting;
@@ -88,12 +89,18 @@ public class BlockNodeApp {
                 .maxMessageSizeBytes(serverConfig.maxMessageSizeBytes())
                 .build();
 
+        final ConnectionConfig connectionConfig = ConnectionConfig.builder()
+                .sendBufferSize(serverConfig.socketSendBufferSizeBytes())
+                .receiveBufferSize(serverConfig.socketSendBufferSizeBytes())
+                .build();
+
         // Build the web server
         final WebServer webServer = webServerBuilder
                 .port(serverConfig.port())
                 .addProtocol(pbjConfig)
                 .addRouting(pbjRouting)
                 .addRouting(httpRouting)
+                .connectionConfig(connectionConfig)
                 .build();
 
         // Update the serviceStatus with the web server
--- a/server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
+++ b/server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
@@ -42,8 +42,8 @@ public final class ServerMappedConfigSourceInitializer {
 
             // Server Config
             new ConfigMapping("server.maxMessageSizeBytes", "SERVER_MAX_MESSAGE_SIZE_BYTES"),
-            new ConfigMapping("server.socketSendBufferSizeBytes", "SOCKET_SEND_BUFFER_SIZE_BYTES"),
-            new ConfigMapping("server.socketReceiveBufferSizeBytes", "SOCKET_RECEIVE_BUFFER_SIZE_BYTES"),
+            new ConfigMapping("server.socketSendBufferSizeBytes", "SERVER_SOCKET_SEND_BUFFER_SIZE_BYTES"),
+            new ConfigMapping("server.socketReceiveBufferSizeBytes", "SERVER_SOCKET_RECEIVE_BUFFER_SIZE_BYTES"),
             new ConfigMapping("server.port", "SERVER_PORT"),
 
             // Service Config
--- a/server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
+++ b/server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
@@ -11,7 +11,7 @@ import java.util.List;
  * A class that extends {@link MappedConfigSource} ir order to have project-relevant initialization.
  */
 public final class ServerMappedConfigSourceInitializer {
-    private static final List<ConfigMapping> MAPPINGS = List.of(
+    static final List<ConfigMapping> MAPPINGS = List.of(
             // Please add properties in alphabetical order
 
             // Consumer Config
@@ -42,6 +42,8 @@ public final class ServerMappedConfigSourceInitializer {
 
             // Server Config
             new ConfigMapping("server.maxMessageSizeBytes", "SERVER_MAX_MESSAGE_SIZE_BYTES"),
+            new ConfigMapping("server.socketSendBufferSizeBytes", "SOCKET_SEND_BUFFER_SIZE_BYTES"),
+            new ConfigMapping("server.socketReceiveBufferSizeBytes", "SOCKET_RECEIVE_BUFFER_SIZE_BYTES"),
             new ConfigMapping("server.port", "SERVER_PORT"),
 
             // Service Config
--- a/server/src/main/java/com/hedera/block/server/consumer/AsyncConsumerStreamResponseObserver.java
+++ b/server/src/main/java/com/hedera/block/server/consumer/AsyncConsumerStreamResponseObserver.java
@@ -8,8 +8,9 @@ import com.hedera.block.server.mediator.SubscriptionHandler;
 import com.hedera.hapi.block.SubscribeStreamResponseUnparsed;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Objects;
+import java.util.concurrent.CompletionService;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 
 /**
  * The AsyncConsumerStreamResponseObserver class is responsible for decoupling the
@@ -19,25 +20,24 @@ import java.util.concurrent.ExecutorService;
 class AsyncConsumerStreamResponseObserver
         implements BlockNodeEventHandler<ObjectEvent<SubscribeStreamResponseUnparsed>> {
 
-    private final ExecutorService executorService;
+    private final CompletionService<Void> completionService;
     private final SubscriptionHandler<SubscribeStreamResponseUnparsed> subscriptionHandler;
     private final BlockNodeEventHandler<ObjectEvent<SubscribeStreamResponseUnparsed>> nextBlockNodeEventHandler;
 
     /**
      * Constructor for the AsyncConsumerStreamResponseObserver class.
      *
-     * @param executorService the executor service to use for asynchronous processing
      * @param subscriptionHandler the handler for managing subscriptions
      * @param nextBlockNodeEventHandler the next block node event handler in the chain
      */
     public AsyncConsumerStreamResponseObserver(
-            @NonNull final ExecutorService executorService,
+            @NonNull final CompletionService<Void> completionService,
             @NonNull final SubscriptionHandler<SubscribeStreamResponseUnparsed> subscriptionHandler,
             @NonNull
                     final BlockNodeEventHandler<ObjectEvent<SubscribeStreamResponseUnparsed>>
                             nextBlockNodeEventHandler) {
 
-        this.executorService = Objects.requireNonNull(executorService);
+        this.completionService = Objects.requireNonNull(completionService);
         this.subscriptionHandler = Objects.requireNonNull(subscriptionHandler);
         this.nextBlockNodeEventHandler = Objects.requireNonNull(nextBlockNodeEventHandler);
     }
@@ -50,9 +50,13 @@ class AsyncConsumerStreamResponseObserver
             @NonNull final ObjectEvent<SubscribeStreamResponseUnparsed> event, final long l, final boolean b) {
 
         try {
-            executorService
-                    .submit(new Task(event, l, b, subscriptionHandler, this, nextBlockNodeEventHandler))
-                    .get();
+            completionService.submit(new Task(event, l, b, subscriptionHandler, this, nextBlockNodeEventHandler));
+
+            Future<Void> future = completionService.poll(); // Non-blocking check
+            if (future != null) {
+                future.get();
+            }
+
         } catch (ExecutionException | InterruptedException e) {
             Throwable cause = e.getCause();
             if (cause instanceof IllegalArgumentException) {
--- a/server/src/main/java/com/hedera/block/server/consumer/AsyncConsumerStreamResponseObserver.java
+++ b/server/src/main/java/com/hedera/block/server/consumer/AsyncConsumerStreamResponseObserver.java
@@ -1,12 +1,10 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.hedera.block.server.consumer;
 
-import com.hedera.block.server.config.BlockNodeContext;
 import com.hedera.block.server.consumer.Functions.Task;
 import com.hedera.block.server.events.BlockNodeEventHandler;
 import com.hedera.block.server.events.ObjectEvent;
 import com.hedera.block.server.mediator.SubscriptionHandler;
-import com.hedera.block.server.metrics.MetricsService;
 import com.hedera.hapi.block.SubscribeStreamResponseUnparsed;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Objects;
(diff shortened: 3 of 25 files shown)


server/src/test/java/com/hedera/block/server/ServerConfigTest.java
```diff
@@ -18,37 +20,97 @@ void tearDown() {}
...
-            new ServerConfig(4_194_304, 1023);
-        });
+        try {
+            new ServerConfig(4_194_304, 32_768, 32_768, 1023);
+            fail("Expected IllegalArgumentException");
+        } catch (IllegalArgumentException e) {
+            String expectedErrorMsg = String.format(
+                    RANGE_ERROR_TEMPLATE, "server.port", 1023, ServerConfig.minPort, ServerConfig.maxPort);
+            assertEquals(expectedErrorMsg, e.getMessage());
+        }
```
ata-nas, line 114: It is never a good idea to use try-catch to assert an exception is thrown. It is very error prone. If you want to test for a custom error message, here an example test that utilizes AssertJ's fluent api to do so:

```java
    @ParameterizedTest
    @MethodSource("com.hedera.block.common.CommonsTestUtility#negativeIntegers")
    void testRequireWholeFail(final int toTest) {
        assertThatIllegalArgumentException()
                .isThrownBy(() -> Preconditions.requireWhole(toTest))
                .withMessage(DEFAULT_REQUIRE_WHOLE_MESSAGE.formatted(toTest));

        final String testMessage = DEFAULT_REQUIRE_WHOLE_MESSAGE.concat(" custom test error message");
        final String expectedTestMessage = testMessage.formatted(toTest);
        assertThatIllegalArgumentException()
                .isThrownBy(() -> Preconditions.requireWhole(toTest, testMessage))
                .withMessage(expectedTestMessage);
    }
```

I would suggest you use that instead of the try-catch. Also, it would not hurt to do parameterized tests for additional edge cases.

P.S. for other types of exceptions, AssertJ supports some well known types like null pointer and illegal argument, for other custom types, AssertJ has the method assertThatExceptionOfType(Class) and then you continue with the fluent api.

---

Side note (unrelated): a personal preference, I would always use AssertJ, it is really powerful and the fluent api allows for a lot of flexibility, worth taking a look at.

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
```diff
@@ -68,6 +68,8 @@ class ServerMappedConfigSourceInitializerTest {
 
         // Server Config
         new ConfigMapping("server.maxMessageSizeBytes", "SERVER_MAX_MESSAGE_SIZE_BYTES"),
+        new ConfigMapping("server.socketSendBufferSizeBytes", "SOCKET_SEND_BUFFER_SIZE_BYTES"),
+        new ConfigMapping("server.socketReceiveBufferSizeBytes", "SOCKET_RECEIVE_BUFFER_SIZE_BYTES"),
```
ata-nas, line 72: Let's keep the convention and use the full key of the config:

```suggestion
        new ConfigMapping("server.socketSendBufferSizeBytes", "SERVER_SOCKET_SEND_BUFFER_SIZE_BYTES"),
        new ConfigMapping("server.socketReceiveBufferSizeBytes", "SERVER_SOCKET_RECEIVE_BUFFER_SIZE_BYTES"),
```

server/src/main/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializer.java
```diff
@@ -42,6 +42,8 @@ public final class ServerMappedConfigSourceInitializer {
 
             // Server Config
             new ConfigMapping("server.maxMessageSizeBytes", "SERVER_MAX_MESSAGE_SIZE_BYTES"),
+            new ConfigMapping("server.socketSendBufferSizeBytes", "SOCKET_SEND_BUFFER_SIZE_BYTES"),
+            new ConfigMapping("server.socketReceiveBufferSizeBytes", "SOCKET_RECEIVE_BUFFER_SIZE_BYTES"),
```
ata-nas, line 46: Let's keep the mapping convention and use the full config key.

```suggestion
            new ConfigMapping("server.socketSendBufferSizeBytes", "SERVER_SOCKET_SEND_BUFFER_SIZE_BYTES"),
            new ConfigMapping("server.socketReceiveBufferSizeBytes", "SERVER_SOCKET_RECEIVE_BUFFER_SIZE_BYTES"),
```

ata-nas (commented):
Here a couple suggestions.

server/src/test/java/com/hedera/block/server/ServerConfigTest.java
mattp-swirldslabs: I refactored these by be ParameterizedTests leveraging the AssertJ fluent assert API. Take a look at the changes and see what you think. I agree their API leads to more readable code and there seems to be solid support for other cases. However, they're also simply using a try-catch under the hood to assert that an exception is thrown. They've just added a lot more syntactic sugar. :)

```
 public static Throwable catchThrowable(ThrowingCallable shouldRaiseThrowable) {
        try {
            shouldRaiseThrowable.call();
            return null;
        } catch (Throwable var2) {
            Throwable throwable = var2;
            return throwable;
        }
    }
```

server/src/main/java/com/hedera/block/server/consumer/Functions.java
```diff
@@ -0,0 +1,85 @@
...
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public Void call() {
+            try {
+                nextBlockNodeEventHandler.onEvent(event, l, b);
+            } catch (IllegalArgumentException e) {
+                throw e;
```
jsync-swirlds, line 67: If all you do is rethrow, you probably shouldn't even catch.

server/src/main/java/com/hedera/block/server/consumer/Functions.java
```diff
@@ -0,0 +1,85 @@
...
+                // service. This should be happening all the time.
+                subscriptionHandler.unsubscribe(asyncConsumerStreamResponseObserver);
+                LOGGER.log(
+                        DEBUG,
+                        "UncheckedIOException caught from Pipeline instance. Unsubscribed consumer observer instance");
+            } catch (Exception e) {
+                subscriptionHandler.unsubscribe(asyncConsumerStreamResponseObserver);
+                LOGGER.log(ERROR, "Exception caught from Pipeline instance. Unsubscribed consumer observer instance.");
+                LOGGER.log(ERROR, e.getMessage());
+            }
```
jsync-swirlds, line 81: Catching Exception is generally not preferred; particularly in a Callable "call" method.

It's defined to `throw Exception` deliberately so that chains can be completed exceptionally.  When you prevent any throw, it breaks a lot of assumptions in things like CompletionStage and Future.

In most cases the call method shouldn't catch at all, so that the ExecutionException (which wraps any exception thrown here) can be thrown on the `get` call and handled then.

server/src/main/java/com/hedera/block/server/consumer/Functions.java
```diff
@@ -0,0 +1,85 @@
...
+ * penalties.
+ */
+final class Functions {
+    private Functions() {}
+
+    /**
+     * Task is a Callable used to stream events asynchronously to
+     * a consumer.
+     */
+    static final class Task implements Callable<Void> {
```
jsync-swirlds, line 27: This name is awfully generic.  Perhaps something that describes what this functional class does a little better?  Something like `HandleNextBlockEvent` or `ProcessNextBlockEvent`?

Functional and callable classes _typically_ read best as a very (similar to a method name), particularly when used in completion pipelines.

server/src/main/java/com/hedera/block/server/mediator/MediatorConfig.java
```diff
@@ -15,10 +15,9 @@
  * @param type use a predefined type string to replace the mediator component implementation.
  *  Non-PRODUCTION values should only be used for troubleshooting and development purposes.
  */
-// 131072 works but not with persistence
 @ConfigData("mediator")
 public record MediatorConfig(
-        @Loggable @ConfigProperty(defaultValue = "4_194_304") int ringBufferSize,
+        @Loggable @ConfigProperty(defaultValue = "1_048_576") int ringBufferSize,
```
jsync-swirlds, line 20: Curiosity, is this bytes or events?

server/src/main/java/com/hedera/block/server/BlockNodeApp.java
```diff
@@ -88,12 +89,18 @@ public void start() throws IOException {
...
+                .receiveBufferSize(serverConfig.socketSendBufferSizeBytes())
+                .build();
+
         // Build the web server
         final WebServer webServer = webServerBuilder
                 .port(serverConfig.port())
                 .addProtocol(pbjConfig)
                 .addRouting(pbjRouting)
                 .addRouting(httpRouting)
+                .connectionConfig(connectionConfig)
```
ata-nas, line 103:
```suggestion
```

server/src/main/java/com/hedera/block/server/pbj/PbjBlockStreamServiceProxy.java
```diff
@@ -71,6 +74,7 @@ public PbjBlockStreamServiceProxy(
         streamMediator.subscribe(Objects.requireNonNull(streamPersistenceHandler));
         streamMediator.subscribe(Objects.requireNonNull(streamVerificationHandler));
         this.streamMediator = Objects.requireNonNull(streamMediator);
+        this.executorService = Executors.newVirtualThreadPerTaskExecutor();
```
jsync-swirlds, line 77: We might consider setting the system property `jdk.virtualThreadScheduler.parallelism` in our container setup. Java is notoriously poor at getting a correct processor count in a container, so setting the parallelism helps ensure it doesn't end up with a single carrier thread trying to handle all of the consumers.

server/src/main/java/com/hedera/block/server/verification/service/NoOpBlockVerificationService.java
```diff
@@ -10,20 +10,19 @@
...
+        LOGGER.log(INFO, "Using " + getClass().getName());
     }
 
     /**
      * Does nothing
      */
     @Override
     public void onBlockItemsReceived(@NonNull List<BlockItemUnparsed> blockItems) {
-        LOGGER.log(INFO, "Using " + getClass().getName());
+        // no-op
```
jsync-swirlds, line 26: Perhaps a log at debug would be better than doing nothing at all?

server/src/main/java/com/hedera/block/server/consumer/Functions.java
```diff
@@ -0,0 +1,85 @@
...
+                // clients disconnecting from the server streaming
+                // service. This should be happening all the time.
+                subscriptionHandler.unsubscribe(asyncConsumerStreamResponseObserver);
+                LOGGER.log(
+                        DEBUG,
+                        "UncheckedIOException caught from Pipeline instance. Unsubscribed consumer observer instance");
+            } catch (Exception e) {
+                subscriptionHandler.unsubscribe(asyncConsumerStreamResponseObserver);
+                LOGGER.log(ERROR, "Exception caught from Pipeline instance. Unsubscribed consumer observer instance.");
+                LOGGER.log(ERROR, e.getMessage());
```
jsync-swirlds, line 80: Have you tried just `LOGGER.log(ERROR, e);`?  Worst case it should include both _class_ and _message_, which is better than just the message.

server/src/main/java/com/hedera/block/server/consumer/AsyncConsumerStreamResponseObserver.java
```diff
@@ -0,0 +1,73 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void onEvent(
+            @NonNull final ObjectEvent<SubscribeStreamResponseUnparsed> event, final long l, final boolean b) {
+
+        try {
+            executorService
+                    .submit(new Task(event, l, b, subscriptionHandler, this, nextBlockNodeEventHandler))
+                    .get();
```
jsync-swirlds, line 55: The method `get()` blocks until the entire thread completes, which makes this almost the same as just calling `new Tast(..).call()` directly.

server/src/main/java/com/hedera/block/server/consumer/Functions.java
```diff
@@ -0,0 +1,85 @@
...
+                @NonNull final ObjectEvent<SubscribeStreamResponseUnparsed> event,
+                final long l,
+                final boolean b,
+                @NonNull final SubscriptionHandler<SubscribeStreamResponseUnparsed> subscriptionHandler,
+                @NonNull
+                        final BlockNodeEventHandler<ObjectEvent<SubscribeStreamResponseUnparsed>>
+                                asyncConsumerStreamResponseObserver,
+                @NonNull
+                        final BlockNodeEventHandler<ObjectEvent<SubscribeStreamResponseUnparsed>>
+                                nextBlockNodeEventHandler) {
```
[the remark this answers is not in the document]
mattp-swirldslabs: Cool trick, we'll have to socialize that to help with these crazy blocks of formatting. I fixed this one.

server/src/main/java/com/hedera/block/server/consumer/AsyncConsumerStreamResponseObserver.java
mattp-swirldslabs: Based on some exploration today with CompletionService, I'm going to keep this implementation. Allowing the blocking thread to return results in client side errors since the order of the sent List<BlockItem> becomes non-deterministic

jsync-swirlds: But then this is not async, nor concurrent.  We're trading time on one thread for the same time on two threads, which seems completely wasted.

Just keep the previous implementation and we can revisit something that is _actually_ concurrent later...

server/src/test/java/com/hedera/block/server/ServerConfigTest.java
ata-nas: The point is if we do the try catch then we can have an actual exception (not the one we intended) be caught and potentially silenced, simply we may write erroneous code. This would cause a false positive. Whereas using the library where it is standardized, it is not possible for us who wrote the test to do such a thing. Also it is not evident immediately what we are trying to achieve, future support will be cumbersome and no less error prone if try catch is used directly.

server/src/main/java/com/hedera/block/server/verification/service/NoOpBlockVerificationService.java
mattp-swirldslabs: Changed

server/src/main/java/com/hedera/block/server/mediator/MediatorConfig.java
mattp-swirldslabs: It's the size of the array used internally by the RingBuffer so it's number of events. I made adjustments to the MediatorConfig and NotifierConfig javadocs

server/src/main/java/com/hedera/block/server/consumer/Functions.java
mattp-swirldslabs: Currently, removal of that separate catch would result in the IllegalArgumentException being caught by the broader case further down. I isolated IllegalArgumentException since it's thrown if there's a spec error with the blockitems. That's a catastrophic failure to be propagated further up for the system to terminate connections and shut the server down. More general exceptions (SocketClosed, etc) will be caught and the individual consumer unsubscribed from the RingBuffer.

jsync-swirlds: Discussed separately.

server/src/main/java/com/hedera/block/server/pbj/PbjBlockStreamServiceProxy.java
mattp-swirldslabs: Makes sense...we can leverage this property if/when we move back to virtual threads

PR #568, opened 2025-02-03
refactor: Move Hasher to `common` sub-project

### LINKED ISSUE #569: Refactor of Hasher
Move Hasher to Common, so it can be re-utilized by both the server and the `simulator`

AlfredoG87:
**Description**:
This PR only moves the `hasher` subpackage of the `:server` subproject to the `:commons` subproject.

This so on a follow-up PR, we can implement the Generation Mode of the simulator with `valid` block_proof and hashes.

**Related issue(s)**:

Fixes #569

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/common/build.gradle.kts
+++ b/common/build.gradle.kts
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
-plugins { id("org.hiero.gradle.module.library") }
+plugins {
+    id("org.hiero.gradle.module.library")
+    id("application")
+}
 
 description = "Commons module with logic that could be abstracted and reused."
 
--- a/common/build.gradle.kts
+++ b/common/build.gradle.kts
@@ -1,7 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-plugins {
-    id("org.hiero.gradle.module.library")
-}
+plugins { id("org.hiero.gradle.module.library") }
 
 description = "Commons module with logic that could be abstracted and reused."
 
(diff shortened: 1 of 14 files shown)


common/build.gradle.kts
```diff
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
-plugins { id("org.hiero.gradle.module.library") }
+plugins {
+    id("org.hiero.gradle.module.library")
+    id("application")
```
mattp-swirldslabs, line 4: Why are we setting up the `common` subproject as an application? It's a library right?

mattp-swirldslabs (commented):
I had a question about setting this up as an application

common/build.gradle.kts
AlfredoG87: I agree, I also thought about it, however the `library` plugin is no longer there:
image


I think it was needed for be able to use the `requires com.hedera.pbj.runtime;` on the `module-info.java` however it seems it was not.

Thanks for noticing and asking, I have remove it already. :)

jsync-swirlds:
Just a note that we had originally designed the simulator to share as little as possible with the block node so that we cannot have a bug in block node that _also_ affects the simulator.  This ensures that when the simulator is sending a stream for testing, we do not "hide" a bug in the block node.

It's a bit more work to maintain multiple codebases that calculate the same hash in different ways, but it also ensures that we aren't tied to a single specific library, and improves the probability that the block node correctly calculates values according to the public specification.

common/build.gradle.kts
AlfredoG87: @mattp-swirldslabs, it was failing the tests, but now it passes 👍 
✅ Done

mattp-swirldslabs (approved):
I'm worried about the transitive dependencies here but I'll approve.

PR #585, opened 2025-02-05
fix: removed UTC timezone flags. we can rely on the linux default

### LINKED ISSUE #450: Set UserTime Zone at the container level (Dockerfile) instead of JAVA level using ENV config
User and user time are set in the container configuration, and we should probably set the user time to UTC at the container level (so we do not need to override in JVM command line).
I second the suggestion to discuss across the team.

_Originally posted by @jsync-swirlds in https://github.com/hashgraph/hedera-block-node/pull/446#discussion_r1904318301_

mattp-swirldslabs:
**Description**:
* Linux is set to UTC by default
* I removed the JVM options to explicitly set UTC

**Related issue(s)**:

Fixes #450 

**Notes for reviewer**:

**Checklist**


jsync-swirlds (approved):

AlfredoG87 (approved):
LGTM

PR #637, opened 2025-02-11
fix: ackHandler order responses under edge cases

### LINKED ISSUE #638: Bug: Under certain `edge` conditions Acks can be sent out of order.
Due to race conditions, is possible for the AckHandler to send Acks out of order.

Cases:

1. When streaming really really fast blocks into the BN (latency = 0)
2. When starting with block higher right after the system has started, meaning the first ack to attempt being sent is 2, and then 1 arrives. (this also breaks the order)

AlfredoG87:
## Reviewer Notes

This PR adds test that helped replicate edge cases when acks were sent out of order.
either due to the workaround to determine the first valid block (where to start sending acks) or
due to random delays occur in either `persistence` or `verification` threads.

Also this PR includes the fixes so those tests always pass instead of always failing or sometimes failing.

Fixes #638

**Before:**
image

**After:** Under the same conditions, same result in multiple passes.
image

### DIFF
--- a/server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
+++ b/server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
@@ -109,7 +109,7 @@ public class AckHandlerImpl implements AckHandler {
     private void attemptAcks() {
         // Temporarily if lastAcknowledgedBlockNumber is -1, we get the first block in the map
         if (lastAcknowledgedBlockNumber == -1) {
-            // todo(147): once we have a way to get the last acknowledged block from the store we should use that
+            // @todo(147): once we have a way to get the last acknowledged block from the store we should use that
             lastAcknowledgedBlockNumber = 0;
         }
 
--- a/server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
+++ b/server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
@@ -109,8 +109,8 @@ public class AckHandlerImpl implements AckHandler {
     private void attemptAcks() {
         // Temporarily if lastAcknowledgedBlockNumber is -1, we get the first block in the map
         if (lastAcknowledgedBlockNumber == -1) {
-            lastAcknowledgedBlockNumber =
-                    blockInfo.keySet().stream().min(Long::compareTo).orElse(1L) - 1;
+            // todo(147): once we have a way to get the last acknowledged block from the store we should use that
+            lastAcknowledgedBlockNumber = 0;
         }
 
         // Keep ACK-ing starting from the next block in sequence
@@ -135,9 +135,6 @@ public class AckHandlerImpl implements AckHandler {
                 // We "won" the race; we do the actual ACK
                 notifier.sendAck(nextBlock, info.getBlockHash(), false);
 
-                // Update last acknowledged
-                lastAcknowledgedBlockNumber = nextBlock;
-
                 // Update the service status
                 serviceStatus.setLatestAckedBlock(info);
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
```diff
@@ -111,8 +111,8 @@ public void blockVerificationFailed(long blockNumber) {
     private void attemptAcks() {
         // Temporarily if lastAcknowledgedBlockNumber is -1, we get the first block in the map
         if (lastAcknowledgedBlockNumber == -1) {
-            lastAcknowledgedBlockNumber =
-                    blockInfo.keySet().stream().min(Long::compareTo).orElse(1L) - 1;
+            // todo(147): once we have a way to get the last acknowledged block from the store we should use that
+            lastAcknowledgedBlockNumber = 0;
```
ata-nas, line 115: As noted, as a very temporary solution, here is started from 0, but needs to change once we have the appropriate logic, otherwise the stream cannot start from another block number.

server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
```diff
@@ -137,19 +137,24 @@ private void attemptAcks() {
...
                 blockInfo.remove(nextBlock);
 
-                // Update metrics
+                // Update metrics and logging
                 metricsService.get(BlockNodeMetricTypes.Counter.AckedBlocked).increment();
-            }
+                LOGGER.log(System.Logger.Level.DEBUG, "ACKed block " + nextBlock);
 
+                // Update last acknowledged
+                lastAcknowledgedBlockNumber = nextBlock;
```
ata-nas, line 151: As we noted, the update to the volatile field is the one that will allow other threads to get inside the if check, so this being done last is important.

AlfredoG87: Correct, this was left to the last step indeed

mattp-swirldslabs (approved):
It looks like the discussions were resolved. Looks good Fredy

server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
```diff
@@ -111,8 +111,8 @@ public void blockVerificationFailed(long blockNumber) {
     private void attemptAcks() {
         // Temporarily if lastAcknowledgedBlockNumber is -1, we get the first block in the map
         if (lastAcknowledgedBlockNumber == -1) {
-            lastAcknowledgedBlockNumber =
-                    blockInfo.keySet().stream().min(Long::compareTo).orElse(1L) - 1;
+            // todo(147): once we have a way to get the last acknowledged block from the store we should use that
```
Nana-EC, line 114: nit: please link to an actual issue so we have it on the board.
I'm not a fan of to-do that can get lost in code.

Nana-EC (commented):
LG
Rebase to resolve conflict and todo comment

server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
AlfredoG87: That is what the `todo(147)` means, 147 is the actual issue number, ohhh am missing the `@` it should be `@todo(147)` 😅  will fix during rebase.

I waited to merge this one on Atanas request, to avoid conflict resolution on his side, since his ticket was way more complicated and large than this one, I yielded in favor of him, but now I can finish it :)

mattp-swirldslabs (approved):

Nana-EC (approved):

PR #679, opened 2025-02-20
fix: fixed broken unit tests from new metric

mattp-swirldslabs:
## Reviewer Notes
* Fixed failing unit tests I broke with my last commit/merge

## Related Issue(s)
Related #670

### DIFF
--- a/server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncBlockAsLocalFileWriterTest.java
+++ b/server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncBlockAsLocalFileWriterTest.java
@@ -61,7 +61,10 @@ class AsyncBlockAsLocalFileWriterTest {
     private MetricsService metricsServiceMock;
 
     @Mock
-    private Counter counterMock;
+    private Counter successfulPersistenceCounterMock;
+
+    @Mock
+    private Counter persistenceErrorCounterMock;
 
     @TempDir
     private Path testTempDir;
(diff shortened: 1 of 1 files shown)


server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncBlockAsLocalFileWriterTest.java
```diff
@@ -352,7 +358,7 @@ private void verifySuccessfulPersistencePublish(final BlockPersistenceResult act
     private void verifyUnsuccessfulPersistencePublish(final BlockPersistenceResult actual) {
         verify(ackHandlerMock, times(1)).blockPersisted(actual);
         verify(metricsServiceMock, never()).get(BlocksPersisted);
-        verify(counterMock, never()).increment();
+        verify(counterMock, times(1)).increment();
```
ata-nas, line 361: Let's make a different counter for every different metric, in this way, we can have the correct asserts. My opinion would be to have a second counter mock call it persistenceErrorCounterMock and maybe rename the other mock to  successfulPersistenceCounterMock so we can easily diff them.

ata-nas (commented):
I have a suggestion so we can easily have the correct asserts.

AlfredoG87 (dismissed):

georgi-l95 (approved):

server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncBlockAsLocalFileWriterTest.java
mattp-swirldslabs: Fixed.

ata-nas (approved):
🔥

georgi-l95 (approved):
👀 👍 👉 👨‍💻

Issue #675, opened 2025-02-19
Add ability to "pull" block item batches from the RingBuffer data channel


mattp-swirldslabs:
### Persona

As a developer

### Request

I want to add the ability to "pull" block item batches from the RingBuffer live data channel rather than accept items "pushed" to me

### Goal

So that I can more easily toggle between historic and live stream items in the consumer observer

### Technical Notes

Prerequisite for #672 
Related to #488

Issue #672, opened 2025-02-19
feat: open-range historic block streaming


mattp-swirldslabs:
### Persona

As a client

### Request

I want to request a streaming range of blocks starting with an existing block and continuing on with the live stream

### Goal

So that I can start with an historic block and catch up to the live stream of blocks

### Technical Notes

Technical details, notes, suggestions, etc...

Design Doc:
#488

PR #689, opened 2025-02-21
00675 pull events from rb

mattp-swirldslabs:
## Related Issue(s)
feat #675


server/src/main/java/com/hedera/block/server/mediator/PollerImpl.java
```diff
@@ -0,0 +1,92 @@
...
+            @NonNull final EventPoller<V> eventPoller,
+            final int batchSize,
+            @NonNull final LongGauge consumerBuffersRemainingCapacity) {
+        this.poller = Objects.requireNonNull(eventPoller);
+        this.polledData = new BatchedData<>(batchSize);
+        this.consumerBuffersRemainingCapacity = Objects.requireNonNull(consumerBuffersRemainingCapacity);
+    }
+
+    @Override
+    public V poll() throws Exception {
```
jsync-swirlds, line 35: We should not declare `Exception` in the implementation.  Even if and interface is declared to throw `Exception` (often to allow wider use), the implementation(s) should only declare the specific exception(s) necessary for that implementation.

server/src/main/java/com/hedera/block/server/mediator/Poller.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+public interface Poller<V> {
+
+    /**
+     * Polls the next event from the stream of events.
+     *
+     * @return the next event
+     * @throws Exception if an error occurs while polling the event
+     */
+    V poll() throws Exception;
```
jsync-swirlds, line 17: Can we make this more specific?  What kinds of exceptions do we expect to be thrown from this internal-to-the-service interface?

server/src/main/java/com/hedera/block/server/mediator/Poller.java
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.block.server.mediator;
+
+/**
+ * The Poller interface defines the contract for polling the next event from the stream of events.
+ *
+ * @param <V> the type of the polled event
+ */
+public interface Poller<V> {
```
jsync-swirlds, line 9: Just a thought.  It looks like this is intended to be used only within the mediator; which suggests we'll only ever have the one implementation.
Could we leave out the interface/impl split and just have a `Poller`?

server/src/main/java/com/hedera/block/server/consumer/PollConsumerStreamResponseObserver.java
```diff
@@ -0,0 +1,142 @@
...
+            } catch (Exception e) {
+                subscriptionHandler.unsubscribePoller(this);
+                throw new RuntimeException(e);
+            }
+            try {
+                Thread.sleep(1);
+            } catch (InterruptedException e) {
+                subscriptionHandler.unsubscribePoller(this);
+                throw new RuntimeException(e);
+            }
```
jsync-swirlds, line 75: When catching the interrupted exception, it's generally best to reset the thread interrupted flag, even if throwing an exception.

In this case, why not just `break` out of the while loop instead of throwing an exception?  That ends this thread just as quickly and with less "fanfare".

server/src/main/java/com/hedera/block/server/consumer/PollConsumerStreamResponseObserver.java
```diff
@@ -0,0 +1,142 @@
...
+                    if (isResponsePermitted.get()) {
+                        final List<BlockItemUnparsed> blockItems = event.get();
+                        send(blockItems);
+                    }
+                    refreshLiveness();
+                }
+            } catch (Exception e) {
+                subscriptionHandler.unsubscribePoller(this);
+                throw new RuntimeException(e);
+            }
```
jsync-swirlds, line 69: If this is all we do with exceptions from the poller; I wonder if we'd be better off just not declaring any thrown exceptions for the poller, catching runtime exception here, and re-throwing instead of wrapping?

server/src/main/java/com/hedera/block/server/consumer/PollConsumerStreamResponseObserver.java
```diff
@@ -0,0 +1,142 @@
...
+     */
+    @Override
+    public boolean isTimeoutExpired() {
+        if (livenessCalculator != null) {
+            subscriptionHandler.unsubscribePoller(this);
+            return livenessCalculator.isTimeoutExpired();
+        }
+
+        return false;
+    }
```
jsync-swirlds, line 96: This seems off.  We unsubscribe every time anything checks if a timeout is expired?  Shouldn't the unsubscribe be conditioned on whether the timeout is expired (or better, controlled by whatever is checking for an expired timeout)?

server/src/main/java/com/hedera/block/server/verification/session/BlockVerificationSessionBase.java
```diff
@@ -145,7 +145,7 @@ protected void finalizeVerification(BlockProof blockProof) {
 
             result = new VerificationResult(blockNumber, blockHash, BlockVerificationStatus.VERIFIED);
         } else {
-            LOGGER.log(INFO, "Block verification failed for block number: {0}", blockNumber);
+            LOGGER.log(WARNING, "Block verification failed for block number: {0}", blockNumber);
```
jsync-swirlds, line 148: I do not believe this severity increase is warranted.
A log level of WARNING (or above) should (in my opinion) be reserved for events that indicate _this process_ is not operating correctly.  Here we just detected that the data sent to us is not verified, which is _normal operation_ for the block node. It may be abnormal for the publisher, but that's for them to wake in the middle of the night about.

simulator/src/main/resources/app.properties
```diff
@@ -24,16 +24,16 @@ prometheus.endpointPortNumber=9998
 # ----------------------------------------------
 # BlockAsFileLargeDataSets config
 # ----------------------------------------------
-#generator.folderRootPath=/Users/user/Downloads/block-0.0.3-perf
-#generator.managerImplementation=BlockAsFileLargeDataSets
+generator.folderRootPath=/Users/mattpeterson/Downloads/BN_Test_Data/block-0.0.3
```
jsync-swirlds, line 27: We should change the simulator to use something like `/tmp` instead of a user folder that has to be changed like this (and exposes local usernames).

Just a side thought.

jsync-swirlds (commented):
The general direction is reasonable.

I know this is in-progress, but I've added comments for things that might need further attention.

You're welcome to just resolve any that are related to the not-yet-finished status.

server/src/main/java/com/hedera/block/server/mediator/PollerImpl.java
mattp-swirldslabs: `com.lmax.disruptor.EventPoller.poll()` throws Exception so we're stuck with it.

server/src/main/java/com/hedera/block/server/mediator/Poller.java
mattp-swirldslabs: See previous

server/src/main/java/com/hedera/block/server/mediator/Poller.java
mattp-swirldslabs: I prefer the split to facilitate testing with mocks, etc

server/src/main/java/com/hedera/block/server/consumer/PollConsumerStreamResponseObserver.java
mattp-swirldslabs: Yeah, I had a similar thought. I've changed this quite a bit with the historic streaming work. I think you'll like the changes. It's much simpler. With the polling approach, I'm better able to handle unsubscribing from the RB here and just exiting the loop. I should have those changes for review next week.

server/src/main/java/com/hedera/block/server/consumer/PollConsumerStreamResponseObserver.java
mattp-swirldslabs: I used the original `ConsumerStreamResponseObserver` as a starting point for this class. That's legacy from that copy/paste. I'll be refactoring that for #672.

simulator/src/main/resources/app.properties
mattp-swirldslabs: That might be a good simulator enhancement. We'll have to talk with Georgi about it.

server/src/main/java/com/hedera/block/server/consumer/PollConsumerStreamResponseObserver.java
mattp-swirldslabs: Yeah, I've revisited the exception handling in #672 . The polling approach should make this a lot cleaner.

mattp-swirldslabs:
Closing this PR since it was meant to foster an early discussion. These changes will be rolled into a future PR

PR #691, opened 2025-02-21
chore: Document some thoughts for nano-service approach

jsync-swirlds:
## Reviewer Notes
* Added a diagram to depict the general structure
* Added a document to gather initial thoughts.


ata-nas (dismissed):
Looks good to me.

docs/design/Nano-Service-Approach.md
```diff
@@ -0,0 +1,103 @@
+# Block Node Nano-Service Approach
```
mattp-swirldslabs, line 1: How is this different from a Micro-Kernel (Plugin) architecture? If it's not different, why rebrand it as nano-services?

docs/design/Nano-Service-Approach.md
```diff
@@ -0,0 +1,103 @@
...
+   service both simple and focused.  This also makes adding, removing, and
+   modifying these services much easier and faster.
+   * It's also much easier to test services with nothing more than a mock of the
+     "Messaging" service; which further improves velocity.
+2. Composing services may be easier to reason about than composing interfaces,
+   and systems composed of independent services are easier to modify and revise
+   than systems with many interacting method calls or complex module
+   interactions.
+3. It is much easier to reason about concurrency for a single focused service
+   than it is for a larger and more interconnected set of components.
```
mattp-swirldslabs, line 103: What about a Possible Drawbacks section and how to mitigate those risks?

mattp-swirldslabs (dismissed):
Looks good. Thanks Joseph!

docs/design/Nano-Service-Approach.md
```diff
@@ -0,0 +1,103 @@
...
+</dl>
+
+## Core Concepts
+
+1. Helidon and API objects are restricted to the API layer.
+   * The less we flow these externally defined interfaces and classes through
+     the system, the more easily we can potentially make outward facing API
+     changes without reworking the internal design.
+2. Most services are not required to be present in every deployment of the block node.
+3. No service should depend on classes or interfaces from another service.
```
georgi-l95, line 35: Should we mention that it's expected for some services, to require other services in order to work correctly? (the case with state managment for example)

georgi-l95 (dismissed):

docs/design/Nano-Service-Approach.md
jsync-swirlds: Micro-kernel has a ton of implications as it's a well understood O/S design concept that carries a great deal of conceptual baggage.

docs/design/Nano-Service-Approach.md
jsync-swirlds: Excellent Idea, I'll add that.

docs/design/Nano-Service-Approach.md
jsync-swirlds: I'll add something in the section that Matt recommended.
I'll also probably name it something like "Considerations and Possible Concerns".

docs/design/Nano-Service-Approach.md
```diff
@@ -0,0 +1,103 @@
...
+   * That service might not be deployed, and each service should be removable
+     without breaking other services.  The exception is the Messaging service.
+   * To this end, services should be independent modules with clearly defined
+     and carefully controlled interactions.
+4. Two services are required for all deployments, Messaging and Status
```
AlfredoG87, line 40: What about modules such as `MetricsService` or `Config`? should not also be a direct call and injected to all modular nano services and system services?

AlfredoG87 (commented):
Should we provide an example of the Record Events Structure?
Should this document delineate the phased and order of execution (plan) for implementing this in a way that disrupts the minimum with parallel progress for more features?

docs/design/Nano-Service-Approach.md
jsync-swirlds: Metrics and config are up to the services (since we won't control all of them).
Any particular service _that we write_ might obtain metrics or config; but we cannot assume all services will do so.

As to how, during service load the service will need to register with the message service, an optional call to that service will offer configuration and/or metrics, if the plugin service wants those.

That's more detailed than this document strives to be, but we'll definitely cover those details in the Message Service documentation.

jsync-swirlds:
> Should we provide an example of the Record Events Structure? Should this document delineate the phased and order of execution (plan) for implementing this in a way that disrupts the minimum with parallel progress for more features?

That's outside the scope of this document, and is important enough to warrant its own effort.  Short answer: We have yet to make any decisions in this regard, and may plan just enough for the next increment of value at any given time.

docs/design/Nano-Service-Approach.md
jsync-swirlds: Third party services won't participate in dagger (there's no way to make that work, as we will not run their build).

Dagger is fine within a service we build, if it is useful there, but it requires that all dependencies be present _at compile time_, so it is not going to work for enabling/adding services in the node or registering services with messaging (things that must be entirely determined at runtime).

If we want to have _global_ metrics and/or configuration available, it has to be something that can be obtained by plugins that do not participate in our build, and cannot receive anything via dependency injection. Having a narrow API for plugins to request those things (e.g. metrics, or configuration) enables us to provide the necessary _capability_ in a consistent manner to all plugins without making it overly constraining for a third party plugin developer.

Note, as an alternative, we might use Dagger/DI to inject metrics for everything _we build_, and make those things blind dependencies (they're injected, but do not make cross-service calls).  That makes some things easier for our services, but makes third-party services a "second class" service as we've effectively excluded them from using metrics or configuration facilities.

To your last question, it's not clear at the moment whether we will continue to use a dependency injection framework like Dagger broadly or we might need a more dynamic setup for some, or even all, dependency objects.

Overall, this discussion highlights a design decision that probably needs more discussion than fits within a PR review.

AlfredoG87: I think that we might end-up using Dagger and some direct calls, within the Core of the `system` everything that is not a "Plugin".
Then there should be some APIs for plugins to be able to interact with from the "Core System".

I already can think of 3 Core Modules that **ALL** Plugins should be able to interact with.
1. MessageService
2. ConfigurationService
3. MetricsService


I guess that we can have separate package called `types` (or something better) that mainly defines the interfaces that both the block node server `core` and the block node server plugins `plugins` can interact with.

for example. we can define an API with some interfaces to register needed metrics, so persistence or verification will be able to register their own metrics, and then, able to use them.

We can include a `Dep` Loading mechanism at start-up, where a plugin basically registers with the different services that are available. (maybe is required to register with the MessageService) but Configuration and Metrics Services should be optional (to not overly constrain the 3rd Party developer xp)

Again, I agree that this by itself is appropriate of its own discussion.

AlfredoG87 (approved):
There are still a lot of open questions and unknows unknows but seems like this is the path we want to take for the future, so approving 👍 

Thank you @jsync-swirlds for leading this effort.

docs/design/Nano-Service-Approach.md
jsync-swirlds: I think we're closer than it might seem.
I see configuration and metrics as "facilities", rather than full services.  Something that's not using or subject to messaging.
The thought of using the Messaging registration as a means to deliver those two facilities (if the caller wishes) was to avoid adding more complexity beyond the (necessary) registration with Messaging.  We might take a slightly different approach by using some form of central "registration" facility that provides all three (messaging, configuration, and metrics) instead.

AlfredoG87: Agreed 👍

PR #692, opened 2025-02-22
fix: CodeCov fix

AlfredoG87:
## Reviewer Notes

This is the fix for CodeCov reports on open PRs.

**FINAL UPDATE:**
is working with @jjohannes suggestion 💯 
image

**Before:**
image

**After:** 🤞 
Try 2:
image

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -92,3 +92,4 @@ jobs:
         uses: codecov/codecov-action@13ce06bfc6bbe3ecf90edbbf1bc32fe5978ca1d3 # v5.3.1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+          files: gradle/aggregation/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml


AlfredoG87:
@jjohannes Can you validate why with new hiero gradle plugins we need to set manually to generate `xml` jacoco reports. seems like only `html` is set by default.

@andrewb1269hg this seems to fix it 👍 🥳

jsync-swirlds (approved):
LGTM.

I second the question raised to Jendrik; it would be good to be able to set the xml requirement globally.

andrewb1269hg (approved):
Approve the `pr-checks.yaml` file.

.github/workflows/pr-checks.yaml
```diff
@@ -90,5 +90,5 @@ jobs:
 
       - name: Upload coverage report
         uses: codecov/codecov-action@13ce06bfc6bbe3ecf90edbbf1bc32fe5978ca1d3 # v5.3.1
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
jjohannes, line 94:
```suggestion
        env:
          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
        with:
          files: gradle/aggregation/build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml
```

See: https://github.com/hiero-ledger/hiero-consensus-node/blob/main/.github/workflows/node-zxc-compile-application-code.yaml#L525

jjohannes (changes requested):

jjohannes:
There are basically two approaches to how Gradle interacts with other coverage tools:
1) generate an XML for each module in isolation, upload all of them to the coverage tool and let the coverage tool aggregate the information.
2) Let [Grade generate one aggregated XML](https://docs.gradle.org/current/userguide/jacoco_report_aggregation_plugin.html) and only upload that.


In the _hiero-consensus-node_ repo we decided for (2) some months ago. In the hope that it is (a bit) more accurate. It will also contain the information if a test in `Module A` covers code in `Module B`. In any case, that is the setup we decided for back then and which is currently what the Gradle setup is designed for.
See: https://github.com/hiero-ledger/hiero-gradle-conventions?tab=readme-ov-file#code-coverage-reports

I don't know how the CodeCov action finds XML reports. But it seems like it does not find the aggregated one by default. That's why we need to configure the path to it (see my suggested change).

The changes to the `build.gradle.kts` files are **not** needed as we do not need the individual XMLs.

.github/workflows/pr-checks.yaml
AlfredoG87: Thank you for the clarification here @jjohannes I was un-aware of such decision and that it was even there.
now you mention it makes sense. 

Also thanks for the suggested changes, I will test it out and if it passes, we can finally close this one 😅 👍

AlfredoG87: Changes done ✅

.github/workflows/pr-checks.yaml
```diff
@@ -92,3 +92,4 @@ jobs:
         uses: codecov/codecov-action@13ce06bfc6bbe3ecf90edbbf1bc32fe5978ca1d3 # v5.3.1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
```
jsync-swirlds, line 94: Does the `token` need to be in `env`, and the `with` section then only has the `files`?

jjohannes: Both solutions seem to work. This is now the same as here (in with block):

https://github.com/hiero-ledger/hiero-consensus-node/blob/main/.github/workflows/node-zxc-compile-application-code.yaml#L520-L525

I would keep it like that.

AlfredoG87: `with` seems to be the general accepted practice.

andrewb1269hg (approved):

PR #702, opened 2025-02-24
fix: adjusted FileHandler and ConsoleHandler logging to ALL to see TRACE

### LINKED ISSUE #701: Change logging handlers to accept ALL to make TRACE level visible
`logging.properties` currently sets these 2 values:

`java.util.logging.FileHandler.level = FINE`
`java.util.logging.ConsoleHandler.level = FINE`

Unfortunately, this will omit logging at the `TRACE` level even if you have an explicit logging entry like:
`com.hedera.block.server.Classname.level=FINER`

Proposed change:

`java.util.logging.FileHandler.level = ALL`
`java.util.logging.ConsoleHandler.level = ALL`

This should re-enable the ability to specify logging at the class level like:
`com.hedera.block.server.Classname.level=FINER`

mattp-swirldslabs:
## Related Issue(s)
Fixes #701

### DIFF
--- a/server/docker/logging.properties
+++ b/server/docker/logging.properties
@@ -50,7 +50,7 @@ handlers = java.util.logging.ConsoleHandler, java.util.logging.FileHandler
 ################################################################################
 # ConsoleHandler configuration
 ################################################################################
-java.util.logging.ConsoleHandler.level = FINE
+java.util.logging.ConsoleHandler.level = ALL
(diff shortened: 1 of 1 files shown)


ata-nas (approved):
Works as intended, TRACE is visible🔥

jsync-swirlds (approved):

PR #715, opened 2025-02-25
feat: add unverified blocks root

### LINKED ISSUE #582: Add unverified blocks root
Instead of storing block files directly in the local destination, we wish to modify the block writers as follows:
1. Store block files initially in an "unverified" folder, just flat with full file name.
1. A separate process initiated when a result is published by the verification service will move the file.
   - When a "verified" result is published this will move the now-verified block (only) to its proper destination
   - When an "unverified" result is published, this will remove the now-failed block (only).
   - When other result types are published this sub-service will only react if appropriate.
1. On startup, the persistence service will remove all unverified files.
   - This might be accomplished by using a folder on tmpfs, which is managed in-memory by the kernel and would significantly improve reliability and performance.
1. All other persistence service elements, and all archive services, will operate only on the trie structure folders containing verified block files.

ata-nas:
## Reviewer Notes

- add a new unverified blocks root
- persistence will first persist blocks under the unverfied root
- after successful persistence under unverified && successful verification, persistence should move the block to live
- due to unavaliable critical infrastructure, a method call has to be made to persistence to move the block

## Related Issue(s)

Closes #582

### DIFF
--- a/server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
+++ b/server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
@@ -1,10 +1,13 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.hedera.block.server.ack;
 
+import static java.lang.System.Logger.Level.ERROR;
+
 import com.hedera.block.server.block.BlockInfo;
 import com.hedera.block.server.metrics.BlockNodeMetricTypes;
 import com.hedera.block.server.metrics.MetricsService;
 import com.hedera.block.server.notifier.Notifier;
+import com.hedera.block.server.persistence.StreamPersistenceHandlerImpl;
 import com.hedera.block.server.persistence.storage.remove.BlockRemover;
 import com.hedera.block.server.persistence.storage.write.BlockPersistenceResult;
 import com.hedera.block.server.persistence.storage.write.BlockPersistenceResult.BlockPersistenceStatus;
@@ -35,6 +38,7 @@ public class AckHandlerImpl implements AckHandler {
     private final ServiceStatus serviceStatus;
     private final BlockRemover blockRemover;
     private final MetricsService metricsService;
+    private StreamPersistenceHandlerImpl streamPersistenceHandler;
 
     /**
      * Constructor. If either skipPersistence or skipVerification is true,
@@ -55,6 +59,11 @@ public class AckHandlerImpl implements AckHandler {
     }
 
     @Override
+    public void registerPersistence(@NonNull final StreamPersistenceHandlerImpl streamPersistenceHandler) {
+        this.streamPersistenceHandler = Objects.requireNonNull(streamPersistenceHandler);
+    }
+
+    @Override
     public void blockPersisted(@NonNull final BlockPersistenceResult blockPersistenceResult) {
         Objects.requireNonNull(blockPersistenceResult);
         if (!skipAcknowledgement) {
@@ -63,10 +72,11 @@ public class AckHandlerImpl implements AckHandler {
                 final BlockInfo info = blockInfo.computeIfAbsent(blockNumber, BlockInfo::new);
                 info.getBlockStatus().setPersisted();
             } else {
-                // @todo(545) handle other cases for the blockPersistenceResult
+                // @todo(774) handle other cases for the blockPersistenceResult
                 //   for now we will simply send an end of stream message
                 //   but more things need to be handled, like ensure the
-                //   blockInfo map will not be inserted
+                //   blockInfo map will not be inserted. We should use a
+                //   persistence failed response code as well.
                 blockVerificationFailed(blockNumber);
                 blockInfo.remove(blockNumber);
             }
@@ -101,9 +111,10 @@ public class AckHandlerImpl implements AckHandler {
     public void blockVerificationFailed(long blockNumber) {
         notifier.sendEndOfStream(lastAcknowledgedBlockNumber, PublishStreamResponseCode.STREAM_ITEMS_BAD_STATE_PROOF);
         try {
-            blockRemover.removeLiveUnverified(blockNumber);
-        } catch (IOException e) {
-            LOGGER.log(System.Logger.Level.ERROR, "Failed to remove block " + blockNumber, e);
+            blockRemover.removeUnverified(blockNumber);
+        } catch (final IOException e) {
+            final String message = "Failed to remove Block with number [%d]".formatted(blockNumber);
+            LOGGER.log(ERROR, message, e);
             throw new RuntimeException(e);
         }
         blockInfo.remove(blockNumber);
@@ -140,6 +151,25 @@ public class AckHandlerImpl implements AckHandler {
 
             // Attempt to mark ACK sent (CAS-protected to avoid duplicates)
             if (info.getBlockStatus().markAckSentIfNotAlready()) {
+                try {
+                    // @todo(582) if we are unable to move the block to the verified state,
+                    //   should we throw or for now simply take the same action as if the block
+                    //   failed persistence (for now since we lack infrastructure we simply
+                    //   call the verification failed method)
+                    streamPersistenceHandler.moveVerified(nextBlock);
+                } catch (final IOException e) {
+                    // @todo(582) if we do this, we must be aware that we will not increment
+                    //   lastAcknowledgedBlockNumber and the verification failed method will
+                    //   remove the info from the map. This means that the data needs to be requested
+                    //   again. What would be the best way to hande inability to move the block with
+                    //   the limitations we current have?
+                    // @todo(774) we should use a response code for failed persistence here
+                    final String message = "Failed to move Block with number [%d] from unverified to live storage"
+                            .formatted(nextBlock);
+                    LOGGER.log(ERROR, message, e);
+                    blockVerificationFailed(nextBlock);
+                    return;
+                }
                 // We "won" the race; we do the actual ACK
                 notifier.sendAck(nextBlock, info.getBlockHash(), false);
 
--- a/server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
@@ -5,6 +5,7 @@ import static com.hedera.block.server.metrics.BlockNodeMetricTypes.Counter.Strea
 import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.ERROR;
 
+import com.hedera.block.common.utils.FileUtilities;
 import com.hedera.block.server.ack.AckHandler;
 import com.hedera.block.server.config.BlockNodeContext;
 import com.hedera.block.server.events.BlockNodeEventHandler;
@@ -15,6 +16,8 @@ import com.hedera.block.server.metrics.MetricsService;
 import com.hedera.block.server.notifier.Notifier;
 import com.hedera.block.server.persistence.storage.PersistenceStorageConfig;
 import com.hedera.block.server.persistence.storage.archive.LocalBlockArchiver;
+import com.hedera.block.server.persistence.storage.path.BlockPathResolver;
+import com.hedera.block.server.persistence.storage.path.UnverifiedBlockPath;
 import com.hedera.block.server.persistence.storage.write.AsyncBlockWriter;
 import com.hedera.block.server.persistence.storage.write.AsyncBlockWriterFactory;
 import com.hedera.block.server.persistence.storage.write.BlockPersistenceResult;
@@ -24,17 +27,21 @@ import com.hedera.hapi.block.BlockItemUnparsed;
 import com.hedera.hapi.block.stream.output.BlockHeader;
 import com.hedera.pbj.runtime.ParseException;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.CompletionService;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorCompletionService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TransferQueue;
+import java.util.stream.Stream;
 import javax.inject.Singleton;
 
 /**
@@ -56,6 +63,7 @@ public class StreamPersistenceHandlerImpl implements BlockNodeEventHandler<Objec
     private final AsyncBlockWriterFactory asyncBlockWriterFactory;
     private final CompletionService<Void> completionService;
     private final LocalBlockArchiver archiver;
+    private final BlockPathResolver pathResolver;
     private TransferQueue<BlockItemUnparsed> currentWriterQueue;
 
     /**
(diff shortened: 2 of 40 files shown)


server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
```diff
@@ -80,21 +86,62 @@ public StreamPersistenceHandlerImpl(
...
+    }
+
+    public void moveVerified(final long blockNumber) throws IOException {
+        final Optional<UnverifiedBlockPath> optUnverified = pathResolver.findUnverifiedBlock(blockNumber);
+        if (optUnverified.isPresent()) {
+            final UnverifiedBlockPath unverifiedBlockPath = optUnverified.get();
+            final Path source = unverifiedBlockPath.dirPath().resolve(unverifiedBlockPath.blockFileName());
+            final Path rawPathToLive = pathResolver.resolveLiveRawPathToBlock(unverifiedBlockPath.blockNumber());
+            final Path target = FileUtilities.appendExtension(
+                    rawPathToLive, unverifiedBlockPath.compressionType().getFileExtension());
```
jsync-swirlds, line 138: I believe the extension of the block file should be `blk` regardless of compression.  The block node fully manages its files and we do not plan to support multiple compression options (nor changing compression) for a block node.  If we ever decide to change that, perhaps then we'll use a different extension, or perhaps we'll just check the file content for compression type (most compression sets a "magic byte" at the beginning of the file).

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
```diff
@@ -36,6 +37,7 @@ public final class BlockAsLocalFilePathResolver implements BlockPathResolver {
     public BlockAsLocalFilePathResolver(@NonNull final PersistenceStorageConfig config) throws IOException {
         this.liveRootPath = Objects.requireNonNull(config.liveRootPath());
         this.archiveRootPath = Objects.requireNonNull(config.archiveRootPath());
+        this.unverifiedRootPath = Objects.requireNonNull(config.unverifiedRootPath());
         this.archiveGroupSize = config.archiveGroupSize();
         this.archiveDirDepth = MAX_LONG_DIGITS - (int) Math.log10(this.archiveGroupSize);
         this.longLeadingZeroesFormat = new DecimalFormat("0".repeat(MAX_LONG_DIGITS));
```
jsync-swirlds, line 43: Instead of using DecimalFormat and constructing here, would it make sense to use a constant string format?
i.e. `private static final String NINETEEN_DIGITS_FORMAT = "%019d";`
Then, when used `NINETEEN_DIGITS_FORMAT.formatted(blockNumber)`

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
```diff
@@ -130,27 +130,36 @@ public Optional<ArchiveBlockPath> findArchivedBlock(final long blockNumber) {
...
         final CompressionType[] allCompressionTypes = CompressionType.values();
         for (int i = 0; i < allCompressionTypes.length; i++) {
-            final CompressionType compressionType = allCompressionTypes[i];
-            final Path compressionExtendedUnverifiedPath = FileUtilities.appendExtension(
-                    pathToUnverifiedBlockNoCompressionExtension, compressionType.getFileExtension());
-            if (Files.exists(compressionExtendedUnverifiedPath)) {
-                doMarkUnverified(compressionExtendedUnverifiedPath);
+            final CompressionType localCompressionType = allCompressionTypes[i];
+            final Path compressionExtendedBlockPath =
+                    FileUtilities.appendExtension(unverifiedPathToBlock, localCompressionType.getFileExtension());
```
jsync-swirlds, line 144: Here too, I think trying to have compression-specific extensions makes things more complex. It might be simpler to just always have `.blk` and assume all compression is ZStd for now. Future work might support other compression options by checking the "magic byte" or similar inspection, to handle the potentially mixed compression types case.

server/src/main/java/com/hedera/block/server/persistence/storage/path/UnverifiedBlockPath.java
```diff
@@ -0,0 +1,25 @@
...
+import com.hedera.block.server.persistence.storage.PersistenceStorageConfig.CompressionType;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.nio.file.Path;
+import java.util.Objects;
+
+/**
+ * A record that represents a detailed path to an unverified Block that is
+ * stored under the unverified storage root.
+ */
+public record UnverifiedBlockPath(
```
jsync-swirlds, line 14: Do we really need all these parts?  Couldn't we have the resolver just always return a `Path` already rendered?

This might be a future PR, but it just seems like perhaps we've overthought this and could simplify...

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
AlfredoG87: Consensus node uses `.gz` as compression type, so files are `00000001.blk.gz` 
Also thinking about this, it would be useful if we can implement several compression algos, since there are some use cases where we want to extract the `.gz` files out of the block_node to feed the simulator or just for manual inspection, and since consensus node uses `.gz` all the tooling for blocks i made around that extension, just noting it for future consideration. definitively not something for this PR.

now the extension I think it should have the compressionType own extension at the end for claritiy and maximun self description of the file.

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
ata-nas: Good point. I guess the question we need to answer is would we ever want (currently) to persist blocks with anything other than `zstd`, that includes no compression at all? If the answer is no, I do see we could make a maintenance chore to clean this up. I also see the point to check for magic bytes to determine if the data is compressed or not, but appending the extension will give us the ability to determine that immediately. I do see both arguments here.

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
ata-nas: As mentioned in above reply, I do see the arguments for both sides. I agree that we could drop different types of compression in favor of always using `zstd` if we are absolutely sure that we never want to do anything else for now, including we will not be able to persist uncompressed data (compression type `NONE`) most likely used for dev/test. But I also agree that we can stick with the different types of compression, no matter that currently we only have no compression or zstd compression. As far as checking for magic bytes vs having an extension, also, I see both ways. I see no problem with resolving the file with an extension and using that to determine how we can read the data. Yes, that does add a touch of complexity, granted, but we should probably make the move to drop the extension only if we will agree that data can only ever be written/read using zstd, otherwise, we are shifting the complexity from appending/resolving an extension to reading magic bytes.

As an action item I see the following:
Q: Can/Should we only write/read zstd compressed data, that is including for test/dev as well as prod?
If yes, we could then drop the extension, do not check for magic bytes as of yet and just use the zstd lib always. (future PR)
If no, we should stick with what we have for now

server/src/main/java/com/hedera/block/server/persistence/storage/path/UnverifiedBlockPath.java
ata-nas: The live and archive block path were introduced initially to help differentiate where the block would come from. With a future PR #741 we will be able to use the zipfs to read an entry directly, thus making it very unlikely that we will actually need these path objects. I see that we can make a housekeeping item for this so we can directly resolve an `Optional<Path>` instead, but the other use for these detailed records is the inclusion of a compression extension, making it easier to determine how to wrap the input stream. I am 50/50 on this currently, I guess we could see what we would require in the near future and make a decision as to what to do with these.

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
jsync-swirlds: - We do not expect to use other compression algorithms.  If we did we likely would not use more than one in the same block node.
- "magic number" (it's not actually one byte anymore) isn't just "is this compressed", it's also which compression, since each algorithm generally sets a different value (and those are documented, for zstd it's `37 a4 30 ec`).
   - Many file types use an easily detected header or magic number, the `libmagic` library uses that as part of it's database to detect file types.
- checking headers is much more reliable than extension, in fact the tools and libraries almost always ignore extension and check the magic number before attempting to decompress.  We might assume the extension is correct, but we'll always need to handle that compression might fail, and may have fallbacks (e.g. try zstd, if fails try gzip, if fails try uncompressed, etc...).

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
jsync-swirlds: Extensions are a terrible choice, you always have to assume the extension may be incorrect.  At best an extension is a _suggestion_.
If we were expecting humans to explore the filesystem, I might see the point of adding extensions for compression, but we are not (and in fact no human should ever try to read our files directly, and definitely should not be modifying them).
I don't see any reason we would use gzip for the compression, given it's terribly slow and not as good compression as zstd.  If we did want to use something different, we would be much more likely to use LZMA (highest compression ratios) or LZ4 (faster than zstd). Consensus uses gzip mostly because it matches the current record stream process and the devops tooling that uploads files to buckets expects it.

For the current setup, I'd recommend assuming always ZStd and leave off extensions as unnecessary.

Note, decompression must always assume extension is incorrect.  It will have to _at least_ handle a failed decompression with fallback to _try_ uncompressed.  We'll probably add other options (e.g. try ZStd, then LZMA, then gzip, then deflate, then uncompressed) eventually, but that can be deferred to after first "production" release.

server/src/main/java/com/hedera/block/server/persistence/storage/path/UnverifiedBlockPath.java
jsync-swirlds: We definitely still need separate paths.  We need to write unverified, verified, and archives to three different locations, and zipfs does nothing to change that.

The use of Zipfs will only be _after_ finding the zip file, and _only_ for reading, not for writing files.  Zipfs is an alternative to reading one zip file, not an alternative to the _general_ filesystem (that would be frighteningly slow).

Regarding extension, that would just be in the Path object.  It's not hard to retrieve an extension from a Path if you need to; but it is also mostly irrelevant.  The code to read files will always need to attempt the configured compression first, then attempt all other available compressions if the configured option fails.  Writing always uses configured compression, so extension really doesn't matter once the Path is determined.

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
```diff
@@ -80,21 +88,68 @@ public StreamPersistenceHandlerImpl(
...
+        final Path unverifiedRootPath = Objects.requireNonNull(persistenceStorageConfig.unverifiedRootPath());
         Files.createDirectories(liveRootPath);
         Files.createDirectories(archiveRootPath);
+        Files.createDirectories(unverifiedRootPath);
+
+        try (final Stream<Path> blockFilesInUnverified = Files.list(unverifiedRootPath)) {
+            // Clean up the unverified directory at startup. Any files under the
+            // unverified root at startup are to be considered unreliable
+            blockFilesInUnverified.map(Path::toFile).forEach(File::delete);
+        }
```
ata-nas, line 114: Is this deletion correct? Here we delete all files that are direct children to the unverified root at component initialization. If two or more instances of the block node have access to the same root, that would be an issue. Do we have such concerns?

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
ata-nas: @jsync-swirlds I do agree that checking for a `magic number` is far more reliable than an extension. Based on what we have discussed, I think indeed we could simplify a lot the current codebase, given we accept that for now we will only need `zstd` compression. We can then do the following:
- drop all compression types but `zstd` and only ever use that for now
- drop all extensions (this does indeed simplify and make less error prone the whole path resolution)
- drop the wrapper classes for live, archive and unverified path, simplifying them to a simple resolved `Path`

I have created #768 as a follow up. Let's close with what we have and improve in a follow up.

server/src/main/java/com/hedera/block/server/persistence/storage/path/BlockAsLocalFilePathResolver.java
ata-nas: See comment above, #768 is created as a follow up.

server/src/main/java/com/hedera/block/server/persistence/storage/path/UnverifiedBlockPath.java
ata-nas: Yeah, what I meant was that the differently typed paths would allow me to type safely determine where the block is, initially I did not see it simpler, resolving a wrapper object that has components of the path broken down made a lot more sense because I can pick and choose what to do later. With simplifications that are to come, meaning the removal of extensions, using zipfs for reading etc. etc. these become less and less needed, relying on the correct method call to determine where the block would be present. (I still kind of like the type safety though).

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
jsync-swirlds: 1. There should never be two block nodes running on the same roots, that would create all sorts of chaos.
2. The unverified root must be ephemeral, it's a mechanism to allow persistence and verification to proceed more-or-less independently without needing to keep blocks (and threads) active and in memory.  It also supports a few possible sue cases around multi-archive and allowing plugin services, but those are secondary, at most.
   * Even if we had two block nodes, they could not share an _unverified_ root, as they would have to be in different containers, with different filesystem namespaces.

ata-nas: @jsync-swirlds my thoughts as well, relying on the ephemeral nature of the unverified root and the fact that two block nodes really should not be running on the same roots will ensure proper isolation of the data. Do you believe we need to take any immediate action here or the pr could be resolved? For now I have put the default unverified root under `/opt`, maybe that could be under `/tmp`. What would be your suggestion or we should proceed like so. In any case, when running the actual system these should be resolved via ENV vars for specifics.

jsync-swirlds: Eventually, we'll want the _default_ and _compiled in_ configuration to match production, but that should be done closer to when we make a production-candidate release.

Using `/opt` is fine, as we'll be mounting multiple filesystems to the various folders there.

I believe this delete is correct and we can proceed with that understanding.

Resolving your comment is up to you 😄 .

server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
```diff
@@ -63,7 +70,7 @@ public void blockPersisted(@NonNull final BlockPersistenceResult blockPersistenc
                 final BlockInfo info = blockInfo.computeIfAbsent(blockNumber, BlockInfo::new);
                 info.getBlockStatus().setPersisted();
             } else {
-                // @todo(545) handle other cases for the blockPersistenceResult
+                // @todo(743) handle other cases for the blockPersistenceResult
```
jsync-swirlds, line 73: Could we update this to 774?  We needed to have creating the code and _using_ the code be different sprints due to the necessity to make the protobuf change in a different repo then wait for their release process to get the updated protocol buffer API specification.

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
AlfredoG87: Yes I agree that deleting `unverified` files makes sense (in any case those would be partial and/or unverified blocks)

AlfredoG87 (dismissed):
LGTM 💯

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
ata-nas: @jsync-swirlds yeah, I was just unsure if you were just making statements or were hinting to a change needed to be made, hence my ask 😄.

@AlfredoG87 yep, the deletion on startup is a must indeed!

So, the root will remain under `/opt` by default, we agree that the delete is correct as it is.

server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
ata-nas: ✅ Sure thing! Done.

server/src/main/java/com/hedera/block/server/persistence/StreamPersistenceHandlerImpl.java
ata-nas: @jsync-swirlds @AlfredoG87 in addition I have made a couple of tiny adjustments. I am now calling the archiver to notify for a persisted block after the move has been successful (inside the moveVerified method). Couple of reasons, first of, we will release the RB thread even faster, not having to go through the archiver on the RB thread. The move verified logic is called from within the thread that runs the writer task. Also, it is much more accurate to call the archiver from there, because we know (current implementation) that the calls to move verified will be ordered and we are absolutely sure that the block that was just now moved is indeed persisted && verified correctly. Also, a small improvement to logging, properly logging stack traces where applicable. These changes can be seen here c0fa3ca43e77bb1f82046a87ca4139949709d148.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
```diff
@@ -71,6 +71,9 @@ public Optional<BlockUnparsed> read(final long blockNumber) throws IOException,
                 final ArchiveBlockPath archiveBlockPath = optArchivedBlock.get();
                 final Path zipFilePath = archiveBlockPath.dirPath().resolve(archiveBlockPath.zipFileName());
                 final BlockUnparsed value;
+                // @todo(736) (depends on 598!) update reader to be able to read
+                //    appended blocks
```
jsync-swirlds, line 75: This todo can be removed.

jsync-swirlds (dismissed):

server/src/main/java/com/hedera/block/server/ack/AckHandlerImpl.java
```diff
@@ -159,8 +163,12 @@ private void attemptAcks() {
...
                     //   again. What would be the best way to hande inability to move the block with
                     //   the limitations we current have?
-                    // @todo(743) we should think of a response code for unsuccessful persistence
-                    //   this is outside of the scope of this PR
+                    // @todo(774) we should use a response code for failed persistence here
+                    LOGGER.log(
+                            ERROR,
+                            "Failed to move Block with number [%d] from unverified to live storage"
+                                    .formatted(nextBlock),
+                            e);
```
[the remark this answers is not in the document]
ata-nas: ✅ Done, readability is improved.

server/src/main/java/com/hedera/block/server/persistence/storage/read/BlockAsLocalFileReader.java
ata-nas: ✅ Done, also removed the one for the writer.

jsync-swirlds (approved):

georgi-l95 (dismissed):

georgi-l95 (approved):

ata-nas:
@jsync-swirlds @georgi-l95 @AlfredoG87 as a last minute cleanup: 0c9dd45bfd0c063c5b0ed3fb081e18c28c1e6a4f

What I did in this change was to simply remove some redundant (and in a sense wrong, but not wrong in a "this should not happen way", but rather in a "we do not even need to do this, it is redundant" way) logic which would delete if exists the unverified block file. The reason for this is that unverified block files should be rewritable, the writer is agnostic to the decision making as to should/shouldn't it rewrite the contents of an existing unverified block file. It simply always does. Whoever created the writer task is responsible for potential race conditions and/or undesired behavior if two tasks that act upon a single unverified block file (with the same block number) exist and run in parallel. Also, I added a test to assert that the rewrite happens. Simply, whenever we use `Files.newOutputStream(path)` it uses default options which are write, create, truncate - this means that if a file does not exist it will be created. If it exists, it will be truncated to 0 bytes and then it will be written the new data.

PR #720, opened 2025-02-25
fix: Exclude `suites` and `tools` from CodeCov Coverage report

### LINKED ISSUE #717: Fix CodeCov CI follow-up
After the CodeCov re-enablement, Project Coverage is failing due to not meeting the previously set threshold of `80%` (currently standing at `~61%`).

This is due in part to PRs merged during the time it was not present lacked the minimum `80%` for their respective changes, but also due to some sub-packages that should not be taken into consideration for UTs. such as `tools` and `suites`.

AlfredoG87:
## Reviewer Notes
Exclude `suites` and `tools` from CodeCov Coverage Report

## Related Issue(s)
Fixes #717

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -24,3 +24,5 @@ ignore:
   - "server/src/main/java/com/hedera/block/server/producer/NoOpProducerObserver.java"
   - "server/src/main/java/com/hedera/block/server/persistence/storage/write/NoOpBlockWriter.java"
   - "simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulator.java"
+  - "suites/**"
+  - "tools/**"


codecov.yml
```diff
@@ -24,3 +24,5 @@ ignore:
   - "server/src/main/java/com/hedera/block/server/producer/NoOpProducerObserver.java"
   - "server/src/main/java/com/hedera/block/server/persistence/storage/write/NoOpBlockWriter.java"
   - "simulator/src/main/java/com/hedera/block/simulator/BlockStreamSimulator.java"
+  - "suites/**"
+  - "tools/**"
```
jsync-swirlds, line 28: Should we exclude documentation, or is that automatic?

jjohannes:
@AlfredoG87 you may consider doing the exclude on the Gradle side. Then the local report corresponds to what we see in CodeCov.  You can find the local report in 
`gradle/aggregation/build/reports/jacoco/testCodeCoverageReport/html/index.html`
after running `./gradlew check`.

You can copy this code. It's a bit involved unfortunately. If we start to use this here and in other projects, I will have a look to simplify this through the plugins.

https://github.com/hiero-ledger/hiero-consensus-node/blob/e9ee21ade19df292c3cb6fdd569b03d1290fcd7d/gradle/aggregation/build.gradle.kts#L25-L47

codecov.yml
AlfredoG87: Since they are not "code" they don't show as needed for Test Coverage. so yeah, is automatic.

ata-nas (approved):
Yaml seems to do the job. We could also try @jjohannes's suggestion.

PR #722, opened 2025-02-26
feat: allow for configuration of the async writers executors

### LINKED ISSUE #587: Allow for configuration of the async writers' executor(s)
### Persona

As a Block Node Operator/Developer

### Request

I want to have configurable values for the thread pool executor that is inside the async writer

### Goal

So that I can have control over the executor's internal properties when it is being created

### Technical Notes

- configurable values need to exist so that the executor (the one that is being submitted async persistence tasks) that is internal to the persistence

georgi-l95:
This PR adds a configurable executor for async writers in the persistence storage module. The implementation allows for different executor types and configurations to optimize performance for various workloads and deployment environments.
#### Key Changes:
- Added `ExecutorType` enum to `PersistenceStorageConfig` with options for `THREAD_POOL`, `SINGLE_THREAD`, and `FORK_JOIN` executors
- Added configuration properties for thread count, thread keep-alive time, virtual threads support, and execution queue limit
- Implemented `AsyncWriterExecutorFactory` to create appropriate executor instances based on configuration
- Updated `PersistenceInjectionModule` to use the factory for creating executors
- Added unit tests for all new components

#### Configuration Options:
- `persistence.storage.executionQueueLimit`: Maximum queue size for pending tasks (64-2048)
- `persistence.storage.executorType`: Type of executor (THREAD_POOL, SINGLE_THREAD, FORK_JOIN)
- `persistence.storage.threadCount`: Number of threads for thread pool (1-16)
- `persistence.storage.threadKeepAliveTime`: Keep-alive time in seconds for idle threads
- `persistence.storage.useVirtualThreads`: Whether to use virtual threads (boolean)
## Related Issue(s)
Fixes #587

### DIFF
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
@@ -29,7 +29,12 @@ public record PersistenceStorageConfig(
         @Loggable @ConfigProperty(defaultValue = "BLOCK_AS_LOCAL_FILE") StorageType type,
         @Loggable @ConfigProperty(defaultValue = "ZSTD") CompressionType compression,
         @Loggable @ConfigProperty(defaultValue = "3") @Min(0) @Max(20) int compressionLevel,
-        @Loggable @ConfigProperty(defaultValue = "1_000") @Min(10) int archiveGroupSize) {
+        @Loggable @ConfigProperty(defaultValue = "1_000") @Min(10) int archiveGroupSize,
+        @Loggable @ConfigProperty(defaultValue = "true") boolean archiveEnabled,
+        @Loggable @ConfigProperty(defaultValue = "THREAD_POOL") ExecutorType executorType,
+        @Loggable @ConfigProperty(defaultValue = "6") @Min(1) @Max(16) int threadCount,
+        @Loggable @ConfigProperty(defaultValue = "false") boolean useVirtualThreads,
+        @Loggable @ConfigProperty(defaultValue = "1024") @Min(64) @Max(2048) int executionQueueLimit) {
     /**
      * Constructor.
      */
@@ -72,6 +77,39 @@ public record PersistenceStorageConfig(
          * This type of storage does nothing.
          */
         NO_OP
+    }
+
+    /**
+     * An enum that defines the type of executor to use for async writers.
+     * <p>
+     * Different executor types have different performance characteristics and are suitable
+     * for different workloads and deployment environments.
+     */
+    public enum ExecutorType {
+        /**
+         * A fixed-size thread pool executor that maintains a specified number of threads.
+         * Recommended for environments where consistent performance and resource usage are important.
+         */
+        THREAD_POOL,
+
+        /**
+         * An executor with a single worker thread.
+         * Suitable for low-throughput environments or when strict ordering of task execution is required.
+         */
+        SINGLE_THREAD,
+
+        /**
+         * Uses the common ForkJoinPool for task execution.
+         * Best for compute-intensive workloads that benefit from work-stealing and when
+         * you want to share thread resources with other components in the application.
+         */
+        FORK_JOIN,
+
+        /**
+         * Executes tasks directly in the calling thread without creating additional threads.
+         * Useful for testing, debugging, or when immediate execution in the current thread is desired.
+         */
+        CALLING_THREAD
     }
 
     /**
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
@@ -30,11 +30,11 @@ public record PersistenceStorageConfig(
         @Loggable @ConfigProperty(defaultValue = "ZSTD") CompressionType compression,
         @Loggable @ConfigProperty(defaultValue = "3") @Min(0) @Max(20) int compressionLevel,
         @Loggable @ConfigProperty(defaultValue = "1_000") @Min(10) int archiveGroupSize,
-        @Loggable @ConfigProperty(defaultValue = "true") boolean archiveEnabled,
+        @Loggable @ConfigProperty(defaultValue = "1024") @Min(64) @Max(2048) int executionQueueLimit,
         @Loggable @ConfigProperty(defaultValue = "THREAD_POOL") ExecutorType executorType,
         @Loggable @ConfigProperty(defaultValue = "6") @Min(1) @Max(16) int threadCount,
-        @Loggable @ConfigProperty(defaultValue = "false") boolean useVirtualThreads,
-        @Loggable @ConfigProperty(defaultValue = "1024") @Min(64) @Max(2048) int executionQueueLimit) {
+        @Loggable @ConfigProperty(defaultValue = "60000") @Min(0) long threadKeepAliveTime,
+        @Loggable @ConfigProperty(defaultValue = "false") boolean useVirtualThreads) {
     /**
      * Constructor.
      */
@@ -43,6 +43,7 @@ public record PersistenceStorageConfig(
         Objects.requireNonNull(archiveRootPath);
         Objects.requireNonNull(unverifiedRootPath);
         Objects.requireNonNull(type);
+        Objects.requireNonNull(executorType);
         compression.verifyCompressionLevel(compressionLevel);
         // @todo(742) verify that the group size has not changed once it has
         //    been set initially
@@ -53,6 +54,15 @@ public record PersistenceStorageConfig(
         Preconditions.requirePositivePowerOf10(
                 archiveGroupSize,
                 "persistence.storage.archiveGroupSize [%d] is required to be a positive power of 10.");
+        Preconditions.requirePositivePowerOf10(archiveGroupSize);
+        Preconditions.requireWhole(threadKeepAliveTime);
+        Preconditions.requireInRange(
+                executionQueueLimit,
+                64,
+                2048,
+                "persistence.storage.executionQueueLimit [%d] is required to be between [%d] and [%d].");
+        Preconditions.requireInRange(
+                threadCount, 1, 16, "persistence.storage.threadCount [%d] is required to be between [%d] and [%d].");
     }
 
     /**
@@ -99,17 +109,10 @@ public record PersistenceStorageConfig(
         SINGLE_THREAD,
 
         /**
-         * Uses the common ForkJoinPool for task execution.
-         * Best for compute-intensive workloads that benefit from work-stealing and when
-         * you want to share thread resources with other components in the application.
+         * Uses the ForkJoinPool for task execution.
+         * Best for compute-intensive workloads that benefit from work-stealing.
          */
         FORK_JOIN,
-
-        /**
-         * Executes tasks directly in the calling thread without creating additional threads.
-         * Useful for testing, debugging, or when immediate execution in the current thread is desired.
-         */
-        CALLING_THREAD
     }
 
     /**
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
@@ -1,0 +1,163 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.block.server.persistence.storage.write;
+
+import static java.lang.System.Logger.Level.TRACE;
+import static java.util.Objects.requireNonNull;
+
+import com.hedera.block.server.persistence.storage.PersistenceStorageConfig;
+import com.hedera.block.server.persistence.storage.PersistenceStorageConfig.ExecutorType;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.ThreadPoolExecutor.CallerRunsPolicy;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Factory for creating executor instances for async writers based on configuration.
+ * <p>
+ * This factory creates different types of executors with specific configurations
+ * to optimize performance for different workloads and deployment environments.
+ * <p>
+ * The factory supports creating:
+ * <ul>
+ *   <li>Thread pool executors with configurable thread count and queue size
+ *   <li>Single-threaded executors for sequential processing
+ *   <li>Fork-join pool executors for work-stealing parallelism
+ * </ul>
+ */
+public final class AsyncWriterExecutorFactory {
+    private static final System.Logger LOGGER = System.getLogger(AsyncWriterExecutorFactory.class.getName());
+    private static final int BASE_THREADS = 4;
+
+    // Private constructor to prevent instantiation
+    private AsyncWriterExecutorFactory() {}
+
+    /**
+     * Creates an executor based on the provided configuration.
+     * <p>
+     * This method creates the appropriate executor type with the specified parameters
+     * depending on the configuration.
+     *
+     * @param config the persistence storage configuration containing executor settings
+     * @return an executor configured according to the settings
+     * @throws NullPointerException if config is null
+     */
+    @NonNull
+    public static Executor createExecutor(@NonNull final PersistenceStorageConfig config) {
+        requireNonNull(config);
+
+        LOGGER.log(TRACE, "Creating async writer executor of type: {0}", config.executorType());
+        final ExecutorType executorType = config.executorType();
+        return switch (executorType) {
+            case THREAD_POOL -> createThreadPoolExecutor(config);
+            case SINGLE_THREAD -> createSingleThreadExecutor();
+            case FORK_JOIN -> createForkJoinExecutor(config);
+        };
+    }
+
+    /**
+     * Creates a thread pool executor with the specified configuration.
+     * <p>
+     * The thread pool uses:
+     * <ul>
+     *   <li>Fixed number of threads as specified in the configuration
+     *   <li>Custom thread factory that creates named daemon threads
+     *   <li>Configurable keep alive time for threads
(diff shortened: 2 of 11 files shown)


server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -66,6 +71,33 @@ public enum StorageType {
...
+        THREAD_POOL,
+
+        /**
+         * An executor with a single worker thread.
+         * Suitable for low-throughput environments or when strict ordering of task execution is required.
+         */
+        SINGLE_THREAD,
+
+        /**
+         * Uses the common ForkJoinPool for task execution.
```
jsync-swirlds, line 94: This should use _a_ fork/join pool, but should not use the _common_ fork/join pool.
The reason for this is that the common pool is almost always overused and in container environments often ends up having much fewer threads than it should.
In my opinion, we should, if this is selected, create a separate fork/join pool with a thread count set by the threadCount config value.

server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @return the common fork-join pool
+     */
+    @NonNull
+    private static ForkJoinPool createForkJoinExecutor() {
+        final ForkJoinPool forkJoinPool = ForkJoinPool.commonPool();
+        LOGGER.log(
+                TRACE,
+                "Using common fork-join pool with parallelism level: %d".formatted(forkJoinPool.getParallelism()));
+        return forkJoinPool;
```
jsync-swirlds, line 134: We should not use the common pool here. We should treat this very much like the thread pool case, just using a fork/join pool instead of a "simple" thread pool.

Also, this method should also react to the virtual threads config by using a thread factory that creates virtual threads if that flag is set.

We also probably want to set the `async` mode on the pool to `true`, rather than the default of `false`.  The default tends to not run tasks until the result is requested or polled, which increases latency.

server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private static class AsyncWriterThreadFactory implements ThreadFactory {
+        private final AtomicInteger threadNumber = new AtomicInteger(1);
+
+        @Override
+        public Thread newThread(@NonNull final Runnable r) {
+            return new Thread(r, "async-writer-" + threadNumber.getAndIncrement());
+        }
+    }
+}
```
jsync-swirlds, line 145: It's not clear what we need this for.  The default thread factory should work well (unless we want virtual threads).

jsync-swirlds (commented):
A few items, mostly relating to the fork/join option.

server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
ata-nas: I do see some benefit in having a custom thread name, it would make it easier to look at a thread dump in order to see which thread is running a writer.

server/src/test/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfigTest.java
```diff
@@ -76,7 +77,12 @@ void testPersistenceStorageConfigStorageTypes(final StorageType storageType) {
                 CompressionType.NONE,
                 DEFAULT_COMPRESSION_LEVEL,
                 DEFAULT_ARCHIVE_ENABLED,
-                DEFAULT_ARCHIVE_BATCH_SIZE);
+                DEFAULT_ARCHIVE_BATCH_SIZE,
+                PersistenceStorageConfig.ExecutorType.THREAD_POOL,
+                6,
+                60,
+                false,
+                2048);
```
ata-nas, line 85: I would suggest that you save these values in a field that would be easily reused. The values should be the config's default values. Reasoning is that first of all, the values are reused. A change in the values that would break the tests with the current value set will only have to be done in a single place (if requirements change so that an assert will no longer pass, the tests need to be adapted).

Consider for all places.

server/src/test/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfigTest.java
```diff
@@ -169,10 +190,132 @@ void testPersistenceStorageConfigCompressionTypes(final CompressionType compress
...
+                DEFAULT_ARCHIVE_ENABLED,
+                DEFAULT_ARCHIVE_BATCH_SIZE,
+                PersistenceStorageConfig.ExecutorType.THREAD_POOL,
+                6,
+                60,
+                false,
+                queueLimit);
+        assertThat(actual).returns(queueLimit, from(PersistenceStorageConfig::executionQueueLimit));
+    }
+
```
ata-nas, line 318: It would be a good idea to parameterize these tests and to supply additional values as well as edge cases. See other tests in the same class for simple examples. The more cases we test the better.

server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+        final PersistenceStorageConfig config = createConfig(ExecutorType.THREAD_POOL, true, 4, 60L, 100);
+
+        // When
+        final Executor executor = AsyncWriterExecutorFactory.createExecutor(config);
+
+        // Then
+        assertNotNull(executor);
+        System.out.println(executor.getClass().getName());
+        // Virtual thread executor invokes underneath ThreadPerTaskExecutor with specific factory for virtual
+        assertTrue(executor.getClass().getName().contains("ThreadPerTaskExecutor"));
```
ata-nas, line 74: I am not sure what this test asserts? Judging by the name, it asserts the creation of an executor with type THREAD_POOL? If that is so, the above test already covers that with the enum source parameterized test.

server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+                PersistenceStorageConfig.StorageType.BLOCK_AS_LOCAL_FILE,
+                PersistenceStorageConfig.CompressionType.NONE,
+                3, // Default compression level
+                true, // Default archive enabled
+                1000, // Default archive batch size
+                executorType,
+                threadCount,
+                threadKeepAliveTime,
+                useVirtualThreads,
+                executionQueueLimit);
```
ata-nas, line 126: @jsync-swirlds mentioned something about test dependency for creating configurations for tests, coming from services config dependency. For a future cleanup, we should take a look at that and use that on our side if applicable. For now, we have the test config builder which we could utilize:

```java
    // a sample code snippet that would help give you the configs you need before each test
    @BeforeEach
    void setUp() throws IOException {
        final Configuration config = new TestConfigBuilder(PersistenceStorageConfig.class)
                .withValue(PERSISTENCE_STORAGE_COMPRESSION_TYPE, "NONE")
                .withValue(PERSISTENCE_STORAGE_ARCHIVE_BATCH_SIZE, String.valueOf(ARCHIVE_GROUP_SIZE))
                .withValue(PERSISTENCE_STORAGE_LIVE_ROOT_PATH_KEY, testTempDir.resolve("live"))
                .withValue(PERSISTENCE_STORAGE_ARCHIVE_ROOT_PATH_KEY, testTempDir.resolve("archive"))
                .getOrCreateConfig();
        persistenceStorageConfig = config.getConfigData(PersistenceStorageConfig.class);
    }
```

You can consider to using this because it is very easy to setup with the specific values you actually need, the rest will be resolved with default values automatically, thus making your setup a lot easier and shorter.

ata-nas (commented):
Good job @georgi-l95. I leave some nits, suggestions for cleanup and some general thoughts.

server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
georgi-l95: I added it for the exact reason @ata-nas mentioned. But I will remove it, if the team decides 👍

jsync-swirlds: It's not a _bad_ item; just more maintenance for a very small possible use case.

If we keep this, it _should_ follow the virtual threads setting, however.

server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
```diff
@@ -0,0 +1,145 @@
...
+                TRACE,
+                "Creating thread pool executor with %d threads and queue limit of %d"
+                        .formatted(threadCount, queueLimit));
+
+        return new ThreadPoolExecutor(
+                threadCount,
+                threadCount,
+                threadKeepAliveTime,
+                TimeUnit.SECONDS,
+                new LinkedBlockingDeque<>(queueLimit),
```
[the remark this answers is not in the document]
jsync-swirlds: In most cases, linked is a more efficient choice here.
Having to rewrite arrays generally costs more than the slight memory overhead.

server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactoryTest.java
jsync-swirlds: The [latest upgrade](https://github.com/hiero-ledger/hiero-block-node/pull/761) to the consensus libraries has already replaced the `TestConfigBuilder` with `ConfigurationBuilder`, which works more-or-less the same.

server/src/test/java/com/hedera/block/server/persistence/PersistenceInjectionModuleTest.java
```diff
@@ -73,6 +75,12 @@ class PersistenceInjectionModuleTest {
     @Mock
     private Executor executorMock;
 
+    @Mock
+    private BlockNodeContext blockNodeContextMock;
+
+    @Mock
+    private Configuration configurationMock;
+
```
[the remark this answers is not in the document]
jsync-swirlds: I definitely support dropping as many mocks as possible.
The less we mock, the more accurate our test results will be.

server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactoryTest.java
AlfredoG87: Yes, if you rebase to the latest `main` head you will no longer be able to use the `TestConfigBuilder`

AlfredoG87 (commented):
Looking good 👍

server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactoryTest.java
ata-nas: Yes, and thanks @AlfredoG87 for introducing the change on our side!

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
georgi-l95: Yes, you are right. I've changed the implementation to use dedicated pool.

server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
georgi-l95: Thank you for catching this. Changed the implementation to use dedicated pool. 👍

server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
georgi-l95: Yes, you're right. Added functionality to respect the virtual threads setting.

server/src/main/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfig.java
```diff
@@ -30,7 +30,12 @@ public record PersistenceStorageConfig(
         @Loggable @ConfigProperty(defaultValue = "ZSTD") CompressionType compression,
         @Loggable @ConfigProperty(defaultValue = "3") @Min(0) @Max(20) int compressionLevel,
         @Loggable @ConfigProperty(defaultValue = "true") boolean archiveEnabled,
-        @Loggable @ConfigProperty(defaultValue = "1_000") int archiveGroupSize) {
+        @Loggable @ConfigProperty(defaultValue = "1_000") int archiveGroupSize,
+        @Loggable @ConfigProperty(defaultValue = "THREAD_POOL") ExecutorType executorType,
+        @Loggable @ConfigProperty(defaultValue = "6") @Min(1) @Max(16) int threadCount,
+        @Loggable @ConfigProperty(defaultValue = "60") @Min(0) long threadKeepAliveTime,
+        @Loggable @ConfigProperty(defaultValue = "false") boolean useVirtualThreads,
+        @Loggable @ConfigProperty(defaultValue = "1024") @Min(64) @Max(2048) int executionQueueLimit) {
```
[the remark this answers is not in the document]
georgi-l95: Added preconditions, thank you 👍

server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @param config the persistence storage configuration containing executor settings
+     * @return an executor configured according to the settings
+     * @throws NullPointerException if config is null
+     */
+    @NonNull
+    public static Executor createExecutor(@NonNull final PersistenceStorageConfig config) {
+        requireNonNull(config);
+
+        LOGGER.log(TRACE, "Creating async writer executor of type: %s".formatted(config.executorType()));
```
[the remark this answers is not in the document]
georgi-l95: Changed to this formating, now that it's fixed 👍

server/src/main/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactory.java
```diff
@@ -0,0 +1,145 @@
...
+        LOGGER.log(
+                TRACE,
+                "Creating thread pool executor with %d threads and queue limit of %d"
+                        .formatted(threadCount, queueLimit));
+
+        return new ThreadPoolExecutor(
+                threadCount,
+                threadCount,
+                threadKeepAliveTime,
+                TimeUnit.SECONDS,
```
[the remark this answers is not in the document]
georgi-l95: You're right, changed.

server/src/test/java/com/hedera/block/server/config/ServerMappedConfigSourceInitializerTest.java
```diff
@@ -59,6 +59,11 @@ class ServerMappedConfigSourceInitializerTest {
         new ConfigMapping("persistence.storage.type", "PERSISTENCE_STORAGE_TYPE"),
         new ConfigMapping("persistence.storage.archiveEnabled", "PERSISTENCE_STORAGE_ARCHIVE_ENABLED"),
         new ConfigMapping("persistence.storage.archiveGroupSize", "PERSISTENCE_STORAGE_ARCHIVE_BATCH_SIZE"),
+        new ConfigMapping("persistence.storage.executorType", "PERSISTENCE_STORAGE_EXECUTOR_TYPE"),
+        new ConfigMapping("persistence.storage.threadCount", "PERSISTENCE_STORAGE_THREAD_COUNT"),
+        new ConfigMapping("persistence.storage.threadKeepAliveTime", "PERSISTENCE_STORAGE_THREAD_KEEP_ALIVE_TIME"),
+        new ConfigMapping("persistence.storage.useVirtualThreads", "PERSISTENCE_STORAGE_USE_VIRTUAL_THREADS"),
+        new ConfigMapping("persistence.storage.executionQueueLimit", "PERSISTENCE_STORAGE_EXECUTION_QUEUE_LIMIT"),
```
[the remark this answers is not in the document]
georgi-l95: Good idea, done.

server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactoryTest.java
georgi-l95: You're right, modified it to use the api you've mentioned.

server/src/test/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfigTest.java
georgi-l95: True, done.

server/src/test/java/com/hedera/block/server/persistence/PersistenceInjectionModuleTest.java
```diff
@@ -204,4 +212,33 @@ void testProvidesStreamValidatorBuilder() throws IOException {
...
+                PersistenceInjectionModule.providesBlockNodeEventHandler(
+                        subscriptionHandlerMock,
+                        notifierMock,
+                        blockNodeContextMock,
+                        serviceStatusMock,
+                        ackHandlerMock,
+                        asyncBlockWriterFactoryMock);
+
+        // Then
+        assertNotNull(handler);
```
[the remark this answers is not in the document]
georgi-l95: You can check again, a lot have changed after the rebase 😆

server/src/test/java/com/hedera/block/server/persistence/storage/PersistenceStorageConfigTest.java
georgi-l95: Yep, done. Thank you

server/src/test/java/com/hedera/block/server/persistence/storage/write/AsyncWriterExecutorFactoryTest.java
georgi-l95: Thanks all, done.

ata-nas (approved):
LGTM. 🔥

jsync-swirlds (approved):

ata-nas (approved):
Good job! :fire:

PR #733, opened 2025-02-28
POC: Block item proof POC

AlfredoG87:
## Reviewer Notes

This is a WIP POC.


AlfredoG87:
@ata-nas  Indeed I have considered it, and although yes, we will not merge this PR, there are some POC stuff here that we might be able to re-use and make the dev of the feature faster once we get back to it.
That is why am adamant of closing the PR, since then the branch might get deleted and the POC code could get lost.

PR #745, opened 2025-02-28
fix: Improvements and fixes to Helm charts and docker compose

### LINKED ISSUE #630: Set-up Basic Simulator and Test Scenarios
Set up a _basic_ minimal simulator and test scenarios for the block node deployed to a testing Kubernetes cluster.  This should be _just enough_ to validate the deployment. Detailed and extensive testing (and the development or documentation of formal testing scenarios) is planned for a later effort.

### LINKED ISSUE #726: Deploy latest version of blockstream-simulator as a Consumer to GCP K8 DevEnv.
### Technical Notes

We may also wish to experiment with different replica counts to simulate more consumers.

### LINKED ISSUE #744: Improvements: Various findings during test deployment to K8s
As part of the deployment experience there have been found some initial set of needed or desired improvements, such as: 
- missed changes needed in Helm charts for correct deployment.
- missed changes to log formatter on helm charts deployment.

AlfredoG87:
## Reviewer Notes
- Changes to the Helm Charts for deployment
- Fix to the ENV defaults on the docker compose dev env
- Consumer Mode support for the deployment of multiple consumers using the same simulator chart.

Fixes #744 
Fixes #726 
Fixes #630


charts/block-node-server/values.yaml
```diff
@@ -59,7 +59,7 @@ ingress:
 resources:
   requests:
     cpu: "2"
-    memory: "8Gi"
+    memory: "16Gi"
```
jsync-swirlds, line 62: This sets container ram the same as the Java max below, which will cause problems.

charts/block-node-server/values.yaml
```diff
@@ -73,8 +73,14 @@ blockNode:
   config:
     # Add any additional env configuration here
     # key: value
-    BLOCKNODE_STORAGE_ROOT_PATH: "/app/storage"
     JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties"
+    JAVA_OPTS: "-Xms8G -Xmx16G"
```
jsync-swirlds, line 77: It's usually best to set `Xms` and `Xmx` to the same value for servers; particularly containerized servers.

jsync-swirlds (commented):
Just a couple items, the most concerning being the overallocation of container RAM.

charts/block-node-server/values.yaml
AlfredoG87: Yes, you are reading it correctly.

I will update to reflect the changes. You are right! 👍 

✅ Done!

charts/block-node-server/values.yaml
AlfredoG87: Yes, I also considered but you recommending this is confirmation. 
Agreed.

✅ Done!

jsync-swirlds (approved):

PR #749, opened 2025-03-01
fix: downgrade of upgrade found to be the culprit of Build broke.

AlfredoG87:
## Reviewer Notes

This PR aims to rollback dependency upgrade: https://github.com/hiero-ledger/hiero-block-node/pull/695

To allow the Build process in `main` to be fixed/unblocked, while a deeper investigation on why exactly this dependency upgrade has broken the build process, specifically for `Hedera-Protobuf` stream package.

Pending: another Issue to continue with the investigation and if possible, re-upgrade to newest version.


AlfredoG87:
This was fixed by itself with next merges. no need to rollback 👍

PR #761, opened 2025-03-03
build(deps): Upgrade Swirlds dependency to `0.59.2`

AlfredoG87:
## Reviewer Notes

- Upgrade of `swirldsVersion` to `0.59.2`
- Removal of `TestConfigBuilder` a Test Class to simulate loading config used by the consensus node and using the PlatformSDK stack.  This was originally made to be able to load test configurations instances on Unit Tests.
- Refactor of places where this Class was used to use a simpler more straightforward approach that is now available.
- The `TestConfigBuilder` was a Class copy-pasted from `services` but now we don't need it anymore, and this class was using a deprecated ConfigurationHolder that was being used as starter for configuration. on This version that Class has been removed, so is forcing us to refactor if we want to support the latest and newer versions of the Platform SDK releases.

Fixes: #755


server/src/test/java/com/hedera/block/server/config/TestConfigBuilder.java
jsync-swirlds, line 1: It appears github thinks this file is still present, just empty.
Is that intended, or a github issue?

jsync-swirlds (approved):

server/src/test/java/com/hedera/block/server/config/TestConfigBuilder.java
AlfredoG87: seems like a weird git issue, just doubled check and the file is removed. 🤔

PR #766, opened 2025-03-04
chore: improvements for test repeatability

### LINKED ISSUE #642: chore: improvements for test repeatability
### Persona

As a Block Node Developer

### Request

I want to have repeatable tests whenever possible

### Goal

So that the tests are not flaky and will not hinder CI/CD

### Technical Notes

Since the development of new async block writers, it became evident that there are some tests failing because some logic runs async. Simply by the time we reach the asserts, the logic that would produce the desired results for the asserts is not yet finished. We should make these (and similar) tests repeatable in the sense that we should *ensure* that the async tasks have enough time to complete before we do the asserts.
- an idea is to inject an executor that will allow us to wait for the tasks to complete, blocking the test thread

Tests that were disabled and need migration:
- `com.hedera.block.server.pbj.PbjBlockStreamServiceIntegrationTest#testFullProducerConsumerHappyPath`
- `com.hedera.block.server.pbj.PbjBlockStreamServiceIntegrationTest#testPublishBlockStreamRegistrationAndExecution`
- `com.hedera.block.server.mediator.LiveStreamMediatorImplTest#testMediatorPersistenceWithoutSubscribers`

Considerations:
- based on [this thread](https://github.com/hashgraph/hedera-block-node/pull/545#discussion_r1945240035) consider to use jimfs and/ or mocks

AlexKehayov:
## Reviewer Notes
- Added a custom BlockingExecutorService to ensure async logic in tests executes and completes before assertions begin.
- Used CountDownLatch technique to synchronize execution in cases where BlockingExecutorService cannot be properly injected.
- Modified PbjBlockStreamServiceProxy to enable custom executor injection for test purposes.

## Related Issue(s)
 Fixes #642

### DIFF
--- a/server/src/main/java/com/hedera/block/server/pbj/PbjBlockStreamServiceProxy.java
+++ b/server/src/main/java/com/hedera/block/server/pbj/PbjBlockStreamServiceProxy.java
@@ -130,7 +130,37 @@ public class PbjBlockStreamServiceProxy implements PbjBlockStreamService {
             final @NonNull RequestOptions options,
             final @NonNull Pipeline<? super Bytes> replies) {
 
-        final var m = (BlockStreamMethod) method;
+        return openPipeline(
+                (BlockStreamMethod) method,
+                options,
+                replies,
+                openRangeHistoricStreamingExecutorService,
+                closedRangeHistoricStreamingExecutorService);
+    }
+
+    @NonNull
+    public Pipeline<? super Bytes> open(
+            final @NonNull Method method,
+            final @NonNull RequestOptions options,
+            final @NonNull Pipeline<? super Bytes> replies,
+            final @NonNull ExecutorService openRangeHistoricStreamingExecutorService,
+            final @NonNull ExecutorService closedRangeHistoricStreamingExecutorService) {
+
+        return openPipeline(
+                (BlockStreamMethod) method,
+                options,
+                replies,
+                openRangeHistoricStreamingExecutorService,
+                closedRangeHistoricStreamingExecutorService);
+    }
+
+    private Pipeline<? super Bytes> openPipeline(
+            @NonNull BlockStreamMethod method,
+            @NonNull RequestOptions options,
+            @NonNull Pipeline<? super Bytes> replies,
+            @NonNull ExecutorService openRangeHistoricStreamingExecutorService,
+            @NonNull ExecutorService closedRangeHistoricStreamingExecutorService) {
+        final var m = method;
         try {
             return switch (m) {
                 case publishBlockStream -> {
@@ -145,7 +175,11 @@ public class PbjBlockStreamServiceProxy implements PbjBlockStreamService {
                 case subscribeBlockStream -> Pipelines
                         .<SubscribeStreamRequest, SubscribeStreamResponseUnparsed>serverStreaming()
                         .mapRequest(bytes -> parseSubscribeStreamRequest(bytes, options))
-                        .method(this::subscribeBlockStream)
+                        .method((subscribeStreamRequest, helidonConsumerObserver) -> subscribeBlockStream(
+                                subscribeStreamRequest,
+                                helidonConsumerObserver,
+                                openRangeHistoricStreamingExecutorService,
+                                closedRangeHistoricStreamingExecutorService))
                         .mapResponse(reply -> createSubscribeStreamResponse(reply, options))
                         .respondTo(replies)
                         .build();
@@ -201,7 +235,9 @@ public class PbjBlockStreamServiceProxy implements PbjBlockStreamService {
      */
     void subscribeBlockStream(
             @NonNull final SubscribeStreamRequest subscribeStreamRequest,
-            @NonNull final Pipeline<? super SubscribeStreamResponseUnparsed> helidonConsumerObserver) {
+            @NonNull final Pipeline<? super SubscribeStreamResponseUnparsed> helidonConsumerObserver,
+            @NonNull final ExecutorService openRangeHistoricStreamingExecutorService,
+            @NonNull final ExecutorService closedRangeHistoricStreamingExecutorService) {
 
         LOGGER.log(DEBUG, "Executing Server Streaming subscribeBlockStream gRPC method");
 
--- a/server/src/test/java/com/hedera/block/server/mediator/LiveStreamMediatorImplTest.java
+++ b/server/src/test/java/com/hedera/block/server/mediator/LiveStreamMediatorImplTest.java
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.hedera.block.server.mediator;
 
+import static com.hedera.block.server.metrics.BlockNodeMetricTypes.Counter.BlocksPersisted;
 import static com.hedera.block.server.metrics.BlockNodeMetricTypes.Counter.LiveBlockItems;
 import static com.hedera.block.server.metrics.BlockNodeMetricTypes.Gauge.Consumers;
 import static com.hedera.block.server.util.PersistTestUtils.PERSISTENCE_STORAGE_ARCHIVE_ROOT_PATH_KEY;
@@ -32,6 +33,7 @@ import com.hedera.block.server.persistence.storage.write.AsyncNoOpWriterFactory;
 import com.hedera.block.server.service.ServiceConfig;
 import com.hedera.block.server.service.ServiceStatus;
 import com.hedera.block.server.service.ServiceStatusImpl;
+import com.hedera.block.server.util.BlockingExecutorService;
 import com.hedera.block.server.util.PersistTestUtils;
 import com.hedera.block.server.util.TestConfigUtil;
 import com.hedera.hapi.block.BlockItemSetUnparsed;
@@ -174,7 +176,9 @@ class LiveStreamMediatorImplTest {
     }
 
     @Test
-    void testMediatorPersistenceWithoutSubscribers() throws IOException {
+    void testMediatorPersistenceWithoutSubscribers() throws IOException, InterruptedException {
+        // 1 block is expected to be processed, so the expected tasks param is set to 1
+        final BlockingExecutorService executor = new BlockingExecutorService(1, 1);
         final ServiceStatus serviceStatus = new ServiceStatusImpl(serviceConfig);
         final LiveStreamMediator streamMediator = LiveStreamMediatorBuilder.newBuilder(
                         metricsService, mediatorConfig, serviceStatus)
@@ -192,7 +196,7 @@ class LiveStreamMediatorImplTest {
                 serviceStatus,
                 ackHandlerMock,
                 writerFactory,
-                executorMock,
+                executor,
                 archiverMock,
                 pathResolverMock,
                 persistenceStorageConfig);
@@ -201,13 +205,12 @@ class LiveStreamMediatorImplTest {
         // Acting as a producer, notify the mediator of a new block
         streamMediator.publish(blockItemUnparsed);
 
+        // Wait all the tasks to complete before the assertions to avoid flakiness
+        executor.waitTasksToComplete();
+
         // Verify the counter was incremented
         assertEquals(10, metricsService.get(LiveBlockItems).get());
-
-        // @todo(642) we need to employ the same technique here to inject a writer that will ensure
-        // the tasks are complete before we can verify the metrics for blocks persisted
-        // the test will pass without flaking if we do that
-        // assertEquals(1, blockNodeContext.metricsService().get(BlocksPersisted).get());
+        assertEquals(1, metricsService.get(BlocksPersisted).get());
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


server/src/test/java/com/hedera/block/server/util/TestUtils.java
```diff
@@ -43,4 +45,14 @@ public static FileAttribute<Set<PosixFilePermission>> getNoRead() {
...
+
+    public static Answer<Void> onEventLatchCountdown(CountDownLatch latch) {
+        return invocation -> {
+            if (latch.getCount() == 0) {
+                throw new IllegalStateException("Event calls exceeded");
+            }
+            latch.countDown();
+            return null;
+        };
+    }
```
AlexKehayov, line 57: This utility method is exposed for easy reuse in tests, such as testPublishBlockStreamRegistrationAndExecution().

server/src/test/java/com/hedera/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
```diff
@@ -155,14 +159,31 @@ void setUp() throws IOException {
...
     @Timeout(value = JUNIT_TIMEOUT, unit = TimeUnit.MILLISECONDS)
-    void testPublishBlockStreamRegistrationAndExecution() throws IOException {
+    void testPublishBlockStreamRegistrationAndExecution() throws IOException, InterruptedException {
         final int numberOfBlocks = 1;
-        final ExecutorService executor = Executors.newFixedThreadPool(numberOfBlocks);
+
+        final ExecutorService persistenceExecutor = Executors.newFixedThreadPool(numberOfBlocks);
+        final BlockingExecutorService subscriberExecutor1 = new BlockingExecutorService(10, 1);
+        final BlockingExecutorService subscriberExecutor2 = new BlockingExecutorService(10, 1);
+        final BlockingExecutorService subscriberExecutor3 = new BlockingExecutorService(10, 1);
```
AlexKehayov, line 170: BlockingExecutorService is used for injection in subscribe stream observers, ensuring that onNext() calls can be awaited before assertions begin.

server/src/test/java/com/hedera/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
```diff
@@ -155,14 +159,31 @@ void setUp() throws IOException {
...
+        CountDownLatch publishStreamObserversLatch = new CountDownLatch(3);
+        doAnswer(onEventLatchCountdown(publishStreamObserversLatch))
+                .when(helidonPublishStreamObserver1)
+                .onNext(any());
+        doAnswer(onEventLatchCountdown(publishStreamObserversLatch))
+                .when(helidonPublishStreamObserver2)
+                .onNext(any());
+        doAnswer(onEventLatchCountdown(publishStreamObserversLatch))
+                .when(helidonPublishStreamObserver3)
+                .onNext(any());
```
AlexKehayov, line 183: CountDownLatch + Mockito.doAnswear() technique is used on publish stream observers, ensuring that onNext() calls can be awaited before assertions begin.

server/src/test/java/com/hedera/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
```diff
@@ -184,22 +205,25 @@ void testPublishBlockStreamRegistrationAndExecution() throws IOException {
         // Register 3 consumers - Opening a pipeline is not enough to register a consumer.
         // pipeline.onNext() must be invoked to register the consumer at the Helidon PBJ layer.
         pbjBlockStreamServiceProxy
-                .open(
+                .openWithSpecifiedExecutor(
```
AlexKehayov, line 208: openWithSpecifiedExecutor is used instead of open, so the custom executors can be injected.

server/src/test/java/com/hedera/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
```diff
@@ -213,49 +237,62 @@ void testPublishBlockStreamRegistrationAndExecution() throws IOException {
...
-        verify(helidonPublishStreamObserver3, timeout(testTimeout).times(1)).onNext(any());
-
         // Close the stream as Helidon does
         helidonPublishStreamObserver1.onComplete();
 
+        // Wait for subscribers and publishers to finish execution before assertions
+        subscriberExecutor1.waitTasksToComplete();
+        subscriberExecutor2.waitTasksToComplete();
+        subscriberExecutor3.waitTasksToComplete();
+        publishStreamObserversLatch.await();
```
AlexKehayov, line 247: Async tasks are awaited here.

server/src/test/java/com/hedera/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
```diff
@@ -213,49 +237,62 @@ void testPublishBlockStreamRegistrationAndExecution() throws IOException {
...
+        verify(subscribeStreamObserver2, times(1)).onNext(buildSubscribeStreamResponse(blockItems.get(9)));
+
+        verify(subscribeStreamObserver3, times(1)).onNext(buildSubscribeStreamResponse(blockItems.getFirst()));
+        verify(subscribeStreamObserver3, times(8)).onNext(buildSubscribeStreamResponse(blockItems.get(1)));
+        verify(subscribeStreamObserver3, times(1)).onNext(buildSubscribeStreamResponse(blockItems.get(9)));
+
+        verify(helidonPublishStreamObserver1, times(1)).onNext(any());
+        verify(helidonPublishStreamObserver2, times(1)).onNext(any());
+        verify(helidonPublishStreamObserver3, times(1)).onNext(any());
+
```
AlexKehayov, line 265: Timeouts are removed from assertions as they are no longer needed.

server/src/test/java/com/hedera/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
```diff
@@ -701,7 +743,8 @@ private static void verifySubscribeStreamResponse(
             int blockItemsToWait,
             int blockItemsToSkip,
             Pipeline<? super Bytes> pipeline,
-            List<BlockItemUnparsed> blockItems) {
+            List<BlockItemUnparsed> blockItems,
+            boolean verifyWithTimeout) {
```
AlexKehayov, line 747: Verification with timeout is now conditional to align with the new BlockingExecutorService approach.

server/src/test/java/com/hedera/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
```diff
@@ -739,8 +788,8 @@ private static Bytes buildEndOfStreamResponse() {
     private BlockVerificationSessionFactory getBlockVerificationSessionFactory() {
         final VerificationConfig config = blockNodeContext.configuration().getConfigData(VerificationConfig.class);
         final SignatureVerifierDummy signatureVerifier = mock(SignatureVerifierDummy.class);
-        when(signatureVerifier.verifySignature(any(), any())).thenReturn(true);
         final ExecutorService executorService = ForkJoinPool.commonPool();
+        lenient().when(signatureVerifier.verifySignature(any(), any())).thenReturn(true);
```
AlexKehayov, line 792: Added lenient() to prevent conditional stubbing exceptions.

server/src/test/java/com/hedera/block/server/mediator/LiveStreamMediatorImplTest.java
```diff
@@ -167,21 +171,20 @@ void testMediatorPersistenceWithoutSubscribers() throws IOException {
                 serviceStatus,
                 ackHandlerMock,
                 writerFactory,
-                executorMock,
+                executor,
```
AlexKehayov, line 174: The new BlockingExecutorService is used, so persistence can be awaited before the assertions.

server/src/test/java/com/hedera/block/server/mediator/LiveStreamMediatorImplTest.java
```diff
@@ -167,21 +171,20 @@ void testMediatorPersistenceWithoutSubscribers() throws IOException {
...
+        executor.waitTasksToComplete();
+
         // Verify the counter was incremented
         assertEquals(10, blockNodeContext.metricsService().get(LiveBlockItems).get());
-
-        // @todo(642) we need to employ the same technique here to inject a writer that will ensure
-        // the tasks are complete before we can verify the metrics for blocks persisted
-        // the test will pass without flaking if we do that
-        // assertEquals(1, blockNodeContext.metricsService().get(BlocksPersisted).get());
+        assertEquals(1, blockNodeContext.metricsService().get(BlocksPersisted).get());
```
AlexKehayov, line 187: Flakiness is now avoided by using executor.waitTasksToComplete().

server/src/main/java/com/hedera/block/server/pbj/PbjBlockStreamServiceProxy.java
```diff
@@ -135,7 +136,42 @@ public Pipeline<? super Bytes> open(
...
+                                subscribeStreamRequest, helidonConsumerObserver, Executors.newSingleThreadExecutor()))
+                        .mapResponse(reply -> createSubscribeStreamResponse(reply, options))
+                        .respondTo(replies)
+                        .build();
+            };
+        } catch (Exception e) {
+            replies.onError(e);
+            return Pipelines.noop();
+        }
+    }
```
AlexKehayov, line 149: Executors.newSingleThreadExecutor() is passed here to preserve the original logic outside the tests.

server/src/main/java/com/hedera/block/server/pbj/PbjBlockStreamServiceProxy.java
```diff
@@ -135,7 +136,42 @@ public Pipeline<? super Bytes> open(
...
+
+    @NonNull
+    public Pipeline<? super Bytes> openWithSpecifiedExecutor(
```
AlexKehayov, line 152: This method allows the use of different executors for test purposes and can be further modified to support additional testing utilities.

AlexKehayov (commented):
self-review

server/src/test/java/com/hedera/block/server/util/BlockingExecutorService.java
```diff
@@ -0,0 +1,35 @@
...
+        completedTasks++;
+        if (getQueue().isEmpty() && completedTasks >= expectedTasks) {
+            countDownLatch.countDown();
+        }
+    }
+
+    public void waitTasksToComplete() throws InterruptedException {
+        countDownLatch.await();
+    }
+}
```
jsync-swirlds, line 35: I'm not sure this is entirely correct.
1. This is limited to cases where we know the task count in advance, and it just counts down from there to zero.
2. This doesn't necessarily handle the queue correctly, since it only counts down when the queue is empty, so if we submit 100 tasks (out of 1000) in rapid succession, we won't count all of them and may never complete.


Suggestion:
In the method `waitTasksToComplete`, just use a busy-wait (with a moderate, perhaps 500 ns pause each cycle), and wait for `getQueue().isEmpty()`, `getActiveCount() == 0`, and `getCompletedTaskCount() >= getTaskCount()` all to be `true`.
e.g.
```java
    while(!(getQueue().isEmpty() && getActiveCount() == 0 && getCompletedTaskCount() >= getTaskCount())) {
        LockSupport.parkNanos(500); // This could be a longer pause, if desired.
        if(Thread.interrupted()) {
            break;
        }
    }
```

That eliminates the need for the countdown latch, the counting of tasks, and the afterExecute override.

jsync-swirlds (commented):
The work here is very much appreciated.
I do have a question on the executor; it might be a bit more complex than necessary, and I'm not sure the count down is entirely correct.

AlfredoG87:
Please rebase with latest `main` so we have an actual PR Code Coverage report and also for the other CI Checks to pass.

server/src/test/java/com/hedera/block/server/util/BlockingExecutorService.java
AlexKehayov: Thank you for the feedback. I will have the open() method overloaded in the next commit.

Regarding the BlockingExecutorService, unfortunately the busy-waiting approach does not give deterministic results (image appended). I'm not sure I understand the issues you mentioned. In the current implementation the countdown is called only once - when the queue is empty and at least the minimum amount of tasks expected are completed. The  afterExecute method assures that the check is made after a task is finished (not just submitted to a worker and possibly still in progress). The expectedTasks param also gives us opportunity to wait for tasks that are submitted to the executor with delay (all previously submitted tasks may already be completed and the conditions you mentioned will be met before additional tasks enter the queue - probably that is the reason for the flakiness with the proposed busy-waiting approach).

image

jsync-swirlds:
The merge modified pre-existing commits (possibly due to an out-of-date main).
It might be helpful to back up and then _rebase_ instead.

server/src/test/java/com/hedera/block/server/util/BlockingExecutorService.java
jsync-swirlds: I think I see, you're not counting down tasks, you're using the count down latch as a flag (it's either 1 or 0).
I wonder if we could do this more simply, but if it works, it's a good first pass and we can investigate other options later.

jsync-swirlds (approved):

AlfredoG87 (approved):
LG, however still needs a rebase from main and also conflicts resolutions 😓

jsync-swirlds (approved):

AlfredoG87 (approved):
LGTM! 👍

AlfredoG87:
Need rebase to latest `main` version and conflict resolution. once is done, I can re-approve.

georgi-l95 (approved):

ata-nas (approved):
LG

jsync-swirlds:
@AlfredoG87 The PR title check fails for some reason (not clear why).  I believe you can still merge this?

Nana-EC:
Looking into failing PR check blocker

Nana-EC:
@AlexKehayov kindly rebase, I just pushed a PR #820 that should help with the GHA that was stuck here

PR #784, opened 2025-03-05
chore: removal of the BlockNodeContext object

### LINKED ISSUE #629: chore: removal of the BlockNodeContext object
### Persona

As a Block Node Developer

### Request

I want to drop the usage of the `BlockNodeContext`

### Goal

So that I can remove redundant complexity of the context which is only used to do DI

### Technical Notes

Currently there exist a `BlockNodeContext` object that is wired with the `Dagger` DI. The context has currently only a reference to the `MetricsService` and the `Configuration` class and is only used to do DI of the metrics service and the configuration.
The configuration for any given type is in some places still acquired via the `Configuration` object, which contradicts other usages, so it is a deprecated usage that should be removed.
The metrics service has no real reason to be injected via the context object.
This refactor will improve a lot of redundant and wrongful usages we currently have and clean up a lot the current DI wiring.

We need to:
- drop the `BlockNodeContext` object
- all places that require the metrics service should simply get it via DI
- all places that require a specific type of configuration should get it via DI
- if there are any `Configuration` specific types that are needed to be injected, but are not provided via `Dagger`, need to be now provided via `Dagger`

AlexKehayov:
## Reviewer Notes
Removed BlockNodeContext usage.
Eliminated raw Configuration usage.
Added ServiceConfig in ConfigInjectionModule.
Refactored constructors and builders to replace BlockNodeContext and raw Configuration with MetricsService and specific configurations.
Updated tests to align with these changes.

## Related Issue(s)
Closes #629


jsync-swirlds:
Just a note; the "deterministic build" and related "baseline" checks will always fail for fork PRs (limitation in runners, as I understand it).

server/src/test/java/com/hedera/block/server/BlockNodeAppTest.java
```diff
@@ -64,7 +66,13 @@ class BlockNodeAppTest {
...
 
     @Mock
-    private BlockNodeContext blockNodeContext;
+    private MetricsService metricsService;
+
+    @Mock
+    private ProducerConfig producerConfig;
+
+    @Mock
+    private ConsumerConfig consumerConfig;
```
jsync-swirlds, line 75: Rather than mocking these, it's generally better to get the real configuration (which may be modified by test environment).
Configuration doesn't have any side effects that would require mocking.

jsync-swirlds (dismissed):
Looks good, just one small item.

server/src/test/java/com/hedera/block/server/BlockNodeAppTest.java
AlexKehayov: You are absolutely right, I will refactor that right away.

jsync-swirlds (approved):

PR #788, opened 2025-03-06
chore: Update issue templates

jsync-swirlds:
* Added type to both forms
* Moved most values to placeholders
* Changed 3 separate questions to one text area for feature/task
   * This improves readability versus the 3 separate headers github inserts
* Removed titles so that is entered more easily
* Fixed projects assignment

Small updates based on user feedback to make the template forms easier to use and less intrusive.


AlfredoG87:
I would be nice to have a Bug Template 😅  😁

ata-nas (approved):
LG, great addition for the bug template. Looks like it needs a spotless run though. Style check fails.

Nana-EC (approved):
LG. Thanks

AlfredoG87 (approved):

PR #795, opened 2025-03-06
feat: BN server Helm various improvements

### LINKED ISSUE #791: PersistentVolumeClaim Helm Templates
### Persona

As a BlockNode  Operator

### Request

I want to be able to persist saved data between restarts into a PVC

### Goal

So that persisted data is indeed persisted to FS or Claim provider.

### Technical Notes

PVC should be highly flexible and configurable.
- an Existing PVC can be provided via `Values.yaml` file.
- it can be Created using the `Values.yaml` configurable options.

this should allow the helm charts to create the PVC in case they are not provided.

### LINKED ISSUE #792: Rename Hedera Resources names to Hiero inside the Block-Node-Server Helm Chart Templates
### Persona

As a Hiero Mainteneer

### Request

I want to Change all instances of hedera to hiero

### Goal

So that repository migration is completed and keeps consistency

### Technical Notes

this refactor should not alter functionality

### LINKED ISSUE #793: Update Resources request to production values and add minikube override
### Persona

As a BlockNode Operator

### Request

I want to have defaults the production requisites

### Goal

So that is easier to deploy to production

### Technical Notes

It would be nice to have a `/override-values/mini.yaml` that provides a layer of overriden values optimized to run in minikube or another dev environment

AlfredoG87:
## Reviewer Notes
- Fixed notes command for correct `port-forward` tunnel into the `block-node-server` using default port of `8080`
- changed all instances were resources were named with the `hedera` prefix and updated them for `hiero` prefix.
- Added Archive and Live PVC
  - given flexibility to use existing external PVC instead of creating one.
- optimized `default` **Values.yaml** file for Production recommendations, and created a mini version that is intended to run on a minikube or similar environment for small tests.
- 

## Related Issue(s)
Fixes #793 
Fixes #792 
Fixes #791

### DIFF
--- a/charts/block-node-server/values-overrides/mini.yaml
+++ b/charts/block-node-server/values-overrides/mini.yaml
@@ -1,0 +1,21 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# This override values for the block-node-server chart to deploy a mini version of the block-node-server.
+# This is useful for testing and development purposes with limited resources.
+# ie: minikube
+
+resources:
+  requests:
+    cpu: "3"
+    memory: "8Gi"
+
+blockNode:
+  config:
+    JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties"
+    JAVA_OPTS: "-Xms5G -Xmx5G"
+    MEDIATOR_RING_BUFFER_SIZE: "2048"
+  persistence:
+    archive:
+      size: 2Gi
+    live:
+      size: 5Gi
--- a/charts/block-node-server/values-overrides/mini.yaml
+++ b/charts/block-node-server/values-overrides/mini.yaml
@@ -16,6 +16,6 @@ blockNode:
     MEDIATOR_RING_BUFFER_SIZE: "2048"
   persistence:
     archive:
-      size: 2Gi
+      size: 6Gi
     live:
-      size: 5Gi
+      size: 1Gi
--- a/charts/block-node-server/values.yaml
+++ b/charts/block-node-server/values.yaml
@@ -80,7 +80,29 @@ blockNode:
     # VERIFICATION_TYPE: "NO_OP"
     # MEDIATOR_TYPE: "NO_OP"
     MEDIATOR_RING_BUFFER_SIZE: "4096"
-
+  persistence:
+    archive:
+      # If false, the chart expects an externally provided PVC
+      create: true
+      # Name of the externally provided PVC
+      existingClaim: ""
+      # If create is true, the following values are used to create the PVC
+      # should match PERSISTENCE_STORAGE_ARCHIVE_ROOT_PATH, leave as is for default.
+      mountPath: "/opt/hashgraph/blocknode/data/archive"
+      size: 2Gi
+      # Optionally add a storage class name if needed
+      # storageClass: "your-storage-class"
+    live:
+      # If false, the chart expects an externally provided PVC
+      create: true
+      # Name of the externally provided PVC
+      existingClaim: ""
+      # If create is true, the following values are used to create the PVC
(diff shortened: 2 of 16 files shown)


charts/block-node-server/values.yaml
```diff
@@ -79,8 +79,30 @@ blockNode:
...
+      # storageClass: "your-storage-class"
+    live:
+      # If false, the chart expects an externally provided PVC
+      create: true
+      # Name of the externally provided PVC
+      existingClaim: ""
+      # If create is true, the following values are used to create the PVC
+      # should match PERSISTENCE_STORAGE_LIVE_ROOT_PATH, leave as is for default.
+      mountPath: "/opt/hashgraph/blocknode/data/live"
+      size: 500Gi
```
AlfredoG87, line 103: @ata-nas, on current GKE Cluster we have 1 TB of disk storage.

lets assume we can allocate 800 GB for the PVCs... what would be the ideal relationship between archive and live. I set it as `200` for archive and `500` for live, but want your thoughts here.

AlfredoG87 (commented):
Self Review.

charts/block-node-server/values.yaml
jsync-swirlds: Commented elsewhere as well, but archive should be almost all of the storage.
Live is only ever a small set of blocks before the archiver migrates the blocks to the archive volume.

In production, live will be on NVMe and Archive will be on a bunch of spinning disks.  The size difference will be about 100:1 archive to live.

AlfredoG87: what about we give 20GB to Live, and 800GB to Archive? would that make sense in current default settings?

charts/block-node-server/values-overrides/mini.yaml
```diff
@@ -0,0 +1,21 @@
...
+blockNode:
+  config:
+    JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties"
+    JAVA_OPTS: "-Xms5G -Xmx5G"
+    MEDIATOR_RING_BUFFER_SIZE: "2048"
+  persistence:
+    archive:
+      size: 2Gi
+    live:
+      size: 5Gi
```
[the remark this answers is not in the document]
AlfredoG87: Will give 1GB to Live and 6 to Archive then 👍

charts/block-node-server/values.yaml
jsync-swirlds: Sounds good.

AlfredoG87: ✅ Done!

jsync-swirlds (dismissed):

jsync-swirlds (dismissed):

charts/block-node-server/values.yaml
ata-nas: @AlfredoG87 what @jsync-swirlds says is correct. The consideration is mostly about what the `archiveGroupSize` will be. Based on the speed of the blocks coming in to be persisted and the block size average, we can assume that the live storage will have the `archiveGroupSize` x2 at the minimum, but not much more than that since the archiving will be coming around to take a group along as the threshold passes. For archived storage, consider that the blocks are zipped with no compression applied to the zip itself (for now), but when the blocks are written they are compressed (by default) using the java `zstd` library. (If memroy is correct, the compression level we use by default is 3).

AlfredoG87: Yes, we need to think if is better to have different PVCs for those needs too

jsync-swirlds: We can probably figure those things out when we get state management built.
That's a bit in the future still.

jsync-swirlds (approved):

PR #797, opened 2025-03-06
chore: fixed disabled tests

### LINKED ISSUE #751: Increase historic and live block streaming test coverage
### Persona

As a block node engineer

### Request

I want to increase test coverage on the historic and live block streaming

### Goal

So that we can catch regression defects, etc

### Technical Notes

* While working on #675 and #672 I refactored much of the streaming consumer classes. I was not able to replace all of the broken tests with tests adapted to the new approach.

mattp-swirldslabs:
## Reviewer Notes
* Realigned disabled unit test to the new consumer polling mechanism
* Fixed broken Producer metric

## Related Issue(s)
Fixes #751

### DIFF
--- a/server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorBuilder.java
+++ b/server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorBuilder.java
@@ -1,12 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.hedera.block.server.mediator;
 
+import com.hedera.block.server.consumer.StreamManager;
 import com.hedera.block.server.events.BlockNodeEventHandler;
 import com.hedera.block.server.events.ObjectEvent;
 import com.hedera.block.server.metrics.MetricsService;
 import com.hedera.block.server.service.ServiceStatus;
 import com.hedera.hapi.block.BlockItemUnparsed;
 import com.lmax.disruptor.BatchEventProcessor;
+import com.lmax.disruptor.EventPoller;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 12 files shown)


server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorBuilder.java
```diff
@@ -70,16 +75,25 @@ public static LiveStreamMediatorBuilder newBuilder(
...
-                                    BatchEventProcessor<ObjectEvent<List<BlockItemUnparsed>>>>
-                            subscribers) {
+            @NonNull final Map<BlockNodeEventHandler<ObjectEvent<List<BlockItemUnparsed>>>,
+                               BatchEventProcessor<ObjectEvent<List<BlockItemUnparsed>>>> subscribers) {
         this.subscribers = subscribers;
         return this;
     }
+    // spotless:on
+
+    // spotless:off
```
AlfredoG87, line 88: should we use this spotless hacks?

I recognize that spotless is not perfect and this might well be one case of this. 

```
    @NonNull
    public LiveStreamMediatorBuilder subscribers(
            @NonNull
                    final Map<
                                    BlockNodeEventHandler<ObjectEvent<List<BlockItemUnparsed>>>,
                                    BatchEventProcessor<ObjectEvent<List<BlockItemUnparsed>>>>
                            subscribers) {
        this.subscribers = subscribers;
        return this;
    }

```

given the spotless format is uglier but do we want to have this kind of comments all over the place and make a precedent on disabling and enabling spotless?

@jsync-swirlds @ata-nas Thoughts?

AlfredoG87 (approved):
LGTM. 

Just one nit/question about the spotless format workaround.

server/src/main/java/com/hedera/block/server/mediator/LiveStreamMediatorBuilder.java
jsync-swirlds: We do this in other places (relatively frequently in pbj).
Spotless is useful to have consistency, but on occasion the results are just unreadble.

Rather than go to exceptional lengths to accommodate spotless a _rare_ exception is fine.

ata-nas: I am pro for using the spotless off exceptionally when spotless would really render our code completely unreadable.

Nana-EC: Sounds like the team is good

PR #798, opened 2025-03-06
chore: rename java packages from com.hedera to org.hiero across the repo

### LINKED ISSUE #709: Rename Java packages from `com.hedera` to `org.hiero` across the repo.
### Persona

As a Hiero Maintainer

### Request

I want to update the java package reference to be `org.hiero.` 

### Goal

So that everything is transitioned to LFDT

### Technical Notes

# CRITCIAL REQUIREMENT
Coordinate all changes across the team so that all active PRs are merged, then this change merges to main.

georgi-l95:
## Reviewer Notes
- Renamed all packages `com.hedera` to `org.hiero` accross the repository.
## Related Issue(s)
Fixes #709


server/docker/logging.properties
```diff
@@ -26,24 +26,24 @@ io.helidon.common.level=INFO
...
-#com.hedera.block.server.config.logging.ConfigurationLoggingImpl.level=OFF
+#logging.config.org.hiero.block.server.ConfigurationLoggingImpl.level=OFF
 
 # Configure specific Block Node loggers
-#com.hedera.block.server.producer.ProducerBlockItemObserver.level=FINE
-#com.hedera.block.server.mediator.LiveStreamMediatorImpl.level=FINE
-#com.hedera.block.server.mediator.LiveStreamPoller.level=FINE
+#producer.org.hiero.block.server.ProducerBlockItemObserver.level=FINE
+#mediator.org.hiero.block.server.LiveStreamMediatorImpl.level=FINE
+#mediator.org.hiero.block.server.LiveStreamPoller.level=FINE
```
jsync-swirlds, line 34: Adding a prefix here will break these.
The logger name exactly matches either a class or package.

server/docker/logging.properties
```diff
@@ -26,24 +26,24 @@ io.helidon.common.level=INFO
...
 # Configure the configuration logging - OFF will suppress all configuration logging
-#com.hedera.block.server.config.logging.ConfigurationLoggingImpl.level=OFF
+#logging.config.org.hiero.block.server.ConfigurationLoggingImpl.level=OFF
 
 # Configure specific Block Node loggers
-#com.hedera.block.server.producer.ProducerBlockItemObserver.level=FINE
-#com.hedera.block.server.mediator.LiveStreamMediatorImpl.level=FINE
-#com.hedera.block.server.mediator.LiveStreamPoller.level=FINE
+#producer.org.hiero.block.server.ProducerBlockItemObserver.level=FINE
+#mediator.org.hiero.block.server.LiveStreamMediatorImpl.level=FINE
```
jsync-swirlds, line 33: This must be exactly `org.hiero.block.server.mediator.LiveStreamMediatorImpl` to match the correct logger name.
You can see where the value is "set" [here](https://github.com/hiero-ledger/hiero-block-node/pull/798/files#diff-75ce2bbdcc28981b21db7edb46207e25d0ceef9a92f01cd40ceb91db609e4b73R32).

server/src/main/java/module-info.java
```diff
@@ -1,52 +1,51 @@
...
+    requires org.hiero.block.stream;
     requires com.github.luben.zstd_jni;
+    requires com.github.spotbugs.annotations;
     requires com.lmax.disruptor;
     requires dagger;
     requires io.helidon.common;
     requires io.helidon.webserver;
     requires javax.inject;
-    requires static transitive com.github.spotbugs.annotations;
-    requires static transitive com.google.auto.service;
```
jsync-swirlds, line 49: Is this correct?  I don't think we want to make added changes other than the package renames in this PR...

simulator/src/main/java/module-info.java
```diff
@@ -1,35 +1,34 @@
...
     requires com.google.protobuf;
     requires dagger;
     requires io.grpc.stub;
     requires io.grpc;
     requires java.logging;
     requires javax.inject;
-    requires static transitive com.github.spotbugs.annotations;
-    requires static transitive com.google.auto.service;
-    requires static java.compiler; // javax.annotation.processing.Generated
+    requires static java.compiler;
```
jsync-swirlds, line 31: Looks like something moved even comments around here...

tools/build.gradle.kts
```diff
@@ -6,12 +6,14 @@ plugins {
...
+description = "Hiero Block Stream Tools"
 
-application { mainClass = "com.hedera.block.tools.BlockStreamTool" }
+application { mainClass = "org.hiero.block.tools.BlockStreamTool" }
 
 mainModuleInfo {
-    requires("com.hedera.block.stream") // use streams module to access protobuf generated classes
+    requiresStatic(
+        "org.hiero.block.stream"
+    ) // use streams module to access protobuf generated classes
```
jsync-swirlds, line 16: Another possibly unnecessary change...
In this case, I'm not sure we can make this change, as the hedera-protobufs hasn't changed anything to match hiero naming.

server/src/main/java/module-info.java
georgi-l95: Possibly, could be incorect. Will revert those and try to build again, but will clear caches before that. I made those because otherwise gradle was complaining. 🤔

tools/build.gradle.kts
georgi-l95: That's a good point. I was wondering about the streams module. Do we change it to hiero or not? Because protobuf hasn't changed, as you mentioned.

jsync-swirlds: We have to use the same packages for the hedera-protobufs until/unless those change to hiero.
Keeping the module the same as the source avoids confusion.

AlfredoG87: I think is a matter of time before the protobufs are moved to hiero. but for the mean time, yes we should only change what we can change, and for this PR let's try to keep it focused only on the package name, other things are less disruptive and can be tackled individually and incrementally.

georgi-l95:
Superseded by #822

PR #802, opened 2025-03-07
feat: Simple Persistence startup data

### LINKED ISSUE #796: Simple Persistence startup data
### Story Form

As a Block Node Developer/Operator/User
I want to know at startup what is the first and last available block (meaning persisted && verified)
So that I can take actions based on that information

### Technical Notes

- currently there is a need to know at system startup what is the first and last available block
- available block is one that is both persisted and verified
- the persistence at startup could now be utilized to determine that
- we will walk the file tree utilizing the trie structure to determine the first and last block that resides in the live storage
- we know that in the live storage exists only persisted and verified blocks
- under the live storage root we also have the links to the archived blocks, which are also persisted and verified so will be taken into account when trying to determine first/last available blocks
- for now we will utilize the server status to effectively publish that data at system startup

ata-nas:
## Reviewer Notes
- Changed the default start block generated from `1` to `0`
- Changed the assumption that Genesis block number was `1` on BN to `0`.
- Persistence will `traverse` the files and provide to `serviceStatus` the `lastBlock` persisted, so that BN can resume streaming from the last received block and not having to start always from genesis.
- Improved AckHandler init process due to now receiving the `lastAckedBlock` from persistence.
- Added UT with coverage for new functionality

The goal of this PR is to enable a simple and partial, but functional way for the BN to **Recover from restarts** so that persisted blocks are truly used and persisted across restarts.

## Related Issue(s)
Closes #796

### DIFF
--- a/server/src/main/java/com/hedera/block/server/persistence/storage/archive/LocalGroupZipArchiveTask.java
+++ b/server/src/main/java/com/hedera/block/server/persistence/storage/archive/LocalGroupZipArchiveTask.java
@@ -75,7 +75,7 @@ public final class LocalGroupZipArchiveTask implements Callable<Long> {
      * @param config valid, non-null {@link PersistenceStorageConfig} instance
      * @param pathResolver valid, non-null {@link BlockPathResolver} instance
      */
-    LocalGroupZipArchiveTask(
+    public LocalGroupZipArchiveTask(
             final long blockNumberThreshold,
             @NonNull final PersistenceStorageConfig config,
             @NonNull final BlockPathResolver pathResolver) {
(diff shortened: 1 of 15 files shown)


AlfredoG87 (approved):
LGTM 👍

server/src/main/java/com/hedera/block/server/persistence/storage/archive/LocalGroupZipArchiveTask.java
```diff
@@ -75,7 +75,7 @@ public final class LocalGroupZipArchiveTask implements Callable<Long> {
      * @param config valid, non-null {@link PersistenceStorageConfig} instance
      * @param pathResolver valid, non-null {@link BlockPathResolver} instance
      */
-    LocalGroupZipArchiveTask(
+    public LocalGroupZipArchiveTask(
```
jsync-swirlds, line 78: hmmm, we shouldn't be making this public (even for testing).
Something to look at in more detail later, though.

jsync-swirlds (approved):

server/src/main/java/com/hedera/block/server/persistence/storage/archive/LocalGroupZipArchiveTask.java
ata-nas: @jsync-swirlds agreed. However the logic is not so simple to reproduce locally in the unit test (at least not easily and fast). Testing the new logic is very important however, hence I made the small exception.

PR #805, opened 2025-03-08
chore: Set unique default server port

### LINKED ISSUE #800: Set new default server port
### Story Form

As a block node operator
I want to have a default server port that is not the default 8080
So that the port is unique and not easy to experience collisions and confusion with other applications 

### Technical Notes

The block node supports port 8080 as the main endpoint.
We should set it to something else (team suggested 40840) sooner rather than later

Nana-EC:
## Reviewer Notes

Set a new unique server port 40840 to reduce likelihood of collisions.
Updates cover charts, docker files, config READMEs, serverconfig, tests files and tests bash files

## Related Issue(s)
Fixes #800


charts/block-node-server/templates/NOTES.txt
```diff
@@ -1,8 +1,8 @@
 1. **Get the application URL** by running the following commands:
 
 ```bash
-kubectl port-forward --namespace {{ .Release.Namespace }} svc/{{.Release.Name}}-{{.Chart.Name}} 8080:{{ .Values.service.port }}
-echo "Visit http://127.0.0.1:8080 to use your application"
+kubectl port-forward --namespace {{ .Release.Namespace }} svc/{{.Release.Name}}-{{.Chart.Name}} 40840:{{ .Values.service.port }}
+echo "Visit http://127.0.0.1:40840 to use your application"
```
AlfredoG87, line 5:
```suggestion
kubectl port-forward --namespace {{ .Release.Namespace }} svc/{{.Release.Name}}-{{.Chart.Name}} {{ .Values.service.port }}:{{ .Values.service.port }}
echo "Visit http://127.0.0.1:{{ .Values.service.port }} to use your application"
```

AlfredoG87 (commented):
Looking good, maybe adding  files:

https://github.com/hiero-ledger/hiero-block-node/blob/c303fa0f19f2dc10406de9fd8d0d2c9562bf7e0d/simulator/src/test/java/com/hedera/block/simulator/config/ConfigInjectionModuleTest.java#L51

and 

https://github.com/hiero-ledger/hiero-block-node/blob/0be7b657458f1e7705a61a5796f333a593dc00a5/simulator/src/main/resources/app.properties#L7

jsync-swirlds (approved):

PR #821, opened 2025-03-11
chore: named injection for components with a common interface

### LINKED ISSUE #628: chore: named injection for components with a common interface
### Persona

As a Block Node Developer

### Request

I want to use `@Named` injection with `Dagger`

### Goal

So that I can have components that share an interface be injectable

### Technical Notes

Currently there exist components that are different in nature/logic which share an interface, for instance the persistence handler and the verification handler are both a `BlockNodeEventHandler` so they can work with the ring buffer.
This does not allow for injection of any `BlockNodeEventHandler` by default.
Different injection targets need to be somehow distinguished.
A way to do that is by using the `@Named` annotation which would allow `Dagger` to wire multiple implementations with the same interface.
These then could be injected as the interface they implement in the appropriate places using the `@Named` qualifier again.

We need to:
- identify all such components, example is the persistence handler and the verification handler which are both `BlockNodeEventHandler`
- migrate the current injection logic to leverage the `@Named` qualifier approach

AlexKehayov:
## Reviewer Notes
- Implemented named injection for StreamPersistenceHandlerImpl and StreamVerificationHandlerImpl.
- Scanned the project for similar cases but didn't find any. Other interfaces with multiple implementations rely on configuration-based injection managed by Dagger.

## Related Issue(s)
Closes #628

### DIFF
--- a/block-node/base/src/main/java/org/hiero/block/server/verification/StreamVerificationHandlerImpl.java
+++ b/block-node/base/src/main/java/org/hiero/block/server/verification/StreamVerificationHandlerImpl.java
@@ -7,7 +7,6 @@ import com.hedera.hapi.block.BlockItemUnparsed;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.List;
 import java.util.Objects;
-import javax.inject.Inject;
 import javax.inject.Singleton;
 import org.hiero.block.server.events.BlockNodeEventHandler;
 import org.hiero.block.server.events.ObjectEvent;
@@ -39,7 +38,6 @@ public class StreamVerificationHandlerImpl implements BlockNodeEventHandler<Obje
      * @param serviceStatus           the service status
      * @param blockVerificationService the block verification service
      */
-    @Inject
     public StreamVerificationHandlerImpl(
             @NonNull final SubscriptionHandler<List<BlockItemUnparsed>> subscriptionHandler,
             @NonNull final Notifier notifier,
--- a/block-node/server/src/main/java/org/hiero/block/server/pbj/PbjBlockStreamServiceProxy.java
+++ b/block-node/server/src/main/java/org/hiero/block/server/pbj/PbjBlockStreamServiceProxy.java
@@ -22,6 +22,7 @@ import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import javax.inject.Inject;
+import javax.inject.Named;
 import org.hiero.block.server.block.BlockInfo;
 import org.hiero.block.server.consumer.ClosedRangeHistoricStreamEventHandlerBuilder;
 import org.hiero.block.server.consumer.ConsumerConfig;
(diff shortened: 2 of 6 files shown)


jsync-swirlds:
We have a 100% rename of files and packages to replace `com.hedera` with `org.hiero` that merged today, and another to split the `server` into `server` and `base` modules (preparatory to extracting more modules later this week, and likely continuing thereafter).  This PR will need to rebase (or probably easier to just re-do) following the merge of PR #824.

block-node/server/src/main/java/org/hiero/block/server/pbj/PbjBlockStreamServiceProxy.java
```diff
@@ -75,8 +75,10 @@ public class PbjBlockStreamServiceProxy implements PbjBlockStreamService {
     public PbjBlockStreamServiceProxy(
             @NonNull final LiveStreamMediator streamMediator,
             @NonNull final ServiceStatus serviceStatus,
-            @NonNull final BlockNodeEventHandler<ObjectEvent<List<BlockItemUnparsed>>> streamPersistenceHandler,
-            @NonNull final StreamVerificationHandlerImpl streamVerificationHandler,
+            @NonNull @Named("StreamPersistenceHandlerImpl")
+                    final BlockNodeEventHandler<ObjectEvent<List<BlockItemUnparsed>>> streamPersistenceHandler,
+            @NonNull @Named("StreamVerificationHandlerImpl")
```
ata-nas, line 80: Personally, I would have these literals in a constants class, I suggest we create a new one, call it `InjectionConstants` or similar. Having to search for string literals to then copy paste where needed is quite cumbersome and any changes that need to be made would have to be made everywhere instead of just once if we use the constants.

ata-nas (commented):
@AlexKehayov LG. A minor suggestion, having the name of the `impl` class for the injection name may not be the best idea, but that is just my opinion. Reasoning is that changing the name of this implementation would imply we need to change the injection name, also, we can shorten them for easier use. On another note, having the injection names provided as static constants would be great as well.

block-node/base/src/main/java/org/hiero/block/server/verification/StreamVerificationHandlerImpl.java
```diff
@@ -39,7 +38,6 @@ public class StreamVerificationHandlerImpl implements BlockNodeEventHandler<Obje
      * @param serviceStatus           the service status
      * @param blockVerificationService the block verification service
      */
-    @Inject
```
ata-nas, line 42: Why the removal of the inject annotation here?

ata-nas (approved):
LG 🔥

jsync-swirlds:
A process note: When you create the PR, it helps to assign to yourself, set milestone, and add a label.  That will resolve the PR formatting check failures.

ata-nas:
@AlexKehayov your CI is failing due to spotless, could you please re-apply that and push?

jsync-swirlds (approved):

PR #824, opened 2025-03-11
chore: Initial Module Split for Block Node Server

jsync-swirlds:
### Pre-Migration Work
* Move ServerConfig to config
* Move Constants to service sub-package
* Move some constants from PbjBlockStreamServiceProxy to Constants

### Module Split
* Moved former "server" down under "block-node" folder
* Moved most code to a "base" module
* Adjusted the remaining "server" code to reference "base" module as a library
* Cleaned up most duplication
   * Test utils are duplicated until we can create a shared testing library/module
* Fixed up the build
* Verified tests all run
   * Three "integration" tests were disabled due to having complex cross-module interactions
      * We will need to rework or replace these.


block-node/base/src/main/java/module-info.java
```diff
@@ -22,29 +21,25 @@
...
+    requires transitive com.hedera.pbj.runtime;
+    requires transitive com.swirlds.common;
+    requires transitive com.swirlds.config.api;
+    requires transitive com.swirlds.config.extensions;
+    requires transitive com.swirlds.metrics.api;
+    requires transitive org.hiero.block.common;
+    requires transitive com.lmax.disruptor;
+    requires transitive dagger;
+    requires transitive io.helidon.webserver;
+    requires transitive javax.inject;
```
AlfredoG87, line 41: is this transitive because they are needed on the `server` module but they are coming from the `base` module?

settings.gradle.kts
```diff
@@ -5,7 +5,24 @@ rootProject.name = "hiero-block-node"
...
+        module("server") { artifact = "block-node-server" }
+        module("base") { artifact = "block-node-base" }
+        // module("persistence") { artifact = "service-persistence" }
+        // module("persistence-api") { artifact = "service-persistence-api" }
+        // module("verification") { artifact = "service-verification" }
+        // module("verification-api") { artifact = "service-verification-api" }
+        // module("messaging") { artifact = "facility-messaging" }
+        // module("configuration") { artifact = "facility-configuration" }
+        // module("registrar") { artifact = "facility-registrar" }
+        // module("facilities-api") { artifact = "facilities-api" }
```
AlfredoG87, line 26: are these comments on purpose?

settings.gradle.kts
```diff
@@ -5,7 +5,24 @@ rootProject.name = "hiero-block-node"
 
 javaModules {
     directory(".") {
-        group = "org.hiero.block"
+        group = "com.hedera.block"
```
AlfredoG87, line 8: should be `org.hiero.block` ?

settings.gradle.kts
```diff
@@ -5,7 +5,24 @@ rootProject.name = "hiero-block-node"
...
         module("tools") // no 'module-info' yet
+        module("suites")
+        module("simulator")
+        module("common")
+        module("stream")
+    }
+    directory("block-node") {
+        group = "com.hedera.block"
```
AlfredoG87, line 16: same here

AlfredoG87 (commented):
Looking really good, I leaved some questions, one of them prevents me from approving just yet.

Well done!

block-node/base/src/main/java/module-info.java
jsync-swirlds: Spotless refused to allow the PR to continue without these.
They don't actually need to be transitive, it's just spotless being picky.

settings.gradle.kts
jsync-swirlds: Yes, deliberate.
I plan to add (some of) these in future PRs.

settings.gradle.kts
jsync-swirlds: Good Catch, will fix.

rbarker-dev (dismissed):
This looks good to me and I was able to build and test locally.

settings.gradle.kts
AlfredoG87: not a blocker, but is there a todo ticket for these that we could associate it to? otherwise looks like leftover comments.

jsync-swirlds: It's part of the same Issue.  This PR is very much partial, to unblock the team.
I'll remove the unused items in the next PR.

AlfredoG87 (approved):
💯

jsync-swirlds:
So changing the group broke the gradle module checks; it won't pass without removing modules that are actually required, so I'm reverting the group change and will investigate that tomorrow.

AlfredoG87:
> BTW, if succeeds locally, and only fails in CI, so it's definitely something weird.
to me it did broke locally too.

AlfredoG87:
> So changing the group broke the gradle module checks; it won't pass without removing modules that are actually required, so I'm reverting the group change and will investigate that tomorrow.

Yes, I agree, we still need to do a lot of more `hedera` to `hiero` follow-up and clean-up.

AlfredoG87 (approved):

block-node/base/src/test/java/org/hiero/block/server/util/TestUtils.java
ata-nas, line 1: This class seems like a redundant copy of the test utils that was made, which currently exists in the `server` module. Since nothing is used from here, this class should be deleted. Logic here is about file perms and dir deletion which really should not exist anymore since we dropped support for block-as-dir.

block-node/server/src/test/java/org/hiero/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
```diff
@@ -199,6 +200,7 @@ void setUp() throws IOException {
         mediatorConfig = config.getConfigData(MediatorConfig.class);
     }
 
+    @Disabled
```
ata-nas, line 203: Any particular reason for disabling?

block-node/server/src/test/java/org/hiero/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
```diff
@@ -502,6 +504,7 @@ void testFullWithSubscribersAddedDynamically() throws IOException, InterruptedEx
         verifySubscribeStreamResponse(numberOfBlocks, 89, numberOfBlocks, subscribeStreamObserver6, blockItems, true);
     }
 
+    @Disabled
```
ata-nas, line 507: Any particular reason for disabling?

ata-nas (approved):
Great job on initial split @jsync-swirlds! I leave a couple of remarks that came up which could be for a future cleanup.

On a completely different note, would it make sense in the future to have common logic for testing, like some utilities that were now duplicated, to be extracted into the common/test?

block-node/base/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+plugins { id("org.hiero.gradle.module.library") }
+
+description = "Hedera Block Node Base"
```
georgi-l95, line 4: Should this be Hiero Block Node Base ?

georgi-l95 (approved):
LG

block-node/server/src/main/java/module-info.java
```diff
@@ -0,0 +1,21 @@
...
+    requires com.swirlds.metrics.api;
+    requires org.hiero.block.base;
+    requires org.hiero.block.common;
+    requires dagger;
+    requires io.helidon.common;
+    requires io.helidon.webserver;
+    requires javax.inject;
+    requires static transitive com.github.spotbugs.annotations;
+    requires static transitive com.google.auto.service;
+    requires static java.compiler; // javax.annotation.processing.Generated
```
ata-nas, line 20: Line 20 here may be redundant? Just a note for a future cleanup. I have seen the java.compiler appear from time to time during dev, removing it was always the right choice I believe. In any case, I do not think this gets added by us explicitly.

rbarker-dev (approved):

block-node/server/src/test/java/org/hiero/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
jsync-swirlds: This test crosses the modules in ways that make it incompatible with a split, so I disabled for now and we'll re-write in follow-up.

block-node/server/src/test/java/org/hiero/block/server/pbj/PbjBlockStreamServiceIntegrationTest.java
jsync-swirlds: This test crosses the modules in ways that make it incompatible with a split, so I disabled for now and we'll re-write in follow-up.

block-node/base/src/test/java/org/hiero/block/server/util/TestUtils.java
jsync-swirlds: I copied everything in test utils because it's used _everywhere_ and we have not yet setup a shareable test module.  That's something we'll need to do in follow-up.

block-node/server/src/test/java/org/hiero/block/server/util/TestUtils.java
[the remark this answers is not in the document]
jsync-swirlds: I copied everything in test utils because it's used everywhere and we have not yet setup a shareable test module. That's something we'll need to do in follow-up.

PR #716, opened 2025-02-25
fix: changed the logging format and added tests

mattp-swirldslabs:
## Reviewer Notes
* Worked with @jsync-swirlds  to arrive at the logging format: `%TF %<TT.%<TL%<Tz %4$-7s [%2$s] %5$s%6$s%n`

## Related Issue(s)
Fixed #703

PR #827, opened 2025-03-12
feat: Centralize Record Configuration Logging for Simulator

### LINKED ISSUE #611: Centralize Record Configuration Logging for Simulator
As an Administrator
I want to see all properties of the Simulator runtime configuration logged at boot time
So that I can confirm the configuration on a node or by inspecting the logs on a remote dashboard for troubleshooting

## Tech Notes
* I was thinking we could follow the pattern established in this [BN ticket](https://github.com/hashgraph/hedera-block-node/issues/471).
* We have [this precedent](https://github.com/hashgraph/hedera-block-node/issues/474) to look at for sensitive properties

### LINKED ISSUE #893: Introduce Custom Logging and Log Configuration to the Simulator
### Story Form

I want to have customized logging for the simulator
So that I will use the customized logging

### Technical Notes

- taking inspiration from the server, we should add:
  - `logging.properties` file with custom options that will be available inside a docker image to be used
  - custom log format (see one in server)
- we should also have some log formatting tests (see tests in server)

ata-nas:
## Reviewer Notes

- this pr is essentially a copy of #501 and #716 but for the simulator
- general future improvements would be an endpoint that could be queried for the config values instead of the log we currently have
- added formatting for the logger of the simulator to match the one we have for the server
- awareness tests could be extended in the future as well (if this implementation would be supported)

## Related Issue(s)

Closes #611
Closes #893

### DIFF
--- a/charts/blockstream-simulator/values-overrides/consumer.yaml
+++ b/charts/blockstream-simulator/values-overrides/consumer.yaml
@@ -3,6 +3,5 @@ replicas: 2
 
 simulator:
   config:
-    JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties"
     GRPC_SERVER_ADDRESS: "my-bn-block-node-helm-chart"
     BLOCK_STREAM_SIMULATOR_MODE: "CONSUMER"
--- a/charts/blockstream-simulator/values-overrides/consumer.yaml
+++ b/charts/blockstream-simulator/values-overrides/consumer.yaml
@@ -3,5 +3,6 @@ replicas: 2
 
 simulator:
   config:
+    JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties"
     GRPC_SERVER_ADDRESS: "my-bn-block-node-helm-chart"
     BLOCK_STREAM_SIMULATOR_MODE: "CONSUMER"
--- a/charts/blockstream-simulator/values.yaml
+++ b/charts/blockstream-simulator/values.yaml
@@ -20,7 +20,6 @@ replicas: 1
 
 simulator:
   config:
-    JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties"
     # either CONSUMER or PUBLISHER
     BLOCK_STREAM_SIMULATOR_MODE: "PUBLISHER_CLIENT"
     GRPC_SERVER_ADDRESS: "my-bn-block-node-helm-chart"
--- a/charts/blockstream-simulator/values.yaml
+++ b/charts/blockstream-simulator/values.yaml
@@ -20,6 +20,7 @@ replicas: 1
 
 simulator:
   config:
+    JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties"
     # either CONSUMER or PUBLISHER
     BLOCK_STREAM_SIMULATOR_MODE: "PUBLISHER_CLIENT"
     GRPC_SERVER_ADDRESS: "my-bn-block-node-helm-chart"
--- a/simulator/docker/Dockerfile
+++ b/simulator/docker/Dockerfile
@@ -11,7 +11,9 @@ WORKDIR /app
 
 # Copy the distribution and resources
 COPY simulator-*.tar ./simulator.tar
-COPY logging.properties .
+
+RUN mkdir -p /app/logs/config
+COPY logging.properties /app/logs/config/logging.properties
 
 # Create resources directory and copy block data
 RUN mkdir -p src/main/resources
@@ -24,7 +26,7 @@ RUN tar -xf simulator.tar && \
     tar -xzf block-0.0.3.tar.gz && \
     rm block-0.0.3.tar.gz && \
     cd /app && \
-    chown -R $UNAME:$UNAME /app
+    chown -R $UID:$GID /app
 
 # Switch to non-root user
 USER $UNAME
--- a/simulator/docker/docker-compose.yml
+++ b/simulator/docker/docker-compose.yml
@@ -17,6 +17,8 @@ services:
       interval: 3s
       timeout: 10s
       retries: 5
+    volumes:
+      - ./logging.properties:/app/logs/config/logging.properties
 
   simulator-consumer:
     container_name: simulator-consumer
@@ -34,6 +36,8 @@ services:
     depends_on:
       simulator-publisher:
         condition: service_healthy
+    volumes:
+      - ./logging.properties:/app/logs/config/logging.properties
 
 networks:
   block-node_default:
--- a/simulator/docker/docker-compose.yml
+++ b/simulator/docker/docker-compose.yml
@@ -17,8 +17,6 @@ services:
       interval: 3s
       timeout: 10s
       retries: 5
-    volumes:
-      - ./logging.properties:/app/logs/config/logging.properties
 
   simulator-consumer:
     container_name: simulator-consumer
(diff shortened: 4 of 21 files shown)


charts/blockstream-simulator/values-overrides/consumer.yaml
```diff
@@ -3,5 +3,6 @@ replicas: 2
 
 simulator:
   config:
+    JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties"
```
ata-nas, line 6: @AlfredoG87 is this addition to chart correct?

charts/blockstream-simulator/values.yaml
```diff
@@ -20,6 +20,7 @@ replicas: 1
 
 simulator:
   config:
+    JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/app/logs/config/logging.properties"
```
ata-nas, line 23: @AlfredoG87 is this addition to chart correct?

simulator/docker/Dockerfile
```diff
@@ -11,7 +11,9 @@ WORKDIR /app
 
 # Copy the distribution and resources
 COPY simulator-*.tar ./simulator.tar
-COPY logging.properties .
+
+RUN mkdir -p /app/logs/config
+COPY logging.properties /app/logs/config/logging.properties
```
ata-nas, line 16: @AlfredoG87 is this addition to oci file correct?

simulator/docker/Dockerfile
```diff
@@ -24,7 +26,7 @@ RUN tar -xf simulator.tar && \
     tar -xzf block-0.0.3.tar.gz && \
     rm block-0.0.3.tar.gz && \
     cd /app && \
-    chown -R $UNAME:$UNAME /app
+    chown -R $UID:$GID /app
```
ata-nas, line 29: Changed the `chown` which used UNAME previously

simulator/docker/docker-compose.yml
```diff
@@ -17,6 +17,8 @@ services:
       interval: 3s
       timeout: 10s
       retries: 5
+    volumes:
+      - ./logging.properties:/app/logs/config/logging.properties
```
ata-nas, line 21: @AlfredoG87 is this addition to compose correct?

simulator/docker/docker-compose.yml
```diff
@@ -34,6 +36,8 @@ services:
     depends_on:
       simulator-publisher:
         condition: service_healthy
+    volumes:
+      - ./logging.properties:/app/logs/config/logging.properties
```
ata-nas, line 40: @AlfredoG87 is this addition to compose correct?

simulator/docker/update-env.sh
```diff
@@ -17,6 +17,9 @@ PUBLISHER_PROMETHEUS_ENDPOINT_PORT_NUMBER=9998
 # For consumer service
 CONSUMER_BLOCK_STREAM_SIMULATOR_MODE=CONSUMER
 CONSUMER_PROMETHEUS_ENDPOINT_PORT_NUMBER=9997
+
+# Logging props
+JAVA_TOOL_OPTIONS=-Djava.util.logging.config.file=/app/logs/config/logging.properties
```
ata-nas, line 22: Adding the java tool opts as env var.

ata-nas (commented):
Just a couple places to take heed.

charts/blockstream-simulator/values-overrides/consumer.yaml
AlfredoG87: No need to repeat the key assignment if is the same value, by definition an override file should only contain the key values that change over the "values.yaml" default values.

charts/blockstream-simulator/values.yaml
AlfredoG87: This is partially correct, this will indeed tell the JVM to use that config file properties from start-up, so all logs get the desired format.

However, this alone would use the defaults always and without an easy (if any) way to change the log level or format.

In the Block-Node-Chart we achive customization via a custom config-map that writes the file `logging.properties` and then we mount this file at the desired path on the container `/app/logs/config/`

see this templates:
https://github.com/hiero-ledger/hiero-block-node/blob/main/charts/block-node-server/templates/configmap-logging.yaml

https://github.com/hiero-ledger/hiero-block-node/blob/main/charts/block-node-server/templates/deployment.yaml#L80-L83


However you can also create a follow-up issue to add `Custom Logging Properties files to simulator Helm Chart `

simulator/docker/Dockerfile
AlfredoG87: Yes, seems correct in order to include the default logging.properties file.

simulator/docker/Dockerfile
AlfredoG87: I guess this accomplish the same result in a more explicit way. 👍

simulator/docker/docker-compose.yml
AlfredoG87: No, it would make sense if you add a logging.properties file to the `/docker` folder.

This is esentially telling Docker to override the default logging.properties with the one you have on the docker folder.

AlfredoG87: I would say to leave it there but add the file to the docker folder, so is easier to change the Logger Level and properties when running on docker.

simulator/docker/docker-compose.yml
AlfredoG87: same as above.

simulator/docker/update-env.sh
AlfredoG87: Yes this is needed for java to load the properties formatter and level on the logging.properties file. 👍

AlfredoG87 (commented):
Looking good, just a few comments.

Nana-EC:
@ata-nas , Fredy answered the q's above. Also since these are all simulator items you should be good to complete this with little conflict from Jaspers changes.

charts/blockstream-simulator/values.yaml
ata-nas: Thanks for the feedback. You are right about addressing this in a follow up. I will do that.

charts/blockstream-simulator/values-overrides/consumer.yaml
ata-nas: I will drop the changes in the charts to be addressed in another issue.

ata-nas: Update: I have created #892 to address as follow up the charts changes.

charts/blockstream-simulator/values.yaml
ata-nas: Update: I have created #892 to address as follow up the charts changes.

simulator/docker/docker-compose.yml
ata-nas: I have removed this for now as it seems unnecessary.

simulator/docker/docker-compose.yml
ata-nas: I have removed this for now as it seems unnecessary.

ata-nas:
> @ata-nas , Fredy answered the q's above. Also since these are all simulator items you should be good to complete this with little conflict from Jaspers changes.

@Nana-EC I was unsure about a couple of things, on another view, I have split the effort into a couple of issues, some to be addressed in a follow up. The current state should allow us to merge without stalling the PR any longer. Mainly, what is achieved here is:
- introduction of the custom logging format and options as seen in the server
- introduction of the config logging logic at simulator startup, again as seen in the server

Nana-EC (approved):

jsync-swirlds (approved):

AlfredoG87 (approved):
LGTM

PR #841, opened 2025-03-12
chore: follow-up for hiero package move

jsync-swirlds:
- Moved the base package for all modules to `org.hiero` in the main build settings
   - Fixed all the references as needed (mostly just the "stream" module name changed).


ata-nas (approved):
🔥

PR #843, opened 2025-03-13
chore: remove block data from the repo

### LINKED ISSUE #429: Remove block data from the repo
As a block node developer
I want to remove all block data accross all subprojects
To be able to maintain easier and not clutter the project

### Additional information
Currently we have the same block data across the repo in several places. Ideally we'd want to remove it all together.

AlexKehayov:
## Reviewer Notes
 - Removed all block data across repo (except for some test data).
 - Modified BlockGeneratorConfig handling for GenerationMode.DIR.
 - Modified tests according to the removal of block data and BlockGeneratorConfig changes.
 - Removed untar from simulator Dockerfile and build.gradle.kts.
## Related Issue(s)
 Closes #429


simulator/src/test/java/org/hiero/block/simulator/config/data/BlockStreamConfigTest.java
```diff
@@ -33,7 +33,7 @@ private BlockStreamConfig.Builder getBlockStreamConfigBuilder() {
     }
 
     private BlockGeneratorConfig.Builder getBlockGeneratorConfigBuilder() {
-        String folderRootPath = "build/resources/main//block-0.0.3/";
+        String folderRootPath = "build/resources/test//block-0.0.3-blk/";
```
ata-nas, line 36: Why the double forward slashs?

ata-nas (commented):
LG, have you tried running the simulator using the file (dir) strategy again in order to see that it still works after your changes? We must not lose that, even though we have now the generator strategy for producing (publishing) data via the simulator.

simulator/src/test/java/org/hiero/block/simulator/TestUtils.java
```diff
@@ -48,4 +49,8 @@ public static int findFreePort() throws IOException {
             return socket.getLocalPort();
         }
     }
+
+    public static String getAbsoluteFolder(String relativePath) {
```
Nana-EC, line 53: Q: Should this have a null guard with a test also?
```suggestion
    public static String getAbsoluteFolder(@NonNull String relativePath) {
```

simulator/src/test/java/org/hiero/block/simulator/generator/BlockAsFileBlockStreamManagerTest.java
```diff
@@ -40,14 +36,14 @@ void getNextBlock() throws IOException, BlockSimulatorParsingException {
 
     @Test
     void getNextBlockItem() throws IOException, BlockSimulatorParsingException {
-        for (int i = 0; i < 35000; i++) {
+        for (int i = 0; i < 72; i++) {
```
Nana-EC, line 39: Q: Why the change to 72? Is the data set smaller and the expectation now only 72 block items?

Nana-EC (commented):
Thanks for the contribution.
1 clarifying question and 1 suggestion

simulator/src/test/java/org/hiero/block/simulator/TestUtils.java
AlexKehayov: You are right. If relativePath is null, calling Paths.get(relativePath) will throw a NullPointerException. Adding @NonNull is suitable for this case.

simulator/src/test/java/org/hiero/block/simulator/generator/BlockAsFileBlockStreamManagerTest.java
AlexKehayov: Yes, I removed the large tar file that was filling "build/resources/main/block-0.0.3/" and switched to "build/resources/test/block-0.0.3-blk/", which contains only 8 blocks.

Nana-EC: Okay, usually a minor comment or a well named variable in the test function helps to clarify for any devs reading

AlexKehayov: You are right, I will add a comment in the test.

Nana-EC:
@AlexKehayov feel free to rebase and update if you're able to now that the refactor to `main` has completed..
A lot has changed so it's possible the issue is no longer applicable or a larger or smaller effort than before.
Take your time and let us know what you find.

AlexKehayov:
@Nana-EC, the conflicts are fixed :)

simulator/src/test/java/org/hiero/block/simulator/TestUtils.java
```diff
@@ -48,4 +49,8 @@ public static int findFreePort() throws IOException {
             return socket.getLocalPort();
         }
     }
+
+    public static String getAbsoluteFolder(@NonNull String relativePath) {
+        return Paths.get(relativePath).toAbsolutePath().toString();
+    }
```
ata-nas, line 55: This method should not exist really, why would it? It simply calls logic that we can call directly. An utility method IMO should bring some value, like make an assert (precondition), a small arbitrary calculation or some steps that would make our life easier.

On another note, a minor nit: `Paths#get` should not be used, but use `Path#of` instead.

ata-nas (commented):
LG, let's run CI to be sure everything is in order 🙌

georgi-l95 (approved):

ata-nas (approved):

PR #848, opened 2025-03-13
chore: Add "facilities" wording to nano-service document.

jsync-swirlds:
## Reviewer Notes

This replaces the clunky "required services" with facilities, which much more clearly expresses the concept.


AlfredoG87 (approved):
LGTM

Nana-EC (changes requested):
Nice update.
Although I see the Status service was made optional and no longer a required service.
This is only reflected in the image and not in the md file.
Was this an omission or change to be considered here?

jsync-swirlds:
It's no longer present in the MD file as a named service (or at all).
It will be very important (and should be present in any block node that offers client APIs), but it turns out there are reasonable deployments without the status service, so it's a service and not a facility.

Nana-EC (approved):
LG,
I'm reminded  that some BNs won't expose the Status API as it won't be needed

PR #849, opened 2025-03-13
feat: Configured Observability stack to use PVC on Helm Charts

### LINKED ISSUE #831: Block-Node-Server Helm Deployment Template Improvement For Custom PVC Sub-Path data folder
### Story Form

As a BlockNode Operator
I want to specify the `subPath` inside the PVC for `live-data` and `archive-data`
So that if I don't want to use the defaults `live-data` and `archive-data` I can use any folder name that I wish

### Technical Notes

https://github.com/hiero-ledger/hiero-block-node/blob/main/charts/block-node-server/templates/deployment.yaml#L44-L89

it should come from https://github.com/hiero-ledger/hiero-block-node/blob/main/charts/block-node-server/values.yaml#L84-L105

### LINKED ISSUE #847: Observability Stack PVCs
### Story Form

As a BlockNode Operator
I want to Keep my metrics, logs and grafana configurations (dashboards, logins, alerts, etc) Persistent across pod reciclyng
So that I can have Observability data persisted across restarts of the whole deployment

### Technical Notes

For that we need to create PVCs on the helm charts for the Observability stack for the following services:
- Grafana
- Prometheus
- Loki

AlfredoG87:
## Reviewer Notes
- Configuring all observability stack to use Persistent PVC instead of ephemeral disks.
- Set `subPath` for `archive-data` and `live-data` configurable via chart values.
- Added override for mini deployments

## Related Issue(s)
Fixes #831 
Fixes #847

### DIFF
--- a/charts/block-node-server/templates/deployment.yaml
+++ b/charts/block-node-server/templates/deployment.yaml
@@ -48,12 +48,12 @@ spec:
             - sh
             - -c
             - |
-              mkdir -p /live-pvc/live-data && \
-              chown 2000:2000 /live-pvc/live-data && \
-              chmod 700 /live-pvc/live-data && \
-              mkdir -p /archive-pvc/archive-data && \
-              chown 2000:2000 /archive-pvc/archive-data && \
-              chmod 700 /archive-pvc/archive-data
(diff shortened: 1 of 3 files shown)


Nana-EC (approved):
LG

jsync-swirlds (approved):

PR #855, opened 2025-03-14
chore: Resolve and Enable E2E tests

### LINKED ISSUE #828: Improve simulator startup sequence in suites
### Story Form

As a Block Node Engineer
I want to be able to easily run simulator in different configurations
So that I can achieve flexability when creating tests

### Technical Notes

In order to achieve flexability and separtion in each tests scenario, we need to add base method/framework for easier simulator startup. It should start a new instance and should start it with passed configuration.

### LINKED ISSUE #845: Resolve and enable all E2E tests
### Story Form

As a Block Node Engineer
I want to have all available E2E tests in `suites` running
So that I have as much confidence and code coverage as possible

### Technical Notes

With some of the recent changes, we had to change disable some of the available E2E tests for persistance. We now need to resolve their errors and enable them once again. 
Test: https://github.com/hiero-ledger/hiero-block-node/blob/main/suites/src/main/java/org/hiero/block/suites/persistence/positive/PositiveDataPersistenceTests.java#L51
and 
https://github.com/hiero-ledger/hiero-block-node/blob/main/suites/src/main/java/org/hiero/block/suites/grpc/positive/PositiveEndpointBehaviourTests.java#L56

georgi-l95:
## Reviewer Notes
This PR:
- Enables all E2E tests that were previously disabled due to changes in the implementations.
- Modify and add additional startup method for the simulator, to allow flexible configuration changes.

## Related Issue(s)
Fixes #845 
Fixes #828

### DIFF
--- a/suites/src/main/java/module-info.java
+++ b/suites/src/main/java/module-info.java
@@ -3,11 +3,11 @@ module org.hiero.block.node.suites {
     requires com.swirlds.config.api;
     requires com.swirlds.config.extensions;
     requires org.hiero.block.simulator;
-    requires com.github.spotbugs.annotations;
     requires io.github.cdimascio.dotenv.java;
     requires org.junit.jupiter.api;
     requires org.junit.platform.suite.api;
     requires org.testcontainers;
+    requires static com.github.spotbugs.annotations;
     requires static dagger;
 
     exports org.hiero.block.suites to
--- a/suites/src/main/java/module-info.java
+++ b/suites/src/main/java/module-info.java
@@ -3,6 +3,7 @@ module org.hiero.block.node.suites {
     requires com.swirlds.config.api;
     requires com.swirlds.config.extensions;
     requires org.hiero.block.simulator;
+    requires com.github.spotbugs.annotations;
     requires io.github.cdimascio.dotenv.java;
     requires org.junit.jupiter.api;
     requires org.junit.platform.suite.api;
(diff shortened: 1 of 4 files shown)


suites/src/main/java/module-info.java
```diff
@@ -3,6 +3,7 @@
     requires com.swirlds.config.api;
     requires com.swirlds.config.extensions;
     requires org.hiero.block.simulator;
+    requires com.github.spotbugs.annotations;
```
jsync-swirlds, line 6: This should be `requires static`.
c.f. [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/base/src/main/java/module-info.java#L43) and [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/simulator/src/main/java/module-info.java#L30).
Note, those locations include `transitive` which _should not_ be needed here, but a run of `./gradlew checkAllModuleInfo` after adding without `transitive` will show you if it is necessary.

suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -142,30 +145,52 @@ protected Future<?> startSimulatorInThread(BlockStreamSimulatorApp blockStreamSi
...
+    }
+
+    /**
+     * Creates a new instance of the block stream simulator with default configuration.
      *
      * @return a new instance of the block stream simulator
      * @throws IOException if an I/O error occurs
      */
     protected BlockStreamSimulatorApp createBlockSimulator() throws IOException {
+        final Map<String, String> emptyMap = Map.of();
```
ata-nas, line 168: Nit: you could utilize the `Collections.emptyMap();` convenience.

ata-nas (commented):
LG, just a small optional nit.

suites/src/main/java/module-info.java
ata-nas: ✅ Done.

jsync-swirlds (approved):

Nana-EC (approved):

AlfredoG87 (approved):
LGTM

PR #856, opened 2025-03-14
chore: BlockAsLocalFile Persistence Test Plan [Phase 2]

### LINKED ISSUE #844: Files Recent Persistence E2E Test Plan
E2E test plan for the `Files Recent` Persistence.

ata-nas:
## Reviewer Notes

- test plan with test cases for `BlockAsLocalFile Persistence (BLFP)`

## Related Issue(s)

Closes #844


block-node/test-specifications/local-persistence/local-persistence-tp.md
jsync-swirlds, line 1: It would be good to start from the [TCK template](https://github.com/hiero-ledger/hiero-sdk-tck/blob/main/docs/test-specifications/TestSpecificationsTemplate.md) and remove the JSON-specific elements but keep the useful parts.

Examples:
- In the table, we cannot implement these tests consitently, the descriptions are not precise enough, and there is no input or output provided.
- There is no traceability to requirements (granted, that's not in the template either). Each test case should refer to a clear description of the requirement it verifies.
- The names here should not be "method" or "class" names.  They should be human-readable.  This isn't a code document, it's a plan that should turn into tests in code or run manually.

General notes:
- Do not use terse abbreviations. Things like `BLFP` should just be written out.
I like the inclusion of key considerations.
- We should have a definitions section (or perhaps a glossary document used across test plans?) for key terms.
- Should we have a "requirements" document, or just a section within each test plan?  One or the other is needed, but it's not clear which is better.

ata-nas: @jsync-swirlds I have made some changes. Please take a look and let me know if further improvements could be made. 🙏

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,69 @@
...
+  the location of a given `Block`. Path resolution utilizes 3 roots: `unverified`,
+  `live` and `archive`.
+- **Trie Data Structure for Path Resolution**: `BlockAsLocalFile Persistence` utilizes a trie data
+  structure to resolve paths for `Block`s. This is done to ensure that the
+  path resolution is efficient and fast.
+- **`unverfied` root path**: Initially all blocks are stored under the
+  `unverified`. After verification on a given `Block` passes, the persisted
+  `Block` is moved to the `live` root, essentially being _published_.
+- **`live` root path**: This is the root path where all verified `Block`s are
+  stored. It usually will not hold on to `Block`s for a long time, as they are
```
jsync-swirlds, line 25:
```suggestion
  stored. It usually will not hold on to `Block`s for a limited time, as they are
```

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,69 @@
...
+  usually archived in groups, leaving behind only a link to the archive. Utilizing
+  a link it means that discoverability of a `Block` is still possible under the
```
jsync-swirlds:
```suggestion
  a link means that discoverability of a `Block` is still possible under the
```

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,69 @@
...
+| E2ETC_BLFP_0001 | `testBlockPersistedUnverifiedLocation`               | `BlockAsLocalFile Persistence` will initially persist a `Block` under `unverified` root. We should be sure that the `Block` is indeed persisted in the proper location.                                                                                                                      | `Block` should be initially written to the correct location under the `unverfied` root.                                                                                                                                                                 |
```
jsync-swirlds: I'm still a bit concerned that we have no input or output and aren't tracing these to _requirements_. How do I _test_ this, and how do I know what changes when requirements change?

Examples:
>>input: Send block 18279 to the persistence service
>>output: a file, `0000000000018279.blk` is stored at `<unverified root>/`
>>requirement: Unverified blocks must be stored in the unverified root without a trie structure.
>
>>input: publish a verification sucess for block 18279 
>>output: a file, `0000000000018279.blk` is no longer stored at `<unverified root>/` and is not stored at `<live root>/0/0/0/0/0/0/0/0/0/0/0/1/8/2/7/0000000000018279.blk`
>>requirement: Verified blocks must be stored in the live root with a trie structure.
>

---
Minor linguistic tip, if you are using "should" in a scenario description, inputs, or outputs, then it's probably a "discussion" and not a "plan".

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
Nana-EC: If you limited it would be
```suggestion
  stored. `live` root path will only hold on to `Block`s for a limited time, as they are
```

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: ✅ Done.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: ✅ Done. Thanks.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: ✅ @jsync-swirlds thank you for the explicit example. I have made some changes, now with better clarification as to what I should aim for based on your example. Please take a look, should be a lot better.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Overview
+
+`BlockAsLocalFile Persistence` is a key component of the Block Node.
+Mainly, it is responsible for storing and retrieving `Blocks` from a local
+storage. It has several moving parts, like path resolution logic, writer tasks,
+readers etc. which when combined, provide a robust and efficient way to store
+and retrieve`Blocks`. This E2E test plan aims to highlight the key scenarios
+that need to be tested for end results in order to ensure the correctness and
+proper working of the `BlockAsLocalFile Persistence`.
```
Nana-EC, line 11: you can link the design doc here for further clarity

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,85 @@
...
...The persisted `Block`'s content is the same as the content that has been received as input.                                                                                                                                                                                                                                                                                                                                                                                                                                                                  | Send `BlockItem`s for `Block` with `BlockNumber`=12345L to the persistence service.                                                                                                                                                                                                              | Content of persisted `Block` is the same as the one received via input.                                                                                                                                                                                                                                                                                  |         N         |
```
Nana-EC, line 70: Q: Is there a consideration for the order of block items when confirming contents? Or does it make sense to check elsewhere?

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,85 @@
...
...Given that `Block` with `BlockNumber` 12345L is persisted as a file under the `unverified` root, verify that it can be overwritten, file content must be verified changed after an overwrite.                                                                                                                                                                                                                                                                                                                                                                | Ensure that `Block` with `BlockNumber`=12345L is persisted as file at `/unverified/0000000000000012345.blk` and is not published to `live` nor archived in `archive`, then send `BlockItem`s for `Block` with `BlockNumber`=12345L to the persistence service.                                   | `Block` persisted as file at `/unverified/0000000000000012345.blk` is overwritten with contents of the newly sent `Block` to the persistence service.                                                                                                                                                                                                    |         N         |
```
Nana-EC, line 71: > file content must be verified changed after an overwrite

To confirm the test will check that the override contents are different than before?
If they are the same is an error expected or is the file not overridden?

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,85 @@
...
...`Block` file and any data that may have resulted during a failed writing task must be cleaned up if persistence or verification fails.                                                                                                                                                                                                                                                                                                                                                                                                                       | Send `BlockItem`s for `Block` with `BlockNumber`=12345L to the persistence service.                                                                                                                                                                                                              | No file/data present under the `unverified` root that are result of a failed writing task.                                                                                                                                                                                                                                                               |         N         |
```
Nana-EC, line 72: Do we log in such a case?
Should log checks be part of the coverage here?

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,85 @@
...
...Any `Block` that is persisted under the `unverified` root must be moved to `live` once verification passes (essentially publishing it as available) and the content of the moved `Block` must be the same as pre-move.                                                                                                                                                                                                                                                                                                                                       | Ensure that `Block` with `BlockNumber`=12345L is persisted as file at `/unverified/0000000000000012345.blk` and is not published to `live` nor archived in `archive`, then verification must publish a successful verification result for `Block` with `BlockNumber`=12345L.                     | Regular file: `0000000000000012345.blk` with location: `/unverified/0000000000000012345.blk` is no longer present, but is now moved as a regular file at `/live/0/0/0/0/0/0/0/0/0/0/0/0/0/0/1/2/3/4/0000000000000012345.blk` and content of the moved file are the same as pre-move.                                                                     |         N         |
```
Nana-EC, line 75:
```suggestion
| E2ETC_BLFP_0007 | `Verify Content of Block Persisted Under Unverified Root After Move to Live Root`  | `BlockAsLocalFile Persistence` will move a `Block` from `unverified` to `live` root after verification passes. After the move happens, the content of the now live file must be the same compared to pre-move.                                                                                                                                                                                                                                                                 | Any `Block` that is persisted under the `unverified` root must be moved to `live` once verification passes (essentially publishing it as available) and the content of the moved `Block` must be the same as pre-move.                                                                                                                                                                                                                                                                                                                                       | Ensure that `Block` with `BlockNumber`=12345L is persisted as file at `/unverified/0000000000000012345.blk` and is not published to `live` nor archived in `archive`, then verification must publish a successful verification result for `Block` with `BlockNumber`=12345L.                     | Regular file: `0000000000000012345.blk` with location: `/unverified/0000000000000012345.blk` is no longer present, but is now moved as a regular file at `/live/0/0/0/0/0/0/0/0/0/0/0/0/0/0/1/2/3/4/0000000000000012345.blk` and content of the moved file are the same as pre-move.                                                                     |         N         |
```

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,85 @@
...
...Any `Block` files persisted under the `unverified` root must be deleted during system startup. The `unverified` root must be completely empty by the time the system is up and running, ready to operate.                                                                                                                                                                                                                                                                                                                                                    | Ensure that an arbitrary amount of `Block` files are present in the `unverified` root, then proceed to startup the system.                                                                                                                                                                       | `unverified` root is completely empty by the time the system is started and ready to operate.                                                                                                                                                                                                                                                            |         N         |
```
Nana-EC, line 76: Why would it not be possible to verify a block file under `/unverified` after startup?
Shouldn't the block proof process cover this?

For recovery reasons did we consider rather checking if a file in `/unverified` is not in `/live` before cleaning up in this case?

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,85 @@
...
...Any `Block` that is made available (published) as a file under the `live` root, is immutable and cannot be overwritten as it is deemed final. Any input with `BlockItem`s for any `Block` that has been published to `live` must result in a `DUPLICATE_BLOCK`.                                                                                                                                                                                                                                                                                              | Ensure that `Block` with `BlockNumber`=12345L is persisted as file at `/live/0/0/0/0/0/0/0/0/0/0/0/0/0/0/1/2/3/4/0000000000000012345.blk` and is not present in `unverified` nor archived in `archive`, then send `BlockItem`s for `Block` with `BlockNumber`=12345L to the persistence service. | Publisher must expect a `PERSISTENCE_FAILURE` response.                                                                                                                                                                                                                                                                                                  |         N         |
```
Nana-EC, line 78: Q: should we go as far as to make sure any related BlockWriter class cannot overwrite a file in `/live`?

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
Nana-EC, line 75: Should we add a test to confirm that 10 files written in succession are all available under the `/live` root dir e.g. 12340 to 12349

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,85 @@
...
...Any `Block` that is archived is discoverable in the sense that it is deemed available and can be read/queried via the `Block Node`'s APIs.                                                                                                                                                                                                                                                                                                                                                                                                                   | Ensure that `Block` with `BlockNumber`=12345L is archived and is not available in `live` nor `unverified`, then send `long` blockNumber=12345L as parameter to the block reader.                                                                                                                 | Block reader should return the queried `Block`.                                                                                                                                                                                                                                                                                                          |         N         |
```
Nana-EC, line 84: Add a test case to confirm the archived zip files contents for each block matches what's it was in `unverified` and then `/live`

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,85 @@
...
...Requirement                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  | Input                                                                                                                                                                                                                                                                                            | Output                                                                                                                                                                                                                                                                                                                                                   | Implemented (Y/N) |
```
Nana-EC, line 67: I would suggest a column for noting the grouping factor of functionality. Example I see `ingestion`, `writing`/`publication`, `reading`, `archiving` as potential values, there's probably better names
Let's think and agree on one before adding it. Doesn't have to be in this PR if we're unsure but under the umbrella of persistence there are a few features that it would be nice to know which combination of features each row is tackling

Nana-EC (changes requested):
Nice suggestions.
I added a few notes

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: Even if contents are the same, the file will still be truncated and overwritten. Always. As it is not possible to assert the state of the currently existing file (the one that will be now overwritten)

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: Good point. That one also needs an update with latest changes (mainly migrating to an async implementation and the fact that we dropped support for `block-as-dir`). I will make an issue for the update of the design doc to be addressed separately. Also, there is a simple chore to update all the `md` file internal links as most of them got broken due to module/renaming/hiero migration changes (just FYI).

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: Which test case are you referring to? If you mean `E2ETC_BLFP_0004`, there are debug logs on all persistence results, but doing assertions based on logs is not a good idea IMO.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: We currently have only one implementation of the writer, the block-as-file writer. And no-op, but that is going away pretty soon with migration to nano-service. no-op is here to simply be able to turn off the writer, in the new nano service approach, simply no persistence will be loaded.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: Which test case are you referring to here?

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: That is a good point. @georgi-l95 @AlfredoG87 @jsync-swirlds opinions?

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: P.S. also consider that this test plan is regarding the `block-as-local-file` persistence. It should theoretically cover and be concerned only with that implementation alone. If we ever have another one, maybe a new test plan is in place. In any case, I would make a wild guess and say that a different implementation of the persistence will probably use a different file structure (to justify it's existence), making these roots incompatible with it, but even if that is not the case, I would still make a wild guess and say that a different implementation will still not use the same root as this implementation.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
jsync-swirlds: It might be possible to calculate a file hash before the overwrite, then again after and assert that the two are not the same... Any MessageDigest algorithm would be sufficient (perhaps [MD5](https://docs.oracle.com/en/java/javase/21/docs/specs/security/standard-names.html#messagedigest-algorithms), to minimize testing load?)

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
jsync-swirlds: Consensus has some pretty big headaches because certain logs are, effectively, part of the operational API and any change to those log entries breaks node management and performance testing tooling.

We should learn from that and make sure that logging is purely informational.

If something is truly important and _must_ be reported, the "container-native" recommendation for doing that would be an `/statusz` endpoint managed by the Server Health facility (along with `/healthz` and `/readyz`) which provides a well-defined JSON response with detailed system status and point-in-time metrics.  We could also offer an `/alertz` endpoint which publishes anything that should be alertable (ideally with much more detail than a single-line log string can offer).

That would be a completely separate and not-inconsiderable effort to implement, however. It would probably make operations staff (possibly including ourselves) very grateful, though.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
jsync-swirlds: I agree in concept, but I also think it's best done separately from this PR.
This affects all of these test plans, and probably should be reflected in a template document which incorporates all of the things brought up for this document.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
jsync-swirlds: I'm not sure that's even possible with how file system permissions work.
It may be possible to use some tricks with linux security (usually intended to protect system files), but that's complex and prone to unexpected failures.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
Nana-EC: I see.
Not necessarily block contents proof.
More so if the input to the server is a batch of list of block items in order, then it would make sense to confirm that when grouped and written as a file that the order was maintained and the file is correct.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
Nana-EC: Yeah in hindsight I'm not a fan of logs in API either.
It's manageable when every case has a test case to capture behaviour but it quickly falls apart.

We are indeed missing the API to expose some details so maybe the full form of this and a few other tests should include an API call

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
Nana-EC: I'm suggesting a new tests case

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
```diff
@@ -0,0 +1,82 @@
...
+> tests, simply we need a reference point so an example could be shown):_
+> - `unverified` root: `/unverified`
+> - `live` root: `/live`
+> - `archive` root: `/archive`
+>
+> _**NOTE**: the trie structure for the `live` and `archive` root is resolving
+> all the digits of a `long` (max. 19 digits). We have a single digit per node
+> or directory up to a max. depth of 18, and then we have the full `BlockNumber`
+> as the file name of any persisted block with all leading zeros for the `live`
+> root and for the `archive` root we have a max. depth of
```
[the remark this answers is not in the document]
ata-nas: ✅ Thanks, done.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: ✅ Typo fixed. Thanks!

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: ✅ Done.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: ✅ Done, test case added.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: @jsync-swirlds @Nana-EC I have created #889 where we will create a template for the test cases and we can address all these additions.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
ata-nas: @jsync-swirlds @Nana-EC where do you believe it would be best to move the `test-specification` directory? Under repo root maybe?

jsync-swirlds: Once [this PR](https://github.com/hiero-ledger/hiero-block-node/pull/886) merges, these will belong in a reasonable structure under `/docs/test-plans` (at least that is the way I read that PR).

ata-nas: Sounds good.

block-node/test-specifications/local-persistence/block-as-local-file-persistence-tp.md
AlfredoG87: I like the suggestion 👍

ata-nas:
@jsync-swirlds @Nana-EC abandoning this. This PR is severely outdated both for the test plan and for the conversations we have had here. Almost none of what we see here is relevant as we have fully refactored the application, we now have persistence plugins which work differently than before and sometimes produce different e2e results (e.g. before we had an unverified path, but we no longer have, tests related to it are no longer relevant and much more).

Issue #662, opened 2025-02-18
Fix notifiable.notifyUnrecoverableError() so that consumers receive broadcast emergency message


mattp-swirldslabs:
As a spec-compliant block node implementation
I want to notify all subscribers in the event of a catastrophic server event
So that consumers receive a spec-compliant message notifying them that the server is going down

With this ticket, https://github.com/hashgraph/hedera-block-node/issues/652, we established a "data channel" and removed the ability for notifier.notifyUnrecoverableError() to broadcast to all subscribed consumers of the live stream. We need to build an equivalent messaging channel again.

HarshSawarkar:
Hi Team, I would be looking into this issue.

HarshSawarkar:
Hi, I have raised a PR for the issue. Could the team please review it?

jsync-swirlds:
Made obsolete by the code reorganization.

PR #858, opened 2025-03-14
chore: added the message error channel to broadcast emergency message

HarshSawarkar:
## Reviewer Notes

- Added a status field to `blockitemunparsed` in `unparsed.proto` for error tracking during streaming.
- Updated `LiveStreamMediatorImpl`.`notifyUnrecoverableError`() to broadcast unrecoverable errors.
- Modified `ConsumerStreamResponseObserver `to detect errors and forward them downstream.

## Related Issue(s)
#662

### DIFF
--- a/stream/src/main/proto/com/hedera/hapi/block/unparsed.proto
+++ b/stream/src/main/proto/com/hedera/hapi/block/unparsed.proto
@@ -56,5 +56,6 @@ message BlockItemUnparsed {
     bytes filtered_item_hash = 8;
     bytes block_proof = 9;
     bytes record_file = 10;
+    com.hedera.hapi.block.SubscribeStreamResponseCode status = 11;
   }
 }
--- a/block-node/base/src/main/java/org/hiero/block/server/consumer/ConsumerStreamResponseObserver.java
+++ b/block-node/base/src/main/java/org/hiero/block/server/consumer/ConsumerStreamResponseObserver.java
@@ -66,6 +66,11 @@ public class ConsumerStreamResponseObserver {
      */
     public void send(@NonNull final List<BlockItemUnparsed> blockItems) throws ParseException {
 
+        if (blockItems.stream().anyMatch(BlockItemUnparsed::hasStatus)) {
+            send(SubscribeStreamResponseCode.READ_STREAM_NOT_AVAILABLE);
+            return;
+        }
+
         // Only start sending BlockItems after we've reached
         // the beginning of a block.
         final BlockItemUnparsed firstBlockItem = blockItems.getFirst();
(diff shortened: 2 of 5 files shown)


jsync-swirlds:
We need to be careful adding to "block item unparsed". That message is purely a means to process Block Stream defined block items _without fully parsing_ the data.  If we add fields, it changes the serialization, which breaks the ability to verify the block stream (because we've now changed the data from what the network signed).

When we discuss messaging, we usually intend reference to the "Messaging Facility" as descried [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/docs/design/Nano-Service-Approach.md).

I know it's difficult to be patient, and we very much appreciate the eagerness to dive in.  We are working to schedule regular open community conference calls to provide both open communication and a forum for people to ask questions and guide participation in a way that avoids confusion or misdirection.

I will make a point to see that you are invited to the series as soon as we have it setup.

jsync-swirlds (changes requested):
This issue needs to hold a bit longer.
1) Because we cannot add fields to "BlockItemUnparsed".
2) We haven't built the Messaging Facility yet, and the issue depends on that facility.

stream/src/main/proto/com/hedera/hapi/block/unparsed.proto
```diff
@@ -56,5 +56,6 @@ message BlockItemUnparsed {
     bytes filtered_item_hash = 8;
     bytes block_proof = 9;
     bytes record_file = 10;
+    com.hedera.hapi.block.SubscribeStreamResponseCode status = 11;
```
jsync-swirlds, line 59: FYI, this is a protocol buffer definition, it _cannot_ reference Java objects; only other protocol buffer messages.

HarshSawarkar:
> We need to be careful adding to "block item unparsed". That message is purely a means to process Block Stream defined block items _without fully parsing_ the data. If we add fields, it changes the serialization, which breaks the ability to verify the block stream (because we've now changed the data from what the network signed).
> 
> When we discuss messaging, we usually intend reference to the "Messaging Facility" as descried [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/docs/design/Nano-Service-Approach.md).
> 
> I know it's difficult to be patient, and we very much appreciate the eagerness to dive in. We are working to schedule regular open community conference calls to provide both open communication and a forum for people to ask questions and guide participation in a way that avoids confusion or misdirection.
> 
> I will make a point to see that you are invited to the series as soon as we have it setup.

Thanks, @jsync-swirlds , for your efforts and support. I would be delighted to attend the calls to clarify my questions and learn from you all. Looking forward to these sessions!

stream/src/main/proto/com/hedera/hapi/block/unparsed.proto
HarshSawarkar: I thought it referred to the enum `SubscribeStreamResponseCode` from `block_service.proto`. Thanks for the clarification.

jsync-swirlds: That would have a different structure, but even then, block _stream_ objects cannot refer to anything in the block _service_ API.
It is a little confusing, but the two are (almost) completely separate.

Block Service defines the API for the Block Node.
Block Stream defines the structure of the stream of data produced (and signed) by the consensus network.

HarshSawarkar: Oh, okay, got it. Thanks for sharing the knowledge.

Issue #763, opened 2025-03-04
Create a Messaging Facility


jsync-swirlds:
### Persona

As a Block Node Service Engineer

### Request

I want to connect a new service to block node messaging

### Goal

So that my service can participate in the block node processing

### Technical Notes

The messaging facility for the block node is central to any service interactions and delivers the block stream data from publishers and other sources.
This feature will create that messaging facility and define the API(s) whereby newly loaded services will register to publish and receive messages.

PR #859, opened 2025-03-14
feat: Add messaging module

jasperpotts:
implements #763

### DIFF
--- a/block-node/messaging/build.gradle.kts
+++ b/block-node/messaging/build.gradle.kts
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+plugins { id("org.hiero.gradle.module.library") }
+
+description = "Hiero Block Node Messaging"
+
+// Remove the following line to enable all 'javac' lint checks that we have turned on by default
+// and then fix the reported issues.
+tasks.withType<JavaCompile>().configureEach { options.compilerArgs.add("-Xlint:-exports") }
+
+mainModuleInfo {
+    annotationProcessor("dagger.compiler")
+    annotationProcessor("com.google.auto.service.processor")
+    runtimeOnly("com.swirlds.config.impl")
+    runtimeOnly("org.apache.logging.log4j.slf4j2.impl")
+    runtimeOnly("io.helidon.logging.jul")
+    runtimeOnly("com.hedera.pbj.grpc.helidon.config")
+}
+
+testModuleInfo {
+    annotationProcessor("dagger.compiler")
+    requires("org.junit.jupiter.api")
+    requires("org.junit.jupiter.params")
+    requires("org.mockito")
+    requires("org.mockito.junit.jupiter")
+    requires("org.assertj.core")
+}
--- a/block-node/messaging/build.gradle.kts
+++ b/block-node/messaging/build.gradle.kts
@@ -17,10 +17,6 @@ mainModuleInfo {
 }
 
 testModuleInfo {
-    annotationProcessor("dagger.compiler")
+    requires("com.hedera.pbj.runtime")
     requires("org.junit.jupiter.api")
-    requires("org.junit.jupiter.params")
-    requires("org.mockito")
-    requires("org.mockito.junit.jupiter")
-    requires("org.assertj.core")
 }
--- a/block-node/messaging/build.gradle.kts
+++ b/block-node/messaging/build.gradle.kts
@@ -19,4 +19,5 @@ mainModuleInfo {
 testModuleInfo {
     requires("com.hedera.pbj.runtime")
     requires("org.junit.jupiter.api")
+    requires("org.junit.jupiter.params")
 }
--- a/block-node/messaging/src/main/java/module-info.java
+++ b/block-node/messaging/src/main/java/module-info.java
@@ -2,18 +2,6 @@
 module org.hiero.block.messaging {
     exports org.hiero.block.server.messaging;
 
-    requires transitive com.hedera.pbj.runtime;
-    requires transitive com.swirlds.common;
-    requires transitive com.swirlds.config.api;
-    requires transitive com.swirlds.config.extensions;
-    requires transitive com.swirlds.metrics.api;
-    requires transitive org.hiero.block.common;
-    requires transitive org.hiero.block.stream;
     requires transitive com.lmax.disruptor;
-    requires transitive dagger;
-    requires transitive io.helidon.webserver;
-    requires transitive javax.inject;
-    requires static transitive com.github.spotbugs.annotations;
-    requires static transitive com.google.auto.service;
-    requires static java.compiler; // javax.annotation.processing.Generated
+    requires org.hiero.block.stream;
 }
--- a/block-node/messaging/src/main/java/module-info.java
+++ b/block-node/messaging/src/main/java/module-info.java
@@ -8,4 +8,7 @@ module org.hiero.block.messaging {
     requires transitive org.hiero.block.stream;
     requires org.hiero.block.base;
     requires com.lmax.disruptor;
+
+    provides org.hiero.block.server.messaging.MessagingService with
+            org.hiero.block.server.messaging.impl.MessagingServiceImpl;
 }
--- a/block-node/messaging/src/main/java/module-info.java
+++ b/block-node/messaging/src/main/java/module-info.java
@@ -1,7 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
 module org.hiero.block.messaging {
+    uses com.swirlds.config.api.spi.ConfigurationBuilderFactory;
+
     exports org.hiero.block.server.messaging;
 
-    requires transitive com.lmax.disruptor;
-    requires org.hiero.block.stream;
+    requires com.lmax.disruptor;
+    requires transitive com.swirlds.config.api;
+    requires transitive org.hiero.block.stream;
+    requires org.hiero.block.base;
 }
--- a/block-node/messaging/src/main/java/module-info.java
+++ b/block-node/messaging/src/main/java/module-info.java
@@ -4,8 +4,8 @@ module org.hiero.block.messaging {
 
     exports org.hiero.block.server.messaging;
 
-    requires com.lmax.disruptor;
     requires transitive com.swirlds.config.api;
     requires transitive org.hiero.block.stream;
     requires org.hiero.block.base;
+    requires com.lmax.disruptor;
 }
--- a/block-node/messaging/src/main/java/module-info.java
+++ b/block-node/messaging/src/main/java/module-info.java
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: Apache-2.0
-
 module org.hiero.block.messaging {
     exports org.hiero.block.server.messaging;
 
--- a/block-node/messaging/src/main/java/module-info.java
+++ b/block-node/messaging/src/main/java/module-info.java
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+
+module org.hiero.block.messaging {
+    exports org.hiero.block.server.messaging;
+
+    requires transitive com.hedera.pbj.runtime;
+    requires transitive com.swirlds.common;
+    requires transitive com.swirlds.config.api;
+    requires transitive com.swirlds.config.extensions;
+    requires transitive com.swirlds.metrics.api;
+    requires transitive org.hiero.block.common;
+    requires transitive org.hiero.block.stream;
+    requires transitive com.lmax.disruptor;
+    requires transitive dagger;
+    requires transitive io.helidon.webserver;
+    requires transitive javax.inject;
+    requires static transitive com.github.spotbugs.annotations;
+    requires static transitive com.google.auto.service;
+    requires static java.compiler; // javax.annotation.processing.Generated
+}
--- a/block-node/messaging/src/main/java/org/hiero/block/server/messaging/BlockNotification.java
+++ b/block-node/messaging/src/main/java/org/hiero/block/server/messaging/BlockNotification.java
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.server.messaging;
+
+/**
+ * Simple record block notifications.
+ *
+ * @param blockNumber the block number this notification is for
+ * @param type the type of notification
+ */
+public record BlockNotification(long blockNumber, Type type) {
+    /**
+     * Enum representing the type of block notification.
+     */
+    public enum Type {
+        BLOCK_VERIFIED,
+        BLOCK_FAILED_VERIFICATION,
+        BLOCK_PERSISTED
+    }
+}
--- a/block-node/messaging/src/main/java/org/hiero/block/server/messaging/MessagingService.java
+++ b/block-node/messaging/src/main/java/org/hiero/block/server/messaging/MessagingService.java
@@ -8,6 +8,7 @@ import org.hiero.block.server.messaging.impl.MessagingServiceImpl;
 /**
  * Service that handles the messaging between the different components of the server. It is used to send block items
  * and block notifications to the different components. It is not meant to become a general purpose messaging service.
+ * Implementations of this service are expected to be thread safe and to handle back pressure.
  */
 public interface MessagingService {
 
@@ -32,38 +33,40 @@ public interface MessagingService {
 
     /**
      * Use this method to register a block item handler. The handler will be called every time new block items arrive.
-     * The calls will be on its own virtual thread, every handler registered has its own virtual thread. It can consume
-     * block items at its own pace, if it is too slow then it will apply back pressure to block item producer. It is too
-     * slow by taking too long in the handleBlockItemsReceived method. If handleBlockItemsReceived is non-blocking, then
-     * it will not apply back pressure.
-     * <br>
-     * It has to be called before starting the service.
+     * The calls will be on its own thread, every handler registered has its own thread. It can consume block items at
+     * its own pace, if it is too slow then it will apply back pressure to block item producer. It is too slow by taking
+     * too long in the handleBlockItemsReceived method. If handleBlockItemsReceived is non-blocking, then it will not
+     * apply back pressure.
      *
-     * @param handler the block item handler to register
+     * @param handler             the block item handler to register
+     * @param cpuIntensiveHandler hint to the service that this handler is CPU intensive vs IO intensive
+     * @param handlerName         the name of the handler, used for thread name and logging
      * @throws IllegalStateException if the service is already started
      */
-    void registerBlockItemHandler(BlockItemHandler handler) throws IllegalStateException;
+    void registerBlockItemHandler(BlockItemHandler handler, boolean cpuIntensiveHandler, String handlerName);
 
     /**
      * Use this method to dynamically register a block item handler. The handler will be called every time new block
-     * items arrive. It will be called on its own virtual thread, every handler registered has its own virtual thread.
-     * It can consume block items at its own pace, if it is too slow then the
+     * items arrive. It will be called on its own thread, every handler registered has its own thread. It can consume
+     * block items at its own pace, if it is too slow then the
      * {@link NoBackPressureBlockItemHandler#onTooFarBehindError} method will be called and the handler will be
      * unregistered.
-     * <br>
-     * This method has to be called after the service is started.
      *
-     * @param handler the block item handler to unregister
+     * @param handler             the block item handler to unregister
+     * @param cpuIntensiveHandler hint to the service that this handler is CPU intensive vs IO intensive
+     * @param handlerName         the name of the handler, used for thread name and logging
      */
-    void registerDynamicNoBackpressureBlockItemHandler(NoBackPressureBlockItemHandler handler);
+    void registerNoBackpressureBlockItemHandler(
+            NoBackPressureBlockItemHandler handler, boolean cpuIntensiveHandler, String handlerName);
 
     /**
-     * Use this method to unregister a dynamic block item handler. The handler will no longer be called when new block
-     * items arrive.
+     * Use this method to unregister any block item handler. The handler will no longer be called when new block
+     * items arrive. You only need to unregister handlers if they need to be unregistered before the service is
+     * shutdown. Shutting down the service will unregister all handlers.
      *
      * @param handler the block item handler to unregister
      */
-    void unregisterDynamicNoBackpressureBlockItemHandler(NoBackPressureBlockItemHandler handler);
+    void unregisterBlockItemHandler(BlockItemHandler handler);
 
     /**
      * Use this method to send block notifications to all registered handlers.
@@ -74,17 +77,26 @@ public interface MessagingService {
 
     /**
      * Use this method to register a block notification handler. The handler will be called every time new block
-     * notifications arrive. The calls will be on its own virtual thread, every handler registered has its own virtual
-     * thread. It can consume block notifications at its own pace, if it is too slow then it will apply back pressure to
-     * block notification producer. It is too slow by taking too long in the handleBlockNotificationReceived method. If
+     * notifications arrive. The calls will be on its own thread, every handler registered has its own thread. It can
+     * consume block notifications at its own pace, if it is too slow then it will apply back pressure to block
+     * notification producer. It is too slow by taking too long in the handleBlockNotificationReceived method. If
      * handleBlockNotificationReceived is non-blocking, then it will not apply back pressure.
-     * <br>
-     * It has to be called before starting the service.
      *
-     * @param handler the block notification handler to register
-     * @throws IllegalStateException if the service is already started
+     * @param handler             the block notification handler to register
+     * @param cpuIntensiveHandler hint to the service that this handler is CPU intensive vs IO intensive
+     * @param handlerName         the name of the handler, used for thread name and logging
      */
-    void registerBlockNotificationHandler(BlockNotificationHandler handler) throws IllegalStateException;
+    void registerBlockNotificationHandler(
+            BlockNotificationHandler handler, boolean cpuIntensiveHandler, String handlerName);
+
+    /**
+     * Use this method to dynamically unregister a block notification handler. The handler will no longer be called when
+     * new block notifications arrive. You only need to unregister handlers if they need to be unregistered before the
+     * service is shutdown. Shutting down the service will unregister all handlers.
+     *
+     * @param handler the block notification handler to unregister
+     */
+    void unregisterBlockNotificationHandler(BlockNotificationHandler handler);
 
     /**
      * Start the messaging service. This will start the internal threads and start processing messages. All non-dynamic
--- a/block-node/messaging/src/main/java/org/hiero/block/server/messaging/MessagingService.java
+++ b/block-node/messaging/src/main/java/org/hiero/block/server/messaging/MessagingService.java
@@ -1,0 +1,100 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.server.messaging;
+
+import com.hedera.hapi.block.BlockItemUnparsed;
+import java.util.List;
+import org.hiero.block.server.messaging.impl.MessagingServiceImpl;
+
+/**
+ * Service that handles the messaging between the different components of the server. It is used to send block items
+ * and block notifications to the different components. It is not meant to become a general purpose messaging service.
+ */
+public interface MessagingService {
+
+    /**
+     * Use this method to get the new instance of the messaging service.
+     *
+     * @return the instance of the messaging service
+     */
+    static MessagingService createMessagingService() {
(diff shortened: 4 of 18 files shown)


block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * The thread factory used to create the virtual threads for the disruptor. Virtual threads are daemon threads by
+     * default.
+     */
+    public static final ThreadFactory VIRTUAL_THREAD_FACTORY = Thread.ofVirtual()
+            .name("messaging-service-handler", 0)
+            .uncaughtExceptionHandler((thread, throwable) -> LOGGER.log(
+                    Level.ERROR,
+                    "Uncaught exception in thread " + thread.getName() + ": " + throwable.getMessage(),
+                    throwable))
```
jsync-swirlds, line 47: Question, does this prevent the thread from throwing if it returns a Future and `get()` is called on that Future?

Also, we need to make sure the stack trace is in the log as follows:
```suggestion
            .uncaughtExceptionHandler((thread, throwable) -> LOGGER.log(
                    Level.ERROR,
                    "Uncaught exception in thread %s: %s.".formatted(thread.getName(), throwable.getMessage(),
                    throwable)) // Pass the throwable parameter to get a stack trace.
```

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,255 @@
...
+import org.hiero.block.server.messaging.NoBackPressureBlockItemHandler;
+
+/**
+ * Implementation of the MessagingService interface. It uses the LMAX Disruptor to handle block item batches and block
+ * notifications.
+ */
+public class MessagingServiceImpl implements MessagingService {
+
+    /** The size of the ring buffer used by the disruptor. */
+    public static final int RING_BUFFER_SIZE = 1024;
```
jsync-swirlds, line 33: We have configuration for this value, and should probably use it.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jasperpotts: these threads are not used with futures and my understanding is if we do not add a handler then the exception will just be silently lost

jasperpotts: This will happen if in any code that handles items or notifications from messaging services and throws an uncaught runtime exception. Feels like a complete nightmare to debug if not logged and is silently ignored.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,255 @@
...
+
+    /** Collect all block notification handlers till start then register them with the disruptor. */
+    private final ArrayList<EventHandler<BlockNotificationRingEvent>> blockNotificationHandlers = new ArrayList<>();
+
+    /** Map of dynamic no back pressure block item handlers to their threads. So that we can stop them */
+    private final Map<NoBackPressureBlockItemHandler, Thread> dynamicNoBackPressureBlockItemHandlers = new HashMap<>();
+
+    /** Map of dynamic no back pressure block item handlers to their event processors. So that we can stop them */
+    private final Map<NoBackPressureBlockItemHandler, BatchEventProcessor<BlockItemBatchRingEvent>>
+            dynamicNoBackPressureBlockItemHandlerToEventProcessor = new HashMap<>();
```
jsync-swirlds, line 131: The back pressure/no-back-pressure items are a little concerning here.  Is there no way we can pre-register a thread pool for "back pressure" services that are dynamic loaded?  That way we have back-pressure, but still dynamic load all of the services.  There might be a limit to how many services we can support with that, but if we make the number of such threads configurable, the operator can set that count to match the number of services in the deployment.  Then we just allocate threads from that pool as each service is registered, and throw a fatal error if the pool is exhausted.

I'm quite concerned we're backing ourselves into a corner where we require services to be present for all block nodes; and I don't think we should require _any_ service (only the facilities) be present in all block nodes.
Even publisher, verification, persistence are all strictly _optional_ because it is _entirely_ reasonable to have block nodes without any or all of them.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * The disruptor that handles the block item batches. It is used to send block items to the different handlers.
+     * It is a single producer, multiple consumer disruptor.
+     */
+    private final Disruptor<BlockItemBatchRingEvent> blockItemDisruptor = new Disruptor<>(
+            BlockItemBatchRingEvent::new,
+            RING_BUFFER_SIZE,
+            VIRTUAL_THREAD_FACTORY,
+            ProducerType.SINGLE,
+            new SleepingWaitStrategy());
```
jsync-swirlds, line 107: Do Virtual Threads really make sense here?
Virtual threads are intended for when the thread spends most of its time blocked, but do we really expect services to spend most of the time blocked (for example, on I/O or Networking)?  Verification and State management, for instance, I expect will be mostly _running_ and only very rarely blocked.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jsync-swirlds: Makes sense, if never used in a way that something else will catch the exception, then we definitely need the handler.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jsync-swirlds: Note, it's not just those two; I can think of many potential third-party services with the same characteristics.

jasperpotts: We can change our mind but my thought is yes. The reason is each of these threads will call into services like persistence, verification etc. Also call into client handlers to send data to streaming clients. Each of those uses should use the thread that messaging calls them on. Not try and offload into background and quickly return. That way if they are slow(expected) then that will apply back pressure. If they return quickly and do internal threads and queues then all back pressure is broken. Each of those services should be taking the block items sent to them by messaging and doing something blocking with them. Writing to disk, sending to network etc. Only one that is CPU bound is verification but hopefully fine.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jasperpotts: By dynamic I only mean after messaging is started. We could make them all dynamic if we need I know how now. But though we should be able to detect all plugins and register everything then start messaging when all plugins are registered. The dynamic is then only needed for new streaming clients. I am assuming no hot-plugging of plugins while server is running. Is that a wrong assumption?

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jasperpotts: Added config

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jsync-swirlds: I think State management will also be CPU bound (hashing).  Third Party services may also be CPU bound (fraud detections, pattern analysis for HFT, arbitrage bots, etc...).
Not suggesting we switch everything to non-virtual threads, just that it _may_ be reasonable to (eventually) add an "are you CPU bound" flag when registering a handler...

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jsync-swirlds: I wasn't sure of the intent.  If we're making Messaging a "late start" facility (it starts _after_ all deployed services register), then it makes sense.

We'll probably need to document somewhat carefully how these are intended to be used.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    public MessagingServiceImpl(final MessagingConfig config) {
+        blockItemDisruptor = new Disruptor<>(
+                BlockItemBatchRingEvent::new,
+                config.queueSize(),
+                VIRTUAL_THREAD_FACTORY,
+                ProducerType.SINGLE,
+                new SleepingWaitStrategy());
+        blockNotificationDisruptor = new Disruptor<>(
+                BlockNotificationRingEvent::new,
+                config.queueSize(),
```
Nana-EC, line 145: Q: should we have different queueSizes for the two ring buffers. As is changing one as part of a perf optimization would modify the other.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,300 @@
...
+                VIRTUAL_THREAD_FACTORY,
+                ProducerType.SINGLE,
+                new SleepingWaitStrategy());
```
Nana-EC, line 148: We discussed the different strategies a while back and how each makes a tradeoff. I believe `SleepingWaitStrategy` optimizes for reducing CPU intensiveness as risk of higher latency, whereas something like `BusySpinWaitStrategy` does the opposite.

Is it worth supporting a config option here to allow operators to set the desired strategy based on their deployment and use case considerations?

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,300 @@
...
+     *
+     * @return the configuration for the messaging service
+     * @throws IllegalStateException if no ConfigurationBuilderFactory implementation is found
+     */
+    public static MessagingConfig getConfig() {
+        ConfigurationBuilderFactory configurationBuilderFactory = null;
+        final ServiceLoader<ConfigurationBuilderFactory> serviceLoader =
+                ServiceLoader.load(ConfigurationBuilderFactory.class, MessagingServiceImpl.class.getClassLoader());
+        final Iterator<ConfigurationBuilderFactory> iterator = serviceLoader.iterator();
+        if (iterator.hasNext()) {
```
Nana-EC, line 163: Note, this assumes we'll only ever have one implementation matching and will pick the first in the iterator if more than one.
Should we loop over and pull the expected service out based on an additional identification criteria?

Probably fine for now but something to consider down the line.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    public void registerDynamicNoBackpressureBlockItemHandler(NoBackPressureBlockItemHandler handler) {
+        final RingBuffer<BlockItemBatchRingEvent> ringBuffer = blockItemDisruptor.getRingBuffer();
+        final SequenceBarrier barrier = ringBuffer.newBarrier();
+        final EventHandler<BlockItemBatchRingEvent> eventHandler = (event, sequence, endOfBatch) -> {
+            // send on the event block items
+            handler.handleBlockItemsReceived(event.get());
+            // check if the event processor is too far behind
+            final double percentageBehindHead =
+                    (100d * ((double) (barrier.getCursor() - sequence) / (double) ringBuffer.getBufferSize()));
+            if (percentageBehindHead > 80) {
```
Nana-EC, line 210: nit: 80% should be a configureable number an operator can set

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,300 @@
...
+     */
+    @Override
+    public void shutdown() {
+        // Shuts down all the threads handling events.
+        blockItemDisruptor.shutdown();
+        blockNotificationDisruptor.shutdown();
+        // Stop all the dynamic no back pressure block item handlers
+        for (Thread thread : dynamicNoBackPressureBlockItemHandlers.values()) {
+            thread.interrupt();
+        }
```
Nana-EC, line 298: Q: Should there be logic here to unregister the non dynamic bloc handlers - per the comments on `shutdown()` in the interface.

Nana-EC (commented):
Nice work, 1st pass.
Will circle back for a 2nd if still open on Tuesday

block-node/messaging/src/main/java/org/hiero/block/server/messaging/MessagingService.java
```diff
@@ -0,0 +1,100 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.server.messaging;
+
+import com.hedera.hapi.block.BlockItemUnparsed;
+import java.util.List;
+import org.hiero.block.server.messaging.impl.MessagingServiceImpl;
+
+/**
+ * Service that handles the messaging between the different components of the server. It is used to send block items
+ * and block notifications to the different components. It is not meant to become a general purpose messaging service.
```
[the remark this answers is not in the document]
Nana-EC: Earlier we'd spoken about considerations for Services to be able to communicate their own details as needed usign the message service.
@jsync-swirlds is that still intended as part of the design for later or something removed and now out of scope?

block-node/messaging/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+
+// Remove the following line to enable all 'javac' lint checks that we have turned on by default
+// and then fix the reported issues.
+tasks.withType<JavaCompile>().configureEach { options.compilerArgs.add("-Xlint:-exports") }
+
+mainModuleInfo {
+    annotationProcessor("dagger.compiler")
+    annotationProcessor("com.google.auto.service.processor")
+    runtimeOnly("com.swirlds.config.impl")
+    runtimeOnly("org.apache.logging.log4j.slf4j2.impl")
```
ata-nas, line 14: Do we need/use `slf4j2`?

block-node/messaging/src/main/java/org/hiero/block/server/messaging/BlockNotification.java
```diff
@@ -0,0 +1,19 @@
...
+ * Simple record block notifications.
+ *
+ * @param blockNumber the block number this notification is for
+ * @param type the type of notification
+ */
+public record BlockNotification(long blockNumber, Type type) {
+    /**
+     * Enum representing the type of block notification.
+     */
+    public enum Type {
```
ata-nas, line 14: For this enum, should we also add a `BLOCK_PERSISTENCE_FAILURE` option? Persistence could fail for multiple reasons.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/MessagingService.java
```diff
@@ -0,0 +1,100 @@
...
+    void registerBlockNotificationHandler(BlockNotificationHandler handler) throws IllegalStateException;
+
+    /**
+     * Start the messaging service. This will start the internal threads and start processing messages. All non-dynamic
+     * handlers must have been registered before calling this.
+     */
+    void start();
+
+    /**
+     * Stop the messaging service. This will stop the internal threads and stop processing messages. All handlers will
```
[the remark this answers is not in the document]
ata-nas: +1 @Nana-EC I was thinking about the same.

block-node/messaging/src/test/java/org/hiero/block/server/messaging/MessagingServiceBlockItemTest.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+public class MessagingServiceBlockItemTest {
+    /**
+     * The number of items to send to the messaging service. This is twice the size of the ring buffer, so that we can
+     * test the back pressure and the slow handler.
+     */
+    public static final int TEST_DATA_COUNT = MessagingServiceImpl.getConfig().queueSize() * 2;
+
+    /**
+     * Simple test to verify that the messaging service can handle multiple block notification handlers and that
```
ata-nas, line 29: Javadoc on line 29 seems to be unfinished I believe?

block-node/messaging/src/test/java/org/hiero/block/server/messaging/MessagingServiceBlockItemTest.java
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @throws Throwable if there was an error during the test
```
ata-nas, line 31: Not having the `@throws` in javadoc for tests is better IMO, less clutter and generally we always expect that if an exception happens there is a problem with the test or env.

block-node/messaging/src/test/java/org/hiero/block/server/messaging/MessagingServiceBlockNotificationTest.java
```diff
@@ -0,0 +1,203 @@
...
+public class MessagingServiceBlockNotificationTest {
+
+    /**
+     * The number of items to send to the messaging service. This is twice the size of the ring buffer, so that we can
+     * test the back pressure and the slow handler.
+     */
+    public static final int TEST_DATA_COUNT = MessagingServiceImpl.getConfig().queueSize() * 2;
+
+    /**
+     * Simple test to verify that the messaging service can handle multiple block notification handlers and that
```
ata-nas, line 31: Javadoc on line 31 seems to be unfinished?

ata-nas (commented):
@jasperpotts looks good on first pass. In addition to what the others have added, I add just a couple of nits.

block-node/messaging/build.gradle.kts
jsync-swirlds: Good catch, we should not be using that.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/MessagingService.java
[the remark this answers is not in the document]
jsync-swirlds: As a first pass, this works well enough.  I suspect we'll need more than this, but I have no direct evidence of that at this time.
We're working to keep the complexity to a minimum, and starting very simple then adding as we find it to be properly necessary makes sense for that approach.

block-node/messaging/src/main/java/module-info.java
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+module org.hiero.block.messaging {
+    uses com.swirlds.config.api.spi.ConfigurationBuilderFactory;
+
+    exports org.hiero.block.server.messaging;
+
+    requires transitive com.swirlds.config.api;
```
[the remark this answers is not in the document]
jasperpotts: The aim is to use platform base services in Block Node. We have gone through a pile of effort to create 3rd party dependency free base services there that are fast and minimal.  We can have a bigger discussion about this in person at some point on if that is right call.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/MessagingService.java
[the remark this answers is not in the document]
jasperpotts: This was very intentional. I have been burnt in the past when generic messaging just becomes another type of mess like having a pile of spaghetti interconnections. So trying to keep messaging as small and tight as possible. We can add extra message times if we have to but they will have to really fight for their existence. Simple dependencies should be handled by services using each other's public API if they depend on each other. Though again we should fight to minimize those dependencies, but they can exist if they are 100% needed.  Messaging is only designed to solve the core bulk communication needed by the majority of services. Which is basically the flow of blocks.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/BlockNotification.java
```diff
@@ -0,0 +1,19 @@
...
+        BLOCK_VERIFIED,
```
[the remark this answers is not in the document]
jasperpotts: Don't think so, we can add if we need it later but they will know that from block items arriving.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/MessagingService.java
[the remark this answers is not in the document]
jasperpotts: The way Disruptor is designed is you create an instance, register all handlers then start. Those handlers have to be registered before starting the services and remain registered till you call shutdown on the disruptor. So there is no need to have API to unregister non-dynamic handlers. As there life cycle is to exist till shutdown. Handlers do not have to be unregistered on shutdown even dynamic ones. It will happen automatically.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jasperpotts: I am trying to avoid config unless absolutely necessary. We have so much config in consensus node and have no idea what to do with it. Most config is just guesses or set and pray, not actually needed.  Most of these settings should be decided based on testing and just hard code in. Only if a setting needs to be configured at runtime by a ops person should it be exposed as config. If we made everything config I could see us having 1000s of config params, can you even imagine trying to document what they all do and best practices for setting them. We should discuss the philosophy with the team.

In this case 1024 should be plenty, if we have issues in performance testing maybe we will tweak it once or twice and then should be happy.

jasperpotts: If everything is well designed then this setting should not be important. In almost all cases we should be within a couple items of the head of the ring buffer.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jasperpotts: Now that we have batched block items then we are really not that latency sensitive. The difference in this choice is less than 1ms, +/- a couple milliseconds should not be a problem for block nodes. So I went with save the CPU for something more important than saving microseconds of latency. Again we can revisit this later if that is the wrong decision it is simple to change in the code but I feel it should be the correct one. Now that we are doing batches the whole Disruptor Ring Buffer is probably over-kill and we could go for a much simpler solution if needed which is why I am trying hard to not expose any disruptor/ring buffer concepts through the public API of the messaging service. So we can swap the whole implementation around as needed later. When we fully understand what is needed from having a complete block node to test we might decide this needs to be more optimized or we go with a lighter weight approach. Hopefully we find what I have built here just find and more on. :-)

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jasperpotts: See above comment on why less configuration. Again in this case 80% should be a fine number if it is not we can change the code. It should not need to be adjusted often or be that sensitive. Ideally it should work fine if it was 90% or 30%. If it needs adjustment by a ops person then we have done something wrong in the design.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
jasperpotts: explained above

block-node/messaging/build.gradle.kts
jasperpotts: I don't think so I am not sure why the gradle scripts said it had to be there. I had to add these otherwise the build failed.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/BlockNotification.java
jasperpotts: This comes down to what would other services do with that information? Off the top of my head I can't see any way they can do something useful with it. It seems the persistence services should retry a few times then if it can no longer persist anything then the block node is in real trouble. In that case we either shut it down or maybe more to a read-only mode where we can't accept new blocks. We should discuss how to handle this.

jasperpotts: The easy answer is to just send a message and leave the handling problem for someone else. But we really need to work out when it can happen and what the right action of the block node as a whole is in that case. Once we have that we can decide if another notification is needed to support that solution. I would totally expect there to probably be a couple more notification types before we finish.

block-node/messaging/src/test/java/org/hiero/block/server/messaging/MessagingServiceBlockNotificationTest.java
jasperpotts: oops

block-node/messaging/src/test/java/org/hiero/block/server/messaging/MessagingServiceBlockItemTest.java
jasperpotts: oops

block-node/messaging/build.gradle.kts
jsync-swirlds: It appears this may be a consequence of having transitive dependencies on the following in Base and Server:
```
    requires transitive com.swirlds.common;
    requires transitive com.swirlds.config.api;
    requires transitive com.swirlds.config.extensions;
```

I see the same SLF4J dependency everywhere except test suites and common, so it may be something to look into further.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/BlockNotification.java
jsync-swirlds: We do have a response to consensus for failed persistence; as that was felt to be something important enough to consensus to report (so the CN's can connect to a different BN immediately).

block-node/messaging/build.gradle.kts
jsync-swirlds: Update, I found a way to remove this dependency without the CI checks complaining.
Sending a PR for that shortly.

ata-nas: @jsync-swirlds do you refer to #871? I guess with the addition that you have made, either this PR or that one will rebase and resolve the unwanted dependency.

block-node/messaging/src/main/java/org/hiero/block/server/messaging/BlockNotification.java
ata-nas: @jasperpotts I see, I guess judging by other discussions we could always add the option if we can ever justify it's existence, and also seems like there should be further discussions on related topics as you mentioned, for example how we should handle certain cases and document the behavior of the system. For now please consider this thread resolved at your discretion (in case there is something else to add).

block-node/messaging/src/main/java/org/hiero/block/server/messaging/impl/MessagingServiceImpl.java
Nana-EC: Makes sense. Thanks

jsync-swirlds (approved):

ata-nas (dismissed):
🔥

jsync-swirlds (approved):

block-node/messaging/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
```
AlfredoG87, line 14: should we link this with a todo?
```suggestion
    // @todo(871) remove as part of ticket
    runtimeOnly("org.apache.logging.log4j.slf4j2.impl")
```

AlfredoG87 (approved):
LGTM

block-node/messaging/build.gradle.kts
AlfredoG87: just a nit,  feel free to resolve and merge.

jsync-swirlds: I'll make sure it's handled as part of 871 if this merges first.

Nana-EC (approved):
LG

PR #871, opened 2025-03-17
chore: Remove unneeded dependency on slf4j

### LINKED ISSUE #870: Remove unnecessary and unused dependency on SLF4J
### Story Form

N/A

### Technical Notes

We have a SLF4J dependency only because some CI checks required it.  We should work out how to remove that dependency declaration, as it is nothing but excess baggage currently.

jsync-swirlds:
* Removed dependency from Common to swirlds-common that only provided 2 constants.
* Removed slf4j from build "runtime" dependencies.
* Verified that gradle checks do not require the dependency

## Reviewer Notes
Just a quick PR to remove an unused runtime dependency from the gradle build files.
This also cleans up an unnecessary dependency in the Common module.  Much as `com.swirlds.common` might seem basic from the name, it is actually a _very_ large module, with a vast dependency tree.

## Related Issue(s)
 Fixes #870.

### DIFF
--- a/common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
+++ b/common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
@@ -4,7 +4,6 @@ package org.hiero.block.common.hasher;
 import com.hedera.hapi.block.BlockItemUnparsed;
 import com.hedera.hapi.block.stream.BlockProof;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
-import com.swirlds.common.crypto.DigestType;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.nio.ByteBuffer;
 import java.security.MessageDigest;
@@ -21,9 +20,18 @@ public final class HashingUtilities {
     }
 
     /**
-     * The size of an SHA-384 hash in bytes.
+     * The size of a SHA-384 hash, in bytes.
      */
-    public static final int HASH_SIZE = DigestType.SHA_384.digestLength();
+    public static final int HASH_SIZE = 48;
+
+    /**
+     * The standard name of the SHA2 384-bit hash algorithm.
+     *
+     * This value must match what is declared for the
+     * <a href="https://docs.oracle.com/en/java/javase/21/docs/specs/security/standard-names.html#messagedigest-algorithms">
+     * standard message digest names</a>.
+     */
+    public static final String HASH_ALGORITHM = "SHA-384";
 
     /**
      * Returns the SHA-384 hash of the given bytes.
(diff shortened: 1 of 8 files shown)


common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -21,9 +20,18 @@ private HashingUtilities() {
...
+    public static final int HASH_SIZE = 48;
+
+    /**
+     * The standard name of the SHA2 384-bit hash algorithm.
+     *
+     * This value must match what is declared for the
+     * <a href="https://docs.oracle.com/en/java/javase/21/docs/specs/security/standard-names.html#messagedigest-algorithms">
+     * standard message digest names</a>.
+     */
+    public static final String HASH_ALGORITHM = "SHA-384";
```
ata-nas, line 34: Why the switch here?

common/src/main/java/org/hiero/block/common/hasher/StreamingTreeHasher.java
```diff
@@ -17,7 +16,7 @@ public interface StreamingTreeHasher {
     /**
      * The length of the hash produced by this hasher.
      */
-    int HASH_LENGTH = DigestType.SHA_384.digestLength();
+    int HASH_LENGTH = 48;
```
ata-nas, line 19: Why the switch here?

ata-nas (dismissed):
LG, just a couple of questions.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
jsync-swirlds: Because having `block-node-common` depend on `swirlds-common` (which itself depends on a huge array of added libraries) dragged a large dependency tree into everything.  It turns out that consensus has been adding a lot of things to `swirlds-common` that are quite specific to consensus and it's no longer a true "base layer" module.  To the extent possible, we should depend on specific lower-level modules and avoid `swirlds-common`.  One of the difficult things in state management will be the need to manage the much larger dependency tree for that plugin.
The _only_ thing we were using from that module is a pair of constant values that will not every change without fairly widespread code changes, so making them local constants is entirely reasonable.

common/src/main/java/org/hiero/block/common/hasher/StreamingTreeHasher.java
jsync-swirlds: same answer as above.

Nana-EC (approved):
Nice one

ata-nas (approved):
🔥

Nana-EC (approved):

AlfredoG87 (approved):

PR #873, opened 2025-03-19
docs: Block content proof design

### LINKED ISSUE #857: Design Document for BlockContentProof
### Story Form

As a BlockNode Developer
I want to review, discuss and plan the changes needed for a BlockContentProof API and feature.
So that needs a Design document reviewed and approved by the team.

### Technical Notes

This will be worked together with the API proposal for the protobuf changes

AlfredoG87:
## Reviewer Notes

Design Document for BlockContentProof Feature

Fixes #857


block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Terms
+
+<dl>
+  <dt>BlockNumber</dt>
+  <dd>A value that represents the unique number (identifier) of a given `Block`.
+  It is a strictly increasing `long` value, starting from zero (`0`).</dd>
+
+  <dt>BlockItem</dt>
+  <dd>Every block is comprised of a repeated set of block_items, there are many types of block_items, but for the purpose of understanding the BlockContentProof we can divide them into 2 large groups, `inputs` and `outputs` each Kind of BlockItem can belong to these groups and this is very important in order to calculate correctly the BlockRootHash and a BlockContentProof. </dd>
```
jsync-swirlds, line 36: I wonder if we need this detail here.  The exact mechanics of computing a block hash are defined in the Block Stream HIP, and _should not_ be relevant here except that the block item is placed in the _correct location_ within the Block Merkle tree (wherever the HIP defines that to be).

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+
+  <dt>BlockItemHash</dt>
+  <dd>Is the SHA-384 hash of the block_item </dd>
+
+  <dt>MerkleSiblingHashes</dt>
+  <dd>Are the sibling missing hashes from the item all the way up to the BlockRootHash.</dd>
```
jsync-swirlds, line 42: These shouldn't be _missing_.
```suggestion
  <dd>The ordered list of hashes for each "sibling" in the binary tree from the item to be proven all the way up to the tree root.</dd>
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+
+  <dt>BlockContentProof</dt>
+  <dd>A data structure that holds all the needed information to prove a given block item is part of a given block number. Is at least comprised of `BlockNumber, `BlockItemHash`, `MerkleSiblingHashes`, `BlockSignature`</dd>
```
jsync-swirlds, line 45: I wonder if this should be closer to what we return:
```suggestion
  <dd>A data structure that holds all the needed information to prove a given block item is part of a given block number. Is at least comprised of `BlockNumber, `BlockItem`, `MerkleSiblingHashes`, `BlockSignature`</dd>
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+
+  <dt>BlockRootHash</dt>
+  <dd>Is a "special" merkle tree defined by the `Hiero` DLT that is comprised by `previousBlockHash`, `inputsMerkleTree`, `outputsMerkleTree` and `stateRootHash`.</dd>
```
jsync-swirlds, line 48: This is defined in the Block Streams HIP, not by the DLT, and we shouldn't specify it here (as it's subject to change out of sequence with this document).
```suggestion
  <dd>The SHA2-384 value at the root of a carefully defined, partially truncated, binary merkle tree defined by the Block Streams HIP (1056).</dd>
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+
+  <dt>Block Signature</dt>
+  <dd>Is the Signed BlockRootHash by the Network, its purpose is to verify that the BlockRootHash of a given block was signed by the network.</dd>
```
jsync-swirlds, line 51: The wording here is a bit awkward.
```suggestion
  <dd>A TSS signature of a BlockRootHash provided by the consensus network in a `BlockProof` item. Its purpose is to verify that the full content of a given block was generated by and agreed upon by the consensus network as a whole.</dd>
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+## API
+
+### block_proof.proto
+
+```protobuf
+message MerkleSiblingHash {
+    bool is_first = 1;
+    bytes sibling_hash = 2;
+}
+```
```
jsync-swirlds, line 64: Not major, but we usually don't want to re-specify a pre-existing protocol buffer message.

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+}
+
+message BlockContentProofResponse {
+    BlockContentProofResponseCode status = 1;
+    com.hedera.hapi.block.stream.BlockContentProof block_content_proof = 2;
+}
+
+service BlockContentProofService {
+    rpc getBlockContentProof(BlockContentProofRequest) returns (BlockContentProofResponse);
+}
```
jsync-swirlds, line 100: For new messages, should we include the full specification?

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+## Entities
+
+### BlockContentProof
+
+BlockContentProof is a new record that contains all the necessary data to compute the block content proof. It contains the following fields:
```
jsync-swirlds, line 107: This isn't all of the necessary data, as we still need the block itself (including block proof item).  Should we add that value here, or reword this sentence?
```suggestion
BlockContentProof is a new record that contains the fields necessary to compute the block content proof when combined with the retrieved block data. It contains the following fields:
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+- block_item: BlockItem,
+- block_item_hash: Bytes,
+- sibling_hashes: List< MerkleSiblingHash>,
+- block_root_hash: Bytes,
+- block_signature: Bytes
+
+### PBJBlockContentProofServiceProxy
+
+PBJBlockContentProofServiceProxy is the entity responsible for handling the block content proof requests. It provides the implementation for the `getBlockContentProof` rpc endpoint.
+Handles the verification of webServerStatus, handles exceptions, and wraps the response from the BlockContentProofService in a `BlockContentProofResponse` message.
```
jsync-swirlds, line 118: We aren't wrapping the value so much as transforming it.
```suggestion
Handles the verification of webServerStatus, handles exceptions, and transforms the response from the BlockContentProofService to a `BlockContentProofResponse` message.
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+
+### BlockContentProofService
+
+Requests the Block to the BlockReader (Persistence Module), waits to get it and once it gets the Block it creates a new BlockMerkleTreeInfo record out of that block, after that it uses it to create the BlockContentProof and returns it to the PBJBlockContentProofServiceProxy.
```
jsync-swirlds, line 122: This is very confusing what the service entity is intended to represent. I'm entirely uncertain whether this is supposed to be the whole service, or a class that has the same name.

This is a guess that the intent was to describe the _service_ and not an implementation class, but perhaps the following is clearer?
```suggestion
A java module that embodies the functionality necessary to respond to request for block content proofs by calculating and returning those proofs.
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+
+### BlockReader
+
+The BlockReader is responsible for reading the block items from the block store and providing them to the BlockContentProofService.
```
jsync-swirlds, line 126: We do _not_ want to directly interact with block readers or the block data store here, as there is no guarantee it exists in any particular form (it could be local, remote, stored in a columnar database, or even requested from other block nodes on request).

Perhaps refer to the "Persistence Service" and describe a block read message/response sequence to obtain the necessary block instead?

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+    participant Client as Client
+    participant Proxy as PBJBlockContentProofServiceProxy
+    participant Service as BlockContentProofService
+    participant Reader as BlockReader
+
+    Client->>Proxy:  BlockContentProofRequest
+    Proxy->>Proxy:  Validates request & check web server status
+    Proxy->>Service: BlockContentProofRequest
+    Service->>Reader: Requests Block Number
+    Reader-->>Service: Returns Block items
```
jsync-swirlds, line 160: As we're dealing with already-complete blocks in this case, the return from whatever service manages stored blocks should be a `Block`, not the individual items.
```suggestion
    Reader-->>Service: Returns Block
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+
+1. BlockContentProofRequest_Success Counter
+2. BlockContentProofRequest_Error Counter
+3. BlockContentProofResponse_Latency Histogram
+
+**Note:** Once we are able to add labels we should add `status` label to the metrics
+
+## Exceptions
+
+Any exception will be handled and return `BLOCK_CONTENT_PROOF_INTERNAL_ERROR` response code to the client.
```
jsync-swirlds, line 180: As Jasper noted, we shouldn't return internal errors:
```suggestion
Any exception will be handled and return `BLOCK_CONTENT_PROOF_NOT_AVAILABLE` response code to the client.
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+        bytes block_item_hash = 2;
+        uint32 block_item_index = 3;
+    }
+}
+
+enum BlockContentProofResponseCode {
+    BLOCK_CONTENT_PROOF_UNKNOWN = 0;
+    BLOCK_CONTENT_PROOF_SUCCESS = 1;
+    BLOCK_CONTENT_PROOF_NOT_FOUND = 2;
+    BLOCK_CONTENT_PROOF_INSUFFICIENT_BALANCE = 3;
```
jsync-swirlds, line 88: We don't even know if there will be any monetization or charging, and we certainly won't be checking balance in this service.
We do need a response for "not available" (e.g. could not get the block, or could not compute the proof).
```suggestion
    BLOCK_CONTENT_PROOF_NOT_AVAILABLE = 3;
```
This would replace `INTERNAL_ERROR` below as well.

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+        com.hedera.hapi.block.stream.BlockItem block_item = 1;
+        bytes block_item_hash = 2;
+        uint32 block_item_index = 3;
+    }
+}
+
+enum BlockContentProofResponseCode {
+    BLOCK_CONTENT_PROOF_UNKNOWN = 0;
+    BLOCK_CONTENT_PROOF_SUCCESS = 1;
+    BLOCK_CONTENT_PROOF_NOT_FOUND = 2;
```
jsync-swirlds, line 87: What does not found mean?  The _item_ was not found, or the _block_ was not found?
I am assuming it means the _item_ was not found.

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+
+1. Client sends a BlockContentProofRequest using `block_item_index` to the BN for a block_item that has a duplicate hash in the block.
+2. BN receives the request and returns a complete BlockContentProofResponse.
+3. Client receives the BlockContentProofResponse and verifies the block_item using the BlockContentProof provided by the BN.
+4. Client verifies that the block_item is the correct one by pre-calculating the merkle proof path before hand for the right item and then comparing the given proof with the expected proof.
+
+### Block Number not found:
+
+1. Client sends a BlockContentProofRequest to the BN for a non-existing block number.
+2. BN receives the request and returns a `BLOCK_CONTENT_PROOF_NOT_FOUND` response code to the client.
```
jsync-swirlds, line 206: If the _block_ is not available we should return NOT_AVAILABLE.
I would reserve NOT_FOUND for not finding the _item_ within the block.

block-node/server/docs/design/block-content-proof.md
AlfredoG87: just removed the message, since adding a link to the whole API spec, they can find it if needed.

block-node/server/docs/design/block-content-proof.md
AlfredoG87: I added a Link to the full spec API, I think that due to the comments, it makes it too large. I wonder if we should not remove the API section altogether once we merge the API PR, and leave only the references here.

block-node/server/docs/design/block-content-proof.md
AlfredoG87: Improved wording 👍 Thank you!!

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,216 @@
...
+The BlockReader is responsible for reading the block items from the block store and providing them to the BlockContentProofService.
+
+### BlockMerkleTreeInfo
+
+Is a new record type that contains all the necessary data to compute any block item proof in the block. it contains:
+- inputsMerkleTree: List< List< Bytes>>,
+- outputsMerkleTree: List< List< Bytes>>,
+- previousBlockHash: Bytes,
+- stateRootHash: Bytes,
+- blockHash: Bytes,
```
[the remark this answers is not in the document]
AlfredoG87: removed!

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,208 @@
+# Design Document Template
```
Nana-EC, line 1:
```suggestion
# Block Contents Proof Design Document
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,208 @@
...
+5. [Design](#design)
+6. [Sequence Diagram](#sequence-diagram)
+7. [Configuration](#configuration)
+8. [Metrics](#metrics)
+9. [Exceptions](#exceptions)
+10. [Acceptance Tests](#acceptance-tests)
+
+## Purpose
+
+The purpose of this document is to define the design of the BlockContentProof feature.
```
Nana-EC, line 18: nit: we should standardize but initially i thought it was block contents.
Here I see Block Content.
I block has contents but I'm guessing the singular form is because I since content items (BlockItem) is under consideration.

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,208 @@
...
+The BlockContentProof API allows the BN to provide a proof of the content of a block to the client.
+
+## Goals
+
+1. Define the BlockContentProof struct
```
Nana-EC, line 23: By struct to did you mean structure? Logic might be better.

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,208 @@
...
+2. Define the BlockContentProof API
+3. BN should be able to calculate the BlockContentProof given a BlockNumber and BlockItem (or BlockItemHash or BlockItemIndex)
```
Nana-EC, line 25:
```suggestion
3. Given a BlockNumber and BlockItem (or BlockItemHash or BlockItemIndex) a BN should support the calculation of a BlockContentProof
```

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,208 @@
...
+  <dd>Every block is comprised of a repeated set of block_items, there are many types of block_items, but for the purpose of understanding the BlockContentProof we can divide them into 2 large groups, `inputs` and `outputs` each Kind of BlockItem can belong to these groups and this is very important in order to calculate correctly the BlockRootHash and a BlockContentProof. </dd>
+
+  <dt>BlockItemHash</dt>
+  <dd>Is the SHA-384 hash of the block_item </dd>
+
+  <dt>MerkleSiblingHashes</dt>
+  <dd>The ordered list of hashes for each "sibling" in the binary tree from the item to be proven all the way up to the tree root.</dd>
+
+  <dt>BlockContentProof</dt>
+  <dd>A data structure that holds all the needed information to prove a given block item is part of a given block number. Is at least comprised of `BlockNumber, `BlockItem`, `MerkleSiblingHashes`, `BlockSignature`</dd>
```
Nana-EC, line 45:
```suggestion
  <dd>A data structure that holds all the needed information to prove a given block item is part of a given block number. Is at least comprised of the `BlockItem`, `MerkleSiblingHashes`, `BlockSignature`</dd>
```

Nana-EC (commented):
1st pass, looking at Acceptance tests next

block-node/server/docs/design/block-content-proof.md
Nana-EC: I think `BLOCK_CONTENT_PROOF_NOT_FOUND` is still unclear no?
Only the block or block items exist to be found. The proof is based on those.
So the not found should refer to one or both of those

block-node/server/docs/design/block-content-proof.md
jsync-swirlds: Just references makes sense, to me, for an in-tree design document.

block-node/server/docs/design/block-content-proof.md
jsync-swirlds: That would probably make this clearer.
```suggestion
    BLOCK_CONTENT_PROOF_ITEM_NOT_FOUND = 2;
```

block-node/server/docs/design/block-content-proof.md
AlfredoG87: Good Catch!! Thank you!

block-node/server/docs/design/block-content-proof.md
AlfredoG87: yes

block-node/server/docs/design/block-content-proof.md
jsync-swirlds: This did not change...

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,197 @@
...
+7. The `PBJBlockContentProofServiceProxy` wraps the `BlockContentProof` in a `BlockContentProofResponse` and sends it back to the client.
+
+## Sequence Diagram
+
+```mermaid
+sequenceDiagram
+    participant Client as Client
+    participant Proxy as PBJBlockContentProofServiceProxy
+    participant Service as BlockContentProofService
+    participant Reader as BlockReader
```
jsync-swirlds, line 135: Should this be `BlockProvider`?

block-node/server/docs/design/block-content-proof.md
```diff
@@ -0,0 +1,197 @@
...
+### BlockMerkleTreeInfo
+
+Is a record type that contains all the necessary data to compute any block item proof in the block.
+
+## Design
+
+1. The `PBJBlockContentProofServiceProxy` receives the block content proof requests.
+2. If the request is valid and the web server is running, it forwards the request to the `BlockContentProofService`.
+3. The `BlockContentProofService` publish a message requests for the block.
+4. The `BlockProvider` provides the Block to the `BlockContentProofService`. If block is Available.
```
jsync-swirlds, line 123: Do we want to note that this component might be a separate facility or service, and is outside the scope of this design document?

block-node/server/docs/design/block-content-proof.md
AlfredoG87: Yes, good catch, should be fixed

jsync-swirlds (approved):

Nana-EC (approved):
LG

That being said I think we're missing an `BLOCK_ITEM_NOT_FOUND ` or related error case.
We can revisit and add later

AlfredoG87:
> LG
> 
> That being said I think we're missing an `BLOCK_ITEM_NOT_FOUND ` or related error case. We can revisit and add later

I think the last 2 cases are reffering to that:

```
### Block item not found:

1. Client sends a BlockContentProofRequest to the BN for a non-existing block item. (variations of block_item, block_item_hash, block_item_index)
2. BN receives the request and returns a `BLOCK_CONTENT_PROOF_ITEM_NOT_FOUND` response code to the client.

### Block item hash is invalid:

1. Client sends a BlockContentProofRequest to the BN for a block_item with an invalid hash.
2. BN receives the request and returns a `BLOCK_CONTENT_PROOF_ITEM_NOT_FOUND` response code to the client.
```

jsync-swirlds (approved):

PR #875, opened 2025-03-19
fix: restart recovery race condition

### LINKED ISSUE #874: Race Condition on Startup is preventing restart recovery for AckHandler
### Description

There is a `Race Condition` between the `StreamPersistenceHandlerImpl` and `AckHandler`.
if `AckHandler` finishes init before `StreamPersistenceHandlerImpl` the `AckHandler` will think it needs Block `0` instead of the actual block it needs, preventing the system from sending ACKs and moving items from `unverified` to `verified`. Preventing restart recovery.

### Steps to Reproduce

Once enough Blocks are persisted it always happens. Long term run is necessary.

### Additional context

This Fix has already been tested on GKE Long standing tests successfully, by deploying a temporarily `0.7.0-SNAPSHOT` version published directly from this `branch`

_No response_

AlfredoG87:
## Reviewer Notes
This issue was found on `0.6.0-rc2` after 233K blocks have been persisted, GCP restarted everything and the block-node was unable to recover sucessfully, while Persistence and Verification were continue to function properly, the ACK handler was initialized in a bad state.

Fixes #874

### DIFF
--- a/block-node/base/src/main/java/org/hiero/block/server/ack/AckHandlerImpl.java
+++ b/block-node/base/src/main/java/org/hiero/block/server/ack/AckHandlerImpl.java
@@ -74,6 +74,10 @@ public class AckHandlerImpl implements AckHandler {
             // @todo(147) we need to handle new instances that need to start from a different block than 0.
             lastAcknowledgedBlockNumber = -1;
         }
+
+        LOGGER.log(
+                System.Logger.Level.INFO,
+                "AckHandler initialized with lastAcknowledgedBlockNumber: " + lastAcknowledgedBlockNumber);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


jsync-swirlds (approved):

block-node/base/src/main/java/org/hiero/block/server/ack/AckHandlerImpl.java
```diff
@@ -56,8 +56,14 @@ public AckHandlerImpl(
...
         this.blockRemover = Objects.requireNonNull(blockRemover);
         this.metricsService = metricsService;
+    }
+
+    @Override
+    public void registerPersistence(@NonNull final StreamPersistenceHandlerImpl streamPersistenceHandler) {
+        this.streamPersistenceHandler = Objects.requireNonNull(streamPersistenceHandler);
 
         // Initialize lastAcknowledgedBlockNumber from the service status if available
+        // When Persistence registers itself, it means it has already calculated is last persisted block.
```
Nana-EC, line 66:
```suggestion
        // When Persistence registers itself, it means it has already calculated the last persisted block.
```
or
```suggestion
        // When Persistence registers itself, it means it has already calculated its last persisted block.
```

Nana-EC (approved):

block-node/base/src/main/java/org/hiero/block/server/ack/AckHandlerImpl.java
AlfredoG87: we will refactor this in the near future, so comment will/should be deleted eventually 😅

PR #885, opened 2025-03-20
chore: Add a document explaining block node types and tiers

### LINKED ISSUE #884: Create a document describing block node tiers and types
### Story Form

As a Block Node client or operator
I want to understand the different block node types, tiers, and services offered
So that I can make reasonable decisions regarding my use or management of block nodes

### Technical Notes

Please document/defined the different block node references made in various places. Multiple names refer to the same type of block node, and multiple block node types sometimes share a name.  This causes significant confusion, so a set of clear definitions and perhaps a diagram would be helpful.

jsync-swirlds:
* Added document and diagram as requested.

## Reviewer Notes

The goal of this document is to standardize the descriptions of block nodes, clarify the difference between _types_ of block nodes and _services_ offered by a block node, and show, visually, how the various network elements relate to one another.

## Related Issue(s)
Resolves #884


docs/Block-Node-Types.md
```diff
@@ -0,0 +1,120 @@
...
+    distributed application.</dd>
+    <dt>Defined Filter</dt>
+    <dd>A theoretical service provided by a block node specialized to run a
+    plugin that filters the block stream according to well-defined criteria,
+    producing a modified block stream with un-desired or private data removed.
+    Critically, the filtered stream is _still verifiable_ using the Block
+    Proof for each block.</dd>
+    </dl>
+</dd>
+<dt>Block Node Types</dt>
```
Nana-EC, line 78: I think the definition of types is still a bit inconsistent.
The listed types mix quantity of storage, public/private, who operates it as well as the type of APIs exposed.
I don't think who operates it and the APIs exposed should affect the types - both items can be modified without modifying deployed code

Should we standardize the criteria to assist the naming convention
- Access: Private/Public
- Block Persistence: Full/Partial/No History
- State Persistence: None, live

e.g. private full history live state BN, public

Nana-EC (commented):
Great description and illustration here.
The more I think on it the more I think we need to revisit the definition of "service" and "type".
Service seems like it should refer more to the appreciable API behaviour, some of the services to me refer more to the use case the BN is applied to.

I'm thinking on it for additional suggestions

docs/Block-Node-Types.md
jsync-swirlds: I think I'm understanding the disconnect here.

1. I have an error in the definition of a private archive.  It should not discuss "access".
2. I am not trying to define every way a node might be operated, that would explode the number of definitions by a geometric product of several dozen items, resulting in a list in the thousands, and there would be no clarity.
3. I'm not defining the _internal_ services or functionality, only externally observable elements.
4. I'm trying to be at least somewhat generic here, the intended audience is not developers or engineers, the intended audience is more "general interested parties".

I'm updating a few things in definitions to fix the error and clarify some other elements.

Nana-EC: I like the improvements, thanks

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,126 @@
...
+    <dl>
+    <dt>Tier 1</dt>
+    <dd>A Block Node that receives its Block Stream data directly from Consensus
+    Nodes. Tier 1 Block Nodes are critical to the operation of the Consensus
+    Network, and it is expected that, eventually, each Consensus Node operator
+    will need to run their own Tier 1 Block Node in
+    a "Full Node" configuration.</dd>
+    <dt>Tier 2</dt>
+    <dd>A Block Node that receives its Block Stream data from another Block
+    Node (often a Tier 1 Block Node). The data may be received via direct stream
```
Nana-EC, line 118: Did you mean ?
```suggestion
    Node (often a Tier 1 Block Node). The data may be received via direct stream,
```

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,126 @@
+# Block Node Types
+
+## Overview
+
+Block Nodes can be a somewhat slippery concept. The core system encompasses a
```
Nana-EC, line 5: Should we drop the first sentence?
I fell like it won't be applicable much longer especially with efforts like this but I get you're highlighting that it's a weighted term. 
```suggestion
Block Nodes  encompass a
```

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,126 @@
...
+offered and operating intent.
+
+## Block Node Types and Tiers Visualized
+
+![Block Node Type Diagram](assets/Block-Node-Tiers-and-Types.svg)
+
+# Definitions
+
+<dl>
+<dt>Block Node</dt>
```
Nana-EC, line 28: Should this at least talk about parsing/ingesting acquired blocks?
Maybe have Block Streams term go first so this has context.

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,126 @@
...
+    filter the block stream according to well-defined criteria, producing a
+    modified block stream with un-desired or private data removed.
+    Critically, the filtered stream is _still verifiable_ using the Block
+    Proof for each block.</dd>
+    </dl>
+</dd>
+<dt>Block Node Types</dt>
+<dd>
+    <dl>
+    <dt>Rolling-History</dt>
```
Nana-EC, line 84: Q: I keep considering the option of a partial history. A block node that doesn't start from genesis but at a time after for efficiency or business reasons e.g. the data in mind exists later and the operator is fine not to support services for data prior to its start?

Nana-EC (commented):
Looks good, a suggestion and a question

docs/Block-Node-Types.md
jsync-swirlds: I avoided including ingesting/parsing blocks for two reasons:
1. We don't actually parse the blocks; we use a protobuf hack to avoid it.
2. How any given block node gets blocks is highly variable, enough so that making ingest/read-in part of the definition seems to be over-defining to me.

docs/Block-Node-Types.md
jsync-swirlds: Good idea.  I'll add that.

docs/Block-Node-Types.md
jsync-swirlds: Updated.

docs/Block-Node-Types.md
jsync-swirlds: Added.

docs/Block-Node-Types.md
jsync-swirlds: Fixed and text reflowed.

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,133 @@
...
```
georgi-l95, line 89: Another name is Light Node. It will make it easier for operators/developers/users coming from other networks to understand what it does as it is standard for such nodes to be called `light`.

georgi-l95 (approved):
LG, one naming suggestion.

docs/Block-Node-Types.md
jsync-swirlds: Thanks for the suggestion.
I'll add that as a synonym for "Limited History" and "Rolling-History".

Nana-EC (dismissed):
LG.
We can more easily maintain definitions and update over time with this.
Thanks

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,138 @@
...
+    than the network as a whole. Examples include cloud buckets, long-term
+    tape, or replicated local disks. Private archives are often intended for
+    disaster recovery or offline analysis.</dd>
+    <dt>Custom Analysis</dt>
+    <dd>A theoretical service provided by a block node specialized to perform
+    some form of data analysis in, or near, real time.</dd>
+    <dt>General Interledger Bridge</dt>
+    <dd>A theoretical service provided by a block node specialized to bridge an
+    Hiero network to another general-purpose network, such as Cardano, Base,
+    Bitcoin, Ethereum, etc...</dd>
```
AlfredoG87, line 74: I don't think this is a service that a block node can provide 🤔 
in any case is something that warrants more discussion, to me this `specialized block node` is the same as a `Defined Filter` and `Dapp Support`.

`Bitcoin` is not a general-purpose network. and for any other I believe the bridge is a too broad concept to put as part of services that the BN by its own can offer.

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,138 @@
...
+    <dt>Dapp Support</dt>
+    <dd>A theoretical service provided by a block node specialized to support
+    only the requirements of a specific distributed application.</dd>
```
AlfredoG87, line 77: A `Dapp Support` could be just a `Defined Filter` and if is not, why don't we add a `Custom Service`  instead of `Custom Analysis`, `Bridge Support`, `Dapp Support`.

and Define custom service mentioning the `3rd party plugin` system that the BN will support.

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,138 @@
...
+    <dd>A type of node that chooses to manage only "recent" history.<br/>
+    May offer other services that do not require more than a day (or other
+    limited duration) of locally stored history. Most Tier 2 nodes are expected
+    to be this type of node. A "Private Archive" service could be provided
+    by this type of node. Rolling-History provides a particular type of
+    "Partial History" service.</dd>
+    <dt>Full Node</dt>
+    <dd>A type of node that chooses to manage all history and state.<br/>
+    Also offers state management, state proof, content proof, query,
+    inter-ledger bridge, and other core services. This type of node may offer
```
AlfredoG87, line 99: `inter-ledger bridge` should not be considered among "core" services imo.

AlfredoG87 (commented):
Do we want to reference to the available plugins here? seems to me that there is some relationship between a block node type with the services it offers, and then the services with the plugins that it runs.

so a section that includes those relationships would be nice too.

In the diagram the ` inter ledger bridge` is not a BN right?

docs/Block-Node-Types.md
jsync-swirlds: Work is starting on defining exactly this functionality for inter-ledger, and the expressed request is to have BN offer this service as a custom plugin (at some point, probably not this year, though).

Bitcoin shouldn't be listed, and I'll remove that item.  All of the rest _should_ be possible.

docs/Block-Node-Types.md
jsync-swirlds: This document is not intended for engineers or operators.
I've deliberately left off the technical details of how things work, and focused on "product like" names, rather than implementation.

docs/Block-Node-Types.md
jsync-swirlds: I'll rearrange, I intended core services to be separate, not inclusive of the preceding.

docs/Block-Node-Types.md
jsync-swirlds: Note, I clarified language in the overview to further emphasize the intended audience, and disambiguate terms.

AlfredoG87 (approved):

AlfredoG87 (approved):

ata-nas (approved):
LGTM 🔥

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,138 @@
...
+query capabilities for that network. A Mirror Node is a client of one or more
+Block Nodes, which deliver the block stream to the Mirror Node for indexing,
+storage, and related purposes.</dd>
+<dt>Consensus Node</dt>
+<dd>One node in the consensus network of an Hiero network. Consensus Nodes
+perform the critical work of executing the network consensus algorithm,
+executing transactions in consensus order, maintaining a current network state,
+and producing the Block Stream.</dd>
+<dt>Private Sphere</dt>
+<dd>An Hiero network operated on behalf of a private entity.<br/>
```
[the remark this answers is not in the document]
jsync-swirlds: Technically, `An` is recommended before words that start with `H`, but either is considered acceptable.

docs/Block-Node-Types.md
```diff
@@ -0,0 +1,138 @@
...
+variety of purposes and in myriad ways. A Block Node may be deployed with any,
+or many, set or subset of service plugins built to provide specific capabilities
+within the context of processing the block stream.</dd>
+<dt>Mirror Node</dt>
+<dd>A service available in Hiero networks that provides extensive history and
+query capabilities for that network. A Mirror Node is a client of one or more
+Block Nodes, which deliver the block stream to the Mirror Node for indexing,
+storage, and related purposes.</dd>
+<dt>Consensus Node</dt>
+<dd>One node in the consensus network of an Hiero network. Consensus Nodes
```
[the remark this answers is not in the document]
jsync-swirlds: Technically, `An` is recommended before words that start with `H`, but either is considered acceptable.

georgi-l95 (approved):

PR #887, opened 2025-03-20
chore: Remove docker determinism checks for now.

### LINKED ISSUE #835: Remove docker determinism checks for now
### Story Form

As a block node engineer
I want to make significant changes to the system efficiently
So that I can deliver value to our users as rapidly as possible.

### Technical Notes

We have a very complex and somewhat brittle process for "docker determinism" in the CI currently that causes quite a bit of rework for an uncertain gain.
We intend to remove this check and related code for now, and build a new CI check, when needed, that is more robust and less sensitive to changes that have no impact on repeatability of OCI image generation.

jsync-swirlds:
* Removed the docker determinism checks
    * We will revisit this in a future issue, when we know what the actual requirements are.

## Reviewer Notes
We're removing these checks for now because they do impose extra time and hassle when making major changes and moves.  We're also not certain they're properly needed, as we do not need to ensure that every block node is running the exact same image (and probably _should not_ do so).

## Related Issue(s)
Resolves #835.


jsync-swirlds:
Note, I found a solution for _partial_ testing, at least enough to know if the syntax is correct, and I have run that validation/testing process as well as actionlint and action-validator to find what issues I could.  Full testing requires all the secrets and infrastructure of the CI environment, which simply isn't available until after this merges, sadly.

ata-nas (approved):
LGTM 🔥

.github/workflows/release-push-image.yaml
```diff
@@ -38,14 +40,6 @@ jobs:
...
-  check-docker:
-    name: Docker
-    uses: ./.github/workflows/zxc-verify-docker-build-determinism.yaml
-    with:
-      ref: ${{ github.event.inputs.ref || '' }}
-      java-distribution: ${{ inputs.java-distribution || 'temurin' }}
-      java-version: ${{ inputs.java-version || '21.0.6' }}
-
   publish:
     needs: [check-gradle, check-docker]
```
AlfredoG87, line 44: `check-docker` should be removed.
```suggestion
    needs: [check-gradle]
```

AlfredoG87 (changes requested):
This will break the Build.

There is a way we can test this without merging. will ping you on how.

.github/workflows/release-push-image.yaml
jsync-swirlds: Thanks.

.github/workflows/release-push-image.yaml
```diff
@@ -38,16 +40,8 @@ jobs:
       java-distribution: ${{ inputs.java-distribution || 'temurin' }}
```
AlfredoG87, line 40: are we keeping the gradle deterministic intentionally?

AlfredoG87 (commented):
Do we want to keep the Gradle Determinism checks?
Do we want to simplify the `Dockerfile` as well, there is a bunch of overhead code that was added to a rather very simple `Dockerfile` when the deterministic was introduced.

AlfredoG87 (approved):
If we want to keep the `Dockerfile` and gradle deterministic check. is passing :)

image

Nana-EC (approved):
LG

.github/workflows/release-push-image.yaml
jsync-swirlds: We could remove it, but I think we might want to keep it.
- Gradle determinism (probably should rename to "repeatable build", as it's not strictly speaking "determinism") has value to our team because it ensures that we have a reliable build process and can reliably re-create a release build when needed.
- It also has no added reliance on folders and files, so it doesn't generally break when the main build is working correctly.

What are your thoughts?  Should we keep the repeatable build checks, or remove them?

jsync-swirlds:
> Do we want to keep the Gradle Determinism checks? Do we want to simplify the `Dockerfile` as well, there is a bunch of overhead code that was added to a rather very simple `Dockerfile` when the deterministic was introduced.

Thanks for the suggestion.

Simplifying the Dockerfile where possible would be great.  I'll add an issue to do that separately, as I suspect we'll _also_ need to make some other changes to bring the new modules and plugins together in a way that's more easily adjusted for various use cases.

.github/workflows/release-push-image.yaml
AlfredoG87: This is how I understand it. 

For a container image to be deterministic, the build also needs to be deterministic like 2 layers of deterministism.

Yes, it also gives us some repeatability (but is just a side effect) and to be honest, we are not building in many different platforms 😅 for that matter. mostly just flavors of linux.
```
      matrix:
        os:
          - ubuntu-24.04
          - ubuntu-22.04
          - hiero-block-node-linux-medium
```

The Check is adding about 5 minutes to the process, so I don't think is impactful.

Let's leave it there 👍

andrewb1269hg (approved):

Issue #505, opened 2025-01-16
EPIC: Support Service Loaders and fully independent Service interfaces (SPI)


mattp-swirldslabs:
As an Block Node User
I want to be able to extend the software with an SPI plugin
So that I can satisfy my own use cases by extending the product without recompilation,
  deploy different configurations without recompilation or reliance on editing a global,
  configuration, support private/proprietary third-party services deployed easily by
  operators of community block nodes, and several other critical use cases.

## Tech Notes
* We need to establish service interfaces which could leverage SPIs to load implementations
   * Each service will need to be a separate module, rather than mingled together in one
   * Services will need to communicate via messages, not direct method calls
   * Cross Service messages need to be generic, not directed.
   * All services must function correctly regardless of the presence or absence of
     any other service or combination of services.
* Persistence Service
   * This may benefit from permitting multiple such services
   * Alternatively, we might design a separate Archive Service
* Verification Service
* Consumer Service
   * There might be more than one type of consumer service in a deployment.
* Producer Service
* Additional (future/planned) services that should be fully pluggable (and only deployed when needed)
   * Status Service
   * Block Item Proof Service
   * Block Filter Service
   * Redaction and Errata Service
   * State Update Service
   * State Query Service
   * State Proof Service
   * State Snapshot Service
   * Reconnect Service
* Privately developed and proprietary services added to privately operated block nodes without modification to the core codebase.

PR #890, opened 2025-03-22
refactor: refactor to plugins and services

### LINKED ISSUE #762: Create core service facilities
### Goal

Create the core facilities for a service-oriented block node.

### Intention

We intend to create a messaging facility, and a server status facility.  We also intend to _possibly_ create a configuration facility and a metrics facility.  We intend to either add service registration to messaging, or create a separate service registration facility.

### LINKED ISSUE #763: Create a Messaging Facility

### LINKED ISSUE #764: Create a Configuration facility
### Persona

As a Block Node Service Engineer

### Request

I want to connect to a configuration facility

### Goal

So that my service can receive configuration information on initialization

### Technical Notes

This feature is to create a Configuration facility that any newly loaded service may (optionally) connect with to obtain its configuration data.  This data is read (at initialization) from environment and system properties based on definitions provided by the newly loaded service.

### LINKED ISSUE #765: Create a Metrics facility
### Persona

As a Block Node Service Engineer

### Request

I want to connect to a Metrics facility

### Goal

So that I can publish relevant metrics for my service

### Technical Notes

A Metrics facility is needed, to be(optionally)  obtained by a newly loaded service at initialization.  The facility must offer a mechanism for a newly loaded service to create and publish metrics, as needed or desired.

### LINKED ISSUE #813: Initial break out of code base out into multiple java modules
### Goal

Ensure an appropriate balance of block node logic encapsulation and decoupling to ensure support for future customized implementations of core block node features as well as better testing and explicit service management

### Intention

Update the block node code to support java modules with clear goals for each modules.

Initial module split
- block-node-base (temporary module to encompass a few remnants)
- block-node-app (this is just the top-level "start the server" code)
- facility-messaging
- hiero-block-node-spi
- block-node-health
- block-node-publisher
- block-node-subscriber
- block-node-verification
- Block Node Providers
    - block-node-blocks-cloud-archive
    - block-node-blocks-cloud-historic
    - block-node-blocks-file-historic
    - block-node-blocks-file-recent

### Considerations

- gradle builds
- nano service architecture

### LINKED ISSUE #852: Extract persistence into a separate module
### Story Form

As a block node operator
I want to deploy block node without persistence or with alternate persistence
So that I can customize the behavior of the deployed system

### Technical Notes

Extract the persistence code into a separate module.  Once extracted also extract an API module and establish proper dependency flows.

This effort depends on a messaging facility to detangle persistence from legacy-messaging, status, ack, validation, consumers, and producers.
This effort may depend on a registrar/initialization facility to replace global dagger dependencies.

### LINKED ISSUE #853: Extract Verification into a separate module
### Story Form

As a block node operator
I want to deploy block node without verification or with alternate verification
So that I can customize the behavior of the deployed system

### Technical Notes

Extract the verification code into a separate module.  Once extracted also extract an API module and establish proper dependency flows.

This effort depends on a messaging facility to detangle verification from legacy-messaging, status, ack, persistence, consumers, and producers.
This effort may depend on a registrar/initialization facility to replace global dagger dependencies.

jasperpotts:
#505 Quite a code explosion

Closes #813
Closes #852
Closes #853
Closes #762
Closes #763
Closes #764
Closes #765


common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -3,6 +3,7 @@
 
 import com.hedera.hapi.block.stream.BlockProof;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.swirlds.common.crypto.DigestType;
```
jsync-swirlds, line 6: This was removed in an earlier PR because we want to _not_ include swirlds common (because that's not a low-level common library, it has a very large dependency tree).

block-node/publisher/src/main/java/org/hiero/block/node/publisher/BlockStreamProducerSession.java
```diff
@@ -0,0 +1,375 @@
...
+                }
+            }
+            System.out.println("BlockStreamProducerSession.onNext currentBlockState=" + currentBlockState);
+            switch (currentBlockState) {
+                case NEW -> newBlockItems.addAll(items);
+                case PRIMARY -> {
+                    // we are in the primary state, so we can send the items directly to the block messaging service
+                    // this will never be the first items in a block so we can always send UNKNOWN_BLOCK_NUMBER for
+                    // block number
+                    sendToBlockMessaging.accept(new BlockItems(items, UNKNOWN_BLOCK_NUMBER));
```
AlfredoG87, line 272: Even if we are in the middle of a block, do we want to send the block_number, maybe we have it from state?

block-node/base/src/main/java/org/hiero/block/node/base/BlockFile.java
```diff
@@ -183,6 +186,27 @@ public static long nestedDirectoriesMaxBlockNumber(Path basePath, CompressionTyp
...
+        final String fullExtension = BLOCK_FILE_EXTENSION + compressionType.extension();
+        try (var stream = Files.walk(basePath)) {
+            stream.filter(Files::isRegularFile)
+                    .filter(path -> path.getFileName().toString().endsWith(fullExtension))
+                    .forEach(blockFile -> blockNumbers.add(blockNumberFromFile(blockFile)));
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return blockNumbers;
+    }
```
jsync-swirlds, line 208: This will be a _very_ large set, with a quite significant memory impact.  Is there any way we could reduce this to _ranges_ of blocks instead (particularly as the most common case should be a single contiguous range)?

block-node/base/src/main/java/org/hiero/block/node/base/BlockFile.java
```diff
@@ -0,0 +1,233 @@
...
+        // remove the last digits from the block number for the files in last directory
+        final String blockNumberDir = blockNumberStr.substring(0, blockNumberStr.length() - filesPerDir);
+        // start building path to zip file
+        Path dirPath = basePath;
+        for (int i = 0; i < blockNumberDir.length(); i += filesPerDir) {
+            final String dirName = blockNumberStr.substring(i, Math.min(i + filesPerDir, blockNumberDir.length()));
+            dirPath = dirPath.resolve(dirName);
+        }
+        // append the block file name
+        return standaloneBlockFilePath(basePath, blockNumber, compressionType);
```
ata-nas, line 98: @jasperpotts minor bugfix, I have applied locally to my unit tests branch, but I think you would want to have it as well, otherwise you will not see the block files in the expected places.

```suggestion
        return standaloneBlockFilePath(dirPath, blockNumber, compressionType);
```

block-node/app/docker/update-env.sh
```diff
@@ -36,9 +30,15 @@ else
   echo "JAVA_OPTS='-Xms16G -Xmx16G'" >> .env
 fi
 
-# file is mounted in the docker-compose.yml, changes to the file will be reflected in the container by simply restarting it
-echo "JAVA_TOOL_OPTIONS='-Djava.util.logging.config.file=/app/logs/config/logging.properties'" >> .env
-
+if [ true = "$is_debug" ]; then
+  # The server will wait for the debugger to attach on port 5005
+  # JProfiler can attach on port 8849
+  echo "JAVA_TOOL_OPTIONS='-Djava.util.logging.config.file=/app/logs/config/logging.properties -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005' -agentpath:/path/to/libjprofilerti.so=port=8849 " >> .env
```
ata-nas, line 36: I think we need to remove the below `'` as they will result in having an environment variable `-agentpath:/path/to/libjprofilerti.so=port=8849` which is not correct (they conflict with the `"`s). Moreover, maybe we can completely remove this part `-agentpath:/path/to/libjprofilerti.so=port=8849` not really sure we need this.
```suggestion
  echo "JAVA_TOOL_OPTIONS=-Djava.util.logging.config.file=/app/logs/config/logging.properties -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005 -agentpath:/path/to/libjprofilerti.so=port=8849" >> .env
```

AlfredoG87: That part is used for Profiling, should only be on debug mode.

Thanks for noticing, somehow debugging remotely on the container is working, but probably the profiling is not, never tested that to be honest.

Have you tried this with your suggested changes?

ata-nas: I noticed the wrong env var when I made changes to the simulator java opts, that one was a copy paste. This will produce the same result. I know about the profiling stuff, I think it never worked (I mean the option, profiling did work if I remember correctly). This is why I think we can omit the profiling opt (could be wrong tho, need to double check).

ata-nas: P.S. debugging is always working because the debugging part is correct even with the `'`s.

ata-nas: But do try it out though.

ata-nas:
During my exploration of the Files Recent Plugin and doing tests for it, there are some points that came up which we need to take a look at. Some of them are already mentioned here, but I will nevertheless mention here again so we have everything in one place:

## Files Recent Plugin

1. Ranged Sets

   - as discussed in [this comment thread](https://github.com/hiero-ledger/hiero-block-node/pull/941#discussion_r2022779315), we saw the need of a RangedSet in order to be able to know the availability of a given Block Provider.
   - also there was a question about the immutability of the files written, mainly if we persist a given block file and it is verified, should/could it ever change

1. Files Recent `UnverifiedHandler` Rework

   - as discussed in [this comment thread](https://github.com/hiero-ledger/hiero-block-node/pull/941#discussion_r2022788320), we should potentially think about and rework the `UnverifiedHandler` to either not do any IO operations and only follow for the correctly flagged persistence and verification notifications, moving all IO operations to the plugin, or move all IO operations to the handler. This would make the class actually more testable since we will not need the `LongConsumer` which we currently cannot unit test from inside the handler.

1. Files Recent Block Accessor Delete Possible Race Conditions

   - would the current implementation of the `BlockFileBlockAccessor#delete()` introduce a possible race condition? The delete method will try to delete empty parents up to the root. Would that not introduce some possible race conditions?
     - imagine we receive block 1 and persist it in live. Then we receive block 2, but before we move it to live, and after we have passed the `Files.createDirectories` logic, a thread calls the `delete` method on block 1. If the parents up to root are empty they will be deleted. This would interfere with the persistence of block 2 since they would share the same parent.
     - roots must be of course deleted when no longer needed, with the previous implementation that came very naturally because the archiver picked a root (parent of group to archive), then zipped everything inside and then simply recursively deleted the picked parent which was a safe operation. There was no problem because we ensured that the data was now discoverable via the zip.
     - do we have other concerns about the deletion logic?

1. Files Recent Plugin Potentially Publishing an Unverified Block

   - what would happen if:
     1. we receive items to persist for block 1
     1. we persist block 1 to unverified
     1. verification does not pass
     1. then block 1 is requested again
     1. now verification passes before the items come in to be persisted (in this case to overwrite the unverified block 1 we persisted a moment ago)
     1. but since there is a block 1 in unverified root, it would be moved to live?
   - a tricky race condition we should be aware of, what would be the correct approach here? Does it tie into a rework of the unverified handler? Something else? Tighter synchronization?
   - should we handle verification failed notifications and cleanup existing block file for the failed verification block number? But also we need to be careful because we do not want to miss this notification, persistence may still be in progress or not yet started, in which case we need to know that the items that are coming in failed verification.
   - because items received and notifications are on different threads, we should be extra careful synching between them, although notifications will be received in order and items as well on their respective thread.
   - the plugin maybe should not be a decision maker, meaning it should probably remain simple and not decide wether to persist or not a block, as long as it received items, it will persist them, as long as it receives notifications it will act upon them and other decision making should probably be the responsibility of another component

---

Are there any other considerations we need to take into account?

ata-nas:
Some general questions that came up when testing the historic plugin, the first 2 kind of block my tests as it will require some rework in the production code, in some places we currently only handle Zstandard compressed data and no compression type tests fail

## Files Historic Plugin

1. Compression extensions (should they exist)

   - we have had some conversations if compression extensions should exist in the past:
     - it is better to check for magic number than to rely on an extension
     - extensions are human readable, but the data we are persisting maybe should not be accessed directly but rather be software managed and be queried via the usage of software rather than manually
     - the implementation of the app becomes quite more complex when we have extensions to think about:
       - for reading (this is valid for all types of plugins, not only historic but recent too), we cannot rely on runtime compression type setting to read a persisted block since the block could have been persisted with a different compression type which means that first of all, path resolution is more complex since we need to check for the file with multiple extension (iterate over a set of available/managed extensions that the app supports generally) and then we should read the data in the appropriate manner.

1. CompressionType (should it exist)

   - as of today, we have only one compression type, that is ZSTD
   - granted, we also have NONE (no compression)
   - the question is should we complicate the code by having the compression type as of today, could we not proceed only with ZSTD compression everywhere and introduce other types as requirements for them come
   - I kind of not see the point to have the no compression option, even for testing, we are still using IO streams, wrapping that into a ZstdIn/Out stream is no hustle
   - we will simplify the existing code
   - if we choose to remain only with ZSTD compression, should we remove the CompressionType enum or should we still have it for easier future introduction of other compression types and for now we simply remove the NONE option?
   - when reading, we will still have the option to have wrapped bytes as `ZSTD_PROTOBUF` or `PROTOBUF` (no compression)

1. Where are Blocks Coming From?

   - currently, we are using the historical block facility to get accessors to blocks we need to archive, but at least I do not really see (know) where the blocks are coming from. I understand this is by design, but I was thinking "Are there wrong places to get blocks to archive them from?". One such place would certainly be the Files Historic Plugin itself (the one that is zipping, to be taking from itself, another would be cloud I believe)
   - should we have some ability to say, only accept blocks if they are coming from a provider with lower/higher priority than x or to target them in some other way?
   - we are somewhat protected by the notifications because when we see a `BLOCK_PERSISTED` notification, we know that this means persisted locally, but we have no assurance really if blocks for the given group we need to archive have not potentially moved by other plugins or circumstances
   - I guess the main question would be is what are the real concerns here regarding this question (if any) and how deeply must they be handled
   - I understand that no plugin should take ownership of any given block, as in reality a block is not the property of any given plugin

1. What Should Happen if the Block Node Restarts Somewhere in the Middle of a Group Being Persisted?

   - I imagine the following scenario:
     - say we group each 1000 blocks
     - files recent persists the first 500 blocks and then the node restarts
     - should the block node then scan to determine latest archived and then follow notifications
     - or should it rely on historical block provider to determine largest stored (if so do we not again hit the above problem "where are blocks coming from"?)
   - what should be the general desired behavior of this plugin at startup?
   - I understand that no plugin should take ownership of any given block, as in reality a block is not the property of any given plugin

---

Are there any other considerations we need to take into account?

cc: @jasperpotts @jsync-swirlds @AlfredoG87 @georgi-l95 @Nana-EC

ata-nas:
# Team Discussion 10-Apr-2025

- We discussed the above questions both for the [`Files Recent`](https://github.com/hiero-ledger/hiero-block-node/pull/890#issuecomment-2783098567) and [`Files Historic`](https://github.com/hiero-ledger/hiero-block-node/pull/890#issuecomment-2789269073) Plugins
- We also have a reply from @jsync-swirlds [here](https://github.com/hiero-ledger/hiero-block-node/pull/890#issuecomment-2784051050)
- The questions sparked quite a lot of discussions:
  - We categorized them in 2 main topics:
    1. Unverified Blocks and what to do with them
    1. "Moving" Blocks between Plugins (Where are Blocks Coming From?)
- There are 2 major takeaways from the discussion
  1. We will rework the current implementation (and API) of the Files Recent Plugin which will solve a lot of race conditions, mainly we will only get the Block after verification on it passes
  1. We still need to think thoroughly about the Files Historic Plugin

Here are some captures from our discussion (checklist order matches the questions in the above links):

## Files Recent Plugin

- [x] Q1: Ranged Sets are pretty much addressed and will be a thing.
- [x] Q2: The Unverified Handler will be removed as there will be a rework in the Files Recent Plugin.
- [ ] Q3: The delete will probably become obsolete as there will probably be changes to the philosophy of how deletes happen, mainly deletion would probably be under the discretion of the respective plugin, reactive.
- [x] Q4: The race conditions will no longer be a problem because we decided that we will move to an implementation where the Files Recent Plugin will only be listening to a successful verification notification which will now also be having a reference to the Block itself. This simplifies a lot the whole dataflow and solves a ton of issues. A note here, we will probably also have to be careful with the ring buffer as it will not release blocks until overwritten, so we will hold refs we will no longer need. Probably a different data structure will be utilized.

## Files Historic Plugin

- [ ] Q1: Team is 50/50 on this one, we will proceed with what we have now and address this in the future.
- [ ] Q2: Team is 50/50 on this one, we will proceed with what we have now and address this in the future, possibly we will simplify by removing the NONE compression?
- [ ] Q3: Something that we will need to think more thoroughly about.
- [ ] Q4: Something that we will need to think more thoroughly about.
- [ ] Q5: Something that we will need to think more thoroughly about.
- [ ] Q6: Something that we will need to think more thoroughly about.
- [ ] Q7: Something that we will need to think more thoroughly about.

---

Please edit this as team sees fit. These were my major captures. Are we missing something?

cc: @jasperpotts @jsync-swirlds @AlfredoG87 @georgi-l95 @Nana-EC

Nana-EC (approved):
LG.

Solid effort from the team and this was a big change.
The new foundation gives us an approach to further expand and make it easy for alternative considerations.

More tests and improvements to come in the near team sprints.

settings.gradle.kts
```diff
@@ -14,15 +14,17 @@ javaModules {
...
+        module("health") { artifact = "block-node-health" }
         module("messaging") { artifact = "facility-messaging" }
-        // module("configuration") { artifact = "facility-configuration" }
-        // module("registrar") { artifact = "facility-registrar" }
-        // module("facilities-api") { artifact = "facilities-api" }
+        module("publisher") { artifact = "block-node-publisher" }
+        module("subscriber") { artifact = "block-node-subscriber" }
+        module("verification") { artifact = "block-node-verification" }
+        module("block-providers/cloud.archive") { artifact = "block-node-blocks-cloud-archive" }
+        module("block-providers/cloud.historic") { artifact = "block-node-blocks-cloud-historic" }
```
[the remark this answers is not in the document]
jsync-swirlds: We need to work on some deployment options.
Including this here is necessary to ensure CI tests cover everything we build, though. What we will do for actual deployment is create alternative distributions; which is a future task.

AlfredoG87 (approved):
Being contributing and working on this long lived branch for what it feels like a month, in that time I've grown confident on this overhaul of the BN architecture, the splitting and re-doing of the different modules of the system.

Also we have attained quite a lot of e2e and UTs coverage, we have added new features, and althought I recognize that improvements and optimizations will be made, and bugs could (and most likely) will be found. I approve as a vote of confidence in this new architecture and optimistic about our path forward with this new `Nano Services approach architecture` 

Congrats team!

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
```diff
@@ -0,0 +1,292 @@
...
+                    LOGGER.log(WARNING, "Failed to list files in directory: " + parentDir, e);
+                }
+                // we did not find any files in the directory, so delete it
+                Files.deleteIfExists(parentDir);
+                // move up to the parent directory
+                parentDir = parentDir.getParent();
+            }
+        } catch (IOException e) {
+            LOGGER.log(WARNING, "Failed to delete block file: " + blockFilePath, e);
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
jsync-swirlds: This is already replaced, but I agree with the idea to avoid top-level or generic exception types in general.

ata-nas (approved):
Great job! Looking forward to moving to main again! As my main focus was the persistence plugins, I have opened a bucket to capture future improvements #1022. Looking forward to pushing more improvements/fixes/tests and to further increase confidence in the new approach. Team did an awesome job with this one!
🔥

georgi-l95 (approved):
👁️ 👍 👉 👨‍💻

PR #900, opened 2025-03-25
feat: Simple Simulator Startup Data

### LINKED ISSUE #868: Simulator Data Persistent Volume
### Story Form

As a BlockNode Developer
I want to have a persistent volume for simulator data and to be able to deploy it long term into K8s
So that it can be used to store various data persistently

### Technical Notes

- we need a persistent volume for the simulator
- we need a persistent volume claim for the simulator
- we need to utilize it with a k8s deployment

### LINKED ISSUE #899: Simple Simulator Startup Data
### Story Form

As a Block Node Developer
I want to store simple simulator startup data
So that for now the simulator will be able to start generating and streaming blocks from block with number one higher than the latest ackd block

### Technical Notes

- we need to load simple startup data that will give us:
  - latest block number ackd
  - latest block hash
- this PR will only be loading and using the startup data if the `CRAFT` generator mode is used

ata-nas:
## Reviewer Notes

- add simple startup data that will give us:
  - the latest ackd block number
  - the latest ackd block hash
- for this PR, the startup data will only be utilized by the `CRAFT` block generation mode

## Related Issue(s)

Closes #899 
Closes #868


simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -78,6 +81,30 @@ public CraftBlockStreamManager(@NonNull BlockGeneratorConfig blockGeneratorConfi
...
+        if (Files.exists(latestAckBlockHashPath)) {
+            try {
+                final byte[] fromFile = Files.readAllBytes(latestAckBlockHashPath);
+                if (fromFile.length == StreamingTreeHasher.HASH_LENGTH) {
+                    previousBlockHash = fromFile;
+                }
+            } catch (final IOException e) {
+                LOGGER.log(DEBUG, "Error reading latestAckBlockHash file.");
+            }
+        }
```
ata-nas, line 107: @georgi-l95 is this correct? The idea is that we save the latest ackd block number + hash, when the simulator restarts, we check for the number and assign the current number to the queried number +1. Then for the hash, if we find any, we assign it as the previousBlockHash. Is the state we initialize with correct? The next created block needs to be able to be verified and to be correctly used the previous block hash.

simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamObserver.java
```diff
@@ -50,8 +76,26 @@ public void onNext(PublishStreamResponse publishStreamResponse) {
...
             lastKnownStatuses.pollFirst();
         }
+        final BlockAcknowledgement ack =
+                publishStreamResponse.getAcknowledgement().getBlockAck();
+        final PublishStreamResponseCode responseCode =
+                publishStreamResponse.getStatus().getStatus();
+        // todo we need the correct response code, currently it seems that
+        //   the response code is not being set correctly? The if check should
+        //   be different and based on the response code, only saving
+        if (PublishStreamResponseCode.STREAM_ITEMS_UNKNOWN == responseCode && !ack.getBlockAlreadyExists()) {
```
ata-nas, line 86: @AlfredoG87 please note the todo, currently it seems that the standard response code we get from a successful ack is the items unknown. Let's speak about this to resolve it properly. This if check is generally wrong, I need to ensure we got this right.

simulator/docker/docker-compose-publisher-consumer.yml
```diff
@@ -17,6 +17,8 @@ services:
       interval: 3s
       timeout: 10s
       retries: 5
+    volumes:
+      - simulator_data:/opt/simulator/data
```
ata-nas, line 21: @AlfredoG87 This volume is probably not needed. When developing, it is generally good to have (possibly), but that is debatable. As we spoke, we may introduce a simple flag that will either ignore or not the startup data files. I am not sure if we should add that as of now. Issue is that if you start the simulator within a container, everything is fine, but if not, then it is a bit of a hassle because of access denied on default options. 

On another note, we need to add the proper pvc for k8s.

simulator/docker/update-env.sh
```diff
@@ -17,11 +17,19 @@ PUBLISHER_PROMETHEUS_ENDPOINT_PORT_NUMBER=9998
...
+debug_arg="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5006"
+# determine if we should include debug opts
+[ "$1" = true ] && is_debug=true || is_debug=false
+if [ true = "$is_debug" ]; then
+  # The server will wait for the debugger to attach on port 5005
+  echo "JAVA_TOOL_OPTIONS=${logging_config_file_arg} ${debug_arg}" >> .env
+else
+  echo "JAVA_TOOL_OPTIONS=${logging_config_file_arg}" >> .env
+fi
+
```
ata-nas, line 32: Added debugging opts for the simulator when starting it with docker.

ata-nas (commented):
Still at draft, need to point out a couple of things, as we need this PR.

simulator/src/main/java/org/hiero/block/simulator/config/data/BlockStreamConfig.java
```diff
@@ -26,7 +29,11 @@ public record BlockStreamConfig(
         @Loggable @ConfigProperty(defaultValue = "100_000") int maxBlockItemsToStream,
         @Loggable @ConfigProperty(defaultValue = "MILLIS_PER_BLOCK") StreamingMode streamingMode,
         @Loggable @ConfigProperty(defaultValue = "1000") int millisecondsPerBlock,
-        @Loggable @ConfigProperty(defaultValue = "1000") int blockItemsBatchSize) {
+        @Loggable @ConfigProperty(defaultValue = "1000") int blockItemsBatchSize,
+        @Loggable @ConfigProperty(defaultValue = "/opt/simulator/data/latestAckBlockNumber")
```
georgi-l95, line 33: We should have a toggle, whether we want to start on a clean slate or continue from previous run.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -78,6 +81,30 @@ public CraftBlockStreamManager(@NonNull BlockGeneratorConfig blockGeneratorConfi
         this.outputTreeHasher = new NaiveStreamingTreeHasher();
 
         LOGGER.log(INFO, "Block Stream Simulator will use Craft mode for block management.");
+
+        final Path rootDataPath = Path.of("/opt/simulator/data");
```
georgi-l95, line 85: All this logic should be extracted in a method IMO.

simulator/src/main/java/org/hiero/block/simulator/config/SimulatorMappedConfigSourceInitializer.java
```diff
@@ -24,6 +24,8 @@ public final class SimulatorMappedConfigSourceInitializer {
             new ConfigMapping("blockStream.streamingMode", "BLOCK_STREAM_STREAMING_MODE"),
             new ConfigMapping("blockStream.millisecondsPerBlock", "BLOCK_STREAM_MILLISECONDS_PER_BLOCK"),
             new ConfigMapping("blockStream.blockItemsBatchSize", "BLOCK_STREAM_BLOCK_ITEMS_BATCH_SIZE"),
+            new ConfigMapping("blockStream.latestAckBlockNumberPath", "BLOCK_STREAM_LATEST_ACK_BLOCK_NUMBER_PATH"),
+            new ConfigMapping("blockStream.latestAckBlockHashPath", "BLOCK_STREAM_LATEST_ACK_BLOCK_HASH_PATH"),
```
georgi-l95, line 28: Can we just use the same path for both, therefore use only one variable? Don't see a reason for them to be in different folders from each other.

georgi-l95 (commented):
Overall looks alright.

simulator/src/main/java/org/hiero/block/simulator/config/SimulatorMappedConfigSourceInitializer.java
ata-nas: It is not easy to differentiate the number from the hash if we put everything in one file. This is an easier and safer approach. I did consider it, the option is to read the length of a long and then the rest is hash, but what if some error happens and we have only the number or hash in the single file? Also the writing logic will be a bit more of a hassle. I see no reason to do it in one file with much increased complexity.

simulator/src/main/java/org/hiero/block/simulator/config/data/BlockStreamConfig.java
ata-nas: I kind of agree with you, however I have some reservations. See[ this comment](https://github.com/hiero-ledger/hiero-block-node/pull/900#discussion_r2012168197). The main thing is that a toggle _might_ make development/testing a bit easier. For the purposes that we need the startup data for, it may be redundant. I have discussed this briefly with @AlfredoG87, but we should indeed come to a conclusion.

ata-nas: This will be added, it will aid testing and development. As an end result, this is something we will need.

simulator/docker/docker-compose-publisher-consumer.yml
AlfredoG87: I think we already discussed this on the call right?

but my take:
If we want to support and test the restart recovery, we should create a volume, otherwise we will always lose the data.

I believe the flag will help with e2e test workflows, but we can also include a volume clean-up step on the `dockerStop` gradle task, so it cleans the container and volumes.

simulator/docker/docker-compose-publisher-consumer.yml
```diff
@@ -39,3 +41,6 @@ networks:
   block-node_default:
     name: block-node_default
     external: true
+
+volumes:
+  simulator_data:
```
AlfredoG87, line 46: missing blank line at EOF

simulator/docker/docker-compose-publisher.yml
```diff
@@ -0,0 +1,31 @@
...
+    volumes:
+      - simulator_data:/opt/simulator/data
+
+networks:
+  block-node_default:
+    name: block-node_default
+    external: true
+
+volumes:
+  simulator_data:
```
AlfredoG87, line 31: missing blank line at EOF

simulator/docker/update-env.sh
AlfredoG87: nice!! we need to do this on the `server` as well, maybe on another PR?
or maybe since you already got it here we can sneak it in, your call! :)

simulator/src/main/java/org/hiero/block/simulator/config/SimulatorMappedConfigSourceInitializer.java
AlfredoG87: I think we can have a base path for the folder (that is the one we map to the PVC or persistent volume, and another Constant for the file-name.

so we are on the same folder, but on different files. win win :)

simulator/src/main/java/org/hiero/block/simulator/config/data/BlockStreamConfig.java
AlfredoG87: @georgi-l95 you read our mind, we were unsure about it but after some testing it became aparent that indeed is something helpful.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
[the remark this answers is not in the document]
AlfredoG87: agreed

AlfredoG87 (commented):
Looking good, I will circle back once is "Ready" for review 👍

simulator/src/main/java/org/hiero/block/simulator/config/data/BlockStreamConfig.java
ata-nas: ✅ Done.

simulator/src/main/java/org/hiero/block/simulator/config/SimulatorMappedConfigSourceInitializer.java
ata-nas: Let's proceed as it is, I do agree we should probably have a base simulator config in the future.

simulator/docker/docker-compose-publisher-consumer.yml
ata-nas: ✅ Done.

simulator/docker/docker-compose-publisher.yml
ata-nas: ✅ Done.

charts/blockstream-simulator/templates/deployment.yaml
```diff
@@ -28,6 +28,23 @@ spec:
...
+            claimName: {{ if .Values.simulator.persistence.data.create }}{{ include "blockstream-simulator-chart.fullname" . }}-data{{ else }}{{ .Values.simulator.persistence.data.existingClaim }}{{ end }}
+      initContainers:
+        - name: init-storage-dirs
+          image: busybox
+          command:
+            - sh
+            - -c
+            - |
+              mkdir -p /data-pvc/{{ .Values.simulator.persistence.data.subPath }} && \
+              chown 2000:2000 /data-pvc/{{ .Values.simulator.persistence.data.subPath }} && \
```
jsync-swirlds, line 43: Should this be `$UID:$GID` instead of `2000:2000`?

simulator/docker/docker-compose-publisher.yml
```diff
@@ -0,0 +1,31 @@
...
+  simulator-publisher:
+    container_name: simulator-publisher
+    build:
+      context: .
+      dockerfile: Dockerfile
+    image: hedera-block-simulator:latest
+    networks:
+      - block-node_default
+    ports:
+      - "5006:5006"
```
jsync-swirlds, line 11: These should probably be environment variables or similar, or a random value if we don't need to expose a callable API.

jsync-swirlds (dismissed):
Looks good, just minor nitpicks that can be resolved in a future PR.

charts/blockstream-simulator/templates/deployment.yaml
AlfredoG87: No, here we need to match the same UID and GID that the `container imagee` is using (the one is set on the Dockerfile) and that will be the one that will be running the app on the main container.

AlfredoG87 (dismissed):
LGTM

simulator/docker/docker-compose-publisher.yml
AlfredoG87: good question, what is the simulator exposing?

charts/blockstream-simulator/templates/deployment.yaml
jsync-swirlds: Interesting.  Is it possible to make these variables/parameters, or are we restricted to constants?

I'm just thinking of a situation where the specific number used is already in use for another purpose.

AlfredoG87: I guess the question is why would we want them as variables, if it is to keep a single centralized place and reference it, not sure if is possible since this is in the helm charts and the other place we need it is on the dockerfile that is used to build the container image. *cant see a way for those 2 too share a parameter.

jsync-swirlds: I see.
Thanks for explaining the situation.

simulator/docker/docker-compose-publisher.yml
ata-nas: This is for the debugger of the simulator.

ata-nas:
@jsync-swirlds @AlfredoG87 I found a couple of wrongful documentation comments and fixed those so it is not misleading. I have  also made minor improvements to the tests to cover a bit more cases and to properly initialize the object under test (as it does produce side effects, mainly it creates files on the filesystem, which reminds me that we should at some point do take a look at `jimfs` for tests and start using it as proposed previously by @jsync-swirlds)

jsync-swirlds (approved):

PR #905, opened 2025-03-25
chore: Adjust workflows to reduce "stuck" jobs via reduced priority and time limits.

### LINKED ISSUE #903: Address CI timeouts by limiting runtime and reducing process priority.
### Story Form

As a Block Node Engineer
I want to have a reliable and time-limited continuous integration suite
So that I do not have to wait overly long for results and have a minimum rate of false errors.

### Technical Notes

The github agent does not play well with long (several minute) CPU intensive build or test processes. Our current build has been starving the agent of CPU and taking up to 6 hours to time out after that point.
To address this we propose two changes.
1. Run CPU intensive processes with `ionice` to reduce process priority.
2. Set much stricter time limits (60 minutes for most jobs, less for "fast" jobs) so any that do time out do so more quickly.

jsync-swirlds:
## Reviewer Notes
* Added time limits to most jobs
* Added ionice to most processor or I/O heavy steps.

## Related Issue(s)
Resolves #903

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -19,10 +19,11 @@ permissions:
   contents: read
 
 env:
-  GRADLE_EXEC: ./gradlew
+  GRADLE_EXEC: "ionice -c 2 -n 2 nice -n 19 ./gradlew "
 
 jobs:
   e2e-tests:
+    timeout-minutes: 30
     runs-on: hiero-block-node-linux-medium
     steps:
       - name: Harden Runner
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -16,13 +16,14 @@ defaults:
     shell: bash
 
 env:
-  GRADLE_EXEC: ./gradlew
+  GRADLE_EXEC: "ionice -c 2 -n 2 nice -n 19 ./gradlew "
 
 permissions:
   contents: read
 
 jobs:
   smoke-test:
+    timeout-minutes: 30
     name: "Smoke Tests"
     runs-on: hiero-block-node-linux-medium
     steps:
--- a/.github/workflows/helm-charts.yaml
+++ b/.github/workflows/helm-charts.yaml
@@ -6,8 +6,12 @@ on: pull_request
 permissions:
   contents: read
 
+env:
+  CGEXEC: "ionice -c 2 -n 2 nice -n 19 "
+
 jobs:
   lint-install-test:
+    timeout-minutes: 20
     name: Lint and Install Charts
     runs-on: hiero-block-node-linux-medium
     steps:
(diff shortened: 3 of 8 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -19,10 +19,11 @@ permissions:
   contents: read
 
 env:
-  GRADLE_EXEC: ./gradlew
+  GRADLE_EXEC: "ionice -c 2 -n 2 nice -n 19 ./gradlew "
```
Nana-EC, line 22: Would be nice to capture the intent and choice of values somewhere for those unfamiliar with the [ionice](https://man7.org/linux/man-pages/man1/ionice.1.html) and [nice](https://man7.org/linux/man-pages/man1/nice.1.html) commands.
Especially if modifying is ever needed

Nana-EC (commented):
LG, might even suggest a few minutes smaller for some of the jobs.
I'm assuming current values are based off of observed behaviour with some buffer so that's fine.

Can we see about capturing the knowledge for the `ionice` and `nice` commands given developers will be of varies backgrounds and linux experience

.github/workflows/smoke-test.yaml
```diff
@@ -16,13 +16,14 @@ defaults:
...
 env:
-  GRADLE_EXEC: ./gradlew
+  GRADLE_EXEC: "ionice -c 2 -n 2 nice -n 19 ./gradlew "
 
 permissions:
   contents: read
 
 jobs:
   smoke-test:
+    timeout-minutes: 60
```
AlfredoG87, line 26: This could be `20 mins` and still be inside the 3-5x the historical average.

.github/workflows/e2e-tests.yaml
```diff
@@ -19,10 +19,11 @@ permissions:
...
 
 jobs:
   e2e-tests:
+    timeout-minutes: 60
```
AlfredoG87, line 26: this we can also lower it to `30 min` for 3x historical average

AlfredoG87 (approved):
LGTM, just 2 nits on times we migh get away lowering them a tad.
So when it gets stuck we can realize it faster 👍

.github/workflows/smoke-test.yaml
jsync-swirlds: I set to 30 to leave a little extra room for growth here.

.github/workflows/e2e-tests.yaml
jsync-swirlds: Set to 30.

AlfredoG87 (approved):

.github/workflows/e2e-tests.yaml
Nana-EC: I know, just trying to save other developers time and make it easy to modify should we need to

jsync-swirlds: I'll add the documentation.  The reasoning for after your reorg PR is to avoid creating unnecessary conflicts.

Nana-EC: That's fine. I wasn't sure if the current files allowed for comments.
I didn't mean a separate so maybe this is a no-op.
Was a nice to have request anyway

Nana-EC (approved):

rbarker-dev (approved):

PR #908, opened 2025-03-26
initial test plan for verification

### LINKED ISSUE #907: Verification Test Plan
### Story Form

As a BlockNode Developer
I want to make sure that Block Verification feature continues to work after any change is done to the codebase via a PR

### Technical Notes

So that there is a test suite that includes all the test scenarios to validate that verification is working as expected, and that this test suite is run as part of a PR CI Checks.

AlfredoG87:
## Reviewer Notes

Fixes #907


docs/test-plan/verification-tp.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Key Considerations
+
+- **Implementation Variants:** Two implementations exist, SYNC for testing and ASYNC for production.
+
+- **ASYNC vs. SYNC:** The ASYNC implementation uses a thread pool with a `concurrent` TreeHasher for calculating the block root hash and is approximately 50% faster than the SYNC version, which uses a `naive` TreeHash.
+
+- **Naive Tree Hasher:** calculates the root hash of the merkle trees in a synchronous manner and using a simple recursive algorithm.
+
+- **Concurrent Tree Hasher:** calculates the root hash of the merkle trees in an asynchronous manner using a thread pool to parallelize the calculation as the blocks items are received and leafs are appended to the "merkle tree".
```
Nana-EC, line 15: Q: Is it important to note the logic of parallelization here in case it informs testing approach or is that an implementation not necessary?

docs/test-plan/verification-tp.md
```diff
@@ -0,0 +1,31 @@
...
+
+- **Verification Criteria:** A block is verified if the calculated block root hash matches the signature provided in the block_proof.
+
+- **Acknowledgement Conditions:** The BlockNode sends a BlockAcknowledgement only if the block has been both verified and persisted successfully.
+
+- **Test Coverage:** End-to-end test scenarios should include both positive and negative cases for both the ASYNC and SYNC implementations.
+
+## Test Scenarios
+
+| Test Case ID |                         Test Name                          |                                                      Scenario Description                                                       |                                                                                                                                                               Expected Behaviour                                                                                                                                                               |                   Input                   |                                              Output                                               | Implemented (Y/N) |
```
Nana-EC, line 25: Note `Test Case ID` should be a unique string of sort, as a we want to note it in the code so that an E2E test can be mapped back to the plan

Nana-EC (commented):
Looking good, I think you noted one form might be going so the test plan might simplify even more correct?

PR #923, opened 2025-03-29
When streaming a block from the simulator, sending the block_header a…

### LINKED ISSUE #922: BlockHeader and BlockProof - Should come alone or as part of a batch?
### Story Form

As a BlockNode Developer
I want to E2E test the new BN Messaging system
So that we can reach the verification and persistence plugins

### Technical Notes

Seems to me by looking at the code and tests that `block_header` and `block_proof` items should go by their own, just them in the batch.

Currently the simulator and the old version of the BN is not expecting that. so the logic does not seem to work properly.

The intention of this ticket is to first determine what is the right approach and second to fix it either on the BN app or on the simulator side.

If this will be a new rule, then we also need to communicate it to the CN team, since I don't think their work contemplates this.

The only rules we used to have is that block_header always is going to be the first item on a batch, and that block_proof always would be the last item in a batch, and was perfectly normal for a single Batch of items to contain both block_heder, items and block_proof on the same batch. 

@jsync-swirlds @jasperpotts might need some help clarifying this new rule before deciding on path forward to unblock e2e test progress

AlfredoG87:
## Reviewer Notes
Modifies the Simulator to stream `block_header` and `block_proof` in its own batches.

**This PR won't be merged, we will need to support block_header and block_proof to be part of a batch with more items, for now, this is a temp workaround to unblock some parts of the e2e testing.**

## Related Issue(s)
Fixes #922


Nana-EC (commented):
Looking good.
Based on #922 does that mean you determined this should be the way and we need to have CN adopt a similar approach?

AlfredoG87:
> Looking good. Based on #922 does that mean you determined this should be the way and we need to have CN adopt a similar approach?

I remember that this was mentioned, but haven't got confirmation, however looking at the tests and code, seems very intentional, so I don't think is a regression, but a "new" rule or change to the protocol.

but since I suspect this to be the case, I went forward and fix it  (optimistically) so we could continue with the e2e and find the next issue in the flow 😅

jasperpotts:
Fixed issue in main code so do not need this workaround any more. Thanks @AlfredoG87

PR #941, opened 2025-04-01
test: Files Recent Plugin Unit Tests

### LINKED ISSUE #942: Files Recent Plugin Unit Tests
### Story Form

As a Block Node Developer
I want to have unit tests for the `Files Recent` Plugin
So that I have confidence in the existing logic and catch regressions

### Technical Notes

- write unit tests for the `Files Recent` Plugin

ata-nas:
## Reviewer Notes

- unit tests for `Files Recent` plugin
- some cleanup && isolation improvements

## Related Issue(s)

Closes #942

### DIFF
--- a/block-node/block-providers/files.recent/build.gradle.kts
+++ b/block-node/block-providers/files.recent/build.gradle.kts
@@ -14,4 +14,11 @@ mainModuleInfo {
     runtimeOnly("com.hedera.pbj.grpc.helidon.config")
 }
 
-testModuleInfo {}
+testModuleInfo {
+    requires("com.google.common.jimfs")
+    requires("org.junit.jupiter.api")
+    requires("org.junit.jupiter.params")
+    requires("org.mockito")
+    requires("org.mockito.junit.jupiter")
+    requires("org.assertj.core")
+}
--- a/block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/UnverifiedHandler.java
+++ b/block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/UnverifiedHandler.java
@@ -7,6 +7,7 @@ import com.hedera.pbj.runtime.io.stream.WritableStreamingData;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.BufferedOutputStream;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.lang.System.Logger.Level;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -103,7 +104,7 @@ final class UnverifiedHandler {
                     "Failed to create unverified block directory: {0}, error: {1}",
                     unverifiedBlockPath.getParent().toAbsolutePath().toString(),
                     e.getMessage());
-            throw new RuntimeException(e);
+            throw new UncheckedIOException(e);
         }
         // write the block to the file
         try (final WritableStreamingData streamingData = new WritableStreamingData(new BufferedOutputStream(
@@ -127,7 +128,7 @@ final class UnverifiedHandler {
                     blockNumber,
                     unverifiedBlockPath.toAbsolutePath().toString(),
                     e.getMessage());
-            throw new RuntimeException(e);
+            throw new UncheckedIOException(e);
         }
     }
 }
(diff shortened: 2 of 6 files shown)


block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/UnverifiedHandler.java
```diff
@@ -81,8 +84,9 @@ public synchronized void blockVerified(long blockNumber) {
      * @param blockNumber the block number of the block to store
      * @return true if the block was stored, false if it was already verified
      */
-    public synchronized boolean storeIfUnverifiedBlock(
-            final List<BlockItemUnparsed> blockItems, final long blockNumber) {
+    synchronized boolean storeIfUnverifiedBlock(
+            @NonNull final List<BlockItemUnparsed> blockItems, final long blockNumber) {
+        Objects.requireNonNull(blockItems);
         // check if the block number is already verified
         if (last100VerifiedBlockNumbers.contains(blockNumber)) {
```
ata-nas, line 91: The check on line 91 is good, it solves the problem locally and for a brief period of time, but I believe we must have another check in general, that would prevent us from creating (writing) a new block in the unverified handler (or maybe the proper place would be the files recent plugin) if any other block provider has the block. 

The point is to ensure immutability, since once written and verified, a block must not be overwritten. In the previous implementation we did have this check, existsBlock(long blockNumber), but with the new implementation I guess we must iterate over all available providers and check for existence of the block, or we must have some other mechanism to do so, asserting that our persistent storage (no matter where it is) contains or not the block with the block number we are now trying to write.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/UnverifiedHandler.java
```diff
@@ -40,9 +43,9 @@ public class UnverifiedHandler {
      * @param config the configuration for this plugin
      * @param moveToLive function to call to move the block to live storage
      */
-    public UnverifiedHandler(FilesRecentConfig config, LongConsumer moveToLive) {
-        this.config = config;
-        this.moveToLive = moveToLive;
+    UnverifiedHandler(@NonNull final FilesRecentConfig config, @NonNull final LongConsumer moveToLive) {
+        this.config = Objects.requireNonNull(config);
+        this.moveToLive = Objects.requireNonNull(moveToLive);
```
ata-nas, line 48: Do we absolutely need the long consumer/callback here? I do see that it uses some internal state of the files recent plugin. The problem with the consumer is that it cannot really be asserted the logic of for unit tests for this class, as in the best case scenario the consumer must be copied (bad idea IMO).

I was thinking, could it not be made that this handler will not do any IO operations, but simply managing pings for persisted and verified blocks, returning the next in line block to be moved to live and then doing the actual move inside the files recent plugin? In this way the IO logic will remain only in the files recent plugin. I am not sure as of yet what would be the best idea here.

For now I will mock it in unit tests for this class, IO operations for the callback will be asserted from the files recent plugin tests.

block-node/health/src/test/java/org/hiero/block/node/health/HealthServiceTest.java
ata-nas, line 1: Changes in this file are because of spotless

block-node/block-providers/files.recent/build.gradle.kts
```diff
@@ -14,4 +14,11 @@ mainModuleInfo {
...
 
-testModuleInfo {}
+testModuleInfo {
+    requires("com.google.common.jimfs")
+    requires("org.junit.jupiter.api")
+    requires("org.junit.jupiter.params")
+    requires("org.mockito")
+    requires("org.mockito.junit.jupiter")
+    requires("org.assertj.core")
+}
```
ata-nas, line 24: Including test dependencies

ata-nas (commented):
Just some comments to take a look at.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/UnverifiedHandler.java
jasperpotts: Well worth thinking about. My aim yesterday was to get things working so we can move forward. We can refactor as needed to get the separation of concerns right and make it testable.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/UnverifiedHandler.java
jasperpotts: Good question. This seems like something we should think about discuss. I have been torn how we keep track of blocks stored. As we have ranges min/max at the moment but that does not account for missing blocks. I am not sure how we efficiently account for what blocks we have and are missing. I am thinking we should maybe implement a RangeSet type data structure and each block provider can express what it has and what it is missing better.  Also not sure it is right to assume that we should only have a block in a single provider. For example the CloudArchive provider should have a copy of all blocks. But we might want to keep local copies to answer queries for as much as we have available disk space. Also need to think do we allow incomplete ranges of blocks to move into providers that need a whole range like for a zip.

jasperpotts (approved):

Issue #427, opened 2024-12-19
Separate Prometheus and Grafana configs


georgi-l95:
As a block node developer
I want to have clear separation between simulator and server metrics configuration
To be able to easily modify and improve without interference between subprojects

HarshSawarkar:
Hi Team ,I have raised a PR for this issue. Could you please review it?

HarshSawarkar:
Also I tried running `./gradlew startDockerContaine`r on a **Windows** system, but it failed with the error below.

```
> Task :simulator:createDockerImage FAILED                                                                                                                                          
> Task :block-node-server:createDockerImage FAILED

FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':simulator:createDockerImage'.
> A problem occurred starting process 'command 'sh''
```

 I explored a couple of workarounds, like modifying the script and using a _Linux container with Docker-in-Docker support_ as the build environment, but the issue persists. Any guidance on resolving this would be greatly appreciated. Thanks!

jsync-swirlds:
It appears the base OCI build environment is missing the `sh` binary.  Some Linux distributions do not link `sh` to the local default shell (or default to csh or other shell options incompatible with the original "bourn" shell).
When preparing the environment where the build process will run, it is necessary to ensure that the "standard" shell is present and linked appropriately.

Note, we do not test or support a Windows build environment _at this time_.  We support Linux primarily (the distributions _tested_ for build are in the CI workflow [here](https://github.com/hiero-ledger/hiero-block-node/blob/ebc899b87b155d529f503db9379bc8fa71bea8a5/.github/workflows/zxc-verify-gradle-build-determinism.yaml#L148)) for the build environment. Other environments (e.g. Windows, Mac, BSD, etc...) may work, but are not necessarily something the maintainers are able to support.

PR #949, opened 2025-04-02
chore: seperated the prometheus and grafana configs for simulator and server

HarshSawarkar:
## Reviewer Notes

- Moved the configs of prometheus and grafana related to simulator in its own module
- Updated the docker files to adapt the changes.

## Related Issue(s)
#427 


AlfredoG87 (changes requested):
Hi @HarshSawarkar,

Thank you for your cotribution, am sorry I was not able to review this before but we were in the middle of changing pretty much everything, so we frozen non-essential merges to `main` for about a month.

I suggest you apply your changes from new version of `main` if you still want to finish this PR. and try to avoid `merge`, always use `rebase`.

Now, I believe the **Reviewer Notes** are not depicting the detailed changes and the intention behind them. And I am not sure this PR has capture correctly the intion of the Issue #427 that it tries to address.

Maybe @georgi-l95 can elaborate more on the intention, goal and how the end result would look like for this issue.

When I try to run both containers `Server` and `Simulator` using gradle tasks: `:block-node-server:startDockerContainer` and `:simulator:startDockerContainer` I get the following error when trying to run the lattter:

```
Error response from daemon: Conflict. The container name "/prometheus" is already in use by container "204ccc93084d0cd7659cb36891f6cd4d7f8a8490f67ef260a017cbee6c016ffc". You have to remove (or rename) that container to be able to reuse that name.

```

Please take into consideration that imo, the objective is to divide between `server` and `simulator` prometheus and grafana server.

Not necessarily between each type of simulator: 1. PublisherClient, 2. PublisherServer and 3. Consumer.

In any case, I should be able start the docker compose stack for the server and for both publisherClient and consumer at the same time and everything most work.

In  my opinion it would be necessary to have a gradle tasks that can start in each of those modes, and that they all are using the same docker-compose stack with just some changes in `.ENV` file. and if starting 2 types of simulators at the same time, they should have their own instance of prom+grafana, running on different ports with different dashboards. but I would prefer @georgi-l95 to elaborate further on the desired outcome since he created the original issue.

Nana-EC:
Hey @HarshSawarkar, feel free to rebase and update if you're able to now that the refactor to `main` has completed..
A lot has changed so it's possible the issue is no longer applicable or a larger or smaller effort than before.
Take your time and let us know what you find.

PR #955, opened 2025-04-03
fix: e2e improvements

### LINKED ISSUE #343: refactor: suites must not use services build files directly
### Problem

Based on [this comment thread](https://github.com/hashgraph/hedera-block-node/pull/337#discussion_r1838506405) we must not use `:service` related build files directly in `:suites`

### Solution

- see [this comment thread](https://github.com/hashgraph/hedera-block-node/pull/337#discussion_r1838506405)
- generally it is best practice to use `gradle` to copy all needed build/environment related resources to local subproject build folder and run tasks from there
- projects must not use build/environment related resources of other projects directly, be that in the build process or in code
- projects must be isolated and the build process must be isolated and deterministic, we should not go out of the respective project's `build` root when building, supplying all needed resources there using `gradle`

### Alternatives

_No response_

georgi-l95:
## Reviewer Notes
This PR fixes the E2E test suite and also fixes a technical debt regarding the coupling of the sub-projects

## Related Issue(s)
Fixes #343

### DIFF
--- a/suites/src/main/java/org/hiero/block/suites/persistence/positive/PositiveDataPersistenceTests.java
+++ b/suites/src/main/java/org/hiero/block/suites/persistence/positive/PositiveDataPersistenceTests.java
@@ -23,7 +23,7 @@ import org.testcontainers.containers.Container;
 public class PositiveDataPersistenceTests extends BaseSuite {
     // @todo(#371) - the default life/archive root path must be absolute starting from /opt
     private final String[] GET_BLOCKS_COMMAND =
-            new String[] {"find", "../opt/hashgraph/blocknode/data/live", "-mindepth", "19", "-maxdepth", "20"};
+            new String[] {"find", "../opt/hashgraph/blocknode/data/live", "-mindepth", "5", "-maxdepth", "7"};
 
     private Future<?> simulatorThread;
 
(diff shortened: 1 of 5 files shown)


suites/src/main/java/org/hiero/block/suites/persistence/positive/PositiveDataPersistenceTests.java
```diff
@@ -23,7 +23,7 @@
 public class PositiveDataPersistenceTests extends BaseSuite {
     // @todo(#371) - the default life/archive root path must be absolute starting from /opt
     private final String[] GET_BLOCKS_COMMAND =
-            new String[] {"find", "../opt/hashgraph/blocknode/data/live", "-mindepth", "19", "-maxdepth", "20"};
+            new String[] {"find", "../opt/hashgraph/blocknode/data/live", "-mindepth", "5", "-maxdepth", "7"};
```
AlfredoG87, line 26: why the change to this command?

AlfredoG87 (approved):
LGTM

suites/src/main/java/org/hiero/block/suites/persistence/positive/PositiveDataPersistenceTests.java
georgi-l95: We previously saved them in such trie structure -> 0/0/0/0/0/0/0/0/0...0/000000001.blk . With depth around 19
Now we save like this -> 000/000/000/000/000/0000000001.blk with depth around 5 to 7.

PR #964, opened 2025-04-03
fix: Simulator Helm Chart Deployment in Consumer Mode

### LINKED ISSUE #963: Bug: Helm deployment of Consumers broke due to PVC added to Publisher
Simulator in `consumer` mode does not need (at least rn) support for persistent storage, however since it was added to the publisher, now the consumers try to use that resource and since it is already taken by the publisher. then the consumers fail to start.

The Solution is to add a `values ` config value for `consumer` or Mode on the helm chart of the simulator, and only add the PVC storage if is enabled.

AlfredoG87:
- Only adding PVC to mounting to container when deploying in Publisher Mode.

## Reviewer Notes

This can be cherry-picked and added to `rc2` then to `0.8.0` GA version.

## Related Issue(s)
Fixes #963

### DIFF
--- a/charts/blockstream-simulator/templates/deployment.yaml
+++ b/charts/blockstream-simulator/templates/deployment.yaml
@@ -32,7 +32,7 @@ spec:
         - name: data-storage
           persistentVolumeClaim:
             claimName: {{ if .Values.simulator.persistence.data.create }}{{ include "blockstream-simulator-chart.fullname" . }}-data{{ else }}{{ .Values.simulator.persistence.data.existingClaim }}{{ end }}
-      {{- if eq .Values.simulator.config.BLOCK_STREAM_SIMULATOR_MODE "PUBLISER_MODE" }}
+      {{- if eq .Values.simulator.config.BLOCK_STREAM_SIMULATOR_MODE "PUBLISHER_CLIENT" }}
       initContainers:
         - name: init-storage-dirs
           image: busybox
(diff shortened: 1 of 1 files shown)


charts/blockstream-simulator/templates/deployment.yaml
```diff
@@ -32,6 +32,7 @@ spec:
         - name: data-storage
           persistentVolumeClaim:
             claimName: {{ if .Values.simulator.persistence.data.create }}{{ include "blockstream-simulator-chart.fullname" . }}-data{{ else }}{{ .Values.simulator.persistence.data.existingClaim }}{{ end }}
+      {{- if eq .Values.simulator.config.BLOCK_STREAM_SIMULATOR_MODE "PUBLISER_MODE" }}
```
jsync-swirlds, line 35: Should this be publis**h**er?
```suggestion
      {{- if eq .Values.simulator.config.BLOCK_STREAM_SIMULATOR_MODE "PUBLISHER_MODE" }}
```

charts/blockstream-simulator/templates/deployment.yaml
```diff
@@ -60,10 +62,12 @@ spec:
               name: {{ include "blockstream-simulator-chart.fullname" . }}-config
           - secretRef:
               name: {{ include "blockstream-simulator-chart.fullname" . }}-secret
+        {{- if eq .Values.simulator.config.BLOCK_STREAM_SIMULATOR_MODE "PUBLISER_MODE" }}
```
jsync-swirlds, line 65: Same here?
```suggestion
        {{- if eq .Values.simulator.config.BLOCK_STREAM_SIMULATOR_MODE "PUBLISHER_MODE" }}
```

jsync-swirlds (approved):
A couple possible typos, otherwise looks good.

jsync-swirlds (approved):

Nana-EC (approved):

PR #966, opened 2025-04-04
feat: added block stream simulator out-of-order streaming

### LINKED ISSUE #522: Add support to the Block Stream Simulator to send out-of-order blocks
The simulator must be able to, on request, send blocks out of order.
Out of order includes all manner of ordering, including something like 1,3,2,4; 1,2,4,3; 1,76,32,43,2,3,4; or 45,90,91,92).
The simulator should also accept instruction as to how "deep" it's cache _appears_ to be, so, for example, it could stream 43,90,91,92 and only have block 90 and later available; which would help in testing how the BN handles when it needs a different source of blocks to catch up.

Other scenarios and options may arise during testing, and should be supported.

AlexKehayov:
## Reviewer Notes
Implemented unordered block streaming in the simulator, configurable via app.properties.
The feature can be enabled by setting unorderedStream.enabled=true.
Two sub-modes are supported:
- Automatic Scrambling: Generates an unordered stream of blocks based on a specified block range and a scrambling coefficient.
- Fixed Sequence: Streams blocks in a predefined order specified in unorderedStream.fixedStreamingSequence.
Added unit tests to cover both sub-modes and the new configuration logic.

## Related Issue(s)
Closes #522

### DIFF
--- a/simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
+++ b/simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.simulator.config.data;
+
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import java.util.regex.Pattern;
+import org.hiero.block.common.utils.Preconditions;
+import org.hiero.block.simulator.config.logging.Loggable;
+
+/**
+ * Defines the configuration data for the unordered block streaming in the Hedera Block Simulator.
+ *
+ * @param enabled                indicates whether to enable the unordered streaming mode
+ * @param availableBlocks        the numbers of the blocks which will be considered available for sending
+ * @param sequenceScrambleLevel  the coefficient used to randomly scramble the stream
+ * @param fixedStreamingSequence the numbers of the blocks which will form the steam (could be from availableBlocks or not)
+ */
+@ConfigData("unorderedStream")
+public record UnorderedStreamConfig(
+        @Loggable @ConfigProperty(defaultValue = "false") boolean enabled,
+        @Loggable @ConfigProperty(defaultValue = "[1-10]") String availableBlocks,
+        @Loggable @ConfigProperty(defaultValue = "0") int sequenceScrambleLevel,
+        @Loggable @ConfigProperty(defaultValue = "1,2,4,3") String fixedStreamingSequence) {
+
+    public UnorderedStreamConfig {
+        availableBlocks = availableBlocks.trim();
+        fixedStreamingSequence = fixedStreamingSequence.trim();
+        final Pattern pattern = Pattern.compile("^(\\[(\\d+)-(\\d+)\\]|\\d+)(,\\s*(\\[(\\d+)-(\\d+)\\]|\\d+))*$");
+
+        Preconditions.requireInRange(sequenceScrambleLevel, 0, 10, "sequenceScrambleLevel must be between 0 and 10");
+        Preconditions.requireNotBlank(
+                fixedStreamingSequence, "fixedStreamingSequence must not be blank when sequenceScrambleLevel is 0");
+        if (!pattern.matcher(fixedStreamingSequence).matches()) {
+            throw new IllegalArgumentException(fixedStreamingSequence + " does not match the expected format");
+        }
+        Preconditions.requireNotBlank(
+                availableBlocks, "availableBlocks must not be blank when sequenceScrambleLevel is larger than 0");
+        if (!pattern.matcher(availableBlocks).matches()) {
+            throw new IllegalArgumentException(availableBlocks + " does not match the expected format");
+        }
+    }
+}
--- a/simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
+++ b/simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
@@ -2,7 +2,10 @@
 package org.hiero.block.simulator.generator;
 
 import static java.lang.System.Logger.Level.DEBUG;
+import static java.lang.System.Logger.Level.ERROR;
 import static java.lang.System.Logger.Level.INFO;
+import static java.lang.System.Logger.Level.WARNING;
+import static java.util.Objects.requireNonNull;
 
 import com.hedera.hapi.block.stream.BlockProof;
 import com.hedera.hapi.block.stream.protoc.Block;
@@ -12,14 +15,23 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
 import java.lang.System.Logger;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Random;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.hiero.block.common.hasher.Hashes;
 import org.hiero.block.common.hasher.HashingUtilities;
 import org.hiero.block.common.hasher.NaiveStreamingTreeHasher;
 import org.hiero.block.common.hasher.StreamingTreeHasher;
 import org.hiero.block.internal.BlockItemUnparsed;
 import org.hiero.block.simulator.config.data.BlockGeneratorConfig;
+import org.hiero.block.simulator.config.data.UnorderedStreamConfig;
 import org.hiero.block.simulator.config.types.GenerationMode;
 import org.hiero.block.simulator.exception.BlockSimulatorParsingException;
 import org.hiero.block.simulator.generator.itemhandler.BlockHeaderHandler;
@@ -56,17 +68,22 @@ public class CraftBlockStreamManager implements BlockStreamManager {
     private StreamingTreeHasher inputTreeHasher;
     private StreamingTreeHasher outputTreeHasher;
 
+    // Unordered streaming
+    private final boolean unorderedStreamingEnabled;
+    private Iterator<Block> unorderedStreamIterator;
+
     /**
      * Constructs a new CraftBlockStreamManager with the specified configuration.
      *
      * @param blockGeneratorConfig Configuration parameters for block generation
-     * including event and transaction counts
+     *                             including event and transaction counts
      * @param simulatorStartupData simulator startup data used for initialization
      * @throws NullPointerException if blockGeneratorConfig is null
      */
     public CraftBlockStreamManager(
             @NonNull final BlockGeneratorConfig blockGeneratorConfig,
-            @NonNull final SimulatorStartupData simulatorStartupData) {
+            @NonNull final SimulatorStartupData simulatorStartupData,
+            @NonNull final UnorderedStreamConfig unorderedStreamConfig) {
         this.generationMode = blockGeneratorConfig.generationMode();
         this.minEventsPerBlock = blockGeneratorConfig.minEventsPerBlock();
         this.maxEventsPerBlock = blockGeneratorConfig.maxEventsPerBlock();
@@ -81,6 +98,39 @@ public class CraftBlockStreamManager implements BlockStreamManager {
         this.currentBlockNumber = simulatorStartupData.getLatestAckBlockNumber() + 1L;
         this.previousBlockHash = simulatorStartupData.getLatestAckBlockHash();
         LOGGER.log(INFO, "Block Stream Simulator will use Craft mode for block management");
+
+        // Unordered streaming
+        unorderedStreamingEnabled = unorderedStreamConfig.enabled();
+        if (unorderedStreamingEnabled) {
+            initUnorderedStreaming(simulatorStartupData, unorderedStreamConfig);
+        }
+    }
+
+    private void initUnorderedStreaming(
+            SimulatorStartupData simulatorStartupData, UnorderedStreamConfig unorderedStreamConfig) {
+
+        if (simulatorStartupData.isEnabled()) {
+            throw new IllegalStateException("Unordered streaming does not support start-up data enabled");
+        }
+        this.currentBlockNumber = 0;
+        final List<Block> blockStreamList;
+        final int scrambleLevel = unorderedStreamConfig.sequenceScrambleLevel();
+        if (scrambleLevel == 0) {
+            // use a predefined order of streaming from the properties file
+            final LinkedHashSet<Long> fixedStreamingSequence =
+                    parseRangeSet(unorderedStreamConfig.fixedStreamingSequence());
+            blockStreamList = getBlockStreamList(fixedStreamingSequence);
+        } else {
+            // put the predefined available blocks in a list and scramble by the provided coefficient
+            final Set<Long> availableBlocks = parseRangeSet(unorderedStreamConfig.availableBlocks());
+            blockStreamList = getBlockStreamList(availableBlocks);
+            scrambleBlocks(blockStreamList, scrambleLevel);
+        }
+        if (blockStreamList.isEmpty()) {
+            throw new IllegalStateException("No blocks are available for streaming with the current configuration");
+        }
+        unorderedStreamIterator = blockStreamList.iterator();
+        LOGGER.log(INFO, "Unordered streaming is enabled");
     }
 
     /**
@@ -109,11 +159,22 @@ public class CraftBlockStreamManager implements BlockStreamManager {
      * Each block includes a header, events with their transactions and results, and a proof.
      *
      * @return A newly generated Block
-     * @throws IOException if there is an error processing block items
+     * @throws IOException                    if there is an error processing block items
      * @throws BlockSimulatorParsingException if there is an error parsing block components
      */
     @Override
     public Block getNextBlock() throws IOException, BlockSimulatorParsingException {
+        if (unorderedStreamingEnabled) {
+            if (unorderedStreamIterator.hasNext()) {
+                return unorderedStreamIterator.next();
+            }
+            return null;
+        } else {
+            return createNextBlock();
+        }
+    }
+
+    private Block createNextBlock() throws BlockSimulatorParsingException {
         LOGGER.log(DEBUG, "Started creation of block number %s.".formatted(currentBlockNumber));
         // todo(683) Refactor common hasher to accept protoc types, in order to avoid the additional overhead of keeping
         // and unparsing.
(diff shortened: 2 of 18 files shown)


simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * Defines the configuration data for the unordered block streaming in the Hedera Block Simulator.
+ *
+ * @param enabled                indicates whether to enable the unordered streaming mode
+ * @param availableBlocks        the numbers of the blocks which will be considered available for sending
+ * @param sequenceScrambleLevel  the coefficient used to randomly scramble the stream
+ * @param fixedStreamingSequence the numbers of the blocks which will form the steam (could be from availableBlocks or not)
+ */
+@ConfigData("unorderedStream")
+public record UnorderedStreamConfig(
```
AlexKehayov, line 21: This record is responsible for the initialization of the four properties managing the unordered streaming. Default values are set if configurations are not available in the app.properties file.
Provided values and their format are verified in the compact constructor and in the parseRangeSet method below.
Additional getters are available for easier usage throughout the code as values of availableBlocks and fixedStreamingSequence are parsed to sets.

The main things to point here are the following:
enabled -> A boolean that activates/deactivates the mode
availableBlocks -> A set of numeric values representing the numbers of block considered available for streaming. All blocks up to the highest number are generated via craft mode, but only the specified here are used for streaming.
sequenceScrambleLevel -> a value from 0 to 10. When set to 0, the stream includes the blocks with numbers from fixedStreamingSequence in the exact specified order. When larger than 0, serves as a coefficient defining how many swaps are performed based on the number of blocks. Closer to 1 means a small amount of swaps of neighbouring elements, closer to 10 means a lot of swaps, including swaps of distant elements.
fixedStreamingSequence -> Used only when sequenceScrambleLevel is set to 0. Provides the exact order of blocks to be streamed. The generation strategy is the same as for availableBlocks.

For utility, availableBlocks and fixedStreamingSequence are set in format (described below) in the app.properties and converted to linked set of Long values by the additional getters (availableBlocksAsSet and fixedStreamingSequenceAsSet).

simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public Set<Long> availableBlocksAsSet() {
+        return parseRangeSet(availableBlocks);
+    }
+
+    public LinkedHashSet<Long> fixedStreamingSequenceAsSet() {
+        return parseRangeSet(fixedStreamingSequence);
+    }
+
+    private static LinkedHashSet<Long> parseRangeSet(String input) {
```
AlexKehayov, line 51: This method is responsible for validating the format of the values of availableBlocks and fixedStreamingSequence and parsing them to sets.
The format that should be matched in the properties file is the following:
comma separated positive numeric values and/or range of positive numbers from lower to higher in square brackets with a slash between them.
e.g. 1, 4, 3, [7-10], 15, [17-19], 2 which equals 1, 4, 3, 7, 8, 9 , 10, 15, 17, 18, 19, 2

Examples and explanations are provided in short comments in the app.properties file.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -79,6 +92,38 @@ public CraftBlockStreamManager(
...
+            // put the predefined available blocks in a list and scramble by the provided coefficient
+            Set<Long> availableBlocks = unorderedStreamConfig.availableBlocksAsSet();
+            blockStreamList = getBlockStreamList(availableBlocks);
+            blockStreamList = scrambleBlocks(blockStreamList, scrambleLevel);
+        }
+        if (blockStreamList.isEmpty()) {
+            throw new IllegalStateException("No blocks are available for streaming with the current configuration");
+        }
+        unorderedStreamIterator = blockStreamList.iterator();
+        LOGGER.log(INFO, "Unordered streaming is enabled");
```
AlexKehayov, line 126: The base unordered streaming logic is initialized as part of the CraftBlockStreamManager as it depends on its properties for creating the blocks that will be used. Streaming in unordered mode is based on streaming in craft mode, supporting millisPerBlockStreaming() and constantRateStreaming().

Only if the unordered mode is enabled, the block sequence is initialized before the beginning of the streaming.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -107,11 +152,22 @@ public BlockItem getNextBlockItem() {
...
     public Block getNextBlock() throws IOException, BlockSimulatorParsingException {
+        if (unorderedStreamingEnabled) {
+            if (unorderedStreamIterator.hasNext()) {
+                return unorderedStreamIterator.next();
+            }
+            return null;
+        } else {
+            return createNextBlock();
+        }
+    }
```
AlexKehayov, line 168: This method is modified so the blocks can be created before the start of the streaming when unordered mode is enabled.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -180,4 +236,56 @@ private void resetState() {
...
+                LOGGER.log(ERROR, e.getMessage(), e);
+                throw new RuntimeException(e);
+            }
+            for (Long blockNbr : blockSequence) {
+                if (craftedBlocksMap.containsKey(blockNbr)) {
+                    blockStreamList.add(craftedBlocksMap.get(blockNbr));
+                }
+            }
+        }
+        return blockStreamList;
```
AlexKehayov, line 262: This method serves for faster and more convenient retrieval of the blocks, specified for streaming, from the collection of all the crafted blocks.

The reason for crafting the full sequence of blocks and picking only the specified ones is to preserve the authentic proofs and keep the stream realistic.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -180,4 +236,56 @@ private void resetState() {
...
+                    } else {
+                        // Higher coefficient allows more distant swaps
+                        index2 = random.nextInt(n);
+                    }
+                    Collections.swap(scrambled, index1, index2);
+                }
+            }
+        }
+        return scrambled;
+    }
```
AlexKehayov, line 290: This method scrambles the available blocks based on the sequenceScrambleLevel and the size of the block list.

simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublisherClientModeHandler.java
```diff
@@ -132,6 +133,10 @@ private void millisPerBlockStreaming() throws IOException, InterruptedException,
                 INFO,
                 "Number of BlockItems sent by the Block Stream Simulator: "
                         + metricsService.get(LiveBlockItemsSent).get());
+        LOGGER.log(
+                INFO,
+                "Number of Blocks sent by the Block Stream Simulator: "
+                        + metricsService.get(LiveBlocksSent).get());
```
AlexKehayov, line 139: Added this log for convenience.

simulator/src/main/java/org/hiero/block/simulator/startup/SimulatorStartupData.java
```diff
@@ -65,4 +65,6 @@ void updateLatestAckBlockStartupData(
      */
     @NonNull
     byte[] getLatestAckBlockHash();
+
+    boolean isEnabled();
```
AlexKehayov, line 69: Added this because unordered mode is not compatible with start-up data and it should be checked on initialization.

simulator/src/main/resources/app.properties
```diff
@@ -41,3 +41,26 @@ prometheus.endpointPortNumber=9998
...
+# Coefficient that determines how scrambled the stream sequence will be (range: 0 to 10).
+# - 0 -> The sequence follows unorderedStream.fixedStreamingSequence exactly.
+# - 1-5 -> Minor sequence variations.
+# - 6-10 -> Major sequence changes.
+#unorderedStream.sequenceScrambleLevel=10
+
+# Defines the exact streaming sequence when sequenceScrambleLevel is 0.
+# Ranges are expanded, e.g., [1-3], 99, [5-6], 4, 55 -> 1, 2, 3, 99, 5, 6, 4, 55.
+# Only the elements that exist in both
+#unorderedStream.fixedStreamingSequence=[1-3], 99, [5-6], 4, 55
```
AlexKehayov, line 66: The key properties are meant to be set here. Additional information and examples are also provided.

AlexKehayov (commented):
A self-review of the changes.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -107,11 +152,22 @@ public BlockItem getNextBlockItem() {
...
      *
      * @return A newly generated Block
-     * @throws IOException if there is an error processing block items
+     * @throws IOException                    if there is an error processing block items
      * @throws BlockSimulatorParsingException if there is an error parsing block components
      */
     @Override
     public Block getNextBlock() throws IOException, BlockSimulatorParsingException {
+        if (unorderedStreamingEnabled) {
+            if (unorderedStreamIterator.hasNext()) {
```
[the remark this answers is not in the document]
AlexKehayov: When the iterator reaches the end of the collection, the program simply ends - both in CONSTANT_RATE and MILLIS_PER_BLOCK streaming mode. Currently, the connection does not seem to be explicitly closed in none of the generation and streaming modes. Maybe it would be a good thing to add  publishStreamGrpcClient.shutdown() in PublisherClientModeHandler -> start()

```  
@Override
    public void start() throws BlockSimulatorParsingException, IOException, InterruptedException {
        LOGGER.log(INFO, "Block Stream Simulator is starting in publisher mode.");
        try {
            if (streamingMode == StreamingMode.MILLIS_PER_BLOCK) {
                millisPerBlockStreaming();
            } else {
                constantRateStreaming();
            }
        } finally {
            publishStreamGrpcClient.shutdown();
        }
        LOGGER.log(INFO, "Block Stream Simulator has stopped streaming.");
    }
```

I will push this change shortly and follow the thread in case any additional changes are needed.

simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
```diff
@@ -0,0 +1,147 @@
...
+                throw new IllegalArgumentException("sequenceScrambleLevel must be between 0 and 10");
+            }
+            if (sequenceScrambleLevel == 0 && fixedStreamingSequence.isBlank()) {
+                throw new IllegalArgumentException(
+                        "fixedStreamingSequence must not be blank when sequenceScrambleLevel is 0");
+            }
+            if (sequenceScrambleLevel != 0 && availableBlocks.isBlank()) {
+                throw new IllegalArgumentException(
+                        "availableBlocks must not be blank when sequenceScrambleLevel is larger than 0");
+            }
```
ata-nas, line 39: You can utilize the preconditions we have created, in this case you can use the requireInRange precondition and others as they fit.

simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
```diff
@@ -0,0 +1,147 @@
...
+                        }
+                        result.add(value);
+                    }
+                }
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Exception in parsing input: " + input, e);
+            }
+        }
+        return result;
+    }
```
ata-nas, line 91: Not blocking, but having logic inside the config classes seems like a bad idea (even for the simulator). Config classes should be value holders and nothing else. It seems this logic that is used only inside the craft manager should reside internally there.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -79,6 +92,38 @@ public CraftBlockStreamManager(
...
+            initUnorderedStreaming(simulatorStartupData, unorderedStreamConfig);
+        }
+    }
+
+    private void initUnorderedStreaming(
+            SimulatorStartupData simulatorStartupData, UnorderedStreamConfig unorderedStreamConfig) {
+
+        if (simulatorStartupData.isEnabled()) {
+            throw new IllegalStateException("Unordered streaming does not support start-up data enabled");
+        }
```
ata-nas, line 108: Why do we throw here? Why does startup data interfere with the unordered streaming? It seems like generally these 2 should not be connected? I remember we discussed that block creation with the unordered stream starts from the beginning. Is that the reason?

As generally I do not expect the unordered streams to be useful for anything else that tests which aim to assert erroneous state/input, I do not think we will need the startup data, so I do think it is absolutely fine to throw here. Startup data is generally useful when we stream correctly.

 Just curious.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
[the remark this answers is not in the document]
ata-nas: Well, the thing here is that I do not think that the simulator is important. After all, we aim to simulate an erroneous stream that the block node should react to in some manner. So do we generally expect/want something from the simulator after it has correctly produced the erroneous result we desire? Closing the connection is an interesting thing, should we not expect the block node to do it? Curious as to how/what we should generally expect as behavior.

simulator/src/test/java/org/hiero/block/simulator/config/data/UnorderedStreamConfigTest.java
ata-nas, line 1: Almost all tests here can and IMO should be parameterized. I would suggest we do so. Also having javadocs to assert the goal of the test is a very good practice since intent can be juxtaposed against test implementation. Optionally, having `@Nested` categorization is also good, splits aspects of the class and can be better read.

Here a good example of some tests we recently introduced that you can take a look if you decide to improve based on the suggestions:

https://github.com/hiero-ledger/hiero-block-node/tree/505-plugins-services/block-node/block-providers/files.recent/src/test/java/org/hiero/block/node/blocks/files/recent

https://github.com/hiero-ledger/hiero-block-node/tree/505-plugins-services/block-node/block-providers/files.historic/src/test/java/org/hiero/block/node/blocks/files/historic

These are non-blocking suggestions.

ata-nas (commented):
LG, some suggestions for cleanup (non-blocking) and some questions, will run it and come back for round 2.

simulator/src/main/resources/app.properties
```diff
@@ -41,3 +41,26 @@ prometheus.endpointPortNumber=9998
 
 # Simulator startup data
 #simulator.startup.data.enabled=true
+
+# ----------------------------------------------
+# Unordered Stream Configuration
+# ----------------------------------------------
+
+# Enables the unordered stream mode. (values: true/false)
+unorderedStream.enabled=false
```
ata-nas, line 50: We should probably comment out line 50, we have the setting false by default no?

```suggestion
# unorderedStream.enabled=false
```

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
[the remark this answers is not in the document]
AlexKehayov: Additional checks and final variables will be pushed shortly too.

simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
AlexKehayov: Would it be fine if the method is moved to a separate util class, so testing is more convenient? Future utility/parsing methods can be added there as well.

simulator/src/main/resources/app.properties
AlexKehayov: You are absolutely correct.

simulator/src/test/java/org/hiero/block/simulator/config/data/UnorderedStreamConfigTest.java
AlexKehayov: Good idea, will do that.

AlexKehayov:
@georgi-l95 @ata-nas Thank you for your feedback and suggestions, all requested changes are pushed. :)

simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
```diff
@@ -0,0 +1,106 @@
...
+            if (sequenceScrambleLevel == 0) {
+                Preconditions.requireNotBlank(
+                        fixedStreamingSequence,
+                        "fixedStreamingSequence must not be blank when sequenceScrambleLevel is 0");
+            } else {
+                Preconditions.requireNotBlank(
+                        availableBlocks,
+                        "availableBlocks must not be blank when sequenceScrambleLevel is larger than 0");
+            }
+        }
```
ata-nas, line 41: Why do we have to over complicate this? It feels like we should simply do the preconditions directly.

simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public Set<Long> availableBlocksAsSet() {
+        return parseRangeSet(availableBlocks);
+    }
+
+    public LinkedHashSet<Long> fixedStreamingSequenceAsSet() {
+        return parseRangeSet(fixedStreamingSequence);
+    }
```
ata-nas, line 50: Yeah, there is still logic inside the config class, I was noting that this whole thing should be internal to the place that needs to use it, the config class just carries the input values. In your case that is `availableBlocks` && `fixedStreamingSequence`

simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
```diff
@@ -0,0 +1,106 @@
...
+    public static class Builder {
+        private boolean enabled = false;
+        private String availableBlocks = "";
+        private int sequenceScrambleLevel = 0;
+        private String fixedStreamingSequence = "";
+
+        /**
+         * Creates a new instance of the {@code Builder} class with default configuration values.
+         */
+        public Builder() {
```
ata-nas, line 73: I question the existence of this builder. We should generally not use this and instantiate configs with the config dependency builder.

simulator/src/main/java/org/hiero/block/simulator/config/util/ParsingUtils.java
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.simulator.config.util;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.LinkedHashSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ParsingUtils {
```
ata-nas, line 10: I see no point in having this class, this seems like logic that must be internal to the craft block manager as it is the only place that uses this.

ata-nas (dismissed):
Non blocking comments. We can improve in the future if we ever need to.

simulator/src/main/java/org/hiero/block/simulator/config/data/UnorderedStreamConfig.java
ata-nas: I would say no, why do that? This logic is only ever used inside the craft block manager and is way too non-generic to be part of a util class. Could be a future improvement.

ata-nas (dismissed):
LGTM, non-blocking items above for acknowledgement

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -180,4 +237,57 @@ private void resetState() {
...
+        requireNonNull(list);
+        if (!list.isEmpty()) {
+            final List<Block> originalList = new ArrayList<>(list);
+            final Random random = new Random();
+            final int n = list.size();
+
+            // The number of swaps is proportional to the coefficient
+            final int swapCount = (int) ((coefficient / 10.0) * (n * 2));
+
+            while (list.equals(originalList)) {
```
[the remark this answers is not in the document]
AlexKehayov: I added some additional checks to prevent the potential problem in case of bad configs provided.

AlexKehayov:
Just pushed some changes and refactoring related to the non-blocking changes that @ata-nas mentioned above.

georgi-l95 (approved):

ata-nas (dismissed):
LG, just a general question.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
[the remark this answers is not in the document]
AlexKehayov: Precondition checks in UnorderedStreamConfig take care of such scenarios. By bad configs I meant a list too small for guaranteed and efficient scrambling. Such input may be intentional in some cases.

simulator/src/main/java/org/hiero/block/simulator/generator/CraftBlockStreamManager.java
```diff
@@ -81,6 +98,39 @@ public CraftBlockStreamManager(
...
+        }
+    }
+
+    private void initUnorderedStreaming(
+            SimulatorStartupData simulatorStartupData, UnorderedStreamConfig unorderedStreamConfig) {
+
+        if (simulatorStartupData.isEnabled()) {
+            throw new IllegalStateException("Unordered streaming does not support start-up data enabled");
+        }
+        this.currentBlockNumber = 1; // start block crafting from block number 1
```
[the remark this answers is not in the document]
AlexKehayov: At the time, I added this line to prevent an issue, that currently I cannot recreate (probably due to other code changes). I will set the value to 0 and push.

georgi-l95 (approved):

ata-nas (approved):

PR #970, opened 2025-04-07
chore: Add UTs for Recent BlockFileBlockAccessor exception handling

### LINKED ISSUE #969: Files Recent Plugin Exception Unit Tests
### Story Form

As a Block Node Developer
I want to have unit tests for the Files Recent Plugin exception cases
So that I can have confidence in existing logic and catch regressions

### Technical Notes

Add additional unit tests for exception handle paths

Nana-EC:
## Reviewer Notes
Add additional UTs for exception handling cases in BlockFileBlockAccessor
- Add nested depth test for additional `delete()` confidence
- Add logic to handle IOException and ParseException in block()
- Add logic to handle IOException and ParseException in blockUnparsed()
- Add logic to handle IOException in blockBytes() 
- Add logic to handle IOException in writeBytesTo() 
- Add logic to handle IOException in writeBytesTo() 
- Move shared setup logic from `createInstanceToTest()` into `buildAndCreateBlockAndGetAssociatedAccessor()` and `createBlockAndGetAssociatedAccessor()`

## Related Issue(s)
Fixes #969

### DIFF
--- a/block-node/block-providers/files.recent/src/test/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessorTest.java
+++ b/block-node/block-providers/files.recent/src/test/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessorTest.java
@@ -21,7 +21,6 @@ import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
-
 import org.assertj.core.api.Assertions;
 import org.hiero.block.node.app.fixtures.blocks.SimpleTestBlockItemBuilder;
 import org.hiero.block.node.base.CompressionType;
@@ -236,7 +235,8 @@ class BlockFileBlockAccessorTest {
             final Block expected = new Block(List.of(blockItems));
             final Bytes protoBytes = Block.PROTOBUF.toBytes(expected);
             // create instance to test
-            final BlockFileBlockAccessor toTest = createBlockAndGetAssociatedAccessor(blockFilePath, compressionType, protoBytes);
+            final BlockFileBlockAccessor toTest =
+                    createBlockAndGetAssociatedAccessor(blockFilePath, compressionType, protoBytes);
             // test accessor.block()
             final Block actual = toTest.block();
             assertThat(actual).isEqualTo(expected);
@@ -257,7 +257,8 @@ class BlockFileBlockAccessorTest {
             final Block expected = new Block(List.of(blockItems));
             final Bytes protoBytes = Block.PROTOBUF.toBytes(expected);
             // create instance to test
-            final BlockFileBlockAccessor toTest = createBlockAndGetAssociatedAccessor(blockFilePath, compressionType, protoBytes);
+            final BlockFileBlockAccessor toTest =
+                    createBlockAndGetAssociatedAccessor(blockFilePath, compressionType, protoBytes);
 
             // delete the file to simulate NoSuchFileException IOException
             Files.delete(blockFilePath);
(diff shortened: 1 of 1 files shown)


block-node/block-providers/files.recent/src/test/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessorTest.java
```diff
@@ -232,12 +235,71 @@ void testBlock(final CompressionType compressionType) throws IOException {
...
+            Files.delete(blockFilePath);
+
+            // test accessor.block() expecting IOException
+            try {
+                toTest.block();
+                assertThat(false).isTrue(); // exception should have been thrown
+            } catch (Throwable e) {
+                // expected
+                Assertions.assertThat(e).isInstanceOf(UncheckedIOException.class);
+            }
```
ata-nas, line 273: AssertJ offers fluent apis for checking exceptions thrown. The below code should be in place of the whole try-catch. It will fail if no exception is thrown.  You can consider this approach to be applied everywhere.

```suggestion
            assertThatExceptionOfType(UncheckedIOException.class).isThrownBy(toTest::block);
```

On another note, if for any reason you want to fail the test, we should use `Assertions.fail("some message");` (again coming from AssertJ) instead of trying to produce a fake assert. But in this case none of this should be used.

ata-nas (commented):
LG, I have one suggestion that will cleanup the try-catches and make the tests more readable.

On another note, while it is good that we have more coverage, I was a bit hesitant to add these (or similar asserts) because there would still be IMO changes to the production logic. This would mean that we will have to revisit this. Catching regressions is mostly in happy path tests (granted not always).

I am still exploring the production changes that happen daily and already have compiled a list of points where we need to be extra extra careful. Just to give a quick example (I will post the list with considerations on @jasperpotts PR 505), what happens when we have received block 1, written it to unverified, but then verification fails and then we receive it again, but this time verification passes before we have overwritten the newly received items. Will then the failed block be moved to live? This is a simple example with some concerns/race conditions etc. etc. Addressing such will change the production logic, hence will have to revisit the tests. Just something to keep in mind. Maybe it is a question of finding the right balance for now and when we are certain in the production logic, we go down on tests 💯.

block-node/block-providers/files.recent/src/test/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessorTest.java
Nana-EC: Agreed, thanks. Will adopt

block-node/block-providers/files.recent/src/test/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessorTest.java
```diff
@@ -1,20 +1,19 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.blocks.files.recent;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
-import static org.assertj.core.api.Assertions.assertThatNoException;
-import static org.assertj.core.api.Assertions.assertThatNullPointerException;
+import static org.assertj.core.api.Assertions.*;
```
ata-nas, line 4: I think the wildcard `import static org.assertj.core.api.Assertions.*;` is wrong? Maybe an IDE automatic change? Does spotless not change that to the correct no wildcard import?

ata-nas (approved):
LGTM, just an import nit.

jasperpotts (approved):

PR #979, opened 2025-04-07
feat: Get single block plugin / module

### LINKED ISSUE #976: Get Single Block API and Module
### Story Form

As a Block Node Developer
I want to be able to fetch a Single Block
So that we regain parity with the current version

### Technical Notes

In order to have feature parity with current version

AlfredoG87:
## Reviewer Notes
- Added new module org.hiero.block.node.access.service
- Added new Plugin BlockAccessServicePlugin
- Added UT for the new plugin
- Adding new module to BN App

## Related Issue(s)
Fixes #976


block-node/single-block/src/main/java/org/hiero/block/node/single/block/GetSingleBlockPlugin.java
```diff
@@ -0,0 +1,152 @@
...
+    private SingleBlockResponse handleSingleBlockRequest(SingleBlockRequest request) {
+        LOGGER.log(
+                System.Logger.Level.DEBUG, "Received SingleBlockRequest for block number: {0}", request.blockNumber());
+
+        try {
+            // Validate request
+            if (request.retrieveLatest() && request.blockNumber() != 0) {
+                LOGGER.log(System.Logger.Level.WARNING, "Invalid request: both block_number and retrieve_latest set");
+                return new SingleBlockResponse(SingleBlockResponseCode.READ_BLOCK_NOT_FOUND, null);
+            }
```
AlfredoG87, line 90: Should we just ignore the `block_number` and return the `latest`? or the other way around?
Thoughts on this?

jsync-swirlds: If I were calling the API, I would expect to get "latest" in that case.  Principle of least surprise suggests that might be the better option.

Of course, we should also make sure it's very clear in the API specifications that this is the behavior, just to be sure.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+                System.Logger.Level.DEBUG, "Received SingleBlockRequest for block number: {0}", request.blockNumber());
+        requestCounter.increment();
+
+        try {
+            // Validate request
+            if (request.retrieveLatest() && request.blockNumber() != 0) {
+                // LOGGER.log(System.Logger.Level.WARNING, "Invalid request: both block_number and retrieve_latest
+                // set");
+                // requestCounterNotFound.increment();
+                // return new SingleBlockResponse(SingleBlockResponseCode.READ_BLOCK_NOT_FOUND, null);
```
ata-nas, line 57: Should we use/delete this commented out code?

ata-nas (commented):
Looks good. One nit.

block-node/single-block/src/main/java/org/hiero/block/node/single/block/GetSingleBlockPlugin.java
AlfredoG87: This has been addressed, however left comments in case someone else in the team disagrees.
@jasperpotts 

I think that the most intuitive case is when you ignore the `latest` and return the specific `block` the contrary of what you are suggestion, my rationale is that maybe they did not Read the spec (as everybody does) and just intuitively use the API, in that case they might Think that latest means something else and set it to true on false assumptions like the latest block, and block_number is very specific and unambiguous and harder to be set under wrong assumptions.

However, this is something minimal and did the change since I am OK with either options. ignore one field and use the other (2 options) or fail (as I originally had it).

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: Just left a comment on the why I left it there, however hope to delete it soon. 👍

block-node/single-block/src/main/java/org/hiero/block/node/single/block/GetSingleBlockPlugin.java
jsync-swirlds: It's easy enough to change either way.  The important thing is to make sure it's well documented.  A single block request is unlikely to be a make-or-break item for an application, and it's both easy to detect and easy to test when making the request.

Side note, it's a lot easier to tell your request is incorrect (in my opinion) if you expect a specific block and get the latest, rather than the reverse.

Further Side Note, if we end up making the API include both a result code and a detail code, we could probably have something like `SUCCESS` for the main code and `BOTH_LATEST_AND_NUMBER_REQUESTED` as a detail code.  Something for future consideration (well outside this PR).

block-node/block-access/build.gradle.kts
```diff
@@ -0,0 +1,20 @@
...
+
+description = "Hiero Block Node - Block Access Service API Module"
+
+// Remove the following line to enable all 'javac' lint checks that we have turned on by default
+// and then fix the reported issues.
+tasks.withType<JavaCompile>().configureEach { options.compilerArgs.add("-Xlint:-exports") }
+
+mainModuleInfo {
+    runtimeOnly("com.swirlds.config.impl")
+    runtimeOnly("org.apache.logging.log4j.slf4j2.impl")
```
jsync-swirlds, line 12: Why is this included?  We should not need this (and definitely should not include it if we can avoid that).

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Handle a request for a single block
+     *
+     * @param request the request containing the block number or latest flag
+     * @return the response containing the block or an error status
+     */
+    private SingleBlockResponse handleSingleBlockRequest(SingleBlockRequest request) {
+        LOGGER.log(
+                System.Logger.Level.DEBUG, "Received SingleBlockRequest for block number: {0}", request.blockNumber());
```
jsync-swirlds, line 48: Why not `import System.Logger.Level` to trim this line down (and other similar lines)?

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+        try {
+            // Validate request
+            if (request.retrieveLatest() && request.blockNumber() != 0) {
+                // LOGGER.log(System.Logger.Level.WARNING, "Invalid request: both block_number and retrieve_latest
+                // set");
+                // requestCounterNotFound.increment();
+                // return new SingleBlockResponse(SingleBlockResponseCode.READ_BLOCK_NOT_FOUND, null);
+
+                LOGGER.log(
+                        System.Logger.Level.WARNING,
```
jsync-swirlds, line 60: This should be `INFO` at most, ideally it would be `DEBUG`, because operations doesn't really care that the client sent an incorrectly formatted request, the block node is not having any kind of trouble.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+                        "Both block_number and retrieve_latest set. Using retrieve_latest instead of block_number: {0}",
+                        request.blockNumber());
+            }
+            long blockNumberToRetrieve;
+
+            // if retrieveLatest is set, get the latest block number
+            if (request.retrieveLatest()) {
+                blockNumberToRetrieve = blockProvider.latestBlockNumber();
+                if (blockNumberToRetrieve < 0) {
+                    LOGGER.log(System.Logger.Level.WARNING, "Latest block number not available");
```
jsync-swirlds, line 70: `WARNING` is usually where operations sets alerts.  I would recommend reducing this to `INFO`.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+                blockNumberToRetrieve = request.blockNumber();
+            }
+
+            // Check if block is within available range
+            long lowestBlockNumber = blockProvider.oldestBlockNumber();
+            long highestBlockNumber = blockProvider.latestBlockNumber();
+
+            if (blockNumberToRetrieve < lowestBlockNumber || blockNumberToRetrieve > highestBlockNumber) {
+                LOGGER.log(
+                        System.Logger.Level.WARNING,
```
jsync-swirlds, line 84: This should be `DEBUG` (it's a client issue, not a server issue).

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+                requestCounterNotAvailable.increment();
+                return new SingleBlockResponse(SingleBlockResponseCode.READ_BLOCK_NOT_AVAILABLE, null);
+            }
+
+            // Retrieve the block
+            Block block = blockProvider.block(blockNumberToRetrieve).block();
+            requestCounterSuccess.increment();
+            return new SingleBlockResponse(SingleBlockResponseCode.READ_BLOCK_SUCCESS, block);
+
+        } catch (Exception e) {
```
jsync-swirlds, line 98: It would be good to be _specific_ for all checked exceptions, and then include only `RuntimeException` for "generic" catch.
Example assuming the only checked exception is an `IOException`.
```suggestion
        } catch (IOException | RuntimeException e) {
```

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+            LOGGER.log(System.Logger.Level.ERROR, "Failed to retrieve block number: {0}", request.blockNumber());
+            requestCounterNotFound.increment();
+            return new SingleBlockResponse(SingleBlockResponseCode.READ_BLOCK_NOT_FOUND, null);
+        }
+    }
+
+    // <editor-fold desc="BlockNodePlugin Methods">
```
jsync-swirlds, line 105: Probably shouldn't include editor-specific comments; it tends to create a lot of clutter  over time.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+        return "BlockAccessService";
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @NonNull
+    @Override
+    public String fullName() {
+        return "com.hedera.hapi.block." + serviceName();
```
jsync-swirlds, line 163: Shouldn't this be `org.hiero.block.node.`?

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+     * {@inheritDoc}
+     *
+     * This is called each time a new request is received.
+     */
+    @NonNull
+    @Override
+    public Pipeline<? super Bytes> open(
+            @NonNull Method method, @NonNull RequestOptions requestOptions, @NonNull Pipeline<? super Bytes> pipeline)
+            throws GrpcException {
+        final BlockAccessServiceMethod blockAccessServiceMethod = (BlockAccessServiceMethod) method;
```
jsync-swirlds, line 185: Should we have an `instanceof` check for this?

jsync-swirlds (approved):
Mostly just mild suggestions or questions.
One case of `hedera` naming that _might_ need to change (not sure).

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: Yeah, I agree that Debug should be ideal, however I would like to know by querying the logs on loki how often and frequent does this happens but we only keep logs by default up to INFO, and I agree that on production Debug should not be enabled more than if necessary for troubleshooting.

AlfredoG87: will make it INFO

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: No, this needs to match with this on the Protobuf.
https://github.com/hashgraph/hedera-protobufs/blob/main/block/block_service.proto#L14

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: We don't have it on other GRPC plugins, and it might be something that might never happen and is actually not needed, other than a mistake that we would found on early testing and it always fails.

The rationale on why this should never happen in that the methods that are given by the same class when implementing a `ServiceInterface`  an enum is created that implements `Method` and is needed for returning `List<Method> methods()` that also is implemented from the said interface, so this 

So I don't think is actually needed to be honest.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: I know, but is really nice to be able to collapse different sections of the code logically (in this case by interface implemented) :)

block-node/block-access/build.gradle.kts
AlfredoG87: I have removed it, however almost all other plugins also have this included, is why I included it, assumed it was necessary.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: ✅  Done, good suggestin! 👍

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: already deleted! ✅

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: ✅  Done

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: ✅ Done!

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: ✅ Done!

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: Agreed 👍 

✅ Done

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
jsync-swirlds: If we want a count then a metric would be more appropriate, IMO.

Something to consider for a future PR, perhaps.

block-node/block-access/build.gradle.kts
jsync-swirlds: We probably need to remove from the other plugins, then.
It's important to avoid cargo-cult entries in build files.

jasperpotts (approved):

Nana-EC (commented):
Late feedback, let's see if Github still captures it

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+                    LOGGER.log(System.Logger.Level.WARNING, "Latest block number not available");
+                    requestCounterNotAvailable.increment();
+                    return new SingleBlockResponse(SingleBlockResponseCode.READ_BLOCK_NOT_AVAILABLE, null);
+                }
+            } else {
+                blockNumberToRetrieve = request.blockNumber();
+            }
+
+            // Check if block is within available range
+            long lowestBlockNumber = blockProvider.oldestBlockNumber();
```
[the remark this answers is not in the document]
AlfredoG87: We haven't defined a caching story just yet, and I think it will be an optimization that we might want to add based on metrics and actual user usage, after we are in production.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+
+        try {
+            // Validate request
+            if (request.retrieveLatest() && request.blockNumber() != 0) {
+                // LOGGER.log(System.Logger.Level.WARNING, "Invalid request: both block_number and retrieve_latest
+                // set");
+                // requestCounterNotFound.increment();
+                // return new SingleBlockResponse(SingleBlockResponseCode.READ_BLOCK_NOT_FOUND, null);
+
+                LOGGER.log(
```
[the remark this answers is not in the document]
AlfredoG87: Will address this on a follow-up PR 👍

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -0,0 +1,198 @@
...
+                }
+            } else {
+                blockNumberToRetrieve = request.blockNumber();
+            }
+
+            // Check if block is within available range
+            long lowestBlockNumber = blockProvider.oldestBlockNumber();
+            long highestBlockNumber = blockProvider.latestBlockNumber();
+
+            if (blockNumberToRetrieve < lowestBlockNumber || blockNumberToRetrieve > highestBlockNumber) {
```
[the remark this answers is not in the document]
AlfredoG87: Good Idea, let me see if we can follow up PR.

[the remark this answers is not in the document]
AlfredoG87: There is already a method for this `contains` the name was so generic I overlooked it 😅

PR #987, opened 2025-04-08
feat: added block stream simulator mid-block sending failure

### LINKED ISSUE #523: Add support to the Block Stream Simulator to fail sending mid-block
When acting as "publisher" for the block node, the simulator needs to be able to simulate many different failure scenarios. This issue asks to support a scenario where a publisher fails or experiences a network interruption in the middle of sending a block (so after sending the BlockHeader, but before sending the last BlockItem or the BlockProof).
Ideally the simulator will calculate a random offset within the block and stop sending data at that point.
The feature should include the option to either disconnect abruptly (without closing the connection or sending an EndOfStream message) or to send an EndOfStream message.
The first simulates a network or system failure, the second simulates a detected error in the publisher (e.g. an ISS in a consensus node).

AlexKehayov:
## Reviewer Notes
Mid-block sending failure can be triggered by setting blockStream.midBlockFailType and blockStream.midBlockFailOffset in app.properties. midBlockFailType takes NONE/ABRUPT/EOS as values. NONE is the default value which disables the feature. ABRUPT causes a runtime exception after sending the header and a random amount of items (without the proof). Currently, EOS (end of stream) causes calling of onError after sending the header and a random amount of items because PublishStreamRequest does not support end of stream messages, only a BlockItemSet.
midBlockFailOffset takes a numeric value and sets the amount on blocks to be sent before the failing block.
Test methods for the new functionality are added.
The streamBlockItem method is removed as it was redundant.

## Related Issue(s)
 Closes #523

### DIFF
--- a/simulator/src/main/java/org/hiero/block/simulator/config/data/BlockStreamConfig.java
+++ b/simulator/src/main/java/org/hiero/block/simulator/config/data/BlockStreamConfig.java
@@ -4,6 +4,7 @@ package org.hiero.block.simulator.config.data;
 import com.swirlds.config.api.ConfigData;
 import com.swirlds.config.api.ConfigProperty;
 import org.hiero.block.simulator.config.logging.Loggable;
+import org.hiero.block.simulator.config.types.MidBlockFailType;
 import org.hiero.block.simulator.config.types.SimulatorMode;
 import org.hiero.block.simulator.config.types.StreamingMode;
 
@@ -26,7 +27,9 @@ public record BlockStreamConfig(
         @Loggable @ConfigProperty(defaultValue = "100_000") int maxBlockItemsToStream,
         @Loggable @ConfigProperty(defaultValue = "MILLIS_PER_BLOCK") StreamingMode streamingMode,
         @Loggable @ConfigProperty(defaultValue = "1000") int millisecondsPerBlock,
-        @Loggable @ConfigProperty(defaultValue = "1000") int blockItemsBatchSize) {
+        @Loggable @ConfigProperty(defaultValue = "1000") int blockItemsBatchSize,
+        @Loggable @ConfigProperty(defaultValue = "NONE") MidBlockFailType midBlockFailType,
+        @Loggable @ConfigProperty(defaultValue = "0") long midBlockFailOffset) {
 
     /**
      * Creates a new {@link Builder} instance for constructing a {@code BlockStreamConfig}.
@@ -48,6 +51,8 @@ public record BlockStreamConfig(
         private StreamingMode streamingMode = StreamingMode.MILLIS_PER_BLOCK;
         private int millisecondsPerBlock = 1000;
         private int blockItemsBatchSize = 1000;
+        private MidBlockFailType midBlockFailType = MidBlockFailType.NONE;
+        private long midBlockFailOffset = 0;
 
         /**
          * Creates a new instance of the {@code Builder} class with default configuration values.
@@ -134,6 +139,28 @@ public record BlockStreamConfig(
         }
 
         /**
+         * Sets a failure type to occur while streaming.
+         *
+         * @param midBlockFailType the failure type
+         * @return this {@code Builder} instance
+         */
+        public Builder midBlockFailType(MidBlockFailType midBlockFailType) {
+            this.midBlockFailType = midBlockFailType;
+            return this;
+        }
+
+        /**
+         * Sets the index of the failing block.
+         *
+         * @param midBlockFailOffset the index of the failing block
+         * @return this {@code Builder} instance
+         */
+        public Builder midBlockFailOffset(long midBlockFailOffset) {
+            this.midBlockFailOffset = midBlockFailOffset;
+            return this;
+        }
+
+        /**
          * Builds a new {@link BlockStreamConfig} instance with the configured values.
          *
          * @return a new {@code BlockStreamConfig}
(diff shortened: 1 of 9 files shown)


simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamGrpcClientImpl.java
```diff
@@ -192,4 +167,28 @@ public void shutdown() throws InterruptedException {
...
+            requestStreamObserver.onNext(PublishStreamRequest.newBuilder()
+                    .setBlockItems(BlockItemSet.newBuilder()
+                            .addAllBlockItems(streamingBatchBeforeFail)
+                            .build())
+                    .build());
+            if (blockStreamConfig.midBlockFailType() == MidBlockFailType.ABRUPT) {
+                throw new RuntimeException("Configured abrupt disconnection occurred");
+            } else if (blockStreamConfig.midBlockFailType() == MidBlockFailType.EOS) {
+                requestStreamObserver.onError(new Exception("Configured failure occurred, calling onError()"));
+            }
```
georgi-l95, line 191:
```suggestion
    private void handleMidBlockFailIfSet(@NonNull final List<BlockItem> streamingBatch) {
        requireNonNull(streamingBatch);
        if (blockStreamConfig.midBlockFailType() == MidBlockFailType.NONE && streamingBatch.size() < 3) {
              return;
         }
         final int failIndex = new Random().nextInt(1, streamingBatch.size() - 1);
         final List<BlockItem> streamingBatchBeforeFail = streamingBatch.subList(0, failIndex);
         requestStreamObserver.onNext(PublishStreamRequest.newBuilder()
                    .setBlockItems(BlockItemSet.newBuilder()
                            .addAllBlockItems(streamingBatchBeforeFail)
                            .build())
                    .build());
            if (blockStreamConfig.midBlockFailType() == MidBlockFailType.ABRUPT) {
                throw new RuntimeException("Configured abrupt disconnection occurred");
            } 
            requestStreamObserver.onError(new Exception("Configured failure occurred, calling onError()"));
```

simulator/src/main/java/org/hiero/block/simulator/config/data/BlockStreamConfig.java
```diff
@@ -26,7 +27,9 @@ public record BlockStreamConfig(
         @Loggable @ConfigProperty(defaultValue = "100_000") int maxBlockItemsToStream,
         @Loggable @ConfigProperty(defaultValue = "MILLIS_PER_BLOCK") StreamingMode streamingMode,
         @Loggable @ConfigProperty(defaultValue = "1000") int millisecondsPerBlock,
-        @Loggable @ConfigProperty(defaultValue = "1000") int blockItemsBatchSize) {
+        @Loggable @ConfigProperty(defaultValue = "1000") int blockItemsBatchSize,
+        @Loggable @ConfigProperty(defaultValue = "NONE") MidBlockFailType midBlockFailType,
+        @Loggable @ConfigProperty(defaultValue = "0") long midBlockFailOffset) {
```
georgi-l95, line 32: I forgot, why do we need the offset ?

simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamGrpcClientImpl.java
georgi-l95: Just a suggestion.

georgi-l95 (changes requested):
Overal looks good, I left a question and suggestion.

simulator/src/main/java/org/hiero/block/simulator/config/data/BlockStreamConfig.java
AlexKehayov: In case we want to successfully send an amount of blocks before the block that is supposed to fail.
For example, if midBlockFailOffset==5, 4 blocks will be sent successfully, the fifth will fail.
The default value is 0, so this param is optional. It may be useful in some test scenarios.

simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamGrpcClientImpl.java
AlexKehayov: Just pushed your proposed changes but I kept the midBlockFailOffset check in the first "if".

simulator/src/test/java/org/hiero/block/simulator/grpc/impl/PublishStreamGrpcClientImplTest.java
```diff
@@ -228,22 +217,94 @@ void testStreamBlock_RejectsAfterShutdown() throws InterruptedException {
...
+                    .addItems(blockItemEventTransaction1)
+                    .addItems(blockItemEventTransaction2)
+                    .addItems(blockItemProof)
+                    .build();
+        } else {
+            return Block.newBuilder()
+                    .addItems(blockItemHeader)
+                    .addItems(blockItemProof)
+                    .build();
+        }
```
ata-nas, line 308: General note: we should probably have testFixtures in the common module for stuff like this. We need tests blocks here and all over the plugins as well.

AlexKehayov: Should we make any changes at this point?

ata-nas: No, this is something that should be done only after 505 merges.

georgi-l95 (approved):
LG

ata-nas (approved):
LGTM.

PR #991, opened 2025-04-09
fix: loading historic with clean state

georgi-l95:
## Reviewer Notes
This PR fixes the recently introduced LongRange class to take into consideration the case, when we pass start as -1 and end as -1. This means that we are starting on clean state with no blocks. Otherwise application throws exception and fails to load rest of the plugins.
```
Exception in thread "main" java.lang.IllegalArgumentException: Range start must be non-negative: -1
	at org.hiero.block.node.spi@0.9.0-SNAPSHOT/org.hiero.block.node.spi.historicalblocks.LongRange.<init>(LongRange.java:32)
	at org.hiero.block.node.base@0.9.0-SNAPSHOT/org.hiero.block.node.base.ranges.ConcurrentLongRangeSet.add(ConcurrentLongRangeSet.java:164)
	at org.hiero.block.node.blocks.files.historic@0.9.0-SNAPSHOT/org.hiero.block.node.blocks.files.historic.BlocksFilesHistoricPlugin.init(BlocksFilesHistoricPlugin.java:69)
	at org.hiero.block.node.app@0.9.0-SNAPSHOT/org.hiero.block.node.app.BlockNodeApp.<init>(BlockNodeApp.java:168)
	at org.hiero.block.node.app@0.9.0-SNAPSHOT/org.hiero.block.node.app.BlockNodeApp.main(BlockNodeApp.java:259)
```
## Related Issue(s)

### DIFF
--- a/block-node/block-providers/files.historic/src/main/java/module-info.java
+++ b/block-node/block-providers/files.historic/src/main/java/module-info.java
@@ -11,12 +11,12 @@ module org.hiero.block.node.blocks.files.historic {
             com.swirlds.config.extensions,
             org.hiero.block.node.app;
 
-    requires com.hedera.pbj.runtime;
-    requires org.hiero.block.stream;
     requires transitive com.swirlds.config.api;
     requires transitive org.hiero.block.node.base;
     requires transitive org.hiero.block.node.spi;
+    requires com.hedera.pbj.runtime;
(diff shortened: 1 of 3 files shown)


AlfredoG87 (approved):
LGTM

PR #993, opened 2025-04-09
chore: Fix threading issues in subscriber tests

jsync-swirlds:
### Update subscriber plugin tests to validate transitions from live to historic and back
 * Added the ability to make the test messaging plugin react as if a handler is behind.
 * Added some testing access to the plugin test base and subscriber service plugin.
    * A very similar access for the plugin test base was added upstream
 * Used added features to enable the test code to publish a series of block item events in messaging and mark the subscriber handler "behind" at any desired point.
 * TODO: Add a few tests using the capabilities added.
 * The subscriber sessions are not registered as handlers, so of course get no messages.
    * This is logging and other changes to try to get them registered.
 * Found some issues in testing fixtures and began fixing
 * Added quite a lot of debug tracing to work out where things were not connecting.
 * Adjustments to wait for slowed down subscribers.
 * Latest `505` branch changes caused some regressions
    * All but four test pass now * We have off-by-one due to not handling EndOfStream properly * Transition from historical to live is not working reliably with the latest block provider changes


block-node/subscriber/src/main/java/org/hiero/block/node/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -67,18 +69,20 @@ public enum SubscriberState {
     /** The thread for the historical stream */
     private Thread historicalStreamThread;
     /** The current block being sent to the client */
-    private long currentBlockBeingSent = UNKNOWN_BLOCK_NUMBER;
+    private final AtomicLong currentBlockBeingSent = new AtomicLong(UNKNOWN_BLOCK_NUMBER);
```
jasperpotts, line 72: The intent here was all internal state was accessed from within a synchronized method or block so state should not need to be thread safe.  If that is not true then I am sure there are other problems. The next question is if these need to be atomic are they ok being individually atomic or do they need to be grouped into a single atomic.

jasperpotts: Looks fine other than not 100% sure on this. Happy to accept for now and can think more later

jsync-swirlds: These were accessed from both synchronized and unsynchronized contexts, and it caused some occasional odd results.  That's why I made them atomics.
We might be able to make it work with more synchronized, but that also adds more locks and more opportunities for deadlock, which seemed less desirable as I was trying to remove a deadlock.
I don't think they need to be mutually atomic; both can change independently and be correct.

And yes, there were definitely a few deadlock and inconsistent read cases.  The atomic variables cleaned up _almost_ all of those.  There's still one path somewhere where the code never returns to reading live; I intend to hunt that down in the next PR.

PR #1019, opened 2025-04-15
test: Files Historic Plugin Unit Tests Pt. 3

### LINKED ISSUE #1009: Files Historic Plugin Unit Tests Pt. 3
### Story Form

As a Block Node Developer
I want to have unit tests for the `Files Historic` Plugin
So that I can have confidence in existing logic and to be able to catch regressions

### Technical Notes

- part 3 of unit tests for the `Files Historic` Plugin

ata-nas:
## Reviewer Notes

- part 3 of unit tests for the files historic plugin
- some changes to the compress method for the ZSTD compression type for consistency with the wrapped stream method
- additional tests for the change in the compress method

## Related Issue(s)

Closes #1009

### DIFF
--- a/block-node/base/src/main/java/org/hiero/block/node/base/CompressionType.java
+++ b/block-node/base/src/main/java/org/hiero/block/node/base/CompressionType.java
@@ -5,9 +5,11 @@ import com.github.luben.zstd.Zstd;
 import com.github.luben.zstd.ZstdInputStream;
 import com.github.luben.zstd.ZstdOutputStream;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.UncheckedIOException;
 import java.util.Objects;
 
 /**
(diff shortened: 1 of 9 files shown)


block-node/base/src/main/java/org/hiero/block/node/base/CompressionType.java
```diff
@@ -87,7 +89,24 @@ public OutputStream wrapStream(@NonNull final OutputStream streamToWrap) throws
...
+                        final OutputStream wrappedBaos = wrapStream(baos)) {
+                    wrappedBaos.write(data);
+                    wrappedBaos.close();
+                    // close before return so that we are absolutely sure that
+                    // all data is written including possible footer and others
+                    yield baos.toByteArray();
+                } catch (final IOException e) {
+                    throw new UncheckedIOException(e);
+                }
+            }
```
ata-nas, line 109: I have introduced this change for consistency reasons. It seems that using the static method produces slightly different output than a wrapped stream would, even if using the same compression level. Is it ok to do this? Is there a better way?
cc: @jasperpotts @jsync-swirlds

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
```diff
@@ -133,7 +139,7 @@ public BlockAccessor blockAccessor(long blockNumber) {
      *
      * @return the minimum block number, or -1 if no block files are found
      */
-    public long minStoredBlockNumber() {
+    long minStoredBlockNumber() {
```
ata-nas, line 142: Minor bug on line 163 (GH did not allow me to include int in the preview as I have not made changes there)
```java
...
L:163    return Long.parseLong(fileName.substring(0, fileName.indexOf('.')));
...
```
 (in the max method is the same). We are neglecting the "s" appended at the back of the filename (e.g. "10s.zip") which produces a number format exception.

- Solution 1: handle with fileName.indexOf('s')
- Solution 2: change the BlockPath resolution logic to exclude the "s"

Any opinions/preferences?

cc: @jsync-swirlds @jasperpotts @AlfredoG87 @Nana-EC @georgi-l95

jsync-swirlds: I would remove the `s`.  I'd also remove the extra `0` characters and just have a single digit filename (regardless of grouping). We really shouldn't make reading these files manually "easy"; we don't want operators meddling in the filesystem, so why make it easier?

block-node/base/src/main/java/org/hiero/block/node/base/CompressionType.java
ata-nas: Update: after a discussion, let's revert this. The difference in bytes is probably because the static method call has more context than the streaming approach and this means that it can do a better job at compressing. This will produce different bytes at the end of the day. As far as the consistency concern, we should probably not take it into account as the library _should_ be non deterministic, based on multiple factors. The proper way to assert would be to always uncompress the data and then juxtapose that against the expected.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
ata-nas: Update: after a discussion, let's proceed with solution 1 for now, revisiting this later if we need to.

jasperpotts (approved):

PR #1043, opened 2025-04-17
feat: Implement Cloud Archive Plugin

jasperpotts:
Adds new base functionality of a S3 compatible client lib. Minimal with no dependencies, just want we need. Then a first implementation of cloud archive


block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
```diff
@@ -1,19 +1,278 @@
...
+ * watch persisted block notifications for when the next batch of blocks can be archived. It will then fetch those
+ * blocks from persistence plugins and upload to the archive.
  */
-public class ArchivePlugin implements BlockNodePlugin {
-    // keep track of the last block number archived
-    // have configuration for the number of blocks to archive at a time
-    // on init register as a block notification handler, listen to persisted notifications
-    // on persisted notification, check if the whole batch size of blocks from last block archived is available
-    // if so, fetch them from the persistence plugin, compress and upload to the archive bucket
+public class ArchivePlugin implements BlockNodePlugin, BlockNotificationHandler {
```
Nana-EC, line 40: As structured the `ArchivePlugin` is focused on the cloud bucket archiving.
Should this class rather be more general and the plugin support integration with local archiving and cloud archiving in a staged fashion based on the enabled services?
For example we already have the ZipBlockArchive under the historical block provider, should that run creating local archive files and then the cloud archive service uploads the already created archived files.
In this way the cloud logic is simpler as it's only looking for the compressed archive files and whether it's uploaded the latest, rather than having to track and calculate how many blocks should go into an archive file which is already done.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
```diff
@@ -0,0 +1,609 @@
...
+import javax.crypto.spec.SecretKeySpec;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+
+/**
+ * Simple standalone S3 client for uploading, downloading and listing objects from S3.
+ */
+@SuppressWarnings("JavadocLinkAsPlainText")
+public class S3Client implements AutoCloseable {
```
Nana-EC, line 45: Q: Why in base vs in the archive Plugin module?

Nana-EC (dismissed):
Lots of good progress.
Approving based on structure to enable further rapid development.
We should consider the organization though as we're expanding. 
Should Archive plugin imply cloud archive only or should we have file archive creation logic in one place and support for different move logic i.e. more to a separate dir or move to cloud

block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
Nana-EC: Synced with @jasperpotts and @jsync-swirlds 
This plugin is indeed intentionally focused on cloud archive into S3 API compatible cloud providers.
This is DR focused thus supporting the different approaches in archival file type i.e. tar vs zip, as it allows for more stream like file creation, also background parallel building (expected to be low resource but needs consideration of large chunk size) and other items.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/SimpleTestBlockItemBuilder.java
```diff
@@ -103,20 +134,80 @@ public static BlockItem[] createNumberOfVerySimpleBlocks(final int numberOfBlock
      * @param endBlockNumber the ending block number, inclusive
      * @return an array of BlockItem objects
      */
-    public static BlockItem[] createNumberOfVerySimpleBlocks(final int startBlockNumber, final int endBlockNumber) {
+    public static BlockItem[] createNumberOfVerySimpleBlocks(final long startBlockNumber, final long endBlockNumber) {
```
jsync-swirlds, line 137: What do we gain making these long?

block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
```diff
@@ -1,19 +1,278 @@
...
+    /** The storage class used for uploading the lastest block file */
+    public static final String LATEST_FILE_STORAGE_CLASS = "STANDARD";
```
jsync-swirlds, line 42: There is a storage class config value, do we still need this constant?

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/SimpleTestBlockItemBuilder.java
jasperpotts: It was just that everywhere we have block numbers as longs, otherwise not sure. I think it was causing be to downcast. Guess all tests could be written with ints, I have just always been using long for block numbers

block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
jasperpotts: The reason was I wanted to be to store the latest file at a storage class different from archive files. For example want archives files to be "Glacier" and need the latest file to remain as "Standard" as we want fast read back and short lifespan.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/SimpleTestBlockItemBuilder.java
jsync-swirlds: I see.  It just seemed an unexpected change, particularly given that it worked for all the existing tests as an int.
It's not a major issue, but could have unexpected behavior if someone wanted to test with specific block number ranges that do not fit within the `int` range.

block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
jsync-swirlds: Do we support read-back of the latest?
Just trying to understand how these elements interact, as we'll need to document it to prevent it being incorrectly modified later.

block-node/archive/src/main/java/module-info.java
```diff
@@ -2,10 +2,25 @@
 import org.hiero.block.node.archive.ArchivePlugin;
 
 // SPDX-License-Identifier: Apache-2.0
-module org.hiero.block.node.blocks.cloud.archive {
+module org.hiero.block.node.archive {
```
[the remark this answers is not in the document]
Nana-EC: Agreed, this was one of the points I was getting at with the single general archive module being focused on cloud s3 archive.

ata-nas (dismissed):
LG @jasperpotts, general comments. Looking forward to discuss future improvements for the cloud archive plugin(s).
Based on what also @Nana-EC said, I also approve so that we facilitate faster development.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
```diff
@@ -24,29 +26,59 @@
...
 
-    public PluginTestBase(P plugin, HistoricalBlockFacility historicalBlockFacility) {
+    /**
+     * Start the test fixture with the given plugin, historical block facility, and configuration overrides.
+     *
+     * @param plugin the plugin to be tested
+     * @param historicalBlockFacility the historical block facility to be used
+     * @param configOverrides a map of configuration overrides to be applied to loaded configuration
+     */
+    public void start(P plugin, HistoricalBlockFacility historicalBlockFacility, Map<String, String> configOverrides) {
```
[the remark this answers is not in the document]
jasperpotts: Here is PR https://github.com/hiero-ledger/hiero-consensus-node/pull/18617#pullrequestreview-2785230368

block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
```diff
@@ -1,19 +1,278 @@
...
+        }
+    }
+
+    // ==== BlockItemHandler Methods ===================================================================================
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void handlePersisted(PersistedNotification notification) {
```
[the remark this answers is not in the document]
jasperpotts: The message handler will never be registered if not enabled so this will never get called. So should not need to check enabled on every block.

block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
```diff
@@ -1,19 +1,278 @@
...
+                || archiveConfig.accessKey().isEmpty()) {
+            LOGGER.log(System.Logger.Level.INFO, "Archive plugin is disabled. No access key provided.");
+            return;
+        } else if (archiveConfig.secretKey() == null
+                || archiveConfig.secretKey().isEmpty()) {
+            LOGGER.log(System.Logger.Level.INFO, "Archive plugin is disabled. No secret key provided.");
+            return;
+        }
+        // plugin is enabled
+        enabled.set(true);
```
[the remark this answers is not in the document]
jasperpotts: It seemed reasonable that the plugin is enabled if you give it a config, if not it is disabled and gives warnings.

block-node/archive/src/test/java/org/hiero/block/node/archive/ArchivePluginTest.java
```diff
@@ -0,0 +1,276 @@
...
+        assertTrue(
+                allObjects.contains("blocks/2025/02/2025-02-20_00-00-00_0000000000000000050-0000000000000000059.tar"));
+        assertTrue(
+                allObjects.contains("blocks/2025/03/2025-03-02_00-00-00_0000000000000000060-0000000000000000069.tar"));
+        assertTrue(
+                allObjects.contains("blocks/2025/03/2025-03-12_00-00-00_0000000000000000070-0000000000000000079.tar"));
+        assertTrue(
+                allObjects.contains("blocks/2025/03/2025-03-22_00-00-00_0000000000000000080-0000000000000000089.tar"));
+        assertTrue(
+                allObjects.contains("blocks/2025/04/2025-04-01_00-00-00_0000000000000000090-0000000000000000099.tar"));
```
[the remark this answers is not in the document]
jasperpotts: Sounds reasonable but can always be done later

block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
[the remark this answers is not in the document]
jsync-swirlds: Also, the "canonical" way to disable a plugin is to simply _not install_ that plugin.

block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
jasperpotts: The block node reads the most recent archived block from file is s3 bucket so is maintaining state in the bucket. This has advantage that you can swap block nodes sending to a bucket and one can take over from other. When first starting up the plugin reads latest block from the bucket.

block-node/archive/src/main/java/org/hiero/block/node/archive/ArchivePlugin.java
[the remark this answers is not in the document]
ata-nas: @jsync-swirlds you are right, but we discussed that we will do that sometime in the future for practical reasons.

ata-nas (dismissed):
@jasperpotts LG, let's continue with what we have in order to facilitate faster development, we will iterate over these changes with improvements and we can address also the comments in [this thread](https://github.com/hiero-ledger/hiero-block-node/pull/1043#pullrequestreview-2783757255) as we mentioned there.

block-node/base/src/main/java/org/hiero/block/node/base/s3/XmlBodyHandler.java
```diff
@@ -0,0 +1,61 @@
...
+        try {
+            DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+            factory.setNamespaceAware(true);
+            DocumentBuilder builder = factory.newDocumentBuilder();
+            return builder.parse(inputStream);
+        } catch (IOException e) {
+            throw new UncheckedIOException("Failed to parse XML", e);
+        } catch (ParserConfigurationException | SAXException e) {
+            throw new UncheckedIOException(new IOException("Failed to parse XML", e));
+        }
```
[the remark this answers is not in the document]
ata-nas: Let's iterate on this in the near future.

AlfredoG87 (approved):
LGTM but seems like it needs a rebase.

georgi-l95 (commented):
LG, but needs rebase

AlfredoG87 (approved):

ata-nas (approved):

PR #1055, opened 2025-04-23
fix: hard shutdown of the producer causes exception when another is connected

### LINKED ISSUE #932: Add Publisher Tests
### Story Form

As a Block Node Developer
I want to have enough code coverage on the Publisher Plugin
So that I can confidently use and deploy the application in production environments.

### Technical Notes

.

### LINKED ISSUE #1053: Hard shutdown of the producer causes exception when another is connected.
### Description

Hard shutdown of a producer causes the plugin to not properly remove it from the list of sessions. This cause problems when another is connected and reaches the lastest available block. This causes the newer producer to throw exception and closes the connection. Further more this keeps the plugin in a deadlock, where it denies new producers when they reach the latest available block.

### Steps to Reproduce

1. Run the block-node application
2. Run a publisher for couple of seconds, until it reaches more than one send blocks
3. Perform a hard stop, using IDE UI for example
4. Start a second publisher and observe.

### Additional context

_No response_

georgi-l95:
## Reviewer Notes
This PR resolves a case where, after hard shutdown of a producer the publisher stops accepting new producers.
Also moves around and cleans up the `suites`, by separating utils that can be used everywhere.

## Related Issue(s)
Fixes #1053 
Fixes #932

### DIFF
--- a/block-node/publisher/src/main/java/org/hiero/block/node/publisher/BlockStreamProducerSession.java
+++ b/block-node/publisher/src/main/java/org/hiero/block/node/publisher/BlockStreamProducerSession.java
@@ -12,7 +12,6 @@ import com.hedera.pbj.runtime.ParseException;
 import com.hedera.pbj.runtime.grpc.Pipeline;
 import com.swirlds.metrics.api.Counter;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.SortedSet;
@@ -309,7 +308,7 @@ public final class BlockStreamProducerSession implements Pipeline<List<BlockItem
 
         try {
             responsePipeline.onNext(response);
-        } catch (UncheckedIOException e) {
+        } catch (RuntimeException e) {
             LOGGER.log(
                     WARNING,
                     "Failed to send response to {0}. Disconnecting this session. Error: {1}",
--- a/suites/src/main/java/org/hiero/block/suites/utils/BlockAccessUtils.java
+++ b/suites/src/main/java/org/hiero/block/suites/utils/BlockAccessUtils.java
@@ -25,7 +25,7 @@ public final class BlockAccessUtils {
      * @param latest Whether to retrieve the latest block
      * @return A SingleBlockRequest object
      */
-    public static BlockRequest createSingleBlockRequest(long blockNumber, boolean latest) {
+    public static BlockRequest createGetBlockRequest(long blockNumber, boolean latest) {
         return BlockRequest.newBuilder()
                 .setBlockNumber(blockNumber)
                 .setRetrieveLatest(latest)
@@ -41,7 +41,7 @@ public final class BlockAccessUtils {
      *   verifying its `BlockProof`
      * @return The SingleBlockResponse from the API
      */
-    public static BlockResponse getSingleBlock(
+    public static BlockResponse getBlock(
             @NonNull final BlockAccessServiceGrpc.BlockAccessServiceBlockingStub blockAccessStub,
             final long blockNumber,
             final boolean allowUnverified) {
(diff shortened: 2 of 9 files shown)


block-node/publisher/src/main/java/org/hiero/block/node/publisher/BlockStreamProducerSession.java
```diff
@@ -309,7 +308,7 @@ private void sendResponse(final PublishStreamResponse response) {
 
         try {
             responsePipeline.onNext(response);
-        } catch (UncheckedIOException e) {
+        } catch (RuntimeException e) {
```
georgi-l95, line 311: It seems that when the exception is lower than RuntimeException (UncheckedIO, Socket) we don't catch all cases of failure. Althouth not sure if the whole flow is incorrect, meaning that we should call onNext and be done with it, if any error occurs on the helidon plugin side, it should trigger onError, but that is not the case. Therefore this is the only way (for now) to avoid locking plugin from accepting new producers in this edge case.

AlfredoG87: Yes Indeed seems like a bug on Helidon.
@mattp-swirldslabs complained about this and I believe there is an outstanding Helidon Issue raised about this. 
Was able to find this one, seems related but not 100% sure if is the same thing 🤔 
https://github.com/helidon-io/helidon/issues/9496

AlfredoG87 (approved):
LGTM, let me know when rebased and ready for review so I can re-review (changes) and approve.

suites/src/main/java/org/hiero/block/suites/utils/BlockAccessUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Creates a SingleBlockRequest to retrieve a specific block.
+     *
+     * @param blockNumber The block number to retrieve
+     * @param latest Whether to retrieve the latest block
+     * @return A SingleBlockRequest object
+     */
+    public static BlockRequest createSingleBlockRequest(long blockNumber, boolean latest) {
```
Nana-EC, line 28: Please note renaming on protobuf size and align there
```suggestion
    public static BlockRequest createGetBlockRequest(long blockNumber, boolean latest) {
```

suites/src/main/java/org/hiero/block/suites/utils/BlockAccessUtils.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Retrieves a single block using the Block Node API.
+     *
+     * @param blockNumber The block number to retrieve
+     * @param allowUnverified A flag to indicate that the requested block may be sent without
+     *   verifying its `BlockProof`
+     * @return The SingleBlockResponse from the API
+     */
+    public static BlockResponse getSingleBlock(
```
Nana-EC, line 44:
```suggestion
    public static BlockResponse getBlock(
```

block-node/publisher/src/main/java/org/hiero/block/node/publisher/BlockStreamProducerSession.java
Nana-EC: Nice catch

Nana-EC (commented):
Looking good.
Please note the naming changes. Best to keep naming convention in line with API

suites/src/main/java/org/hiero/block/suites/utils/BlockAccessUtils.java
ata-nas, line 1: +1 for this reusable utility. 👍

ata-nas (dismissed):
LG @georgi-l95

suites/src/main/java/org/hiero/block/suites/utils/BlockAccessUtils.java
georgi-l95: Thanks, changed 👍

suites/src/main/java/org/hiero/block/suites/utils/BlockAccessUtils.java
georgi-l95: Thanks, changed 👍

ata-nas (dismissed):
🔥

AlfredoG87 (approved):
re-approving!

AlfredoG87 (approved):

jsync-swirlds (approved):

Issue #1051, opened 2025-04-21
Enhance/Extend subscriber to feature-match pre-modular version


jsync-swirlds:
### Story Form

As a Block Node Client
I want to subscribe to a block stream that seamlessly merges live and historical blocks.
So that I can process a block stream efficiently and in near-real-time.

### Technical Notes

We have partial support for "live" blocks, but switching between live and historical no longer works correctly in the reworked codebase (after the `505` epic).
This is a rewrite of the subscriber plugin (probably a rename as well) to change the thread model and data flow so that we have stable, consistent, and reliable processing of the client stream that seamlessly moves between live and historical block sources as needed.
The threading model is specifically designed to minimize (or completely avoid) the need for difficult or complicated synchronization during unit tests while remaining correct and lock-free in production.

PR #1058, opened 2025-04-23
feat: Create new stream subscriber plugin

jsync-swirlds:
## Reviewer Notes
Adding a new plugin
 * Instead of beating on the old implementation, implement a new subscriber plugin
   that uses a single client thread and feeds that alternately from messaging or history
   This avoids the countdown latch and state transition logic.
   Also avoids a state machine, just take the next block from one source or the other,
   always getting just the next block to send the client.
   Does require a secondary queue from the messaging, but it's small and simple.
* Adjust the subscriber "method" class to be reused, so we track sessions properly
* Added a latch so the pipeline thread holds until the session is registered with
  messaging before returning. This has minimal impact in production, and makes
  it possible to test reliably.
* Added live block push to queue, live blocks stream correctly.
* Added handling for request ahead of live condition
    * Session just skips over live blocks until live reaches the start block.
* Added proper shutdown of all sessions if the plugin is stopped.
* History and Live both work, and the test disables history to prove it
    * This is necessary because otherwise the test "history" plugin is too fast
      and we always serve history.

### TODO
* Add more unit tests
* Add testing that forces a client to fall behind and catch up
* Add some end-to-end test suites to exercise this plugin better
* Add more and better javadoc
* Clean up the (excessive) trace logging
* Add better metrics, the "transition" metrics aren't super helpful.
* Suggested metrics
    * Live batches sent to client (would be good if we can tag this with a client ID)
    * History blocks sent to client (would be good if we can tag this with a client ID)
    * Count of times the client thread must wait for live blocks, and time spent waiting
    * Count of times the client cannot read from history, but subsequently reads from live
    * Client connection and time from connect to session established

## Related Issue(s)
 Resolves: #1051

### DIFF
--- a/block-node/stream-subscriber/build.gradle.kts
+++ b/block-node/stream-subscriber/build.gradle.kts
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+plugins { id("org.hiero.gradle.module.library") }
+
+description = "Hiero Block Node Subscriber Service"
+
+// Remove the following line to enable all 'javac' lint checks that we have turned on by default
+// and then fix the reported issues.
+tasks.withType<JavaCompile>().configureEach { options.compilerArgs.add("-Xlint:-exports") }
+
+mainModuleInfo {
+    runtimeOnly("com.swirlds.config.impl")
+    runtimeOnly("org.apache.logging.log4j.slf4j2.impl")
+    runtimeOnly("io.helidon.logging.jul")
+    runtimeOnly("com.hedera.pbj.grpc.helidon.config")
+}
+
+testModuleInfo {
+    requires("org.junit.jupiter.api")
+    requires("org.hiero.block.node.app.test.fixtures")
+    requires("org.assertj.core.api")
+}
--- a/block-node/stream-subscriber/src/main/java/module-info.java
+++ b/block-node/stream-subscriber/src/main/java/module-info.java
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
+import org.hiero.block.node.stream.subscriber.SubscriberServicePlugin;
+
+// SPDX-License-Identifier: Apache-2.0
+module org.hiero.block.node.stream.subscriber {
+    uses com.swirlds.config.api.spi.ConfigurationBuilderFactory;
+
+    // export configuration classes to the config module and app
+    exports org.hiero.block.node.stream.subscriber to
+            com.swirlds.config.impl,
+            com.swirlds.config.extensions,
+            org.hiero.block.node.app;
+
+    requires transitive com.hedera.pbj.runtime;
+    requires transitive com.swirlds.config.api;
+    requires transitive org.hiero.block.node.spi;
+    requires transitive org.hiero.block.protobuf;
+    requires com.swirlds.metrics.api;
+    requires org.hiero.block.node.base;
+    requires com.github.spotbugs.annotations;
+
+    provides org.hiero.block.node.spi.BlockNodePlugin with
+            SubscriberServicePlugin;
+}
--- a/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
+++ b/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
@@ -1,0 +1,574 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.stream.subscriber;
+
+import static java.util.concurrent.locks.LockSupport.parkNanos;
+import static org.hiero.block.node.spi.BlockNodePlugin.METRICS_CATEGORY;
+import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
+
+import com.hedera.hapi.block.stream.output.BlockHeader;
+import com.hedera.pbj.runtime.OneOf;
+import com.hedera.pbj.runtime.ParseException;
+import com.hedera.pbj.runtime.grpc.Pipeline;
+import com.swirlds.metrics.api.Counter;
+import com.swirlds.metrics.api.Counter.Config;
+import java.lang.System.Logger;
+import java.lang.System.Logger.Level;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Flow.Subscription;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import org.hiero.block.api.SubscribeStreamRequest;
+import org.hiero.block.api.SubscribeStreamResponse;
+import org.hiero.block.internal.BlockItemSetUnparsed;
+import org.hiero.block.internal.BlockItemUnparsed;
+import org.hiero.block.internal.BlockUnparsed;
+import org.hiero.block.internal.SubscribeStreamResponseUnparsed;
+import org.hiero.block.internal.SubscribeStreamResponseUnparsed.Builder;
+import org.hiero.block.internal.SubscribeStreamResponseUnparsed.ResponseOneOfType;
+import org.hiero.block.node.spi.BlockNodeContext;
+import org.hiero.block.node.spi.BlockNodePlugin;
+import org.hiero.block.node.spi.blockmessaging.BlockItems;
+import org.hiero.block.node.spi.blockmessaging.NoBackPressureBlockItemHandler;
+import org.hiero.block.node.spi.historicalblocks.BlockAccessor;
+
+/**
+ * This class is used to represent a session for a single BlockStream subscriber that has connected to the block node.
+ * At least to start with we will only support receiving a single SubscribeStreamRequest per session.
+ * <p>
+ * This session supports two primary modes of operation: live-streaming and historical streaming. Also switching between
+ * on demand as needed.<p>
+ * #### Threading<br/>
+ * This class is called from many threads from web server, the block messaging system and its own background historical
+ * fetching thread. To make its state thread safe it uses synchronized methods around all entry points from other
+ * threads (which are minimized to avoid potential for deadlock or contention). Currently only the close method is
+ * synchronized.
+ */
+public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscriberSession> {
+    /** The logger for this class. */
+    private final Logger LOGGER = System.getLogger(getClass().getName());
+
+    /** The maximum time units to wait for a live block to be available */
+    private static final long MAX_LIVE_POLL_DELAY = 500;
+    /** The time unit for the maximum live poll delay */
+    private static final TimeUnit LIVE_POLL_UNITS = TimeUnit.MILLISECONDS;
+    /** The "resolution", in ns, to use for a polling delays. */
+    private static final long PARK_DELAY_TIME = 100_000;
+
+    /** The client id for this session */
+    private final long clientId;
+    /** The first block number to stream */
+    private final long startBlockNumber;
+    /** The last block number to stream, can be {@value BlockNodePlugin#UNKNOWN_BLOCK_NUMBER } to mean infinite */
+    private final long endBlockNumber;
+    /** The pipeline to send responses to the client */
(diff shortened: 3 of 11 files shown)


block-node/stream-subscriber/src/main/java/module-info.java
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+import org.hiero.block.node.stream.subscriber.SubscriberServicePlugin;
+
+// SPDX-License-Identifier: Apache-2.0
+module org.hiero.block.node.stream.subscriber {
+    uses com.swirlds.config.api.spi.ConfigurationBuilderFactory;
+
```
georgi-l95, line 7:
```suggestion
    // export configuration classes to the config module and app
    exports org.hiero.block.node.stream.subscriber to
            com.swirlds.config.impl,
            com.swirlds.config.extensions,
            org.hiero.block.node.app;

```

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -0,0 +1,255 @@
...
+        clientHandler.stop();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @NonNull
+    public String serviceName() {
+        return "BlockStreamService";
```
georgi-l95, line 89:
```suggestion
        String[] parts = fullName().split("\\.");
        return parts[parts.length - 1];
```

Otherwise it fails with 👇 , when you try to connect consumer

```
2025-04-25 16:12:39.973+0300 SEVERE  [org.hiero.block.simulator.grpc.impl.ConsumerStreamObserver onError] Error io.grpc.StatusRuntimeException: UNKNOWN: HTTP status code 200
invalid content-type: null
trailers: Metadata(:status=200,grpc-status=5) with status Status{code=UNKNOWN, description=HTTP status code 200
invalid content-type: null
trailers: Metadata(:status=200,grpc-status=5), cause=null}.
io.grpc.StatusRuntimeException: UNKNOWN: HTTP status code 200
invalid content-type: null
trailers: Metadata(:status=200,grpc-status=5)
```
Change fixes this, it was discovered in this [PR](https://github.com/hiero-ledger/hiero-block-node/pull/891)

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @NonNull
+    public String fullName() {
+        return "com.hedera.hapi.block." + serviceName();
```
georgi-l95, line 98:
```suggestion
        return BlockStreamSubscribeServiceGrpc.SERVICE_NAME;
```


Otherwise it fails with 👇 , when you try to connect consumer

```
2025-04-25 16:12:39.973+0300 SEVERE  [org.hiero.block.simulator.grpc.impl.ConsumerStreamObserver onError] Error io.grpc.StatusRuntimeException: UNKNOWN: HTTP status code 200
invalid content-type: null
trailers: Metadata(:status=200,grpc-status=5) with status Status{code=UNKNOWN, description=HTTP status code 200
invalid content-type: null
trailers: Metadata(:status=200,grpc-status=5), cause=null}.
io.grpc.StatusRuntimeException: UNKNOWN: HTTP status code 200
invalid content-type: null
trailers: Metadata(:status=200,grpc-status=5)
```
Change fixes this, it was discovered in this [PR](https://github.com/hiero-ledger/hiero-block-node/pull/891)

georgi-l95 (commented):
Also we'd need to replace in `block-node/app/build.gradle.kts`:
`    runtimeOnly("org.hiero.block.node.subscriber")`
with: 
`    runtimeOnly("org.hiero.block.node.stream.subscriber")`

Looking good, done some manual testing, which will be transformed into E2E tests. Those included:
- Live streaming - Send request with -1 start block to -1 end block. ✅ 
- Historic streaming - Send request with 0 start block to 10 end block. ✅ 
- Historic -> Live -> Historic -> Live streaming - Send request with 0 start block to -1 end block. We started receiving blocks, then (with some modifications on simulator side) waited couple of seconds to get behind, then the plugin switched us to historic and when we caught up, we switched to live. ✅ 

Things left to try:
- Live streaming on two or more consumers
- Historic streaming on two or more consumers
- Historic -> Live -> Historic -> Live streaming on two or more consumers

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
jsync-swirlds: Done.

block-node/stream-subscriber/src/main/java/module-info.java
jsync-swirlds: Done.

georgi-l95 (approved):

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberConfig.java
```diff
@@ -0,0 +1,32 @@
...
+ *     that the session will try to maintain.  If there is less than this much
+ *     capacity available, the session will drop the oldest full blocks (at the queue head)
+ *     in the queue until at least this many batches can be added without blocking.<br/>
+ *     This value should typically be around 10% of the live queue size.
+ */
+@ConfigData("subscriber")
+public record SubscriberConfig(
+        @Loggable @ConfigProperty(defaultValue = "4000") @Min(100) int liveQueueSize,
+        @Loggable @ConfigProperty(defaultValue = "4000") @Min(10) long maximumFutureRequest,
+        @Loggable @ConfigProperty(defaultValue = "400") @Min(10) int minimumLiveQueueCapacity) {}
```
ata-nas, line 32: Should we not have max values also? Also we could utilize the preconditions require in range here in a follow up.

ata-nas (approved):
@jsync-swirlds looks good! Looking forward to see the improvements done based on the todo. 🔥

Nana-EC (commented):
Nice work.
Made it a 1/3 of the way through and will circle back

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberConfig.java
jsync-swirlds: These don't really need maximums.  The maximum is whatever resources we wish to allocate, which is not something we really need to constrain.  The minimums set here are just to ensure we don't end up with a system that fails because it lacks enough room in the queue for one block, or denies requests unnecessarily.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -0,0 +1,574 @@
...
+        this.responsePipeline = Objects.requireNonNull(responsePipeline);
+        this.context = Objects.requireNonNull(context);
+        this.sessionReadyLatch = Objects.requireNonNull(sessionReadyLatch);
+        latestLiveStreamBlock = new AtomicLong(UNKNOWN_BLOCK_NUMBER - 1);
+        pluginConfiguration = context.configuration().getConfigData(SubscriberConfig.class);
+        // Next block to send depends on what was requested and what is available.
+        nextBlockToSend = startBlockNumber < 0 ? getLatestKnownBlock() : startBlockNumber;
+        handlerName = "Live stream client " + clientId;
+        // create metrics
+        // Note: These two may, or may not, be useful...
```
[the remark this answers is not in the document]
jsync-swirlds: A few points:

1. Transition to history has nothing to do with performance of the _server_, it happens when the _client_ is slow.
2. These aren't currently set anywhere because there isn't a "clean" transition. There is no place in the code where we can say "this was history, now it's live" or "this was live, now it's history". Regardless of source the code just tries to send the next available block and is "greedy" in terms of sending as many from the current source as possible.
3. Recording history blocks sent, live blocks sent, and time to send by batch is probably more useful if we want to track performance (though even that's going to have a ton of noise; the most a "slow" would say is that the _client_ is slow, not the server, server performance is more likely measured at the messaging facility).

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -0,0 +1,574 @@
...
+     * <p>
+     * This is true if all of the following are true:
+     *     1. The stream is not interrupted
+     *     2. We have not run past the latest live block
+     *     3. All requested blocks have not been sent
+     *     4. The next block to send is not available from the live stream.
+     */
+    private boolean isHistoryPermitted() {
+        return !(interruptedStream.get()
+                || latestLiveStreamBlock.get() < nextBlockToSend
```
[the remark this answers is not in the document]
jsync-swirlds: That check is two lines later and is different from what we check on this line.

This line checks whether we're out in front of live (which can happen in some situations), not whether live has the next block (which wouldn't return true when this does).

The javadoc list above matches the order of the lines here.

AlfredoG87 (approved):
LGTM

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -0,0 +1,574 @@
...
+    /** The pipeline to send responses to the client */
+    private final Pipeline<? super SubscribeStreamResponseUnparsed> responsePipeline;
+    /** The context for the block node */
+    private final BlockNodeContext context;
+    /** The configuration for the "owning" plugin */
+    private final SubscriberConfig pluginConfiguration;
+    /** The number of historic to live stream transitions metric */
+    private final Counter historicToLiveStreamTransitions;
+    /** The number of live to historic stream transitions metric */
+    private final Counter liveToHistoricStreamTransitions;
```
AlfredoG87, line 78: This metrics are declared but then are not used, .increase()

AlfredoG87 (commented):
Just a nit.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
AlfredoG87: Also we are missing another Metric of Type `Gauge` **numberOfSubscribers**

AlfredoG87: As a matter of fact, these are the mestric so far that we have identified with @jasperpotts that need to be created for the subscriiber plugin:
<google-sheets-html-origin><style type="text/css"></style>
subscriber_historic_to_live | Count of historic to live stream transitions
-- | --
subscriber_live_to_historic | Count of live to historic stream transitions
subscriber_open_connections | No of Connected Subscribers
subscriber_newest_block_sent | The newest block sent to any subscriber, if this is far from publisher_current_block_number_inbound we have issues
subscriber_errors | Count of errors while streaming to subscribers

jsync-swirlds: As I detailed in the review notes, the metrics are a TODO item, and there are several to add.

There is some difficulty with some metrics in that several of those (like newest sent) need to be specific to each subscriber, do we have tag support (or similar) so we can display per-subscriber in dashboards (without creating thousands of individual metric streams)?

I have notes in code and above about the "transition" metrics; I don't think they're that useful, nor really applicable in this modified approach; but that's certainly something to consider in follow-up PR(s).

block-node/stream-subscriber/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+plugins { id("org.hiero.gradle.module.library") }
+
+description = "Hiero Block Node Subscriber Service"
```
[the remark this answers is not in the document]
jsync-swirlds: Will make this change in follow-up PR(s), if that's OK.

jasperpotts:
I want to go through in detail but don't want to hold up as I will not get a chance till after devcon. So can merge and I will review later.

PR #1064, opened 2025-04-24
test: Files Historic Plugin Unit Tests Pt. 5

### LINKED ISSUE #1039: Files Historic Plugin Unit Tests Pt. 5
### Story Form

As a Block Node Developer
I want to have unit tests for the `Files Historic` Plugin
So that I can have confidence in existing logic and catch regression

### Technical Notes

- part 5 of unit tests for the `Files Historic` Plugin

ata-nas:
## Reviewer Notes

- Part 5 of tests for the `Files Historic` Plugin.
- We test the actual plugin here.
- Introduced a new blocking serial executor service for testing purposes that will allow us to serially execute tasks in the order they have been submitted, all on the caller thread, this ensure that tests will always be in assertable state and debugging is a breeze.
- Found a bug when merging a range in a ranged set, we allow the `LongRange` record to have -1 for start and end, but the merging logic does not allow that, which produces a runtime exception when we initialize a ranged set with `-1 -> -1` and then we try to add anything to it, for example if we add `0 - 9` to it, the ranged set will attempt to produce the following values `-1 -> 9` which the constructor of the `LongRange` record does not allow. For now, I have handled the ranged set in the plugin, the available ranges to only insert anything if we have found some values initially when starting up the server. Otherwise the first insert will be after the first successful zip.
- Another bug was in the `start()` where the range was not correctly established
  - I have extracted and reused the zipping logic from the notification handling method
  - A concern here is again "where are the blocks coming from?"
  - I think we might need to extend the historical block facility method to provide us with accessors from a provider with some priority filter, but I am not sure as of yet if we really need that + if that will over complicate things, potentially hanging some places to wait for accessors that may never come but are available in other places. What are the tradeoffs?
- Am I missing some good cases that I can add as tests?

## Related Issue(s)

Closes #1039

### DIFF
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
@@ -1,0 +1,125 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.app.fixtures.async;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A very simple executor to be used only for testing! This executor will
+ * collect all submitted tasks to it and will hold them in order. After that
+ * when the {@link #executeSerially()} method is called, all tasks will be
+ * executed in order serially on the same thread that called the method. This
+ * will ensure that all tasks will complete before doing any asserts
+ */
+public class BlockingSerialExecutor extends ThreadPoolExecutor {
+    /** The work queue that will be used to hold the tasks */
+    private final BlockingQueue<Runnable> workQueue;
+
+    /**
+     * Constructor.
+     */
+    public BlockingSerialExecutor(@NonNull final BlockingQueue<Runnable> workQueue) {
+        // supply super with arbitrary values, they will not be used
+        super(
+                1,
+                1,
+                0L,
+                TimeUnit.MILLISECONDS,
+                Objects.requireNonNull(workQueue), // super will not use this queue
+                Executors.defaultThreadFactory(),
+                new AbortPolicy());
+        this.workQueue = workQueue; // actual work queue
+    }
+
+    /**
+     * Overriding this method in order to make sure this logic will be called
+     * every time we will submit a task to the pool.
+     */
+    @Override
+    @SuppressWarnings("all")
+    public void execute(@NonNull final Runnable command) {
+        workQueue.offer(command);
+    }
+
+    /**
+     * Invoke this method once you have submitted all the tasks that need to be
+     * run by this executor. This method will then proceed to poll each task
+     * (in the same order that they have been submitted) and will execute the
+     * tasks serially on the same thread that called this method. Once this
+     * method returns (or throws), we are certain that all tasks have run and
+     * can safely assert based on that. This method will throw an
+     * {@link IllegalStateException} to indicate a broken state, when the queue
+     * is empty.
+     */
+    public void executeSerially() {
+        if (workQueue.isEmpty()) {
+            throw new IllegalStateException("Queue is empty");
+        } else {
+            while (!workQueue.isEmpty()) {
+                workQueue.poll().run();
+            }
+        }
+    }
+
+    @Override
+    public void close() {
+        shutdownNow();
+    }
+
+    @Override
+    public void shutdown() {
+        shutdownNow();
+    }
+
+    @Override
+    public boolean awaitTermination(final long timeout, final TimeUnit unit) {
+        shutdownNow();
+        return true;
+    }
+
+    /**
+     * Operation currently not supported!
+     */
+    @Override
+    @NonNull
+    public <T> T invokeAny(@NonNull final Collection<? extends Callable<T>> tasks) {
+        throw new UnsupportedOperationException("Operation not supported, to be extended as needed");
+    }
+
+    /**
+     * Operation currently not supported!
+     */
+    @Override
+    @NonNull
+    public <T> T invokeAny(
+            @NonNull final Collection<? extends Callable<T>> tasks, final long timeout, @NonNull final TimeUnit unit) {
+        throw new UnsupportedOperationException("Operation not supported, to be extended as needed");
+    }
+
+    /**
+     * Operation currently not supported!
+     */
+    @Override
+    @NonNull
+    public <T> List<Future<T>> invokeAll(@NonNull final Collection<? extends Callable<T>> tasks) {
+        throw new UnsupportedOperationException("Operation not supported, to be extended as needed");
+    }
+
+    /**
+     * Operation currently not supported!
+     */
+    @Override
+    @NonNull
+    public <T> List<Future<T>> invokeAll(
+            @NonNull final Collection<? extends Callable<T>> tasks, final long timeout, @NonNull final TimeUnit unit) {
+        throw new UnsupportedOperationException("Operation not supported, to be extended as needed");
+    }
+}
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
@@ -20,8 +20,10 @@ import java.util.concurrent.TimeUnit;
  * will ensure that all tasks will complete before doing any asserts
  */
 public class BlockingSerialExecutor extends ThreadPoolExecutor {
-    /** The work queue that will be used to hold the tasks */
+    /** The work queue that will be used to hold the tasks. */
     private final BlockingQueue<Runnable> workQueue;
+    /** Counter to indicate total submitted tasks. */
+    private int tasksSubmitted;
 
     /**
      * Constructor.
@@ -47,6 +49,7 @@ public class BlockingSerialExecutor extends ThreadPoolExecutor {
     @SuppressWarnings("all")
     public void execute(@NonNull final Runnable command) {
         workQueue.offer(command);
+        tasksSubmitted++;
     }
 
     /**
@@ -67,6 +70,24 @@ public class BlockingSerialExecutor extends ThreadPoolExecutor {
                 workQueue.poll().run();
             }
         }
+    }
+
+    /**
+     * This method indicates if any task was ever submitted to this executor.
+     * This is useful during tests in order to assert that the pool was
+     * essentially not interacted with. An example would be if we have a test
+     * where we want to assert that some production logic will never submit a
+     * task to the executor given some condition, then we can use this method
+     * to assert that. This method does not reflect the current state of the
+     * queue, meaning the queue might be empty due to a call to the
+     * {@link #executeSerially()} method, but this method will still return
+     * true if any task was submitted before that.
+     *
+     * @return boolean value, true if any task was ever submitted, false
+     * otherwise
+     */
+    public boolean wasAnyTaskSubmitted() {
+        return tasksSubmitted > 0;
     }
 
     @Override
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/NoOpServiceBuilder.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/NoOpServiceBuilder.java
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.app.fixtures.plugintest;
+
+import com.hedera.pbj.runtime.grpc.ServiceInterface;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.helidon.webserver.http.HttpService;
+import org.hiero.block.node.spi.ServiceBuilder;
+
+/**
+ * A simple no-op implementation of {@link ServiceBuilder} that does nothing.
+ * To be used for testing purposes only where we need a non-null implementation
+ * that we do not want to act upon.
+ */
+public final class NoOpServiceBuilder implements ServiceBuilder {
+    /** No-op implementation, does nothing. */
+    @Override
+    public void registerHttpService(final String path, final HttpService... service) {}
+
+    /** No-op implementation, does nothing. */
+    @Override
+    public void registerGrpcService(@NonNull final ServiceInterface service) {}
+}
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/NoOpServiceBuilder.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/NoOpServiceBuilder.java
@@ -12,11 +12,19 @@ import org.hiero.block.node.spi.ServiceBuilder;
  * that we do not want to act upon.
  */
 public final class NoOpServiceBuilder implements ServiceBuilder {
-    /** No-op implementation, does nothing. */
+    /**
+     * No-op implementation, does nothing.
+     */
     @Override
-    public void registerHttpService(final String path, final HttpService... service) {}
+    public void registerHttpService(final String path, final HttpService... service) {
+        // No-op implementation, does nothing.
+    }
 
-    /** No-op implementation, does nothing. */
+    /**
+     * No-op implementation, does nothing.
+     */
     @Override
-    public void registerGrpcService(@NonNull final ServiceInterface service) {}
+    public void registerGrpcService(@NonNull final ServiceInterface service) {
+        // No-op implementation, does nothing.
+    }
 }
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/SimpleInMemoryHistoricalBlockFacility.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/SimpleInMemoryHistoricalBlockFacility.java
@@ -43,11 +43,16 @@ public class SimpleInMemoryHistoricalBlockFacility implements HistoricalBlockFac
      */
     @Override
     public void handleBlockItemsReceived(BlockItems blockItems) {
+        handleBlockItemsReceived(blockItems, true);
+    }
+
+    public void handleBlockItemsReceived(BlockItems blockItems, final boolean sendNotification) {
         if (!disablePlugin.get()) {
             if (blockItems.isStartOfNewBlock()) {
                 if (!partialBlock.isEmpty()) {
                     throw new RuntimeException(
-                            "Something went wrong, partitionedBlock is not empty. So we never got a end block for current block");
+                            "Something went wrong, partitionedBlock is not empty. So we never got a end "
+                                    + "block for current block");
                 }
                 currentBlockNumber.set(blockItems.newBlockNumber());
             }
@@ -63,9 +68,11 @@ public class SimpleInMemoryHistoricalBlockFacility implements HistoricalBlockFac
                 availableBlocks.add(blockNumber);
                 partialBlock.clear();
                 // send block persisted message
-                blockNodeContext
-                        .blockMessaging()
-                        .sendBlockPersisted(new PersistedNotification(blockNumber, blockNumber, 2000));
+                if (sendNotification) {
+                    blockNodeContext
+                            .blockMessaging()
+                            .sendBlockPersisted(new PersistedNotification(blockNumber, blockNumber, 2000));
+                }
             }
         }
     }
--- a/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
+++ b/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
@@ -71,8 +71,8 @@ public final class BlocksFilesHistoricPlugin implements BlockProviderPlugin, Blo
      * {@inheritDoc}
      */
     @Override
-    public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
-        this.context = context;
+    public void init(final BlockNodeContext context, final ServiceBuilder serviceBuilder) {
+        this.context = Objects.requireNonNull(context);
         final FilesHistoricConfig localConfig =
                 this.config == null ? context.configuration().getConfigData(FilesHistoricConfig.class) : this.config;
         // create plugin data root directory if it does not exist
--- a/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
+++ b/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
@@ -220,26 +220,29 @@ public final class BlocksFilesHistoricPlugin implements BlockProviderPlugin, Blo
                     batchFirstBlockNumber,
                     batchLastBlockNumber);
             zipBlockArchive.writeNewZipFile(batchFirstBlockNumber);
-            // update the first and last block numbers
-            availableBlocks.add(batchFirstBlockNumber, batchLastBlockNumber);
-            // log done
-            LOGGER.log(
-                    System.Logger.Level.INFO,
-                    "Moved batch of blocks[%d -> %d] to zip file",
-                    batchFirstBlockNumber,
-                    batchLastBlockNumber);
-            // now all the blocks are in the zip file and accessible, send notification
-            context.blockMessaging()
-                    .sendBlockPersisted(
-                            new PersistedNotification(batchFirstBlockNumber, batchLastBlockNumber, defaultPriority()));
-            // remove the batch of blocks from in progress ranges
-            inProgressZipRanges.remove(batchRange);
-        } catch (final Exception e) {
+        } catch (final IOException e) {
             LOGGER.log(
                     System.Logger.Level.ERROR,
                     "Failed to move batch of blocks[" + batchFirstBlockNumber + " -> " + batchLastBlockNumber
                             + "] to zip file",
                     e);
+            return;
         }
+        // if we have reached here, then the batch of blocks has been zipped,
+        // now we need to make some updates
+        // update the first and last block numbers
+        availableBlocks.add(batchFirstBlockNumber, batchLastBlockNumber);
+        // log done
+        LOGGER.log(
+                System.Logger.Level.INFO,
+                "Moved batch of blocks[%d -> %d] to zip file",
+                batchFirstBlockNumber,
+                batchLastBlockNumber);
+        // now all the blocks are in the zip file and accessible, send notification
+        context.blockMessaging()
+                .sendBlockPersisted(
+                        new PersistedNotification(batchFirstBlockNumber, batchLastBlockNumber, defaultPriority()));
+        // remove the batch of blocks from in progress ranges
+        inProgressZipRanges.remove(batchRange);
     }
 }
--- a/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
+++ b/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
@@ -6,6 +6,7 @@ import java.io.IOException;
 import java.lang.System.Logger.Level;
 import java.nio.file.Files;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -23,11 +24,11 @@ import org.hiero.block.node.spi.historicalblocks.LongRange;
  * This plugin provides a block provider that stores historical blocks in file. It is designed to store them in the
  * most compressed optimal way possible. It is designed to be used with the
  */
-public class BlocksFilesHistoricPlugin implements BlockProviderPlugin, BlockNotificationHandler {
+public final class BlocksFilesHistoricPlugin implements BlockProviderPlugin, BlockNotificationHandler {
     /** The logger for this class. */
     private final System.Logger LOGGER = System.getLogger(getClass().getName());
     /** The executor service for moving blocks to zip files in a background thread. */
-    private final ExecutorService zipMoveExecutorService = Executors.newSingleThreadExecutor();
+    private final ExecutorService zipMoveExecutorService;
     /** The block node context. */
     private BlockNodeContext context;
     /** The zip block archive. */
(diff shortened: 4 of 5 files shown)


block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
ata-nas, line 1: I have re-introduced this. We had a blocking executor in the past that worked similarly to this one, but this one is actually serial and is called on the same thread. This means that this is way easier to debug and also allows us to do some neat asserts. Calling the `executeSerially()` ensures that after we pop from it, all tasks have run which means that asserts in the tests can be safely done every time (removing any potential flakiness), but the trade off is that running tasks serially is not the same as running them async but waiting (blocking) until all complete (which was a past implementation of this one). This **should be the better choice** here and also for other places, but in places where we have high concurrency and things are very sensitive, we may want to introduce the async runner blocking pool again.

ata-nas (commented):
Self review to help other reviewers with the PR.

block-node/block-providers/files.historic/src/test/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPluginTest.java
```diff
@@ -0,0 +1,647 @@
...
+            // call the start method, we expect that it will queue a new task
+            // that we can execute
+            toTest.start();
+            // execute serially to ensure all tasks are completed
+            pluginExecutor.executeSerially();
+            // assert that the first 10 blocks are zipped now
+            for (int i = 0; i < 10; i++) {
+                assertThat(BlockPath.computeExistingBlockPath(testConfig, i)).isNotNull();
+            }
+        }
```
ata-nas, line 645: The `start()` reuses the same logic as the notification handler method. I have added only the happy path here, I see no value in having all the tests done again, but called from the start method. This should be sufficient.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -38,6 +39,22 @@ public class BlocksFilesHistoricPlugin implements BlockProviderPlugin, BlockNoti
...
+    /** Constructor, used for normal plugin loading */
+    public BlocksFilesHistoricPlugin() {
+        this.zipMoveExecutorService = Executors.newSingleThreadExecutor();
+    }
+
+    /** Constructor, used only for testing temporarily while we wait for
+     * extended functionality to support overriding config converters
+     * PR #18617 in hiero-consensus-node.
+     */
+    BlocksFilesHistoricPlugin(final FilesHistoricConfig config, final ExecutorService zipMoveExecutorService) {
```
jasperpotts, line 54: I don't love adding API/Code to production for testing. This burnt us in the last week on consensus node as we could not do something because of the way a class was using the code. This was a huge problem and after many hours investigation we found the only reason the class existed still that was using that code, was because it was used by a for test only method. So we almost said we could not deliver a feature because it was impossible, just because of an old for test method that should have gone away.

I like your executor service swap. For config I added ability to override config for plugin test classes. I wonder if we should have a createExecutorService(int numThreads, String namePrefix) to BlockNodeContext. That way we can add central tracking, startup logging of number of threads created and by who. We have issue in consensus node that we create executors all over the place and do not track how many threads are created or by who. We have a plan to add ThreadManager one day there but not got to it today. If we added that to context then in plugin test it can swap out the executor in test fixture and require no special code in production plugin. Thoughts?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -56,42 +73,41 @@ public List<Class<? extends Record>> configDataTypes() {
...
+            // we never expect to enter here, if we do, we have an issue that
+            // needs to be investigated
+            throw new IllegalStateException(
+                    "Latest zipped block number [%d] cannot be greater than the first zipped block number [%d]"
+                            .formatted(latestZippedBlock, firstZippedBlock));
+        }
+        if (firstZippedBlock >= 0) {
+            // add the blocks to the available blocks only if the range is a valid one (positive)
+            availableBlocks.add(firstZippedBlock, latestZippedBlock);
+        }
```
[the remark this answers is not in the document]
jasperpotts: This brings up a big design question. That would be huge to change but worth asking the question. That is:

inclusive vs exclusive for all block number ranges in block node?

So far it is "inclusive" everywhere which is fine other than it doesn't allow for an empty range as 0->0 includes 0 rather than being empty. If we move to exclusive then we could avoid `-1`s to have empty. This would be a horrible change to code base but easier now than later. As code base will only get bigger 😃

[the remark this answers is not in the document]
ata-nas: I think maybe this is simpler. The way I see it that currently it is fine and behaves as expected. Inclusive is I think the right way to go (as it is now). So fore instance as you mention `0 -> 0` will return true only if we query for a 0. To me the range set class is empty when it has no `LongRange` instance internally in it. In fact this is why the code change is working. When the ranged set is instantiated initially, it is empty and retruns false on everything for the sole reason that there are no `LongRange` instances inside it. So, for an empty ranged set, can we simply not have a set that has no `LongRange` instances inside of it (this is an implementation detail for the `ConcurrentLongRangeSet` we currently use). Am I missing something?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
ata-nas: If by adding code you meant the constructor, I completely agree. We were waiting on the changes (still waiting I think) for the custom converters. But in this case I am using the jUnit temp dir because jimfs does not support the `File` abstraction unfortunately and it is used for the zips :/ But yeah, looks like you did introduce some new logic with latest merge to main, I will check it out and attempt to pass the path through there.

As far as the executor service factory method, I really like that as well. It will neatly fit into the testing scenarios as well, since we will not have to pass test instance of executors through constructors and introduce "test only" logic in production like it happened now. So if we make the change you suggested, we can rework this as well. Shall I attempt that tomorrow or would you like to take a look into that factory method?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -190,26 +216,29 @@ private void moveBatchOfBlocksToZipFile(final LongRange batchRange) {
...
+                System.Logger.Level.INFO,
+                "Moved batch of blocks[%d -> %d] to zip file",
+                batchFirstBlockNumber,
+                batchLastBlockNumber);
+        // now all the blocks are in the zip file and accessible, send notification
+        context.blockMessaging()
+                .sendBlockPersisted(
+                        new PersistedNotification(batchFirstBlockNumber, batchLastBlockNumber, defaultPriority()));
+        // remove the batch of blocks from in progress ranges
+        inProgressZipRanges.remove(batchRange);
```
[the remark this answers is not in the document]
ata-nas: P.S. do we want to catch any exceptions produced by the calls here? If so, we can revert, but we need to make sure that in the catch, we also leave the plugin in a proper state, reverting available blocks changes, not removing in progress ranges etc etc

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
Nana-EC: I also like the factory within the context idea to centralize thread management.
@ata-nas let's open a ticket for that and add it to the sprint backlog and verification of refactor code EPIC.
If it becomes required you can pick it up next week otherwise we can get to it soon.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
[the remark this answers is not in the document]
Nana-EC: I'd also prefer inclusive. 
Although I'm not a fan of the support for start = end in the range class without practical usage.
What's the value of an empty range?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -133,24 +149,34 @@ public BlockRangeSet availableBlocks() {
...
-                minBlockNumber += numberOfBlocksPerZipFile;
-                maxBlockNumber += numberOfBlocksPerZipFile;
-            }
-        }
+            attemptZipping();
+        } // todo this is not enough of an assertion that the blocks will be coming from the right place
+        //     as notifications are async and things can happen, when we get the accessors later, we should
+        //     be able to get accessors only from places that have higher priority than us. We should probably
+        //     have that as a feature in the block accessor api. (meaning we should be able to query the
+        //     historical block facility for blocks that are coming from higher priority plugins)
```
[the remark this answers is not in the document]
Nana-EC: Definitely something to watch.
I believe right now priority may not be accurately set on new notifications. 
I see defaultPriority set but not much beyond that. Need to look more

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
ata-nas: @Nana-EC as requested, I have created the issue: #1067

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
jsync-swirlds: One small (possible) improvement would be to have the serial executor use its own thread to execute all the tasks, and when you call `executeSerially` you `join` that thread instead of executing everything on the original thread.  That helps flesh out more possible asynchronous bugs and also avoids some very-hard-to-diagnose deadlock opportunities.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Operation currently not supported!
+     */
+    @Override
+    @NonNull
+    public <T> List<Future<T>> invokeAll(
+            @NonNull final Collection<? extends Callable<T>> tasks, final long timeout, @NonNull final TimeUnit unit) {
+        throw new UnsupportedOperationException("Operation not supported, to be extended as needed");
+    }
```
jsync-swirlds, line 145: Just a thought.
Instead of the custom `executeSerially` method, we could use these two as follows:
1. build a list from the work queue
    * Also create an empty "results" list.
2. for each item in the list, execute that item and put it's (completed) Future (e.g. `list.add(new CompletedFuture(task.call())`) in the results list.
3. Return the results list.  As every Future is already completed in that list, the results are immediately available.

This is not necessarily appropriate for this PR, but perhaps worth documenting for follow-up.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/SimpleInMemoryHistoricalBlockFacility.java
```diff
@@ -61,9 +65,11 @@ public void handleBlockItemsReceived(BlockItems blockItems) {
             availableBlocks.add(blockNumber);
             partialBlock.clear();
             // send block persisted message
-            blockNodeContext
-                    .blockMessaging()
-                    .sendBlockPersisted(new PersistedNotification(blockNumber, blockNumber, 2000));
+            if (sendNotification) {
+                blockNodeContext
+                        .blockMessaging()
```
jsync-swirlds, line 70: What is the magic number `2000` for?  Should that be a constant or configuration item?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -38,6 +39,22 @@ public class BlocksFilesHistoricPlugin implements BlockProviderPlugin, BlockNoti
     private final ConcurrentLongRangeSet availableBlocks = new ConcurrentLongRangeSet();
     /** List of all zip ranges that are in progress, so we do not start a duplicate job. */
     private final CopyOnWriteArrayList<LongRange> inProgressZipRanges = new CopyOnWriteArrayList<>();
+    /** Config used internally. Should come from the context, this is temporary until we have some needed logic. */
+    private FilesHistoricConfig config;
+
+    /** Constructor, used for normal plugin loading */
+    public BlocksFilesHistoricPlugin() {
+        this.zipMoveExecutorService = Executors.newSingleThreadExecutor();
```
jsync-swirlds, line 47: This definitely should be created via Context or some other provided factory...
Ideally we wouldn't be creating single thread executors in general, and would use task pools, optionally with virtual threads.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -190,26 +216,29 @@ private void moveBatchOfBlocksToZipFile(final LongRange batchRange) {
...
-            // now all the blocks are in the zip file and accessible, send notification
-            context.blockMessaging()
-                    .sendBlockPersisted(
-                            new PersistedNotification(batchFirstBlockNumber, batchLastBlockNumber, defaultPriority()));
-            // remove the batch of blocks from in progress ranges
-            inProgressZipRanges.remove(batchRange);
-        } catch (Exception e) {
+        } catch (final IOException e) {
             LOGGER.log(
                     System.Logger.Level.ERROR,
```
jsync-swirlds, line 221: This raises a question that the team should probably discuss.
the `ERROR` level is usually a "wake the team" log level, and unless the block node is in a failure state, we probably shouldn't use it.
I would recommend `WARNING` here (as this is serious, but not fatal); but I think the team should discuss (and then document) what the 7 log levels mean for this project and when to use them.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -190,26 +216,29 @@ private void moveBatchOfBlocksToZipFile(final LongRange batchRange) {
...
+        // if we have reached here, then the batch of blocks has been zipped,
+        // now we need to make some updates
+        // update the first and last block numbers
+        availableBlocks.add(batchFirstBlockNumber, batchLastBlockNumber);
+        // log done
+        LOGGER.log(
+                System.Logger.Level.INFO,
+                "Moved batch of blocks[%d -> %d] to zip file",
+                batchFirstBlockNumber,
+                batchLastBlockNumber);
```
jsync-swirlds, line 236: This is another log level question (there are others, but I'll stop with this one).
INFO is a significant level, and (contrary to the name) should only contain _significant_ events (e.g. things that require operations attention).
This log (which is "we successfully did the thing we are supposed to do") is more often a `TRACE` level, or `DEBUG` at most.

jsync-swirlds (dismissed):
Some thoughts and questions.  Nothing here is blocking.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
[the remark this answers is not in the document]
jasperpotts: I just saw your comment and made me question the inclusive decision. I have been fixing anywhere I had seen ranges of block numbers even in test fixtures to always be inclusive so we do not have to think on case by case basis and know if working with block numbers and in block node code base it is always inclusive. So strongly feel we should have one kind everywhere. Just wanted to make sure I had not missed something and chosen the wrong one.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
jasperpotts: I think you need both executor factory method in context and the new method for config override that I added. Just need to make sure it works with non-string values. It may not. In which case we need the PR from Hendrik and should push to get it reviewed and merged.

jasperpotts (approved):
Happy for merge as well but would be good to think about some of the issues brought up in future improvements .

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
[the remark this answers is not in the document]
ata-nas: Good discussion.

Here are a couple of thoughts summarized:
1. I think ranges should be positive or zero
1. I think ranges should be inclusive always
1. I think we should not have the notion of an `empty` range, but simply one that not exist, this will drastically reduce complexity and fix current issues with exceptions thrown when merging a range that starts with `-1` with another range
1. I also agree that we should agree upon a single approach everywhere

As an action item in a follow up:
1. I think we should remove the possibility to have `-1 -> -1` range (`LongRange`) and treat it as an empty range
1. Update logic to support the removal of the `-1 -> -1` "empty" range
1. The ranged sets will be empty (behave like we would expect them to behave empty) when they simply have no ranges (`LongRange`) currently inside them.

I think that these changes will drastically reduce complexity (already it is working because of the change I made, but we need to update the logic to ensure no negative ranges and agree to switch to the new approach) and do cover our needs.

@jasperpotts @Nana-EC @jsync-swirlds opinions?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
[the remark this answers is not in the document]
ata-nas: @Nana-EC I think of making an issue for this so we can not lose this. We can chose to do something about this in the future as needed.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
ata-nas: Neat.
I think that we can extend the executor to support both as well. Let's think about this in the future, there are a couple of ways to do this and depending on what we need, we can either support multiple simultaneous calls of the `executeSerially` method or one only. It is hard for me to think about that without having the actual need to do so. In the test environment I do see the need to have both options as some asserts (and/or behavior) can be done via one option or the other.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
ata-nas: Cool idea. I am thinking of adding an issue to extend/improve the executor. We will need the extensions for sure in the future and I want to capture your ideas here. I hope that this class will be our go-to executor when testing and as @jasperpotts suggested, it will be injectable via the context once we add that support as well. This means that we will no longer need to have non-production logic just facilitate the possibility to test. 🔥

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/SimpleInMemoryHistoricalBlockFacility.java
ata-nas: Interesting, well this is from the past looks like. Since this class is not a `BlockProviderPlugin`, it does not implement the `BlockProviderPlugin#defaultPriority`. Technically the historical block facility consists of multiple registered plugins which have the default priority and execute their logic respectively. Here this number is I believe one that fits the test cases we have up until now. I believe we should be setting this via the constructor so we will have control over this and we are mindful (my fear is that not having the correct number might produce false positives in our tests). Good catch @jsync-swirlds 🚀

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
ata-nas: I suspect that we can facilitate one of the future internal design discussions to include the point about logging, the process we should agree to use to determine levels etc., but also it would be good to discuss when we should consider to shutdown the server (call the shutdown on the health facility). Sound like a good pair of (somewhat) connected items to discuss in a sitting.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
ata-nas: @jasperpotts in our case, supplying a string value of the path (a jUnit `@TempDir` for these tests) works just fine, because the temp dir is resolved directly to the os (somewhere under `/var` randomly and gets cleaned up after tests). The config dependency will indeed create a new instance of `Path` (`UnixPath` specifically in our case) during initialization, but that is fine, because the string value of the path will be the same and we know that it will point physically to the same place, so cleanup will not be an issue.

The overrides you introduced recently, the ones I started using with latest commit here, only work with string values: api from consensus lib for configs, and we will go through their resolution logic once we build the configs. This is problematic when we want to use `jimfs` because there we need an instance of their implementation of the `Path` interface, which the config dependency will never build (hence why we are waiting for the change you have requested for the ability to override converters for classes supported by default, so we can supply the `jimfs` specific `Path`, to be used in other places)

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -56,42 +73,41 @@ public List<Class<? extends Record>> configDataTypes() {
     @Override
     public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
         this.context = context;
-        final FilesHistoricConfig config = context.configuration().getConfigData(FilesHistoricConfig.class);
+        final FilesHistoricConfig localConfig =
+                this.config == null ? context.configuration().getConfigData(FilesHistoricConfig.class) : this.config;
```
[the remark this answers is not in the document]
ata-nas: No longer relevant, some overriding changes were introduced and we no longer have this concern for this test. This has been reverted with latest commits.

ata-nas:
@jasperpotts @jsync-swirlds thanks for your approvals, had to rebase because of @jsync-swirlds's latest merge to main. But also took the opportunity to do some additional cleanup.

I kindly ask for your re-approvals. I have also captured all the feedback and created some issues so that we do not lose it. Will iterate with some improvements.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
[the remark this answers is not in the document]
ata-nas: @Nana-EC I have created #1069 to capture this and not lose it. Let's explore this concern and act upon it if we deem necessary. Otherwise we can close and keep history.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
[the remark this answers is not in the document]
ata-nas: Let's proceed like this for now, as mentioned elsewhere, let's discuss logging + when to shutdown server and strategies to use for throwing/catching etc. in an upcoming internal design meeting.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
ata-nas: P.S. I disabled these for now because they hang and the process cannot exit, but your proposal will put these into proper use.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
ata-nas: @jsync-swirlds @Nana-EC I have made #1070 to capture @jsync-swirlds improvement proposals.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
ata-nas: @jsync-swirlds @Nana-EC I have made #1070 to capture @jsync-swirlds improvement proposals.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/SimpleInMemoryHistoricalBlockFacility.java
ata-nas: P.S. is there any action item here? Creating an issue or something else? @jasperpotts @jsync-swirlds @Nana-EC ?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
ata-nas: Same as [here](https://github.com/hiero-ledger/hiero-block-node/pull/1064#discussion_r2063120950)

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/SimpleInMemoryHistoricalBlockFacility.java
Nana-EC: Same question.
I think @jasperpotts may have an idea here.
I see `BlocksFilesHistoricPlugin` has a defaultPriority of 100, `BlocksFilesRecentPlugin` and `BlocksCloudHistoricPlugin` have a default of 2000.
I'd say it's prioritizing recent and cloud over historic but we don't seem to have the priority logic documented so i'm not sure of the other options.

I agree not being clear here will have tests with low confidence.
@ata-nas you can open a ticket to clarify priority values of the Plugins so it's clear as right now it seems a little arbitrary.

Nana-EC (approved):
Unblocking.
Please open a ticket around Historical Plugin that captures the comments or open requests in this PR so we can follow up.
If they are straightforward go ahead and tackle if not let's discuss as a team

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -133,24 +145,34 @@ public BlockRangeSet availableBlocks() {
...
     // ==== Private Methods ============================================================================================
+    private void attemptZipping() {
+        // compute the min and max block in next batch to zip
+        long minBlockNumber = availableBlocks().max() + 1;
+        long maxBlockNumber = minBlockNumber + numberOfBlocksPerZipFile - 1;
+        // check all those blocks are available
+        final BlockRangeSet historicalAvailable =
+                context.historicalBlockProvider().availableBlocks();
+        // while we can zip blocks, we must keep zipping
+        while (historicalAvailable.max() >= maxBlockNumber) {
```
Nana-EC, line 165: Q: could the logic to decide whether to zip be simplified.
I see a few calls and calculations to manage ranges, would these be simplified if `ZipBlockArchive` tracked on its own the max block number it's had archived?
The number would be calculated once on `BlockFilesHistoricPlugin` init and then updated at the end of the zip and move.
Still thinking through the implications but I'm curious

Nana-EC (approved):
Left additional considerations.
Comments can be resolved once captured in a follow up ticket if applicable

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
georgi-l95: Good remark from Joseph. I agree that we should discuss and document 👍

georgi-l95 (approved):
LG

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingSerialExecutor.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A very simple executor to be used only for testing! This executor will
+ * collect all submitted tasks to it and will hold them in order. After that
+ * when the {@link #executeSerially()} method is called, all tasks will be
+ * executed in order serially on the same thread that called the method. This
+ * will ensure that all tasks will complete before doing any asserts
+ */
+public class BlockingSerialExecutor extends ThreadPoolExecutor {
```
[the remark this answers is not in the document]
ata-nas: @Nana-EC this is test fixtures. Available for test only.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/NoOpServiceBuilder.java
```diff
@@ -0,0 +1,30 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.helidon.webserver.http.HttpService;
+import org.hiero.block.node.spi.ServiceBuilder;
+
+/**
+ * A simple no-op implementation of {@link ServiceBuilder} that does nothing.
+ * To be used for testing purposes only where we need a non-null implementation
+ * that we do not want to act upon.
+ */
+public final class NoOpServiceBuilder implements ServiceBuilder {
```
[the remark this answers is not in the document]
ata-nas: @Nana-EC this is test fixtures. Available for test only.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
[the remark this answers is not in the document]
ata-nas: @Nana-EC @jasperpotts @jsync-swirlds #1072 is made to address (finish) this discussion and so we do not miss this.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
ata-nas: I would counter here. Deciding when and what to zip IMO should not be a concern of the zip archiver. The archiver should only do it's job and should not be a decision maker. As long as it is called and given parameters to work with, it will do it's job. IMO if it has some internal state that will follow next in line will in fact complicate the whole thing. As the source of truth regarding what blocks are available remains the plugin itself, let's manage starting zips and updating availables via it.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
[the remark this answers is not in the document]
ata-nas: @Nana-EC @jsync-swirlds I have made a slight extension to the description and title of #1069. I have added the following:
- should we also assert that all blocks in the given range are present (we have the correct accessors for)?
  - if so, then what do we do if there are gaps, do we fail the zip or are gaps in the zip OK?

PR #1076, opened 2025-04-30
test: Add subscriber plugin Tests

### LINKED ISSUE #584: Simulator integration tests for closed-range historic streaming
* Add integration tests using the Simulator to verify the API functionality
* Verify the design outlined here: https://github.com/hashgraph/hedera-block-node/pull/577

### LINKED ISSUE #933: Add Consumer Plugin Tests
### Story Form

As a Block Node Developer
I want to have enough code coverage on the Stream Subscriber Plugin
So that I can confidently use and deploy the application in production environments.

### Technical Notes

This effort should focus on improving unit tests for the plugin created under issue #1051.

georgi-l95:
## Reviewer Notes
This PR aim to:
- Introduce minor improvements to object checks, allocations and more.
- Add unit tests for config and session objects, with enough coverage and cases to ensure that we have adequte confidence in the plugin.
- Add E2E covering both negative and positive scenarios for single consumer cases and multiple consumer case.
- Fix issue, where we don't close the session and stop the streaming, when we execute succesfully the request.
- Enable new plugin as default.

Things to consider:
- One unit test disabled due to unstable behavior. Passes when run as standalone, fails when run with all others from the same class.
- Added check in method `allRequestedBlocksSent()` to ensure that when consumer requests 0 as end block, we continue to stream indefinitely.
- Added `close()` session method call in `call()`, when we execute the request, in cases where it's valid and consumer does not want live streaming(0 as `endblock`).

## Related Issue(s)
Fixes #933 
Fixes #584

### DIFF
--- a/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
+++ b/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
@@ -117,9 +117,6 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
             @NonNull final Pipeline<? super SubscribeStreamResponseUnparsed> responsePipeline,
             @NonNull final BlockNodeContext context,
             @NonNull final CountDownLatch sessionReadyLatch) {
-        requireNonNull(request);
-
-        LOGGER.log(Level.TRACE, request.toString());
         this.clientId = clientId;
         this.startBlockNumber = request.startBlockNumber();
         this.endBlockNumber = request.endBlockNumber();
@@ -177,14 +174,13 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
                     // then process live blocks, if available.
                     sendLiveBlocksIfAvailable();
                 }
-                if (!interruptedStream.get()) {
+                if (allRequestedBlocksSent()) {
                     // We've sent all request blocks. Therefore, we close, according to the protocol.
                     close(SubscribeStreamResponse.Code.READ_STREAM_SUCCESS);
                 }
             }
         } catch (RuntimeException | ParseException | InterruptedException e) {
             sessionFailedCause = e;
-            interruptedStream.set(true);
             close(SubscribeStreamResponse.Code.READ_STREAM_SUCCESS); // Need an "INCOMPLETE" code...
         }
         // Need to record a metric here with client ID tag, so we can record
@@ -210,7 +206,7 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
             } else {
                 // Only give up if this is an historical block, otherwise just
                 // go back up and see if live has the block.
-                if (!(nextBlockToSend < 0 || nextBlockToSend >= getLatestKnownBlock())) {
+                if (!(nextBlockToSend < 0 || nextBlockToSend >= getLatestHistoricalBlock())) {
                     // We cannot get the block needed, something has failed.
                     // close the stream with an "unavailable" response.
                     final String message = "Unable to read historical block {0}.";
@@ -236,7 +232,7 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
      */
     private boolean isHistoryPermitted() {
         return !(interruptedStream.get()
-                || latestLiveStreamBlock.get() < nextBlockToSend
+                || (latestLiveStreamBlock.get() < nextBlockToSend && latestLiveStreamBlock.get() >= 0)
                 || allRequestedBlocksSent()
                 || nextBatchIsLive());
     }
@@ -259,9 +255,6 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
     }
 
     private boolean allRequestedBlocksSent() {
-        if (endBlockNumber == 0L) { // We are requesting block indefinitely.
-            return false;
-        }
         return endBlockNumber != UNKNOWN_BLOCK_NUMBER && nextBlockToSend > endBlockNumber;
     }
 
@@ -282,21 +275,31 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
         // If we run out, get ahead of live, or have to send historical blocks,
         // then we'll also break out of the loop and return to the caller.
         while (!liveBlockQueue.isEmpty()) {
-            // take the block item from the queue and process it
-            final BlockItems blockItems = liveBlockQueue.poll();
-            // Live _might_ be behind the next expected block (particularly if
-            // the requested start block is in the future), skip this block, in that case.
-            if (blockItems.isStartOfNewBlock() && blockItems.newBlockNumber() != nextBlockToSend) {
+            // Peek at the block item from the queue and _possibly_ process it
+            BlockItems blockItems = liveBlockQueue.peek();
+            // Live _might_ be ahead or behind the next expected block (particularly if
+            // the requested start block is in the future), don't send this block, in that case.
+            // If behind, remove that item from the queue; if ahead leave it in the queue.
+            if (blockItems.isStartOfNewBlock() && blockItems.newBlockNumber() < nextBlockToSend) {
                 LOGGER.log(Level.TRACE, "Skipping block {0} for client {1}", blockItems.newBlockNumber(), clientId);
                 skipCurrentBlockInQueue(blockItems);
-            } else {
+            } else if (blockItems.newBlockNumber() == nextBlockToSend) {
+                blockItems = liveBlockQueue.poll();
                 if (blockItems != null) {
                     sendOneBlockItemSet(blockItems);
+                    if (blockItems.isEndOfBlock()) {
+                        nextBlockToSend++;
+                        trimBlockItemQueue(nextBlockToSend);
+                    }
                 }
-                if (blockItems.isEndOfBlock()) {
-                    nextBlockToSend++;
-                    trimBlockItemQueue(nextBlockToSend);
-                }
+            } else if (blockItems.isStartOfNewBlock() && blockItems.newBlockNumber() > nextBlockToSend) {
+                // This block is _future_, so we need to wait, and try to get the next block from history
+                // first, then come back to this block.
+                LOGGER.log(Level.TRACE, "Retaining future block {0} for client {1}", blockItems.newBlockNumber(), clientId);
+            } else {
+                // This is a past or future _partial_ block, so we need to trim the queue.
+                liveBlockQueue.poll(); // discard _this batch only_.
+                trimBlockItemQueue(nextBlockToSend);
             }
             // Note: We depend here on the rule that there are no gaps between blocks.
             //       The header for block N immediately follows the proof for block N-1.
@@ -349,20 +352,26 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
     /**
      * Remove the remainder of a block from the queue.
      * <p>
-     * This method assumes that a possible header batch has already been removed
-     * from the queue (and is provided). The provided item is checked, and if it
-     * is a header block, the remainder of that block, up to the next header
-     * batch (which might be the next item) is removed from the queue.<br/>
-     * Note, the item provided and the next item are not removed from the queue,
-     * so it is important to only call this method after polling and item, and
-     * when this method returns, the next item in the queue will be the start of
-     * a new block (or else the queue will be empty).
+     * This method assumes that a possible header batch has already been viewed
+     * but not removed from the queue (and is provided). The provided item is
+     * checked, and if it is a header block, it is removed and then the
+     * remainder of that block, up to the next header batch (which might be the
+     * next item) is removed from the queue.<br/>
+     * Note, the item provided _may be_ removed from the queue (even if it's not
+     * a block header), so it is important to only call this method after
+     * peeking at the item without removing it, and when this method returns,
+     * the next item in the queue will be the start of a new block (or else
+     * the queue will be empty).
      */
     private void skipCurrentBlockInQueue(BlockItems queueHead) {
-        // The "head" entry is already removed, remove the rest of its block if it's a block header.
-        if (queueHead != null && queueHead.isStartOfNewBlock()) {
-            queueHead = liveBlockQueue.peek();
+        // The "head" entry is _not_ already removed, remove it, and the rest of
+        // its block. This also handles a partial block at the head of the queue
+        // when we cannot process that block (e.g. it's in the future or past from
+        // the block we currently need to send).
+        if (queueHead != null) {
+            liveBlockQueue.poll(); // remove the "head" entry
             // Now remove "head" entries until the _next_ item is a block header
+            queueHead = liveBlockQueue.peek(); // peek at the next item.
             while (queueHead != null && !(queueHead.isStartOfNewBlock())) {
                 liveBlockQueue.poll();
                 queueHead = liveBlockQueue.peek();
@@ -476,11 +485,6 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
         return nextBlockToSend;
     }
 
-    // Visible for testing.
-    LiveBlockHandler getLiveBlockHandler() {
-        return liveBlockHandler;
-    }
-
     /**
      * Close this session. This will unregister us from the block messaging system and cancel the subscription.
      */
@@ -491,11 +495,13 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
             sessionReadyLatch.countDown();
             LOGGER.log(Level.DEBUG, "Session ready latch was not counted down on close, releasing now");
         }
-        // unregister us from the block messaging system, if we are not registered then this is noop
-        context.blockMessaging().unregisterBlockItemHandler(liveBlockHandler);
-        final Builder response = SubscribeStreamResponseUnparsed.newBuilder().status(endStreamResponseCode);
-        responsePipeline.onNext(response.build());
-        responsePipeline.onComplete();
+        if(!interruptedStream.get()) {
+            // unregister us from the block messaging system, if we are not registered then this is noop
+            context.blockMessaging().unregisterBlockItemHandler(liveBlockHandler);
+            final Builder response = SubscribeStreamResponseUnparsed.newBuilder().status(endStreamResponseCode);
+            responsePipeline.onNext(response.build());
+            responsePipeline.onComplete();
+        }
         if (subscription != null) {
             subscription.cancel();
             subscription = null;
@@ -542,7 +548,7 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
     //       This means we must not modify any state in the session object directly.
     //       Instead we must use a transfer queue to pass the block items to the session
     //       and possibly set an atomic flag if we are "too far behind".
-    static class LiveBlockHandler implements NoBackPressureBlockItemHandler {
+    private static class LiveBlockHandler implements NoBackPressureBlockItemHandler {
         /** The logger for this class. */
         private final Logger LOGGER = System.getLogger(getClass().getName());
 
--- a/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
+++ b/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
@@ -177,6 +177,7 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
                     // then process live blocks, if available.
                     sendLiveBlocksIfAvailable();
                 }
+                close(SubscribeStreamResponse.Code.READ_STREAM_SUCCESS); // Need an "INCOMPLETE" code...
             }
         } catch (RuntimeException | ParseException | InterruptedException e) {
             sessionFailedCause = e;
@@ -535,7 +536,7 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
     //       This means we must not modify any state in the session object directly.
     //       Instead we must use a transfer queue to pass the block items to the session
     //       and possibly set an atomic flag if we are "too far behind".
-    private static class LiveBlockHandler implements NoBackPressureBlockItemHandler {
+    static class LiveBlockHandler implements NoBackPressureBlockItemHandler {
         /** The logger for this class. */
         private final Logger LOGGER = System.getLogger(getClass().getName());
 
--- a/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
+++ b/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.stream.subscriber;
 
+import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.locks.LockSupport.parkNanos;
 import static org.hiero.block.node.spi.BlockNodePlugin.METRICS_CATEGORY;
 import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
@@ -11,10 +12,10 @@ import com.hedera.pbj.runtime.ParseException;
 import com.hedera.pbj.runtime.grpc.Pipeline;
 import com.swirlds.metrics.api.Counter;
 import com.swirlds.metrics.api.Counter.Config;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import java.lang.System.Logger;
 import java.lang.System.Logger.Level;
 import java.util.List;
-import java.util.Objects;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Callable;
@@ -111,18 +112,20 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
      * @param context The context for the block node
      */
     public BlockStreamSubscriberSession(
-            long clientId,
-            SubscribeStreamRequest request,
-            Pipeline<? super SubscribeStreamResponseUnparsed> responsePipeline,
-            BlockNodeContext context,
-            final CountDownLatch sessionReadyLatch) {
+            final long clientId,
+            @NonNull final SubscribeStreamRequest request,
+            @NonNull final Pipeline<? super SubscribeStreamResponseUnparsed> responsePipeline,
+            @NonNull final BlockNodeContext context,
+            @NonNull final CountDownLatch sessionReadyLatch) {
+        requireNonNull(request);
+
         LOGGER.log(Level.TRACE, request.toString());
         this.clientId = clientId;
         this.startBlockNumber = request.startBlockNumber();
         this.endBlockNumber = request.endBlockNumber();
-        this.responsePipeline = Objects.requireNonNull(responsePipeline);
-        this.context = Objects.requireNonNull(context);
-        this.sessionReadyLatch = Objects.requireNonNull(sessionReadyLatch);
+        this.responsePipeline = requireNonNull(responsePipeline);
+        this.context = requireNonNull(context);
+        this.sessionReadyLatch = requireNonNull(sessionReadyLatch);
         latestLiveStreamBlock = new AtomicLong(UNKNOWN_BLOCK_NUMBER - 1);
         pluginConfiguration = context.configuration().getConfigData(SubscriberConfig.class);
         // Next block to send depends on what was requested and what is available.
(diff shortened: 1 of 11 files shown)


block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -111,18 +112,20 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
...
-            SubscribeStreamRequest request,
-            Pipeline<? super SubscribeStreamResponseUnparsed> responsePipeline,
-            BlockNodeContext context,
-            final CountDownLatch sessionReadyLatch) {
+            final long clientId,
+            @NonNull final SubscribeStreamRequest request,
+            @NonNull final Pipeline<? super SubscribeStreamResponseUnparsed> responsePipeline,
+            @NonNull final BlockNodeContext context,
+            @NonNull final CountDownLatch sessionReadyLatch) {
+        requireNonNull(request);
```
jsync-swirlds, line 120: This line isn't needed, the very first use of request will throw if request is null.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -111,18 +112,20 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
...
+
         LOGGER.log(Level.TRACE, request.toString());
```
jsync-swirlds, line 122: Should probably remove this log line, as it's pretty noisy.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -174,6 +177,10 @@ public BlockStreamSubscriberSession call() {
                     // then process live blocks, if available.
                     sendLiveBlocksIfAvailable();
                 }
+                if (!interruptedStream.get()) {
+                    // We've sent all request blocks. Therefore, we close, according to the protocol.
+                    close(SubscribeStreamResponse.Code.READ_STREAM_SUCCESS);
+                }
```
jsync-swirlds, line 183: This is not correct, interrupted does not imply we succeeded, it's an interrupt, not all blocks sent.
The close is correct, but only because we don't have an "interrupted" response code.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -251,6 +259,9 @@ private void awaitNewLiveEntries() {
     }
 
     private boolean allRequestedBlocksSent() {
+        if (endBlockNumber == 0L) { // We are requesting block indefinitely.
+            return false;
+        }
```
jsync-swirlds, line 264: This is not correct.  `0L` as end block is valid if I wish to request block `0`.
Infinite is `UNKNOWN_BLOCK_NUMBER` and is handled in the next line.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -532,7 +542,7 @@ private void sendOneBlockItemSet(final List<BlockItemUnparsed> blockItems) {
     //       This means we must not modify any state in the session object directly.
     //       Instead we must use a transfer queue to pass the block items to the session
     //       and possibly set an atomic flag if we are "too far behind".
-    private static class LiveBlockHandler implements NoBackPressureBlockItemHandler {
+    static class LiveBlockHandler implements NoBackPressureBlockItemHandler {
```
jsync-swirlds, line 545: This was private deliberately, as we shouldn't be trying to test this implementation detail or exposing it for testing.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -557,7 +567,7 @@ public void onTooFarBehindError() {
         }
 
         @Override
-        public void handleBlockItemsReceived(final BlockItems blockItems) {
+        public void handleBlockItemsReceived(@NonNull final BlockItems blockItems) {
             if (blockItems.newBlockNumber() > latestLiveStreamBlock.get()) {
                 latestLiveStreamBlock.set(blockItems.newBlockNumber());
                 LOGGER.log(Level.TRACE, "Updated latest block to {0}.", latestLiveStreamBlock);
```
jsync-swirlds, line 573: Probably should remove most of these `TRACE` logs.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -31,7 +33,22 @@
...
+/**
+ * Provides implementation for the block stream subscriber endpoints of the server. These handle incoming requests for block
+ * stream from consumers.
+ *
+ * <p>This plugin is responsible for:
+ * <ul>
+ *   <li>Managing subscriber sessions for clients requesting block streams</li>
+ *   <li>Handling both live-streaming and historical block data requests</li>
+ *   <li>Implementing the gRPC service interface for block stream subscriptions</li>
+ *   <li>Coordinating between multiple publishers when they connect to the block node</li>
```
jsync-swirlds, line 45: This is an hallucination; this plugin has nothing to do with publishers.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -31,7 +33,22 @@
...
+ *   <li>Maintaining metrics about active subscribers and stream health</li>
```
jsync-swirlds, line 46: This isn't quite true, stream health isn't really monitored and probably cannot be; it's a nebulous concept at best.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -157,25 +176,38 @@ Map<Long, BlockStreamSubscriberSession> getOpenSessions() {
...
+    /**
+     * Handler for block stream subscription requests from clients.
+     * Responsible for:
+     * <ul>
+     *   <li>Managing client sessions for block stream subscribers</li>
+     *   <li>Processing subscription requests and streaming block data to clients</li>
+     *   <li>Tracking active subscriber sessions and their lifecycle</li>
+     *   <li>Handling graceful termination of subscriber connections</li>
+     *   <li>Reporting metrics about active subscribers</li>
+     * </ul>
```
jsync-swirlds, line 188: This is highly misleading. Most of these "responsibilities" are _mostly_ handled by the `Session` thread, this class really just supplies the methods to initiate a session and manage the _futures_ for those sessions.
I'd recommend removing "responsible for" sections entirely (particularly as copilot generally gets it completely wrong when generating these).
Perhaps focus on a more prosaic description of the class initially.

jsync-swirlds (commented):
Quick initial look at the non-test classes, will look at more on Monday.
A few incorrect documentation items, and a couple over-zealous adjustments.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -0,0 +1,398 @@
...
+    private CountDownLatch sessionReadyLatch;
+    private static final long CLIENT_ID = 0L;
+
+    @BeforeEach
+    void setUp() {
+        sessionReadyLatch = new CountDownLatch(1);
+        final Configuration configuration = ConfigurationBuilder.create()
+                .withConfigDataType(SubscriberConfig.class)
+                .build();
+        responsePipeline = spy(new ResponsePipeline());
```
jsync-swirlds, line 72: Using a `spy` to assert invocation counts is concerning; we end up testing far too specifically in almost all cases. I think we'd be better off validating the _result_ of a pipeline rather than trying to count method invocations (particularly as it's entirely reasonable to rework the pipeline quite dramatically but tests for the Session shouldn't fail as a result).

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -0,0 +1,398 @@
...
+            boolean historicalProvided = false;
+            while (!historicalProvided) {
+                try {
+                    verify(responsePipeline, atLeast((int) MAX_AVAILABLE_BLOCK))
+                            .onNext(any(SubscribeStreamResponseUnparsed.class));
+                    historicalProvided = true;
+                } catch (TooFewActualInvocations e) {
+                    Thread.sleep(100);
+                }
+            }
```
jsync-swirlds, line 137: Would be good to check the responses sent into a buffer, rather than checking a spy method was called.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -0,0 +1,398 @@
...
+                        .blockItems(blockItems.blockItems())
+                        .build();
+
+                lenient().when(blockAccessor.blockUnparsed()).thenReturn(blockUnparsed);
+                lenient().when(context.historicalBlockProvider().block(i)).thenReturn(blockAccessor);
+            }
+
+            // Start session in separate thread
+            Thread sessionThread = new Thread(session::call);
+            sessionThread.start();
```
jsync-swirlds, line 125: After start, it might be good to check that the "session ready" countdown latch is unlocked before continuing (that's partially why it exists).

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -0,0 +1,398 @@
...
+                final SubscribeStreamResponseUnparsed.Builder response = SubscribeStreamResponseUnparsed.newBuilder()
+                        .blockItems(BlockItemSetUnparsed.newBuilder().blockItems(blockItems.blockItems()));
+                Thread.sleep(100);
+                verify(responsePipeline, times(1)).onNext(response.build());
+            }
+
+            // Verify final pipeline state
+            verify(responsePipeline, times(1)).onComplete();
+            verify(responsePipeline, never()).onError(any(Throwable.class));
+        }
```
jsync-swirlds, line 153: Here again, it would be nice to check the actual responses from the pipeline rather than spying and checking invocations...

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -0,0 +1,398 @@
...
+    @Nested
+    @DisplayName("Input Validation Tests")
+    class ValidationTests {
+
+        /**
+         * Tests the handling of a start block number that is neither available in history
+         * nor in the live stream range.
+         */
+        @Test
+        @DisplayName("Should reject request with start block outside available range")
```
jsync-swirlds, line 215: A lot of these duplicate tests in the plugin tests. Would it, perhaps, be better to focus on testing things here that those tests do not cover?

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -0,0 +1,398 @@
...
+     * Creates a sample block items set for testing.
+     *
+     * @param blockNumber the block number to create items for
+     * @return a BlockItems object containing sample block items
+     */
+    private BlockItems createSampleBlockItems(long blockNumber) {
+        BlockItemUnparsed blockHeader = toBlockItemUnparsed(sampleBlockHeader(blockNumber));
+        BlockItemUnparsed roundHeader = toBlockItemUnparsed(sampleRoundHeader(blockNumber));
+        BlockItemUnparsed blockProof = toBlockItemUnparsed(sampleBlockProof(blockNumber));
+        return new BlockItems(List.of(blockHeader, roundHeader, blockProof), blockNumber);
```
jsync-swirlds, line 342: We have a fixture method that does this.  Might be good to use that.

It could also be helpful to extend the fixture to include batches that are not complete blocks.  This might enable tests (particularly for live blocks) to verify that the code (using a fixture history provider instead of a mock so history assembles the full blocks) handles a live stream of blocks split across multiple batches (perhaps including several different items, beyond just round header).

Perhaps something to add in the next PR...

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -0,0 +1,398 @@
...
+     * @param maxBlock the maximum available block number
+     */
+    private void setupHistoricalBlockProvider(long minBlock, long maxBlock) {
+        final BlockRangeSet availableBlocks = mock(BlockRangeSet.class);
+        when(availableBlocks.min()).thenReturn(minBlock);
+        when(availableBlocks.max()).thenReturn(maxBlock);
+        when(context.historicalBlockProvider()).thenReturn(historicalBlockFacility);
+        when(context.blockMessaging()).thenReturn(blockMessagingFacility);
+        when(context.historicalBlockProvider().availableBlocks()).thenReturn(availableBlocks);
+    }
```
jsync-swirlds, line 358: Instead of a mock, this could use the fixture in-memory history provider and just feed in the blocks to be returned...

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -0,0 +1,398 @@
...
+        return SubscribeStreamRequest.newBuilder()
+                .startBlockNumber(startNumber)
+                .endBlockNumber(endNumber)
+                .build();
+    }
+
+    /**
+     * A simple implementation of Pipeline for testing purposes.
+     */
+    private class ResponsePipeline implements Pipeline<SubscribeStreamResponseUnparsed> {
```
jsync-swirlds, line 377: There is a similar implementation in fixtures (although it's buried and an anonymous inner class) perhaps we could extract that to a top level class to make it easier to use here instead of duplicating it?
```java
        new Pipeline<>() {
            @Override
            public void clientEndStreamReceived() {
                LOGGER.log(TRACE, "clientEndStreamReceived");
            }

            @Override
            public void onNext(Bytes item) throws RuntimeException {
                fromPluginBytes.add(item);
            }

            @Override
            public void onSubscribe(Subscription subscription) {
                LOGGER.log(TRACE, "onSubscribe");
            }

            @Override
                fail("onError: ", throwable);
            }

            @Override
            public void onComplete() {
                LOGGER.log(TRACE, "onComplete");
            }
        };
```

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/SubscriberConfigTest.java
```diff
@@ -0,0 +1,105 @@
...
+import com.swirlds.config.api.Configuration;
+import com.swirlds.config.api.ConfigurationBuilder;
+import com.swirlds.config.api.validation.ConfigViolationException;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests for the {@link SubscriberConfig} class.
+ */
+public class SubscriberConfigTest {
```
jsync-swirlds, line 15: Why are we testing configuration?
It seems awfully low value, and also extremely brittle.  I should be able to modify configuration values and defaults without also having to modify test classes.

Note, I know this is different from how we've done things in the past, but one of the things discussed at Devcon was that we should not have tests so tightly coupled to implementation that refactoring is made _more difficult_.  Part of the value of unit tests is to provide confidence when refactoring (because unmodified tests still pass), not add extra work and complexity.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/SubscriberTest.java
```diff
@@ -225,6 +225,7 @@ void testSubscriberBlockStreamInMiddleNoHistory() throws ParseException {
         activePlugin.stop(); // request the plugin to end all client streams.
     }
 
+    @Disabled("Disabled until issue with it is resolved. Works when run standalone")
```
jsync-swirlds, line 228: hmmm... This _was_ working, but it was also one of the hardest tests to get working correctly.
Perhaps send me a DM with what's not working or how it's failing?  I'd like to understand this better.

suites/src/main/java/org/hiero/block/suites/subscriber/positive/PositiveSingleSubscriberTests.java
```diff
@@ -0,0 +1,173 @@
...
+            } catch (InterruptedException e) {
+                // Do nothing, this is not mandatory, we try to shut down cleaner and graceful
+            }
+        });
+        simulators.forEach(simulator -> simulator.cancel(true));
+        simulators.clear();
+    }
+
+    @Test
+    @DisplayName("Should subscribe to stream and receive both historical and live")
```
jsync-swirlds, line 53: Just a note here.  It's almost impossible (from outside the node) to know if you received a block from history or live.  The subscriber is intended to conceal the source as effectively as possible so that the subscriber just receives the stream, regardless of live or historical.
For this test, all we really can know for certain is that we receive the stream timely; the effective source of each block is pretty obscured.

jsync-swirlds (commented):
Read through the rest.
Some concerns with use of mocks when fixtures (which already exist) might be better.
Also a few other small suggestions or questions.

Thanks for writing the Suite tests, those help a lot with confidence that everything works together the way it should.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
georgi-l95: My understanding was the same (-1 is indefinetely only). However then I saw the protocol definitions, [which](https://github.com/hashgraph/hedera-protobufs/blob/main/block/block_service.proto#L589-L609) states that: 
```
     * This field MAY be zero (`0`) to indicate the stream SHOULD continue
     * indefinitely, streaming new blocks as each becomes available.<br/>
     * If this value is greater than zero (`0`)
```

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
georgi-l95: I was copying and pasting some templates for the javadoc around and didn't edit them correctly. My bad, thank you

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
georgi-l95: You're right. Will remove.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
georgi-l95: Thank you for resolving this. 🙏

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
georgi-l95: Thank you for resolving this. 🙏

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
georgi-l95: Thank you for resolving this. 🙏

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
georgi-l95: I also think that -1 should be the only case for indefinitely streaming. That means we need to change the documentation in the protobufs.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
georgi-l95: Yes, I was trying to test it out. Thank you for resolving this.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
georgi-l95: Removed.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
georgi-l95: Edited

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
georgi-l95: Removed.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -41,8 +58,9 @@ public class SubscriberServicePlugin implements BlockNodePlugin, ServiceInterfac
     private final List<SubscribeBlockStreamHandler> clientHandlers = new LinkedList<>();
     /** The block node context */
     private BlockNodeContext context;
-
+    /** The subscriber plugin configuration */
     private SubscriberConfig pluginConfiguration;
+    /** The client handler for client subscriptions*/
```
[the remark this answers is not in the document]
georgi-l95: Gotcha, edited. Thanks

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -51,8 +69,9 @@ public class SubscriberServicePlugin implements BlockNodePlugin, ServiceInterfac
      * {@inheritDoc}
      */
     @Override
-    public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
-        this.context = context;
+    public void init(@NonNull final BlockNodeContext context, @NonNull final ServiceBuilder serviceBuilder) {
+        requireNonNull(serviceBuilder);
```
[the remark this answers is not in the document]
georgi-l95: Removed. You're right.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -157,25 +176,38 @@ Map<Long, BlockStreamSubscriberSession> getOpenSessions() {
...
         private final Logger LOGGER = System.getLogger(getClass().getName());
         /** Count of active sessions, because LongGauge doesn't support increment/decrement */
-        private AtomicLong sessionCount = new AtomicLong(0L);
+        private final AtomicLong sessionCount = new AtomicLong(0L);
         /** The next client id to use when a new client session is created */
         private final AtomicLong nextClientId = new AtomicLong(0);
-
+        /** The block node context */
         private final BlockNodeContext context;
+        /** The subscriber plugin responsible for handling incoming requests*/
```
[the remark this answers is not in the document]
georgi-l95: Edited. thanks

suites/src/main/java/org/hiero/block/suites/subscriber/positive/PositiveSingleSubscriberTests.java
georgi-l95: Yes, in real scenario we wouldn't know. Now we can determine now due to the simulator. You're right.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
georgi-l95: I can open an issue for us to adress that. I think there is a similar pipeline in the publisher tests also, which could be addressed in the said issue 🤔

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
georgi-l95: Thank you for adressing this.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
georgi-l95: Agree, wanted to make sure we cover some edge cases, but you're right.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
georgi-l95: Agree with you 👍  Thank you for addressing this

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
georgi-l95: Yes, thanks for adding addition lists and getters for us to check this in the pipeline class.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -0,0 +1,398 @@
...
+    private BlockNodeContext context;
+
+    @Mock
+    private Metrics metrics;
+
+    @Mock
+    private HistoricalBlockFacility historicalBlockFacility;
+
+    @Mock
+    private BlockMessagingFacility blockMessagingFacility;
```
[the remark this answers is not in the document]
georgi-l95: Thanks for addressing this.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
georgi-l95: Good idea, opened a issue https://github.com/hiero-ledger/hiero-block-node/issues/1086

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
jsync-swirlds: The proto documentation is definitely wrong.  I keep saying I should make a clean pass to ensure it's all "correct"; time just hasn't been on our side in that regard.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/SubscriberConfigTest.java
jsync-swirlds: I would lean toward just not having testing for config classes directly; any "does this work correctly with incorrect values" should probably be in the unit tests for the code _consuming_ the config, rather than the config class.

suites/src/main/java/org/hiero/block/suites/subscriber/positive/PositiveSingleSubscriberTests.java
jsync-swirlds: I had not considered that the simulator could be used to ensure the node reads from one source or another; a useful feature.

jsync-swirlds (dismissed):
Looks good.  Thanks for working through this.

jsync-swirlds (approved):

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/SubscriberConfigTest.java
georgi-l95: Opened a issue to remove this class and others similar (publisher also has). So that all can be addressed in another effort and we don't get out of scope for this one. https://github.com/hiero-ledger/hiero-block-node/issues/1090

PR #1077, opened 2025-05-02
chore: Update SECURITY.md

jwagantall:
The SECURITY.md file update includes the additional information approved in the TSC call and the reference regarding the bounty program.


jwagantall:
SECURITY.md will be moved to .github

PR #1083, opened 2025-05-05
feat: BlockNodeContext Executor Service Factory Method

### LINKED ISSUE #1067: `BlockNodeContext` Executor Service Factory Method
### Story Form

As a Block Node Developer
I want to have a streamlined, clean and manageable approach to creating/managing thread pools
So that I have a greater control of thread pool creation and an easier maintainance if need be

### Technical Notes

- As suggested in this [comment thread](https://github.com/hiero-ledger/hiero-block-node/pull/1064#discussion_r2058797092) we should create a factory method within the `BlockNodeContext` that will allow us to create thread pools.
- As suggested, we should have some configurable parameters, the factory method in itself is easily extendable if need be.
- As suggested, we should also consider logging/monitoring for the pools created and who issued the creation.

ata-nas:
## Reviewer Notes

- introduce a new `ThreadPoolManager`
  - the manager is accessible system-wide via the `BlockNodeContext`
- introduce a test `ThreadPoolManager`
  - the test manager is accessible to children of `PluginTestBase`
  - the test manager utilizes a `BlockingSerialExecutor`
  - the test manager cleans up automatically after tests

## Related Issue(s)

Closes #1067


block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
```diff
@@ -0,0 +1,44 @@
...
+     * be positive or 0 (zero)
+     * @param timeUnit the time unit of the keep alive time, must not be null
+     * @param workQueue the work queue to use for the thread pool, must not be
+     * null
+     * @param threadFactory the thread factory to use for the thread pool, must
+     * not be null
+     * @return a new instance of the BlockNodeContext record
+     */
+    @NonNull
+    ExecutorService createExecutorService(
```
jasperpotts, line 37: I think I would take a different design approach here. I would add some high level API and hide implementation details. That way we have single place to tune the implementation details when we get to performance work. So I would add some methods like `createSingleThreadExecutor(String name)` then we will have defaults for thread factory, queue, timeout etc.  Then maybe we have a second method to create an executor with N threads. Hopefully those two will cover all use cases.

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/PrefixedIncrementingThreadFactory.java
```diff
@@ -0,0 +1,70 @@
...
+ * <b>Thread naming example</b>:
+ * Suppose we initialize the factory with the prefix "MyThread".
+ * The counter always starts at 0, so the first thread created will be named:
+ *     <i>"MyThread-0"</i>
+ * The second thread created will be named:
+ *     <i>"MyThread-1"</i>
+ * and so on and so forth.
+ * </pre>
+ */
+public final class PrefixedIncrementingThreadFactory implements ThreadFactory {
```
jasperpotts, line 28: This should be an implementation detail not part of SPI, SPI is for API interfaces not concrete classes. With the proposed design change below this can be a inner class or package private inside DefaultThreadPoolManager.

jasperpotts (changes requested):

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
jsync-swirlds: I almost wonder if it might be better to offer (in addition to "getExecutor" for cases that don't otherwise fit), a "start thread" method (similar to `ExecutorService.submit`) that returns a `Future`, so the mechanism of how the thread is managed is further hidden (so we could, for example, change implementation from `ExecutorService` to `ForkJoinPool` transparently if we choose).
Not sure if that leaves us _too_ abstract, however.

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/PrefixedIncrementingThreadFactory.java
ata-nas: In my initial implementation before the push, it was in fact an inner class to the default manager. I was simply not too sure what to do with this one and how/what the API of the manager must be. But I do agree it should not be in SPI, will move it back as we will simplify the API as well.

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
ata-nas: @jasperpotts thanks for this, I was wondering what level of abstraction we should have, I wired the whole thing up, but I will indeed introduce a higher level of abstraction.

@jsync-swirlds cool idea, just so that I do not misunderstand, are you proposing that we have some pool running internally in the manager that could be submitted the tasks via the "start thread" method, or are you thinking of something else?

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/PrefixedIncrementingThreadFactory.java
ata-nas: P.S. I guess I should think of the manager as something more than a mere factory, maybe that way of looking at it will make for a better decision making.

ata-nas: This factory has been removed in favor of the built in factory builder.

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
ata-nas: I've made some changes. For this PR I want to limit the capabilities of the manager to what I need to use them for. Let's extend it as we see fit when the time arrives 😊

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -84,20 +79,24 @@ public List<Class<? extends Record>> configDataTypes() {
...
-                || archiveConfig.accessKey().isEmpty()) {
-            LOGGER.log(System.Logger.Level.INFO, "Archive plugin is disabled. No access key provided.");
+        } else if (StringUtilities.isBlank(archiveConfig.accessKey())) {
+            LOGGER.log(System.Logger.Level.INFO, "S3 Archive plugin is disabled. No access key provided.");
             return;
-        } else if (archiveConfig.secretKey() == null
-                || archiveConfig.secretKey().isEmpty()) {
-            LOGGER.log(System.Logger.Level.INFO, "Archive plugin is disabled. No secret key provided.");
+        } else if (StringUtilities.isBlank(archiveConfig.secretKey())) {
+            LOGGER.log(System.Logger.Level.INFO, "S3 Archive plugin is disabled. No secret key provided.");
```
ata-nas, line 91: Migrated this to use the "isBlank" check.

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
jasperpotts: @jsync-swirlds I think we want to maintain the ability for plugins to have there own single threaded executors so they can have a simple threading model knowing that everything used by that executor can be single threaded. Expect later we might have bigger pool and maybe a common shared pool or two.

ata-nas:
> Looks good for first step. Think we might want to add metrics or logging or other central tracking later but not sure what we need yet.

@jasperpotts same here, I was also not sure at all what we need for metrics and logging. Good thing is that everything is wired now, so extending this ad hoc will be a breeze.

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
jasperpotts: We will also need to work out how this integrates with Foundation team thread management as that API will be used in state management modules we get from consensus.

AlfredoG87 (approved):
LGTM

PR #1104, opened 2025-05-09
Run tests on module path

jjohannes:
- [ ] TODO requires 0.4.0 release of _hiero-gradle-conventions_ to be finalized


block-node/app/build.gradle.kts
```diff
@@ -23,7 +22,6 @@ tasks.register<JavaExec>("runWithCleanStorage") {
     description = "Run the block node, deleting storage first"
     group = "application"
 
-    modularity.inferModulePath = true
```
jjohannes, line 26: `true` is the default. Hence I removed these redundant configuration lines.

jjohannes:
@AlfredoG87 I figured that we temporarily can add patch code for `com.swirlds.common` to workaround the renaming issue until `0.63` is released. Then this does not block the other updates.

protobuf/src/test/java/.gitignore
jsync-swirlds, line 1: We can also consider suppressing the test generation by PBJ so that we do not create or run those tests at all (it is an option in the PBJ compiler now).
Those tests verify that _PBJ_ is correct, which isn't particularly necessary to run here.

AlfredoG87:
> @AlfredoG87 @jasperpotts With moving tests to the Module Path, the test `BlockNodeAppTest.testMain()` behaves different. The service registry is now available at test time. Also, loading one of a duplicated resource in the root of the module/class path behaves different. It is an unstable thing to rely on in any case. There is a duplication with `app/src/main/resources/app.properties` and `app/src/test/resources/app.properties`. I propose two changes, but please let me know if we should solve this differently:
> 
> * Rename `app/src/test/resources/app.properties` to `app/src/test/resources/app-test.properties` and prefer the latter one if available. This is [similar to what Spring boot supports for such a use case](https://www.baeldung.com/spring-tests-override-properties#springProfile).
> * Adjust `main()` method test code to no longer expect an Exception.

I agree with the approach 👍

AlfredoG87 (approved):
LGTM

AlfredoG87 (approved):

block-node/app/src/main/java/module-info.java
```diff
@@ -14,7 +14,8 @@
     // export configuration classes to the config module
     exports org.hiero.block.node.app to
             com.swirlds.config.impl,
-            com.swirlds.config.extensions;
+            com.swirlds.config.extensions,
```
jsync-swirlds, line 17: This is a bit concerning, as this module _both_ `requires` and `exports to` the `com.swirlds.config.extensions` module.

As noted above, this module really _should not_ be exporting anything, so we probably need to fix whatever is needing us to export here.

Outside the scope of this PR, but I've filed an [issue](https://github.com/hiero-ledger/hiero-block-node/issues/1244) to investigate and resolve.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -128,10 +129,13 @@ public class BlockNodeApp implements HealthFacility {
...
+                ? APPLICATION_TEST_PROPERTIES
+                : APPLICATION_PROPERTIES;
         //noinspection unchecked
         final ConfigurationBuilder configurationBuilder = ConfigurationBuilder.create()
                 .withSource(new AutomaticEnvironmentVariableConfigSource(allConfigDataTypes, System::getenv))
-                .withSources(new ClasspathFileConfigSource(Path.of(APPLICATION_PROPERTIES)))
+                .withSources(new ClasspathFileConfigSource(Path.of(appProperties)))
                 .withConfigDataType(com.swirlds.common.metrics.config.MetricsConfig.class)
                 .withConfigDataType(com.swirlds.common.metrics.platform.prometheus.PrometheusConfig.class)
                 .withConfigDataTypes(allConfigDataTypes.toArray(new Class[0]));
```
jsync-swirlds, line 141: This whole setup feels off.
We should not be aware of testing in production code, and configuration should be "automagic" not needing explicit setup here in the app module...

I wonder if this is a leftover from early PoC efforts before the team had learned enough about the foundation config libraries...

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
```diff
@@ -128,9 +128,8 @@ void testStart() {
 
     @Test
     @DisplayName("Test main method")
-    void testMain() {
-        // The main will try to start the app with default services loader but not find any messaging facility and hence
-        // throw an exception. This tests both the main method and the error handling for missing facility.
-        assertThrows(IllegalStateException.class, () -> BlockNodeApp.main(new String[] {}));
```
jsync-swirlds, line 134: Do we not want to test that main fails when there is no messaging facility?

Seems to me we probably need _both_ of these tests (start with test config and _also_ verify failure when no messaging facility is present).

jsync-swirlds (approved):
Thanks for working through this change.

I've left a few comments, but none are blocking or need to be resolved here; more just questions and items for the block node team to resolve separately.

jsync-swirlds:
> > @AlfredoG87 @jasperpotts With moving tests to the Module Path, the test `BlockNodeAppTest.testMain()` behaves different. The service registry is now available at test time. Also, loading one of a duplicated resource in the root of the module/class path behaves different. It is an unstable thing to rely on in any case. There is a duplication with `app/src/main/resources/app.properties` and `app/src/test/resources/app.properties`. I propose two changes, but please let me know if we should solve this differently:
> > 
> > * Rename `app/src/test/resources/app.properties` to `app/src/test/resources/app-test.properties` and prefer the latter one if available. This is [similar to what Spring boot supports for such a use case](https://www.baeldung.com/spring-tests-override-properties#springProfile).
> > * Adjust `main()` method test code to no longer expect an Exception.
> 
> I agree with the approach 👍

Perhaps not in this PR (as it's not a small addition), but we should definitely _not_ be adding awareness of unit test behaviors in the production `main` method.
Ideally we'd be using the test config builder where needed and have only one properties file supported (which is _not present_ in the main build because all defaults match production settings).

common/src/main/java/org/hiero/block/common/constants/StringsConstants.java
```diff
@@ -8,6 +8,11 @@ public final class StringsConstants {
      */
     public static final String APPLICATION_PROPERTIES = "app.properties";
 
+    /**
+     * File name for application test properties (may override APPLICATION_PROPERTIES)
+     */
+    public static final String APPLICATION_TEST_PROPERTIES = "app-test.properties";
+
```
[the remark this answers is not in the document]
jjohannes: I am happy to not add this. 

As I am not familiar with different _configuration builder_ options, would it maybe be okay to just disable the exiting `testMain()` test in this PR and not do any of these changes here (also in light of your other comment https://github.com/hiero-ledger/hiero-block-node/pull/1104/files#r2124989424).

And then create a follow up ticket for one of you to follow up on?

PR #1113, opened 2025-05-09
docs: Create a logging guidelines document

### LINKED ISSUE #1102: Document logging levels and recommendations
### Story Form

As a Block Node Operator
I want to have useful, actionable, and consistent log data
So that I can effectively manage my block node

### Technical Notes

One important requirement for a well managed system is to ensure all log data is clear, consistent, actionable, and useful.  In support of this goal we intend to clearly and thoroughly document the intent and recommended use for each log level as well as guidelines for ensuring that only necessary log statements are in the code and log statements do not unnecessarily waste resources with excessive or "noisy" logs.

jsync-swirlds:
## Reviewer Notes
* Added logging-guidelines.md to the docs folder
   * Filled in Purpose
   * Filled in Level definitions
   * Filled in recommended usage
   * Filled in things to avoid
   * Mangled formatting until Spotless would accept it
       * Codacy still flags all definitions, though.

## Related Issue(s)
 Resolves #1102

### DIFF
--- a/docs/logging-guidelines.md
+++ b/docs/logging-guidelines.md
@@ -1,0 +1,160 @@
+# Logging Guidelines for Block Node
+
+## Table of Contents
+
+1. [Purpose](#purpose)
+2. [Level Definitions](#level-definitions)
+3. [Recommended Usage](#recommended-usage)
+4. [Thigs to Avoid](#things-to-avoid)
+
+## Purpose
+
+This document seeks to provide a common baseline of understanding for all
+engineers working on the Block Node project. Herein we provide an agreed
+definition for each logging level, recommendations for using log statements,
+and some common anti-patterns that should be avoided.
+
+## Level Definitions
+
+<dl>
+  <dt>ERROR</dt>
+  <dd>Events that indicate the system is entering a failure state and
+    requires immediate attention.<br/>
+    This level almost always triggers a critical alert to the entire operations
+    team, and should be considered only for situations that require immediate
+    attention to prevent catastrophic results.<br/>
+    Good examples might include detecting a condition that will result in loss
+    of the entire local archive, or detection of a _high probability_ that the
+    system has suffered a substantial cybersecurity breach.<br/>
+    <p style='margin-left:2em;margin-top:0em'>Example:<br/>
+        In the "Recent Blocks" plugin, a disk full error might be logged at
+        <code>ERROR</code> level, as that block node cannot continue its
+        primary purpose when the disk is full.
+    </p>
+    </dd>
+  <dt>WARNING</dt>
+  <dd>Events that indicate the system is not healthy, but not immediately
+    failing.<br/>
+    This level almost always triggers a call to the on-call operations engineer
+    and this should be considered whenever writing a log at this level.<br/>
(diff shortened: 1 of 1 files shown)


AlfredoG87 (dismissed):
LGTM, just one question, would be advisable to make a kind of map between FINE, FINER, etc and the Log.Levels described here?

or that should be part of another document, one that describes implementation rather than guidelines.

jsync-swirlds:
> LGTM, just one question, would be advisable to make a kind of map between FINE, FINER, etc and the Log.Levels described here?
> 
> or that should be part of another document, one that describes implementation rather than guidelines.

That's definitely implementation, particularly as we might not be using the `java.util.logging` backend for the `System.Logger` facade forever. If we do switch over to the foundation Logger library, we would have completely different mappings.

In the meantime, that table already exists [here](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/System.Logger.Level.html).

ata-nas (approved):
@jsync-swirlds looks good.

docs/logging-guidelines.md
```diff
@@ -0,0 +1,120 @@
...
+There are a number of common anti-patterns that should be avoided when writing
+log statements.
+
+* Avoid writing log statements in loops.
+  * If you need to log a large amount of information, you are almost certainly
+    misusing logs for some purpose other than operational diagnostics.
+  * Consider why this information might be logged, and explore other options
+    such as using a metrics library or transmitting the information to a
+    dedicated storage system.
+* Avoid log statements that require significant amounts of memory or processing.
```
Nana-EC, line 103: This is a great example of where checking if the log level is supported is applicable. May be worth mentioning here or at the top

docs/logging-guidelines.md
```diff
@@ -0,0 +1,120 @@
...
+    test or production deployments except when highly targeted and
+    strongly warranted.</dd>
+</dl>
+
+## Recommended Usage
+
+Writing log statements is a common and useful practice, but it is important to
+use them judiciously.
+
+* Use the appropriate level for the information being logged.
```
Nana-EC, line 63: Might be a good idea to add an example of a log for each level so developers have a clear reference

Nana-EC (dismissed):
LG.
Left some non-blocking feedback to help make it more clear

docs/logging-guidelines.md
jsync-swirlds: Already mentioned, line 77.

docs/logging-guidelines.md
jsync-swirlds: hmmm... Might be good to add that in the definition list; will see how that looks.

jsync-swirlds: I think I have something that works; although github rendering is less than ideal (it scrubs indentation and other style elements).
If you view the document in the IDE rendered view it reads a lot better.

ata-nas: @Nana-EC that is a good idea indeed.

ata-nas (approved):
@jsync-swirlds good addition to logging examples, I find that quite helpful. Thanks @Nana-EC for the idea.

PR #1120, opened 2025-05-12
docs: Add metrics document

### LINKED ISSUE #1131: Metrics Documentation
### Story Form

### Technical Notes

This ticket is just for a Metrics Document with an overview with all the expected metrics to be found in the app and the provisioned dashboards.

AlfredoG87:
## Reviewer Notes

This document aims to add all defined metrics by the BlockNode App and it's default plugins.

For team members to access, this is the original spredsheeet where we worked and tracked the metrics with @jasperpotts 
https://docs.google.com/spreadsheets/d/16Pt55fydA_W1Fuk9B_GRpoyLpIOgOrzc3nZnkEhoTSk/edit?usp=sharing

Fixes #1131

### DIFF
--- a/docs/block-node/metrics.md
+++ b/docs/block-node/metrics.md
@@ -80,6 +80,10 @@ Observes inbound streams from publishers.
 | Counter | `publisher_blocks_ack_sent`              | Block‑ack messages sent                             |
 | Gauge   | `publisher_latest_block_number_ack_sent` | Latest Block Number Ack Sent from Publisher         |
 | Counter | `publisher_stream_errors`                | Publisher connection streams that end in an error   |
+| Counter | `publisher_blocks_skips_sent`            | Block‑ack skips sent                                |
+| Counter | `publisher_blocks_resend_sent`           | Block Resend messages sent                          |
+| Counter | `publisher_blocks_endofstrem_sent`       | Block End-of-Stream messages sent                   |
+| Counter | `publisher_blocks_endofstrem_received`   | Block End-of-Stream messages received               |
 
 ---
 
@@ -88,13 +92,10 @@ Observes inbound streams from publishers.
 **Plugin:** `stream-subscriber` [block-node-stream-subscriber]`
 Observes outbound streams served to subscribers.
 
-|  Type   |              Name              |                          Description                           |
-|---------|--------------------------------|----------------------------------------------------------------|
-| Counter | `subscriber_historic_to_live`  | Historic‑to‑live stream transitions                            |
-| Counter | `subscriber_live_to_historic`  | Live‑to‑historic stream transitions                            |
-| Gauge   | `subscriber_open_connections`  | Connected subscribers                                          |
-| Gauge   | `subscriber_newest_block_sent` | Newest block number sent (lag indicator vs. publisher inbound) |
-| Counter | `subscriber_errors`            | Errors while streaming to subscribers                          |
+|  Type   |             Name              |              Description              |
+|---------|-------------------------------|---------------------------------------|
+| Gauge   | `subscriber_open_connections` | Connected subscribers                 |
+| Counter | `subscriber_errors`           | Errors while streaming to subscribers |
 
 ---
 
@@ -118,13 +119,13 @@ Measures block‑verification throughput and success rate.
 **Plugin:** `block-providers/files.recent [block-node-blocks-file-recent]`
 Activity and utilization of the recent on‑disk tier.
 
-|  Type   |                Name                |           Description           |
-|---------|------------------------------------|---------------------------------|
-| Counter | `files_recent_blocks_written`      | Blocks written to recent tier   |
-| Counter | `files_recent_blocks_read`         | Blocks read from recent tier    |
-| Counter | `files_recent_blocks_deleted`      | Blocks deleted from recent tier |
-| Gauge   | `files_recent_blocks_stored`       | Blocks stored in recent tier    |
-| Gauge   | `files_recent_total_bytes_storerd` | Bytes stored in recent tier     |
+|  Type   |               Name                |           Description           |
+|---------|-----------------------------------|---------------------------------|
+| Counter | `files_recent_blocks_written`     | Blocks written to recent tier   |
+| Counter | `files_recent_blocks_read`        | Blocks read from recent tier    |
+| Counter | `files_recent_blocks_deleted`     | Blocks deleted from recent tier |
+| Gauge   | `files_recent_blocks_stored`      | Blocks stored in recent tier    |
+| Gauge   | `files_recent_total_bytes_stored` | Bytes stored in recent tier     |
 
 ---
 
@@ -133,28 +134,12 @@ Activity and utilization of the recent on‑disk tier.
 **Plugin:** `block-providers/files.historic [block-node-blocks-file-historic`
 Activity and utilization of the historic on‑disk tier.
 
-|  Type   |                Name                 |            Description            |
-|---------|-------------------------------------|-----------------------------------|
-| Counter | `files_historic_blocks_written`     | Blocks written to historic tier   |
-| Counter | `files_historic_blocks_read`        | Blocks read from historic tier    |
-| Counter | `files_historic_blocks_deleted`     | Blocks deleted from historic tier |
-| Gauge   | `files_historic_blocks_stored`      | Blocks stored in historic tier    |
-| Gauge   | `files_historic_total_bytes_stored` | Bytes stored in historic tier     |
-
----
-
-## cloud.historic
-
-**Plugin:** `block-providers/cloud.historic [block-node-blocks-cloud-historic]`
-Activity and utilization of the cloud on‑disk tier.
-
-|  Type   |                Name                 |          Description           |
-|---------|-------------------------------------|--------------------------------|
-| Counter | `cloud_historic_blocks_written`     | Blocks written to cloud tier   |
-| Counter | `cloud_historic_blocks_read`        | Blocks read from cloud tier    |
-| Counter | `cloud_historic_blocks_deleted`     | Blocks deleted from cloud tier |
-| Gauge   | `cloud_historic_blocks_stored`      | Blocks stored in cloud tier    |
-| Gauge   | `cloud_historic_total_bytes_stored` | Bytes stored in cloud tier     |
+|  Type   |                Name                 |           Description           |
+|---------|-------------------------------------|---------------------------------|
+| Counter | `files_historic_blocks_written`     | Blocks written to historic tier |
+| Counter | `files_historic_blocks_read`        | Blocks read from historic tier  |
+| Gauge   | `files_historic_blocks_stored`      | Blocks stored in historic tier  |
+| Gauge   | `files_historic_total_bytes_stored` | Bytes stored in historic tier   |
 
 ---
 
@@ -163,10 +148,13 @@ Activity and utilization of the cloud on‑disk tier.
 **Plugin:** `s3-archive [hiero-block-node.s3-archive]`
 Tracks long‑term archival jobs that push blocks to S3.
 
-|  Type   |              Name               |              Description               |
-|---------|---------------------------------|----------------------------------------|
-| Counter | `s3_archive_blocks_written`     | Blocks archived to S3                  |
-| Gauge   | `s3_archive_latest_block`       | Latest block number archived           |
-| Counter | `s3_archive_tasks_failed_total` | Failed archival tasks                  |
-| Counter | `s3_archive_tasks_sucess_total` | Successful archival tasks              |
-| Gauge   | `s3_archive_total_bytes_stored` | Total bytes stored in S3 (cost metric) |
+|  Type   |              Name               |               Description               |
+|---------|---------------------------------|-----------------------------------------|
+| Counter | `s3_archive_blocks_written`     | Blocks archived to S3                   |
+| Gauge   | `s3_archive_latest_block`       | Latest block number archived            |
+| Counter | `s3_archive_tasks_failed_total` | Failed archival tasks                   |
+| Counter | `s3_archive_tasks_sucess_total` | Successful archival tasks               |
+| Gauge   | `s3_archive_total_bytes_stored` | Total bytes stored in S3 (cost metric)  |
+| Counter | `s3_archive_chunks_uploaded`    | Chunks uploaded to S3                   |
+| Gauge   | `s3_archive_chunks_opened`      | Open chucks currently                   |
+| Counter | `s3_archive_files_closed`       | Total number of files closed, finished. |
--- a/docs/block-node/metrics.md
+++ b/docs/block-node/metrics.md
@@ -73,7 +73,6 @@ Observes inbound streams from publishers.
 |  Type   |                   Name                   |                     Description                     |
 |---------|------------------------------------------|-----------------------------------------------------|
 | Counter | `publisher_block_items_received`         | Live block items received (sum over all publishers) |
-| Counter | `live_block_items_messaged`              | Legacy metric – kept for compatibility              |
 | Gauge   | `publisher_lowest_block_number_inbound`  | Oldest inbound block number                         |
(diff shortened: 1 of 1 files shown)


docs/block-node/metrics.md
```diff
@@ -0,0 +1,170 @@
...
+## app
+
+**Plugin:** `app`
+Node‑level state and current block numbers.
+
+| Type  |             Name              |              Description              |
+|-------|-------------------------------|---------------------------------------|
+| Gauge | `app_historical_oldest_block` | Oldest block the BN currently stores  |
+| Gauge | `app_historical_newest_block` | Newest block the BN currently stores  |
+| Gauge | `app_state_status`            | 0=Starting, 1=Running, 2Shutting Down |
```
Nana-EC, line 33:
```suggestion
| Gauge | `app_state_status`            | 0=Starting, 1=Running, 2=Shutting Down |
```

docs/block-node/metrics.md
```diff
@@ -0,0 +1,170 @@
...
+---
+
+## Block Access
+
+**Plugin:** `block-access [block-access-service]`
+Observes the block access service that serves requests for single blocks.
+
+|  Type   |                 Name                  |                 Description                 |
+|---------|---------------------------------------|---------------------------------------------|
+| Counter | `single-block-requests`               | Number of single block requests             |
```
Nana-EC, line 44: API name changed so this and below should be get block vs single block
```suggestion
| Counter | `get-block-requests`               | Number of get block requests             |
```

docs/block-node/metrics.md
```diff
@@ -0,0 +1,170 @@
...
+Observes inbound streams from publishers.
+
+|  Type   |                   Name                   |                     Description                     |
+|---------|------------------------------------------|-----------------------------------------------------|
+| Counter | `publisher_block_items_received`         | Live block items received (sum over all publishers) |
+| Gauge   | `publisher_lowest_block_number_inbound`  | Oldest inbound block number                         |
+| Gauge   | `publisher_current_block_number_inbound` | Current block number from primary publisher         |
+| Gauge   | `publisher_highest_block_number_inbound` | Newest inbound block number                         |
+| Gauge   | `publisher_open_connections`             | Connected publishers                                |
+| Counter | `publisher_blocks_ack_sent`              | Block‑ack messages sent                             |
```
Nana-EC, line 80: 🤔, should we rather note the last block acknowledgement sent vs the number of blocks we're acknowledged?
We can probably inform the later from the number of notifications.

docs/block-node/metrics.md
```diff
@@ -0,0 +1,170 @@
...
+---
+
+## Subscriber
+
+**Plugin:** `stream-subscriber` [block-node-stream-subscriber]`
+Observes outbound streams served to subscribers.
+
+|  Type   |              Name              |                          Description                           |
+|---------|--------------------------------|----------------------------------------------------------------|
+| Counter | `subscriber_historic_to_live`  | Historic‑to‑live stream transitions                            |
```
Nana-EC, line 91: Are these referring to the number of active listeners or is thus a count of all listeners that went through the transition in a block node life cycle?

docs/block-node/metrics.md
```diff
@@ -0,0 +1,170 @@
...
+| Counter | `messaging_block_verification_notifications` | Notifications issued after verification   |
+| Counter | `messaging_block_persisted_notifications`    | Notifications issued after persistence    |
+| Gauge   | `messaging_no_of_item_listeners`             | Active item listeners                     |
+| Gauge   | `messaging_no_of_notification_listeners`     | Active notification listeners             |
+| Gauge   | `messaging_item_queue_percent_used`          | Percent of item queue utilised            |
+| Gauge   | `messaging_notification_queue_percent_used`  | Percent of notification queue utilised    |
+
+---
+
+## Publisher
```
Nana-EC, line 68: Q: should we have a metric to help capture publish streams that end in an error?

Nana-EC (commented):
Nice start, did a quick review will circle back.

docs/block-node/metrics.md
AlfredoG87: it is a counter of `transitions` that have occurred, not the active listeners, all transitions.
This metric would in theory help us know how the BN app plugin is being used the most.

The number of active_listeners overall is the Gauge: `subscriber_open_connections`

docs/block-node/metrics.md
AlfredoG87: Good Catch, I think there will be connection and other kind of errors, we need this metric.
`publisher_stream_errors`.
✅ Done

docs/block-node/metrics.md
AlfredoG87: I agree with you, we need a last_block_acked gauge. but not sure we want to get rid of the counter for ack sents. since it might happen that the notification is sent but the publisher (due to a bug or miss).

Metric added as `publisher_latest_block_number_ack_sent` type `Gauge`.

✅ Done.

docs/block-node/metrics.md
AlfredoG87: ✅ Done

docs/block-node/metrics.md
jsync-swirlds: Might also want to track how many `SkipBlock` and `ResendBlock` items sent.
Also how many `EndStream` _received_ and `EndOfStream` sent.
Not sure if we want that level of detail, but perhaps?

docs/block-node/metrics.md
jsync-swirlds: There's basically no place to set this metric, and it would be number of transitions for all subscribers with no differentiation (unless we have labels).
Transitions are not particularly meaningful with the new stream subscriber, as it will switch back and forth very seamlessly (it's no longer a significant impact so it's done whenever one or the other source is "faster").  Switching between the two doesn't indicate slowness, or anything really, other than "this block is immediately available from that source".

The same applies to live-to-historic below.

docs/block-node/metrics.md
```diff
@@ -0,0 +1,172 @@
...
+
+**Plugin:** `stream-subscriber` [block-node-stream-subscriber]`
+Observes outbound streams served to subscribers.
+
+|  Type   |              Name              |                          Description                           |
+|---------|--------------------------------|----------------------------------------------------------------|
+| Counter | `subscriber_historic_to_live`  | Historic‑to‑live stream transitions                            |
+| Counter | `subscriber_live_to_historic`  | Live‑to‑historic stream transitions                            |
+| Gauge   | `subscriber_open_connections`  | Connected subscribers                                          |
+| Gauge   | `subscriber_newest_block_sent` | Newest block number sent (lag indicator vs. publisher inbound) |
```
jsync-swirlds, line 96: This is not meaningful unless we have labels.  Without labels it will jump around as each subscriber sets the same metric for latest block sent.
Example:
1. Subscriber 1 asks for latest onward, receives block 982
2. Subscriber 2 asks for blocks from 0 onward, receives blocks 0-5
3. Subscriber 1 receives block 983
4. Subscriber 2 receives block 6
The metric would record the following values, in order 982, 0, 1, 2, 3, 4, 5, 983, 6.
I struggle to see how such a metric, bouncing somewhat randomly all over the number space, would be useful.

docs/block-node/metrics.md
```diff
@@ -0,0 +1,172 @@
...
+**Plugin:** `block-providers/files.recent [block-node-blocks-file-recent]`
+Activity and utilization of the recent on‑disk tier.
+
+|  Type   |                Name                |           Description           |
+|---------|------------------------------------|---------------------------------|
+| Counter | `files_recent_blocks_written`      | Blocks written to recent tier   |
+| Counter | `files_recent_blocks_read`         | Blocks read from recent tier    |
+| Counter | `files_recent_blocks_deleted`      | Blocks deleted from recent tier |
+| Gauge   | `files_recent_blocks_stored`       | Blocks stored in recent tier    |
+| Gauge   | `files_recent_total_bytes_storerd` | Bytes stored in recent tier     |
```
jsync-swirlds, line 127: Small typo
```suggestion
| Gauge   | `files_recent_total_bytes_stored` | Bytes stored in recent tier     |
```

docs/block-node/metrics.md
```diff
@@ -0,0 +1,172 @@
...
+## files.historic
+
+**Plugin:** `block-providers/files.historic [block-node-blocks-file-historic`
+Activity and utilization of the historic on‑disk tier.
+
+|  Type   |                Name                 |            Description            |
+|---------|-------------------------------------|-----------------------------------|
+| Counter | `files_historic_blocks_written`     | Blocks written to historic tier   |
+| Counter | `files_historic_blocks_read`        | Blocks read from historic tier    |
+| Counter | `files_historic_blocks_deleted`     | Blocks deleted from historic tier |
```
jsync-swirlds, line 140: This _should_ be 0 forever.

docs/block-node/metrics.md
```diff
@@ -0,0 +1,172 @@
...
+## cloud.historic
+
+**Plugin:** `block-providers/cloud.historic [block-node-blocks-cloud-historic]`
+Activity and utilization of the cloud on‑disk tier.
+
+|  Type   |                Name                 |          Description           |
+|---------|-------------------------------------|--------------------------------|
+| Counter | `cloud_historic_blocks_written`     | Blocks written to cloud tier   |
+| Counter | `cloud_historic_blocks_read`        | Blocks read from cloud tier    |
+| Counter | `cloud_historic_blocks_deleted`     | Blocks deleted from cloud tier |
```
jsync-swirlds, line 155: This _should_ also be 0 forever.

docs/block-node/metrics.md
```diff
@@ -0,0 +1,172 @@
...
+|---------|-------------------------------------|-----------------------------------|
+| Counter | `files_historic_blocks_written`     | Blocks written to historic tier   |
+| Counter | `files_historic_blocks_read`        | Blocks read from historic tier    |
+| Counter | `files_historic_blocks_deleted`     | Blocks deleted from historic tier |
+| Gauge   | `files_historic_blocks_stored`      | Blocks stored in historic tier    |
+| Gauge   | `files_historic_total_bytes_stored` | Bytes stored in historic tier     |
+
+---
+
+## cloud.historic
```
jsync-swirlds, line 146: What is this plugin, and how is it different from the s3-archive below?
We don't have a use case for a "store to cloud" _provider_ plugin, that I know of.

docs/block-node/metrics.md
AlfredoG87: @jsync-swirlds, just to be clear? so you are suggesting to remove these 2 metrics? 
`subscriber_historic_to_live` and `subscriber_live_to_historic`

docs/block-node/metrics.md
AlfredoG87: oh no, the idea here is for live streams and the value can only go up. so it will be the faster (sooner) newest block that was sent to any subscriber, I imagine that we can update an atomic long with the latest only if is bigger than the current, and only then update the metric.

docs/block-node/metrics.md
AlfredoG87: Thank you, good catch of a repeated typo 😅

docs/block-node/metrics.md
AlfredoG87: good question, this plugin is not yet implemented, but added it here since the boilerplate is already there, just not implemented. I was not sure either on the use case, maybe @jasperpotts can elaborate on it?

docs/block-node/metrics.md
```diff
@@ -0,0 +1,172 @@
...
+| Counter | `cloud_historic_blocks_deleted`     | Blocks deleted from cloud tier |
+| Gauge   | `cloud_historic_blocks_stored`      | Blocks stored in cloud tier    |
+| Gauge   | `cloud_historic_total_bytes_stored` | Bytes stored in cloud tier     |
+
+---
+
+## s3‑archive
+
+**Plugin:** `s3-archive [hiero-block-node.s3-archive]`
+Tracks long‑term archival jobs that push blocks to S3.
```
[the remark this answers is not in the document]
AlfredoG87: not sure, we are using blocks as a natural boundary for observing metrics, so is consistent across plugins and maybe "easier" to interpret.

docs/block-node/metrics.md
jsync-swirlds: Yes, I don't see how they're useful (at least not until/unless we have labels; and perhaps not even then).
If we intended those to be proxies for "slow" clients, then we should add that specifically via a "slow_subscriber_client_count" metric (as that's something we _might_ be able to determine and record in the stream subscriber plugin).

docs/block-node/metrics.md
jsync-swirlds: All streams are both live and historic; there's no differentiation.  Trying to figure out which client is furthest ahead is tricky, and just tracking a forever-incrementing value could easily lead to a very misleading metric (i.e. is it stopped because we have no subscribers?  Is it going up, but only one of 100 subscribers is functional?, etc...)

As far as a "high-water-mark" metric, I think the last acknowledged is reasonable, but why would we try to separately track latest sent?  I believe it's going to vary all over the place and doesn't usefully indicate health or non-health of any subsystem...

docs/block-node/metrics.md
jsync-swirlds: I don't think it's necessary.  It might be useful as a "wait, why isn't this `0`" indicator, but not sure if that's worth the overhead.

docs/block-node/metrics.md
jsync-swirlds: I think it was originally going to be what because the `s3-archive` plugin (which we did once it became clear that there was no good use-case for "read" from the DR archive, except in very rare circumstance).

docs/block-node/metrics.md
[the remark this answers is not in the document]
jsync-swirlds: The S3 archive doesn't meaningfully work with blocks; but I am thinking that _adding_ these other items for that plugin could be very helpful to understand the actual health and progress for that particular plugin (which is quite different from all the other plugins).

docs/block-node/metrics.md
AlfredoG87: ✅  Have done the suggested changes.

docs/block-node/metrics.md
AlfredoG87: I have removed the whole plugin section metrics.

✅ Done!

docs/block-node/metrics.md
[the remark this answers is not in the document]
AlfredoG87: ✅  Done. 
Added the following metrics:

| Counter | `s3_archive_chunks_uploaded`    | Chunks uploaded to S3                   |
| Gauge   | `s3_archive_chunks_opened`      | Open chucks currently                   |
| Counter | `s3_archive_files_closed`       | Total number of files closed, finished. |

docs/block-node/metrics.md
AlfredoG87: ✅ Done!

docs/block-node/metrics.md
AlfredoG87: ✅  Done! Fixed!

docs/block-node/metrics.md
AlfredoG87: Removed them ✅

docs/block-node/metrics.md
AlfredoG87: ✅ Done!

georgi-l95 (commented):
LG. We might want to add server status, when it's merged. It can track for now recieved requests and we can extend as it goes.

jsync-swirlds (approved):

georgi-l95 (approved):

Nana-EC (approved):
LG
Nice work @AlfredoG87 .
I think in a future PR we can update these to note which metrics are P0 alert worthy.
Depends if we decide to include alert specs in repo or not as as suggestion for operators. Might be good for DevEx

docs/block-node/metrics.md
Nana-EC: When labels 😄

PR #1133, opened 2025-05-13
chore: Finalize the Stream Subscriber plugin

### LINKED ISSUE #1093: `SubscribeBlockStreamHandler` must configure internal thread pool using the `ThreadPoolManager`
### Story Form

As a Block Node Developer
I want to use the `ThreadPoolManager` when setting up thread pools
So that my thread pools are easily manageable and testable

### Technical Notes

- the `SubscribeBlockStreamHandler` should use the `ThreadPoolManager` to set up its thread pool
- this requires care in the unit tests because the plugin test base automatically injects a blocking executor which will prevent testing actual multi-threading; so the plugin test probably needs to inject its own executor to use "real" (likely virtual) threads.

jsync-swirlds:
## Reviewer Notes
* Removed prior subscriber plugin (Obsolete)
* Added virtual thread methods to ThreadPoolManager and default implementation class
* Adjusted grpc and base test plugin superclasses to permit setting the thread pool manager.
* Modified the stream subscriber plugin to use the ThreadPoolManager to get the thread executor.
* Fixed flakey plugin test

## Related Issue(s)
 Resolves #1093.


AlfredoG87 (approved):
LGTM

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
```diff
@@ -17,6 +21,75 @@
...
+    @NonNull
+    @Override
+    public ExecutorService getVirtualThreadExecutor(
+            @Nullable final String threadName, @Nullable final UncaughtExceptionHandler uncaughtExceptionHandler) {
+        if (threadName != null) {
+            return createVirtualThreadExecutor(threadName, uncaughtExceptionHandler);
+        } else if (uncaughtExceptionHandler != null) {
+            if (!VIRTUAL_THREAD_MAP.containsKey(uncaughtExceptionHandler)) {
+                final ExecutorService newExecutor = createVirtualThreadExecutor(threadName, uncaughtExceptionHandler);
+                VIRTUAL_THREAD_MAP.put(uncaughtExceptionHandler, newExecutor);
```
Nana-EC, line 62: nit:
```suggestion
                VIRTUAL_THREAD_MAP.put(uncaughtExceptionHandler, createVirtualThreadExecutor(threadName, uncaughtExceptionHandler));
```

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
```diff
@@ -17,6 +21,75 @@
...
+     *         or a cached (based on exception handler instance) executor service otherwise.
+     */
+    private static ExecutorService createVirtualThreadExecutor(
+            @Nullable final String threadName, @Nullable final UncaughtExceptionHandler uncaughtExceptionHandler) {
+        Thread.Builder factoryBuilder = Thread.ofVirtual();
+        if (uncaughtExceptionHandler != null) {
+            factoryBuilder.uncaughtExceptionHandler(uncaughtExceptionHandler);
+        }
+        if (threadName != null) {
+            factoryBuilder.name(threadName, 0);
```
Nana-EC, line 85: Q: Should we check for conflicts of existence?
For example if I called `createVirtualThreadExecutor()` twice with the same `threadName` what would the expected behavior be?

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -191,7 +191,6 @@ private void sendHistoricalBlocks() throws ParseException {
         // Don't send anything from history if this stream is interrupted, has sent
         // every requested block, or we can send the next block from "live".
         while (isHistoryPermitted()) {
-            LOGGER.log(Level.TRACE, "Sending historical block {0} to {1}", nextBlockToSend, handlerName);
```
Nana-EC, line 194: In line with the teams discussions on logs I see why we are removing a few of these.
However, with removal of all of them is there a log that gives an operator viewing logs a sending that a block was sent out?
Or do we believe it's too noisy to have at all and operators should rely on metrics to see things flowing.
Issues observed at a handler level of course would show up an appropriate error message.

Nana-EC (commented):
Looking good.
1 nit and 2 questions

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
georgi-l95: If i remember correctly 0 indicates the number it starts with. So if  we pass subscriber for threadName and 0. First one should be - subsriber0, next one subscriber1 and so on. 🤔

ata-nas: Yes, each builder will produce it's own factory and the number is indeed the start number, appended as a suffix. If no number is provided, there will not be any counters (that is for the platform builder, I think here should be the same, have not looked for diffs in the virtual thread factory).

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
ata-nas, line 1: With the new addition of the map that collects the pools, should we not consider to make the `ThreadPoolManager` an `AutoCloseable` so that we have the shutdown logic (shutdown the cached pools), or should shutting down of pools remain a responsibility of the user of the thread pool manager?

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
jsync-swirlds: Thread names are not unique; if you create two factories with the same prefix and start number, then you'll get threads with the same name, but thread names are only visible when debugging locally, so impact is minimal (in my opinion).

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
ata-nas: P.S. maybe not an auto-closeable, but a `shutdown` method? Depending on what fits better.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
jsync-swirlds: Sending out thousands of "I sent a block" messages in logs is nothing but noise.  Operators should not care about individual blocks sent (and TRACE would never be enabled in production anyway).  Metrics are how we expose the _count_ of messages sent, which is about the only thing I would expect operators to, potentially, care about.

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
jsync-swirlds: I don't see a need to add `AutoCloseable` because the thread manager should not be used in a try-with-resources; it should always live far beyond a single method.
The manager _should_ live the lifetime of the block node app process, so shutdown isn't a significant concern, in my opinion.
If we _do_ want a shutdown process, we might add a `shutdownAllServices` method that can be called affirmatively during orderly shutdown.

ata-nas: @jsync-swirlds indeed, the auto-closeable does not fit... and yes, a shutdown method would be a good idea indeed if we ever find the need for it.

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
jsync-swirlds: The variable prevents spotless mangling the line.
In my view, it's more compact and clear with the variable than what spotless does.

settings.gradle.kts
```diff
@@ -21,7 +21,6 @@ javaModules {
         module("health") { artifact = "block-node-health" }
         module("messaging") { artifact = "facility-messaging" }
         module("publisher") { artifact = "block-node-publisher" }
-        // module("subscriber") { artifact = "block-node-subscriber" }
         module("stream-subscriber") { artifact = "block-node-stream-subscriber" }
```
jasperpotts, line 24: would it make sense to rename plugin just "subscriber" or rename publisher to "stream-publisher" to match?

ata-nas: I personally like the `stream-publisher/subscriber` pair. Just an opinion.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
ata-nas: +1 for the metrics point. A lot of the logs that people generally make are indeed just metrics and not logs. Good way to look at it.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
ata-nas, line 1: There appear to be unused fields:
- `historicToLiveStreamTransitions`
- `liveToHistoricStreamTransitions`

Comment for awareness.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
ata-nas, line 1: There appears to be an unused field:
- `pluginConfiguration`

Also, what are we doing with the `clientHandlers`? It appears that we use it to add items to the collection, but nothing else?

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
```diff
@@ -17,6 +21,75 @@
  * implementation is used systemwide to manage the thread pools.
  */
 final class DefaultThreadPoolManager implements ThreadPoolManager {
+    private static final ExecutorService DEFAULT_VIRTUAL_THREAD_EXECUTOR = Executors.newVirtualThreadPerTaskExecutor();
+    private static final Map<UncaughtExceptionHandler, ExecutorService> VIRTUAL_THREAD_MAP = new ConcurrentHashMap<>();
```
ata-nas, line 25: Curious, why have we opted to use an `UncaughtExceptionHandler` as key? It is not always null. Interesting approach.

P.S. I did read the docs from L:45-51

ata-nas (approved):
@jsync-swirlds looks good. I had a couple questions, for awareness mostly. A couple of fields were maybe supposed to be used, but are left unused, or we plan to use them in a follow up?

georgi-l95 (dismissed):

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
Nana-EC: I"m not suggesting a change, just calling out to make sure we've mapped to the appropriate metric here.
FYI to @AlfredoG87 to confirm coverage that these logs would relate to.

settings.gradle.kts
Nana-EC: I like `stream-publisher` and `stream-subscriber`, it's consistent and explicit.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
jsync-swirlds: Those metrics are, questioned.
1. It's hard to see how they're valuable in a structure that may switch between live and historic as often as every block (depending on client timing) and does so seamlessly.
2. There isn't a particularly clear point to where they might be updated.
There might still be some value, however, so they may be added back in later.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
jsync-swirlds: Good catch, removed.

ata-nas (approved):
🔥

AlfredoG87 (approved):
LGTM

Nana-EC (approved):
LG thanks

PR #1135, opened 2025-05-13
chore: Improve block node protobuf devex

### LINKED ISSUE #1095: Add gradle protobuf publish logic
### Story Form

As a block node developer
I want to publish Java protobuf artifacts
So that other Java community products can take my API as a dependency to build against

### Technical Notes

We need to add a gradle process that will take the protobuf and generates class and publish them so that consensus node and mirror node initially can take versioned dependencies on block node proto without having to copy files.

Nana-EC:
## Reviewer Notes
There were a few areas to improve protobuf logic.

In this PR we
- add publishing logic to provide a single directory `block-node-protobuf` with all the protobuf needed to interface with the Block Node covering both BN and CN proto
- switch protobuf dependency logic to pull from `hiero-ledger/hiero-consensus-node` instead of `hedera-protobufs`
- update protobuf version and resolve old references
- match the protobuf file hierarchy in CN repo to make it easier for devs working across repos

## Note for reviewers
The pull of the whole CN repo is undesirable but gets us the latest proto without taking jar dependencies. I'd like this to be a temporary workaround and we can explore consuming a similar proto source artifact from CN once they adopt a similar update

## Related Issue(s)
Fixes #1095


jsync-swirlds (commented):
A few questions about possible alternatives.

One significant issue with the "flatten" process where the output folders are incorrectly retained, which resulted in changing import paths (something we _must not_ do).

.github/workflows/release-push-image.yaml
```diff
@@ -45,6 +50,53 @@ jobs:
...
+          fetch-depth: 0
+
+      - name: Install JDK
+        uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4.7.1
+        with:
+          distribution: ${{ env.JAVA_DISTRIBUTION }}
+          java-version: ${{ env.JAVA_VERSION }}
+
+      - name: Build
+        run: ${GRADLE_EXEC} clean assemble
```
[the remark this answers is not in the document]
Nana-EC: Good question.
A build might not be needed.
I'll look into the earliest task that configures the SourceSet.

.github/workflows/release-push-image.yaml
```diff
@@ -45,6 +50,53 @@ jobs:
       java-distribution: ${{ inputs.java-distribution || 'temurin' }}
       java-version: ${{ inputs.java-version || '21.0.6' }}
 
+  package-block-node-protobuf:
+    timeout-minutes: 30
```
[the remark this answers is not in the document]
Nana-EC: Agreed, leftover from a copy. 5 seems more than enough

.github/workflows/release-push-image.yaml
```diff
@@ -45,6 +50,53 @@ jobs:
...
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+
+      - name: Extract CN proto from Jar
+        run: ${GRADLE_EXEC} :block-node-protobuf:extractProtoFromConsensusNodeProtoJar
+
+      - name: Export BN proto
+        run: ${GRADLE_EXEC} :block-node-protobuf:exportBlockNodeProto
+
+      - name: Produce artifact
+        run: tar -czf block-node-protobuf-${{ env.VERSION }}.tgz -C ${{ github.workspace }}/block-node-protobuf/build/block-node-protobuf .
```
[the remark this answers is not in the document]
Nana-EC: Yeah, good point. Could essentially replace this with a script that does the git clone logic  and tars.
Indeed this would totally skip Gradle need.
Let me look into it.

.github/workflows/release-push-image.yaml
```diff
@@ -109,14 +161,49 @@ jobs:
...
+
+      - name: Extract version
+        id: extract_version
+        run: |
+          VERSION=$(cat version.txt)
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+
+      - name: Publish protobuf jar to maven local repo
+        if: ${{ env.INPUT_PUBLISH == 'true' }}
+        run: ${GRADLE_EXEC} :block-node-protobuf:releaseMavenCentral -PpublishingPackageGroup=org.hiero.block --no-configuration-cache
```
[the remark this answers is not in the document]
Nana-EC: It's publishing the compiled Java

protobuf/build.gradle.kts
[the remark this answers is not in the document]
Nana-EC: Yep, the end result is the goal.
I'm hoping to refine this process and have it replicated on CN at least.
The circular dependency is unavoidable given CN publishes the stream but then also communicates with BN over gRPC.

Agreed and similar to the previous comment. Looking into it.
I guess with this I don't **have to** optimize the GitClone Task after all 😅

protobuf/src/main/proto/org/hiero/block/api/block_access_service.proto
```diff
@@ -7,7 +7,7 @@ option java_package = "org.hiero.block.api.protoc";
 // <<<pbj.java_package = "org.hiero.block.api">>> This comment is special code for setting PBJ Compiler java package
 option java_multiple_files = true;
 
-import "stream/block.proto";
+import "block/stream/block.proto";
```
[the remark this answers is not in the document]
Nana-EC: My thinking here was the opposite. More so that we had it wrong the first time with the depth of the imports.
Initially I thought it was a BN mistake but seems it was rather the hedera-protbuf imports hierarchy didn't match the folder structure.
In terms of DevEx it feels clear for the hierarchy of imports to match the view from the root path of the given product.
If not it's odd that developers have to realize that a separate hidden process does a flatmap.

CN node seems to have realized this and also resolve this issue.
This matches that and makes it more consistent going forward.
Even if the proto is now split across repos it would be nice so say you can put them all together in a single dir and have a clear idea of the proto layout across the products

protobuf/build.gradle.kts
```diff
@@ -17,37 +17,131 @@ tasks.withType<JavaCompile>().configureEach {
...
+        description =
+            "Copies the protobuf files from block api to a combined CN & BN block-streams related protobuf directory"
+        group = "protobuf"
+
+        from(
+            layout.buildDirectory.dir(
+                "hiero-consensus-node/hapi/hedera-protobuf-java-api/src/main/proto/block"
+            )
+        )
+        into(layout.buildDirectory.dir("block-node-protobuf/block"))
```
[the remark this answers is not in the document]
Nana-EC: See comment before, I think it's best to maintain the structure as we acquired it so it's easy for devs to follow the files they may find in one product to the origin product specification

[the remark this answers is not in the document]
Nana-EC: Same response for `/platform`, `/services` and `/streams`

protobuf/build.gradle.kts
```diff
@@ -17,37 +17,131 @@ tasks.withType<JavaCompile>().configureEach {
...
+                .asFile
+                .delete()
+        }
+        doLast {
+            localCloneDirectory
+                .file("hapi/hedera-protobuf-java-api/src/main/proto/sdk/transaction_list.proto")
+                .get()
+                .asFile
+                .delete()
+        }
```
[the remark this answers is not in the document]
Nana-EC: Should have added a comment.
Ease of build. When they are pulled in there are protoc conflicts requiring more dependency management.
The other reason is I didn't want any files that aren't applicable to CN.
Removing these files removes the `/mirror` and `/sdk` proto.

For instance in the git clone follow up i'll do i'll only clone these folders and long term CN should probably remove those dirs from the repo and they should only live in the MN and SDK repos.

protobuf/src/main/proto/org/hiero/block/api/block_access_service.proto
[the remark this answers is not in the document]
jsync-swirlds: Surprising; I had not noticed they changed every import in their repo (consequences of having the API mingled in with a huge mass of other code, it's very hard to follow changes to the API documents in all the noise).
Personally I think that's the least good option, as it adds a very artificial division in the _published_ file structure (basically exposing some historical team structures that no longer exist).

It's not something worth debating, however, so I suppose we will just have to match what they've done.

protobuf/build.gradle.kts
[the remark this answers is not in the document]
jsync-swirlds: I see, it seems `consensus-node-repo` made some changes that were not well communicated.

Thanks for the clarification.

protobuf/build.gradle.kts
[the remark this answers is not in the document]
jsync-swirlds: A comment would be very helpful, particularly to know those APIs require additional dependencies.

.github/workflows/release-push-image.yaml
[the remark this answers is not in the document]
Nana-EC: NO longer applicable with update

.github/workflows/release-push-image.yaml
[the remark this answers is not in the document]
Nana-EC: Updated

.github/workflows/release-push-image.yaml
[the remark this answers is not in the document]
Nana-EC: Updated swapping for script run instead

protobuf/build.gradle.kts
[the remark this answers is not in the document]
Nana-EC: Updated.

georgi-l95 (dismissed):

protobuf/build.gradle.kts
```diff
@@ -17,131 +15,61 @@ tasks.withType<JavaCompile>().configureEach {
...
+
+        // clean up intermediate files generated from build-bn-proto.sh run
+        commandLine(
+            "rm",
+            "-rf",
+            "${layout.projectDirectory}/block-node-protobuf",
+            "&&",
+            "rm",
+            "-rf",
+            "${layout.projectDirectory}/hiero-consensus-node",
```
jsync-swirlds, line 50: Could elide this "clean up" step if we modified the script to (as a default, subject to override) write all source, temporary, intermediate, and output files into the build subfolder?

protobuf/scripts/build-bn-proto.sh
```diff
@@ -0,0 +1,138 @@
...
+if [[ -z $cleanup ]]; then
+    cleanup=false
+fi
+if [[ -z $include_bn_api ]]; then
+    include_bn_api=true
+fi
+if [[ -z $bn_api_path ]]; then
+    echo "Set 'bn_api_path' default to 'src/main/proto/org/hiero/block/api'"
+    bn_api_path="src/main/proto/org/hiero/block/api"
+fi
```
jsync-swirlds, line 75: It might be possible to simplify this using variable defaults:
```shell
    bn_api_path="${bn_api_path:-src/main/proto/org/hiero/block/api}"
```

AlfredoG87 (approved):
LGTM, there is some good feedback from Joseph though nothing blocking. will re-approve if needed.

protobuf/build.gradle.kts
Nana-EC: Agreed, I tried that in a previous attempt but it was causing some build failures that were hindering progress.
I'll circle back. 
If it's not a quick fix I'll follow with a subsequent ticket to capture it

protobuf/scripts/build-bn-proto.sh
Nana-EC: Good call.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/SimpleTestBlockItemBuilder.java
```diff
@@ -113,11 +116,11 @@ public static BlockItem sampleRoundHeader(final long roundNumber) {
...
 
     /**
-     * Create an EventHeader with a large 0.5MB signature data.
+     * Create an EventHeader with no parents and no signature middle bit.
      */
     public static BlockItem sampleLargeEventHeader() {
-        return new BlockItem(
-                new OneOf<>(ItemOneOfType.EVENT_HEADER, new EventHeader(EventCore.DEFAULT, RANDOM_HALF_MB)));
+        return new BlockItem(new OneOf<>(
+                ItemOneOfType.EVENT_HEADER, new EventHeader(EventCore.DEFAULT, Collections.emptyList(), false)));
```
jsync-swirlds, line 123: Is there no value to a non-empty event?
Just wondering if we should have an issue to go back through and re-add the option to produce more fully filled in block streams (also, sub-block batching).

jsync-swirlds (dismissed):
Looks good, a few non-blocking items we might tackle in follow-up.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/SimpleTestBlockItemBuilder.java
Nana-EC: Good call, we should add coverage here. Will capture an item to confirm and update

.github/workflows/release-push-image.yaml
```diff
@@ -109,14 +147,49 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
           platforms: linux/amd64, linux/arm64
-          push: true
+          push: ${{ env.INPUT_PUBLISH == 'true' }}
```
andrewb1269hg, line 150: Same as above.

```suggestion
          push: "${{ github.event_name != 'workflow_dispatch' || github.event.inputs.publish == 'true' }}"
```

.github/workflows/release-push-image.yaml
```diff
@@ -45,6 +50,39 @@ jobs:
...
+        run: |
+          VERSION=$(cat version.txt)
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+
+      - name: Produce artifact
+        working-directory: protobuf
+        run: scripts/build-bn-proto.sh -t "efb0134e921b32ed6302da9c93874d65492e876f" -v ${{ env.VERSION }} -o "block-node-protobuf" # CN 0.62.2
+
+      - name: Upload artifact
+        if: ${{ env.INPUT_PUBLISH == 'true' }}
```
andrewb1269hg, line 80: I don't see `INPUT_PUBLISH` being set anywhere. I'd use the variable directly from the input, like this:
```suggestion
        if: ${{ github.event_name != 'workflow_dispatch' || github.event.inputs.publish == 'true' }}
```

.github/workflows/release-push-image.yaml
```diff
@@ -109,14 +147,49 @@ jobs:
...
+        run: ${GRADLE_EXEC} clean assemble
+
+      - name: Extract version
+        id: extract_version
+        run: |
+          VERSION=$(cat version.txt)
+          echo "VERSION=${VERSION}" >> $GITHUB_ENV
+
+      - name: Publish protobuf jar to maven local repo
+        if: ${{ env.INPUT_PUBLISH == 'true' }}
```
andrewb1269hg, line 190:
```suggestion
        if: ${{ github.event_name != 'workflow_dispatch' || github.event.inputs.publish == 'true' }}
```

.github/workflows/release-push-image.yaml
```diff
@@ -221,6 +294,7 @@ jobs:
           helm package charts/block-node-server
 
       - name: Push helm chart
+        if: ${{ env.INPUT_PUBLISH == 'true' }}
```
andrewb1269hg, line 297:
```suggestion
        if: ${{ github.event_name != 'workflow_dispatch' || github.event.inputs.publish == 'true' }}
```

.github/workflows/release-push-image.yaml
```diff
@@ -260,5 +334,6 @@ jobs:
           helm package charts/blockstream-simulator
 
       - name: Simulator Chart - Push helm chart
+        if: ${{ env.INPUT_PUBLISH == 'true' }}
```
andrewb1269hg, line 337:
```suggestion
        if: ${{ github.event_name != 'workflow_dispatch' || github.event.inputs.publish == 'true' }}
```

.github/workflows/release-push-image.yaml
```diff
@@ -180,7 +253,7 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
           platforms: linux/amd64, linux/arm64
-          push: true
+          push: ${{ env.INPUT_PUBLISH == 'true' }}
```
andrewb1269hg, line 256:
```suggestion
          push: "${{ github.event_name != 'workflow_dispatch' || github.event.inputs.publish == 'true' }}"
```

.gitignore
```diff
@@ -59,3 +59,7 @@ server/data/
 
 # manual test files
 server/src/test/resources/test_output/
+
+# protobuf artifcate related dirs
+protobuf/block-node-protobuf/
+protobuf/hiero-consensus-node/
```
andrewb1269hg, line 65:
```suggestion
protobuf/hiero-consensus-node/

```

andrewb1269hg (changes requested):
See suggestions for details.

protobuf/scripts/build-bn-proto.sh
Nana-EC: Updated with suggestion

.gitignore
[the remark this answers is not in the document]
Nana-EC: Updated with suggestion

.github/workflows/release-push-image.yaml
Nana-EC: Updated with suggestion

.github/workflows/release-push-image.yaml
Nana-EC: Updated with suggestion

.github/workflows/release-push-image.yaml
Nana-EC: Updated with suggestion

.github/workflows/release-push-image.yaml
Nana-EC: Updated with suggestion

.github/workflows/release-push-image.yaml
Nana-EC: Updated with suggestion

.gitignore
Nana-EC: Updated with suggestion

andrewb1269hg (approved):
Approve workflow and git ignore files.

jsync-swirlds (approved):

PR #1137, opened 2025-05-13
feat: create server status plugin

### LINKED ISSUE #781: External Test Suites
### Persona

As a BlockNode Operator

### Request

I want to be sure that the service status api and persistance is working on an automated test

### Goal

So that when changes are performed to the Block-Node server, that unintentionally break this feature, we don't merge the changes until fixed.

### Technical Notes

Use the `suites` project and define a new set of scenarios:

Scenario 1: Query ServerStatus and check with expected data based on test details.
Scenario 2: Restart BlockNode Server, then Query ServerStatus and verify that response matches expected info.

### LINKED ISSUE #1111: Create Server Status Plugin
### Story Form

As a Block Node Operator
I want to be able to call unary rpc service - `BlockNodeService`
So that can receive server status information about this Block Node

### Technical Notes

We need to introduce new plugin called server-status, which opens new service and handles requests for the status of the Block Node.

georgi-l95:
## Reviewer Notes
This PR aims to:
- Create new plugin responsible for opening a unary endpoint for serving `ServerStatusRequest`.
- Handling of this requests is responsibility of the plugin, which returns:
```
message ServerStatusResponse {
    uint64 first_available_block = 1;
    uint64 last_available_block = 2;
    bool only_latest_state = 3;
    BlockNodeVersions version_information = 4;
}
```
- Add unit tests validating the correct behavior, using already existing test fixtures.
- Add E2E tests validating the behaviour of the endpoint.

Left out for next PRs:
- Handling of `only_latest_state` information, for now we just set it to false.
- Handling the construction of BlockNodeVersions object. We currently does not have this information at hand. Will require some change in order to be availble across the project. 

## Related Issue(s)
Fixes #1111
Fixes #781

### DIFF
--- a/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
+++ b/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
@@ -1,0 +1,125 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.server.status;
+
+import static java.lang.System.Logger.Level.DEBUG;
+import static java.util.Objects.requireNonNull;
+
+import com.hedera.pbj.runtime.grpc.GrpcException;
+import com.hedera.pbj.runtime.grpc.Pipeline;
+import com.hedera.pbj.runtime.grpc.Pipelines;
+import com.hedera.pbj.runtime.grpc.ServiceInterface;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.swirlds.metrics.api.Counter;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Arrays;
+import java.util.List;
+import org.hiero.block.api.ServerStatusRequest;
+import org.hiero.block.api.ServerStatusResponse;
+import org.hiero.block.api.protoc.BlockNodeServiceGrpc;
+import org.hiero.block.node.spi.BlockNodeContext;
+import org.hiero.block.node.spi.BlockNodePlugin;
+import org.hiero.block.node.spi.ServiceBuilder;
+import org.hiero.block.node.spi.historicalblocks.HistoricalBlockFacility;
+
+/**
+ * Plugin that implements the BlockNodeService and provides the 'serverStatus' RPC.
+ */
+public class ServerStatusServicePlugin implements BlockNodePlugin, ServiceInterface {
+    /** The logger for this class. */
+    private final System.Logger LOGGER = System.getLogger(getClass().getName());
+    /** The block provider */
+    private HistoricalBlockFacility blockProvider;
+    /** Counter for the number of requests */
+    private Counter requestCounter;
+    /** Counter for the number of responses Success */
+    private Counter responseCounterSuccess;
+    /** Counter for the number of responses Failed */
+    private Counter responseCounterFailed;
+
+    private ServerStatusResponse handleServiceStatusRequest(@NonNull final ServerStatusRequest request) {
+        LOGGER.log(DEBUG, "Received ServerStatusRequest: {0}", request);
+        requestCounter.increment();
+
+        return new ServerStatusResponse(UNKNOWN_BLOCK_NUMBER, UNKNOWN_BLOCK_NUMBER, false, null);
+    }
+
+    // ==== BlockNodePlugin Methods ====================================================================================
+    @Override
+    public String name() {
+        return "ServerStatusServicePlugin";
+    }
+
+    @Override
+    public void init(@NonNull final BlockNodeContext context, @NonNull final ServiceBuilder serviceBuilder) {
+        requireNonNull(serviceBuilder);
+        this.blockProvider = requireNonNull(context.historicalBlockProvider());
+
+        // Create the metrics
+        requestCounter = context.metrics()
+                .getOrCreate(new Counter.Config(METRICS_CATEGORY, "server-status-requests")
+                        .withDescription("Number of server status requests"));
+
+        // Register this service
+        serviceBuilder.registerGrpcService(this);
+    }
+    // ==== ServiceInterface Methods ===================================================================================
+    /**
+     * BlockNodeService methods define the gRPC methods available on the BlockNodeService.
+     */
+    enum BlockNodeServiceMethod implements Method {
+        /**
+         * The serverStatus method retrieves the status of this Block Node.
+         */
+        serverStatus
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @NonNull
+    @Override
+    public String serviceName() {
+        String[] parts = fullName().split("\\.");
+        return parts[parts.length - 1];
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @NonNull
+    @Override
+    public String fullName() {
+        return BlockNodeServiceGrpc.SERVICE_NAME;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @NonNull
+    @Override
+    public List<Method> methods() {
+        return Arrays.asList(BlockNodeServiceMethod.values());
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * This is called each time a new request is received.
+     */
+    @NonNull
+    @Override
+    public Pipeline<? super Bytes> open(
+            @NonNull Method method, @NonNull RequestOptions requestOptions, @NonNull Pipeline<? super Bytes> pipeline)
+            throws GrpcException {
+        final BlockNodeServiceMethod blockNodeServiceMethod = (BlockNodeServiceMethod) method;
+        return switch (blockNodeServiceMethod) {
+            case serverStatus:
+                yield Pipelines.<ServerStatusRequest, ServerStatusResponse>unary()
+                        .mapRequest(ServerStatusRequest.PROTOBUF::parse)
+                        .method(this::handleServiceStatusRequest)
+                        .mapResponse(ServerStatusResponse.PROTOBUF::toBytes)
+                        .respondTo(pipeline)
+                        .build();
+        };
+    }
+}
--- a/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
+++ b/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
@@ -1,21 +1,13 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.server.status;
 
-import static java.lang.System.Logger.Level.DEBUG;
 import static java.util.Objects.requireNonNull;
 
-import com.hedera.pbj.runtime.grpc.GrpcException;
-import com.hedera.pbj.runtime.grpc.Pipeline;
-import com.hedera.pbj.runtime.grpc.Pipelines;
-import com.hedera.pbj.runtime.grpc.ServiceInterface;
-import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.metrics.api.Counter;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import java.util.Arrays;
-import java.util.List;
+import org.hiero.block.api.BlockNodeServiceInterface;
 import org.hiero.block.api.ServerStatusRequest;
 import org.hiero.block.api.ServerStatusResponse;
-import org.hiero.block.api.protoc.BlockNodeServiceGrpc;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
 import org.hiero.block.node.spi.ServiceBuilder;
@@ -24,7 +16,7 @@ import org.hiero.block.node.spi.historicalblocks.HistoricalBlockFacility;
 /**
  * Plugin that implements the BlockNodeService and provides the 'serverStatus' RPC.
  */
-public class ServerStatusServicePlugin implements BlockNodePlugin, ServiceInterface {
+public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServiceInterface {
     /** The logger for this class. */
     private final System.Logger LOGGER = System.getLogger(getClass().getName());
     /** The block provider */
@@ -40,8 +32,7 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, ServiceInterf
      * @param request the request containing the available blocks, state snapshot status and software version
      * @return the response containing the block or an error status
      */
-    private ServerStatusResponse handleServiceStatusRequest(@NonNull final ServerStatusRequest request) {
-        LOGGER.log(DEBUG, "Handling ServerStatusRequest");
+    public ServerStatusResponse serverStatus(@NonNull final ServerStatusRequest request) {
         requestCounter.increment();
 
         final ServerStatusResponse.Builder serverStatusResponse = ServerStatusResponse.newBuilder();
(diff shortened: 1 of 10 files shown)


block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @NonNull
+    @Override
+    public String serviceName() {
+        String[] parts = fullName().split("\\.");
+        return parts[parts.length - 1];
```
ata-nas, line 104: I've seen this in some other places, I think this should become an utility maybe?

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -0,0 +1,146 @@
...
+     * @param request the request containing the available blocks, state snapshot status and software version
+     * @return the response containing the block or an error status
+     */
+    private ServerStatusResponse handleServiceStatusRequest(@NonNull final ServerStatusRequest request) {
+        LOGGER.log(DEBUG, "Handling ServerStatusRequest");
+        requestCounter.increment();
+
+        final ServerStatusResponse.Builder serverStatusResponse = ServerStatusResponse.newBuilder();
+        final long firstAvailableBlock = blockProvider.availableBlocks().min();
+        final long lastAvailableBlock = blockProvider.availableBlocks().max();
```
ata-nas, line 49: I have a general question. Do we have to think about gaps? Here a hypothetical scenario:

1. We have blocks available: 1, 2, 3, 5, 6 ,7
1. Min will be 1
1. Max will be 7
1. If we query for 4 however, we will not see it

Should we generally extend these requests to return ranges in some form which then could be parsed?
This is something for the future, I was wondering if this is relevant in general?

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -0,0 +1,146 @@
...
+
+        // TODO(#579) Should get from state config or status, which would be provided by the context from responsible
+        // facility
+        boolean onlyLatestState = false;
```
ata-nas, line 53: Question: what is the meaning of this boolean?

ata-nas (commented):
@georgi-l95 looks good. A couple of general questions.

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
```diff
@@ -0,0 +1,125 @@
...
+        //        sendBlocks(5);
+        final ServerStatusRequest request = ServerStatusRequest.newBuilder().build();
+        toPluginPipe.onNext(ServerStatusRequest.PROTOBUF.toBytes(request));
+        assertEquals(1, fromPluginBytes.size());
+
+        final ServerStatusResponse response = ServerStatusResponse.PROTOBUF.parse(fromPluginBytes.getFirst());
+
+        assertNotNull(response);
+        assertEquals(0, response.firstAvailableBlock());
+        assertEquals(0, response.lastAvailableBlock());
```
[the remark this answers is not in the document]
georgi-l95: Ah, thank you for noticing. Should be -1L, yes. I forgot it from some previous testing that I did 👍

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
georgi-l95: This is indicating whether this block node offers the latest state snapshot. However we don't have state implemented currently, so it's irrelevant. If we don't set it, it will automatically be false. I left it here for visibility with the TODO comment.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
georgi-l95: Well, I think it's a valid consideration; however, this endpoint wouldn't care. I mean, the StatusServer query only aims to provide information about what should be available in the block-node. If something is missing, it's the responsibility of another plugin. There's currently an issue related to filling gaps—generally, if we say we have data between 0 and 7, then we should have it.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
georgi-l95: Good point, opened a ticket to track this #1158

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
jsync-swirlds: Agreed.  Gaps are the responsibility of the backfill plugin (to be developed).  The status service is intended to provide the edges (earliest and latest); not every detail.

That said, nothing prevents us from adding an extension in the future if there is demand.

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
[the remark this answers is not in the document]
jsync-swirlds: Generally `-1` represents "the highest possible block number" (because we prefer to use unsigned values in the protobuf, which are bitwise transliterated with Java `long`) and is presented to (non-java) API clients as `uint64_max`.

We use `-1L` in Java code because we should never reach that high of a block number (at least, not for an hundred billion years or so).

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -0,0 +1,145 @@
...
+
+        // TODO(#1139) Should get construct a block node version object from application config, which would be provided
+        // by
+        // the context from responsible facility
```
jsync-swirlds, line 56: I think this plugin _is_ the responsible facility.
The server status is the only place block node version is used, and should probably be filled in by the build from `version.txt`.

I don't think that becomes config so much as the build can insert the value into the code (as it's always constant for a given build output). That might require some build changes, however, so perhaps it would make sense to have it in config at least temporarily. Thoughts?

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
Nana-EC: Good question.
The ranges could be unbounded as the block history on the server increases.
Fo this I agree status should be simple and say min and max only for now.
I think the API docs we note there may be gaps which will be surfaced by get requests.
Where it becomes interesting is for streaming.
A block node that has a gap has to share with the consumer of the gap and potentially the next available block they can request.
Something we should discuss as even with a backfill plugin there may be occasions where a BN serves blocks before the gaps are filled.

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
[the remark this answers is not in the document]
Nana-EC: > Generally -1 represents "the highest possible block number" (because we prefer to use unsigned values in the protobuf, which are bitwise transliterated with Java long) and is presented to (non-java) API clients as uint64_max.

Good points but this should be an implementation details of the node and should be exposed in the API behaviour or the value a consumer sees.
If a community member creates a Go/NodeJS/Rust version the response should be the same for consistency.

suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -8,6 +8,7 @@
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import org.hiero.block.api.protoc.BlockAccessServiceGrpc;
+import org.hiero.block.api.protoc.BlockNodeServiceGrpc;
```
Nana-EC, line 11: Let's avoid the protoc usage.
With the new PBJ version you should be able to use the Service Interface generated.
```suggestion
import org.hiero.block.api..BlockNodeServiceInterface;
```

Nana-EC (commented):
Good additions.
Responded to a few comments and 2nd the note to adopt non-protoc approach since this is new

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
```diff
@@ -0,0 +1,125 @@
...
+        assertEquals(1, fromPluginBytes.size());
+
+        final ServerStatusResponse response = ServerStatusResponse.PROTOBUF.parse(fromPluginBytes.getLast());
+
+        assertNotNull(response);
+        assertEquals(0, response.firstAvailableBlock());
+        assertEquals(blocks - 1, response.lastAvailableBlock());
+        assertFalse(response.onlyLatestState());
+
+        // TODO() Remove when block node version information is implemented.
```
ata-nas, line 107: This todo on line 107 (and some other lines here) looks like it could use a GH issue to track it.

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Tests that the server status response is valid when no blocks are available.
+     * Verifies the first and last available block numbers and other response properties.
+     *
+     * @throws ParseException if there is an error parsing the response
+     */
+    @Test
+    @DisplayName("Should return valid Server Status when no blocks available")
+    void shouldReturnValidServerStatus() throws ParseException {
+        //        sendBlocks(5);
```
ata-nas, line 69: Commented out code on line 69, maybe delete or use?

suites/src/main/java/org/hiero/block/suites/server/status/positive/PositiveServerStatusTests.java
```diff
@@ -0,0 +1,95 @@
...
+    void teardownEnvironment() {
+        simulatorAppsRef.forEach(simulator -> {
+            try {
+                simulator.stop();
+                while (simulator.isRunning()) {
+                    Thread.sleep(100);
+                }
+            } catch (InterruptedException e) {
+                // Do nothing, this is not mandatory, we try to shut down cleaner and graceful
+            }
```
ata-nas, line 40: I think it would still be a good idea to call:
```java
```
to truly ensure (at least from our perspective) proper cleanup in case when an interrupt may have happened.

georgi-l95: No need, I interupt them on the next row.

ata-nas: This is for the thread itself, resetting the interrupted flag, generally programs rely on the correct state of that flag to be able to shutdown gracefully, as an inability to reset it may indicate to other places that an interrupt may not have happened, but indeed it has.

ata-nas: P.S. I did see your cancel call on the futures.

georgi-l95: The main ideas, because I didn't restore the flas is because: 
1. JUnit is using the same worker thread for the next tests, so setting wrongly the flag here could result in flaky behavior for the next tests. 
2. Simulators are running in completly different threads, so we don't really care if they about them, we try to stop them gracefully but that's it. I didn't want to "dirty" the main thread.
However I saw that some versions ago JUnit added some clearing procedure of the main working thread, between tests, so It's okay if the set the flag to interrupted here. It will get cleared by the next test. 👍

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
georgi-l95: Thanks, removed. 👍

suites/src/main/java/org/hiero/block/suites/BaseSuite.java
georgi-l95: Didn't we settle down to use protoc in the suites and simulator ? Not 100% sure anymore.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -0,0 +1,145 @@
...
+import com.hedera.pbj.runtime.grpc.Pipelines;
+import com.hedera.pbj.runtime.grpc.ServiceInterface;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.swirlds.metrics.api.Counter;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Arrays;
+import java.util.List;
+import org.hiero.block.api.ServerStatusRequest;
+import org.hiero.block.api.ServerStatusResponse;
+import org.hiero.block.api.protoc.BlockNodeServiceGrpc;
```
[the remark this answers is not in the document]
georgi-l95: Done, thank you 👍

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
georgi-l95: I see your point, If this is the only place where the Block Node Version will be needed, I agree that this is the responsible plugin.

suites/src/main/java/org/hiero/block/suites/server/status/positive/PositiveServerStatusTests.java
ata-nas: Interesting, noted. Thanks for the clarification.

georgi-l95: Forgot to write that, I've added it :D Thanks for the suggestion

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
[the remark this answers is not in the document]
ata-nas: @georgi-l95 are there any changes pending to be made here?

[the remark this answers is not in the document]
georgi-l95: I think it's okay to leave it as -1 for now, but I'm okay either way.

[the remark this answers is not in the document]
ata-nas: @georgi-l95 maybe I am unable to see some change, but I see the same 0 present L:77-78?
```java
        assertEquals(0, response.firstAvailableBlock());
        assertEquals(0, response.lastAvailableBlock());
```
Or am I mistaken somehow? Or we are changing this in the future?
Sorry for my misunderstanding.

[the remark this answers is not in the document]
jsync-swirlds: To be clear, @Nana-EC, by using `-1` we _have_ the same behavior regardless of client.
We _document_ that max_value is what the API expects because that's how `-1` shows up to a client.
The need for Java clients to know how protocol buffers encode max value for Java is present regardless (it's just an unfortunate consequence of Java's lack of unsigned support).

suites/src/main/java/org/hiero/block/suites/BaseSuite.java
jsync-swirlds: We did settle on that (I think it is just easy to forget).
Suites and Simulator should act like "real" clients as much as possible, and almost all clients would be using SDKs derived using protoc.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
georgi-l95: Leaving as TODO for now, until it's settled.

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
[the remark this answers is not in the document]
ata-nas: @georgi-l95 interesting, I am not sure that is/should be the case. The intent is that if there are no available ranges, `-1` is to be returned (the `UNKNOWN_BLOCK_NUMBER` value). If that is not the case, then there might be a bug somewhere? Below a snip of the implementation of the `ConcurrentLongRangeSet` 
image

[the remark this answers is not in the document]
georgi-l95: You're right. Production uses this implementation, for the unit tests we are using another implementation, which returns 0 for min/max.
E2E test validates that -1 is returned.
image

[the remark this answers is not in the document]
ata-nas: I see, thank you for specifying.

ata-nas (approved):
🔥

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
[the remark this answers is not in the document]
jsync-swirlds: Sounds like the test fixture might need an update, but that's probably outside the scope of this PR.

jsync-swirlds (approved):

AlfredoG87 (approved):

PR #1146, opened 2025-05-14
fix: Fix BlockAccessService - getBlock - when using only `retrieve_latest=true`

AlfredoG87:
## Reviewer Notes

current request does not work:
```

grpcurl -plaintext -import-path  <insertDir> -proto block_access_service.proto -d '{"retrieve_latest":"true"}' 127.0.0.1:8080 org.hiero.block.api.BlockAccessService.getBlock

```

with this temporary fix, we make it work.

but it could also work when:

```
grpcurl -plaintext -import-path  <insertDir> -proto block_access_service.proto -d '{"block_number": "18446744073709551615", "retrieve_latest":"true"}' 127.0.0.1:8080 org.hiero.block.api.BlockAccessService.getBlock
```

### DIFF
--- a/block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
+++ b/block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
@@ -54,8 +54,9 @@ public class BlockAccessServicePlugin implements BlockNodePlugin, ServiceInterfa
         requestCounter.increment();
 
         try {
-            // Log in case both block_number and retrieve_latest are set, this should not happen
-            if (request.retrieveLatest() && request.blockNumber() != -1) {
+            // Log and return an error if both block_number and retrieve_latest are set.
+            // blockNumber is considered set if it is different from 0 (default)
+            if (request.retrieveLatest() && request.blockNumber() != 0) {
                 LOGGER.log(
                         INFO,
                         "Both block_number and retrieve_latest set. Using retrieve_latest instead of block_number: {0}",
@@ -63,15 +64,7 @@ public class BlockAccessServicePlugin implements BlockNodePlugin, ServiceInterfa
                 responseCounterNotFound.increment();
                 return new BlockResponse(Code.READ_BLOCK_NOT_FOUND, null);
             }
-            // when block_number is -1 and retrieve_latest is false, return an NOT_FOUND error
-            if (request.blockNumber() == -1 && !request.retrieveLatest()) {
-                LOGGER.log(INFO, "Block number is -1 and retrieve_latest is false");
-                responseCounterNotFound.increment();
-                return new BlockResponse(Code.READ_BLOCK_NOT_FOUND, null);
-            }
-
             long blockNumberToRetrieve;
-
             // if retrieveLatest is set, get the latest block number
             if (request.retrieveLatest()) {
                 blockNumberToRetrieve = blockProvider.availableBlocks().max();
--- a/block-node/block-access/src/test/java/org/hiero/block/node/access/service/BlockAccessServicePluginTest.java
+++ b/block-node/block-access/src/test/java/org/hiero/block/node/access/service/BlockAccessServicePluginTest.java
@@ -100,11 +100,8 @@ public class BlockAccessServicePluginTest extends GrpcPluginTestBase<BlockAccess
     @Test
     @DisplayName("Request Latest Block")
     void testRequestLatestBlock() throws ParseException {
-        final BlockRequest request = BlockRequest.newBuilder()
-                .blockNumber(-1)
-                .allowUnverified(true)
-                .retrieveLatest(true)
-                .build();
+        final BlockRequest request =
+                BlockRequest.newBuilder().retrieveLatest(true).build();
         toPluginPipe.onNext(BlockRequest.PROTOBUF.toBytes(request));
         // Check we get a response
         assertEquals(1, fromPluginBytes.size());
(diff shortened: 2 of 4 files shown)


Nana-EC (commented):
Looking good but one suggestion to include more coverage

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -55,23 +55,15 @@ private BlockResponse handleBlockRequest(BlockRequest request) {
 
         try {
             // Log in case both block_number and retrieve_latest are set, this should not happen
-            if (request.retrieveLatest() && request.blockNumber() != -1) {
+            if (request.retrieveLatest() && request.blockNumber() != 0) {
```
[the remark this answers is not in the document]
AlfredoG87: No, this condition is checking that if `retrieveLatest=true` then block_number=0, otherwise it fails.
Maybe it needs a more descriptive comment and condition.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -55,23 +55,15 @@ private BlockResponse handleBlockRequest(BlockRequest request) {
...
                 LOGGER.log(
                         INFO,
                         "Both block_number and retrieve_latest set. Using retrieve_latest instead of block_number: {0}",
                         request.blockNumber());
                 responseCounterNotFound.increment();
                 return new BlockResponse(Code.READ_BLOCK_NOT_FOUND, null);
```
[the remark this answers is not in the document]
AlfredoG87: Yes I agree, but at the time the team decided to limit the amount of error codes, so this was the only left available 😅 

This is a good discussion for tomorrow design meeeting.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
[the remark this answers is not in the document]
AlfredoG87: Improved wording on the comment to make it clearer. ✅

block-node/block-access/src/test/java/org/hiero/block/node/access/service/BlockAccessServicePluginTest.java
```diff
@@ -117,37 +114,34 @@ void testRequestLatestBlock() throws ParseException {
...
-                .retrieveLatest(false)
+                .retrieveLatest(true)
                 .build();
         toPluginPipe.onNext(BlockRequest.PROTOBUF.toBytes(request));
         // Check we get a response
         assertEquals(1, fromPluginBytes.size());
         // parse the response
         BlockResponse response = BlockResponse.PROTOBUF.parse(fromPluginBytes.get(0));
-        // check that the status is READ_BLOCK_NOT_FOUND
+        // check that the status is success
```
Nana-EC, line 144: nit:
```suggestion
        // check that the status is READ_BLOCK_NOT_FOUND
```

block-node/block-access/src/test/java/org/hiero/block/node/access/service/BlockAccessServicePluginTest.java
```diff
@@ -117,37 +114,34 @@ void testRequestLatestBlock() throws ParseException {
...
-        assertEquals(Code.READ_BLOCK_NOT_FOUND, response.status());
+        // check that the status is READ_BLOCK_NOT_FOUND
+        assertEquals(Code.READ_BLOCK_SUCCESS, response.status());
+        assertEquals(0, response.block().items().getFirst().blockHeader().number());
     }
 
     @Test
-    @DisplayName("block_number is -1 and retrieve_latest is false - should return READ_BLOCK_NOT_FOUND")
-    void testBlockNumberIsMinusOneAndRetrieveLatestIsFalse() throws ParseException {
+    @DisplayName("Request Latest and a specific Block different from latest, should fail with READ_BLOCK_NOT_FOUND")
```
Nana-EC, line 131: Potential for clarity but could be related in follow up
```suggestion
    @DisplayName("Request Latest and any Block, should fail with READ_BLOCK_NOT_FOUND")
```

Nana-EC (approved):
LG

Nana-EC (approved):
LG

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -54,24 +54,17 @@ private BlockResponse handleBlockRequest(BlockRequest request) {
         requestCounter.increment();
 
         try {
-            // Log in case both block_number and retrieve_latest are set, this should not happen
-            if (request.retrieveLatest() && request.blockNumber() != -1) {
+            // Log and return an error if both block_number and retrieve_latest are set.
+            // blockNumber is considered set if it is different from 0 (default)
+            if (request.retrieveLatest() && request.blockNumber() != 0) {
```
jsync-swirlds, line 59: Please ensure we have instructions in the bug issue to revert this when we properly fix the bug.
Block 0 is a valid block to request and must not act as a flag value.

jsync-swirlds:
This PR should not apply to main, as it creates a bug to work around a bad assumption.
Can we target this _only_ to the release branch as a temporary patch/workaround, and rely on the bug issue to actually fix this correctly in main?

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
jsync-swirlds, line 59: I think we actually need to modify the protobuf here (not in this PR, though, as a proper bug fix to main).
The current definition is too difficult to use correctly, and requires us to check things that PBJ makes unnecessarily difficult (field presence).

AlfredoG87: So your proposal is to modify the protobuf to use `int64` instead of `uint64` so we can accept `-1` ?

jsync-swirlds: No, my proposal is to change the protobuf to not allow both fields to be filled in at all.

jsync-swirlds: The branch [is pushed](https://github.com/hiero-ledger/hiero-block-node/tree/block-access-proto-fix) with the suggested change.

Nana-EC (approved):
Appropriate as work around for  v0.10.0 testing

AlfredoG87:
superseeded by https://github.com/hiero-ledger/hiero-block-node/pull/1147

and founded another workaround so no need to rush this change into current release.

PR #1147, opened 2025-05-14
fix: Adjust block access service to be correct without workarounds

### LINKED ISSUE #1148: Handling of "latest" and "last block number" in block access API is not entirely correct
### Description

While investigating bug #1144 we encountered some difficulty setting both retrieve latest and `-1` for block number.  Partly this was a misunderstanding of protobuf java encoding, but it also highlighted some overly complex and not entirely correct code in the plugin.
We need to make the presence for those two fields explicit, so we aren't checking for default values or flag values and trying to determine in code if one or both are set.

I propose to change the proto definition to a oneOf for the two fields, and also remove the allow_unverified flag which we do not intend to support, and which has no apparent value.

### Steps to Reproduce

This is a code-inspection discovered bug; the in-code complexities are in BlockAccessServicePlugin.java around lines 58-70.

jsync-swirlds:
## Reviewer Notes

* Modified the proto definition to remove unverified and prevent calling with both latest and a block number.
* Added an "invalid" response and removed the verified/unverified responses
* Removed unnecessary prefixes in the response enum names
* Removed check for both fields set in Java (it's no longer possible).
* Fixed up all references
* Removed tests that set both fields as it's no longer possible (the last one set is the only one set).

## Related Issue(s)
 Fixes #1148

### DIFF
--- a/protobuf/src/main/proto/org/hiero/block/api/block_access_service.proto
+++ b/protobuf/src/main/proto/org/hiero/block/api/block_access_service.proto
@@ -13,50 +13,29 @@ import "stream/block.proto";
  * A request to read a single block.
  *
  * A client system SHALL send this message to request a single block,
- * including the block state proof.<br/>
- * A client MAY request that the block be sent without verification.
- * A compliant Block Node MAY respond to requests that allow unverified
- * responses by returning the full requested block before verifying
- * the included block proof.<br/>
- * A compliant Block Node MAY support _only_ requests that allow unverified
- * blocks, but MUST clearly document that limitation, and MUST respond to
- * a request that does not allow unverified blocks with the
- * `ALLOW_UNVERIFIED_REQUIRED` response code.
+ * including the block state proof.
  */
 message BlockRequest {
-    /**
-     * The block number of a block to retrieve.
-     * <p>
-     * The requested block MUST exist on the block node.<br/>
-     * This value MUST NOT be set if `retrieve_latest` is set `true`.<br/>
-     * This value MUST be set to a valid block number if `retrieve_latest` is
-     * unset or is set `false`.
-     */
-    uint64 block_number = 1;
+    oneof block_specifier {
+        /**
+         * The block number of a block to retrieve.
+         * <p>
+         * The requested block MUST exist on the block node.<br/>
+         * A request MAY specify `uint64_max` to signal that the last possible
+         * block should be returned (which is subtly different from setting
+         * `retrieve_latest` instead of setting a `block_number`, though the
+         * result will always be the same, in most implementations).
+         */
+        uint64 block_number = 1;
 
-    /**
-     * A flag to indicate that the requested block may be sent without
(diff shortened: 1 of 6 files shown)


protobuf/src/main/proto/org/hiero/block/api/block_access_service.proto
```diff
@@ -13,50 +13,29 @@ import "stream/block.proto";
...
-     * SHALL be ignored.
-     */
-    bool retrieve_latest = 3;
+        /**
+         * A flag to request the latest available block.
+         * <p>
+         * This value SHOULD be set `true`, instead of setting `block_number`
+         * if the intent is to request the latest block available.
+         */
+        bool retrieve_latest = 3;
```
Nana-EC, line 37:
```suggestion
        bool retrieve_latest = 2;
```

protobuf/src/main/proto/org/hiero/block/api/block_access_service.proto
```diff
@@ -104,33 +83,23 @@ message BlockResponse {
...
+         * The request cannot be fulfilled.<br/>
+         * The client sent a malformed or structurally incorrect request.
          * <p>
-         * The client MAY retry the request after a short delay
-         * (typically 2 seconds or more).
+         * The client MAY retry the request after correcting the form and
+         * structure.
          */
-        VERIFIED_BLOCK_UNAVAILABLE = 6;
+        INVALID_REQUEST = 5;
```
Nana-EC, line 102: nit: I might switch the order of `INSUFFICIENT_BALANCE` and `INVALID_REQUEST`.
There's an argument to suggest we haven't designed charging flow properly yet so the `INSUFFICIENT_BALANCE` code should not be present till we need it as it could change or move.
I'm okay okay with it in as the API is still in flux but as we solidify let's revisit so that we're intentional with additions and proto is applicable around the time of inclusion

Nana-EC (commented):
1st pass looks looks good on simplifying the logic

protobuf/src/main/proto/org/hiero/block/api/block_access_service.proto
jsync-swirlds: Good thoughts.  I've reordered to something that might be a good pattern for all the APIs; with SUCCESS as `1` and INVALID_REQUEST as `2`.
INSUFFICIENT_BALANCE is removed with the assumption it would be part of a separate plugin, if anyone ever adds monetization/service charges.

AlfredoG87 (approved):
LGTM

fun fact:

you can actually request the latest block either way:

1.
```
{
    "retrieve_latest": true
}
```
2.
```
{
    "block_number": "18446744073709551615"
}
```

jsync-swirlds:
> LGTM
> 
> fun fact:
> 
> you can actually request the latest block either way:
> ```

Yes, I thought it might smooth the change if either works.  We can always make the small change so it only works with the flag set in a later update.

Issue #1052, opened 2025-04-21
`S3ArchivePlugin` Starts Uploads Sooner


jasperpotts:
### Story Form

I would like archive block backups to be started as soon as possible to minimize risk of loosing block data.

### Technical Notes

~The first implementation waits till all blocks are available before starting uploading a backup archive file. So for example if the backup size is set at 10,000 blocks, the archive plugin waits till the next 10,000 blocks are available and then starts uploading them into new "tar" file. This means that if a disaster happens while waiting for 10,000 blocks then up to 10,000 blocks could be lost forever.~

The archive plugin uses S3 multi-part uploads, there is no time limit to complete upload. So we could start uploading as soon as there is a 5MB chunk of block data to upload. We will need to carefully store metadata of the multi-part key and the last block written and offset in that block. That way if block node is restarted in the middle of upload it can restart where it left off. In a disaster recovery case you will be able to manually close the multi-part upload and download its contents. It will have recoverable contents or all blocks other than last 5MB of data. So worst case scenario you only loose 5Mb of block data.

Note 5MB is configurable.

PR #1149, opened 2025-05-15
feat: S3 Archive Plugin Cleanup and Error Handling Improvements

### LINKED ISSUE #1118: S3 Archive Plugin Cleanup and Error Handling Improvements
### Story Form

As a Block Node Developer
I want to have a cleaned up S3 client with good error handling
So that I can utilize the client easily and be able to extend it easily as needed

### Technical Notes

- the S3 Client and S3 Archive Plugin are in need of some general cleanups that will make these classes easier to read/extend/reuse
- the error handling for the client should be improved, it would be a good idea to have custom checked exception(s) that will be thrown by the public methods in order to improve error handling

ata-nas:
## Reviewer Notes

- some cleanups for the S3 client
- I have removed the xml handler as it was creating some issues (or rather, was not flexible enough)
  - we cannot rely on `Content-Lenght` header to determine if there is a body or not unfortunately. It is not consistent.
  - some requests are not expected to have a body if successful, but will have if an error occurs. We need more flexibility to be able to dynamically parse or not a body.
- currently tested with AWS S3, will test also with GCP
  - GCP _should_ be compatible (request wise), the hope is that the same client can be reused (maybe authorization header will be different)

## Related Issue(s)

Closes #1118


block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ClientException.java
ata-nas, line 1: I've created some checked exceptions that will help us with error handling. In case of failed requests, we will be able to know what happened (for the most part) and we will be able to make some decision, maybe a retry or similar.

I have been generally going back and forth for a couple of days, finding the best way to do that. Early feedback is appreciated. Exception abstractions could be further extended.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ResponseException.java
```diff
@@ -0,0 +1,80 @@
...
+        return sb.toString();
+    }
+
+    private String getRawContentOfTag(final Document responseBody, final String tagName) {
+        return responseBody
+                .getDocumentElement()
+                .getElementsByTagName(tagName)
+                .item(0)
+                .getTextContent();
+    }
```
ata-nas, line 79: This will cleanly print some xml nodes that should be in general always expected whenever an exception occurs (3xx+ response code from S3).

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
```diff
@@ -481,26 +633,32 @@ private static String computeSignatureForAuthorizationHeader(
...
+                .filter(entry -> !entry.toLowerCase(Locale.ENGLISH).equals("x-amz-content-sha256"))
                 .map(header -> header.toLowerCase(Locale.ENGLISH))
                 .sorted(String.CASE_INSENSITIVE_ORDER)
                 .collect(Collectors.joining(";"));
         // The canonical header requires value entries in sorted order, and multiple white spaces in the values should
         // be compressed to a single space.
         final String canonicalizedHeaders = headers.entrySet().stream()
+                        // todo is this added filter needed?
+                        .filter(entry ->
+                                !entry.getKey().toLowerCase(Locale.ENGLISH).equals("x-amz-content-sha256"))
```
ata-nas, line 661: These added filters are maybe not needed. Need to double check and confirm.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
```diff
@@ -322,83 +484,75 @@ void completeMultipartUpload(String key, String uploadId, List<String> eTags) {
...
-            HttpRequest request = requestBuilder.build();
-
+            final HttpRequest request = requestBuilder.build();
             return httpClient.send(request, bodyHandler);
-        } catch (IOException e) {
+        } catch (final IOException e) {
             throw new UncheckedIOException(e);
-        } catch (InterruptedException | URISyntaxException e) {
+        } catch (final InterruptedException | URISyntaxException e) {
+            // todo what would be the correct handling for the InterruptedException?
```
ata-nas, line 555: Ideas? What should we do here if we are interrupted? Close the client? That also brings me to the question, should we throw if a client is closes, but we make calls after that? Or we rely on the http client to do that?

ata-nas (commented):
Just some general questions. I have generally added some cleanup, a couple of new methods. I have tested with AWS S3, it is all working properly. GCP to follow. Collecting early feedback. Going back and forth on how to utilize the checked exceptions to do some error handling. Opted to remove the xml handler as it was causing issues: we cannot rely on the "Content-Length" header to determine if there is a body. Also the handler is not flexible enough for our needs, there are some requests that have no body if successful, but will have an xml body if the request fails (3xx+ response), and the handler will throw if no body or no xml body.

**PLEASE NOTE:** I have not yet gotten to testing the plugin itself. I have tested the client only. I will test the plugin itself as well, but will have to update with results. Also, more tests are to be added, have not gotten to that as well as of now.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ResponseException.java
jsync-swirlds: Are you saying a `3xx` response would result in an error?  In most cases `3xx` responses should be followed through, not treated as errors.  We should also be handling the "incomplete" `1xx` and `2xx` codes.
> Examples:
  `100` by continuing to send
  `201`-`204` as completed
  `205` by retrying
  `206` by requesting more content

The HTTP "error" code ranges are `4xx` for a client error (we sent something wrong) and `5xx` for a server error (the server could not respond correctly).

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
jsync-swirlds: Language, in particular, might not be something we should filter...

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
jsync-swirlds: If a client is closed, that means our plugin closed it, and we should either open a new client or quit.
If a thread is interrupted, that generally means the system either is shutting down or is requesting this plugin to stop running.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
```diff
@@ -242,77 +313,168 @@ public boolean uploadFile(
...
+        // build the request URL
+        final String url = endpoint + bucketName + "/" + key + "?" + canonicalQueryString;
+        final HttpResponse<InputStream> response =
+                request(url, DELETE, Collections.emptyMap(), null, BodyHandlers.ofInputStream());
+        final int statusCode = response.statusCode();
+        if (statusCode != 204) {
+            final Document body = parseDocument(response.body());
+            final String message = "Failed to abort multipart upload: key=%s, uploadId=%s".formatted(key, uploadId);
+            throw new S3ResponseException(statusCode, body, message);
+        }
```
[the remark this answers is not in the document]
jsync-swirlds: I do not think we can rely on local data in this case.
We should close any in-progress multi-part uploads, then read back the archive and determine from that what blocks were stored.
From there we can back up, re-create the archive, and re-do the upload, after which we can remove the partial file.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ResponseException.java
ata-nas: Generally no, I do agree 4xx and 5xx are error codes as you describe, but I do follow the amazon's error handling practices and the api they give us: https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html

ata-nas: And by their docs 5xx should be retried.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
ata-nas: I have no recollection as to why this filter was added, this builds the auth header. Need to double check.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ResponseException.java
ata-nas: P.S. each request we build handles it's own response to the level that we either do nothing, return the result or throw a checked exception. From then on, it's the caller.

jsync-swirlds: I am mostly just noting that anything other than `5xx` or `4xx` should not be an error; and should probably be handled in this client code (though it can be a separate issue as some of these are complex).

I generally prefer to see a client library like this handle things like `3xx` codes because redirections, in particular, are easy to get wrong (and almost nobody will correctly handle things like `100` correctly in a caller).

It's common for APIs to suggest retry of `5xx` codes (in case it was a transient server failure), but that usually needs a limit because a server _can_ continue to fail persistently for a rather long time.

georgi-l95 (commented):
Did one round, will do another when tests are added. Solid progress 🙌

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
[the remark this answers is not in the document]
ata-nas: In general yes, I was just thinking out loud. How will we determine which multipart upload of all listed we should close? Simply we need to know which one(s) are active currently. What would be the best way to do that?

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
ata-nas: Hmm. In the case of an interrupt I would ask should we drop the work of the client forcefully and abort whatever we were doing (if can abort) or should we simply wait. I guess that depends on the plugin itself and the work it does and how critical it is.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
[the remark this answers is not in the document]
jsync-swirlds: Having reasonable guarantees that what is in the deep archive is clean and correct is vastly more important for this plugin than saving a bit of work and upload bandwidth.

I would advocate that we strongly bias all decisions for this plugin on "the archive is complete and correct" over "save cost or effort".

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
jsync-swirlds: It's very important to _not wait_ when handling an interrupted exception (or thread interrupt flag).
Waiting for things to finish (if it isn't _very_ quick) in interrupt, or ignoring the interrupt, is how we get so many Java systems that cannot be cleanly shut down with an interrupt signal (SIG_INT/Ctrl-C).

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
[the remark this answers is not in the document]
jsync-swirlds: BTW, there is no point detecting which uploads are "active", just close them all.  No two block nodes running this plugin should ever be using the same account simultaneously.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ResponseException.java
ata-nas: Recently, I really dislike my approach here. I am thinking of simply printing the body and not make any assumptions about it's structure. @jsync-swirlds (about the toString method)

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ResponseException.java
```diff
@@ -0,0 +1,66 @@
...
+            sb.append(": ").append(message);
+        }
+        sb.append(System.lineSeparator());
+        // append the response status code
+        sb.append("    Response status code: ").append(statusCode);
+        if (responseBody != null && responseBody.length > 0) {
+            // if there is a response body, append it
+            sb.append(System.lineSeparator());
+            sb.append("    Response body: ").append(new String(responseBody));
+        }
```
jsync-swirlds, line 63: This might be easier to follow as a string format:
```java
final String bodyString = responseBody != null && !responseBody.isEmpty() ? "%n    Response body: %s".formatted(responseBody) : "";
return "%s%s%s%n    Response status code: %s%s".formatted(className, message != null ? "" : ": ", message, statusCode, responseBody);
```

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -212,7 +214,14 @@ private void scheduleBatchArchiving(final long nextBatchStartBlockNumber, final
                                     System.Logger.Level.INFO,
                                     "Uploaded archive block batch: " + nextBatchStartBlockNumber + "-"
                                             + nextBatchEndBlockNumber);
-                        } catch (Exception e) {
+                        } catch (final S3ResponseException e) {
+                            // todo we could retry here
+                            LOGGER.log(
+                                    System.Logger.Level.WARNING,
```
jsync-swirlds, line 220: This isn't a warning.  This is informational at most.
It might be a warning if we had repeated failures.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ResponseException.java
ata-nas: I know, but I do like addressing people directly, just a habit. If it is too much of a spam I will tone it down 😄

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -147,25 +150,21 @@ public void stop() {
...
+        // compute the end of the next batch to archive
+        long nextEnd = nextStart + archiveConfig.blocksPerFile() - 1;
+        // while the next batch end is less than or equal to the latest
+        // persisted block number, we can schedule a batch
+        while (nextEnd <= latestPersisted) {
+            // schedule the batch archiving
+            scheduleBatchArchiving(nextStart, nextEnd);
+            // move to the next batch and try again
+            nextStart = nextEnd + 1;
+            nextEnd = nextStart + archiveConfig.blocksPerFile() - 1;
```
ata-nas, line 167: I've simplified this. The previous approach was not calculating successfully and skipped to every other batch (due to the async nature of updating the atomic long as well).

This is not ideal however. Moreover, the logic that will get the accessors for the batch itself will allow gaps.

We do plan to change the approach here. We want to be uploading smaller batches with a few blocks at a time, appending them to the current tar or generating a new one if need be. In that way we will not need this calculation. See #1052

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
ata-nas: I've reverted this. I tested without this filter and everything worked. My assumption is that I added these when testing and was questioning if we need this. I do not believe so.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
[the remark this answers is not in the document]
ata-nas: As we have discussed with the team, we will use these methods during startup in order to do cleanups. We will not utilize local key-value storage to write metadata that will follow current uploads.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ResponseException.java
ata-nas: As we discussed in another place, for now, let's limit the printable size of the headers and body to 2KB each and let's only print for now. When running, we will detect what additional handling we will need. I've added the size limit in the toString method.

AlfredoG87: > It's common for APIs to suggest retry of 5xx codes (in case it was a transient server failure), but that usually needs a limit because a server can continue to fail persistently for a rather long time.

That is why transient retries need to have exponential back off and or a number limit before looking for alternatives or crash.

AlfredoG87 (dismissed):
LGTM 👍

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
ata-nas: +1, I think the same. Let's see in a follow up how that would look like potentially.

block-node/base/src/main/java/org/hiero/block/node/base/s3/S3ResponseException.java
ata-nas: Yes, any retries must have a limit for sure.

AlfredoG87 (approved):
LG!

PR #1155, opened 2025-05-16
fix: Moved GRPC services over to use PBJ generated ones

### LINKED ISSUE #1088: Replace protoc gRPC implementations with PBJ versions
### Story Form

As a Block Node Engineer
I want to use PBJ generated classes and avoid protoc generated classes
So that I have no dependencies on protoc and can remove that step from the build

### Technical Notes

An earlier issue removed the protoc gRPC stubs and switched to PBJ stubs, but the refactor somehow switched everything back to protoc stubs.  We would very much like to remove the protoc compile step from the build entirely so we don't have the dependency tree in the deployed system and also avoid accidentally importing the wrong version of protobuf-defined classes.

jasperpotts:
## Reviewer Notes
Stops exporting protoc to anything other than simulator and suites

## Related Issue(s)
Fixes #1088

### DIFF
--- a/block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
+++ b/block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
@@ -6,20 +6,12 @@ import static java.lang.System.Logger.Level.ERROR;
 import static java.lang.System.Logger.Level.INFO;
 
 import com.hedera.hapi.block.stream.Block;
-import com.hedera.pbj.runtime.grpc.GrpcException;
-import com.hedera.pbj.runtime.grpc.Pipeline;
-import com.hedera.pbj.runtime.grpc.Pipelines;
-import com.hedera.pbj.runtime.grpc.ServiceInterface;
-import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.metrics.api.Counter;
-import edu.umd.cs.findbugs.annotations.NonNull;
-import java.util.Arrays;
-import java.util.List;
+import org.hiero.block.api.BlockAccessServiceInterface;
 import org.hiero.block.api.BlockRequest;
 import org.hiero.block.api.BlockResponse;
 import org.hiero.block.api.BlockResponse.Code;
 // PBJ doesn't generate GRPC stubs for some reason, also the proto file is broken when PBJ compiles it...
-import org.hiero.block.api.protoc.BlockAccessServiceGrpc;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
 import org.hiero.block.node.spi.ServiceBuilder;
@@ -28,7 +20,7 @@ import org.hiero.block.node.spi.historicalblocks.HistoricalBlockFacility;
 /**
  * Plugin that implements the BlockAccessService and provides the 'block' RPC.
  */
-public class BlockAccessServicePlugin implements BlockNodePlugin, ServiceInterface {
+public class BlockAccessServicePlugin implements BlockNodePlugin, BlockAccessServiceInterface {
 
     /** The logger for this class. */
     private final System.Logger LOGGER = System.getLogger(getClass().getName());
@@ -43,13 +35,15 @@ public class BlockAccessServicePlugin implements BlockNodePlugin, ServiceInterfa
     /** Counter for the number of responses not found */
     private Counter responseCounterNotFound;
 
+    // ==== BlockAccessServiceInterface Methods ========================================================================
+
     /**
      * Handle a request for a single block
      *
      * @param request the request containing the block number or latest flag
      * @return the response containing the block or an error status
      */
-    private BlockResponse handleBlockRequest(BlockRequest request) {
+    public BlockResponse getBlock(BlockRequest request) {
         LOGGER.log(DEBUG, "Received BlockRequest for block number: {0}", request.blockNumber());
         requestCounter.increment();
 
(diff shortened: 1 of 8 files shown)


block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -6,20 +6,12 @@
...
-import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.swirlds.metrics.api.Counter;
-import edu.umd.cs.findbugs.annotations.NonNull;
-import java.util.Arrays;
-import java.util.List;
+import org.hiero.block.api.BlockAccessServiceInterface;
 import org.hiero.block.api.BlockRequest;
 import org.hiero.block.api.BlockResponse;
 import org.hiero.block.api.BlockResponse.Code;
 // PBJ doesn't generate GRPC stubs for some reason, also the proto file is broken when PBJ compiles it...
```
jsync-swirlds, line 14: Might want to remove this comment as well.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -171,21 +128,17 @@ static class SubscribeBlockStreamHandler
...
         private final ExecutorCompletionService<BlockStreamSubscriberSession> streamSessions;
 
-        private SubscribeBlockStreamHandler(
-                @NonNull final BlockNodeContext context, @NonNull final SubscriberServicePlugin plugin) {
+        private SubscribeBlockStreamHandler(@NonNull final BlockNodeContext context) {
             this.context = requireNonNull(context);
-            this.plugin = requireNonNull(plugin);
             this.openSessions = new ConcurrentSkipListMap<>();
-            virtualThreadExecutor = context.threadPoolManager().getVirtualThreadExecutor(null);
+            virtualThreadExecutor = context.threadPoolManager().getVirtualThreadExecutor("SubscribeBlockStreamHandler");
```
jsync-swirlds, line 141: This was null on purpose, so we don't create unique executors for every single request.

jsync-swirlds (dismissed):
A couple of non-blocking items.

Thanks for tackling this.

AlfredoG87 (approved):
LGTM, other than the comments leaved by @jsync-swirlds 
Also spotless is compalining.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
Nana-EC: Maybe we should make it more explicit. 
Setting a name hiding implications of executor service and handler relationships is bound to be missed.
Not for this PR but should that interface be made more explicit with an enum that forces clear specification of intention by a dev when calling this?

Nana-EC (dismissed):
LG, gets us closer to no more protoc

georgi-l95 (approved):

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
jsync-swirlds: Possibly.  It's very tricky to define an API that's both clean for routine use and clear for every option.
The impacts of thread name, in particular, because it's a debugging detail, are specific to the implementation (name isn't guaranteed to be used at all), so it's a very difficult item to define and document properly.
The thread name item is particularly pernicious as we've made some somewhat flawed assumptions all over the codebase after switching to this thread manager model, and the name parameter ties into most of those.
Certainly we can document these things better; though we need to be careful that we don't document _implementation_ details in the _API_ documents.

jasperpotts: I changed it as it was flagged as error in IntelliJ as the method parameter is marked as `@NotNull` so it is flagged when you pass null. Maybe you can do another PR and either remove the @NotNull and document the null behavior or remove the special null handling and add a extra method for that specific use case. I like the latter better I think.

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
jasperpotts: done

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
jsync-swirlds: I'll send a PR to fix the annotation, and perhaps also to try to explain this implementation detail in the default implementation class.

jsync-swirlds:
Complete side note, but github is not verifying the signature on your commits; so there may be a setup issue somewhere.
image

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -99,11 +92,18 @@ private BlockResponse handleBlockRequest(BlockRequest request) {
     }
 
     // ==== BlockNodePlugin Methods ====================================================================================
+
+    /**
+     * {@inheritDoc}
+     */
```
jsync-swirlds, line 98: Just curious, why do we keep adding these?
The javadoc will be duplicated properly when there is no comment at all; so this is just extra clutter.

jsync-swirlds (approved):

AlfredoG87 (approved):

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
jasperpotts: if you have "Render All Doc Comments" on in IntelliJ then you get comments in for methods if you include inherit. 

Screenshot 2025-05-20 at 3 07 16 PM

PR #1160, opened 2025-05-16
chore: Update MAINTAINERS.md names and affiliations

jwagantall:


Nana-EC (changes requested):
Not the case, current table contents are correct.
PR can be closed.
Thanks @jwagantall

PR #1162, opened 2025-05-17
chore: Small Helm and project improvements

AlfredoG87:
## Reviewer Notes

1. improving consumer override value, to avoid permission denied issue when deploying.

2. adding mid-size.yaml for medium size deployments, usually what we deploy on dev k8 env.

3. extra housekeeping of the settings.gradle.kts file, so it matches 1-1 with the modules, keeps consistent aliases, and also are shown in the same order (alphabetical) so is easier to compare the 2 in the future

## Related Issue(s)

### DIFF
--- a/charts/blockstream-simulator/values-overrides/consumer.yaml
+++ b/charts/blockstream-simulator/values-overrides/consumer.yaml
@@ -1,7 +1,8 @@
 # 2 consumers
-replicas: 2
+replicas: 1
 
 simulator:
   config:
     GRPC_SERVER_ADDRESS: "my-bn-block-node-helm-chart"
     BLOCK_STREAM_SIMULATOR_MODE: "CONSUMER"
+    SIMULATOR_STARTUP_DATA_ENABLED: "false"
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -15,17 +15,17 @@ javaModules {
     directory("block-node") {
         group = "org.hiero.block"
         module("app") { artifact = "block-node-app" }
-        module("archive") { artifact = "block-node-archive" }
         module("base") { artifact = "block-node-base" }
-        module("plugins") { artifact = "block-node-plugins" }
-        module("health") { artifact = "block-node-health" }
-        module("messaging") { artifact = "facility-messaging" }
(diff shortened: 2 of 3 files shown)


charts/blockstream-simulator/values-overrides/consumer.yaml
```diff
@@ -1,7 +1,8 @@
 # 2 consumers
-replicas: 2
+replicas: 1
```
Nana-EC, line 2: Q: Should this be changed?
Comment above still says 2 and I imagine have 2 consumers is valuable.

Nana-EC (commented):
Looking good, 1 clarifcation

charts/blockstream-simulator/values-overrides/consumer.yaml
AlfredoG87: good catch, yes, I was doing some tests with 1 consumer, but 2 consumers by default is better.

AlfredoG87: ✅  Done

Nana-EC (approved):

settings.gradle.kts
```diff
@@ -15,17 +15,17 @@ javaModules {
     directory("block-node") {
         group = "org.hiero.block"
         module("app") { artifact = "block-node-app" }
-        module("archive") { artifact = "block-node-archive" }
         module("base") { artifact = "block-node-base" }
-        module("plugins") { artifact = "block-node-plugins" }
+        module("block-access") { artifact = "block-access-service" }
+        module("block-providers/cloud.historic") { artifact = "block-node-blocks-cloud-historic" }
```
jsync-swirlds, line 20: This plugin is just a stub, and also has no known requirement to be built.
I would vote that we just remove it entirely; it's wasting effort to maintain.

jsync-swirlds (approved):
Just one note for an item we might want to remove (in a separate issue).

settings.gradle.kts
AlfredoG87: I would agree, similar comment on `plugin metrics` will create another PR to clean-up this.

PR #1174, opened 2025-05-20
chore: Clean up ThreadPoolManager API

jsync-swirlds:
## Reviewer Notes
* Added default implementations for reduced-call overloads.
* Adjusted both default and test implementations to clean up overloads
* Adjusted default implementation to handle null thread names
* Cleaned up JavaDoc and explained potential caching impacts.

### DIFF
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/TestThreadPoolManager.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/TestThreadPoolManager.java
@@ -23,20 +23,9 @@ public class TestThreadPoolManager<T extends ExecutorService> implements ThreadP
         this.executor = Objects.requireNonNull(executor);
     }
 
-    @NonNull
-    @Override
-    public T getVirtualThreadExecutor(@NonNull final String threadName) {
-        return executor;
-    }
-
-    @NonNull
-    @Override
-    public T getVirtualThreadExecutor(
-            @NonNull final String threadName, @Nullable final UncaughtExceptionHandler uncaughtExceptionHandler) {
-        return executor;
-    }
-
     /**
+     * {@inheritDoc}
+     * <p>
      * Test implementation, always returns the same executor service that was
      * passed to the constructor of this class.
      *
@@ -44,11 +33,14 @@ public class TestThreadPoolManager<T extends ExecutorService> implements ThreadP
      */
     @NonNull
     @Override
-    public T createSingleThreadExecutor(@NonNull final String threadName) {
+    public T getVirtualThreadExecutor(
+            @Nullable final String threadName, @Nullable final UncaughtExceptionHandler uncaughtExceptionHandler) {
         return executor;
     }
 
     /**
+     * {@inheritDoc}
+     * <p>
      * Test implementation, always returns the same executor service that was
      * passed to the constructor of this class.
      *
@@ -57,7 +49,7 @@ public class TestThreadPoolManager<T extends ExecutorService> implements ThreadP
     @NonNull
     @Override
     public T createSingleThreadExecutor(
-            @NonNull final String threadName,
+            @Nullable final String threadName,
             @Nullable final Thread.UncaughtExceptionHandler uncaughtExceptionHandler) {
         return executor;
     }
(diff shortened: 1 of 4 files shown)


block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
```diff
@@ -11,49 +11,103 @@
...
+     * Factory method for a single thread executor.
+     * Creates a new, unnamed, single thread {@link ExecutorService}.
      *
-     * @param threadName the thread's name, must not be blank
      * @return a new single thread executor service
      */
     @NonNull
-    ExecutorService createSingleThreadExecutor(@NonNull final String threadName);
+    default ExecutorService createSingleThreadExecutor() {
+        return createSingleThreadExecutor(null, null);
```
ata-nas, line 72: I like all the `default` additions on interface level here. 👍

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
```diff
@@ -11,49 +11,103 @@
...
+     * caching or reusing executor services, and may result in more overhead.
+     *
      * @param threadName the thread name prefix, must not be blank
      * @return a new virtual thread-per-task executor service
      */
     @NonNull
-    ExecutorService getVirtualThreadExecutor(@NonNull final String threadName);
+    default ExecutorService getVirtualThreadExecutor(@NonNull final String threadName) {
+        return getVirtualThreadExecutor(threadName, null);
+    }
```
ata-nas, line 40: Let's enforce non-null (not blank) here. Implementations will most likely not override this default.

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
```diff
@@ -11,49 +11,103 @@
...
+     * Note, the thread name is a debugging aid. Naming threads may prevent
+     * caching or reusing executor services, and may result in more overhead.
      *
      * @param threadName the thread's name, must not be blank
+     * @return a new single thread executor service
+     */
+    @NonNull
+    default ExecutorService createSingleThreadExecutor(@NonNull final String threadName) {
+        return createSingleThreadExecutor(threadName, null);
+    }
```
ata-nas, line 88: Let's enforce non-null (not blank) here. Implementations will most likely not override this default.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/TestThreadPoolManager.java
```diff
@@ -23,32 +23,24 @@ public TestThreadPoolManager(@NonNull T executor) {
...
-    @NonNull
-    @Override
-    public T getVirtualThreadExecutor(
-            @NonNull final String threadName, @Nullable final UncaughtExceptionHandler uncaughtExceptionHandler) {
-        return executor;
-    }
-
     /**
+     * {@inheritDoc}
+     * <p>
```
ata-nas, line 28: Why do we inherit doc here? This test implementation differs completely from what the interface defines.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/TestThreadPoolManager.java
```diff
@@ -23,32 +23,24 @@ public TestThreadPoolManager(@NonNull T executor) {
...
     @Override
-    public T createSingleThreadExecutor(@NonNull final String threadName) {
+    public T getVirtualThreadExecutor(
+            @Nullable final String threadName, @Nullable final UncaughtExceptionHandler uncaughtExceptionHandler) {
         return executor;
     }
 
     /**
+     * {@inheritDoc}
+     * <p>
```
ata-nas, line 43: Same as above why do we inherit doc here? This test implementation differs completely from what the interface defines.

ata-nas (commented):
Looks good. A few questions, cleanup suggestions.

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
jsync-swirlds: I believe non-blank should be implementation (as it doesn't interact with the static analysis tools directly); but asserting non-null here makes sense.

Added.

block-node/spi/src/main/java/org/hiero/block/node/spi/threading/ThreadPoolManager.java
jsync-swirlds: Added non-null.

ata-nas (approved):
🔥

Nana-EC (approved):

PR #1175, opened 2025-05-21
docs: Update the publisher protocol document to include the multi-publisher extension

### LINKED ISSUE #1141: Add multi-publisher protocol changes to publisher protocol documentation
### Story Form

As a Block Stream Publisher
I want to have clear and complete publisher protocol documentation
So that I can correctly implement the required protocol

### Technical Notes

We defined a single publisher protocol for block nodes, and separately defined extensions for multiple publishers.  What we have not done (or lost in the big refactors and codebase reorganizations) is merge the multiple publisher extensions into the publisher protocol documentation.
We intend to correct that oversight with this issue.

jsync-swirlds:
## Reviewer Notes
* Added multiple publisher extension sections to the publisher protocol document
* Added SVG diagram for context
* Updated core protocol with adjustements for multiple publisher extension points
* Adjusted mermaid diagrams as needed
* Added a description for time-out, example, and limitation(s)

## Related Issue(s)
 Resolves #1141


jsync-swirlds:
Note on spotless:
Spotless insists on changes to the structure of the document (in ways that break the result layout); it simply cannot handle some of the markdown structures used here.

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -72,113 +77,248 @@ stream and retry (either to another Block Node, or after a short delay).
...
+    * > Note, a Block-Node can (re)enter "catch up" _any_ time that Block-Node
+        gets the next block from Publisher with a block number that is not what
+        the Block Node expects. This simplifies logic for working out when to
+        retry or reset a stream.
+  * A Publisher _may_ choose to implement routine reset and rebalance.
+    * Publisher, strictly after sending a `BlockProof` and before sending the
+      next `BlockHeader` shall send a `EndStream` with code `RESET`.
+    * Block-Node shall close the connection. If the Publisher failed to follow
+      protocol and sent this reset in the middle of a block, the partially
+      complete block _must_ be discarded and the data received from other
```
Nana-EC, line 127: > and the data received from other Publishers or another Block-Node.

I'm unclear what this part implies. It makes sense to get rid of the partial block from the producer but what data is to be discarded from other publishers and why?

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -72,113 +77,248 @@ stream and retry (either to another Block Node, or after a short delay).
...
+  * If this publisher is _not_ the first to send that block header, then
+    the Block-Node will reply with "SkipBlock".
+    * The Publisher will not send the remainder of that block, but will resume
+      sending data with the next Block Header.
+      * This Publisher _may_ be able to resume _before_ the `BlockProof` for
+        the current block is available, and _should_ do so.
+    * The Publisher _must_ still expect acknowledgement for the "skipped" block
+      and _must not_ remove that block from cache until it is acknowledged.
+    * A Publisher _may_ receive a "ResendBlock" response from the Block-Node
+      if the Publisher must resend the "skipped" block. The Published should
```
Nana-EC, line 225:
```suggestion
      if the Publisher must resend the "skipped" block. The Publisher should
```

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -72,113 +77,248 @@ stream and retry (either to another Block Node, or after a short delay).
...
+      then resume sending data from the "skipped" block Header for the block
+      number specified in `ResendBlock`.
+      * Note, it is entirely possible to receive another "SkipBlock" after
+        receiving a "ResendBlock" response, and should be handled as above.
+        This is possible because a Block-Node will send the "ResendBlock" to
+        all Publishers at the same time, and another Publisher may still
+        "win the race" to send the full block.
+  * If, while receiving a block from one node, another node _completes_ a later
```
Nana-EC, line 233: Can you clarify what complete implies here?
Do you mean a BN completes the caching of a future Block from another Node?

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -72,113 +77,248 @@ stream and retry (either to another Block Node, or after a short delay).
...
+    block, the Block-Node _may_ choose to send `EndStream` with status `TIMEOUT`
```
Nana-EC, line 234: Why timeout? I'm unclear of this scenario and why. the change of publishers with a timeout is needed.

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -72,113 +77,248 @@ stream and retry (either to another Block Node, or after a short delay).
...
+    > * Alice "wins" the race and starts sending block 9
+    >   * Bob receives a "SkipBlock"
+    >   * Carol receives a "SkipBlock"
+    > * Bob "wins" the race and starts sending block 10
+    >   * Carol receives a "SkipBlock"
+    >   * Alice is still sending block 9
+    > * Carol "wins" and starts sending block 11 (this is unusual, and may already indicate a concern)
+    > * Bob sends the `BlockProof` for block 10
+    >   * Block tries to send block 11 and receives a `SkipBlock`
+    >     * Here we _might_ want to send `EndOfStream` to Alice, and `ResendBlock(9)` to Bob and Carol.
```
Nana-EC, line 248: Why request for a resend of Block 9 if a BN is already processing it?

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -72,113 +77,248 @@ stream and retry (either to another Block Node, or after a short delay).
...
+        receiving a "ResendBlock" response, and should be handled as above.
+        This is possible because a Block-Node will send the "ResendBlock" to
+        all Publishers at the same time, and another Publisher may still
+        "win the race" to send the full block.
+  * If, while receiving a block from one node, another node _completes_ a later
+    block, the Block-Node _may_ choose to send `EndStream` with status `TIMEOUT`
+    to the current streaming Publisher for the earlier block, and _must_ then
+    send `ResendBlock` to all other publishers, specifying the block number that
+    _was_ being sent by the first Publisher.
+  * Example
```
Nana-EC, line 238: I like this example as an illustration. I think we need more of them to highlight the appropriateness of many of the codes

Nana-EC (commented):
Ist pass.
Thanks for elaborating and illustrating.
Had some questions.
Also I think we need to work on simplification where possible or in follow up.

docs/design/communication-protocol/publish-block-stream.md
jsync-swirlds: Because the source of that block is too slow to be acceptable _in this specific situation_.

docs/design/communication-protocol/publish-block-stream.md
Nana-EC: > the partially  complete block _must_ be discarded and the data received from other Publishers or another Block-Node.

Ah I see the issue.
The above read as "the data received from other Publishers or another Block-Node" should also be deleted, but you're saying after the discard take the data that you "received from other Publishers or another Block-Node"

docs/design/communication-protocol/publish-block-stream.md
Nana-EC: I see. 👍🏾

Nana-EC: 💯 on the needs a diagram.
If we can find a simple structure to illustrate many of these scenarios then we can add a simple diagram to each to confirm

jsync-swirlds: I hope to add a timing diagram; perhaps later today or tomorrow.

docs/design/communication-protocol/publish-block-stream.md
Nana-EC: Inference makes sense.
Although I'm wondering about  naming of `TIMEOUT`.
Would something more indicative of the BN assessing that the producer is slow and doesn't like the delays be better for clarity?
On the same API the Request and Response have a different definition for `TIMEOUT`.

docs/design/communication-protocol/publish-block-stream.md
Nana-EC: I see, we'll need logic to determine what too slow is. There may be cases where continuing to stream might be faster that starting a whole new one.
However, if could be argued it'd hard to determine and may not be often so in general it's better to switch.
🤔

docs/design/communication-protocol/publish-block-stream.md
jsync-swirlds: On the same API request and response have different definitions for everything; which is why the code is a different enum.
Assuming request and response codes are the same thing is a flawed assumption (kind of like assuming that the "gas" fueling an engine is the same as the "gas" exhausted because the names are the same).

docs/design/communication-protocol/publish-block-stream.md
jsync-swirlds: That's what the earlier comment is about.  We do have certain points where we _know_, without doubt, that "Alice" is slow (because "Bob" already sent us a later BlockProof, for example).

AlfredoG87 (approved):
Leaved a couple nits, otherwise seems spot on.

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -35,26 +35,26 @@ is broken; with disastrous consequences. Consensus nodes are one of the
 core clients of this protocol as Publisher, so a core assumption of this
 protocol is that if anything occurs to suggest to a publisher that a block
 node might fail to store blocks permanently, the publisher should end the
-stream and retry (either to another Block Node, or after a short delay).
+stream and retry (either to another Block-Node, or after a short delay).
```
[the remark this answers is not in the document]
jsync-swirlds: How does this imply holding 0 blocks?  Publishers are expressly required to hold blocks _at least_ until acknowledged (but may hold them arbitrarily longer).

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -35,26 +35,26 @@ is broken; with disastrous consequences. Consensus nodes are one of the
...
 
 <dt>Block Number</dt>
 <dd>A monotonically increasing number assigned by consensus to each block produced
   by the network.</dd>
 
 <dt>Publisher</dt>
-<dd>An entity publishing blocks to a Block Node via the `publishBlockStream` API
-  This is typically a Consensus Node or another Block Node.</dd>
+<dd>An entity publishing blocks to a Block-Node via the `publishBlockStream` API
+  This is typically a Consensus Node or another Block-Node.</dd>
```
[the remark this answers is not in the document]
jsync-swirlds: I don't recall any _restriction_ that a block node cannot send blocks to another block node via the `publishBlockStream` API, only that _Hedera_ likely won't deploy any in that mode.

The Block Node can be used by anyone, including a company that might find it more efficient to use one block node as a pass-through fan-out distributor to several special-purpose block nodes...

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -72,113 +77,248 @@ stream and retry (either to another Block Node, or after a short delay).
...
+  * Example
+    > * Alice "wins" the race and starts sending block 9
+    >   * Bob receives a "SkipBlock"
+    >   * Carol receives a "SkipBlock"
+    > * Bob "wins" the race and starts sending block 10
+    >   * Carol receives a "SkipBlock"
+    >   * Alice is still sending block 9
+    > * Carol "wins" and starts sending block 11 (this is unusual, and may already indicate a concern)
+    > * Bob sends the `BlockProof` for block 10
+    >   * Block tries to send block 11 and receives a `SkipBlock`
```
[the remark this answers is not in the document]
jsync-swirlds: Thanks.  Fixed.

docs/design/communication-protocol/publish-block-stream.md
jsync-swirlds: Thanks. Fixed.

docs/design/communication-protocol/publish-block-stream.md
jsync-swirlds: I believe this is worded better now.

docs/design/communication-protocol/publish-block-stream.md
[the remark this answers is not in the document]
AlfredoG87: I mean in the case of a BN Disaster case, but I think I was thinking in that case when originally read the document. I don't think it is implying it anymore.

georgi-l95 (approved):

Nana-EC (approved):
LG, we can address any minor details in follow up but I believe this has a good basis to explain and walk through the flows

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -72,113 +80,297 @@ stream and retry (either to another Block Node, or after a short delay).
...
+      Publisher-->>BlockNode: Send from block L+1 or send EndStream and retry with exponential backoff
+      Note over Publisher: Includes earliest and latest available blocks with EndStream
+      Note over BlockNode: Needs to catch up from another Block-Node
+    end
+    critical Validate latest streamed block
+      alt Block is verified and persisted
+        BlockNode-->>Publisher: Send `BlockAcknowledgement` with block number N
+        Note over Publisher: Continue with next block header
+      else Block fails verification
+        BlockNode-->>Publisher: Send `EndOfStream` "Bad State Proof" with block number L
```
ata-nas, line 169: Based on #1214, shall we update the diagram here with the updated definitions? I see `Bad State Proof` -> `Bad Block Proof`. Consider for other places as well.

docs/design/communication-protocol/publish-block-stream.md
```diff
@@ -63,6 +66,11 @@ stream and retry (either to another Block Node, or after a short delay).
...
 ## Base Protocol
 
+This protocol describes the basic interactions when a single Publisher is
+communicating with a single Block-Node. A subsequent section will describe
+the interactions when multiple Publishers are communicating with a single Block
+Node.
+
 * Publisher, on connect, sends a block header, this contains a block number.
   * If this is the next block, no problem, start streaming.
   * If this is less than last known verified block, respond with
```
ata-nas, line 76: Had the question for L:76, do we have an edge case here, where the `N < lastKnown` is actually not contained (known) by the block node?

Then answered below in lines L:82-88. Block-Node attempts to fill the gap, but still sends latest known. This is quite the critical distinction that this protocol is about streaming of live data and not backfill. 👍

ata-nas (approved):
@jsync-swirlds awesome job here!

Nana-EC (approved):
LG

PR #1179, opened 2025-05-22
bug: Fix of metrics start and addition of app level metrics

### LINKED ISSUE #1121: Metrics for App Module
### Story Form

### Technical Notes

Metrics for the App module:

Plugin | Type | Name | Description
-- | -- | -- | --
app | Gauge | app_historical_oldest_block | The oldest block the BN has
app | Gauge | app_historical_newest_block | The newest block the BN has
app | Gauge | app_state_status | Mapped to (0) Starting, (1) Running and (2) Shutting_Down on the Dashboard.

As part of this ticket, changes should be made to the BN Dashboard in both the Docker compose dev env and the Charts.

### LINKED ISSUE #1178: bug: Metrics are not being updated
### Description

While working on implementing new metrics I realized that NONE of the current metrics are working, their values remain the initialization ones no matter how many times they are set or incremented in runtime.

### Steps to Reproduce

Start BN
Stream a few Blocks
Get Single Block Latest a couple timees
Check :9999/metrics endpoint
Verify that none of the metrics that should have been update has, including `hiero_block_node_single_block_requests_total`

 

### Additional context

_No response_

AlfredoG87:
- Fixed Bug on how metrics are being started, so they actually start.
- Added App core metrics.

Fixes #1121
Fixes #1178

### DIFF
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -6,12 +6,14 @@ import static java.lang.System.Logger.Level.INFO;
 import static org.hiero.block.common.constants.StringsConstants.APPLICATION_PROPERTIES;
 import static org.hiero.block.node.app.logging.CleanColorfulFormatter.GREY;
 import static org.hiero.block.node.app.logging.CleanColorfulFormatter.LIGHT_GREEN;
+import static org.hiero.block.node.spi.BlockNodePlugin.METRICS_CATEGORY;
 
 import com.hedera.pbj.grpc.helidon.config.PbjConfig;
 import com.swirlds.common.metrics.platform.DefaultMetricsProvider;
 import com.swirlds.config.api.Configuration;
 import com.swirlds.config.api.ConfigurationBuilder;
 import com.swirlds.config.extensions.sources.ClasspathFileConfigSource;
+import com.swirlds.metrics.api.LongGauge;
 import com.swirlds.metrics.api.Metrics;
 import io.helidon.webserver.ConnectionConfig;
 import io.helidon.webserver.WebServer;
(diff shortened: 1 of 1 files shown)


block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -179,7 +197,7 @@ void start() {
         // Start the web server
         webServer.start();
         // Start metrics
-        blockNodeContext.metrics().start();
+        metricsProvider.start();
```
AlfredoG87, line 200: This is the FIX, the rest of the code is for the new metrics that will be updated by the `app` module.
Not really urgent as this fix is for `0.11.0` if preferred, I can make another PR for them, and leave this only with the fix and potentially a test.

Nana-EC (approved):

georgi-l95 (approved):

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -168,6 +176,16 @@ public class BlockNodeApp implements HealthFacility {
...
                         .build())
                 .build();
+
+        // Init the app metrics
+        historicalOldestBlockGauge =
+                metrics.getOrCreate(new LongGauge.Config(METRICS_CATEGORY, "app_historical_oldest_block")
+                        .withDescription("The oldest block the BN has access to"));
+        historicalLatestBlockGauge =
+                metrics.getOrCreate(new LongGauge.Config(METRICS_CATEGORY, "app_historical_newest_block")
+                        .withDescription("The newest block the BN has"));
```
jsync-swirlds, line 186: Just a note that these should _probably_ be part of the block provider plugin.  The app can be deployed _without any_ plugin that supports storing and providing blocks (and in fact we have real use cases asking for this now).

PR #1181, opened 2025-05-22
chore: rename `publisher` to `stream-publisher`

### LINKED ISSUE #1145: Rename the `publisher` plugin to `stream-publisher`
### Story Form

As a Block Node Operator
I want to have clearly named plugins
So that I can easily understand what each plugin offers

### Technical Notes

The subscriber was replaced with `stream-subscriber`, and [comments on the relevant PR](https://github.com/hiero-ledger/hiero-block-node/pull/1133#discussion_r2087252751) asked to rename the publisher in similar manner.  This issue is to complete that request.

georgi-l95:
## Reviewer Notes
Changes:
- Renamed module from `org.hiero.block.node.publisher` to `org.hiero.block.node.stream.publisher` for better alignment with related modules
- Updated package structure across all related files to reflect this new naming convention
- Modified `settings.gradle.kts` to reflect the renamed module
- Updated `block-node/app/build.gradle.kts` to use the new module name in runtime dependencies

## Related Issue(s)
Fixes #1145

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -21,10 +21,10 @@ javaModules {
         module("block-providers/files.recent") { artifact = "block-node-blocks-file-recent" }
         module("health") { artifact = "block-node-health" }
         module("messaging") { artifact = "facility-messaging" }
-        module("stream-publisher") { artifact = "block-node-stream-publisher" }
         module("s3-archive") { artifact = "block-node-s3-archive" }
         module("server-status") { artifact = "block-node-server-status" }
         module("spi") { artifact = "block-node-spi-plugins" }
+        module("stream-publisher") { artifact = "block-node-stream-publisher" }
         module("stream-subscriber") { artifact = "block-node-stream-subscriber" }
         module("verification") { artifact = "block-node-verification" }
     }


settings.gradle.kts
```diff
@@ -21,7 +21,7 @@ javaModules {
         module("block-providers/files.recent") { artifact = "block-node-blocks-file-recent" }
         module("health") { artifact = "block-node-health" }
         module("messaging") { artifact = "facility-messaging" }
-        module("publisher") { artifact = "block-node-publisher" }
+        module("stream-publisher") { artifact = "block-node-stream-publisher" }
```
AlfredoG87, line 24: these names are supposed to be alphabetically ordered same as they appear on the IDE for easy reference, however a refactor like this that could impact many conflicts, I don't wanna hold it up for such a nit, we can address this in the future on any other small improvement

AlfredoG87 (dismissed):
LGTM, just one small nit that can be addressed later.

Nana-EC (dismissed):
LG

settings.gradle.kts
georgi-l95: You're right. Fixed it 👍

AlfredoG87 (approved):
LG 💯

Issue #1184, opened 2025-05-22
Protobuf artifact fails upload


Nana-EC:
### Description

The https://github.com/hiero-ledger/hiero-block-node/blob/main/.github/workflows/release-push-image.yaml#L79 tries to upload an artifact but fails for a mismatch of paths

### Steps to Reproduce

Run release flow.
Th job succeeds but fails to upload with a 
```
Warning: No files were found with the provided path: ./block-node-protobuf-0.11.0-rc1.tgz. No artifacts will be uploaded.
```

### Additional context

See https://github.com/hiero-ledger/hiero-block-node/actions/runs/15192545249/job/42729324247 for an example

The issue is the path.
The tar creation is in the `protobuf` working dir

Nana-EC:
Closed with https://github.com/hiero-ledger/hiero-block-node/pull/1186

PR #1185, opened 2025-05-22
fix: (release/0.11) Add protobuf path to artificat upload and fail on error

Nana-EC:
## Reviewer Notes

The `package-block-node-protobuf` jobs upload step succeeded but doesn't upload anything
- Update path of uplaod to use `protobuf`
- Set `if-no-files-found` to `error` so we don't have silent failures

## Related Issue(s)
Partially addresses #1184


AlfredoG87 (approved):
LGTM

andrewb1269hg (approved):

PR #1191, opened 2025-05-23
fix: Opening for reflection, so Simulator can run in Consumer mode using modules (Docker or K8s)

### LINKED ISSUE #1190: Blockstrem simulator on Consumer mode fails at start-up due to reflection issues
### Description

Similar to the [issue](https://github.com/hiero-ledger/hiero-block-node/issues/1173#issuecomment-2896191233)
But a different path is needed for simulator on consumer mode.  

### Steps to Reproduce

Run Simulator on consumer mode in docker or k8s

### Additional context

there might be more packages to add after the first one is added, need to verify until is working

AlfredoG87:
Changes:
- Adding Opens for all the required proto packages in consumer mode and publisher mode.

Fixes #1190

### DIFF
--- a/protobuf/src/main/java/module-info.java
+++ b/protobuf/src/main/java/module-info.java
@@ -98,6 +98,25 @@ module org.hiero.block.protobuf {
     requires static com.github.spotbugs.annotations;
     requires static java.annotation;
 
+    // only open protoc to com.google.protobuf
+    opens com.hedera.hapi.platform.state.legacy to
+            com.google.protobuf;
     opens org.hiero.block.api.protoc to
+            com.google.protobuf;
+    opens org.hiero.block.internal.protoc to
+            com.google.protobuf;
+    opens com.hedera.services.stream.proto to
+            com.google.protobuf;
+    opens com.hederahashgraph.api.proto.java to
+            com.google.protobuf;
+    opens com.hederahashgraph.service.proto.java to
+            com.google.protobuf;
+    opens com.hedera.hapi.block.stream.protoc to
+            com.google.protobuf;
+    opens com.hedera.hapi.block.stream.input.protoc to
+            com.google.protobuf;
+    opens com.hedera.hapi.block.stream.output.protoc to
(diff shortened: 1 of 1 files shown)


protobuf/src/main/java/module-info.java
```diff
@@ -98,6 +98,26 @@
...
+    opens com.hedera.hapi.block.stream.protoc to
+            com.google.protobuf;
+    opens com.hedera.hapi.block.stream.input.protoc to
+            com.google.protobuf;
+    opens com.hedera.hapi.block.stream.output.protoc to
+            com.google.protobuf;
+    opens com.hedera.hapi.platform.event.legacy to
+            com.google.protobuf;
+
 }
```
AlfredoG87, line 123: should we not just open it in general? 

```suggestion
                // open for reflection commonly used by com.google.protobuf
    opens com.hedera.hapi.platform.state.legacy;
    opens org.hiero.block.api.protoc;
    opens org.hiero.block.internal.protoc;
    opens com.hedera.services.stream.proto;
    opens com.hederahashgraph.api.proto.java;
    opens com.hederahashgraph.service.proto.java;
    opens com.hedera.hapi.block.stream.protoc;
    opens com.hedera.hapi.block.stream.input.protoc;
    opens com.hedera.hapi.block.stream.output.protoc;
    opens com.hedera.hapi.platform.event.legacy;

}
```

Nana-EC (approved):
LG
Let's circle back in the future to see if there's a different approach.

georgi-l95 (approved):

protobuf/src/main/java/module-info.java
jsync-swirlds: We _just_ restricted these classes.
the protoc compiled classes are toxic to code that's trying to use PBJ, and we don't want them in the build.
The proper resolution for these is to split the code that uses protoc (simulator and E2E suites) from the main build and have _only_ protoc used for that build while _only_ PBJ is used for the main build.

PR #1196, opened 2025-05-23
fix: Opening for reflection, so Simulator can run in Consumer mode

### LINKED ISSUE #1190: Blockstrem simulator on Consumer mode fails at start-up due to reflection issues

AlfredoG87:
Cherry-pick of (#1191) into `release/0.11` so we can unblock complete testing workflows

Fixes #1190

### DIFF
--- a/protobuf/src/main/java/module-info.java
+++ b/protobuf/src/main/java/module-info.java
@@ -98,6 +98,25 @@ module org.hiero.block.protobuf {
     requires static com.github.spotbugs.annotations;
     requires static java.annotation;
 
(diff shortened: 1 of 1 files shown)


Nana-EC (approved):
LG

PR #1235, opened 2025-06-02
feat: Add Metrics for Block-Providers Files.Recent, Files.Historic, Server Status and Summary Full History Dashboard

### LINKED ISSUE #1127: Metrics for `block-providere/files.recent` plugin
### Story Form

### Technical Notes

Plugin | Type | Name | Description
-- | -- | -- | --
files.recent | Counter | files_recent_blocks_written | The number of blocks written to disk
files.recent | Counter | files_recent_blocks_read | The number of blocks read from disk
files.recent | Counter | files_recent_blocks_deleted | The number of blocks deleted from disk
files.recent | Gauge | files_recent_blocks_stored | The number of blocks stored on disk for this plugin
files.recent | Gauge | files_recent_total_bytes_storerd | The total number of bytes stored on disk for this plugin

### LINKED ISSUE #1128: Metrics for `Block-Providers/files.historic` plugin
### Story Form

### Technical Notes

Plugin | Type | Name | Description
-- | -- | -- | --
files.historic | Counter | files_historic_blocks_written | The number of blocks written to disk
files.historic | Counter | files_historic_blocks_read | The number of blocks read from disk
files.historic | Counter | files_historic_blocks_deleted | The number of blocks deleted from disk
files.historic | Gauge | files_historic_blocks_stored | The number of blocks stored on disk for this plugin
files.historic | Gauge | files_historic_total_bytes_stored | The total number of bytes stored on disk for this plugin

### LINKED ISSUE #1132: Metrics Dashboard Overview
### Story Form

### Technical Notes

Each of the Plugins will create it's own set of metrics and also its own dashboard for Grafana.

However this ticket is to create a General Overview of the system health, pick and choosing the most important metrics accross plugins and show an unified, compact view of the whole system on a single dashboard.

### LINKED ISSUE #1242: Metrics for Server Status Plugin
### Story Form

As a BN Operator
I want to measure request to server status
So that I can make calculations of load on given endpoints

### Technical Notes

Metric is already implemented, missing dashboard.

AlfredoG87:
## Reviewer Notes
- Added Metrics for Block.Provider **Files.Recent** Plugin.
- Added Grafana Dashboard for the above plugin.
- Added Metrics for BlockProvider **Files.Historic** Plugin
- Grafana Dashboard for Historic Provider Plugin
- Improved ServerStatus Metrics **server status** Plugin
- Grafana Dashboard for Server Status Metrics
- **Block-Node: Full-History** Dashboard as a Compact, most important Dashboard for a quick overview of the whole system.

- Bonus, Fixed/Improved naming of server status request and added it to the metrics document.
- Bonus 2, Fixed Metric (blockItemsReceivedCounter) on Messaging facility, was severely under-reeporting, was reporting batches, instead of block_items.

## Screenshots:

**Files.Recent**
image

**Files.Historic**
image

**Server Status**
image

**Block-Node: Full-History**
image
image

## Related Issue(s)
Fixes #1127 
Fixes #1128 
Fixes #1242 
Fixes #1132

### DIFF
--- a/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
+++ b/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
@@ -49,6 +49,11 @@ class ZipBlockArchive {
     private final Format format;
 
     /**
+     * Record to hold the result of creating a zip file including the block accessors and file size.
+     */
+    record ZipFileCreationResult(List<BlockAccessor> blockAccessors, long zipFileSize) {}
+
+    /**
      * Constructor for ZipBlockArchive.
      *
      * @param context The block node context
@@ -69,9 +74,9 @@ class ZipBlockArchive {
      *
      * @param firstBlockNumber The first block number to write
      * @throws IOException If an error occurs writing the block
-     * @return A list of block accessors for the blocks written to the zip file, can be used to delete the blocks
+     * @return A record containing block accessors and the size of the zip file created
      */
-    List<BlockAccessor> writeNewZipFile(long firstBlockNumber) throws IOException {
+    ZipFileCreationResult writeNewZipFile(long firstBlockNumber) throws IOException {
         final long lastBlockNumber = firstBlockNumber + numberOfBlocksPerZipFile - 1;
         // compute block path
         final BlockPath firstBlockPath = computeBlockPath(config, firstBlockNumber);
@@ -116,10 +121,12 @@ class ZipBlockArchive {
                 zipOutputStream.closeEntry();
             }
         }
+        // Compute zip file size without scanning the entire filesystem
+        long zipFileSize = Files.size(firstBlockPath.zipFilePath());
         // return block accessors
         // todo should these accessors be returned? they were here because a delete
         //   was supposed to be called on them, but we have changed the approach
-        return blockAccessors;
+        return new ZipFileCreationResult(blockAccessors, zipFileSize);
     }
 
     /**
@@ -240,5 +247,32 @@ class ZipBlockArchive {
             }
         }
         return -1;
+    }
+
+    /**
+     * Calculates the total size of all stored zip files in the archive.
+     *
+     * @return The total bytes stored in all zip files
+     */
+    long calculateTotalStoredBytes() {
+        try (Stream<Path> pathStream = Files.walk(config.rootPath())) {
+            return pathStream
+                    .filter(Files::isRegularFile)
+                    .filter(path -> path.getFileName().toString().endsWith(".zip"))
+                    .mapToLong(file -> {
+                        try {
+                            return Files.size(file);
+                        } catch (IOException e) {
+                            LOGGER.log(System.Logger.Level.WARNING,
+                                "Failed to get size of file: " + file, e);
+                            return 0;
+                        }
+                    })
+                    .sum();
+        } catch (IOException e) {
+            LOGGER.log(System.Logger.Level.ERROR,
+                "Error walking directory structure to calculate total bytes stored", e);
+            return 0;
+        }
     }
 }
--- a/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
+++ b/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
@@ -51,11 +51,6 @@ class ZipBlockArchive {
     private final Format format;
 
     /**
-     * Record to hold the result of creating a zip file including the block accessors and file size.
-     */
-    record ZipFileCreationResult(List<BlockAccessor> blockAccessors, long zipFileSize) {}
-
-    /**
      * Constructor for ZipBlockArchive.
      *
      * @param context The block node context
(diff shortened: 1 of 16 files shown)


block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
```diff
@@ -116,10 +123,12 @@ List<BlockAccessor> writeNewZipFile(long firstBlockNumber) throws IOException {
                 zipOutputStream.closeEntry();
             }
         }
+        // Compute zip file size without scanning the entire filesystem
+        long zipFileSize = Files.size(firstBlockPath.zipFilePath());
         // return block accessors
         // todo should these accessors be returned? they were here because a delete
         //   was supposed to be called on them, but we have changed the approach
```
jsync-swirlds, line 130: This is a good question.
We don't really have a reason to be returning the accessors; so perhaps remove that and just return the zip file size?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
```diff
@@ -241,4 +250,29 @@ long maxStoredBlockNumber() {
         }
         return -1;
     }
+
+    /**
+     * Calculates the total size of all stored zip files in the archive.
```
jsync-swirlds, line 255: Might be good to add a warning here that this should not be called except on startup, and even then we might later remove this because it can take a very long time (potentially many minutes, if we have a lot of files and file size queries are expensive).

ata-nas (commented):
Looks good @AlfredoG87! A few comments.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
ata-nas: P.S. for the zips, do we really expect massive amount of files? If we generally batch 100k blocks in a single zip file, the historic root will be growing very slowly (relatively speaking) with 1 zip for each day and a bit (86k seconds in a day, streaming at 1 block a sec). Just a thought. If we batch 10k however, this changes things, but still, having ~350 files per year to ~3500 does not seem that much or am I mistaken?

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
```diff
@@ -115,7 +135,53 @@ public void init(final BlockNodeContext context, final ServiceBuilder serviceBui
...
+                    try {
+                        Path blockFilePath = BlockFile.nestedDirectoriesBlockFilePath(
+                                config.liveRootPath(), blockNumber, config.compression(), config.maxFilesPerDir());
+                        if (Files.exists(blockFilePath)) {
+                            totalBytesStored.addAndGet(Files.size(blockFilePath));
+                        }
+                    } catch (IOException e) {
+                        LOGGER.log(WARNING, "Failed to get size of block file for block " + blockNumber, e);
+                    }
+                });
```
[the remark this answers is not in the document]
jsync-swirlds: It is a bit tricky here.  There are two concerns:
1. The live root will be shared with state management (in full history/full service nodes); and state management demands maximum disk performance
2. The live root might be up to a terabyte or so in size (depending on _operator_ configuration).

Given those two concerns; it is worth being _careful_ to not scan the filesystem any more than strictly necessary.  Certainly (in my opinion) we _should not_ scan the filesystem for every block (perhaps every thousandth is acceptable, probably not any more than that).

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
AlfredoG87: I've create this ticket https://github.com/hiero-ledger/hiero-block-node/issues/1249

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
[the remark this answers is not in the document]
ata-nas: Interesting. What would be the idea behind having the same physical space for blocks and state? Still, I would  imagine the actual live root will be different than state or are they literally the same? (e.g. /opt/live/blocks - /opt/live/state). In that case no redundant files will be scanned and also this operation is once per startup before the node is up and running, so performance wise I do not think it is a concern. But the excessively large amount of blocks might delay a startup for sure.

ata-nas (dismissed):
Good improvements and cleanup @AlfredoG87, let's not forget to add the tests mentioned in my previous thread in a follow up. May I suggest we open an issue to track that?

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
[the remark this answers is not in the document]
AlfredoG87: I guess the NVMe drives are way more critical for State Management than for Live Blocks, so it makes sense for them to use the `live` mount that has NVME based storage PVC. 👍 

In the files.recent plugin the method does not exist, since there we were doing the scan anyways, I just added the measurement of the File there and not on a separate method.

we could in both cases do as a low priority background process that will just add to the metric when finishes or as it goes. (will add this idea to Improvement follow-up issue)

[Refactor: Optimize how to read existing files on historic and recent files plugin at startup.](https://github.com/hiero-ledger/hiero-block-node/issues/1249)

ata-nas (approved):

jsync-swirlds (approved):

PR #1245, opened 2025-06-03
fix: 1227 subscriber plugin does not close failed streams

### LINKED ISSUE #1227: bug: Subscriber plugin is not calling handleCompletedStream
### Description

While working on the Metrics for Subscriber plugin I was able to identify an issue, seems like we are not closing subscriber connections, or if we are, we are not able to get the right number.

So every time I subscribe, a new session is created and the session even if is over, is not removed, since it looks like the `handleCompletedStream` is not being called.

### Steps to Reproduce

1. Start BN Server using DockerStart
2. Start Simulator in both publisher and consumer
3. Go to Dashboards for BN `localhost:3000` and find the Subscriber plugin Dashboard.
4. Initially should be 1 open connection
5. Open and close several connections using Postman
6. Go to dashboard again and update it, see that open connections never goes down.

### Additional context

Image

Metrics for plugin are not yet merged to `main` but you can use branch: `new-metrics-3` while they get merged.

jsync-swirlds:
## Reviewer Notes
* The Helidon plugin throws a `RuntimeException` to indicate a stream that was closed (even if _orderly_).
    * This change handles that exception by closing the session and ending the thread.

## Related Issue(s)
 Fixes #1227


block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -518,7 +518,17 @@ private void sendOneBlockItemSet(final BlockItems nextBatch) {
...
+        try {
+            responsePipeline.onNext(new SubscribeStreamResponseUnparsed(responseOneOf));
+        } catch (RuntimeException e) {
+            // If the pipeline is in an error state; close this session.
+            // Unfortunately this is the "standard" way to end a stream, so log
+            // at debug rather than emitting noise in the logs.
+            final String message =
+                    "Client error sending block items for client %d: %s".formatted(clientId, e.getMessage());
+            LOGGER.log(Level.DEBUG, message, e);
+            close(null); // cannot send the end stream response, just close the stream.
```
ata-nas, line 530: The synchronized `close` method is now called by a lot of places that look like they could be competing for it at first glance. Just noting.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -148,14 +142,90 @@ void shouldStreamHistoricalBlocksSuccessfully() {
...
+            // Initialize session
+            final SubscribeStreamRequest subscribeStreamRequest = createRequest(START_BLOCK, END_BLOCK);
+            setupHistoricalBlockProvider(MIN_AVAILABLE_BLOCK, MAX_AVAILABLE_BLOCK + 1);
+            session = new BlockStreamSubscriberSession(
+                    CLIENT_ID, subscribeStreamRequest, disconnectedPipeline, context, sessionReadyLatch);
+            try (final ExecutorService sessionExecutor = Executors.newVirtualThreadPerTaskExecutor()) {
+                Future<BlockStreamSubscriberSession> sessionFuture = sessionExecutor.submit(session);
+                sessionReadyLatch.await();
+                // Phase 1: Historical Block Streaming
+                final int expectedResponseCount = END_BLOCK - START_BLOCK;
```
ata-nas, line 184: L:184 is unused:
```java
final int expectedResponseCount = END_BLOCK - START_BLOCK;
```
Should we remove this or do something with it?

ata-nas (approved):
@jsync-swirlds looks good. I noticed an unused variable.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSessionTest.java
```diff
@@ -148,14 +142,90 @@ void shouldStreamHistoricalBlocksSuccessfully() {
...
+            final ResponsePipeline disconnectedPipeline = new ResponsePipeline(true);
+
+            // Initialize session
+            final SubscribeStreamRequest subscribeStreamRequest = createRequest(START_BLOCK, END_BLOCK);
+            setupHistoricalBlockProvider(MIN_AVAILABLE_BLOCK, MAX_AVAILABLE_BLOCK + 1);
+            session = new BlockStreamSubscriberSession(
+                    CLIENT_ID, subscribeStreamRequest, disconnectedPipeline, context, sessionReadyLatch);
+            try (final ExecutorService sessionExecutor = Executors.newVirtualThreadPerTaskExecutor()) {
+                Future<BlockStreamSubscriberSession> sessionFuture = sessionExecutor.submit(session);
+                sessionReadyLatch.await();
```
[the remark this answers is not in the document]
jsync-swirlds: Added timeouts for 10 seconds on all of the test methods; just in case.

ata-nas (approved):
:fire:

PR #1261, opened 2025-06-05
feat: add a "slow mode" feature to the simulator in consumer mode

### LINKED ISSUE #837: Add a "slow mode" feature to the simulator in consumer mode
### Story Form

As a block node engineer
I want to test the block node response to a "slow" consumer client
So that I can verify "backpressure", transitions to historical mode, transitions from historical to live, and related functionality.

### Technical Notes

We wish to accomplish all of the following:
- Add options to control how fast the simulator accepts blocks
   - Accept blocks instantly, assume good, do not parse, discard the data
   - Accept blocks normally, parse, verify the block proof, discard after verification
   - Accept blocks normally, inject "slow down" delays, parse, verify the block proof, discard after verification
      -  The base "slow down" is a pause for a random time in μs within a configured range for each received batch.
         - Have timers where we wait a random number of blocks, within a configured range,  before we "go slow"
         - Have timers to "go slow" for a random number of blocks within a configured range
         - Return to full speed after the "go slow" period elapses.
      - This variable cycle _may_ be repeated, waiting a random number of blocks, within a range, between repeats.
      - For CI testing where we need repeatable results, ranges would have the same start/end value, so "random" only selects that single value.
- Integrate these options with other options for errors and error handling (e.g. failed verification).
- Add a few end-to-end tests that make use of this new functionality.

As part of this effort, we may also wish to replace [this test](https://github.com/hiero-ledger/hiero-block-node/tree/main/block-node/server/src/test/network-latency-simulator) with other testing that makes use of the newly added "slow" options.

mustafauzunn:
## Reviewer Notes

The base "slow down" is a pause for a random time in μs within a configured range for each received batch.
- Have timers where we wait a random number of blocks, within a configured range, before we "go slow"
- Have timers to "go slow" for a random number of blocks within a configured range
- Return to full speed after the "go slow" period elapses.

## Related Issue(s)
 
 Fixes #837


tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -47,11 +55,24 @@ public ConsumerStreamObserver(
...
+            this.slowDownBlockRangeSet = parseSlowDownForBlockRange(consumerConfig.slowDownForBlockRange());
+        } else if (consumerConfig.randomSlowDown()) {
+            List<Long> blockRange = parseBlockRange(consumerConfig.slowDownForBlockRange());
+            this.slowDownBlockRangeSet = randomBlockRangeSet(blockRange.get(0), blockRange.get(1));
+        } else if (consumerConfig.randomBlocksToWaitBeforeSlowDown()) {
+            List<Long> blockRange = parseBlockRange(consumerConfig.slowDownForBlockRange());
+            this.slowDownBlockRangeSet = parseSlowDownForBlockRange(consumerConfig.slowDownForBlockRange());
+            this.randomBlocksToWait =
+                    new Random().nextLong(blockRange.get(1) - blockRange.get(0) + 1) + blockRange.get(0);
+        }
```
ata-nas, line 75: I feel like this should be simpler. This if-else looks like a switch-case for an enum. I think that if we transform these (what seems to be) strategies to an enum, we will have only 1 config field for them (less configuration) and the code would be much simpler.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -47,11 +55,24 @@ public ConsumerStreamObserver(
...
             @NonNull final CountDownLatch streamLatch,
             @NonNull final Deque<String> lastKnownStatuses,
-            final int lastKnownStatusesCapacity) {
+            final int lastKnownStatusesCapacity,
+            @NonNull final ConsumerConfig consumerConfig) {
         this.metricsService = requireNonNull(metricsService);
         this.streamLatch = requireNonNull(streamLatch);
         this.lastKnownStatuses = requireNonNull(lastKnownStatuses);
         this.lastKnownStatusesCapacity = lastKnownStatusesCapacity;
+        this.consumerConfig = requireNonNull(consumerConfig);
```
ata-nas, line 64: No need for the null check on L:64, we are calling a method on that reference, if it is null, we will throw. If we only assign to a field however, the check is mandatory as otherwise we will have to search for the null's origin a lot.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -34,6 +38,10 @@ public class ConsumerStreamObserver implements StreamObserver<SubscribeStreamRes
     private final Deque<String> lastKnownStatuses;
     private final AtomicLong blocksConsumed = new AtomicLong(0);
 
+    private final ConsumerConfig consumerConfig;
+    private Set<Long> slowDownBlockRangeSet = new HashSet<>();
```
ata-nas, line 42: The set can be `final`, no need to create an instance here, if the slowdown checks fail, we make an empty one in an `else`.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -122,4 +155,65 @@ private void logNonAscendingBlockNumbers(long blockNumber) {
             }
         }
     }
+
+    private void slowDownProcessing(final String message) {
+        try {
+            LOGGER.log(INFO, "Slowing down processing " + message);
```
ata-nas, line 161: This log may be a `DEBUG`. This seems a bit noisy maybe. We have a document that you can refer to (I think we should also use that for the simulator as well - `logging-guidelines.md`)

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -99,6 +120,18 @@ public void onCompleted() {
     }
 
     private void processBlockItems(List<BlockItem> blockItems) {
+        long lastBlockConsumed = blocksConsumed.get();
+        if (consumerConfig.slowDown() || consumerConfig.randomSlowDown()) {
```
ata-nas, line 124: If we need to pass the `consumerConfig.slowDown()` in order to have a slowdown, does that also mean that we should pass that in all other `else-if` cases? Otherwise if the first `if` fails, because `consumerConfig.slowDown()` is false, the second one might pass as it does not check again. (this would all be made simpler if we move to an enum)

tools-and-tests/simulator/src/test/java/org/hiero/block/simulator/grpc/impl/ConsumerStreamObserverTest.java
```diff
@@ -34,30 +36,35 @@ class ConsumerStreamObserverTest {
     private ConsumerStreamObserver observer;
     private int lastKnownStatusesCapacity;
 
+    @Mock
+    private ConsumerConfig consumerConfig;
+
```
ata-nas, line 41: Let's not mock configs, but rather deliver them via the config api like we generally do for tests (some are not migrated, yes).

ata-nas (commented):
@mustafauzunn looks good on first pass, I leave some cleanup suggestions.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/config/data/ConsumerConfig.java
```diff
@@ -8,4 +8,9 @@
 @ConfigData("consumer")
 public record ConsumerConfig(
         @Loggable @ConfigProperty(defaultValue = "-1") long startBlockNumber,
-        @Loggable @ConfigProperty(defaultValue = "-1") long endBlockNumber) {}
+        @Loggable @ConfigProperty(defaultValue = "-1") long endBlockNumber,
+        @Loggable @ConfigProperty(defaultValue = "false") boolean slowDown,
+        @Loggable @ConfigProperty(defaultValue = "false") boolean randomSlowDown,
+        @Loggable @ConfigProperty(defaultValue = "2") long slowDownMilliseconds,
+        @Loggable @ConfigProperty(defaultValue = "false") boolean randomBlocksToWaitBeforeSlowDown,
+        @Loggable @ConfigProperty(defaultValue = "10-30") String slowDownForBlockRange) {}
```
AlfredoG87, line 16: agree with @ata-nas  that we should try to simplify config values here and that an enum between ` NONE`, `SLOW_DOWN`, `RANDOM_SLOW_DOWN`, `SLOW_DOWN_FOR_RANGE`. could dbe called **SlowDownType**

and simplify not only configuration, but also implementation logic.

Finally, do we have a config document with the detaile description of each config?

AlfredoG87 (commented):
Agree with @ata-nas  suggestion to simplify both config and impl.

jsync-swirlds:
Quick note, you can save round-trips to CI by running the same checks locally:
```shell
export GRADLE_EXEC=`nice -n 20 ./gradlew `
export GRADLE_OPTS='--rerun-tasks --no-daemon --no-build-cache --no-configuration-cache --warning-mode=summary --parallel --max-workers=12 -Dtestlogger.showPassed=false -Dtestlogger.showSkipped=false -Dtestlogger.showPassedStandardStreams=false -Dtestlogger.theme=mocha-parallel '
# Run the QualityChecks and UTs task
${GRADLE_EXEC} ${GRADLE_OPTS} clean check
# Run the E2E test suites task
${GRADLE_EXEC} ${GRADLE_OPTS} clean runSuites
```

ata-nas (approved):
Great job @mustafauzunn! I leave some non-blocking nits 🙌 🔥

ata-nas (approved):
🔥

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/config/data/ConsumerConfig.java
jsync-swirlds: Also, adding javadoc _in addition_ is an excellent idea.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/ConsumerStreamObserver.java
```diff
@@ -122,4 +137,14 @@ private void logNonAscendingBlockNumbers(long blockNumber) {
...
         }
     }
+
+    private void slowDownProcessing(final String message) {
+        try {
+            LOGGER.log(DEBUG, "Slowing down processing " + message);
+            Thread.sleep(consumerConfig.slowDownMilliseconds());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            LOGGER.log(ERROR, "Stream processing interrupted during slowdown", e);
```
jsync-swirlds, line 147: This isn't an error; I would recommend that `INFO` be used here.  Interruption is fairly normal and most often represents a requested process exit.

ata-nas (approved):
🔥

Issue #738, opened 2025-02-28
Process to attempt re-running failed archiving tasks


ata-nas:
### Persona

As a Block Node Operator/User

### Request

I want to have the ability to re-trigger a failed archiving task

### Goal

So that my archive will be in order

### Technical Notes

- the implementation of the local block archiver introduced in #517 will run an archiving task to completion, be that a successful run or not
- if a run is unsuccessful, then the live data should be intact and we should be able to attempt another archive task

**IMPORTANT:**
- a failed run might be unsuccessful for different reasons, we need to make sure we can differentiate these reasons so we can make proper decisions
- or we must ensure that whatever the reason for failure is, we need to ensure proper environment for the archiver so it can run again, being agnostic to any differences
   - a couple of considerations here would be:
      1. if an archive (zipping) is successful, but then deleting of live blocks is unsuccessful be that during the move of the live data to an undiscoverable root to be deleted, making way for the archive to be discovered, or the delete itself fails, it makes little to no sense to delete the already zipped data, only the removal of the live data needs to be finished in another archive task run
      1. if an archive fails during zipping, the zip should be removed if it exists and the whole operation should be attempted again
   - these edge cases would potentially benefit if the archiver is publishing meaningful persistent results for each run which then could be acted upon accordingly

ata-nas:
@Nana-EC @jasperpotts @jsync-swirlds @AlfredoG87 @georgi-l95 I think this is still a good idea, we should transform this issue to fit the new logic and we should move this. A retry attempt is a good idea.

Objections?

Nana-EC:
> [@Nana-EC](https://github.com/Nana-EC) [@jasperpotts](https://github.com/jasperpotts) [@jsync-swirlds](https://github.com/jsync-swirlds) [@AlfredoG87](https://github.com/AlfredoG87) [@georgi-l95](https://github.com/georgi-l95) I think this is still a good idea, we should transform this issue to fit the new logic and we should move this. A retry attempt is a good idea.
> 
> Objections?

None, agreed.

PR #1262, opened 2025-06-05
refactor: Files Historic Plugin Ensure No Gaps in a Zip Can Happen

### LINKED ISSUE #1217: Files Historic Plugin Ensure No Gaps in a Zip Can Happen
### Story Form

I want to make sure that whenever I zip a range of blocks there will not be any gaps in the zipped range
So that I can be absolutely sure all blocks in the range are zipped and remove all complexity in handling gaps otherwise

### Technical Notes

- we recently discussed that whenever we try to zip a range of blocks, we should not allow any gaps in the range
- we should zip only if we are sure we have all the blocks in the range available
- if we notice a gap, this is not an erroneous state, but rather we should keep trying until we have the full range to zip

ata-nas:
## Reviewer Notes

- refactored the batching logic to not allow gaps in the zips
- the zip archiver class is no longer the one that batches as it should not be it's responsibility
  - the zip archiver takes a readied batch of block accessors and then proceeds to zip everything

## Related Issue(s)

Resolves #1217

### DIFF
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/InMemoryBlockAccessor.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/InMemoryBlockAccessor.java
@@ -3,6 +3,7 @@ package org.hiero.block.node.app.fixtures.blocks;
 
 import com.hedera.hapi.block.stream.Block;
 import com.hedera.hapi.block.stream.BlockItem;
+import com.hedera.hapi.block.stream.output.BlockHeader;
 import java.util.List;
 import org.hiero.block.node.spi.historicalblocks.BlockAccessor;
 
@@ -17,32 +18,53 @@ public final class InMemoryBlockAccessor implements BlockAccessor {
     private final long blockNumber;
 
     /**
-     * Constructor.
+     * Constructor. Enforces preconditions on the input block items.
      *
-     * @param blockItems of the block, must not be empty, first item must be a
-     * {@link com.hedera.hapi.block.stream.output.BlockHeader} and last item
-     * must be a {@link com.hedera.hapi.block.stream.BlockProof}.
-     * @throws IllegalArgumentException if preconditions are not met
+     * @param blockItems of the block, if no block header is present, the
+     * block number will be set to
+     * {@value org.hiero.block.node.spi.BlockNodePlugin#UNKNOWN_BLOCK_NUMBER}
      */
     public InMemoryBlockAccessor(final List<BlockItem> blockItems) {
-        if (blockItems.isEmpty()) {
-            throw new IllegalArgumentException("Block items list cannot be empty");
-        }
-        // Ensure the first item is a BlockHeader
-        if (!blockItems.getFirst().hasBlockHeader()) {
-            throw new IllegalArgumentException("First block item must be a BlockHeader");
-        }
-        // Ensure the last item is a BlockProof
-        if (!blockItems.getLast().hasBlockProof()) {
-            throw new IllegalArgumentException("Last block item must be a BlockProof");
+        this(blockItems, true);
+    }
+
+    /**
+     * Constructor.
+     *
+     * @param blockItems of the block, if no block header is present, the
+     * block number will be set to
+     * {@value org.hiero.block.node.spi.BlockNodePlugin#UNKNOWN_BLOCK_NUMBER}
+     * @param enforcePreconditions if true, will enforce preconditions on the
+     * input block items
+     * @throws IllegalArgumentException if preconditions are not met but
+     * enforced
+     */
+    public InMemoryBlockAccessor(final List<BlockItem> blockItems, final boolean enforcePreconditions) {
+        if (enforcePreconditions) {
+            if (blockItems.isEmpty()) {
+                throw new IllegalArgumentException("Block items list cannot be empty");
+            }
+            // Ensure the first item is a BlockHeader
+            if (!blockItems.getFirst().hasBlockHeader()) {
+                throw new IllegalArgumentException("First block item must be a BlockHeader");
+            }
+            // Ensure the last item is a BlockProof
+            if (!blockItems.getLast().hasBlockProof()) {
+                throw new IllegalArgumentException("Last block item must be a BlockProof");
+            }
         }
         // Create a Block from the provided block items
         block = Block.newBuilder().items(blockItems).build();
-        blockNumber = block.items().getFirst().blockHeader().number();
+        final BlockHeader header = block.items().getFirst().blockHeader();
+        blockNumber = header != null ? header.number() : -1;
     }
 
     /**
      * {@inheritDoc}
+     * <p>
+     * This method returns the block number of the accessible block or
+     * {@value org.hiero.block.node.spi.BlockNodePlugin#UNKNOWN_BLOCK_NUMBER}
+     * if the block header is not present as the first item in the block.
      */
     @Override
     public long blockNumber() {


block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
```diff
@@ -118,10 +100,17 @@ long writeNewZipFile(long firstBlockNumber) throws IOException {
                 // close zip entry
                 zipOutputStream.closeEntry();
             }
+        } catch (final UncheckedIOException e) {
+            // adding this because we are potentially throwing an unchecked
+            // io exception, if that is the case, we want to throw the cause
+            // which is expected, but if that is not the case, that may indicate
+            // we have an issue that we should look at
+            throw e.getCause();
```
ata-nas, line 108: Not a fan of this. The idea here is to catch unchecked IO exceptions and to propagate them so that the plugin then can handle itself properly. I really thing that the `BlockAccessor`'s API should throw `IOException`s for most of it's methods as it is very possible that even if we have created an accessor, it may not be able to actually read (or no longer can). Having the checked exception is very helpful.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
```diff
@@ -43,69 +42,52 @@ class ZipBlockArchive {
...
+        // create zip file
+        final Path zipFilePath = firstBlockPath.zipFilePath();
+        try (final ZipOutputStream zipOutputStream = new ZipOutputStream(new BufferedOutputStream(
+                Files.newOutputStream(zipFilePath, StandardOpenOption.CREATE, StandardOpenOption.WRITE),
                 1024 * 1204))) {
             // don't compress the zip file as files are already compressed
             zipOutputStream.setMethod(ZipOutputStream.STORED);
-            // todo should we not also set the level to Deflater.NO_COMPRESSION
-            for (long blockNumber = firstBlockNumber; blockNumber <= lastBlockNumber; blockNumber++) {
+            zipOutputStream.setLevel(Deflater.NO_COMPRESSION);
```
ata-nas, line 81: Setting the lever to no compression here:
```java
zipOutputStream.setLevel(Deflater.NO_COMPRESSION);
```

ata-nas (commented):
There are 3 key areas where we can check and catch gaps:
1. After a notification is received in the plugin, we can do a quick pre-check for min/max available in the historical block facility
1. If the above passes, then by the time we start to collect the batch, some accessors may become unavailable
1. If we have collected all of the above, then during zipping, an accessor might throw due to whatever reason, for example the block is no longer physically there (this is due to the async nature of the app)

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -207,12 +213,20 @@ public void handlePersisted(PersistedNotification notification) {
     // ==== Private Methods ============================================================================================
     private void attemptZipping() {
         // compute the min and max block in next batch to zip
-        long minBlockNumber = availableBlocks().max() + 1;
+        // since we ensure no gaps in the zip file are possible, and also we
+        // have a power of 10 number of blocks per zip file, it is safe to
+        // simply add +1 to the latest available block number and have that as
+        // the start of the next batch of blocks to zip
+        long minBlockNumber = availableBlocks.max() + 1;
```
ata-nas, line 220: We should probably check if the min % blocksPerZip == 0, if no, then we are somewhere in between the batch, we should skip. This will happen when the BN starts at some arbitrary number, if that is the case, a gap must be filled which is not the responsibility of this plugin. We should effectively skip this unsatisfiable (for now) batch. We will be able to redo it later, see #738.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
ata-nas: Interestig. This is also a way to do that. We should make sure one or the other is the case however, as it is clear we should generally be able to assert in some way that the accessor cannot supply the block in question.

jsync-swirlds: I believe returning null would be the correct option, yes.
I also would assert that any exception (checked or unchecked) is a risk we do not want.  Particularly for block accessors, it's always safe to just return that the block is not available and let the caller try again or send an appropriate response to a client.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
ata-nas: This has been added in the while loop as we need to make attempts to resolve the currentBatch + 1 if the currentBatch starts somewhere in the middle. After the currentBatch + 1 zips, the min will be updated. The skipped batch must be backfilled and retried (that is a separate issue).

AlfredoG87 (dismissed):
LGTM

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -234,17 +260,19 @@ private void attemptZipping() {
...
                     "Batch of blocks[{%1} -> {%2}] is already in progress",
                     batchRange.start(),
                     batchRange.end());
-            return;
+        } else {
+            // if the batch is not in progress, we must submit a task
+            // add the batch of blocks to the in progress ranges
+            inProgressZipRanges.add(batchRange);
+            // move the batch of blocks to a zip file (submit a task)
+            zipMoveExecutorService.submit(() -> moveBatchOfBlocksToZipFile(batchRange));
```
jsync-swirlds, line 274: Rather than a lambda, this could be a static inner class.
```suggestion
            zipMoveExecutorService.submit(new BatchToZipFileMover(batchRange));
```
It's a small difference, but it makes maintenance easier and reduces the probability of unplanned threading issues.

jsync-swirlds (approved):
A few suggestions.
Only recommendations, not required.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/InMemoryBlockAccessor.java
```diff
@@ -0,0 +1,61 @@
...
+            throw new IllegalArgumentException("First block item must be a BlockHeader");
+        }
+        // Ensure the last item is a BlockProof
+        if (!blockItems.getLast().hasBlockProof()) {
+            throw new IllegalArgumentException("Last block item must be a BlockProof");
+        }
+        // Create a Block from the provided block items
+        block = Block.newBuilder().items(blockItems).build();
+        blockNumber =
+                Objects.requireNonNull(block.items().getFirst().blockHeader()).number();
```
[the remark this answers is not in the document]
ata-nas: Hmm, that was probably an automatic resolution of an IDE warning...

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/SimpleInMemoryHistoricalBlockFacility.java
```diff
@@ -115,11 +118,40 @@ public void clearDisablePlugin() {
...
+     * original available blocks set, which are always updated. An example usage
+     * here would be to simulate a condition where a check for available blocks
+     * gives a specific set of blocks, but then later when we ask the facility
+     * to give us the available blocks, it will return what is in the original
+     * available blocks set. This simulates the condition where the check might
+     * become stale or outdated and what we get in terms of accessible blocks
+     * is what is currently available in the facility.
+     */
+    public void setTemporaryAvailableBlocks(@NonNull final SimpleBlockRangeSet availableBlocks) {
+        this.availableBlocksOverride = Objects.requireNonNull(availableBlocks);
```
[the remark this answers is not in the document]
ata-nas: If we provide null here this will be very misleading and will produce false positives possibly. I rely on null to either supply the overridden instance or not. If in my test I call this method and somehow I supply null, nothing will change and I will have the regular behavior. This was very intentional.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/InMemoryBlockAccessor.java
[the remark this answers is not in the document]
ata-nas: ✅ This has been reverted, thanks!

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/SimpleInMemoryHistoricalBlockFacility.java
```diff
@@ -115,11 +118,40 @@ public void clearDisablePlugin() {
...
+     * @see #setTemporaryAvailableBlocks(SimpleBlockRangeSet) for details.
+     */
+    public void clearTemporaryAvailableBlocks() {
+        this.availableBlocksOverride = null;
+    }
+
     /**
      * {@inheritDoc}
+     *
+     * <p>
```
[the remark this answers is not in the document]
ata-nas: ✅ Great catch, this is indeed incomplete, I have added the rest.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -253,47 +285,62 @@ private void startMovingBatchOfBlocksToZipFile(final LongRange batchRange) {
...
+                    // Metrics updates
+                    // Update total bytes stored with the new zip file size
+                    totalBytesStored.addAndGet(zipFileSize);
+                    // Increment the blocks written counter
+                    blocksWrittenCounter.add(numberOfBlocksPerZipFile);
+                } catch (final IOException e) {
+                    final String message = "Failed to move batch of blocks [%d -> %d] to zip file"
+                            .formatted(batchFirstBlockNumber, batchLastBlockNumber);
+                    LOGGER.log(WARNING, message, e);
+                    return;
```
ata-nas, line 326: I've changed the level here to `WARNING`. It used to be `ERROR`. I thing warning is better and aligned with our document for logging.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/SimpleInMemoryHistoricalBlockFacility.java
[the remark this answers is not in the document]
jsync-swirlds: Forcing NPE in a test fixture is not a good way to verify your test is written correctly, in my opinion.
Ideally test fixtures allow unexpected usage patterns, they don't protect the caller.

It's mostly a matter of opinion, however.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/InMemoryBlockAccessor.java
```diff
@@ -0,0 +1,61 @@
...
+            throw new IllegalArgumentException("Block items list cannot be empty");
+        }
+        // Ensure the first item is a BlockHeader
+        if (!blockItems.getFirst().hasBlockHeader()) {
+            throw new IllegalArgumentException("First block item must be a BlockHeader");
+        }
+        // Ensure the last item is a BlockProof
+        if (!blockItems.getLast().hasBlockProof()) {
+            throw new IllegalArgumentException("Last block item must be a BlockProof");
+        }
```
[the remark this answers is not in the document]
ata-nas: ✅ This is added.

PR #1266, opened 2025-06-06
docs: Storage Retention Policy Design Documentation

### LINKED ISSUE #1263: Storage Retention Policy Design Documentation
### Story Form

As a Block Node Developer/User/Operator
I want to have a very detailed and well structured design documentation for the retention policy
So that I am very aware of how the functionality works

### Technical Notes

- let's create a design documentation of the Storage Retention Policy
  - consider also what metrics we need for it

ata-nas:
## Reviewer Notes

- design documentation for Retention Policy

## Related Issue(s)

Resolves #1263

### DIFF
--- a/docs/design/persistence/block-node-retention-policy.md
+++ b/docs/design/persistence/block-node-retention-policy.md
@@ -19,19 +19,20 @@ Effectively, this policy will allow us to achieve rolling history.
 ## Goals
 
 There are a few things that the Retention Policy aims to achieve:
+
 1. Each distinct persistence plugin is completely independent and implements its
-own Retention Policy, this includes:
-- Each plugin will be responsible for respecting its own Retention Policy.
-- Each plugin will choose its own level of granularity by which it will
-remove blocks that exceed the threshold.
+   own Retention Policy, this includes:
+   - Each plugin will be responsible for respecting its own Retention Policy.
+   - Each plugin will choose its own level of granularity by which it will
+     remove blocks that exceed the threshold.
 2. The Retention Policy should be configurable, allowing users to set the
-desired block count threshold.
+   desired block count threshold.
 3. The Retention Policy is **NOT** a hard limit, meaning that the Block-Node
-will keep accepting blocks even if the threshold is reached.
+   will keep accepting blocks even if the threshold is reached.
 4. The unit of the Retention Policy is block count.
 5. The Retention Policy is about limiting storage space - cleanup should happen
-when new data is received. There is no reason to clean data that is already
-there if no new data comes in.
+   when new data is received. There is no reason to clean data that is already
+   there if no new data comes in.
 
 ## Terms
 
@@ -52,25 +53,27 @@ listed in the [Goals section](#goals) above.
 ### `Files Recent Persistence`
 
 The `Files Recent Persistence` implements the Retention Policy:
+
 - A configurable parameter defines how many blocks to keep in storage.
 - The Policy is implemented by removing the oldest N amount of blocks (based on
-the configured threshold).
+  the configured threshold).
 - Blocks are removed only when new data is received.
 - The granularity of the removal is one whole block.
 - By default, the Retention Policy is set to `-1`, which means that there is no
-limit on the amount of blocks that can be stored.
+  limit on the amount of blocks that can be stored.
 
 ### `Files Historic Persistence`
 
 The `Files Historic Persistence` implements the Retention Policy:
+
 - A configurable parameter defines how many blocks to keep in storage.
 - The Policy is implemented by removing the oldest N amount of blocks (based on
-the configured threshold).
+  the configured threshold).
 - Blocks are removed only when new data is received.
 - The granularity of the removal is one whole zipped file, which contains one
-archived batch of blocks.
+  archived batch of blocks.
 - By default, the Retention Policy is set to `-1`, which means that there is no
-limit on the amount of blocks that can be stored.
+  limit on the amount of blocks that can be stored.
 
 ## Configuration
 
--- a/docs/design/persistence/block-node-retention-policy.md
+++ b/docs/design/persistence/block-node-retention-policy.md
@@ -59,8 +59,8 @@ The `Files Recent Persistence` implements the Retention Policy:
   the configured threshold).
 - Blocks are removed only when new data is received.
 - The granularity of the removal is one whole block.
-- By default, the Retention Policy is set to `-1`, which means that there is no
-  limit on the amount of blocks that can be stored.
+- By default, the Retention Policy is set to `96400`, which is a bit more than
+  a day when blocks are streamed at a rate of one per second.
 
 ### `Files Historic Persistence`
 
--- a/docs/design/persistence/block-node-retention-policy.md
+++ b/docs/design/persistence/block-node-retention-policy.md
@@ -1,0 +1,77 @@
+# Retention Policy Design Document
+
+## Table of Contents
+
+1. [Purpose](#purpose)
+2. [Goals](#goals)
+3. [Terms](#terms)
+4. [Design](#design)
+5. [Configuration](#configuration)
+6. [Metrics](#metrics)
+7. [Acceptance Tests](#acceptance-tests)
+
+## Purpose
+
+The Retention Policy is a threshold, measured in amount of blocks (block count),
+that defines how many blocks the Block-Node should keep in its storage.
+Effectively, this policy will allow us to achieve rolling history.
+
+## Goals
+
+There are a few things that the Retention Policy should achieve:
+1. Each Block-Node will give access to the Retention Policy as it is defined as
+a core concept of the Block-Node.
+2. The Retention Policy should be configurable, allowing users to set the
+desired block count threshold.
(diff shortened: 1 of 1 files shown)


jasperpotts (approved):
Looks great

docs/design/persistence/block-node-retention-policy.md
```diff
@@ -0,0 +1,77 @@
...
+desired block count threshold.
+3. The Retention Policy is **NOT** a hard limit, meaning that the Block-Node
+will keep accepting blocks even if the threshold is reached.
+4. Each Persistence Plugin is responsible for respecting the Retention
+Policy and removing blocks that exceed the threshold.
+5. Each Persistence Plugin will choose its own level of granularity
+for the Retention Policy, meaning that it can remove blocks in batches or
+one by one, or for example if the blocks are stored in a zip that contains
+'x' amount of blocks, it can delete the whole zip whenever the threshold
+passes at least one 'x' above. This really depends on implementation.
```
jsync-swirlds, line 34: I do not think it is wise to impose a retention policy on every persistence plugin.
This violates the composability of block nodes, and returns to the "configure everything" approach that created so much trouble in the older codebase.

We should not, in my opinion, be applying a policy to plugins (not least because we are then making assumptions about plugins we might not create).
There should be a different plugin for "rolling history" versus "full history", and that separate plugin should store blocks for a limited time.

Having a separate plugin isolates the (considerable) complexity of a rolling history and retention policies to the rolling history plugin, and avoids the complexity of long-term retention and archive in the rolling history plugin.

My view:
Recent Files Plugin - This is a rolling history plugin with (e.g.) 100,000 blocks retention.
Historic Files Plugin - This is _not_ a rolling history plugin, it simply retains data forever.
We may add other plugins; some with rolling history, some without.  It's important, however, to not add any requirement to a plugin, or set of plugins, globally.

docs/design/persistence/block-node-retention-policy.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Design
+
+The Retention Policy is a core concept of the Block-Node. It should be a very
+simple mechanism that will simply allow Persistence Plugins to be able to make
+decisions about what to clean up.
+
+Simply, the Retention Policy will be a configuration parameter that will be
+available to all Plugins in the Block-Node via the system-wide server
+configuration. Persistence Plugins should respect this policy.
```
jsync-swirlds, line 57: This is, in my opinion, very wrong.
Retention is, and should be, plugin-specific.  There is no reason, nor value, in making retention global.

docs/design/persistence/block-node-retention-policy.md
ata-nas: I see your point of view. In fact, initially I thought of something similar. I do see the point where the policy does not exactly fit everywhere. But after syncing with @jasperpotts, we talked about making this as simple as possible. The document is the result of our discussion. We should reach some consensus on this one.

In any case, implementation-wise changes are minimal no matter if we go either-or. Are we getting the config from a global object or local for the plugin config is all the same, the logic does not change after that.

docs/design/persistence/block-node-retention-policy.md
```diff
@@ -0,0 +1,77 @@
...
+
+The Retention Policy is a threshold, measured in amount of blocks (block count),
+that defines how many blocks the Block-Node should keep in its storage.
+Effectively, this policy will allow us to achieve rolling history.
+
+## Goals
+
+There are a few things that the Retention Policy should achieve:
+1. Each Block-Node will give access to the Retention Policy as it is defined as
+a core concept of the Block-Node.
```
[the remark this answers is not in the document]
jsync-swirlds: System-wide storage limit is _completely disjoint_ from retention policy.
Who knows how many different plugins are loaded and how those plugins use storage beyond just a simple block retention time/count.

I agree that each plugin that needs to pay any attention to retention should have its own logic.

If operators want to expose storage limits (which I'm not certain is needed); I believe even that is likely to be a per-plugin limit. Even in the planned first-release Tier 1 nodes we would have multiple storage limits for different plugins because we have multiple different storage pools (e.g. NVMe shared between recent and state management, JBOD for local archive, etc...).  Even within a single storage pool different plugins may very reasonably need different limits.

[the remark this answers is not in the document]
ata-nas: Hmm. It sounds like we need to address the system-wide storage as well. I wonder, how does that look like for us? Do we simply see failure due to reached storage limit or do we do something else (from the plugin's perspective and from the app's). Do we try to recover? Do we pre-check before persist? I feel like this is also something that we need to include in this design doc (at a high level). Suggestions?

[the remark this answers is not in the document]
jsync-swirlds: System level storage limits, in my experience, are best handled by the operator as a human process informed by system level metrics.
In most cases the "solution" to running low on storage is to physically add more, which is often handled by tracking total usage metrics and planning well in advance to add more storage (or migrate to new hardware) during a routine maintenance period.

I wouldn't concern ourselves with actual storage and limitations in this design, as what we intend to design is a tool for the operator to manage their system, not a mechanism to remove that responsibility from them.

docs/design/persistence/block-node-retention-policy.md
jasperpotts: I have a differing opinion. First it seems like we differ on if Historic Files Plugin is involved in rolling history. My thought is if you have a longer rolling history like 3mths, 1 year or 5 years which we have talked about for spheres. Then for those cases you do not want to say they can't use the Historic Files Plugin because they have a rolling history. 

Also I think there are a couple different things we are optimizing for:
- Joseph - minimizing global config and concepts
- Jasper - all storage is private inside a plugin. How and what data is stored by a persistence plugin is up to that plugin and no one else needs to know how it is stored.

jsync-swirlds: @jasperpotts and I had a chat.  Here's what we're thinking:
1. Plugins will be _completely_ independent.
1. Files.historic will no longer attempt to delete files when archived; it will _just_ archive the blocks according to its own configuration.  That configuration _may_ include a retention period, but the default should be "forever".
1. Files.recent will no longer support a "delete" operation.  It stores and retrieves blocks, and decides for itself when to delete.  Files.recent _must_ include a retention configuration, and will remove blocks based on that.  Examples include "10,000 blocks", "80,000 blocks", and perhaps "32 hours" (time based might be a future enhancement).
1. We will document _thoroughly_ that the retention period for `files.recent` _must_ exceed the archive size for `files.historic` in order for both to work together (if the operator deploys both).
1. The default configuration for `files.historic` _should_ include two values, "delay before archiving" (how far behind "latest" does it archive, default 100 blocks) and "amount to archive" (how many blocks to archive, default 1,000 or 10,000).
    * Example: With 100 and 10,000 a plugin would wait until the latest _archived_ block is 10,100 behind the "latest" block, and archive the oldest 10,000.  Thus a completed archive would typically be at most a couple hundred blocks behind "live", but there might be a bit over 10,000 blocks not-yet-archived at any given time.
    * In this example, the retention period for `files.recent` would be at least 12,000 (archive size + 20%) for reliability, but might be 86,000 so that ~1 day of blocks is always "rapidly" available.
1. There is nothing in the design that requires either or both plugins.  The `files.historic` plugin will work equally well with any block provider plugin that can return previously unarchived blocks.  The `files.recent` plugin sees `files.historic` as merely another client and does not know, or care, what it does with blocks.

AlfredoG87: I like this approach 👍 💯

ata-nas: @jsync-swirlds based on what you have said, should I still finish this document here or should we do something completely different (the mentioned operations documentation)? I feel like the retention policy as we have defined it here should simply move and be part of the design documentation of each persistence plugin with specifics there.

ata-nas: I've made changes to this document. Thanks for the feedback! Please take a look again so we can clean it up.

docs/design/persistence/block-node-retention-policy.md
```diff
@@ -0,0 +1,96 @@
...
+
+The `Files Recent Persistence` implements the Retention Policy:
+
+- A configurable parameter defines how many blocks to keep in storage.
+- The Policy is implemented by removing the oldest N amount of blocks (based on
+  the configured threshold).
+- Blocks are removed only when new data is received.
+- The granularity of the removal is one whole block.
+- By default, the Retention Policy is set to `-1`, which means that there is no
+  limit on the amount of blocks that can be stored.
```
jsync-swirlds, line 63: Default should be 96400 blocks.  The recent plugin should default to "not forever" and a bit over a day is a reasonable default.

jsync-swirlds (approved):
Just one minor suggestion; non-blocking.

docs/design/persistence/block-node-retention-policy.md
ata-nas: ✅ I've added this.

jsync-swirlds (approved):

PR #1268, opened 2025-06-06
feat: Storage Retention Policy Implementation

### LINKED ISSUE #1267: Storage Retention Policy Implementation
### Story Form

As a Block Node User/Operator
I want to have a Storage Retention Policy
So that I will be able to configure it and use it

### Technical Notes

- implement the Storage Retention Policy based on the #1263

ata-nas:
## Reviewer Notes

- implementing the retention policy
- persistence plugins are extended to support it

## Related Issue(s)

Resolves #1267

### DIFF
--- a/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
+++ b/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
@@ -44,6 +44,8 @@ public final class BlocksFilesHistoricPlugin implements BlockProviderPlugin, Blo
     private final CopyOnWriteArrayList<LongRange> inProgressZipRanges = new CopyOnWriteArrayList<>();
     /** Running total of bytes stored in the historic tier */
     private final AtomicLong totalBytesStored = new AtomicLong(0);
+    /** The Storage Retention Policy Threshold */
+    private long retentionPolicyThreshold;
 
     // Metrics
     /** Counter for blocks written to the historic tier */
@@ -73,10 +75,9 @@ public final class BlocksFilesHistoricPlugin implements BlockProviderPlugin, Blo
     public void init(final BlockNodeContext context, final ServiceBuilder serviceBuilder) {
         this.context = Objects.requireNonNull(context);
         final FilesHistoricConfig config = context.configuration().getConfigData(FilesHistoricConfig.class);
-
+        this.retentionPolicyThreshold = context.storageRetentionPolicyThreshold();
         // Initialize metrics
         initMetrics(context.metrics());
-
         // create plugin data root directory if it does not exist
         try {
             Files.createDirectories(config.rootPath());
--- a/block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
+++ b/block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
@@ -73,7 +73,8 @@ public final class BlocksFilesRecentPlugin implements BlockProviderPlugin, Block
     private final ConcurrentLongRangeSet availableBlocks = new ConcurrentLongRangeSet();
     /** Running total of bytes stored in the recent tier */
     private final AtomicLong totalBytesStored = new AtomicLong(0);
-
+    /** The Storage Retention Policy Threshold */
+    private long retentionPolicyThreshold;
     // Metrics
     /** Counter for blocks written to the recent tier */
     private Counter blocksWrittenCounter;
@@ -120,6 +121,7 @@ public final class BlocksFilesRecentPlugin implements BlockProviderPlugin, Block
         if (this.config == null) {
             this.config = context.configuration().getConfigData(FilesRecentConfig.class);
         }
+        this.retentionPolicyThreshold = context.storageRetentionPolicyThreshold();
         this.blockMessaging = context.blockMessaging();
         // Initialize metrics
         initMetrics(context.metrics());
--- a/block-node/spi/src/main/java/org/hiero/block/node/spi/BlockNodeContext.java
+++ b/block-node/spi/src/main/java/org/hiero/block/node/spi/BlockNodeContext.java
@@ -29,6 +29,9 @@ import org.hiero.block.node.spi.threading.ThreadPoolManager;
  * @param blockMessaging the block messaging service of the block node
  * @param historicalBlockProvider the historical block provider of the block node
  * @param serviceLoader the service loader function to use to load services
+ * @param threadPoolManager the thread pool manager for the block node
+ * @param storageRetentionPolicyThreshold the threshold for the storage retention policy,
+ * -1 defaults to no retention policy (unlimited storage)
  */
 public record BlockNodeContext(
         Configuration configuration,
@@ -37,4 +40,5 @@ public record BlockNodeContext(
         BlockMessagingFacility blockMessaging,
         HistoricalBlockFacility historicalBlockProvider,
         ServiceLoaderFunction serviceLoader,
-        ThreadPoolManager threadPoolManager) {}
+        ThreadPoolManager threadPoolManager,
+        long storageRetentionPolicyThreshold) {}
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -158,7 +158,8 @@ public class BlockNodeApp implements HealthFacility {
                 blockMessagingService,
                 historicalBlockFacility,
                 serviceLoader,
-                threadPoolManager);
+                threadPoolManager,
+                serverConfig.storageRetentionPolicyThreshold());
         // ==== CREATE ROUTING BUILDERS ================================================================================
         // Create HTTP & GRPC routing builders
         final ServiceBuilderImpl serviceBuilder = new ServiceBuilderImpl();
--- a/block-node/app/src/main/java/org/hiero/block/node/app/ServerConfig.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/ServerConfig.java
@@ -24,4 +24,5 @@ public record ServerConfig(
         @Loggable @ConfigProperty(defaultValue = "32768") @Min(32768) @Max(Integer.MAX_VALUE)
                 int socketReceiveBufferSizeBytes,
         @Loggable @ConfigProperty(defaultValue = "8080") @Min(1024) @Max(65_535) int port,
-        @Loggable @ConfigProperty(defaultValue = "500") int shutdownDelayMillis) {}
+        @Loggable @ConfigProperty(defaultValue = "500") int shutdownDelayMillis,
+        @Loggable @ConfigProperty(defaultValue = "-1") @Min(-1) long storageRetentionPolicyThreshold) {}
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/GrpcPluginTestBase.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/GrpcPluginTestBase.java
@@ -50,17 +50,19 @@ public abstract class GrpcPluginTestBase<P extends BlockNodePlugin> extends Plug
      * @param plugin the plugin to be tested
      * @param historicalBlockFacility the historical block facility to be used
      */
-    public void start(P plugin, Method method, HistoricalBlockFacility historicalBlockFacility) {
-        start(plugin, method, historicalBlockFacility, null, null);
+    public void start(final P plugin, final Method method, final HistoricalBlockFacility historicalBlockFacility) {
+        start(plugin, method, historicalBlockFacility, null, null, -1L);
     }
 
     public void start(
-            P plugin,
-            Method method,
-            HistoricalBlockFacility historicalBlockFacility,
-            ThreadPoolManager testThreadManager,
-            Map<String, String> configOverrides) {
-        super.start(plugin, historicalBlockFacility, testThreadManager, configOverrides);
+            final P plugin,
+            final Method method,
+            final HistoricalBlockFacility historicalBlockFacility,
+            final ThreadPoolManager testThreadManager,
+            final Map<String, String> configOverrides,
+            final long storageRetentionPolicyThreshold) {
+        super.start(
+                plugin, historicalBlockFacility, testThreadManager, configOverrides, storageRetentionPolicyThreshold);
         // setup to receive bytes from the plugin
         fromPluginPipe = new Pipeline<>() {
             @Override
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
@@ -64,8 +64,8 @@ public abstract class PluginTestBase<P extends BlockNodePlugin> {
      * @param plugin the plugin to be tested
      * @param historicalBlockFacility the historical block facility to be used
      */
-    public void start(P plugin, HistoricalBlockFacility historicalBlockFacility) {
-        start(plugin, historicalBlockFacility, testThreadPoolManager, null);
+    public void start(final P plugin, final HistoricalBlockFacility historicalBlockFacility) {
+        start(plugin, historicalBlockFacility, testThreadPoolManager, null, -1L);
     }
 
     /**
@@ -77,8 +77,11 @@ public abstract class PluginTestBase<P extends BlockNodePlugin> {
      * @param historicalBlockFacility the historical block facility to be used
      * @param testThreadManager the thread pool manager to be used
      */
-    public void start(P plugin, HistoricalBlockFacility historicalBlockFacility, ThreadPoolManager testThreadManager) {
-        start(plugin, historicalBlockFacility, testThreadManager, null);
+    public void start(
+            final P plugin,
+            final HistoricalBlockFacility historicalBlockFacility,
+            final ThreadPoolManager testThreadManager) {
+        start(plugin, historicalBlockFacility, testThreadManager, null, -1L);
     }
 
     /**
@@ -90,8 +93,11 @@ public abstract class PluginTestBase<P extends BlockNodePlugin> {
      * @param historicalBlockFacility the historical block facility to be used
      * @param configOverrides a map of configuration overrides to be applied to loaded configuration
      */
-    public void start(P plugin, HistoricalBlockFacility historicalBlockFacility, Map<String, String> configOverrides) {
-        start(plugin, historicalBlockFacility, testThreadPoolManager, configOverrides);
+    public void start(
+            final P plugin,
+            final HistoricalBlockFacility historicalBlockFacility,
+            final Map<String, String> configOverrides) {
+        start(plugin, historicalBlockFacility, testThreadPoolManager, configOverrides, -1L);
     }
 
     /**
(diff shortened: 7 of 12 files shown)


block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
```diff
@@ -251,14 +254,9 @@ public void handleVerification(VerificationNotification notification) {
...
-            availableBlocks.remove(notification.startBlockNumber(), notification.endBlockNumber());
-            // delete all files in range
-            for (long blockNumber = notification.startBlockNumber();
-                    blockNumber <= notification.endBlockNumber();
-                    blockNumber++) {
-                delete(blockNumber);
-            }
+            final long totalStored = availableBlocks.size();
+            long excess = totalStored - retentionPolicyThreshold;
+            availableBlocks.stream().sorted().limit(excess).forEach(this::delete);
```
ata-nas, line 259: Updated deletion logic. Not sure if we still would want to filter by notification provider priority. In theory the new retention logic no longer supports that. We no longer delete the range of the notification. More than that, if we have only this plugin that will fire such notifications, effectively we will never cleanup. I think that should also be changed. Am I missing something?

block-node/spi/src/main/java/org/hiero/block/node/spi/BlockNodeContext.java
```diff
@@ -37,4 +40,5 @@ public record BlockNodeContext(
         BlockMessagingFacility blockMessaging,
         HistoricalBlockFacility historicalBlockProvider,
         ServiceLoaderFunction serviceLoader,
-        ThreadPoolManager threadPoolManager) {}
+        ThreadPoolManager threadPoolManager,
+        long storageRetentionPolicyThreshold) {}
```
ata-nas, line 44: Chose to add this field here. Reason is that the `ServerConfig` is local to the `app` module and in order for plugins to see it, we need to have it as `requires` (not sure but I think that this will be a circular dep, haven't tried). Since the retention policy will be a "core" feature, I think it is ok to have it in the context. Other option was to move the server config into something like the `spi` module, but that seemed like an even worse idea.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
```diff
@@ -338,21 +337,28 @@ private void delete(long blockNumber) {
...
+        } catch (final IOException e) {
             LOGGER.log(WARNING, "Failed to delete block file: " + blockFilePath, e);
-            throw new RuntimeException(e);
+            throw new UncheckedIOException(e);
+            // @todo(1268) if we fail here, should we really throw or just log and continue?
+            //   a retry will essentially happen on the next persisted notification.
+            //   With the current implementation, this method remains problematic because
+            //   other IO happens asynchronously in the background (new persists) which can lead
+            //   to exceptions during the delete of the parent(s). Do we have a better approach to
+            //   clean up directories?
```
ata-nas, line 361: Ideas? This method has some race conditions naturally. What should we do here? Catch all and do nothing, hoping it will be resolved on the next run of the retention policy cleanup (next persisted notification) or something else? not sure if also a delete of the block file itself will throw while s/o is reading, or that's handled for us by default by the os or the java nio.

ata-nas (commented):
Just some initial stuff for the POC. Left some considerations that came up while implementing this.

block-node/spi/src/main/java/org/hiero/block/node/spi/BlockNodeContext.java
ata-nas: This has been made redundant, we updated the implementation.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
ata-nas: Resolved with implementation changes.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
```diff
@@ -251,15 +254,20 @@ public void handleVerification(VerificationNotification notification) {
...
+        // @todo(1268) Not sure if we still would want to filter by notification provider priority.
+        //    In theory the new retention logic no longer supports that. We no longer delete the
+        //    range of the notification. More than that, if we have only this plugin that will fire
+        //    such notifications, effectively we will never cleanup. I think that should also be
+        //    changed. Another way to do this is to delete in the verification notification method.
+        //    In theory from there we get the new data and the idea of the policy is to be bound to
+        //    block count and delete only if new data arrives. Maybe we could drop the handle
+        //    persisted here all together? But we have to be careful because a big delete might
+        //    take a long while and we do not want to hang the notification thread too long cause
+        //    persistence relies on it (verification notification method). Am I missing something?
```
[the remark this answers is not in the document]
ata-nas: Discussed with team, the right place to do retention is after data is successfully received.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
```diff
@@ -314,45 +307,46 @@ private void writeBlockToLivePath(final BlockUnparsed block, final long blockNum
...
+                LOGGER.log(WARNING, "Failed to delete block file: " + blockFilePath, e);
+                // @todo(1268) do not throw, increment a metric and log info
             }
-
             // clean up any empty parent directories up to the base directory
             Path parentDir = blockFilePath.getParent();
             while (parentDir != null && !parentDir.equals(config.liveRootPath())) {
-                try (var filesList = Files.list(parentDir)) {
+                try (final Stream<Path> filesList = Files.list(parentDir)) {
                     if (filesList.findAny().isPresent()) {
```
jsync-swirlds, line 337: Might be better to use `findFirst` here...

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
```diff
@@ -314,45 +307,46 @@ private void writeBlockToLivePath(final BlockUnparsed block, final long blockNum
...
+                    } else {
+                        // we did not find any files in the directory, so delete it
+                        Files.deleteIfExists(parentDir);
+                        // move up to the parent directory
+                        parentDir = parentDir.getParent();
                     }
-                } catch (IOException e) {
-                    LOGGER.log(WARNING, "Failed to list files in directory: " + parentDir, e);
+                } catch (final IOException e) {
+                    LOGGER.log(WARNING, "Failed remove parent directory: " + parentDir, e);
```
jsync-swirlds, line 346: This probably should be `INFO`.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/FilesRecentConfig.java
```diff
@@ -14,21 +14,24 @@
  *
  * @param liveRootPath provides the root path for saving blocks live
  * @param compression compression type to use for the storage. It is assumed this never changes while a node is running
- *                    and has existing files.
+ * and has existing files.
  * @param maxFilesPerDir number of files per directory. This is used to limit the number of files in a directory to avoid
- *                    file system issues.
+ * file system issues.
+ * @param retentionPolicyThreshold the retention policy threshold (count of blocks to keep). Must be positive.
```
jsync-swirlds, line 20: positive implies strictly `> 0`.  Do we allow `0` (perhaps to mean "infinite")?

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/FilesRecentConfig.java
```diff
@@ -14,21 +14,24 @@
...
  */
 @ConfigData("files.recent")
 public record FilesRecentConfig(
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/data/live") Path liveRootPath,
         @Loggable @ConfigProperty(defaultValue = "ZSTD") CompressionType compression,
-        @Loggable @ConfigProperty(defaultValue = "3") int maxFilesPerDir) {
+        @Loggable @ConfigProperty(defaultValue = "3") int maxFilesPerDir,
+        @Loggable @ConfigProperty(defaultValue = "96_000") long retentionPolicyThreshold) {
```
jsync-swirlds, line 27: nitpick:
`blockRetentionThreshold` reads better.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
```diff
@@ -120,6 +124,7 @@ public void init(final BlockNodeContext context, final ServiceBuilder serviceBui
         if (this.config == null) {
             this.config = context.configuration().getConfigData(FilesRecentConfig.class);
         }
+        this.retentionPolicyThreshold = this.config.retentionPolicyThreshold();
```
jsync-swirlds, line 127: The prefix `this` is not needed here.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
```diff
@@ -236,29 +241,17 @@ public BlockRangeSet availableBlocks() {
...
+        // after writing the block, we need to trigger the retention policy
+        // calculate excess
+        final long excess = availableBlocks.size() - retentionPolicyThreshold;
+        final long lowestBlockNumber = availableBlocks.min();
+        // determine how many blocks to delete, up to the retention round limit
+        final long blocksToDelete = Math.min(excess, RETENTION_ROUND_LIMIT);
+        // delete the blocks from the lowest block number up to calculated max
+        // gaps will be retried on subsequent retention runs, which are very
+        // frequent
+        for (long i = lowestBlockNumber; i < lowestBlockNumber + blocksToDelete; i++) {
```
jsync-swirlds, line 253: Instead of calculating the endpoint every iteration, the endpoint could be calculated before entering the loop.

jsync-swirlds (commented):
A few suggestions.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -206,6 +211,11 @@ public void handlePersisted(PersistedNotification notification) {
         // calculations and decide if we will submit a task or not
         if (notification.blockProviderPriority() > defaultPriority()) {
             attemptZipping();
+            // @todo(1268) like in the recents plugin, should we cleanup only if the priority check passes?
+            //   here I think yes, because we only get new data if the check passes, so there is no issue
+            //   to keep within the bounds of the retention policy threshold. But for the recents plugin,
+            //   maybe we need to think about it since it gets new data via the verification notification.
```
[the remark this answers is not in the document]
ata-nas: This has been removed, you are looking at an older commit.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
```diff
@@ -16,23 +16,27 @@
...
+ * @param retentionPolicyThreshold the retention policy threshold (count of blocks to keep). If set to -1, no retention
+ * policy is applied.
  */
 @ConfigData("files.historic")
 public record FilesHistoricConfig(
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/data/historic") Path rootPath,
         @Loggable @ConfigProperty(defaultValue = "ZSTD") CompressionType compression,
-        @Loggable @ConfigProperty(defaultValue = "4") @Min(1) @Max(6) int powersOfTenPerZipFileContents) {
+        @Loggable @ConfigProperty(defaultValue = "4") @Min(1) @Max(6) int powersOfTenPerZipFileContents,
+        @Loggable @ConfigProperty(defaultValue = "-1") long retentionPolicyThreshold) {
```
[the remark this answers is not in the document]
ata-nas: I'd say that is reasonable. I will make the change. Mind you, the operator is responsible of setting this properly however. The granularity here is we can delete a whole zip at the minimum, so if the zip batch size is 1000 blocks per zip, it makes no sense to set the retention policy to 100 blocks, as this might interfere with calculations.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
ata-nas: Interesting, I thought the compiler would optimize for this one (I remember we've seen similar calculations in the past where we discussed that was the case). I'll extract it.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
```diff
@@ -314,45 +307,46 @@ private void writeBlockToLivePath(final BlockUnparsed block, final long blockNum
...
-                    blocksDeletedCounter.increment();
-                    totalBytesStored.addAndGet(-fileSize);
-                }
+                Files.delete(blockFilePath);
+                LOGGER.log(DEBUG, "Successfully deleted block file: " + blockFilePath);
+                availableBlocks.remove(blockNumber);
+                blocksDeletedCounter.increment();
+                totalBytesStored.addAndGet(-fileSize);
+            } catch (final IOException e) {
+                LOGGER.log(WARNING, "Failed to delete block file: " + blockFilePath, e);
```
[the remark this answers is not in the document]
ata-nas: I have the todo already, needed some assistance from @AlfredoG87 for the metrics, this is why you do not see it yet.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -255,6 +265,53 @@ private void attemptZipping() {
...
+        // we only take action if the threshold is greater than -1
+        if (retentionPolicyThreshold > -1) {
+            final long totalStored = availableBlocks.size();
+            long excess = totalStored - retentionPolicyThreshold;
+            // the numberOfBlocksPerZipFile should generally be immutable once set
+            // for the first time when the block node was originally started
+            // we can rely on the check below to ensure we are deleting the correct
+            // number of blocks
+            while (excess >= numberOfBlocksPerZipFile) {
+                // if we have passed the above check, we can delete at least one zip file
```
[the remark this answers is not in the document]
ata-nas: This is kind of a design decision. It changes the retention policy semantically. I am pro for this change, but if we do so, we need to fundamentally change the design docs and the logic as well.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
jsync-swirlds: findFirst is the same if the stream is not parallel.  findAny only performs better (potentially) on a parallel stream.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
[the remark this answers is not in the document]
jsync-swirlds: That's why I suggested a count of zip files elsewhere as an alternative.
Hoping the operator will set both values "correctly" when both refer to block counts is less than ideal.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -255,6 +265,53 @@ private void attemptZipping() {
...
+                // type.
+                final Path zipToDelete =
+                        BlockPath.computeBlockPath(config, minBlockNumberStored).zipFilePath();
+                if (Files.exists(zipToDelete)) {
+                    try {
+                        // since we keep track of the whole zip file size, that is
+                        // what we should decrement the total bytes stored by
+                        final long zipFileSize = Files.size(zipToDelete);
+                        Files.delete(zipToDelete);
+                        totalBytesStored.addAndGet(-zipFileSize);
```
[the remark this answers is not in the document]
ata-nas: I see what you mean. I guess we can revisit the metrics in another PR.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
ata-nas: Cool! I will keep that in mind. Thanks @jsync-swirlds, I always learn something new 🙌

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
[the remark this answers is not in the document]
ata-nas: Not entirely sure that this is an implementation detail per se. I see granularity as a design detail for each plugin.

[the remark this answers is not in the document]
jsync-swirlds: Perhaps, I think my point was that it's specific to the plugin, rather than generic to all plugins (which is what this document purports to be).

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlocksFilesHistoricPlugin.java
```diff
@@ -78,11 +84,10 @@ public List<Class<? extends Record>> configDataTypes() {
     @Override
     public void init(final BlockNodeContext context, final ServiceBuilder serviceBuilder) {
         this.context = Objects.requireNonNull(context);
-        final FilesHistoricConfig config = context.configuration().getConfigData(FilesHistoricConfig.class);
-
+        this.config = context.configuration().getConfigData(FilesHistoricConfig.class);
```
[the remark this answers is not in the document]
ata-nas: Updated.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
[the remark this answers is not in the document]
ata-nas: Updated, metrics are now added and we log info.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/FilesRecentConfig.java
ata-nas: Updated in both plugins.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlocksFilesRecentPlugin.java
ata-nas: Updated.

AlfredoG87 (approved):
Nice! 💯

PR #1271, opened 2025-06-07
chore: #1159 Fix incorrect timeout for publisher

### LINKED ISSUE #1159: Incorrect timeout for publisher
### Description

The default timeout for the publisher plugin at [`PublisherConfig::timeoutThresholdMillis`](https://github.com/hiero-ledger/hiero-block-node/blob/v0.10.0-rc2/block-node/publisher/src/main/java/org/hiero/block/node/publisher/PublisherConfig.java#L20) is incorrect (too short). It should be at least 4 seconds to allow for consensus node block time defaults, real-world delays, and network slowness.
Ideally it would be long enough to tolerate 2 additional block-times of delay (i.e. 8 seconds total), in case the publishers need to construct a block proof from signatures on a later block.

### Steps to Reproduce

1. Run a block node
2. Run a consensus node as publisher
3. observe this error in logs
    * `2025-05-16 18:41:57.837+0000 WARNING PublisherServicePlugin#onSessionUpdate       currentPrimaryHasTimedOut, primarySession={1}`

### Additional context

_No response_

jsync-swirlds:
## Reviewer Notes
* Increased the default timeout to account for block proof delays and block interleave in a multi-publisher scenario.

## Related Issue(s)
 Resolves #1159.

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
@@ -17,7 +17,7 @@ import org.hiero.block.node.base.Loggable;
 @ConfigData("producer")
 public record PublisherConfig(
         @Loggable @ConfigProperty(defaultValue = "PRODUCTION") PublisherType type,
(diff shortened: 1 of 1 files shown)


ata-nas (approved):
LG

PR #1288, opened 2025-06-10
Design: Add Design document for Backfill Plugin

### LINKED ISSUE #1278: Backfill Plugin Design Doc Creation, Review and Approval.
Based on design meeting and discussion with the team, use notes and recording to propose a Backfill design document.

AlfredoG87:
## Reviewer Notes

Based on discussion last Thursday over the design meeting.
Adds a Document that captures the mechanism and high level logic for the backfill plugin functionality.

Fixes #1278

### DIFF
--- a/docs/design/backfill-plugin.md
+++ b/docs/design/backfill-plugin.md
@@ -1,0 +1,156 @@
+# Backfill Plugin Design Document
+
+## Table of Contents
+
+1. [Purpose](#purpose)
+2. [Goals](#goals)
+3. [Terms](#terms)
+4. [Entities](#entities)
+5. [Design](#design)
+6. [Diagram](#diagram)
+7. [Configuration](#configuration)
+8. [Metrics](#metrics)
+9. [Exceptions](#exceptions)
+10. [Acceptance Tests](#acceptance-tests)
+
+## Purpose
+
+This plugin purpose is to detect missing gaps in the intended stored block sequence, then autonomously and asynchronously fetch the missing blocks from another source and store them.
+
+## Goals
+
+1. Detect gaps on start-up an while running.
+2. Fetch missing blocks from another BN, most be configurable.
+3. Asynchronously recovery (fetch and store) the missing blocks, without blocking the live ingestion path (new coming blocks).
+4. Instrumentation, logging, metrics and error handling for the backfill process.
+
+## Terms
+
+<dl>
+  <dt>Missing Gap</dt>
+  <dd>Its a contiguous range of missing blocks, could be a single one.</dd>
+  <dt>Backfill</dt>
+  <dd>The process of fetching and storing missing blocks in the local storage.</dd>
+    <dt>Grpc Client</dt>
+    <dd> A client that connects to another Block Node to fetch missing blocks.</dd>
+
+</dl>
+
+## Entities
+
+### BackfillPlugin
+
+- The main plugin that manages the backfill process.
+- Orchestrates runs, gap detection, fetching, and message processing.
+- Handles configuration and initialization.
+- Implements the `Block Node Plugin` interface.
+
+### MessagingFacility
+
+Block Node's messaging system used for communication between plugins and components.
+
+### HistoricalBlockFacility
+
+A facility that provides access to historical blocks, this is available for all plugins.
+
+## Design
+
+1. At start-up a loop is defined that runs every `backfill.scanInterval`
+2. At every interval the plugin detects missing gaps in the intended block range against the actual stored blocks using the `HistoricalBlockFacility`.
+3. If gaps are found, it initiates the backfill process.
+4. The plugin uses a gRPC client to connect to another Block Node (configured via `backfill_sources`) to fetch the missing blocks.
+5. Once the blocks are fetched, the plugin creates a `BlockNotification` of type `BackfilledBlockNotification` and sends it to the `MessagingFacility`.
+6. The `VerificationPlugin` will then process the `BackfilledBlockNotification` and if the block is valid, it will create a `VerificationNotification` and send it to the `MessagingFacility` for further processing.
+7. The PersistencePlugin will then store the block in the local storage.
+8. The `BackfillPlugin` will receive the `PersistenceNotification` and update its internal state accordingly, marking the backfill process as complete for that block.
+
+## Diagram
+
+```mermaid
+sequenceDiagram
+    autonumber
+    participant BackfillPlugin
+    participant HistoricalBlockFacility
+    participant GrpcClient
+    participant MessagingFacility
+    participant VerificationPlugin
+    participant PersistencePlugin
+
+    loop Every backfill.scanInterval
+        BackfillPlugin->>HistoricalBlockFacility: detectMissingGaps(firstBlock, lastBlock)
+        alt Gaps found
+            BackfillPlugin->>+GrpcClient: fetchMissingBlocks(gapRange, batchSize)
+            GrpcClient-->>-BackfillPlugin: blocks[]
+            loop For each block
+                BackfillPlugin->>MessagingFacility: BackfilledBlockNotification(block)
+            end
+        else No gaps
+            Note right of BackfillPlugin: nothing to backfill
+        end
+    end
+
+%% Verification path
+    MessagingFacility->>VerificationPlugin: BackfilledBlockNotification(block)
+    activate VerificationPlugin
+    VerificationPlugin-->>MessagingFacility: VerificationNotification(block, valid)
+    deactivate VerificationPlugin
+
+%% Persistence path
+    MessagingFacility->>PersistencePlugin: VerificationNotification(block, valid)
+    activate PersistencePlugin
+    PersistencePlugin-->>MessagingFacility: PersistenceNotification(block)
+    deactivate PersistencePlugin
+
+%% Backfill completion
+    MessagingFacility->>BackfillPlugin: PersistenceNotification(block)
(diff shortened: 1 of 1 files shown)


docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+## Terms
+
+<dl>
+  <dt>Missing Gap</dt>
+  <dd>Its a contiguous range of missing blocks, could be a single one.</dd>
+  <dt>Backfill</dt>
+  <dd>The process of fetching and storing missing blocks in the local storage.</dd>
+    <dt>Grpc Client</dt>
+    <dd> A client that connects to another Block Node to fetch missing blocks.</dd>
+
```
jsync-swirlds, line 36: Small spacing error here.

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+        BackfillPlugin-->>MessagingFacility: no gaps → no action
+    end
+
+```
+
+### gRPC Client
+
+The gRPC client is used to connect to another Block Node to fetch the missing blocks. It will be configured with the `backfill_sources` parameter, which is a list of HOST:PORT pairs of the Block Nodes to connect to.
+- Fetching blocks will be done in batches, the size of which can be configured with the `backfill.fetchBatchSize` parameter.
+- For each BN configured, the client will perform an `BlockNodeService/serverStatus` call to check if the missing gap is available in the remote BN, if it is not available, it will skip that BN and continue with the next one.
```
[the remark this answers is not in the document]
AlfredoG87: I disagree, I believe that a query to `server/status` is really cheap and fast, and the backfill plugin knows if it has a partial range available, a full range available or none.

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+- If the remote BN is not available, it will log an error and continue with the next one.
```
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+9. [Exceptions](#exceptions)
+10. [Acceptance Tests](#acceptance-tests)
+
+## Purpose
+
+This plugin purpose is to detect missing gaps in the intended stored block sequence, then autonomously and asynchronously fetch the missing blocks from another source and store them.
+
+## Goals
+
+1. Detect gaps on start-up an while running.
```
ata-nas, line 22:
```suggestion
1. Detect gaps on start-up and while running.
```

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+2. Fetch missing blocks from another BN, most be configurable.
```
ata-nas, line 23: Maybe a typo here?
```suggestion
2. Fetch missing blocks from another BN, must be configurable.
```

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+3. Asynchronously recovery (fetch and store) the missing blocks, without blocking the live ingestion path (new coming blocks).
```
ata-nas, line 24:
```suggestion
3. Asynchronously recover (fetch and store) the missing blocks, without blocking the live ingestion path (new incoming blocks).
```

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+4. Instrumentation, logging, metrics and error handling for the backfill process.
+
+## Terms
+
+<dl>
+  <dt>Missing Gap</dt>
+  <dd>Its a contiguous range of missing blocks, could be a single one.</dd>
+  <dt>Backfill</dt>
+  <dd>The process of fetching and storing missing blocks in the local storage.</dd>
```
ata-nas, line 33: Do we go through verification after fetching?

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+    H -->|Yes| E3[Log No configured BN had missing blocks]
+    E3 --> I[Exit, retry after scanInterval]
+
+```
+
+## Configuration
+
+|     Configuration Property     |                                     Description                                      | Default |
+|--------------------------------|--------------------------------------------------------------------------------------|---------|
+| `backfill.firstBlockAvailable` | The first block that this BN deploy wants to have available                          | 0       |
```
ata-nas, line 185: firstBlockAvailable == 0 default is probably not a good idea. As a default maybe we should have -1 to mean whatever is the first block you receive via the regular stream (or if you already have something persisted, the lowest number you have is the one you want to have lowest available), that would be the first one you want to have. Zero is a valid block number and would essentially mean full history.

ata-nas (commented):
@AlfredoG87 looks good, diagrams are 🔥, I leave some cleanup suggestions.

docs/design/backfill-plugin.md
AlfredoG87: Yes, we cannot Persist a Block that has not gonne trought Verification

docs/design/backfill-plugin.md
AlfredoG87: Let me think more about this, thanks for the idea.

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+        BackfillPlugin->>MessagingFacility: BackfilledBlockNotification(blocks)
+    else No gaps
+        BackfillPlugin-->>MessagingFacility: no gaps → no action
+    end
+
+```
+
+### gRPC Client
+
+The gRPC client is used to connect to another Block Node to fetch the missing blocks. It will be configured with the `backfill_sources` parameter, which is a list of HOST:PORT pairs of the Block Nodes to connect to.
```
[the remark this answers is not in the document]
jsync-swirlds: An external file (pointed to by config) is a good idea.
My primary point is just that host:port pairs are probably not sufficient.

docs/design/backfill-plugin.md
jsync-swirlds: I would advocate for a default of `0`.  The default should be "full history" and any node not wanting that can set a different value.
It's probably lower risk to make the most critical systems (Tier 1 full history) the defaults.

AlfredoG87: I agree with @jsync-swirlds in this one, "The first block received" is too ambiguous and prone to indeterministic behaviour and tests

docs/design/backfill-plugin.md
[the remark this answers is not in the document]
AlfredoG87: My original idea was to set priority by Order of appearance on the list, and since we have some in-flux design for BN Address Book, I would go with this simpler approach atm, to simplify and expedite the feature to live, and once we have more certainty of the BN Address Book inner working and a more final structure for the BN json file. we can change it down the road, is always easier to add complexity than to remove it.

docs/design/backfill-plugin.md
AlfredoG87: ✅  Done!

docs/design/backfill-plugin.md
AlfredoG87: ✅  Done!

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Terms
+
+<dl>
+  <dt>Missing Gap</dt>
+  <dd>Its a contiguous range of missing blocks, could be a single one.</dd>
+  <dt>Backfill</dt>
+  <dd>The process of fetching and storing missing blocks in the local storage.</dd>
+    <dt>Grpc Client</dt>
+    <dd>A client that connects to another Block Node to fetch missing blocks.</dd>
```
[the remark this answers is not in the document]
AlfredoG87: ✅ Done! now it should be correct 👍

jsync-swirlds (dismissed):

ata-nas (dismissed):
Great job @AlfredoG87! 🔥

docs/design/backfill-plugin.md
```diff
@@ -0,0 +1,258 @@
...
+  <dt>Missing Gap</dt>
+  <dd>Its a contiguous range of missing blocks, could be a single one.</dd>
+  <dt>Backfill</dt>
+  <dd>The process of fetching and storing missing blocks in the local storage.</dd>
+  <dt>Grpc Client</dt>
+  <dd>A client that connects to another Block Node to fetch missing blocks.</dd>
+  <dt>BackfilledBlockNotification</dt>
+  <dd>A new Notification Type that can be published to the Messaging Facility, it is intended to contain a whole block that was fetched from another source and is being backfilled into the system. </dd>
+  <dt>NewestBlockKnownToNetwork</dt>
+  <dd>Commonly sent by the Publisher Plugin to indicate that the BlockNode is behind and that needs to get up-to-date, commonly picked handled by Backfill plugin and filling the gap immediately.</dd>
```
jsync-swirlds, line 39: Wording suggestion:
```suggestion
  <dd>Notification sent by a plugin, including a "publisher" plugin, to indicate that the BlockNode is behind and must be brought up-to-date. This is often handled by a "backfill" plugin by immediately requesting the missing blocks.</dd>
```

jsync-swirlds (dismissed):
Looks good.
One small wording update, but that can be added later.

ata-nas (approved):

ata-nas (approved):

PR #1289, opened 2025-06-10
chore: Improving secret handling for BN chart deployments

### LINKED ISSUE #1005: Improve Secret Handling of Helm Chart to better match RE Workflows
> @AlfredoG87 Yes, this is exactly what I have in mind as well. If the secret is not used by the block node then I think removing it completely would be better than keeping it as a placeholder as it causes more confusion on whether or not it is required. However, if it is used and is necessary for the deployment then what you have stated in the template above should work.
> If you want to keep it around as a placeholder then `blockNode.secret.enabled: false` might be a way to make sure that no one specifies both secret and secretRef together.
> So something like this might work better...
> 
> ```
> {{ if .Values.blockNode.secret.enabled }}
>   - secretRef:
>       name: {{ include "hiero-block-node.fullname" . }}-secret
> {{ else if .Values.blockNode.secretRef }}
>   - secretRef:
>       name: {{ .Values.blockNode.secretRef }}
> {{ end }}
> ```
> 
> The `else if` ensures that only one of the conditions (secret.enabled or secretRef) is evaluated.
> If secret.enabled is true, it uses the first block. Otherwise, it checks for secretRef.
> 
> Or this should also work without the `enabled` flag.
> 
> ```
> {{ if .Values.blockNode.secret }}
>   - secretRef:
>       name: {{ include "hiero-block-node.fullname" . }}-secret
> {{ else if .Values.blockNode.secretRef }}
>   - secretRef:
>       name: {{ .Values.blockNode.secretRef }}
> {{ end }}
> ```
> 
> I'll leave it to you on how you want to implement the conditional logic. 

 _Originally posted by @shezaan-hashgraph in [#999](https://github.com/hiero-ledger/hiero-block-node/issues/999#issuecomment-2797625668)_

AlfredoG87:
## Reviewer Notes

Due to how much Operators plan to deploy their BNs the best practice is to pre-create a secret source and then just specify it to the blockNode using a .Values override.

If for testing purposes still desires to create a secret object with entries provided in the collection of values, is still possible but discouraged other than for testing purposes

Fixes #1005

### DIFF
--- a/charts/block-node-server/templates/deployment.yaml
+++ b/charts/block-node-server/templates/deployment.yaml
@@ -83,8 +83,13 @@ spec:
         envFrom:
           - configMapRef:
               name: {{ include "hiero-block-node.fullname" . }}-config
+          {{ if .Values.blockNode.secret }}
           - secretRef:
-              name: {{ include "hiero-block-node.fullname" . }}-secret
+             name: {{ include "hiero-block-node.fullname" . }}-secret
+          {{ else if .Values.blockNode.secretRef }}
+          - secretRef:
+              name: {{ .Values.blockNode.secretRef }}
+          {{ end }}
         volumeMounts:
           - name: logging-config
             mountPath: {{ .Values.blockNode.logs.configMountPath }}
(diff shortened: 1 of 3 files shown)


jsync-swirlds (approved):

ata-nas (approved):
🔥

PR #1295, opened 2025-06-11
build: turn :block-node-protobuf-pbj into an alternative to :hapi

### LINKED ISSUE #1279: Modules `org.hiero.block.protobuf` and `com.hedera.node.hapi` can not not overlap
This was discovered when building the latest versions of both `hiero-block-node` and `hiero-consensus-node` from source in combination. It looks like `org.hiero.block.protobuf` contains generated code that is already generated into `com.hedera.node.hapi`. The latter however, is a transitive dependency of swirlds modules that block node modules depend on.

```
module org.hiero.block.common reads package com.hedera.hapi.node.base from both org.hiero.block.protobuf and com.hedera.node.hapi
```

### Reproduce

Clone both repos next to each other. Then:

1. Add `includeBuild("../hiero-consensus-node")` to `hiero-block-node/settings.gradle.kts`
2. Remove the [workaround patch block](https://github.com/hiero-ledger/hiero-block-node/blob/53f31e9e04fcc9573d8d4308886d8266e4f473af/settings.gradle.kts#L35-L54) in `hiero-block-node/settings.gradle.kts`
3. Add this list to `hiero-block-node/gradle/modules.properties` (needed due to recent group change for swirlds)
```
com.swirlds.base=com.hedera.hashgraph:swirlds-base
com.swirlds.common=com.hedera.hashgraph:swirlds-common
com.swirlds.config.api=com.hedera.hashgraph:swirlds-config-api
com.swirlds.config.extensions=com.hedera.hashgraph:swirlds-config-extensions
com.swirlds.config.impl=com.hedera.hashgraph:swirlds-config-impl
com.swirlds.metrics.api=com.hedera.hashgraph:swirlds-metrics-api
com.swirlds.state.api=com.hedera.hashgraph:swirlds-state-api
```

Run `./gradlew assemble` in `hiero-block-node`

jjohannes:
Using the same Module Name and Gradle Capabilities, the Modules become interchangeable. If both are on the Module Path, because `:hapi` is referenced transitively, Gradle will resolve the conflict by picking `:block-node-protobuf-pbj` only.

## Reviewer Notes

Resolves #1279 inthe current setup, where Protobuf java files are (re-)generated in multiple modules. 

## Related Issue(s)

Resolves #1279

### DIFF
--- a/block-node/app/src/testFixtures/java/module-info.java
+++ b/block-node/app/src/testFixtures/java/module-info.java
@@ -5,12 +5,12 @@ module org.hiero.block.node.app.test.fixtures {
     exports org.hiero.block.node.app.fixtures.blocks;
     exports org.hiero.block.node.app.fixtures.plugintest;
 
+    requires com.hedera.node.hapi;
     requires com.hedera.pbj.runtime;
     requires com.swirlds.common;
     requires com.swirlds.config.api;
     requires com.swirlds.metrics.api;
     requires org.hiero.block.node.spi;
-    requires org.hiero.block.protobuf.pbj;
     requires io.helidon.webserver;
     requires java.logging;
     requires org.junit.jupiter.api;
--- a/block-node/base/build.gradle.kts
+++ b/block-node/base/build.gradle.kts
@@ -22,5 +22,5 @@ testModuleInfo {
     requires("org.testcontainers")
     requires("io.minio")
     requires("junit")
-    requires("org.hiero.block.protobuf.pbj")
+    requires("com.hedera.node.hapi")
 }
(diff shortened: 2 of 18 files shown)


jsync-swirlds:
I don't know that we want to make these resolve to the same.
HAPI will not (cannot) include the block node API classes; and we'd prefer to avoid the `com.hedera` naming if we can.

We may _include_ some HAPI classes, but we're not copying it.

mhess-swl (approved):
LGTM from Execution side

jsync-swirlds (changes requested):
Re-iterating that block node artifacts should not be equivalent to HAPI.
The "correct" resolution would be to ensure that block node artifacts do not include HAPI proto classes/files.

jsync-swirlds:
> @jjohannes any progress on this PR ? @jsync-swirlds should we close the PR as @jjohannes suggested?

As I stated earlier, I do not believe this change should be made.

Closing as unnecessary.

PR #1298, opened 2025-06-11
test: Final metrics improvements

### LINKED ISSUE #1180: Verify Metrics are being updated and are working using `:9999/metrics` endpoint
### Story Form

As a BlockNode Developer and Operatorr
I want to make sure we don't have a regression and metrics stop working
So that we guarantee continuity of metrics.

### Technical Notes

Recently we found that the metrics were never starting correctly, hence they were not being updated and values remained as defaults when created.

We need an e2e test on the suites that makes sure that the metrics are being updated and working properly.

### LINKED ISSUE #1296: Unify Metrics Documents
### Story Form

As a BlockNode Dev
I want to only 1 metrics document for reference
So that I need to deduplicate the existing 2 metrics documents with some overlapping information and unify them into a single 1

### Technical Notes

Also some room for incremental improvements and updates

### LINKED ISSUE #1297: Add Grafana Dashboards to Helm Deployment Grafana provisioned Dashboards
### Story Form

As a BlockNode Operator
I want to Make use of provisioned dashboards with up-to-date new dashboards
So that I need to copy the recently created dashboards for the docker dev env into the grafana dashboard folder for helm deployments

### Technical Notes

Verify that they are appearing.

AlfredoG87:
## Reviewer Notes
- Add E2E Test Suite for Metrics verification
- Added Metrics Dashboards to Helm provisioning
- Unify and improved metrics documentation document

**- Bonus:** Update codecov exceptions, these have drastically change since last big refactor. 

Fixes #1180 
Fixes #1296 
Fixes #1297


codecov.yml
```diff
@@ -22,10 +22,7 @@ coverage:
...
 
 ignore:
-  - "block-node/server/src/main/java/org/hiero/block/server/Server.java"
-  - "block-node/base/src/main/java/org/hiero/block/server/mediator/NoOpLiveStreamMediator.java"
-  - "block-node/base/src/main/java/org/hiero/block/server/producer/NoOpProducerObserver.java"
-  - "block-node/base/src/main/java/org/hiero/block/server/persistence/storage/write/NoOpBlockWriter.java"
   - "simulator/src/main/java/org/hiero/block/simulator/BlockStreamSimulator.java"
   - "suites/**"
   - "tools/**"
+  - "**/src/test/**"
```
ata-nas, line 28: I believe here the `tools` is not correct. It is not at repo root level. Also, `suites` too.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/metrics/MetricsCommonTests.java
```diff
@@ -0,0 +1,133 @@
...
+public class MetricsCommonTests extends BaseSuite {
+
+    // Simulator instance to be used for testing
+    private BlockStreamSimulatorApp blockStreamSimulatorApp;
+
+    // Thread to run the simulator
+    private Future<?> simulatorThread;
+
+    @BeforeEach
+    void publishSomeBlocks() throws IOException, InterruptedException {
```
ata-nas, line 35: Not sure if we should propagate any `InterruptedException`s from the test methods. I am not sure how the library handles that. Maybe it is safer to simply call `Thread.currentThread().interrupt();` and then `fail()` the test if that is what an interrupt means for us, or continue with the test.

ata-nas (commented):
Reviewed all except `*.json` files. Looks good, I leave a couple of questions and a wildcard import we should probably remove.

codecov.yml
AlfredoG87: good catch!
✅ Done!

codecov.yml
```diff
@@ -22,10 +22,7 @@ coverage:
...
-  - "block-node/server/src/main/java/org/hiero/block/server/Server.java"
-  - "block-node/base/src/main/java/org/hiero/block/server/mediator/NoOpLiveStreamMediator.java"
-  - "block-node/base/src/main/java/org/hiero/block/server/producer/NoOpProducerObserver.java"
-  - "block-node/base/src/main/java/org/hiero/block/server/persistence/storage/write/NoOpBlockWriter.java"
   - "simulator/src/main/java/org/hiero/block/simulator/BlockStreamSimulator.java"
-  - "suites/**"
-  - "tools/**"
+  - "tools-and-tests/tools/**"
+  - "tools-and-tests/suites/**"
+  - "**/src/test/**"
```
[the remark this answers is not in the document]
AlfredoG87: 😅 yes, I forgot about that one, will create a ticket so we don't forget to add it in the next PR.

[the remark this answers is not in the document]
AlfredoG87: https://github.com/hiero-ledger/hiero-block-node/issues/1304

PR #1305, opened 2025-06-13
feat:  Support non-zero shard and realm entities in simulator transactions

### LINKED ISSUE #700: Support non-zero shard and realm entities in simulator transactions
### Persona

As a Simulator Operator

### Request

I want to be able to craft and stream transactions with non-zero shard/realm entity.

### Goal

So that we can fully test all possible scenarios of the block stream

### Technical Notes

New Generation mode was recently added in the Simulator. However it currently support only zero shard/realm transactions. We should make the neccesery changes to support this.

mustafauzunn:
## Reviewer Notes

Add support for non-zero shard and realm in sumulator

## Related Issue(s)
Fixes #700


jsync-swirlds (commented):
I think the issue for this is unclear.
Supporting non-zero shard and realm involves setting a (single, global) shard and realm for the simulator, then ensuring the following:
1. The shard and realm are set in all entities that support it
1. The simulator can (once the block node enables this) send shard and realm in a request to the block node to verify the block node matches the publisher
1. The simulator can (once the block node enables this) query block node as a consumer to verify it supports the same shard and realm the simulated consumer expects.

jsync-swirlds (approved):

PR #1324, opened 2025-06-20
feat: Simulator Add Publisher Support for EndOfStream Message

### LINKED ISSUE #1109: Simulator Add Publisher Support for `EndOfStream` Message
### Story Form

As a block node engineer
I want to test `EndOfStream` handling

### Technical Notes

Add support in the block stream simulator, publisher mode, for receiving a `EndOfStream` message from the block node and correctly handling that response.

mustafauzunn:
## Reviewer Notes

Add support in the block stream simulator, publisher mode, for receiving a EndOfStream message from the block node and correctly handling that response.

## Related Issue(s)
Fixes #1109

### DIFF
--- a/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
+++ b/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
@@ -65,16 +65,25 @@ public class PublishClientManager implements SimulatorModeHandler {
         currentHandler.stop();
     }
 
-    private void initializeNewClientAndHandler(Block nextBlock, PublishStreamResponse publishStreamResponse) {
-        streamEnabled = new AtomicBoolean(true);
-        currentClient = new PublishStreamGrpcClientImpl(
-                grpcConfig, blockStreamConfig, metricsService, streamEnabled, startupData);
-        currentHandler =
-                new PublisherClientModeHandler(blockStreamConfig, currentClient, blockStreamManager, metricsService);
+    public void handleResponse(Block nextBlock, PublishStreamResponse publishStreamResponse) {
+        if (publishStreamResponse.hasEndStream()) {
+            handleEndStream(nextBlock, publishStreamResponse);
+        }
+    }
 
-        currentHandler.setPublishClientManager(this);
-        currentHandler.init();
+    private void handleEndStream(Block nextBlock, PublishStreamResponse publishStreamResponse) {
+        try {
+            stop();
+            adjustStreamManager(nextBlock, publishStreamResponse);
+            initializeNewClientAndHandler();
 
+            start();
+        } catch (BlockSimulatorParsingException | IOException | InterruptedException e) {
+            throw new RuntimeException("Failed to restart handler", e);
+        }
+    }
+
+    private void adjustStreamManager(Block nextBlock, PublishStreamResponse publishStreamResponse) {
         if (nextBlock != null) {
             long blockNumber = publishStreamResponse.getEndStream().getBlockNumber();
             Code status = publishStreamResponse.getEndStream().getStatus();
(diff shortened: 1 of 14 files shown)


jsync-swirlds (commented):
A few possible improvements to consider.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
```diff
@@ -0,0 +1,113 @@
...
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+
+        initializeNewClientAndHandler(nextBlock, publishStreamResponse);
+
+        try {
+            currentHandler.start();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to restart handler", e);
```
ata-nas, line 102: Catching an exception just to throw it right after that probably means we should simply propagate the exception upwards (we do not know how to deal with it here.

ata-nas (commented):
First pass, some general comments and questions

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
```diff
@@ -0,0 +1,116 @@
...
+        if (nextBlock != null) {
+            long blockNumber = publishStreamResponse.getEndStream().getBlockNumber();
+            Code status = publishStreamResponse.getEndStream().getStatus();
+
+            switch (status) {
+                case TIMEOUT, BAD_BLOCK_PROOF -> blockStreamManager.resetToBlock(blockNumber - 1);
+                case BEHIND, DUPLICATE_BLOCK, SUCCESS -> blockStreamManager.resetToBlock(blockNumber + 1);
+                case INTERNAL_ERROR, PERSISTENCE_FAILED -> blockStreamManager.resetToBlock(blockNumber);
+            }
+        }
```
jsync-swirlds, line 87: I'm a bit confused by this block here.
It looks like you're trying to handle a response in the `initialize` method, which seems off.  Has the full request/response cycle been processed here?

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public void handleResponse(Block nextBlock, PublishStreamResponse publishStreamResponse) {
+        if (publishStreamResponse.hasEndStream()) {
+            handleEndStream(nextBlock, publishStreamResponse);
+        } else {
+            throw new IllegalArgumentException("Unknown PublishStreamResponse type: " + publishStreamResponse);
```
jsync-swirlds, line 94: This is not correct.  In addition to EndOfStream we should also get ResendBlock, SkipBlock, and Acknowledgement responses.  This else clause will cause the simulator to fail after the first acknowledged block every time.

We shouldn't be throwing for _any_ responses.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
jsync-swirlds: I think I get what's happening, this block really kind of wants to be a different method called from the `handleEndStream` method below, after the call to `stop` and _before_ the call to `initializeNewClientAndHandler`.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
jsync-swirlds: It looks like we're wrapping the exception here, which is a somewhat common approach to avoid declaring checked exceptions.
It's not my favorite pattern, but it is very popular.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublisherClientModeHandler.java
```diff
@@ -108,12 +111,16 @@ public void start() throws BlockSimulatorParsingException, IOException, Interrup
 
     private void millisPerBlockStreaming() throws IOException, InterruptedException, BlockSimulatorParsingException {
         final long secondsPerBlockNanos = (long) millisecondsPerBlock * NANOS_PER_MILLI;
+        PublishStreamResponse[] responseHolder = new PublishStreamResponse[1];
```
jsync-swirlds, line 114: This could just be `PublishStreamResponse currentResponse;`, and the lambda below then becomes `response -> currentResponse = response`...

jsync-swirlds (commented):
Significant improvement.
Still a few awkward constructs we might want to adjust.

PR #1337, opened 2025-06-27
refactor: Introduce Customized Logging for Simulator 

### LINKED ISSUE #892: Introduce Customized Logging for Simulator Charts
### Story Form

I want to have customized logging for the simulator when it is deployed in charts
So that I can use the custom logs

### Technical Notes

- taking inspiration from the server, we should enable the custom logging options and template to be available to a simulator if it is deployed via the charts

mustafauzunn:
## Reviewer Notes

- use  `CleanColorfulFormatter`  for simulator

> [!NOTE]  
> In a discussion with the author of the task agreed just to add `CleanColorfulFormatter` and use it in the simulator

## Related Issue(s)
 Fixes #892


jsync-swirlds:
> I'd say the logger is a good exception. Any updates we do to it will affect all usages. Copy-pasting this particular piece of code will mean that we need to support it in two places. I 100% agree however that the common module is not to share code between the block node and the simulator. It is there to share code between distinct module for our production code (block node and all related to it). But I also agree that we should drop the dependency of the common module to the simulator. So I am torn about this TBH.

I would not even use the common module to share code within the block node.  Plugins aren't actually part of the block node; they're separate and eventually we need to build them separately in order to make deployment independent and support the required deployment configurations.

In my opinion (and to be fair that's just one opinion) we should remove the common module entirely.

ata-nas (approved):
LG, as we spoke, we will copy the formatter instead of extracting it into commons.

jsync-swirlds (approved):

PR #1377, opened 2025-07-09
chore(protobuf): Upgrade of CN proto to latest release `0.63.9`

AlfredoG87:
## Related Issue(s)


jsync-swirlds (approved):

protobuf-sources/scripts/build-bn-proto.sh
```diff
@@ -90,9 +90,16 @@ fi
 rm -rf $output_dir
 mkdir -p $output_dir
 
-# Copy CN 'block' protobuf files to the output_dir directory, remove block_service.proto to avoid conflicts
+# Copy CN 'block' protobuf files to the output_dir directory, remove duplicated files to avoid conflicts
 cp -r ./hiero-consensus-node/hapi/hedera-protobuf-java-api/src/main/proto/block "$output_dir"
-rm -f ./block-node-protobuf/block/block_service.proto
+rm -f ./block-node-protobuf/block/block_access_service.proto
```
Nana-EC, line 95: I think there's more files here than needed.
I only see `block_stream_publish_service.proto` and `shared_message_types.proto`

Nana-EC (approved):
LG, minor edits for non-applicable files

protobuf-sources/scripts/build-bn-proto.sh
jsync-swirlds: The idea here is to be sure that our build is completely protected from copies, even if CN copies additional API files.
`rm -f` won't error if the file to be removed isn't present.

PR #1403, opened 2025-07-18
test: Support for Multiple BNs Test Suites

### LINKED ISSUE #1349: feat(Suites): Support for Multiple BNs Test Suites
### Support for Multiple BNs Test Suites 

Validate and create a testing framework to launch Several BNs with fill them with a certain defined (parameter) range. for consuption on another test.

**Acceptance Criteria:**

BlockNode Dev should be able to create a Test Suite, in which it can "launch" several BNs (each on its own port) and with different available Ranges (This will involve simulator configuration for each one).

Dev would need to specify Port and Range for each BN launched.

mustafauzunn:
## Reviewer Notes

BlockNode Dev should be able to create a Test Suite, in which several BNs could be started

## Related Issue(s)
 Fixes #1349


tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -95,6 +108,43 @@ public void teardown() throws InterruptedException {
...
+     */
+    protected void teardownBlockNodes() throws InterruptedException {
+        for (GenericContainer<?> container : blockNodeContainers) {
+            container.stop();
+            container.close();
+        }
+        blockNodeContainers.clear();
+
+        for (ManagedChannel ch : channels.values()) {
+            ch.shutdown().awaitTermination(5, TimeUnit.SECONDS);
```
ata-nas, line 122: Should we terminate immediately and not await any termination?

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -127,6 +177,25 @@ protected static GenericContainer<?> createContainer() {
         return blockNodeContainer;
     }
 
+    protected static GenericContainer<?> createContainer(BlockNodeConfig config) {
+        String blockNodeVersion = BaseSuite.getBlockNodeVersion();
+        var blockNodePort = config.getPort();
+        var blockNodeMetricsPort = config.getMetricPort();
```
ata-nas, line 183: Nit: let's not use `var`, maybe with very rare exceptions. I'd personally never use it.
Non-blocking.

ata-nas (dismissed):
🔥

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
jsync-swirlds: waiting up to 5 seconds is pretty short as it is.  Container shutdown can take a while if a lot of files are open (which is something block nodes may well encounter).

jsync-swirlds:
This is a somewhat _unrelated_ issue, but it arose while looking at this PR.

Have we checked the open files limit for the block node user in the containers?  The "default" limit for most systems is much smaller than we'll need for a production block node, and even testing, with this multiple-node support, might encounter open files limits.

If we haven't specifically set the open files for the block node process, we should open an issue to do so (I can open the issue, if we're sure we haven't already addressed this).

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -45,12 +50,20 @@ public abstract class BaseSuite {
     /** gRPC channel for connecting to Block Node */
     protected static ManagedChannel channel;
 
+    /** Map from port to gRPC channel */
+    protected Map<Integer, ManagedChannel> channels = new HashMap<>();
```
jsync-swirlds, line 54: Might be good to use a `LinkedHashMap` here to help keep debugging output consistent.  That helps when trying to reproduce failures.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -127,6 +177,25 @@ protected static GenericContainer<?> createContainer() {
         return blockNodeContainer;
     }
 
+    protected static GenericContainer<?> createContainer(BlockNodeConfig config) {
+        String blockNodeVersion = BaseSuite.getBlockNodeVersion();
+        var blockNodePort = config.getPort();
+        var blockNodeMetricsPort = config.getMetricPort();
+        List<String> portBindings = new ArrayList<>();
+        portBindings.add(String.format("%d:%2d", blockNodePort, blockNodePort));
+        portBindings.add(String.format("%d:%2d", blockNodeMetricsPort, blockNodeMetricsPort));
```
jsync-swirlds, line 186: Might be good to replace these strings with constants:
```suggestion
  private static final String PORT_BINDING_FORMAT = "%d:%2d"
        portBindings.add(PORT_BINDING_FORMAT.formatted(blockNodePort, blockNodePort));
        portBindings.add(PORT_BINDING_FORMAT.formatted(blockNodeMetricsPort, blockNodeMetricsPort));
```

Also, just curious, why do we have the 2 digit minimum width (space padded) in the second half?
for example, "%d:%2d".formatted(17635, 17635) produces "17635:17635" but "%d:%2d".formatted(5, 5) produces "5: 5", which is a bit confusing...
I wonder if this was intended to be "%d:%02d" or "%d:%05d" (or perhaps "%d:%2$d" if the intent was to select the second argument).

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BlockNodeConfig.java
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.suites;
+
+/**
+ * Configuration for a single Block Node instance in test suites.
+ */
+public class BlockNodeConfig {
```
jsync-swirlds, line 7: Why not use a record here?
Also, perhaps name `BlockNodeContainerConfig`, to differentiate from configuration for the block node software.

jsync-swirlds (approved):
Thanks for working on this.
A few improvements, but nothing that should block the PR.

Nana-EC (approved):
Good addition.

Kindly resolve current comments but otherwise looks good.

It would also be good to add the simplest of tests case as an example of how your expect this to be utilized.
Maybe a test that starts up 3 BNs and get hit the status API of each to confirm block progression.

ata-nas (approved):

ata-nas (approved):
Thanks for these changes, this will be very handy when testing the multipublisher plugin!
Great job @mustafauzunn! 🔥

ata-nas:
> > This is a somewhat _unrelated_ issue, but it arose while looking at this PR.
> > Have we checked the open files limit for the block node user in the containers? The "default" limit for most systems is much smaller than we'll need for a production block node, and even testing, with this multiple-node support, might encounter open files limits.
> > If we haven't specifically set the open files for the block node process, we should open an issue to do so (I can open the issue, if we're sure we haven't already addressed this).
> 
> I don't think we addressed this. You can open the issue

I also do not think we have tackled this.

jsync-swirlds (approved):

jsync-swirlds:
> > This is a somewhat _unrelated_ issue, but it arose while looking at this PR.
> > Have we checked the open files limit for the block node user in the containers? The "default" limit for most systems is much smaller than we'll need for a production block node, and even testing, with this multiple-node support, might encounter open files limits.
> > If we haven't specifically set the open files for the block node process, we should open an issue to do so (I can open the issue, if we're sure we haven't already addressed this).
> 
> I don't think we addressed this. You can open the issue

Done, Issue #1430.

PR #1407, opened 2025-07-22
chore: Remove BlockAcknowledgement.block_root_hash from Publisher API

### LINKED ISSUE #1395: Remove BlockAcknowledgement.block_root_hash from Publisher API
### Story Form

As a Block Node Operator
I want to have an accurate set of BlockAcknowledgement responses as part of the publisher API without block_root_hash
So that there is clarity on the behaviour of the Block Node when acknowledging a block

### Technical Notes

Remove BlockAcknowledgement.block_root_hash as this is outdated and not believed to be utilized by CN

Nana-EC:
## Reviewer Notes
`BlockAcknowledgement.block_root_hash` is present in the publisher API but outdated and not utilized by CN

- Remove BlockAcknowledgement.block_root_hash from Publisher API (`block_stream_publish_service.proto`)
- Remove `blockHash` from SimulatorStartupData.updateLatestAckBlockStartupData that was utilized as the `blockRootHash` on an acknowledged block. ALso simplify logic given fewer cases
- Remove `blockRootHash` logic from `SimulatorStartupDataImplTest` logic 
- Increase `SimulatorStartupDataImplTest` coverage with `testFailedInitializationUnavailableBlockNumberFile`
- Fix and increase `PublishStreamObserverTest` with `verifyUpdateLatestAckBlockStartupDataHandlesIOException()` and a correction of BlockAcknowledgment  in response

## Related Issue(s)
Fixes #1395

### DIFF
--- a/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/startup/impl/SimulatorStartupDataImpl.java
+++ b/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/startup/impl/SimulatorStartupDataImpl.java
@@ -55,6 +55,8 @@ public final class SimulatorStartupDataImpl implements SimulatorStartupData {
                         // state. We must investigate why this is happening.
                         // Generally we never ever expect to enter here, but
                         // we cannot continue to initialize the simulator
+                        throw new IllegalStateException(
+                                "Failed to initialize Simulator Startup Data, only one startup data file exists!");
                     }
                     case 2 -> {
                         // entering here means that both files exist, so now we
--- a/tools-and-tests/simulator/src/test/java/org/hiero/block/simulator/startup/impl/SimulatorStartupDataImplTest.java
+++ b/tools-and-tests/simulator/src/test/java/org/hiero/block/simulator/startup/impl/SimulatorStartupDataImplTest.java
@@ -25,11 +25,18 @@ class SimulatorStartupDataImplTest {
     @TempDir
     private Path tempDir;
 
+    private byte[] validSimulatedBlockHash;
     private Path latestAckBlockNumberPath;
+    private Path latestAckBlockHashPath;
 
     @BeforeEach
     void setup() {
+        validSimulatedBlockHash = new byte[StreamingTreeHasher.HASH_LENGTH];
+        for (byte i = 0; i < StreamingTreeHasher.HASH_LENGTH; i++) {
+            validSimulatedBlockHash[i] = i;
+        }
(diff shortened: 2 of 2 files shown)


AlfredoG87 (approved):
LGTM 💯

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamObserver.java
```diff
@@ -66,8 +66,7 @@ public void onNext(final PublishStreamResponse publishStreamResponse) {
         if (publishStreamResponse.hasAcknowledgement()) {
             final BlockAcknowledgement ack = publishStreamResponse.getAcknowledgement();
             try {
-                startupData.updateLatestAckBlockStartupData(
-                        ack.getBlockNumber(), ack.getBlockRootHash().toByteArray());
+                startupData.updateLatestAckBlockStartupData(ack.getBlockNumber());
```
ata-nas, line 69: Please note that the startup data will stop functioning w/o the hash to be saved on disk. This means that the next startup will be w/o previous hash and everything will be broken after that. We should add a way to save the current hash (as it is generated by the simulator) and then retrieving and storing that in the proper place once ack is received.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/startup/impl/SimulatorStartupDataImpl.java
```diff
@@ -104,13 +74,12 @@ public boolean isEnabled() {
...
 
     @Override
-    public void updateLatestAckBlockStartupData(final long blockNumber, final byte[] blockHash) throws IOException {
+    public void updateLatestAckBlockStartupData(final long blockNumber) throws IOException {
         if (enabled) {
             // @todo(904) we need the correct response code, currently it seems that
             //   the response code is not being set correctly? The if check should
             //   be different and based on the response code, only saving
             Files.write(latestAckBlockNumberPath, String.valueOf(blockNumber).getBytes());
-            Files.write(latestAckBlockHashPath, blockHash);
```
ata-nas, line 113: I do not think we can omit this, we simply need another way to supply the hash.

ata-nas (commented):
LG, I have a concern however for the simulator startup data. @AlfredoG87 this could interfere with you long-running tests?

Nana-EC:
@mustafauzunn can you kindly complete this?

PR #1410, opened 2025-07-22
chore: Create basic config file .gitvote.yml

jwagantall:
For more information: https://github.com/cncf/gitvote


Nana-EC:
Thanks @jwagantall.
I can see this being utilized well on Issues.
In the PR case when a repo has a min required signs offs where do you think the voting system would be valuable in comparison to just having the PR being approved?

Nana-EC (approved):

Issue #1420, opened 2025-07-23
Multi-Publisher Plugin Documentation


ata-nas:
### Story Form

As a Block Node User/Operator/Engineer
I want to have well written and detailed documentation
So that I can understand the plugin well

### Technical Notes

**DEPENDS ON**: #1370
- the plugin rework requires documentation update
- we should update all javadocs and/or design docs as necessary

Issue #1419, opened 2025-07-23
Multi-Publisher Plugin Unit Tests


ata-nas:
### Story Form

As a Block Node Engineer/Operator/User
I want to have unit tests for my classes
So that I can have confidence in existing logic and catch regressions

### Technical Notes

**DEPENDS ON**: #1370
- we have some unit tests already
- we must add additional unit tests to boost confidence

Issue #1418, opened 2025-07-23
Multi-Publisher Plugin E2E Test Plan


ata-nas:
### Story Form

As a Block Node Operator/User/Engineer
I want to have a well detailed E2E test plan
So that I can follow it to implement tests in order to have confidence in my plugin

### Technical Notes

- create E2E test plan for the Multi-Publisher plugin

Issue #1417, opened 2025-07-23
Multi-Publisher Async Acknowledgements


ata-nas:
### Story Form

As a Block Node Operator/Engineer
I want to send acknowledgements asynchronously
So that I will not hold the messaging thread for a long time

### Technical Notes

**DEPENDS ON**: #1370
- we are currently sending acknowledgements for verified and persisted blocks synchronously and we hold the messaging thread
- we should send acknowledgements asynchronously

PR #1421, opened 2025-07-23
feat(1370): Rebuild Stream Publisher for Finalized Protocol Support

### LINKED ISSUE #1370: Rebuild Stream Publisher for Finalized Protocol Support
### Story Form

N/A

### Technical Notes

The stream publisher plugin was originally built with a one-way publish assumption, extensive synchronization, and acknowledgement sent for error conditions. These assumptions and criteria turned out to be incorrect, so we must rebuild the plugin to match the design required by the full interactive protocol between publisher and block node.
This includes, but is not limited to, support for sending `EndOfStream` messages for errors or to limit the duration of connections, support for multiple publishers, support for timeout based on receiving a later block proof from another publisher, support for receiving multiple blocks from multiple publishers in semi-parallel, etc...

Handling certain specific items is detailed in other features under the same parent epic, and is not part of this feature.

ata-nas:
## Reviewer Notes
* Updated the unparsed proto to match updated request and response types
* Removed the old plugin classes
   * Those classes assumed no control messages, which doesn't work
* Defined the pipeline methods for the plugin
* Started filling in the Manager
* Split the metrics appropriately between Manager and Handler
* Mostly completed the Handler (thanks, @ata-nas)
* Added tests for the new Handler
* Made `StreamPublisherManager` an interface to help with testing
   * Implemented test fixture, renamed existing class to `LiveStreamPublisherManager`.
* Moved notification handling to the manager class (live only for now)
* Started handling multiple streams in manager
* Handled sending batches to messaging, including catching up with future blocks where available.
* Began to add all the details and testing needed
* Added `@todo` items and line comments for several next steps.
* Also upgraded the Java patch version to reduce reported build warnings.

**UPDATE**:
As follow up items, the following were added:
- #1413 
- #1414 
- #1415 
- #1416 
- #1417 
- #1418 
- #1419 
- #1420 
- #1422 

## Related Issue(s)

Closes #1370


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -0,0 +1,149 @@
...
+        return switch (blockStreamPublisherServiceMethod) {
+            case publishBlockStream ->
+                Pipelines.<PublishStreamRequestUnparsed, PublishStreamResponse>bidiStreaming()
+                        .mapRequest(PublishStreamRequestUnparsed.PROTOBUF::parse)
+                        .method(this::initiatePublisherHandler)
+                        .respondTo(replies)
+                        .mapResponse(PublishStreamResponse.PROTOBUF::toBytes)
+                        .build();
+        };
+    }
```
AlfredoG87, line 81: would be worth noting why we need to use `open` instead of `publishBlockStream` in the javadoc or comments for this method?
Edit: I do see the comment on the dead method below.

gradle/toolchain-versions.properties
```diff
@@ -1 +1,3 @@
-jdk=21.0.6
+# SPDX-License-Identifier: Apache-2.0
+
+jdk=21.0.7
```
AlfredoG87, line 3: why are we updating the jdk version here?

tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
```diff
@@ -140,6 +140,7 @@ public void shouldSwitchToFasterPublisherWhenCaughtUp() throws IOException, Inte
      * a consistent and chronological order of block processing, even when multiple publishers with
      * different block timings are connected simultaneously.
      */
+    @Disabled("Temporarily disabled whiles publisher plugin is being rewritten.")
```
AlfredoG87, line 143: should we re-enable this test or something?

AlfredoG87 (approved):
Amazing work, I loved how every piece falls into place in such elegant way, there is a lot of comments to tackle but the core functionality and logic is there and is working, great stuff!! 💯 

Just leaved a couple minor questions.

Also, did some e2e manual testing using some pretty crazy load and speed with several publishers (scenarios that the old plugin would die with) and this one passed them with flying colors, great work @jsync-swirlds  and @ata-nas!!

gradle/toolchain-versions.properties
jsync-swirlds: Two reasons:
1. gradle complains about the earlier java version, and it was causing automation blips
2. We need to upgrade Java for the latest bug fixes anyway, so it seemed reasonable to do so here.

Side note, we should prepare to upgrade to Java 25 in late September (when it's GA).  Tons of major improvements (also, JEP 467, which cannot arrive soon enough) for both engineering and performance in that LTS release.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
jsync-swirlds: Yes, one of the tasks in this next sprint will re-enable and/or replace this test.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -0,0 +1,469 @@
...
+        // called when the _gRPC layer_ receives an HTTP end stream from the client.
+        // THIS IS NOT the same as the `EndStream` message in the API.
+        shutdown(); // @todo(1416) is this all the handling we need to do here?
+    }
+
+    @Override
+    public void onSubscribe(@NonNull final Subscription subscription) {
+        // Check javadoc, this "starts" the subscription for this handler.
+        // mostly we need to call subscription.request to start the flow of items.
+        // @todo(1416) should we do something here?
```
Nana-EC, line 127: Potentially a metric capturing subscription.
May not be needed as I'm sure there's a metric higher up capturing this.

jsync-swirlds: There aren't any application-level subscriptions here (the naming here is super unfortunate, but comes from PBJ and Helidon).  This is just internal wiring for the plugin.

PR #1437, opened 2025-07-28
Task: Make Publisher Acknowledgements Asynchronous

### LINKED ISSUE #1417: Multi-Publisher Async Acknowledgements

jsync-swirlds:
## Reviewer Notes
* Adjusted some javadoc in BlockNotificationHandler that was slightly vague and did not follow recommend JavaDoc forms.
* Adjusted the `LiveStreamPublisherManager.handlePersisted` method as follows:
    1. Moved inline comments to JavaDoc for better documentation of requirements and design decisions.
    1. Changed the for loop to use the Streams API and `parallelStream` along with `unordered` to encourage maximum parallelism.
* This requires testing; if the tasks are not sufficiently parallelized by the Streams API, we may need to implement our own fork/join processing.

## Related Issue(s)
 Resolves #1417


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -263,26 +263,41 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
-        //     need to check that new persisted is > last persisted
-        //     before doing more.
         final long newLastPersistedBlock = notification.endBlockNumber();
-        // @note it is _not_ OK to block here. What we need is a forkJoin task
-        // that forks for each handler and sends acks for all in thread(s)
-        // then joins at the end. That can be added next sprint, however.
-        if (newLastPersistedBlock >= lastPersistedBlockNumber.get()) {
-            for (final PublisherHandler handler : handlers.values()) {
+        if (newLastPersistedBlock > lastPersistedBlockNumber.get()) {
+            handlers.values().parallelStream().unordered().forEach(handler -> {
```
ata-nas, line 294: Cool solution!
Correct me if I am wrong, but the parallel stream will use the common fork join pool by default. Wouldn't it be worth it if we can control what pool will be used to execute the parallel operations? I believe it is possible to do so with the stream API, but I might be wrong?

jsync-swirlds: It would be nice to get a dedicated fork/join pool, but there is no version of the `parallelStream` method that supports providing a pool or executor.

ata-nas (approved):
🔥

georgi-l95 (approved):
LG

Issue #1236, opened 2025-06-02
Add Support to Stream-Publisher for Handling `EndStream` Messages


jsync-swirlds:
### Story Form

As a Publisher
I want to send `EndStream` messages
So that I can orderly end publish streams and inform the Block Node of the reason

### Technical Notes

The `EndStream` message is part of the Block Node `publish` API, and the Block Node `stream-publisher` plugin must properly implement handling as described in the publish protocol [documentation](https://github.com/hiero-ledger/hiero-block-node/blob/main/docs/design/communication-protocol/publish-block-stream.md).

This story assumes that the basic protocol and plugin functionality is already complete.

PR #1444, opened 2025-07-30
feat(backfill): On-Demand Backfilling and other improvements

### LINKED ISSUE #1355: feat(backfill): support NewestBlockKnownToNetwork trigger
### On-Demand Backfill Trigger
 
- subscribe to NewestBlockKnownToNetwork on the bus; when received, recalculate gaps up to the indicated height and, if any, enqueue fetch immediately
- implement concurrency guard so on-demand runs don’t overlap with scheduled scan.

**Acceptance:**
Given a live BN missing blocks and receiving NewestBlockKnownToNetwork, then it backfills.

AlfredoG87:
- Added On-Demand Backfill flow
- Modified Backfill plugin to support both flows to run at the same time.
- Improved UTs to make them simpler and less repetitive, also improved reliability and all tests are 100% independent from one another.
- Added several new UTs related to Backfilling
- Allowed for partial chunks to be fetched instead of the complete batch (however this will required several scan iterations to complete a batch in certain scenarios)
- Added a new UT where a single gap has to be fetched from several BNs. 
- Modified publisherHandler to add capability to handle `TOO_FAR_BEHIND`, *EndStream* type message Request.
- Gave publisher MessagingFacility to be able to send Notifications. (minimal modifications of tests)
- Added missing metrics for new Block notifications types

Fixes #1355

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -2,7 +2,7 @@
 package org.hiero.block.node.backfill;
 
 import static java.lang.System.Logger.Level.INFO;
-import static java.lang.System.Logger.Level.WARNING;
+import static java.lang.System.Logger.Level.TRACE;
 
 import com.hedera.hapi.block.stream.output.BlockHeader;
 import com.hedera.pbj.runtime.ParseException;
@@ -18,6 +18,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 import org.hiero.block.internal.BlockUnparsed;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
@@ -25,6 +26,7 @@ import org.hiero.block.node.spi.ServiceBuilder;
 import org.hiero.block.node.spi.blockmessaging.BackfilledBlockNotification;
 import org.hiero.block.node.spi.blockmessaging.BlockNotificationHandler;
 import org.hiero.block.node.spi.blockmessaging.BlockSource;
+import org.hiero.block.node.spi.blockmessaging.NewestBlockKnownToNetworkNotification;
 import org.hiero.block.node.spi.blockmessaging.PersistedNotification;
 import org.hiero.block.node.spi.blockmessaging.VerificationNotification;
 import org.hiero.block.node.spi.historicalblocks.LongRange;
@@ -51,9 +53,13 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
     // Backfill state
     private List<LongRange> detectedGaps = new ArrayList<>();
     private BackfillGrpcClient backfillGrpcClient;
-    private AtomicLong backfillPendingBlocks = new AtomicLong(0);
-    private volatile boolean isDetectingGaps = false;
-    private CountDownLatch pendingVerificationAndPersistenceLatch;
+
+    // State touched by multiple threads
+    private final AtomicReference<CountDownLatch> autonomousLatch = new AtomicReference<>(new CountDownLatch(0));
+    private final AtomicReference<CountDownLatch> onDemandLatch = new AtomicReference<>(new CountDownLatch(0));
+    private volatile boolean autonomousError = false;
+    private volatile boolean onDemandError = false;
+    private final AtomicLong onDemandBackfillStartBlock = new AtomicLong(-1);
 
     // Metrics
     private Counter backfillGapsDetected;
@@ -61,8 +67,21 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
     private Counter backfillBlocksBackfilled;
     private Counter backfillFetchErrors;
     private Counter backfillRetries;
-    private LongGauge backfillStatus;
+    private LongGauge backfillStatus; // 0 = idle, 1 = running, 2 = autonomous-error, 3 = on-demand-error
     private LongGauge backfillPendingBlocksGauge;
+
+    private AtomicReference<CountDownLatch> getLatch(BackfillType t) {
+        if (t == BackfillType.AUTONOMOUS) {
+            return autonomousLatch;
+        } else if (t == BackfillType.ON_DEMAND) {
+            return onDemandLatch;
+        } else {
+            // This should never happen (as the code is right now is impossible)
+            // but we throw an exception to be safe in the future in case the enum is extended
+            LOGGER.log(INFO, "Unknown backfill type: {0}", t);
+            throw new IllegalArgumentException("Unknown backfill type: " + t);
+        }
+    }
 
     /**
      * {@inheritDoc}
@@ -94,6 +113,38 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
         backfillPendingBlocksGauge =
                 metrics.getOrCreate(new LongGauge.Config(METRICS_CATEGORY, "backfill_pending_blocks")
                         .withDescription("Current amount of blocks pending to be backfilled."));
+
+        metrics.addUpdater(this::updateMetrics);
+    }
+
+    private void updateMetrics() {
+        // calculate the gauge of pending blocks metrics
+        long pendingBackfillBlocks =
+                autonomousLatch.get().getCount() + onDemandLatch.get().getCount();
+        backfillPendingBlocksGauge.set(pendingBackfillBlocks);
+
+        final BackfillStatus status;
+        if (pendingBackfillBlocks > 0) {
+            status = BackfillStatus.RUNNING;
+        } else if (pendingBackfillBlocks == 0 && !autonomousError && !onDemandError) {
+            status = BackfillStatus.IDLE;
+        } else if (autonomousError) {
+            status = BackfillStatus.ERROR;
+        } else if (onDemandError) { // onDemandError
+            status = BackfillStatus.ON_DEMAND_ERROR;
+        } else {
+            status = BackfillStatus.UNKNOWN;
+        }
+
+        backfillStatus.set(status.ordinal());
+    }
+
+    private enum BackfillStatus {
+        IDLE, // 0
+        RUNNING, // 1
+        ERROR, // 2
+        ON_DEMAND_ERROR, // 3
+        UNKNOWN // 4
     }
 
     /**
@@ -103,22 +154,23 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
     public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
         this.context = context;
         backfillConfiguration = context.configuration().getConfigData(BackfillConfiguration.class);
+        autonomousLatch.set(new CountDownLatch(0));
+        onDemandLatch.set(new CountDownLatch(0));
 
         // Initialize metrics
         initMetrics();
-        backfillStatus.set(0); // 0 = idle
 
         // Validate block node sources configuration
         final String sourcesPath = backfillConfiguration.blockNodeSourcesPath();
         if (sourcesPath == null || sourcesPath.isBlank()) {
-            LOGGER.log(INFO, "No block node sources path configured, backfill will not run");
+            LOGGER.log(TRACE, "No block node sources path configured, backfill will not run");
             return;
         }
 
         Path blockNodeSourcesPath = Path.of(backfillConfiguration.blockNodeSourcesPath());
         if (!Files.isRegularFile(blockNodeSourcesPath)) {
             LOGGER.log(
-                    INFO,
+                    TRACE,
                     "Block node sources path file does not exist: {0}, backfill will not run",
                     backfillConfiguration.blockNodeSourcesPath());
             return;
@@ -131,9 +183,9 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
                     backfillConfiguration.maxRetries(),
                     this.backfillRetries,
                     backfillConfiguration.initialRetryDelayMs());
-            LOGGER.log(INFO, "Initialized gRPC client with sources path: {0}", blockNodeSourcesPath);
+            LOGGER.log(TRACE, "Initialized gRPC client with sources path: {0}", blockNodeSourcesPath);
         } catch (Exception e) {
-            LOGGER.log(WARNING, "Failed to initialize gRPC client: {0}", e.getMessage());
+            LOGGER.log(INFO, "Failed to initialize gRPC client: {0}", e.getMessage());
             hasBNSourcesPath = false;
             return;
         }
@@ -153,7 +205,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             return;
         }
         LOGGER.log(
-                INFO,
+                TRACE,
                 "Scheduling backfill process to start in {0} milliseconds",
                 backfillConfiguration.initialDelayMs());
         scheduler = Executors.newSingleThreadScheduledExecutor();
@@ -173,30 +225,40 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             scheduler.shutdownNow();
             try {
                 if (!scheduler.awaitTermination(5, TimeUnit.SECONDS)) {
-                    LOGGER.log(INFO, "Scheduler did not terminate in time");
+                    LOGGER.log(TRACE, "Scheduler did not terminate in time");
                     // terminate forcefully
                     scheduler.shutdown();
                 }
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
             }
-            LOGGER.log(INFO, "Stopped backfill process");
+            LOGGER.log(TRACE, "Stopped backfill process");
         }
+    }
+
+    private boolean isAutonomousBackfillRunning() {
+        return autonomousLatch.get().getCount() > 0;
+    }
+
+    private boolean isOnDemandBackfillRunning() {
+        return onDemandBackfillStartBlock.get() != -1;
     }
 
     private void detectGaps() {
         // Skip if already running
-        if (isDetectingGaps) {
-            LOGGER.log(INFO, "Gap detection already in progress, skipping this execution");
+        if (isAutonomousBackfillRunning()) {
+            LOGGER.log(TRACE, "Gap detection already in progress, skipping this execution");
             return;
         }
 
-        // Set running state
-        isDetectingGaps = true;
-        backfillStatus.set(1); // 1 = running
+        // Skip if OnDemand backfill is running
+        if (isOnDemandBackfillRunning()) {
+            LOGGER.log(TRACE, "On-Demand backfill is running, skipping autonomous gap detection");
+            return;
+        }
 
         try {
-            LOGGER.log(INFO, "Detecting gaps in historical blocks");
+            LOGGER.log(TRACE, "Detecting gaps in historical blocks");
             detectedGaps = new ArrayList<>();
 
             // Get the configured first block available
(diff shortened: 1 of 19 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -158,6 +171,42 @@ public void onNext(@NonNull final PublishStreamRequestUnparsed request) {
...
+            blockMessagingFacility.sendNewestBlockKnownToNetwork(notification);
+
+            // log the newest block known to network
+            LOGGER.log(
+                    TRACE,
+                    "Attempted to recover from TOO_FAR_BEHIND end stream request, latest block number: {0}, sending NewestBlockKnownToNetworkNotification",
+                    endStreamLatestBlockNumber);
+        }
+    }
+
```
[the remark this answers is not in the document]
AlfredoG87: This is the part where I needed the most feedback from you @ata-nas  and @jsync-swirlds, I created a `handleEndStreamRequest` since currently is a TODO, however, I am un-aware if this is how those will be handled, I tried to keep it simple in case of a needed refactor once the complete api spec gets done.

I would say that either, this PR is ready (just in need of approval and feedback addressing), so imo I would say that the last to be merged should refactor accordingly.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -94,6 +99,27 @@ private void initMetrics() {
...
+
+        // Update the backfill status based on the current state
+        if (pendingBackfillBlocks > 0) {
+            backfillStatus.set(1); // 1 = running
+        } else if (pendingBackfillBlocks == 0 && !autonomousError && !onDemandError) {
+            backfillStatus.set(0); // 0 = idle
+        } else if (autonomousError) {
+            backfillStatus.set(2); // 2 = error
+        } else if (onDemandError) {
+            backfillStatus.set(3); // 3 = on-demand error
```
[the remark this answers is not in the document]
jsync-swirlds: Even better, use an enum, if you need an integer you can assign it from the enum `ordinal()` value.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -265,46 +295,56 @@ private void processDetectedGaps() {
...
+            backfillLatches.put(backfillType, new CountDownLatch(batchOfBlocks.size()));
 
             // Process each fetched block
             for (BlockUnparsed blockUnparsed : batchOfBlocks) {
                 long blockNumber = extractBlockNumber(blockUnparsed);
                 context.blockMessaging()
                         .sendBackfilledBlockNotification(new BackfilledBlockNotification(blockNumber, blockUnparsed));
 
+                // Give messaging system some time to process
+                parkNanos(500_000); // 500 microseconds
```
jsync-swirlds, line 320: We shouldn't need to delay here, we wait on the latch after this loop...

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -359,12 +408,83 @@ public void handlePersisted(PersistedNotification notification) {
...
+            try {
+                // Set the start block for on-demand backfill BEFORE scheduling the task
+                onDemandBackfillStartBlock = gap.start();
+
+                // use the scheduler to run the backfill in its own thread (only call backfillGap once)
+                scheduler.execute(() -> {
+                    try {
+                        LOGGER.log(INFO, "Starting on-demand backfill for gap: {0}", gap);
+                        backfillGap(gap, BackfillType.ON_DEMAND);
+                    } catch (Exception e) {
```
jsync-swirlds, line 458: We should catch `RuntimeException` here instead of `Exception`.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -88,8 +88,13 @@ public PublisherHandler addHandler(
...
         final long handlerId = nextHandlerId.getAndIncrement();
-        final PublisherHandler newHandler =
-                new PublisherHandler(handlerId, replies, handlerMetrics, this, registerTransferQueue(handlerId));
+        final PublisherHandler newHandler = new PublisherHandler(
+                handlerId,
+                replies,
+                handlerMetrics,
+                this,
+                registerTransferQueue(handlerId),
+                serverContext.blockMessaging());
```
jsync-swirlds, line 97: The Handlers should not have any connection to messaging.  We very deliberately keep that restricted to the publisher manager.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -184,16 +211,28 @@ public void stop() {
         }
     }
 
+    private boolean isAutonomousBackfillRunning() {
+        long pendingAutonomousBlocks =
+                backfillLatches.get(BackfillType.AUTONOMOUS).getCount();
+        return pendingAutonomousBlocks > 0;
```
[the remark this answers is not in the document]
AlfredoG87: I dont think these are strictly statuses, just would be RUNNING, NOT_RUNNING, so a boolean is simpler.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -184,16 +211,28 @@ public void stop() {
...
+    }
+
+    private boolean isOnDemandBackfillRunning() {
+        return onDemandBackfillStartBlock != -1;
```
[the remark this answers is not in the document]
AlfredoG87: same as above.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -184,16 +211,28 @@ public void stop() {
...
             LOGGER.log(INFO, "Gap detection already in progress, skipping this execution");
             return;
         }
 
-        // Set running state
-        isDetectingGaps = true;
-        backfillStatus.set(1); // 1 = running
+        // Skip if OnDemand backfill is running
+        if (isOnDemandBackfillRunning()) {
+            LOGGER.log(INFO, "On-Demand backfill is running, skipping autonomous gap detection");
```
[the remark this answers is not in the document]
AlfredoG87: Yes, since Backfill is still in the works I have almost all logs at INFO Level, but we might as well change them to DEBUG sooner rather than later.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -265,46 +295,56 @@ private void processDetectedGaps() {
...
+                parkNanos(500_000); // 500 microseconds
+
                 LOGGER.log(INFO, "Backfilling block {0}", blockNumber);
                 backfillFetchedBlocks.increment();
             }
 
             // Wait for verification and persistence to complete
-            pendingVerificationAndPersistenceLatch.await();
+            // give about 1 second for each block to be processed (should be much faster)
+            boolean backfillFinished = backfillLatches.get(backfillType).await(batchOfBlocks.size(), TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
AlfredoG87: yes, and should not happen, but in case it does, the plugin self-recovers, so we might as well just timeout and continue.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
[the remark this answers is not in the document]
AlfredoG87: ✅  Done!

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Yeah, I add it just out of abundance of caution to be somewhat gentle on the messaging facility.
but will remove it as it should not be necessary.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: the method can return `ParseException` or `InterruptedException` which both are checked exceptions.
I would create a catch for each but the handle would be the same so I decided to go Generic with Exception

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
AlfredoG87: would you recommend to create a public method on the `PublisherManager` instead?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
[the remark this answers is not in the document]
AlfredoG87: My thought was that there might be some EndStream that might be better handled by the publisherHandler and some others that might be handled by the publishManager. 

we have: UNKNOW, RESET, TIMEOUT, ERROR and finally TOO_FAR_BEHIND.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
AlfredoG87: I was not able to make the method package public since the handler uses an Interface of the StreamPublisherManager.

That said, I was able to simplify the code by moving the handling of the EndStream outside the handler into the manager.

✅ Done!

AlfredoG87: I was not able to make the method package public since the handler uses an Interface of the StreamPublisherManager.

That said, I was able to simplify the code by moving the handling of the EndStream outside the handler into the manager.

✅ Done!

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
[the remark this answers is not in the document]
AlfredoG87: @ata-nas I've addressed this to potentially align it more with current WIP #1236 and reference on a todo

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.backfill;
 
+// import static java.lang.System.Logger.Level.TRACE;
```
ata-nas, line 4: Commented out import.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -51,19 +54,28 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
     // Backfill state
     private List<LongRange> detectedGaps = new ArrayList<>();
     private BackfillGrpcClient backfillGrpcClient;
-    private AtomicLong backfillPendingBlocks = new AtomicLong(0);
-    private volatile boolean isDetectingGaps = false;
-    private CountDownLatch pendingVerificationAndPersistenceLatch;
+
+    // State touched by multiple threads
+    // private final Map<BackfillType, CountDownLatch> backfillLatches = new HashMap<>();
```
ata-nas, line 59: Commented out field.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -51,19 +54,28 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
...
+    private final AtomicReference<CountDownLatch> autonomousLatch = new AtomicReference<>(new CountDownLatch(0));
+    private final AtomicReference<CountDownLatch> onDemandLatch = new AtomicReference<>(new CountDownLatch(0));
+    private volatile boolean autonomousError = false;
+    private volatile boolean onDemandError = false;
```
ata-nas, line 63: Is volatility for these booleans sufficient here? These are not thread safe. Just making sure.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -51,19 +54,28 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
...
     private Counter backfillBlocksBackfilled;
     private Counter backfillFetchErrors;
     private Counter backfillRetries;
-    private LongGauge backfillStatus;
+    private LongGauge backfillStatus; // 0 = idle, 1 = running, 2 = autonomous-error, 3 = on-demand-error
     private LongGauge backfillPendingBlocksGauge;
 
+    private AtomicReference<CountDownLatch> getLatch(BackfillType t) {
+        return (t == BackfillType.AUTONOMOUS) ? autonomousLatch : onDemandLatch;
+    }
```
ata-nas, line 77: Nit, (more of a preference) we should probably `switch` this enum. Reason is that if we change the enum's values, this code will still work, but we might not want it to. Just thinking about the future.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -51,19 +54,28 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
...
     private List<LongRange> detectedGaps = new ArrayList<>();
     private BackfillGrpcClient backfillGrpcClient;
-    private AtomicLong backfillPendingBlocks = new AtomicLong(0);
-    private volatile boolean isDetectingGaps = false;
-    private CountDownLatch pendingVerificationAndPersistenceLatch;
+
+    // State touched by multiple threads
+    // private final Map<BackfillType, CountDownLatch> backfillLatches = new HashMap<>();
+    private final AtomicReference<CountDownLatch> autonomousLatch = new AtomicReference<>(new CountDownLatch(0));
+    private final AtomicReference<CountDownLatch> onDemandLatch = new AtomicReference<>(new CountDownLatch(0));
```
ata-nas, line 61: Interesting, why do we wrap a countdown latch in an atomic reference?

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillPluginTest.java
```diff
@@ -207,60 +519,89 @@ public void handleVerification(VerificationNotification notification) {
...
+            return Map.of(
+                    "backfill.blockNodeSourcesPath", backfillSourcePath,
+                    "backfill.fetchBatchSize", String.valueOf(fetchBatchSize),
+                    "backfill.delayBetweenBatchesMs", String.valueOf(delayBetweenBatchesMs),
+                    "backfill.initialDelayMs", String.valueOf(initialDelayMs),
+                    "backfill.initialRetryDelayMs", String.valueOf(initialRetryDelayMs),
+                    "backfill.maxRetries", String.valueOf(maxRetries),
+                    "backfill.scanIntervalMs", String.valueOf(scanIntervalMs),
+                    "backfill.startBlock", String.valueOf(startBlock),
+                    "backfill.endBlock", String.valueOf(endBlock));
```
ata-nas, line 604: Really cool reusable way to do configs for testing 👍. A builder that produces a map which is used for overrides in the swirlds config api.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -252,6 +254,47 @@ private Future<Long> launchQueueForwarder() {
                 .submit(new MessagingForwarderTask(serverContext, this, queueByBlockMap));
     }
 
+    /***
+     * todo(1236) Temporal method to handle end of stream specifically to test On-Demand backfill.
+     * This should be improved and completed by 1236.
+     */
+    @Override
+    public void handleEndStreamRequest(final PublishStreamRequest.EndStream endStream) {
```
ata-nas, line 262: I do not believe this is correct. The handler must handle the request. It can call the manager if it is needed to do so, for instance to signal that a notification has to be sent or to shut itself down etc. etc.
Let's leave the handle request method in the handler and only call the manager to signal that a notification has to be sent in the case where we need to do that.

ata-nas: P.S. especially raised on API level, this breaks the whole structure.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherManager.java
```diff
@@ -57,6 +58,8 @@ BlockAction getActionForBlock(
      */
     long getLatestBlockNumber();
 
+    void handleEndStreamRequest(PublishStreamRequest.EndStream endStream);
+
```
ata-nas, line 62: This, in my opinion, breaks the whole api. The manager must not handle requests, that is what handlers are for. The manager's functions are different.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherHandlerTest.java
```diff
@@ -153,6 +157,8 @@ class FunctionalityTest {
         /** The handler under test. */
         private PublisherHandler toTest;
 
+        private final BlockMessagingFacility blockMessagingFacility = new TestBlockMessagingFacility();
```
ata-nas, line 160: I do not think any of the changes in this file are used at all? I think this is possibly a forgotten revert?

ata-nas (commented):
More cleanup suggestions.
I do not think that we should move the handleEndStreamRequest method to the manager at all. That should probably remain in the handlers, which should call the manager if needed. At least, that is my opinion.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Yes, I revisit the whole concurrent state variables that we are using among the AUTO and On-DEMAND flows.
we don't have contentious writes to those variables so yeah, when reading by making it volatile, we get the change immediately, so is cheaper and faster for our use case than using an atomic boolean.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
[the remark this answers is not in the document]
jsync-swirlds: Handlers handle all of the interaction with the publisher (each handler is focused on a single publisher).
Manager handles all of the interaction with messaging (There is only one manager and only one messaging facility).
Handlers and Manager make calls to each other when necessary.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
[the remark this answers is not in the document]
jsync-swirlds: I'm just recommending that er have the time calculated with some large-ish multiplier and a divisor obtained from configuration.
That is not a blocking suggestion, it can be done later if needed.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: What @ata-nas said.


In future follow up I believe that the Manager actually can detect the need for notification without any direct interaction from the handler (I was thinking about this last night and realized there is a path for this).
We can keep all handling of EndStream, behind, too far behind, etc... in the Handler and the methods already used to determine next step will signal enough for the manager to work out publishing the notification.

If we pull this back to "Handler handles message, _if necessary_ calls method on manager that takes block number and _only_ sends the notification (nothing more)", we can clean it up later to have the manager detect the necessary conditions without a separate method call.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: This would be a lot simpler to merge forward if the Handler did everything except create and send the notification.
Then calls a method here (perhaps `notifyTooFarBehind(final long newestKnownBlockNumber`) that _just_ creates and sends the notification.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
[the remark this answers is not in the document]
AlfredoG87: Yes, I've added the configuration property. Documentations updates is scheduled for next and final PR.
✅ Done!

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
AlfredoG87: I've pushed a new version that is more in line with what we have discussed on several conversations in this PR.
Resolving all but the last one 😅 

https://github.com/hiero-ledger/hiero-block-node/pull/1444#discussion_r2271176026

Changes are done and ready for review 👍 

✅ Done!

jsync-swirlds (approved):

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/fixtures/TestStreamPublisherManager.java
```diff
@@ -75,6 +75,11 @@ public long getLatestBlockNumber() {
         return latestBlockNumber;
     }
 
+    @Override
+    public void notifyTooFarBehind(long newestKnownBlockNumber) {
+        // Do nothing.
```
ata-nas, line 80: Throwing is better, otherwise false positives might occur. We must be very explicit with fixtures


```suggestion
        throw new UnsupportedOperationException("implement notifyTooFarBehind in fixture class");
```

ata-nas (approved):
Great job @AlfredoG87 🔥

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
ata-nas: Yeah, we can revisit the tasks if you see improvement possibilities. We can open an issue for the future. Most definitely this is a separate peace of work. 👍

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/fixtures/TestStreamPublisherManager.java
ata-nas: P.S. feel free to not do this change, just pointing out.

AlfredoG87: please do add it to your follow up, I agree is better to throw thought. 👍

Issue #1374, opened 2025-07-08
Revisit `SimpleBlockRangeSet` Test Fixture Implementation


ata-nas:
### Story Form

As a Block Node Developer/User
I want to have correct test fixtures
So that I can be confident when using them and to be confident about my tests and production logic

### Technical Notes

- currently the `SimpleBlockRangeSet`'s min/max methods return `0` (zero) when the set is empty.
  - this is erroneous because it does not comply with the ranged set API
- as a result, we have some tests that are erroneous and assert wrong and we also have some tests that produce false positives (we pass the test with a wrong production logic)
  - [example test with wrong assert](https://github.com/hiero-ledger/hiero-block-node/pull/1366#discussion_r2182674514): here the server config returns 0 for min/max when the node is empty which is erroneous
  - [example test with false positive](https://github.com/hiero-ledger/hiero-block-node/pull/1366#discussion_r2182701447): here the production logic allows for a field to be of value -1, but that is not caught in tests because the `SimpleBlockRangeSet` returns 0 instead of -1 for min/max when the set is empty.
- we need to change the behavior of the `SimpleBlockRangeSet` to correctly comply with the ranged set's API
- we need to fix all tests that will break as a result and to modify any production logic that is not correct
- we need to be extra careful here as there could be hidden issues, we need to manually check anywhere this fixture is used, even if nothing breaks when making it correct

ata-nas:
Note: this is partially implemented in #1445. The correct working of the ranged set fixture was critical for the publisher tests. Changes in that PR show tags `@todo(1374)` where actions need to be taken, mostly in the subscriber plugin. There was also added a javadoc with what the team discussed should be the correct actions based on first/last from the request and latest persisted or the lack thereof. Logic should be validated based on that, but again that is tagged with the todo.

PR #1445, opened 2025-07-31
test: Multi-Publisher Plugin Unit Tests

### LINKED ISSUE #1419: Multi-Publisher Plugin Unit Tests

ata-nas:
## Reviewer Notes

* Follow up of interim PR
* Unit tests added:
  * StreamPublisherPlugin:
    * Added test for sending a valid block
  * LiveStreamPublisherManager:
    * Added constructor tests
    * Added tests for latestKnownBlockNumber
    * Added tests for getActionForBlock
    * Added tests for handlePersisted
    * Added placeholder for closeBlock (not yet implemented)
    * Added placeholder for handleVerification (not yet implemented)
* Implementing GH issue #1374 since it is needed in order to correctly assert plugin logic
  * Fixing various tests that assert wrong because of the ranged set test fixture not behaving according to API
* Dependencies:
  * Added Awaitility as dependency (present in consensus node repo, but newest version)
* Test Fixtures improvements
  * Renamed BlockingSerialExecutor to BlockingExecutor and added async abilities to it

## Related Issue(s)

Closes #1419


ata-nas:
Please note:
I decided to not implement #1374 completely because it seems like the subscriber logic is wrong based on cases/conditions described below when determining what items must be streamed:
```text
Cases:
1. Start and end are both negative --> give me live, whatever that is.
2. Start negative, end positive --> client error
3. Start positive, end negative --> Start at N continue forever
4. Start positive, end positive --> Stream from N to M
5. Start > End --> Client Error

If I cannot fulfill the request:
1. First known block is greater than start, and start is positive --> Server error
2. First known block is negative, and start is positive --> Server Error
3. First known block is negative, and start is negative --> Stream live if available, otherwise Server Error
```
It seems that the wrongful behavior of the ranged set test fixture has produced a lot of false positives.

This PR is quite big on it's own. I have however changed the behavior of the ranged set fixture to be correct based on the API (partially resolves #1374) because I need it to correctly test the publisher logic, otherwise more than 10 tests will fail, but also to mitigate the potential for other false positives and/or wrongful asserts.

Because of the complexity of the cases above and that it seems like the failing tests assert wrong, and also the size and current importance of this unit tests PR (it will unblock other's progress), I've decided to address the rest of #1374 as a follow up.

block-node/stream-publisher/build.gradle.kts
```diff
@@ -20,4 +20,5 @@ testModuleInfo {
     requires("org.junit.jupiter.params")
     requires("org.assertj.core")
     requires("org.hiero.block.node.app.test.fixtures")
+    requires("awaitility")
```
jsync-swirlds, line 23: This appears to be due to an IDE insertion, and should be removed.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/StreamPublisherPluginTest.java
```diff
@@ -2,6 +2,7 @@
 package org.hiero.block.node.stream.publisher;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
```
jsync-swirlds, line 5: This is a utility we haven't discussed or added to reviewed dependencies.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/StreamPublisherPluginTest.java
```diff
@@ -171,36 +175,30 @@ void testPublishUnsetOneOf() {
...
-                    .isNotNull()
-                    .returns(ResponseOneOfType.ACKNOWLEDGEMENT, responseKindExtractor);
+            await().atMost(2, TimeUnit.SECONDS)
+                    .pollInterval(100, TimeUnit.MILLISECONDS)
+                    .untilAsserted(() -> assertThat(fromPluginBytes)
+                            .hasSize(1)
+                            .first()
+                            .extracting(bytesToPublishStreamResponseMapper)
+                            .isNotNull()
+                            .returns(ResponseOneOfType.ACKNOWLEDGEMENT, responseKindExtractor));
```
jsync-swirlds, line 201: We should work out a different approach here; a simple `parkNanos` in a loop is probably ample.

jsync-swirlds (commented):
Mostly looks good.
One added dependency that needs more discussion.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/StreamPublisherPluginTest.java
ata-nas: I saw it being used in the consensus node repo, so I figured we could utilize this. I've removed this however in favor of a simpler solution. It would be good to add this as it is quite useful.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/StreamPublisherPluginTest.java
ata-nas: I've opted for a simple sleep.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingExecutor.java
```diff
@@ -0,0 +1,310 @@
...
+                } catch (final TimeoutException e) {
+                    // this is not expected to happen, if we throw here, this means
+                    // that either the timeout is too low, or the task is
+                    // potentially blocking forever.
+                    final String message =
+                            "A task that executed in the %s took more than the allowed time of %d seconds. It is possible that the task is blocking forever."
+                                    .formatted(getClass().getName(), RUNNABLE_FUTURE_MAX_TIMEOUT_SECONDS);
+                    throw new RuntimeException(message, e);
+                }
+            });
```
ata-nas, line 95: I've noticed an issue today with the blocking executor.

The `submit()` methods in the super, which is a `ThreadPoolExecutor`, wrap any submitted tasks in a `FutureTask`. When running this task (which is both a `Runnable` and a `Future`), all exceptions are captured so that they can be `queried` via the `Future`'s API. This means that when running these tasks with the `executeAsync()` methods, exceptions will always be effectively silenced. To fix this, I am now wrapping any `RunnableFuture`s in a custom `Runnable` that will simply `run` and `get` the actual `RunnableFuture` which will ensure that exceptions will be thrown later (which is something that we want). Otherwise false positives will inevitably happen.

I've thought about making a custom `RunnableFuture`, and overriding the `submit()` methods of super, but that is way too complex and error prone. I've added a hard timeout however in order to ensure that tasks do not run indefinitely during tests.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingExecutor.java
```diff
@@ -0,0 +1,260 @@
...
+     * @return List of {@link CompletableFuture} objects, each representing a
+     * submitted task. The list will contain the same number of futures as the
+     * number of tasks submitted to the executor.
+     */
+    public List<CompletableFuture<Void>> executeAsync(
+            final boolean blockUntilDone, final long blockTimeoutMillis, final boolean throwOnExceptionalCompletion) {
+        if (blockTimeoutMillis <= 0) {
+            throw new IllegalArgumentException("Timeout per task must be greater than 0");
+        } else {
+            final List<CompletableFuture<Void>> futures = new ArrayList<>();
```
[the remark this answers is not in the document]
jsync-swirlds: What you're doing here is wrapping every item in a completion stage, and asynchronously running all of those with individual timeouts; and adding each to a list.
Then you gather them into an array, and create _another_ completable future, then call join on the one future.
That merges exceptions (of which you catch too many; should only catch ExecutionException) and only really throws one (if any).

Instead you could have something like this:
for each item in the work queue
    submit that item to the completion service
    increment submitted count
done
if block until done
    remaining count = submitted count
    while remaining count > 0 and global timeout not reached
        poll next item from completion service
        get result, add to result list
          OR (if completed exceptionally)
        get exception, add to exception list
        decrement remaining count
    done
    return results and exceptions
done

Note, this is simpler (less try/catch, less wrapping, cleaner stack traces, etc...) and also allows for returning results (if desired), which the current approach forecloses (because `CompletableFuture.allOf` returns a `CompletableFuture<Void>`).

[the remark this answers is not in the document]
ata-nas: About returning the results, this is not entirely true. Although you are right that `allOf` will not reflect the results themselves, remember, we've already collected all completable futures in a list, these are mutable and when we return them, they will be in a state which reflects the actual result. This will not be the case if we do not block until done, but then the caller can use these completable futures to see if they are done, get an immediate result, etc. etc.

I see no added value to the completion service vs `allOf`, so I think your first proposal with `allOf` is sufficient and actually simpler to reason and less error prone.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/StreamPublisherPluginTest.java
jsync-swirlds: Consensus has a lot of dependencies they need to remove, eventually.
We've tried to be careful to limit adding dependencies to block node so we keep our supply chain small.
Please bring this dependency up in the next design meeting, or on channel, so the team can discuss whether it's worth adding it to the system.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingExecutor.java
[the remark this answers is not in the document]
jsync-swirlds: I think the "catch everything and wrap it in runtime exception" is probably the primary pattern that bothers me about this code.
I'm very strongly opposed to that pattern, as it's really saying "something is wrong, so hide it and hope something else fails instead".
I know this is test code, but it still seems awkward to have a "catch everything, then turn it into a runtime exception, maybe" process that has a tendency to hide all sources of exceptions and result in a test failure with an unusable stack trace (the wrapping, particularly if it crosses threads, tends to make unit test frameworks never print the real source of the exception).

The ideal situation is to _never_ wrap an exception; if we need to, at the least we should try not to do so any more than necessary (_particularly_ in unit tests where we might not get full detail).

jsync-swirlds (approved):
I still believe the test executor introduced here needs additional work, but that can be further developed in a future PR.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/BlockingExecutor.java
ata-nas: Unfortunately `Runnable` does not throw checked exceptions, otherwise we would not have thrown. Not calling "get" effectively silences any exceptions that might have been thrown when running, as the `FutureTask` is both a `Runnable` and a `Future`. Exceptions from execution are only visible when using the `Future`'s API.

PR #1451, opened 2025-08-01
test: add test for DUPLICATE_BLOCK status with starting block number 0

### LINKED ISSUE #1449: bug(publisher): BN responds with BEHIND instead of Duplicate Block.
### Description

At another test the BN was already with 23 blocks, and CN attempted to send block 0.
BN responded with BEHIND, when we should have expected to respond with Duplicate Block.

### Steps to Reproduce

Start a BN and stream some blocks, lets say 100.
Then start a CN and begin a new stream from block 0.

CN receives BEHIND. Expected result is CN receives `EndOfStream` for `DUPLICATE` block with last known block number `100`.

mustafauzunn:
## Reviewer Notes

>At another test the BN was already with 23 blocks, and CN attempted to send block 0.
>BN responded with BEHIND, when we should have expected to respond with Duplicate Block.

Verifying that the bug is not reproducible with a test.

## Related Issue(s)
Closes #1449


AlfredoG87:
@mustafauzunn  can you comment on this ticket so I can assign it to you please?

AlfredoG87 (approved):
LGTM. this helped me identify the real root cause of the error they are experiencing on the CN side.

Thanks!

ata-nas (approved):

PR #1462, opened 2025-08-06
fix: Setting property generator.endBlockNumber is not stopping simulator

### LINKED ISSUE #1459: Simulator(CraftMode): Setting property `generator.endBlockNumber` is not stopping simulator
### Description

Setting property `generator.endBlockNumber` is not stopping simulator from continuing to stream blocks undefinitively

### Steps to Reproduce

Set property to 100, start the simulator and wait and see that the simulator goes over an keeps streaming indefinitively.

### Additional context

_No response_

mustafauzunn:
## Reviewer Notes

Setting the property `generator.endBlockNumber` prevents the simulator from continuing to stream blocks indefinitely.

## Related Issue(s)
Fixes #1459


ata-nas:
> > LG. Do we have a unit test for this?
> 
> I was wondering when to include tests, now or with another PR. I don't like that everything is mocked in those tests.
> 
> However I added the test with mocks for consistency for the PR, and we can refactor other tests in the future. I don't want to change the scope of the task.

Usually, when we make a new PR, we should aim to add all relevant tests for it in the same PR. Some exceptions apply. But that is the general case.

PR #1464, opened 2025-08-06
feat:  Simulator add publisher support for ResendBlock message

### LINKED ISSUE #840: Simulator add publisher support for `ResendBlock` message
### Story Form

As a block node engineer
I want to test `ResendBlock` handling

### Technical Notes

Add support in the block stream simulator, publisher mode, for receiving a `ResendBlock` message from the block node and correctly resending that block then streaming following blocks after that as normal.

mustafauzunn:
## Reviewer Notes

Add support in the block stream simulator, publisher mode, for receiving a `ResendBlock` message from the block node and correctly resending that block then streaming following blocks after that as normal.

## Related Issue(s)
Fixes #840


AlfredoG87 (approved):
LGTM

PR #1465, opened 2025-08-07
feat: Multi-Publisher Responds with `BAD_BLOCK_PROOF` if Verification Fails

### LINKED ISSUE #1422: Multi-Publisher Responds with `BAD_BLOCK_PROOF` if Verification Fails
### Story Form

As a Block Node Publisher
I want to receive a `BAD_BLOCK_PROOF` response when verification for the block I just sent fails
So that I can handle that on my side

### Technical Notes

- we must handle failed verifications in the Multi-Publisher Plugin
    - Handler must track the last block it completed
    - Manager calls a method on each handler for "bad block" with failed block number.
    - The publisher that completed that block sends `EndOfStream` for `BAD_BLOCK_PROOF`.
    - All other publishers must receive `Resend`, specifying the last known and persisted block (from the manager query).

ata-nas:
## Reviewer Notes

* Added wrapping logic for calling onNext on the response pipeline
* Added a way to track and determine if a given handler has streamed a given block number
* Added logic to send BAD_BLOCK_PROOF
* Added logic to flush queues in case of a failed verification
* Added descriptive documentation
* Added tests:
  * LiveStreamPublisherManagerTest:
    * Added tests for no activity if handle condition does not pass
    * Added tests for BAD_BLOCK_PROOF sent to publisher that published the failed block
    * Added tests for RESEND sent to publishers that did not publish the failed block
    * Added tests for continued normal operation after a failed block proof
  * PublisherHandlerTest:
    * Added tests for BAD_BLOCK_PROOF when publisher has streamed the failed block
    * Added tests for RESEND when publisher has not streamed the failed block

## Related Issue(s)

Closes #1422

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -257,11 +257,65 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         return lastPersistedBlockNumber.get();
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * This method handles verification notifications from the block messaging
+     * facility. It checks if the notification is from a publisher and if the
+     * verification was unsuccessful. If so, it forks each handler and calls
+     * the {@link PublisherHandler#handleFailedVerification(long)} on each.
+     * After each invocation, it clears the transfer queue for that handler
+     * if it exists.
+     * <blockquote>
+     * NOTE: The handler that was responsible for sending the block that failed
+     * verification will proceed to send an {@link PublishStreamResponse.EndOfStream}
+     * with {@link PublishStreamResponse.EndOfStream.Code#BAD_BLOCK_PROOF}. The
+     * handler will then shut down. All other handlers will send
+     * a {@link PublishStreamResponse.ResendBlock}, if the resend was
+     * successfully sent, then the handlers will continue to operate normally.
+     * Otherwise, the handlers will also shut down.
+     * <br>
+     * After we exit from the fork, we will clear all transfer queues.
+     * Naturally, handlers that have shut down will not have the queue
+     * available, i.e. it will be {@code null}.
+     * </blockquote>
+     * We will handle only if the verification has failed, notification block
+     * number is greater than the {@link #lastPersistedBlockNumber}
+     * (which has also passed verification naturally), and less than the
+     * {@link #nextUnstreamedBlockNumber} Blocks will not be known unless they have
+     * passed verification and have been persisted. There is always expected
+     * to be a gap between the {@link #lastPersistedBlockNumber} and the
+     * {@link #nextUnstreamedBlockNumber} because a block cannot start
+     * verification until streamed in full, and the
+     * {@link #nextUnstreamedBlockNumber} is incremented at the moment we start
+     * * streaming the block.
+     * <br>
+     * In practice, if the {@link #lastPersistedBlockNumber} is 100, the
+     * {@link #nextUnstreamedBlockNumber} will be at least 101. When we start
+     * streaming block 101, the {@link #nextUnstreamedBlockNumber} will be
+     * incremented to 102, and the {@link #lastPersistedBlockNumber} will be
+     * 100. If the block 101 fails verification, it will be captured by
+     * (block greater than 100 and block lower than 102).
+     */
     @Override
     public void handleVerification(@NonNull final VerificationNotification notification) {
-        // Need to check, but should only handle the "failed" case.
-        // on success we should probably do nothing.
-        // @todo(1422) implement
+        final long blockNumber = notification.blockNumber();
+        final boolean shouldHandle = !notification.success()
+                && blockNumber > lastPersistedBlockNumber.get()
+                && blockNumber < nextUnstreamedBlockNumber.get();
+        if (shouldHandle) {
+            nextUnstreamedBlockNumber.set(blockNumber);
+            currentStreamingBlockNumber.set(blockNumber);
+            handlers.values().parallelStream().unordered().forEach(handler -> {
+                final long handlerId = handler.handleFailedVerification(blockNumber);
+                final String qId = getQueueNameForHandlerId(handlerId);
+                final BlockingQueue<BlockItemSetUnparsed> queue = transferQueueMap.get(qId);
+                if (queue != null) {
+                    // If a queue exists for the handler, we need to flush it
+                    queue.clear();
+                }
+            });
+        }
     }
 
     /**
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.stream.publisher;
 
+import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.INFO;
 import static org.hiero.block.node.spi.BlockNodePlugin.METRICS_CATEGORY;
 import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
@@ -15,8 +16,10 @@ import com.swirlds.metrics.api.Metrics;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.lang.System.Logger;
 import java.util.List;
+import java.util.NavigableSet;
 import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.Flow.Subscription;
 import java.util.concurrent.atomic.AtomicLong;
 import org.hiero.block.api.PublishStreamResponse;
@@ -63,8 +66,13 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     private final long handlerId;
     /** The current streaming block number. This is used to track the current block being streamed. */
     private final AtomicLong currentStreamingBlockNumber;
-    /** The current block action. This is tracked to help the manager determine
-     * what to do with the current block. */
+    /** The unacknowledged yet blocks that were streamed to completion by this handler. */
+    private final NavigableSet<Long> unacknowledgedStreamedBlocks;
+    /**
+     * The current block action.
+     * This is tracked to help the manager determine what to do with the current
+     * block.
+     */
     private BlockAction blockAction;
 
     /**
@@ -88,6 +96,7 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
         publisherManager = Objects.requireNonNull(manager);
         blockItemsQueue = Objects.requireNonNull(transferQueue);
         currentStreamingBlockNumber = new AtomicLong(UNKNOWN_BLOCK_NUMBER);
+        unacknowledgedStreamedBlocks = new ConcurrentSkipListSet<>();
     }
 
     // ==== Flow Methods =======================================================
@@ -245,16 +254,25 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
                     case END_DUPLICATE -> handleEndDuplicate();
                     case END_ERROR -> handleEndError();
                 };
+        handleBlockActionResult(handleResult);
+        // we must now check if the last item in the set is a proof.
+        final BlockItemUnparsed last = blockItems.getLast();
+        if (last.hasBlockProof()) {
+            handleBlockProof(last.blockProof(), blockNumber);
+        }
+    }
+
+    /**
+     * This method handles the result of a block action handle.
+     *
+     * @param handleResult the result to handle
+     */
+    private void handleBlockActionResult(final BatchHandleResult handleResult) {
         if (handleResult.shouldReset()) {
             resetState();
         }
         if (handleResult.shouldShutdown()) {
             shutdown();
-        }
-        // we must now check if the last item in the set is a proof.
-        final BlockItemUnparsed last = blockItems.getLast();
-        if (last.hasBlockProof()) {
-            handleBlockProof(last.blockProof());
         }
     }
 
@@ -272,14 +290,18 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
      * initiated.
      *
      * @param blockProofBytes the bytes of the block proof to handle
+     * @param blockNumber the block number that this proof is for
      */
-    private void handleBlockProof(final Bytes blockProofBytes) {
+    private void handleBlockProof(final Bytes blockProofBytes, final long blockNumber) {
         try {
             publisherManager.closeBlock(BlockProof.PROTOBUF.parse(blockProofBytes), handlerId);
-        } catch (ParseException e) {
+        } catch (final ParseException e) {
+            // If parsing failed here, items will still be propagated as if
+            // the block is complete. Verification later might succeed.
             publisherManager.closeBlock(null, handlerId);
             LOGGER.log(INFO, "Failed to parse block proof: {}", e.getMessage());
         }
+        unacknowledgedStreamedBlocks.add(blockNumber);
         resetState();
     }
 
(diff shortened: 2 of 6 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -259,9 +261,22 @@ public long getLatestBlockNumber() {
...
+            final long blockNumber = notification.blockNumber();
+            // use a copy of the handlers to avoid concurrent modification exceptions
+            // even though we are using concurrent maps/collections.
+            Set.copyOf(handlers.values()).parallelStream().unordered().forEach(handler -> {
+                final long handlerId = handler.handleFailedVerification(blockNumber);
+                final String qId = getQueueNameForHandlerId(handlerId);
+                final BlockingQueue<BlockItemSetUnparsed> q = transferQueueMap.get(qId);
+                if (q != null) {
+                    // If a queue exists for the handler, we need to flush it
+                    q.clear();
```
ata-nas, line 274: Is this clear sufficient? In theory, a handler, which's queue we've cleared, can proceed to publish more items. Maybe the solution here would be to update next and current streaming atomic longs before we start this fork?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -272,14 +288,18 @@ private void handleBlockItemsRequest(
...
         try {
             publisherManager.closeBlock(BlockProof.PROTOBUF.parse(blockProofBytes), handlerId);
-        } catch (ParseException e) {
+        } catch (final ParseException e) {
+            // If parsing failed here, items will still be propagated as if
+            // the block is complete. Verification later might succeed.
             publisherManager.closeBlock(null, handlerId);
             LOGGER.log(INFO, "Failed to parse block proof: {}", e.getMessage());
         }
+        unacknowledgedStreamedBlocks.add(blockNumber);
```
ata-nas, line 302: Update that this handler has just finished supplying the block with number X.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -316,9 +346,58 @@ private void sendEndOfStream(final EndOfStream.Code codeToSend) {
...
+    private boolean sendResponse(final PublishStreamResponse response) {
+        try {
+            replies.onNext(response);
+            return true;
+        } catch (final Exception e) {
+            final String message = "Failed to send response: %s".formatted(e.getMessage());
+            LOGGER.log(DEBUG, message, e);
+            return false;
+        }
+    }
```
ata-nas, line 374: I am wrapping all invocations of the `onNext` method in a try catch. They might fail for various reasons. We return a boolean result if the request was a success. The caller can act based on that.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -316,9 +346,58 @@ private void sendEndOfStream(final EndOfStream.Code codeToSend) {
...
+     * If this handler was the one that streamed the block, we will attempt to
+     * send an {@link EndOfStream} with a {@link Code#BAD_BLOCK_PROOF} and
+     * proceed to shut down the handler.
+     * Otherwise, we will send a {@link ResendBlock} response to the publisher.
+     *
+     * @param blockNumber of the block that failed verification
+     * @return the id of this handler
+     */
+    public long handleFailedVerification(final long blockNumber) {
+        if (unacknowledgedStreamedBlocks.remove(blockNumber)) {
```
ata-nas, line 387: Send `BAD_BLOCK_PROOF` if the set of sent unacknowledged items contains the failed block number. This means that this handler was responsible for streaming the failed block. Otherwise, we send `RESEND`.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -305,8 +325,18 @@ public void sendAcknowledgement(final long newLastAcknowledgedBlockNumber) {
...
                 PublishStreamResponse.newBuilder().acknowledgement(ack).build();
-        replies.onNext(response);
-        metrics.blockAcknowledgementsSent.increment(); // @todo(1415) add label
+        if (sendResponse(response)) {
+            // if response was sent successfully, we can remove
+            // all unacknowledged blocks that are less than or equal to the
+            // new last acknowledged block number.
+            unacknowledgedStreamedBlocks
+                    .headSet(newLastAcknowledgedBlockNumber, true)
+                    .clear();
```
ata-nas, line 334: The skiplist set is ordered. We can use the headSet and then clear to neatly remove all blocks that have been acknowledged. There could be multiple blocks acknowledged which were sent by this handler.

ata-nas (commented):
Self-review. Aimed to help reviewers.

My main concerns are the race conditions and if what we currently have is sufficient. We can also utilize protocol level agreements to help us solve situations that seem tricky.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I agree that the update to the block numbers should be before we clear the queues.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -259,9 +261,22 @@ public long getLatestBlockNumber() {
 
     @Override
     public void handleVerification(@NonNull final VerificationNotification notification) {
-        // Need to check, but should only handle the "failed" case.
-        // on success we should probably do nothing.
-        // @todo(1422) implement
+        if (BlockSource.PUBLISHER == notification.source() && !notification.success()) {
```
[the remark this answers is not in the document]
jsync-swirlds: I think we need this to be somewhat different.  I don't think the source matters.

What we want to check is whether the failed verification is for a block number _strictly greater than_ the last acknowledged block.  If it _is_ then we need to respond/react.  If it _is not_ then this is a backfill issue (because we already validated and acknowledged this or a later block, so backfill must be involved).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -63,8 +65,12 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     private final long handlerId;
     /** The current streaming block number. This is used to track the current block being streamed. */
     private final AtomicLong currentStreamingBlockNumber;
-    /** The current block action. This is tracked to help the manager determine
-     * what to do with the current block. */
+    /** The unacknowledged yet blocks that were streamed to completion by this handler. */
+    private final ConcurrentSkipListSet<Long> unacknowledgedStreamedBlocks;
```
jsync-swirlds, line 69: I don't think we want, or need, to track this.
It's extra complexity that creates new opportunities to get stuck somewhere.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: I think I see why we might need to track block numbers now.
That is... not ideal.

I wonder if we can manage this better...

I suppose this is OK for now, but it would be good to find an improvement.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -257,11 +259,47 @@ public long getLatestBlockNumber() {
...
+            final long blockNumber = notification.blockNumber();
+            // use a copy of the handlers to avoid concurrent modification exceptions
+            // even though we are using concurrent maps/collections.
+            Set.copyOf(handlers.values()).parallelStream().unordered().forEach(handler -> {
+                final long handlerId = handler.handleFailedVerification(blockNumber);
+                final String qId = getQueueNameForHandlerId(handlerId);
+                final BlockingQueue<BlockItemSetUnparsed> q = transferQueueMap.get(qId);
```
jsync-swirlds, line 294: Can we spell out `queue` here?  Single letter names (and their extensions) are not good for readability.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -63,8 +65,12 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
...
+    /**
+     * The current block action. This is tracked to help the manager determine
+     * what to do with the current block.
```
jsync-swirlds, line 72: Always prefer to have the first sentence end with a newline for Javadoc.
```suggestion
     * The current block action.
     * This is tracked to help the manager determine what to do with the current block.
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds, line 69: Might be good to use `Set` or `NavigableSet` as the variable type.

jsync-swirlds (approved):
Just a few small concerns.  Nothing blocking.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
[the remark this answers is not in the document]
ata-nas: Interesting. So source is something we do not care about at all? I am not sure. I think we've spoken about an edge case where we might be backfilling forward (that sounds kind of impossible)? Not sure about the details there. That's why I included the source. But agreed that we do need to add a number strictly greater than the last ackd.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
ata-nas: +1 on catching exceptions. We should think about this a bit more. Let's proceed for now.

georgi-l95 (approved):

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: I think a shutdown in the catch is fine.  Shutdown should be idempotent and accept repeated calls without error.
Adding the boolean return makes it more likely we miss the need to call shutdown (which has happened once already).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: A queue may have the full content of a block that should stream in the future, and still return a SKIP or other response to the _next_ block from that handler.
We really need to be careful to only clear a queue when we _know_ any prior content is invalid (e.g. when we have a bad block verification).

I cannot think of another case where clearing all queues is correct off hand.

As far as clearing queues getting out of order, I don't think that's reasonably possible, as we would be sending resend messages, which takes about three orders of magnitude longer than this method will take to complete (even with very strange thread scheduling).

We could add a lock, but that's incredibly expensive.  We use concurrent queues precisely to avoid needing that.

If we are truly concerned, we could block the action response until this method completes.
That's not great, but might be acceptable.

jsync-swirlds: Note, it is _entirely acceptable_ for the block node to receive the same block multiple times.  We make some effort to avoid that, but the block node must not fail just because it received a duplicate block.  Also, a subscriber should handle multiple copies of the same block; there is enough external asynchrony and network delays to make that a possibility more-or-less regardless of our efforts to prevent it.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -257,11 +259,50 @@ public long getLatestBlockNumber() {
...
-        // Need to check, but should only handle the "failed" case.
-        // on success we should probably do nothing.
-        // @todo(1422) implement
+        final long blockNumber = notification.blockNumber();
+        final boolean shouldHandle = BlockSource.PUBLISHER == notification.source()
+                && !notification.success()
+                && blockNumber > getLatestBlockNumber();
+        if (shouldHandle) {
+            // use a copy of the handlers to avoid concurrent modification exceptions
+            // even though we are using concurrent maps/collections.
```
jsync-swirlds, line 293: We should not need to copy these.  ConcurrentSkipListMap won't throw concurrent modification exceptions...

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -292,7 +333,9 @@ public void handleVerification(@NonNull final VerificationNotification notificat
     public void handlePersisted(@NonNull final PersistedNotification notification) {
         final long newLastPersistedBlock = notification.endBlockNumber();
         if (newLastPersistedBlock > lastPersistedBlockNumber.get()) {
-            handlers.values().parallelStream().unordered().forEach(handler -> {
+            // use a copy of the handlers to avoid concurrent modification exceptions
+            // even though we are using concurrent maps/collections.
```
jsync-swirlds, line 337: Again, I do not see any reason to copy these.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
ata-nas: The idempotency argument for the shutdown method is very valid. The method is indeed idempotent and should remain like so. I will keep the boolean however because it indicates if other actions specific for the caller must be made. Things like metrics and such.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: Yeah, even though we do use concurrent collections, I still thought this should be beneficial. I will revert.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: > As far as clearing queues getting out of order, I don't think that's reasonably possible, as we would be sending resend messages, which takes about three orders of magnitude longer than this method will take to complete (even with very strange thread scheduling).
> 
> We could add a lock, but that's incredibly expensive. We use concurrent queues precisely to avoid needing that.
> 
> If we are truly concerned, we could block the action response until this method completes.
> That's not great, but might be acceptable.

I decided to open #1478 so we do not lose this. Let's collect some empirical data from running the node and then we can validate these concerns. That sounds reasonable to me. In the meantime, I agree with you that we should clear the numbers and then proceed to call the handlers. 🙌

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -259,9 +261,22 @@ public long getLatestBlockNumber() {
...
+                final long handlerId = handler.handleFailedVerification(blockNumber);
+                final String qId = getQueueNameForHandlerId(handlerId);
+                final BlockingQueue<BlockItemSetUnparsed> q = transferQueueMap.get(qId);
+                if (q != null) {
+                    // If a queue exists for the handler, we need to flush it
+                    q.clear();
+                }
+            });
+            nextUnstreamedBlockNumber.set(blockNumber);
+            currentStreamingBlockNumber.set(blockNumber);
```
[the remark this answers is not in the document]
ata-nas: See [here](https://github.com/hiero-ledger/hiero-block-node/pull/1465#discussion_r2266022784).

georgi-l95 (approved):
LG

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -257,11 +259,50 @@ public long getLatestBlockNumber() {
...
+     * verification will proceed to send an {@link PublishStreamResponse.EndOfStream}
+     * with {@link PublishStreamResponse.EndOfStream.Code#BAD_BLOCK_PROOF}. The
+     * handler will then shut down. All other handler will send
+     * a {@link PublishStreamResponse.ResendBlock}, if the resend was
+     * successfully sent, then the handler will continue to operate normally.
+     * Otherwise, the handler will also shut down.
+     * <br>
+     * After we exit from the fork, we will clear all transfer queues.
+     * Naturally, handlers that have shut down will not have the queue
+     * available, i.e. it will be {@code null}.
```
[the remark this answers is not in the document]
jsync-swirlds: Here is my thought (and "ownership" is exactly the way to look at this).

I see it as Handlers "own" the queue, and the manager just "reads" from the queue.

It could be seen as the reverse, though.  Both views are reasonable.

jsync-swirlds (approved):

PR #1469, opened 2025-08-07
fix: Change the incorrect shutdown of a possibly shared executor service in the stream subscriber plugin.

### LINKED ISSUE #1439: `SubscribeBlockStreamHandler` Shuts Down a Shared Thread Pool
#### What Happened 
- The `SubscribeBlockStreamHandler` uses a thread pool which it gets from the `ThreadPoolManager`, see [L148](https://github.com/hiero-ledger/hiero-block-node/blob/95f67d65b398ae70f23677c246e840b4aefc18b6/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java#L148).
  - This pool is a common one and could be reused by all users.
  - If any user shuts down this common pool, this will affect all other current and future users.
  - The handler needs to ensure that currently active futures get cancelled.
- The plugin then shuts down this pool, see [L161](https://github.com/hiero-ledger/hiero-block-node/blob/95f67d65b398ae70f23677c246e840b4aefc18b6/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java#L161)

#### What Should Happen
- No plugin should ever shutdown any common (reused) thread pools.
- (?) No plugin/component must ever manage the lifecycle of pools, this should be done by the `ThreadPoolManager` and users should notify when they want to stop using a given pool (the user keeps a reference to the given pool while using it), the manager then decides what happens?

### Steps to Reproduce

- This behavior is always happening when the subscriber plugin shuts down (the `stop` method gets called).

### Suggestions

- One way to tackle this is to ensure that we are getting a dedicated thread pool from the manager.
  - We have this option, but we should think about if we should hand such control to end users of the manager, does that defeat the purpose of the manager?
  - Would it be better to be able to get a pool (the user is not bothered if the pool is dedicated or not) and when the user desires to shutdown the pool, it simply notifies the thread pool manager that the given pool will no longer be used, the pool manager can then decide what to do.
- Another way to tackle this is to collect all futures of tasks submitted to the pool and then to manually cancel them.
  - If our pool is dedicated however, we should shut it down and drop the reference, it might as well be easier to just let the pool manager handle this also.

jsync-swirlds:
## Reviewer Notes
Modified the stop behavior to not shut down the executor service
* Instead of shutting down the executor, we set the completion service null
    * Also set the open sessions null
    * We use a local reference to shut down all the tasks
* Handle possibly missing (shutting down) executor service by responding to client with "not available".

## Related Issue(s)
 Resolves #1439.

### DIFF
--- a/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
+++ b/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
@@ -16,6 +16,7 @@ import java.lang.System.Logger.Level;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletionService;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
@@ -26,7 +27,9 @@ import java.util.concurrent.atomic.AtomicLong;
 import org.hiero.block.api.BlockStreamSubscribeServiceInterface;
 import org.hiero.block.api.SubscribeStreamRequest;
 import org.hiero.block.api.SubscribeStreamResponse;
+import org.hiero.block.api.SubscribeStreamResponse.Code;
 import org.hiero.block.internal.SubscribeStreamResponseUnparsed;
+import org.hiero.block.internal.SubscribeStreamResponseUnparsed.Builder;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
 import org.hiero.block.node.spi.ServiceBuilder;
@@ -132,7 +135,7 @@ public class SubscriberServicePlugin implements BlockNodePlugin, BlockStreamSubs
         /** A context that applies to the pipeline this handler supports. */
         private final BlockNodeContext context;
         /** Set of open client sessions */
-        private final Map<Long, BlockStreamSubscriberSession> openSessions;
+        private volatile Map<Long, BlockStreamSubscriberSession> openSessions;
         // Metrics
         /** Counter for errors while streaming to subscribers */
         private final Counter subscriberErrorsCounter;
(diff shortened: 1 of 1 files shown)


block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -158,17 +161,20 @@ private SubscribeBlockStreamHandler(@NonNull final BlockNodeContext context) {
...
                 session.close(SubscribeStreamResponse.Code.SUCCESS);
             }
             // Make sure all the threads complete.
-            while (!openSessions.isEmpty()) {
+            while (!closeableSessions.isEmpty()) {
                 try {
                     // This blocks until the session thread ends, but the close
                     // calls above _should have_ ended all the threads already.
-                    openSessions.remove(streamSessions.take().get().clientId());
+                    closeableSessions.remove(sessionsToClose.take().get().clientId());
```
ata-nas, line 177: Would it be good idea if we take with timeout?

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
```diff
@@ -208,7 +238,7 @@ private void handleCompletedStream(final Future<BlockStreamSubscriberSession> co
                 BlockStreamSubscriberSession completedSession = completedSessionFuture.get();
                 long clientId = completedSession.clientId();
                 // Remove the completed session from open sessions.
-                openSessions.remove(clientId);
+                if (openSessions != null) openSessions.remove(clientId);
```
ata-nas, line 241: Is there a possibility that this check-then-act condition fails due to an async drop of this reference?

ata-nas (approved):
@jsync-swirlds LGTM. A couple of non-blocking questions. 🔥

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberServicePlugin.java
jsync-swirlds: Possible, but unlikely.
I'll change these references to volatile (or possibly atomic references) to limit that probability.

ata-nas: I think atomic refs would be the way to go here.

jsync-swirlds: Looking through it volatile was the best choice, as I had to assign to local variables either way.

ata-nas: Ah, yes! Assigning to local variables ensures that the local variables will be valid for the duration of the method call. Indeed, volatile is the way to go here. 🔥

ata-nas (dismissed):
🔥

AlfredoG87 (approved):
LGTM

PR #1471, opened 2025-08-08
test(messaging): Fix broken test on main

### LINKED ISSUE #1470: test(messaging): testBlockItemHandlerException is failing since added more logs
### Description

Seems that by luck when the PR for adding more logs was added the test did not fail.

Since then it has always failed on me until I changed the LEVEL of logger for the test, so we only care about INFO and above logs.

### Steps to Reproduce

just run testBlockItemHandlerException test on main

### Additional context

_No response_

AlfredoG87:
## Reviewer Notes

This test did not really broke, and it passed on the PR that broke it, since sometimes it does pass, but most of the time, fails.

By Setting the logger to INFO, we ignore the TRACE level logs and the assertions are able to find the logs that they are expecting to see when an exception happens.

## Related Issue(s)
Fixes #1470


ata-nas (approved):
It seems that we rely on log checking for tests? That should probably not be the case. Interesting.

PR #1473, opened 2025-08-08
feat: Simulator add publisher support for SkipBlock message 

### LINKED ISSUE #839: Simulator add publisher support for `SkipBlock` message
### Story Form

As a block node engineer
I want to test `SkipBlock` handling

### Technical Notes

Add support in the block stream simulator, publisher mode, for receiving a `SkipBlock` message from the block node and correctly skipping that block (with a delay that is random within a configured range), but still awaiting the `Acknowledgement` message for that block, while streaming the next.

mustafauzunn:
## Reviewer Notes

Add support in the block stream simulator, publisher mode, for receiving a SkipBlock message from the block node and correctly skipping that block.

## Related Issue(s)
Fixes #839

### DIFF
--- a/tools-and-tests/simulator/src/test/java/org/hiero/block/simulator/grpc/impl/PublishStreamGrpcClientImplTest.java
+++ b/tools-and-tests/simulator/src/test/java/org/hiero/block/simulator/grpc/impl/PublishStreamGrpcClientImplTest.java
@@ -309,7 +309,7 @@ class PublishStreamGrpcClientImplTest {
 
         assertTrue(
                 publishStreamGrpcClient.getLastKnownStatuses().getFirst().contains("skip_block"),
-                "lastKnownStatuses should contain the resend block message");
+                "lastKnownStatuses should contain the skip block message");
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamGrpcClientImpl.java
```diff
@@ -117,8 +117,9 @@ public boolean streamBlock(
                 ChunkUtils.chunkify(block.getItemsList(), blockStreamConfig.blockItemsBatchSize());
         for (List<BlockItem> streamingBatch : streamingBatches) {
             if (publishStreamObserver.getPublishStreamResponse() != null
-                    && publishStreamObserver.getPublishStreamResponse().hasResendBlock()) {
-                LOGGER.log(DEBUG, "Received resend block request, stopping streaming for this block");
+                    && (publishStreamObserver.getPublishStreamResponse().hasResendBlock()
+                            || publishStreamObserver.getPublishStreamResponse().hasSkipBlock())) {
+                LOGGER.log(DEBUG, "Received resend block or skip block request, stopping streaming for this block");
```
jsync-swirlds, line 122: Do we handle the difference in response for resend and skip?
Specifically, do we set the next block to send back to the resend requested block in the resend case, and just the next block in the skip case?

jsync-swirlds (approved):

PR #1496, opened 2025-08-14
test: E2E Happy path test for Autonomous Backfilling 

### LINKED ISSUE #1475: test(backfill): E2E Happy path test for Autonomous Backfilling
In this scenario,

we need to start a Source (Block Node to Backfill from) with Blocks from 0 to 200
and open on a different port other than default, could be `40882`

then we need to start a Target BN with blocks from 100 to 200.
this BN needs to have a valid json config source loaded and referenced it on property so the BN source is available and reachable.

Just wait for the BN Target to trigger the Autonomous Backfill Process

Finally verify that the BN Backfilled the blocks successfully by requesting blocks `99`, `50`. and check logs for completition success message. (we could also check files to make 100% sure, but not sure on feasibility)

mustafauzunn:
## Reviewer Notes

- Add e2e scenario for Autonomous Backfilling 
- Enhance e2e docker containers

## Related Issue(s)
Fixes #1475

### DIFF
--- a/tools-and-tests/suites/src/main/java/module-info.java
+++ b/tools-and-tests/suites/src/main/java/module-info.java
@@ -4,6 +4,7 @@ open module org.hiero.block.node.suites {
     requires com.swirlds.config.extensions;
     requires org.hiero.block.protobuf.protoc;
     requires org.hiero.block.simulator;
+    requires com.github.dockerjava.api;
     requires io.grpc;
     requires java.net.http;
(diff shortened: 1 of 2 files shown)


AlfredoG87 (dismissed):
Looking good, however you need to solve the stylefix issue of the CI Gate:

```
  Please add the following requires directives:
      requires com.github.dockerjava.api;
      
```

georgi-l95 (approved):

PR #1512, opened 2025-08-18
docs(backfill): Update Design, Metrics and Config Documentation regarding (mostly) Backfill 

### LINKED ISSUE #1501: doc(backfill): Update Documentation to include all things related to Backfill
### Story Form

As a BlockNode Operator
I want to have updated documentation regarding the BN
So that in case see Design, Config and Metrics Documents and they will reflect the current state of the BlockNode.

### Technical Notes

- [ ] Update Metrics Doc
- [ ] Update Config Doc
- [ ] Update Design Doc

AlfredoG87:
## Reviewer Notes
Updating Design Doc for Backfill, updating Metrics doc and Configuration doc.

While updating configuration, realized that the current file was very much outdated since it was still referencing pre re-do of BN, so I updated to match current state. 

Also, simplified VerificationConfig and leave it empty since we know in the future it will be needed so we might as well keep the underlying plumbing.

## Related Issue(s)
Fixes #1501

### DIFF
--- a/block-node/verification/src/main/java/module-info.java
+++ b/block-node/verification/src/main/java/module-info.java
@@ -17,7 +17,6 @@ module org.hiero.block.node.verification {
     requires transitive org.hiero.block.node.spi;
     requires transitive org.hiero.block.protobuf.pbj;
     requires com.swirlds.metrics.api;
-    requires org.hiero.block.node.base;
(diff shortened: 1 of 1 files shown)


jsync-swirlds (approved):

Nana-EC (approved):
LG

PR #1515, opened 2025-08-19
fix: Failing `testBackfillOnDemand` After Publisher Correctly Closes Handler Connection

### LINKED ISSUE #1498: Failing `testBackfillOnDemand` After Publisher Correctly Closes Handler Connection
### Story Form

I want to have passing e2e tests
So that I can have confidence in the application and catch regressions

### Technical Notes

- with the introduction of #1497 we've added logic that calls the `onComplete` method on the gRPC pipeline for handlers that are shutting down
  - this has broken the `testBackfillOnDemand`, most likely due to a simulator issue (the simulator could be attempting to use the shut down connection pipeline)
- we must explore the root of the issue and fix this test, which possibly means adding changes to the simulator to properly handle `onComplete`

ata-nas:
## Reviewer Notes

* Fix simulator reuses closed pipeline to send messages
* Re-enabled E2E test: `testBacfkillOnDemand`

Update:
- as suspected, the issue was indeed with the simulator
  - in cases where the block node returns an `EndOfStream`, the simulator must assume that the connection is already closed and needs to open a new one before sending any new requests.
  - by reordering the end stream code handling to be after the setting up of a new connection, we are able to observe the desired result

## Related Issue(s)

Fixes #1498

### DIFF
--- a/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
+++ b/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
@@ -80,17 +80,15 @@ public class PublishClientManager implements SimulatorModeHandler {
 
     private void handleEndStream(Block nextBlock, PublishStreamResponse publishStreamResponse)
             throws InterruptedException, BlockSimulatorParsingException, IOException {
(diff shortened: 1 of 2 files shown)


AlfredoG87 (approved):
LGTM

PR #1408, opened 2025-07-22
chore: Update CN proto to 0.65.0-rc1

Nana-EC:
## Reviewer Notes

CN proto was update to 0.65.0-rc1 and contains notable updates around TraceData and other items

- Bump CN proto to 0.65.0-rc1 in git clone gradle task
- Add export for `com.hedera.hapi.block.stream.trace` in `block-node/protobuf-pbj/src/main/java/module-info.java` to ensure classes are available
- Add export for  `com.hedera.hapi.block.stream.trace.protoc` to `tools-and-tests/protobuf-protoc/src/main/java/module-info.java` to ensure classes are available

## Related Issue(s)
Fixes #1386 
Fixes #1525

**Note for reviewers**
Once there's a 0.65.0 tag we can easily update the hash and there shouldn't be other changes but it's worthwhile to get this fixed to unblock testing with Solo

I'm still investigating the UT failure below. Can't seem to repro it locally 
```
* What went wrong:
Execution failed for task ':protobuf-protoc:compileJava'.
> Compilation failed; see the compiler output below.
  /home/runner/_work/hiero-block-node/hiero-block-node/tools-and-tests/protobuf-protoc/src/main/java/module-info.java:12: error: package is empty or does not exist: com.hedera.hapi.block.stream.trace.protoc
      exports com.hedera.hapi.block.stream.trace.protoc;
                                                ^
  1 error
```

PR #1524, opened 2025-08-21
feat(proto): Update CN Proto version to latest 0.65 release

### LINKED ISSUE #1386: chore(protobuf): Update BN to use latest version of CN protobuf.
### Story Form

As a BN Developer
I want to stay up to date on Proto Spec from CN
So that we implement timely the Spec up to `0.64.x`

### Technical Notes

Once `0.64.x` is released, we should adopt the new protobuf spec, for that we need to update the commit hash of the CN repo that we are using.

see issue #1377 for reference

AlfredoG87:
## Reviewer Notes

- using hash tag from latest commit on https://github.com/hiero-ledger/hiero-consensus-node/tree/release/0.65
- Brute Force changed everything that said EventTransaction for SignedTransaction, and just removed the semantic version thing completely. (this affects both the simulator and the block node tools cli)
- Might still need a deeper review, but made so I can test with solo integration a snapshot version

## Related Issue(s)

Fixes #1386

Related: 
https://github.com/hiero-ledger/hiero-block-node/pull/1408

### DIFF
--- a/block-node/protobuf-pbj/src/main/java/module-info.java
+++ b/block-node/protobuf-pbj/src/main/java/module-info.java
@@ -3,6 +3,7 @@ module org.hiero.block.protobuf.pbj {
     exports com.hedera.hapi.block.stream;
     exports com.hedera.hapi.block.stream.input;
     exports com.hedera.hapi.block.stream.output;
+    exports com.hedera.hapi.block.stream.trace;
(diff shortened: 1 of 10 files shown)


AlfredoG87:
Closing in favor or https://github.com/hiero-ledger/hiero-block-node/pull/1408

Issue #1540, opened 2025-08-26
Automatic HPROF dump in case of OutOfMemoryError


alex-kuzmin-hg:
### Story Form

Currently, if BN crashes with OOMError, it does not save HPROF and pod restarts leaving no artifacts for further debugging OOM.

Suggesting to add the following to Docker defs to java args:
-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="/tmp/dump.hprof"

### Technical Notes

Suggesting to add the following to Docker defs to java args:
-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="/tmp/dump.hprof"

jsync-swirlds:
Parameters added to the command line on main.

PR #1543, opened 2025-08-26
fix(1523): Prevent Unexpected OutOfMemoryError under high load

### LINKED ISSUE #1523: Unexpected OutOfMemoryError under load
in 100M-accounts SDPT run all BlockNodes crash every 30 mins with OOM, e.g. 

https://perf.analytics.eng.hashgraph.io/ephemeral/test_runs/20642-blocknodes-with-sdpt_AdHoc5_124/report/CryptoTransferLoadTest/podlog_solo-sdpt-n5/block-node-0-b767bb5b8-2qrg4_logs/blocknode-0.log

Memory given to BN process: -Xms16G -Xmx16G

Repro in 30 min in SDPT run

jsync-swirlds:
## Reviewer Notes
* Reduced default queue sizes for block items and block notifications.
   * Notifications, in particular, need to be smaller, mostly because
     blocks at high TPS can be 400M or larger.
* Adjusted values files for helm charts in the
  base values, mini values, and mid-size values files.
* Automatic HPROF dump on error
   * Added requested command line flags

## Related Issue(s)
 May address #1523 and #1540.

### DIFF
--- a/block-node/app/docker/update-env.sh
+++ b/block-node/app/docker/update-env.sh
@@ -24,12 +24,10 @@ echo "BACKFILL_BLOCK_NODE_SOURCES_PATH=/opt/hiero/block-node/backfill/backfill-s
 
 if [ true = "$is_smoke_test" ]; then
   # add smoke test variables
-  echo "MEDIATOR_RING_BUFFER_SIZE=1024" >> .env
-  echo "NOTIFIER_RING_BUFFER_SIZE=1024" >> .env
   echo "JAVA_OPTS='-Xms4G -Xmx4G'" >> .env
 else
   # Set the production default values
-  echo "JAVA_OPTS='-Xms16G -Xmx16G'" >> .env
+  echo "JAVA_OPTS='-Xms16G -Xmx16G -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/dump.hprof'" >> .env
 fi
 
 if [ true = "$is_debug" ]; then
--- a/charts/block-node-server/values-overrides/mid-size.yaml
+++ b/charts/block-node-server/values-overrides/mid-size.yaml
@@ -1,4 +1,18 @@
 resources:
   requests:
-    cpu: "4"
+    cpu: "8"
     memory: "12Gi"
+
+blockNode:
+  config:
+    JAVA_OPTS: '-Xms8G -Xmx8G -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="/tmp/dump.hprof"'
+    MESSAGING_BLOCK_ITEM_QUEUE_SIZE: "256"
+    MESSAGING_BLOCK_NOTIFICATION_QUEUE_SIZE: "16"
+    BACKFILL_FETCH_BATCH_SIZE: "8"
+  persistence:
+    archive:
+      size: 16Gi
+    live:
+      size: 4Gi
+    logging:
+      size: 2Gi
--- a/charts/block-node-server/values-overrides/mini.yaml
+++ b/charts/block-node-server/values-overrides/mini.yaml
@@ -6,13 +6,15 @@
 
 resources:
   requests:
-    cpu: "3"
+    cpu: "4"
     memory: "8Gi"
 
 blockNode:
   config:
-    JAVA_OPTS: "-Xms5G -Xmx5G"
-    MEDIATOR_RING_BUFFER_SIZE: "2048"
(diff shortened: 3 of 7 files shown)


block-node/app/docker/update-env.sh
```diff
@@ -24,12 +24,12 @@ echo "BACKFILL_BLOCK_NODE_SOURCES_PATH=/opt/hiero/block-node/backfill/backfill-s
 
 if [ true = "$is_smoke_test" ]; then
   # add smoke test variables
-  echo "MEDIATOR_RING_BUFFER_SIZE=1024" >> .env
-  echo "NOTIFIER_RING_BUFFER_SIZE=1024" >> .env
+  echo "MEDIATOR_RING_BUFFER_SIZE=512" >> .env
+  echo "NOTIFIER_RING_BUFFER_SIZE=16" >> .env
```
AlfredoG87, line 28: Are we still using this configurations?? I believe these are legacy and should be removed. would be surprised if any plugin is using them..

charts/block-node-server/values-overrides/mid-size.yaml
```diff
@@ -1,4 +1,18 @@
 resources:
   requests:
-    cpu: "4"
+    cpu: "8"
     memory: "12Gi"
+
+blockNode:
+  config:
+    JAVA_OPTS: '-Xms8G -Xmx8G -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="/tmp/dump.hprof"'
+    MEDIATOR_RING_BUFFER_SIZE: "256"
```
AlfredoG87, line 9: are we using the mediator prefix?

charts/block-node-server/values-overrides/mini.yaml
```diff
@@ -6,13 +6,15 @@
...
-    cpu: "3"
+    cpu: "4"
     memory: "8Gi"
 
 blockNode:
   config:
-    JAVA_OPTS: "-Xms5G -Xmx5G"
-    MEDIATOR_RING_BUFFER_SIZE: "2048"
+    JAVA_OPTS: '-Xms4G -Xmx4G  -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="/tmp/dump.hprof"'
+    MEDIATOR_RING_BUFFER_SIZE: "64"
```
AlfredoG87, line 15: same as above

charts/block-node-server/values.yaml
```diff
@@ -98,18 +98,20 @@ blockNode:
...
     # key: value
     JAVA_TOOL_OPTIONS: "-Djava.util.logging.config.file=/opt/hiero/block-node/logs/config/logging.properties"
-    JAVA_OPTS: "-Xms16G -Xmx16G"
+    JAVA_OPTS: '-Xms16G -Xmx16G -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="/tmp/dump.hprof"'
     # PRODUCER_TYPE: "NO_OP"
     # PERSISTENCE_STORAGE_TYPE: "NO_OP"
     # VERIFICATION_TYPE: "NO_OP"
     # MEDIATOR_TYPE: "NO_OP"
-    MEDIATOR_RING_BUFFER_SIZE: "8192"
+    # MEDIATOR_RING_BUFFER_SIZE: "512"
```
AlfredoG87, line 106: I think we can safely remove anything referencing the "mediator"

AlfredoG87 (approved):
LGTM, just some nits related to the same thing.

block-node/app/docker/update-env.sh
jsync-swirlds: I don't actually know, so I'm keeping it safe in this PR.
I'll file an issue to investigate and remove these if possible.

charts/block-node-server/values-overrides/mid-size.yaml
jsync-swirlds: Nope, copy-pasta error, will fix.

Nana-EC (commented):
Agree with Fredy's comments.
One request is to update the Config README with the new defaults

charts/block-node-server/values.yaml
jsync-swirlds: Done.

AlfredoG87 (dismissed):

AlfredoG87 (dismissed):

AlfredoG87 (approved):

jsync-swirlds:
Apologies for all the spacing changes in the configuration markdown, that was the result of a spotless reformat.

PR #1544, opened 2025-08-26
chore: Remove stack trace from certain debug logs

jsync-swirlds:
## Reviewer Notes
The stack trace when a remote entity closes a connection makes readers assume the block node crashed or something else is _wrong_.  This is a false assumption that wastes a lot of time explaining that a socket exception is a _normal_ client end-of-communication, so stop printing a stack trace in those cases.


Nana-EC (approved):
LG

PR #1557, opened 2025-08-28
test: add initial assertions

mustafauzunn:
## Reviewer Notes

Add `Get firstBlock` and ` Get lastAvailableBlock mirror node`

## Related Issue(s)

### DIFF
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -178,23 +178,22 @@ jobs:
               registry: docker.io
               repository: xinatswirlds/importer
               tag: e741ea2b2
-            config:
-              hiero:
-                mirror:
-                  importer:
-                    block:
-                      enabled: true
-                      nodes:
-                        - host: block-node-0.${NAMESPACE}.svc.cluster.local
-                          port: 40840
-                      sourceType: BLOCK_NODE
-                  downloader:
-                    record:
-                      enabled: false
-                  startDate: 1970-01-01T00:00:00Z
+          hiero:
+            mirror:
+              importer:
+                block:
+                  enabled: true
+                  nodes:
+                    - host: block-node-0.${NAMESPACE}.svc.cluster.local
+                      port: 40840
+                  sourceType: BLOCK_NODE
                   stream:
                     maxSubscribeAttempts: 10
                     responseTimeout: 10s
+                startDate: 1970-01-01T00:00:00Z
+                downloader:
+                  record:
+                    enabled: false
           EOF
 
       - name: Verify Mirror Values Override File Created


.github/workflows/solo-e2e-test.yml
```diff
@@ -260,4 +260,40 @@ jobs:
           echo "Block is $BLOCK"
           echo "BLOCK=$BLOCK" >> $GITHUB_ENV
 
-      # TODO verify MN using curl and REST API
+      - name: Get firstBlock
+        run: |
+          FIRSTBLOCK=$(grpcurl \
+            -plaintext \
+            -import-path protobuf-sources/src/main/proto \
```
Nana-EC, line 267: Q: Why do both paths need to be imported. The `block-node-protobuf` folder should have everything no?

.github/workflows/solo-e2e-test.yml
```diff
@@ -260,4 +260,40 @@ jobs:
...
+
+      - name: Port Forward to Mirror Node
+        run: |
+          kubectl port-forward --namespace "${NAMESPACE}" svc/mirror-rest  5551:80
+          # wait for port-forward to be established
+          sleep 10
+
+      - name: Get lastAvailableBlock mirror node
+        run: |
+          BLOCKMIRROR=$(curl -s "http://127.0.0.1:5551/api/v1/blocks?limit=1&order=desc" | jq -r '.name')
```
Nana-EC, line 291: This query should get `$FIRSTBLOCK`
This serves as a validation of the block being streamed over in this setup.
Just do a get on block `$FIRSTBLOCK`

AlfredoG87 (approved):
LGTM

.github/workflows/solo-e2e-test.yml
```diff
@@ -260,4 +260,40 @@ jobs:
...
 
-      # TODO verify MN using curl and REST API
+      - name: Get firstBlock
+        run: |
+          FIRSTBLOCK=$(grpcurl \
+            -plaintext \
+            -import-path protobuf-sources/src/main/proto \
+            -import-path protobuf-sources/block-node-protobuf \
+            -proto block-node/api/block_access_service.proto \
+            -d '{"block_number": "0"}' \
```
Nana-EC, line 270: I would get the first block based on the earliest block from serverStatus.
In the previous section we should make it get earliest and latest.
The getBlock on the BN and MN can use those values.
Start with the earliest for MN though. Depending on latency we'll want to be smart on when to query for latest on MN, it should work but something to watch out for and potentially separate in case you see issue initially

Nana-EC (commented):
Nice simple query use.
Left some suggestions

PR #1558, opened 2025-08-28
fix: Reduce the CPU time spent waiting for new block data

jsync-swirlds:
## Reviewer Notes
* Changed the timeout on the condition wait from 500 microseconds to 5 milliseconds.
* Replaced a sleep with parkNanos in a unit test to improve reliability

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -44,8 +44,7 @@ import org.hiero.block.node.spi.threading.ThreadPoolManager;
  */
 public final class LiveStreamPublisherManager implements StreamPublisherManager {
     private static final String QUEUE_ID_FORMAT = "Q%016d";
-    private static final int DATA_READY_WAIT_MICROSECONDS = 500;
-    // @todo(1413) utilize the logger
+    private static final int DATA_READY_WAIT_MICROSECONDS = 5000;
     private final System.Logger LOGGER = System.getLogger(LiveStreamPublisherManager.class.getName());
     private final MetricsHolder metrics;
     private final BlockNodeContext serverContext;
--- a/block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/StreamPublisherPluginTest.java
+++ b/block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/StreamPublisherPluginTest.java
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.stream.publisher;
 
+import static java.util.concurrent.locks.LockSupport.parkNanos;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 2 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -44,8 +44,7 @@
  */
 public final class LiveStreamPublisherManager implements StreamPublisherManager {
     private static final String QUEUE_ID_FORMAT = "Q%016d";
-    private static final int DATA_READY_WAIT_MICROSECONDS = 500;
-    // @todo(1413) utilize the logger
+    private static final int DATA_READY_WAIT_MICROSECONDS = 5000;
```
Nana-EC, line 47: This should be a configureable item now that we've observed it relevance and impact.
We should have to have a PR each time we need to tweak the value in a test environment.

Nana-EC (approved):
Minor fix to support quick testing but we should make this more extensible in follow up

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: We try to minimize configuration values, so I've avoided that here as it's not something an operator should ever change.  Once we have a reasonable value it should be fixed in the code.  This isn't operator-level tuning, but a fixed parameter that was set overly aggressively initially.

ata-nas: +1 on this not being configurable. I also think that we better have this well tested and part of the code. We _could_ however, if we consider configuring this, to very much limit a lot the possible values?

ata-nas (approved):
Looks good. Out of curiosity, why do we generally opt to using parkNanos instead of a sleep? I've read different opinions.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
Nana-EC: It's a fine balance. I'm also of the don't make everything configurable camp.
If we believe fine tuning it once is applicable to all scenarios (TPS, node types, etc) then it's fine to not expose it. If it's something an operator should tweak to ensure an optimal and reduced cost node then we should consider "if" and "when" to expose configs of that nature.

Leaving it not exposed is fine for now.

jsync-swirlds:
> Looks good. Out of curiosity, why do we generally opt to using parkNanos instead of a sleep? I've read different opinions.

Sleep calls `parkNanaos` under the covers in a loop, so sleep wakes more often just to check for interruption (so it can throw the exception).
Calling the `parkNanos` method directly is generally more efficient, as long as you aren't sleeping more than 10 ms it's generally a better choice (and sometimes it's better for longer pauses).

It also doesn't throw a checked exception that is the most widely mishandled exception in Java...

alex-kuzmin-hg:
With this fix, the problem is below sensitivity
image

image
Recommended to merge

PR #1564, opened 2025-08-29
fix: Clear `queueByMap` When Verification Fails

### LINKED ISSUE #1565: Clear `queueByMap` When Verification Fails
### Story Form

As a Block Node Operator/User/Engineer
I want to properly manage the internal state of the publisher manager
So that the publisher functionality will work accordingly

### Technical Notes

- the publisher manager has to clear all queues by block number when verification fails

ata-nas:
## Reviewer Notes

* Clear the queueByMap on failed verficiations
* Added todos with considerations for upcoming changes
* Improved the handlerIsEnding method with a CAS for the next unstreamed

## Related Issue(s)

Fixes #1565

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -114,6 +114,9 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         // block contained in the queue is incomplete!
         // It takes just as long to loop the map as to call `containsValue`.
         // so just loop through and remove the entry if it's found.
+        // @todo(1239): we need to redo the loop below. We could walk the
+        //    queueByBlockMap in reverse order and only remove the first
+        //    time we encounter an incomplete block that matches this handler.
         for (final Map.Entry<Long, BlockingDeque<BlockItemSetUnparsed>> nextEntry : queueByBlockMap.entrySet()) {
             final BlockingDeque<BlockItemSetUnparsed> queueByBlock = nextEntry.getValue();
             // Note: we are using ConcurrentMap which does not allow any null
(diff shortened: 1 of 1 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -392,25 +395,24 @@ public void handlerIsEnding(final long blockNumber, final long handlerId) {
...
-                    // from the queueByBlockMap, as it will be in an incomplete
-                    // state when the handler shuts down, which happens on the
-                    // same thread that called this method immediately after.
-                    // It is important that this queue is removed from the
-                    // queueByBlockMap, because it is not guaranteed that the
-                    // same publisher will supply the next unstreamed block we
-                    // just decremented.
-                    queue.clear();
-                }
+            nextUnstreamedBlockNumber.compareAndSet(blockNumber + 1, blockNumber);
```
Nana-EC, line 398: nit: the previous use of `decrementAndGet()` made it clear that the `blockNumber` was being decremented. 
It's a bit more foggy now but can be worked out.
This is a good location for a good comment covering why we decrement by this amount and in this order

Nana-EC (approved):
LG.
Could use more comments at key locations but this is non blocking for this PR

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Comment added.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -146,6 +150,26 @@ public void removeHandler(final long handlerId) {
...
+        // Remove items until we find a block proof or the queue is empty.
+        while (last != null && !(last.blockItems().getLast().hasBlockProof())) {
+            queue.removeLast();
+            itemsRemoved++;
+            last = queue.peekLast();
+        }
+        // @todo(1416) add an "items discarded" metric.
+        return itemsRemoved > 0;
+    }
+
```
jsync-swirlds, line 172: Note, Added this method (and called in two places) because the smaller change failed tests due to being _too_ aggressive at queue clearing.  This fixes that and also makes the general case more correct.

jsync-swirlds (approved):

Nana-EC (approved):

PR #1589, opened 2025-09-04
ci(solo-e2e): Various Improvements to Solo E2E CI Test

### LINKED ISSUE #1818: feat(ci): Finish e2e solo integration WF and enable it for automatic testing.
### Story Form

As a BlockNode Dev
I want to have an e2e integration automated test
So that I can keep assured that new versions of the BN, CN or MN does not break compatibility.

### Technical Notes

This is follow-up of Quest, now that the MN supports the blockstream we should be able to finish this WF and enabled it on the CI regular testing.

AlfredoG87:
## Reviewer Notes
- migrate to use latest version of solo command structure
- Latest versions of CN, BN and MN by defaults. 
- Removed hardcoded MN Image, Custom built by XIN as a temporary workaround.
- Compacted the WF Steps by grouping them similar to CITR WF on CN Repo
- Added TCK Regression Test (disabled by default) and limited to only 1 file test, same as CN Repo CITR.
- Removed Obsolete Incomplete E2E Test attempt on Hedera Local Node.

**Pending to decide the cadence of the run for this WF, right now only under demand**

Recent pass with TCK Regression Suites enabled: 
https://github.com/hiero-ledger/hiero-block-node/actions/runs/19021065941

## Related Issue(s)
Fixes #1818


.github/workflows/solo-e2e-test.yml
```diff
@@ -7,15 +7,15 @@ on:
       block-node-helm-chart-version:
         description: "Block Node Helm Chart Version"
         required: true
-        default: "v0.17.1"
+        default: "v0.21.2"
```
jsync-swirlds: This doesn't run on PR's, correct?

Assuming that's true, we might want to add this line to the places versions are updated by the release workflow.

jsync-swirlds (approved):
Looks good, one item to follow up in a separate PR/Issue.
Also, E2E tests need a fix.

ata-nas (approved):

PR #1604, opened 2025-09-09
feat: Simulator add publisher support to send EndStream message

### LINKED ISSUE #1168: Simulator add publisher support to send `EndStream` message
### Story Form

As a Block Node Engineer
I want to test handling of `EndStream` messages from publishers
So that I can verify correct behavior across all likely scenarios

### Technical Notes

Add support in the block stream simulator, publisher mode, for _sending_ an `EndStream` message to the block node when the block node is too far behind this publisher, every `N` blocks (to simulate occasionally switching block nodes for resilience), and other scenarios as detailed in the protocol documentation.

mustafauzunn:
## Reviewer Notes

Add support in the block stream simulator, publisher mode, for sending an EndStream message to the block node.

## Related Issue(s)
Fixes #1168

### DIFF
--- a/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
+++ b/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
@@ -102,7 +102,7 @@ public class PublishClientManager implements SimulatorModeHandler {
                     case TIMEOUT -> EndStream.Code.TIMEOUT;
                     case ERROR -> EndStream.Code.ERROR;
                     case TOO_FAR_BEHIND -> EndStream.Code.TOO_FAR_BEHIND;
-                    default -> null;
+                    case NONE -> null;
                 };
 
         if (code != null) {
--- a/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublisherClientModeHandler.java
+++ b/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublisherClientModeHandler.java
@@ -114,7 +114,7 @@ public class PublisherClientModeHandler implements SimulatorModeHandler {
         final long secondsPerBlockNanos = (long) millisecondsPerBlock * NANOS_PER_MILLI;
         PublishStreamResponse[] responseHolder = new PublishStreamResponse[1];
         long blockCount = 0;
-        long endStreamInterval = blockStreamConfig.endStreamEveryNBlocks();
+        long endStreamInterval = blockStreamConfig.endStreamFrequency();
 
         Block nextBlock = blockStreamManager.getNextBlock();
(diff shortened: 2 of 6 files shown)


tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
```diff
@@ -70,28 +72,48 @@ public void stop() throws InterruptedException {
...
+                    default -> null;
+                };
+
+        if (code != null) {
+            currentClient.handleEndStreamModeIfSet(code);
+        }
+
+        stop();
+        initializeNewClientAndHandler();
+        start();
```
ata-nas, line 114: THis is probably the wrong place to stop -> init -> start again. If we do this after handling end stream mode, this caused an issue in the past because the block node will close the stream. Sending anything will result in an exception. See #1515.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublisherClientModeHandler.java
```diff
@@ -112,9 +113,18 @@ public void start() throws BlockSimulatorParsingException, IOException, Interrup
...
 
         Block nextBlock = blockStreamManager.getNextBlock();
         while (nextBlock != null && shouldPublish.get()) {
+            blockCount++;
+            if (blockStreamConfig.endStreamMode() != EndStreamMode.NONE
+                    && endStreamInterval > 0
+                    && blockCount % endStreamInterval == 0) {
+                publishClientManager.sendEndStream(blockStreamConfig.endStreamMode());
+                break;
+            }
```
ata-nas, line 127: Do we restart the stream after we send the endStream, cause we break. Also, maybe it should be configurable if we restart the stream after and EndStream is sent? Valid for both usages.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublishClientManager.java
```diff
@@ -70,28 +72,48 @@ public void stop() throws InterruptedException {
...
+    public void sendEndStream(EndStreamMode endStreamMode)
+            throws BlockSimulatorParsingException, IOException, InterruptedException {
+        EndStream.Code code =
+                switch (endStreamMode) {
+                    case RESET -> EndStream.Code.RESET;
+                    case TIMEOUT -> EndStream.Code.TIMEOUT;
+                    case ERROR -> EndStream.Code.ERROR;
+                    case TOO_FAR_BEHIND -> EndStream.Code.TOO_FAR_BEHIND;
+                    default -> null;
+                };
```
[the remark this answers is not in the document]
Nana-EC: Agreed, we don't want unexpected bahviour to go unrecognized.

PR #1605, opened 2025-09-09
test: Combined Backfilling Flows running in parallel

### LINKED ISSUE #1503: test(backfill): Combined Backfilling Flows running on parallel.
### Story Form

As a Block Node Maintainer
I want to 
So that ?

Let a BN start Autonomous Backfilling for a very large amount of batches and some time between batches that forces it to go slow.
in the mean while, send an TOO_FAR_BEHIND. for a few blocks. (using simulator or grpcurl)

### Technical Notes

Verify that Autonomous task is running.

Send TOO_FAR_BEHIND

Verify that the On-Demand task get executed on parallel, and that is able to even finish sooner than the first one.

Is not necessary to wait for the Autonomous to complete.

mustafauzunn:
## Reviewer Notes

Verify that Autonomous backfill is running.
Send TOO_FAR_BEHIND.
Verify that the On-Demand task gets executed in parallel.

## Related Issue(s)
Fixes #1503

### DIFF
--- a/tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
+++ b/tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
@@ -200,6 +200,7 @@ public class PositiveMultiplePublishersTests extends BaseSuite {
     @Test
     @DisplayName(
             "Autonomous backfill should fill the gaps and Publisher should send TOO_FAR_BEHIND to activate backfill on demand")
+    @Timeout(90)
     public void testBackfillOnDemandAndAutonomousBackfill() throws IOException, InterruptedException {
 
         //  2 Block Nodes, 1 Source (40840), Subject BN: Backfill will happen here. (8082)


AlfredoG87 (approved):
LGTM

tools-and-tests/suites/src/main/resources/logging.properties
jsync-swirlds, line 1: This concerns me a bit.

The hiero-consensus-node codebase has a problem with log properties proliferation that makes identifying the correct properties to change quite challenging.  I would prefer we avoid that in this codebase.

Can we use other mechanisms to set the logging properties for testing from within the tests perhaps?

Also (and this is just my opinion), the default logging for all tests should, in general, be complete silence. Anything run in CI should report _only_ failures and just enough to identify where a failure is (so it can be rerun with high detail in a local environment), because otherwise we easily end up with so much garbage in the CI log that github truncates the output so we get nothing for actual failures. Even if we keep the full output, the amount of unimportant logs can make identifying what actually failed excessively difficult.

AlfredoG87: > Can we use other mechanisms to set the logging properties for testing from within the tests perhaps?

We could, however this is the way we are loading the logging properties on docker-compose and helm so is consistent on e2e.
is easy to change locally by just adjusting the file accordingly rather than re-compile code.

> Also (and this is just my opinion), the default logging for all tests should, in general, be complete silence....

I personally find value in having tests run in verbose mode, that way I can quicker and easily triage issues by just looking at CI logs. But understand the need of balance in general and be careful of this.

Just for context on why we added this to this PR.
Since the level of complexity of this e2e test, we needed to take good look at the logs for debugging at we realized that the testContainers that were created were outputting logs using the colorful formatter that the BN is still using byDefault in Dev Mode.

And also we needed to increase the verbosity to FINEST, in order to triage and find out what was happening during the test.

So while I agree that local runs of e2e tests should have more verbosity than CI tests, for now I think this is better that what we used to have, which was nothing to fine tune and configure the logging.properties of BNs in e2e Tests.

I have created an issue to allows to continue to improve on logging as we go along and don't forget the things that could be improved and others that need fixing.
https://github.com/hiero-ledger/hiero-block-node/issues/1670

tools-and-tests/suites/src/main/resources/logging.properties
```diff
@@ -0,0 +1,104 @@
...
+#org.hiero.block.server.persistence.storage.write.level=FINE
+#org.hiero.block.server.persistence.storage.archive.LocalGroupZipArchiveTask.level=FINE
+
+# Helidon PBJ Plugin loggers
+#org.hiero.pbj.grpc.helidon.PbjProtocolHandler.level=FINE
+
+################################################################################
+# Handlers configuration
+################################################################################
+handlers = java.util.logging.ConsoleHandler, java.util.logging.FileHandler
```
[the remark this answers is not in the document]
AlfredoG87: added it to https://github.com/hiero-ledger/hiero-block-node/issues/1670

PR #1641, opened 2025-09-17
ci(release): Update WF to use new helm chart name

AlfredoG87:
- Updated the Release Workflow to use new helm chart name.
- Updated Chart docs to reflect the new name and bump the doc example to latest version on main ("0.20.0-SNAPSHOT") 

## Related Issue(s)
 Use keywords `Fix, Fixes, Fixed, Close, Closes, Closed, Resolve, Resolves, Resolved`
to connect an issue to be closed by this pull request.

PR #1602, opened 2025-09-08
fix: Conversion to statefulset

shezaan-hashgraph:
# Summary: Deployment to StatefulSet Conversion Changes
## Core Workload Changes
- Changed workload type: Deployment → StatefulSet
- Added serviceName: Required field for StatefulSet (`serviceName: {{ include "hiero-block-node.fullname" . }}`)
- Set replicas to default to 1: StatefulSets typically run single instances for stateful apps; however, there is a provision to increase the replicas if ever in the future more than 1 BN server can be run. Currently, it is not possible to run more than 1 BN at a time.

## Volume Management Overhaul
- Replaced separate PVC templates with volumeClaimTemplates in StatefulSet
- Removed files: `pvc-archive.yaml`, `pvc-logging.yaml`, `pvc-live.yaml`, `deployment.yaml`
- Added automatic PVC naming: PVCs now named `<volume-name>-<statefulset-name>-<pod-index>`
  - `logging-storage-<name>-0`
  - `archive-storage-<name>-0`
  - `live-storage-<name>-0`
- Fix minor PVC name inconsistencies i.e. `logging-data` to `logging-storage` to be consistent with `archive-storage` and `live-storage`
## Flexible PVC Support
- Added `existingClaim` option to `values.yaml` for each volume (archive, logging, live)
- Implemented mixed mode support: Some volumes can use existing PVCs, others auto-created
- Volume logic:
  - If `existingClaim` specified → uses existing PVC in volumes section
  - If no `existingClaim` → creates PVC via `volumeClaimTemplates`

## Configuration Updates
- Fixed chart name: `block-node-helm-chart` → `block-node-server` (removed -helm-chart suffix. The suffix does not look good on pods, statefulsets or even deployments previously)
- Added existingClaim comments to `values.yaml` persistence sections
- Updated `nano.yaml` with existingClaim examples using the mixed mode support

## Volume Mount Structure
- Preserved subPath usage:
  - Archive: `subPath: archive-data`
  - Live: `subPath: live-data`
  - Logging: No subPath (mounts to root) previously unused in deployment as well
- Maintained mount paths:
  - Archive: `/opt/hiero/block-node/data/historic`
  - Live: `/opt/hiero/block-node/data/live`
  - Logging: `/opt/hiero/block-node/logs`
 
## Init Container Considerations
- Kept existing init container for directory setup
- Init container creates:
  - `/live-pvc/live-data` (matches subPath: live-data)
[...]

PR #1646, opened 2025-09-18
ci(publish): Revert chart rename to preserve backwards and forwards compatibility

### LINKED ISSUE #1645: bug(release): New Chart name for the BN breaks Solo backwards and forwards compatibility
### Description

Since Solo is a user of the BN Charts, changing its name and package would need changes on SOLO, and would also complicate the backwards compatibility and forwards. So unless is absolutely necessary we should not change, and if so, we should do it after Solo team is ready.

### Steps to Reproduce

_No response_

### Additional context

_No response_

AlfredoG87:
## Reviewer Notes

As part of an extensive set of improvements made on PR: https://github.com/hiero-ledger/hiero-block-node/pull/1602
BlockNode **Chart Name, was changed** to keep consistency with its parent folder and simplify the name.

This broke the Publish workflow which was expecting one chart package name and was not finding it.

This PR fixed that and improve another thing. https://github.com/hiero-ledger/hiero-block-node/pull/1641

As a result we where able to publish sucessfully but potentially breaking solo and even if fixing solo, making it hard to do backwards and forwards compatibility, so **this PR aims to revert those 2 changes mentioned in the respective PRs and return the BN Chart name to its previous one to preserve stability.**

## Related Issue(s)

#1602 
#1641 

Fixes #1645

### DIFF
--- a/.github/workflows/release-push-image.yaml
+++ b/.github/workflows/release-push-image.yaml
@@ -317,7 +317,7 @@ jobs:
       - name: Push helm chart
         if: "${{ github.event_name != 'workflow_dispatch' || github.event.inputs.publish == 'true' }}"
         run: |
-          helm push block-node-server-${{ env.VERSION }}.tgz oci://ghcr.io/hiero-ledger/hiero-block-node
+          helm push block-node-helm-chart-${{ env.VERSION }}.tgz oci://ghcr.io/hiero-ledger/hiero-block-node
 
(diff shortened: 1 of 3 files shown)


jeromy-cannon (approved):

AlfredoG87:
Closing this PR as will not merge, we decided to go in the direction of updating Solo.

Issue #1576, opened 2025-09-03
Deadlock on the notification queue 


mustafauzunn:
### Description

When the queue is full that creates deadlock on the notification queue. This is possible root cause to several bugs.

When one block node activates both autonomous and on-demand backfill, we have flaky behaviour.
Sometimes the block node that requests backfilling has the following behaviour:

```
2025-09-02 15:06:32.932+0300 FINER   BackfillPlugin#backfillGap               Backfill for gap 30->34 did not finish in time
2025-09-02 15:06:32.932+0300 FINER   BackfillPlugin#backfillGap               Backfill for gap 121->125 did not finish in time
2025-09-02 15:06:33.942+0300 INFO    BackfillGrpcClient#fetchBlocks           Requesting blocks for range: 126 to 130
2025-09-02 15:06:33.943+0300 INFO    BackfillGrpcClient#fetchBlocks           Requesting blocks for range: 35 to 39
2025-09-02 15:06:33.946+0300 INFO    BackfillGrpcClient#fetchBlocks           Trying Block Node: localhost:8,081
2025-09-02 15:06:33.946+0300 INFO    BackfillGrpcClient#fetchBlocks           Trying Block Node: localhost:8,081
2025-09-02 15:06:36.234+0300 FINER   BlockStreamSubscribeUnparsedClient$SubscribePipeline#onComplete received onComplete
2025-09-02 15:06:36.234+0300 FINER   BlockStreamSubscribeUnparsedClient$SubscribePipeline#onComplete received onComplete
[...]

PR #1704, opened 2025-10-01
chore(config): Improve configs defaults and chart overrides regarding queueSize and backfillBatchSize

AlfredoG87:
## Reviewer Notes

- As we noted on ticket https://github.com/hiero-ledger/hiero-block-node/issues/1576 we have sometimes a deadlock under certain circustances, even when there is another ticket to prevent that from happening, it makes sense to improve the default properties values and also con helm charts overrides to improve performance and minimize the risk of queues getting filled.

## Related Issue(s)

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
@@ -32,9 +32,9 @@ public record BackfillConfiguration(
         @Loggable @ConfigProperty(defaultValue = "60000") @Min(100) int scanInterval,
         @Loggable @ConfigProperty(defaultValue = "3") @Min(0) @Max(10) int maxRetries,
         @Loggable @ConfigProperty(defaultValue = "5000") @Min(500) int initialRetryDelay,
-        @Loggable @ConfigProperty(defaultValue = "25") @Min(1) @Max(10_000) int fetchBatchSize,
+        @Loggable @ConfigProperty(defaultValue = "10") @Min(1) @Max(1024) int fetchBatchSize,
         @Loggable @ConfigProperty(defaultValue = "1000") @Min(100) int delayBetweenBatches,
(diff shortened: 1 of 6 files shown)


block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
```diff
@@ -32,9 +32,9 @@ public record BackfillConfiguration(
         @Loggable @ConfigProperty(defaultValue = "60000") @Min(100) int scanInterval,
         @Loggable @ConfigProperty(defaultValue = "3") @Min(0) @Max(10) int maxRetries,
         @Loggable @ConfigProperty(defaultValue = "5000") @Min(500) int initialRetryDelay,
-        @Loggable @ConfigProperty(defaultValue = "25") @Min(1) @Max(10_000) int fetchBatchSize,
+        @Loggable @ConfigProperty(defaultValue = "10") @Min(1) @Max(1024) int fetchBatchSize,
```
Nana-EC, line 35: Q: Was this meant to be 16 to be half of the `blockNotificationQueueSize`?

Nana-EC (commented):
Q: Were we going for batch size being 1/3 or 1/2 of the notification queue size?

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
AlfredoG87: Good question, this is a missunderstanding shared by @jsync-swirlds aswell, in the rare case that `Autonomous` and `On-Demand` Backfilling are working at the same time, then the batchSize is Per Flow, so we need an extra cushion.

AlfredoG87: So we are going for approx 1/3

jsync-swirlds (approved):

PR #1714, opened 2025-10-02
fix(log): Chart default log level to show Trace as intended.

### LINKED ISSUE #1713: Bug(Chart): Log level is supposed to be set to trace by default on chart deployments but its not.
### Description

Currently when you deploy a BN using the Charts, the logging level for the package `org.hiero.block.*` is supposed to be as trace level as per configuration
```
      org.hiero.block.level: "FINEST"

```

however on practice is not showing expected logs

### Steps to Reproduce

Deploy a BN using default chart values

### Additional context

_No response_

AlfredoG87:
Fixing Log Handlers levels to show by default, as intended right now, TRACE logs for org.hiero.block.level package.

and fixing a wrongly defined number, that might not be parseable.

Fixes #1713


Nana-EC (approved):

AlfredoG87:
> Approved for this release, but I question setting this level of detail as default.
We are saying that by default we will completely overwhelm systems with logging that is almost entirely irrelevant (with the occasional needle of value in a haystack of logs).
That's not a good way to deploy a system and we really shouldn't turn on trace logging globally ever in my opinion. Defaults should be very quiet (WARNING) and only overridden when necessary.

The plan is to revert this after Beta2.

PR #1717, opened 2025-10-02
fix: Send Acknowledgements Only On Successful Persistence

### LINKED ISSUE #1718: Send Acknowledgements Only On Successful Persistence
### Description

- sending acknowledgements is possible on failed verification
- acknowledgements must be sent only on successful verification

ata-nas:
## Reviewer Notes

* Adding success check when handling a persistence notifications
* Adding tests for this logic

## Related Issue(s)

Fixes #1718

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -566,13 +566,15 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     public void handlePersisted(@NonNull final PersistedNotification notification) {
         if (notification != null) {
             final long newLastPersistedBlock = notification.blockNumber();
-            if (newLastPersistedBlock > lastPersistedBlockNumber.get()) {
-                handlers.values().parallelStream().unordered().forEach(handler -> {
-                    // _Important_, we only need the last persisted block number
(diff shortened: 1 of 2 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -566,13 +566,15 @@ public void handleVerification(@NonNull final VerificationNotification notificat
     public void handlePersisted(@NonNull final PersistedNotification notification) {
         if (notification != null) {
```
jsync-swirlds, line 567: Move the if succeeded here to make this a bit simpler and clearer.
```suggestion
        if (notification != null && notification.succeeded()) {
```

jsync-swirlds: We can do an if null in the else; which seems more reasonable.

ata-nas: I personally find that to be error prone. It is easy to miss/neglect yet anotehr null check. I think we should stick with a single point to check if the notification is null and afterwards we are safe to use the reference.

jsync-swirlds: Fair point.

jsync-swirlds (approved):

PR #1721, opened 2025-10-03
chore: Disable FileHandler for Java System Logger for E2E Tests

### LINKED ISSUE #1633: Disable FileHandler for Java System Logger for E2E Tests
### Story Form

As a Block Node Engineer
I want to remove file handlers for logging during E2E CI runs
Because they are redundant and cause issues

### Technical Notes

- currently we are observing exceptions when the java system logger cannot resolve a file for a file handler it has configured
- disable the file handler for CI items (e2e tests)

ata-nas:
## Reviewer Notes

* Add the ability to overwrite the default logging properties when running in CI

## Related Issue(s)

Resolves #1633

### DIFF
--- a/block-node/app/build.gradle.kts
+++ b/block-node/app/build.gradle.kts
@@ -90,6 +90,18 @@ val copyDockerFolder: TaskProvider<Copy> =
         into(dockerBuildRootDirectory)
     }
 
+val copyDockerFolderCI: TaskProvider<Copy> =
+    tasks.register<Copy>("copyDockerFolderCI") {
+        description =
+            "Copies the docker folder to the build root directory with CI specific files. To be used only in CI environments, for instance when running E2E tests."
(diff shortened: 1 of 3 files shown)


ata-nas:
@jsync-swirlds these changes _should_ be sufficient to disable the file handlers when running in CI, I hope no failures will be seen on CI here. Note that we need the file handlers for the simulator for long running tests (deployments BN agains the simulator). If, however, we find no need for them, we could entirely remove them. What I have done (for the simulator) is to change the `logging.properties` of the suites. That should be sufficient to disable the handler for the simulator as we start it on a thread and not on docker (testcontainers) which we do for the BN. For the BN I have introduced a new build task for CI which moves the `ci-logging.properties` to `logging.properties` which then gets picked up.

jsync-swirlds (approved):

Issue #1587, opened 2025-09-04
Multi-Publisher Plugin Support for `EndOfStream.Code.PERSISTENCE_FAILED`


ata-nas:
### Story Form

As a Publisher
I want to receive an `EndOfStream.Code.PERSISTENCE_FAILED` message when a block I've published fails to persist
So that I can take appropriate action on my side

### Technical Notes

- we need to add support to send the `EndOfStream.Code.PERSISTENCE_FAILED` when a block has failed persistence

Note:
 - This notification goes to all handlers
 - The end stream closes the connection, so all handlers will disconnect
 - It doesn't matter if there are other queued blocks; they might also fail to persist, but we won't have any handlers to send them to (or if we do they're new connections that also should be closed the same way).

Nana-EC:
@ata-nas do we have considerations for what appropriate action options we have? Is it logging, retrying or maybe more complex.
Or does this ticket need to figure that out if not prescribed in the communication protocol

PR #1722, opened 2025-10-03
feat: Multi-Publisher Plugin Support for EndOfStream.Code.PERSISTENCE_FAILED

mustafauzunn:
## Reviewer Notes

We need to add support to send the `EndOfStream.Code.PERSISTENCE_FAILED` when a block has failed persistence

## Related Issue(s)
Fixes: #1587


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -466,6 +466,25 @@ public long handleFailedVerification(final long blockNumber) {
...
+    /**
+     * This method must be called when persistence fails for a given block.
+     * We will attempt to send an {@link EndOfStream} with a {@link Code#PERSISTENCE_FAILED} and
+     * proceed to shut down the handler.
+     *
+     * @param blockNumber of the block that failed persistence
+     * @return the id of this handler
+     */
+    public long handleFailedPersistence(final long blockNumber) {
+        unacknowledgedStreamedBlocks.remove(blockNumber);
```
ata-nas, line 478: Not sure if we should remove the block number from this set. This is relevant for failed verification. Since we are closing the stream for sure, there is no value in decrementing this. Also, there is a race condition here if a verification fails and the handler get's called to handle the failed verification, it will not be accurate the remove from the set thereof.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -575,6 +575,20 @@ public void handlePersisted(@NonNull final PersistedNotification notification) {
                     });
                     lastPersistedBlockNumber.set(newLastPersistedBlock);
                 }
+            } else {
+                final long blockNumber = notification.blockNumber();
+                nextUnstreamedBlockNumber.set(blockNumber);
+                currentStreamingBlockNumber.set(blockNumber);
```
ata-nas, line 581: Just like in the verification failed method, we noted a very rare race condition where we have reset these values and some active publisher supplies the next unstreamed one before we close all connection. This will leave the value set to blockNumber + 1 (or more) and we will drop all blocks potentially before even propagating the newly come items, and once we close all connections and drop all queues, we could miss a block being supplied. Just pointing this out, as we have to address this (already captured in GH issue).

ata-nas (approved):
LGTM, nice addition.

jsync-swirlds (approved):

Nana-EC (approved):

PR #1739, opened 2025-10-08
chore: Multi-Publisher Plugin Logging

### LINKED ISSUE #1413: Multi-Publisher Plugin Logging
### Story Form

As a Block Node Operator/User/Developer
I want to have proper logging
So that I can diagnose issues when running the plugin and be able to take appropriate actions

### Technical Notes

**DEPENDS ON**: #1370
- we need to introduce logging for the multi-publisher plugin and related classes

mustafauzunn:
## Reviewer Notes

We need to introduce logging for the multi-publisher plugin and related classes

## Related Issue(s)
Fixes #1413

### DIFF
--- a/.github/workflows/helm-charts.yaml
+++ b/.github/workflows/helm-charts.yaml
@@ -33,8 +33,8 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
         with:
-          python-version: "3.x"
-          check-latest: true
+          python-version: "3.11" # we need 3.11 as latest since ct is not compatible with > 3.12 yet
+          check-latest: false
 
       - name: Set up chart-testing
         uses: helm/chart-testing-action@0d28d3144d3a25ea2cc349d6e59901c4ff469b3b # v2.7.0
--- a/.github/workflows/release-automation.yaml
+++ b/.github/workflows/release-automation.yaml
@@ -131,8 +131,8 @@ jobs:
         with:
           allowUpdates: true
           commit: ${{ env.RELEASE_BRANCH }}
-          draft: ${{ steps.version_parser.outputs.prerelease == '' }}
-          generateReleaseNotes: true
+          draft: true
+          generateReleaseNotes: false
           name: ${{ env.RELEASE_TAG }}
           omitBody: ${{ steps.milestone.outputs.milestone_id == '' }}
           prerelease: ${{ steps.version_parser.outputs.prerelease != '' }}
--- a/.github/workflows/release-push-image.yaml
+++ b/.github/workflows/release-push-image.yaml
@@ -28,7 +28,7 @@ defaults:
 
 permissions:
   id-token: write
-  contents: read
+  contents: write
   packages: write
 
 env:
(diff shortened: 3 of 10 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -374,13 +375,18 @@ public void closeBlock(final BlockProof blockEndProof, final long handlerId) {
...
-            //     completed-but-not-forwarded blocks).
+            // Also log completed blocks metric and any other relevant
+            // actions. Also check if we have incomplete blocks lower than the
+            // block that completed, and possibly enter the resend process to
+            // have handlers go back and get the block that was too slow resent
+            // from a different publisher (don't forget to keep/track last
+            // completed block, and retain data in queue(s) for
+            // completed-but-not-forwarded blocks).
+            LOGGER.log(
+                    DEBUG,
```
ata-nas, line 386: That could be TRACE.

Also, printing the whole metric type seems very redundant to me, here an example if we print the whole metric:

`PlatformCounter[category=hiero_block_node,name=publisher_blocks_ack_sent,description=Block‑ack messages sent,unit=,format=%d,dataType=INT,value=0]
`

Instead, we could just get the value thereof. This to be considered everywhere.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -110,6 +113,7 @@ public void start() {
         context.blockMessaging()
                 .registerBlockNotificationHandler(
                         publisherManager, false, LiveStreamPublisherManager.class.getSimpleName());
+        LOGGER.log(TRACE, "StreamPublisherPlugin started successfully.");
```
ata-nas, line 116: Redundant, all plugins when successfully initialized and started, they get logged by the app.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -374,13 +375,18 @@ public void closeBlock(final BlockProof blockEndProof, final long handlerId) {
...
-            //     from a different publisher (don't forget to keep/track last
-            //     completed block, and retain data in queue(s) for
-            //     completed-but-not-forwarded blocks).
+            // Also log completed blocks metric and any other relevant
+            // actions. Also check if we have incomplete blocks lower than the
+            // block that completed, and possibly enter the resend process to
+            // have handlers go back and get the block that was too slow resent
+            // from a different publisher (don't forget to keep/track last
+            // completed block, and retain data in queue(s) for
+            // completed-but-not-forwarded blocks).
```
jsync-swirlds, line 384: This makes the todo end after the first line; the reformat is not correct.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Why are we logging metrics? The metric is already published in a more useful form.
This just wastes resources.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -553,6 +567,7 @@ private BatchHandleResult handleAccept(
      * Handle the SKIP action for a block.
      */
     private BatchHandleResult handleSkip(final long blockNumber) {
+        LOGGER.log(DEBUG, "Handler %d is sending SKIP for block %d".formatted(handlerId, blockNumber));
```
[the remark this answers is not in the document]
ata-nas: Ah, yes. That's absolutely true. Thanks for clarification.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -600,6 +615,7 @@ private BatchHandleResult handleEndBehind() {
      * Handle the END_DUPLICATE action for a block.
      */
     private BatchHandleResult handleEndDuplicate() {
+        LOGGER.log(DEBUG, "Handler {0} is sending DUPLICATE_BLOCK", handlerId);
```
jsync-swirlds: Same here, include the last known persisted block number:
```suggestion
        LOGGER.log(DEBUG, "Handler {0} is sending DUPLICATE_BLOCK({1}).", handlerId, publisherManager.getLatestBlockNumber());
```

PR #1755, opened 2025-10-15
chore: Multi-Publisher Metrics Dashboards

### LINKED ISSUE #1414: Multi-Publisher Metrics Dashboards
### Story Form

As a Block Node Operator/User
I want to have metrics dashboards
So that I can view live metrics

### Technical Notes

**DEPENDS ON**: #1370
- we are already using metrics counters and/or gauges
- we need the dashboards in the frontend

mustafauzunn:
## Reviewer Notes

Setting multi publisher metrics and adjusting dashboard

## Related Issue(s)
Fixes #1414 

P.S this dashboard data is from local run where simulator produce small and fast blocks. For that we have same values for block metrics

Screenshot 2025-10-16 at 15 07 54

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -347,7 +347,9 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             // This should result in new data being available, so we
             // count down the data ready latch.
             signalDataReady();
-            metrics.highestBlockNumber.set(blockNumber);
+            if (blockNumber > metrics.highestBlockNumber.get()) {
+                metrics.highestBlockNumber.set(blockNumber);
+            }
             // We're one of the handlers currently streaming, keep going.
             return BlockAction.ACCEPT;
         } else if (blockNumber == nextUnstreamedBlockNumber.get()) {


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -250,6 +250,7 @@ private BlockAction getActionForHeader(final long blockNumber, final long handle
      * todo(1420) add documentation
      */
     private BlockAction addHandlerQueueForBlock(final long blockNumber, final long handlerId) {
+        metrics.lowestBlockNumber.set(blockNumber);
```
Nana-EC: Here and below are we confident that the numbers will always match the expected lowest and highest.
Should we be checking the current is either < (lowest) or > (highest) that the previous?

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -741,6 +744,7 @@ void testHandleVerificationBadBlockProof() {
                         // below block number in the response is the latest known, -1L because none are stored
                         .returns(-1L, endStreamBlockNumberExtractor);
                 assertThat(responsePipeline.getOnCompleteCalls().get()).isEqualTo(1);
+                assertThat(sharedHandlerMetrics.endOfStreamsSent().get()).isEqualTo(1);
```
ata-nas: Let's not mix up metrics with other types of asserts. I love this addition, but we should really have the test where we solely assert metrics ideally. Also, it is very important in my opinion, that we assert all metrics always to ensure that we have the correct picture. Example is if we expect that only this metric is incremented, then all others should be at default values. If the test fails with unexpected metrics update, that could indicate we have an issue.

On another note, not sure if we should at all assert the shared handler metrics because these tests are for the manager. We have the shared metrics created because we need to instantiate publisher handlers. We should probably only assert manager metrics in this suite. Handler metrics have their own tests in the handler suite.

ata-nas (commented):
Good start, some items to consider. Like @Nana-EC, I also think we should rethink the way we handle metrics here.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -156,15 +159,15 @@ void setup() {
             // Initialize the historical block facility with the context.
             historicalBlockFacility.init(context, null);
             // Create the metrics hodler for the manager.
```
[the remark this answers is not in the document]
mustafauzunn: typo fixed

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
mustafauzunn: I see your point. But to have different test for metrics we need to copy/paste all the logic and just add the assert. I don't like this approach because of the duplication. Asserting all metrics will cause again duplication of code and these are unit tests which calls given method. There is no way wrong metric to be updated.

For the second paragraph if this is the right place for these asserts I wanted just to enhance tests. Currently we don't have tests in Handler for the `sendEndOfStream` method where we update the metric for example. Also I don't see why not to assert metrics in Manager if the code gets to this logic.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -374,6 +375,7 @@ public void closeBlock(final BlockProof blockEndProof, final long handlerId) {
             // here we just update metrics.
             metrics.blocksClosedIncomplete.increment();
         } else {
+            metrics.highestBlockNumber.set(blockEndProof.block());
```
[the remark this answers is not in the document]
mustafauzunn: Changed `highestBlockNumber` to be set on accept

PR #1762, opened 2025-10-16
feat(metrics): Performance metrics and Dashboard

### LINKED ISSUE #1763: Improve Performance Metrics
### Story Form

As a Block-Node Dev and Operator
I want to know where the latency of the BN is coming from
So that I can troubleshoot what ever is the bottleneck

### Technical Notes

During performance tests we noticed that the BN lags behind, and consider the issue was on verification plugin, on messaging ring buffer, on publisher handler, on helidon-pbj and it was hard to diagnose without good metrics, as is hard to calculate the current capacity of the system without metrics that gives us the AckLatency at any given TPS load.

AlfredoG87:
## Reviewer Notes

This introduces latency metrics for persistence, for hashing, fixes measurement for verification and adds in-transit and proof latency metrics.

minor improvements and some trace level logs to have better granularity that supports the metrics.

image
image

## Related Issue

Fixes #1763

### DIFF
--- a/docs/block-node/metrics.md
+++ b/docs/block-node/metrics.md
@@ -129,6 +129,7 @@ Observes inbound streams from publishers.
 | Counter | `publisher_blocks_resend_sent`           | Block Resend messages sent                          |
 | Counter | `publisher_block_endofstream_sent`       | Block End-of-Stream messages sent                   |
 | Counter | `publisher_block_endstream_received`     | Block End-Stream messages received                  |
+| Counter | `publisher_receive_latency_ns`           | Time taken to receive a block (ns=nanoseconds)      |
 
 ---
 
@@ -156,6 +157,7 @@ Measures block‑verification throughput and success rate.
 | Counter | `verification_blocks_failed`   | Blocks that failed verification              |
 | Counter | `verification_blocks_error`    | Internal errors during verification          |
 | Counter | `verification_block_time`      | Verification time per block (ns=nanoseconds) |
+| Counter | `hashing_block_time`           | Hashing time per block (ns=nanoseconds)      |
 
 ---
 
@@ -164,14 +166,15 @@ Measures block‑verification throughput and success rate.
 **Plugin:** `block-providers/files.recent [block-node-blocks-file-recent]`
 Activity and utilization of the recent on‑disk tier.
 
-|  Type   |                 Name                 |               Description               |
-|---------|--------------------------------------|-----------------------------------------|
-| Counter | `files_recent_blocks_written`        | Blocks written to recent tier           |
-| Counter | `files_recent_blocks_read`           | Blocks read from recent tier            |
-| Counter | `files_recent_blocks_deleted`        | Blocks deleted from recent tier         |
-| Counter | `files_recent_blocks_deleted_failed` | Blocks failed deletion from recent tier |
-| Gauge   | `files_recent_blocks_stored`         | Blocks stored in recent tier            |
-| Gauge   | `files_recent_total_bytes_stored`    | Bytes stored in recent tier             |
+|  Type   |                    Name                    |                  Description                   |
+|---------|--------------------------------------------|------------------------------------------------|
+| Counter | `files_recent_blocks_written`              | Blocks written to recent tier                  |
+| Counter | `files_recent_blocks_read`                 | Blocks read from recent tier                   |
+| Counter | `files_recent_blocks_deleted`              | Blocks deleted from recent tier                |
+| Counter | `files_recent_blocks_deleted_failed`       | Blocks failed deletion from recent tier        |
+| Gauge   | `files_recent_blocks_stored`               | Blocks stored in recent tier                   |
+| Gauge   | `files_recent_total_bytes_stored`          | Bytes stored in recent tier                    |
+| Counter | `files_recent_persistence_time_latency_ns` | Time taken to persist a block (ns=nanoseconds) |
 
 ---
 


block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -273,6 +283,12 @@ public void handleVerification(VerificationNotification notification) {
                     }
                 }
             }
+            final long endTime = System.nanoTime() - startTime;
```
jsync-swirlds: Naming is a little confusing
```suggestion
            final long totalTime = System.nanoTime() - startTime;
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -73,6 +75,9 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     private final AtomicLong currentStreamingBlockNumber;
     /** The unacknowledged yet blocks that were streamed to completion by this handler. */
     private final NavigableSet<Long> unacknowledgedStreamedBlocks;
+
+    private Map<Long, Long> blockStreamStartTimes = new HashMap<>();
+    private Map<Long, Long> blockStreamEndTimes = new HashMap<>();
```
jsync-swirlds: We absolutely should not be doing maps here, that's excessive cost.
I think there are some misunderstandings that lead to these two maps, and we should avoid that if possible.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherHandlerTest.java
```diff
@@ -2211,6 +2211,9 @@ private MetricsHolder createMetrics() {
                 new DefaultCounter(new Config("category", "name")),
                 new DefaultCounter(new Config("category", "name")),
                 new DefaultCounter(new Config("category", "name")),
+                new DefaultCounter(new Config("category", "name")),
+                new DefaultCounter(new Config("category", "name")),
+                new DefaultCounter(new Config("category", "name")),
```
jsync-swirlds: This test would be much easier to follow if the name or category matched the metric name, instead of all being the same value.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -397,6 +405,20 @@ public void sendAcknowledgement(final long newLastAcknowledgedBlockNumber) {
...
+            }
+            if (blockStreamEndTimes.containsKey(newLastAcknowledgedBlockNumber)) {
+                metrics.proofLatencyNs.add(System.nanoTime() - blockStreamEndTimes.get(newLastAcknowledgedBlockNumber));
+                blockStreamEndTimes.remove(newLastAcknowledgedBlockNumber);
+            }
+
+            LOGGER.log(
+                    TRACE,
+                    "Sent acknowledgement for block %d from handler %d"
+                            .formatted(newLastAcknowledgedBlockNumber, handlerId));
```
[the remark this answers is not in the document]
AlfredoG87: These are for redundancy, so we can correlate metrics with logs and help in diagnostics when Trace is enabled and we are troubleshooting

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -543,6 +568,15 @@ private void handleBlockProof(final Bytes blockProofBytes, final long blockNumbe
...
         }
+        long proofReceivedTime = System.nanoTime();
+        blockStreamEndTimes.put(blockNumber, proofReceivedTime);
+        // This metric measures the time taken from the start of streaming to the end of the streaming of a block.
+        // it can be thought of as the "streaming latency" of a block. "in-transit" and "network latency" is included
+        // here.
+        if (blockStreamStartTimes.containsKey(blockNumber)) {
+            var inTransitTime = proofReceivedTime - blockStreamStartTimes.get(blockNumber);
+            metrics.inTransitLatencyNs.add(inTransitTime);
+        }
```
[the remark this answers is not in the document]
AlfredoG87: Same as above, we can improve once implemented tracing.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/mode/impl/PublisherClientModeHandler.java
```diff
@@ -137,6 +137,11 @@ private void millisPerBlockStreaming() throws IOException, InterruptedException,
             long elapsedTime = System.nanoTime() - startTime;
             long timeToDelay = secondsPerBlockNanos - elapsedTime;
             if (timeToDelay > 0) {
+                LOGGER.log(
+                        INFO,
+                        "Streaming block took: " + (elapsedTime / 1_000_000) + "ms. Delaying for "
+                                + (timeToDelay / 1_000_000) + "ms to maintain a rate of " + millisecondsPerBlock
+                                + " milliseconds per block.");
```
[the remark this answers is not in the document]
AlfredoG87: This is for the simulator, but happy to optimize it

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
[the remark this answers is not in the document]
jsync-swirlds: That's kind of my point.  The logs should _produce_ the metrics (using the Loki and Grafana metrics queries feature), rather than adding this code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: We can avoid ever adding these by just logging at start and end with an easily parseable format.

e.g.
```java
    final String message = "metric-end-to-end-latency-by-block {0},{1}ns";
    LOGGER.log(TRACE, message, blockNumber, System.nanoTime());
```

Log that when the header starts and again when the block is acknowledged, and Loki can filter, parse, and calculate the duration.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
AlfredoG87: ✅ Done!

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
AlfredoG87: ✅  Done!

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -273,6 +283,12 @@ public void handleVerification(VerificationNotification notification) {
                     }
                 }
             }
+            final long totalTime = System.nanoTime() - startTime;
+            persistenceLatencyNs.add(totalTime);
+            LOGGER.log(
+                    TRACE,
+                    "Persistence Handle verification finished for block: %d, and it took %d ns to complete"
+                            .formatted(notification.blockNumber(), totalTime));
```
jsync-swirlds:
```suggestion
                    "Persistence Handle verification finished for block {0}, and it took {1} ns to complete", notification.blockNumber(), totalTime);
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -143,7 +146,9 @@ public void onSubscribe(@NonNull final Subscription subscription) {
     @Override
     public void onNext(@NonNull final PublishStreamRequestUnparsed request) {
         try {
+            LOGGER.log(TRACE, "Handler %d received request".formatted(handlerId));
```
jsync-swirlds:
```suggestion
            LOGGER.log(TRACE, "Handler {0} received request", handlerId);
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -397,6 +407,15 @@ public void sendAcknowledgement(final long newLastAcknowledgedBlockNumber) {
...
             metrics.blockAcknowledgementsSent.increment(); // @todo(1415) add label
+
+            if (LOGGER.isLoggable(TRACE)) {
+                final String traceMessage = "metric-end-to-end-latency-by-block-end {0},{1}ns";
+                LOGGER.log(TRACE, traceMessage, newLastAcknowledgedBlockNumber, System.nanoTime());
+                LOGGER.log(
+                        TRACE,
+                        "Sent acknowledgement for block %d from handler %d"
+                                .formatted(newLastAcknowledgedBlockNumber, handlerId));
+            }
```
jsync-swirlds: Same here
```suggestion
            final String ackMessage = "Sent acknowledgement for block {0} from handler {1}";
            final String traceMessage = "metric-end-to-end-latency-by-block-end {0},{1}ns";
            LOGGER.log(TRACE, traceMessage, newLastAcknowledgedBlockNumber, System.nanoTime());
            LOGGER.log(TRACE, askMessage, newLastAcknowledgedBlockNumber, handlerId);
            }
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

jsync-swirlds (dismissed):

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -225,6 +230,9 @@ private void handleBlockItemsRequest(
                 blockNumber = header.number();
                 // this means that we are starting a new block, so we can
                 // update the current streaming block number
+                final String traceMessage = "metric-end-to-end-latency-by-block-start {0},{1}ns";
+                LOGGER.log(TRACE, traceMessage, blockNumber, System.nanoTime());
+                currentStreamingBlockHeaderReceivedTime = System.nanoTime();
```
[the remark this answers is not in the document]
AlfredoG87: ✅  Done!

Nana-EC (approved):
LG, thanks for adopting the feedback.

jsync-swirlds:
Excellent work.

jsync-swirlds (approved):

PR #1764, opened 2025-10-16
feat(config): Exposes Helidon Http2 config for fine-tunning WebServer

### LINKED ISSUE #1759: Expose Helidon Http2 Protocol Configs
### Story Form

As a block node operator
I want to set the Helidon HTTP2 config values
So that I can configure my note appropriately based on my network and business needs

### Technical Notes

Helidon exposes `Http2Config` where you can customize the Http2 protocol settings usch as flowControl and windowSize - https://helidon.io/docs/v4/config/io_helidon_webclient_http2_Http2ClientProtocolConfig

For optimal performance these need to be exposed to an operator.
We should expose these either in the `ServerConfig` or a new `Http2Config` class

AlfredoG87:
## Reviewer Notes

- Adds a new Config Data Object WebServerHttp2Config that exposes Http2Config for the Helidon WebServer
- Uses the defined configurable properties provided when instantiating the webserver

## Related Issue(s)
Fixes #1759

### DIFF
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
@@ -7,6 +7,7 @@ import org.hiero.block.node.base.Loggable;
 
 /**
  * Use this configuration across the server features
+ * More info at <a href="https://helidon.io/docs/v4/apidocs/io.helidon.webserver.http2/io/helidon/webserver/http2/Http2Config.html">...</a>
  *
  * <p>ServerConfig will have settings for the server.
  *


block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
```diff
@@ -0,0 +1,35 @@
...
+ * @param maxEmptyFrames Maximum number of consecutive empty frames allowed on connection.
+ * @param maxFrameSize The size of the largest frame payload that the sender is willing to receive in bytes. Default value is 16384 and maximum value is 224-1 = 16777215 bytes. See RFC 9113 section 6.5.2 for details.
+ * @param maxHeaderListSize The maximum field section size that the sender is prepared to accept in bytes. See RFC 9113 section 6.5.2 for details. Default is 8192.
+ * @param maxRapidResets Maximum number of rapid resets(stream RST sent by client before any data have been sent by server). When reached within rapidResetCheckPeriod(), GOAWAY is sent to client and connection is closed. Default value is 50.
+ * @param rapidResetCheckPeriod Period for counting rapid resets(stream RST sent by client before any data have been sent by server). Default value is PT10S.
+ */
+@ConfigData("server.http2")
+public record WebServerHttp2Config(
+        @Loggable @ConfigProperty(defaultValue = "PT15S") String flowControlTimeout,
```
jsync-swirlds: This is super awkward syntax.  Why is this not just a number (e.g. 155) and interpreted as ms?

Also, 15 seconds is insanely long.  This is flow control, and typically should be similar magnitude to the network end-to-end latency.  In our usage 50ms is a good default, but anything from 10-100ms is reasonable.

It sets how long the server _waits_ before rechecking when the client is "full".

block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
```diff
@@ -0,0 +1,35 @@
...
+        @Loggable @ConfigProperty(defaultValue = "1048576") int initialWindowSize,
+        @Loggable @ConfigProperty(defaultValue = "8192") long maxConcurrentStreams,
```
jsync-swirlds: This will kill our RAM, as we need buffers for each concurrent stream.
Default should be very small (4 or 8).

block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
AlfredoG87: https://helidon.io/docs/v4/apidocs/io.helidon.webserver.http2/io/helidon/webserver/http2/Http2Config.html#flowControlTimeout()

Nana-EC: Fredy's just using the default values as specified by Helidon here per the offline team discussion.
We can change the unit to be in ms if it helps but this format is an example of an approach that allows the config to not enforce the unit of time. Duration just parses the value and allows for clearer easier spec e.g.PT15S vs 15_000 and having to read the variable name to know what unit it is.

I think leaving it as is, is fine

AlfredoG87: ✅  Done, changed it to `ms` instead of the horrible syntax. 

Also defaulted it to `50ms`

block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
AlfredoG87: Leaved it to 8, lets see how it does 🤞

jsync-swirlds: Thanks. Helidon defaults are definitely not what we need (hence why we have latency issues). Let's see how this goes with the changes.

jsync-swirlds (dismissed):

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -178,9 +183,23 @@ public class BlockNodeApp implements HealthFacility {
                 .name(PBJ_PROTOCOL_PROVIDER_CONFIG_NAME)
                 .maxMessageSizeBytes(serverConfig.maxMessageSizeBytes())
                 .build();
+
+        // Http2 Config
+        final Http2Config http2Config = Http2Config.builder()
+                .flowControlTimeout(Duration.ofMillis(webServerHttp2Config.flowControlTimeout()))
```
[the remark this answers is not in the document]
AlfredoG87: Thanks for your flexibility 🙏

block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.app.config;
+
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import org.hiero.block.node.base.Loggable;
+
+/**
+ * Use this configuration across the server features
```
Nana-EC: Could leave a link to the Helidon config page in comments here.
For later

Nana-EC (dismissed):
LG.
Follow up doc suggestions

jsync-swirlds (dismissed):

Nana-EC (approved):

PR #1782, opened 2025-10-23
chore: Enhance rolling log logic

### LINKED ISSUE #1774: Enhance Rolling Logs Support
### Story Form

As a block node operator
I want to ensure my logs roll over to new file as time goes
So that I have logs that manageable and predictably sized logs files over time

### Technical Notes

Confirm and update FileHandler log configs.
Ensure and set appropriate size and naming conventions

Nana-EC:
## Reviewer Notes

Add support for date based rolling log files.
Node operators engaging in load tests and trace logs will observe large multiple log files.
This is a node operator experience improvement to make management and configuration easier.

## Related Issue(s)
Fixes #1774

### DIFF
--- a/block-node/app/src/main/java/module-info.java
+++ b/block-node/app/src/main/java/module-info.java
@@ -29,5 +29,6 @@ module org.hiero.block.node.app {
     requires static transitive com.github.spotbugs.annotations;
     requires static java.compiler;
 
-    exports org.hiero.block.node.app.logging to java.logging;
+    exports org.hiero.block.node.app.logging to
+            java.logging;
 }
--- a/block-node/app/src/main/java/module-info.java
+++ b/block-node/app/src/main/java/module-info.java
@@ -28,4 +28,6 @@ module org.hiero.block.node.app {
     requires java.logging; // javax.annotation.processing.Generated
     requires static transitive com.github.spotbugs.annotations;
     requires static java.compiler;
+
+    exports org.hiero.block.node.app.logging to java.logging;
 }
--- a/block-node/app/src/main/java/org/hiero/block/node/app/logging/RollingFileHandler.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/logging/RollingFileHandler.java
@@ -2,8 +2,9 @@
 package org.hiero.block.node.app.logging;
 
 import java.io.IOException;
-import java.text.SimpleDateFormat;
-import java.util.Date;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.logging.FileHandler;
 import java.util.logging.LogManager;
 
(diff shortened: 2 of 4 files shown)


AlfredoG87 (dismissed):
Nice work. LGTM
talking about logs we should consider getting rid of the colorful formmater altogether, I think it adds unnecessary complexity and think that @jsync-swirlds agrees with me on this one.

Nana-EC:
> 2\. Using [LogRotate](https://linux.die.net/man/8/logrotate) in the O/S to rotate and compress log files on a schedule (e.g. hourly or daily). Further usage information is also available [here](https://www.baeldung.com/linux/rotating-logs-logrotate).
> 
> LogRotate, perhaps in combination with the basic rotation in the `FileHandler`, seems a much cleaner overall option with a lot more flexibility.

I agree LogRate in addition to FileHandler works fine. This update doesn't preclude a node operator from doing more refined management if they want. This is supposed to be a simple default config to make life easier for a node operator that doesn't customize the management.
I think as is is fine

Nana-EC:
> > > Did we consider either of the following rather than implementing this in Java code:
> > > 
> > > 1. Use the existing [FileHandler](https://docs.oracle.com/en/java/javase/21/docs/api/java.logging/java/util/logging/FileHandler.html) pattern and limits.
> > 
> > 
> > Yes for 1. FileHandler out of the box doesn't support the timestamp injection. It has a limited set of items you can inject into the pattern. If we don't care for the timestamp then FileHandler on it's won would have been fine, except block-node-0.log required you have to peek the file before knowing when the logs started.
> 
> The create timestamp for the file should let us know when a log file started (or the timestamp of the previous file). That's a pretty typical approach (using file timestamps) to at-a-glance determinations.
> 
> If we _really_ want thorough logfile management, we could also consider using the Socket handler and forward the data to journald or syslog so it's managed with other system-level log files.

To confirm you're suggesting using the more detailed `yyyy-MM-dd HH:mm:ss` instead of `yyyy-MM-dd` to zone in on the first log?

To your second point I think that's out of scope for this PR but a good consideration. Again this is focused on giving a working and simple default config that a node operator can use without extra overhead with the current logs being produced

jsync-swirlds:
> To confirm you're suggesting using the more detailed `yyyy-MM-dd HH:mm:ss` instead of `yyyy-MM-dd` to zone in on the first log? I get the idea but with the way FileHandler works it pulls the file name each time which would mean a new file every second. The value of the `%g` gives you and incrementing unique id. So you'll get 0, 1, 2 every time a new file name is needed because `java.util.logging.FileHandler.limit` has been exceeded on the previous file. The goal here is that `block-node-2025-10-29-0.log, block-node-2025-10-29-1.log, block-node-2025-10-29-2.log` would be more informative than `blocknode-0.log, blocknode-1.log, blocknode-2.log`

I'm not suggesting do more here, I'm saying that the O/S tracks file timestamps so it is not necessary to put that in the file name at all.  I do not think this entire PR is necessary, to be honest, we're fixing a problem with code that should be managed via system operations.

> 
> To your second point I think that's out of scope for this PR but a good consideration. Again this is focused on giving a working and simple default config that a node operator can use without extra overhead with the current logs being produced

I think it _replaces_ this PR.  That's been my point, we're writing new code for something that should be an O/S and Operations activity.

PR #1783, opened 2025-10-23
chore: Updated log formats to be more kv log metric friendly

### LINKED ISSUE #1775: Modify logs to support log metric telemetry
### Story Form

As a block node operator
I want to have logs that are well formatted when considering `key=value` cases
So that I can easily extract metrics where applicable and for easier reading

### Technical Notes

Currently many of our logs have cases where we logs a key and a value, however, they aren't consistent and sometimes hard to read or format.

Some telemetry options will support formatting of logs and more easily pick it up when written as a key=value pair.

Update observed locations of log metrics to make statements easier to format

Nana-EC:
## Reviewer Notes
A few areas can benefit from both readable and parseable logs with values.
Also some general improvements I happened upon

- Update a few log formats to be more key=value statement friendly
- defer to e over e.getMessage where log message formats are used
- format instances of block number and ns as integers when in log message
- fix a test cases name

## Related Issue(s)
Fixes #1775

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillGrpcClient.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillGrpcClient.java
@@ -177,7 +177,6 @@ public class BackfillGrpcClient {
                                 "Failed to fetch blocks from node address={0} port={1,number,#} error={2}",
                                 node.address(),
                                 node.port(),
-                                e.getMessage(),
                                 e);
 
                     } else {
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -196,7 +196,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
 
             LOGGER.log(TRACE, "Initialized gRPC client with sources path: {0}", blockNodeSourcesPath);
         } catch (Exception e) {
-            LOGGER.log(INFO, "Failed to initialize gRPC client: {0}", e.getMessage(), e);
+            LOGGER.log(INFO, "Failed to initialize gRPC client: {0}", e);
             hasBNSourcesPath = false;
             return;
         }
@@ -313,7 +313,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             }
             autonomousError = false;
         } catch (Exception e) {
-            LOGGER.log(TRACE, "Error during backfill autonomous process: {0}", e.getMessage(), e);
+            LOGGER.log(TRACE, "Error during backfill autonomous process: {0}", e);
             autonomousError = true;
         }
     }
@@ -511,7 +511,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
                         LOGGER.log(TRACE, "Starting on-demand backfill for gap: {0}", gap);
                         backfillGap(gap, BackfillType.ON_DEMAND);
                     } catch (ParseException | InterruptedException | RuntimeException e) {
-                        LOGGER.log(TRACE, "Error backfilling gap {0}: {1}", gap, e.getMessage(), e);
+                        LOGGER.log(TRACE, "Error backfilling gap {0}: {1}", gap, e);
                         backfillFetchErrors.add(1);
                         onDemandError = true;
                         onDemandBackfillStartBlock.set(-1); // Reset on error to allow new backfills
@@ -519,7 +519,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
                 });
 
             } catch (RuntimeException e) {
-                LOGGER.log(TRACE, "Error scheduling backfill for gap {0}: {1}", gap, e.getMessage(), e);
+                LOGGER.log(TRACE, "Error scheduling backfill for gap {0}: {1}", gap, e);
                 backfillFetchErrors.add(1);
                 onDemandError = true;
                 onDemandBackfillStartBlock.set(-1); // Reset on error to allow new backfills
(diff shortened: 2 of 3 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -457,7 +459,8 @@ private boolean sendResponse(final PublishStreamResponse response) {
             return false;
         } catch (final RuntimeException e) {
             shutdown(); // this method is idempotent and can be called multiple times
-            final String message = "Failed to send response for handler %d: %s".formatted(handlerId, e.getMessage());
+            final String message =
+                    "Failed to send response '%s' for handler %d: %s".formatted(response, handlerId, e.getMessage());
```
jsync-swirlds: Not a significant issue, but including the response object here is probably excessive.
The response _type_ is probably all that's useful.

Also, including `e` instead of just `e.getMessage()` is generally recommended.  Many common exceptions have null messages because the exception _type_ is enough for normal usage.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -33,19 +33,19 @@ public class ExtendedMerkleTreeSession implements VerificationSession {
...
-        LOGGER.log(INFO, "Created ExtendedMerkleTreeVerificationSessionV0680 for block {}", blockNumber);
+        LOGGER.log(INFO, "Created ExtendedMerkleTreeVerificationSessionV0680 for block {0}", blockNumber);
     }
 
     // todo(1661) implement the real logic here, for now just return true if last item has block proof.
     @Override
     public VerificationNotification processBlockItems(List<BlockItemUnparsed> blockItems) throws ParseException {
         this.blockItems.addAll(blockItems);
-        LOGGER.log(TRACE, "Processed {} block items for block {}", blockItems.size(), blockNumber);
+        LOGGER.log(TRACE, "Processed {0} block items for block {1}", blockItems.size(), blockNumber);
```
jsync-swirlds: Thanks for fixing these.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -291,15 +291,19 @@ private void detectGaps() {
                     LongRange gap = new LongRange(previousRangeEnd + 1, range.start() - 1);
                     detectedGaps.add(gap);
                     pendingBlocks += gap.size();
-                    LOGGER.log(TRACE, "Detected gap in historical blocks from {0} to {1}", gap.start(), gap.end());
+                    LOGGER.log(
+                            TRACE,
+                            "Detected gap in historical blocks from start={0} to end={1}",
```
jsync-swirlds:
```suggestion
                            "Detected gap in historical blocks from start={0,number,#} to end={1,number,#}",
```

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -511,15 +511,15 @@ public void handleNewestBlockKnownToNetwork(NewestBlockKnownToNetworkNotificatio
...
+                        LOGGER.log(TRACE, "Error backfilling gap {0}: {1}", gap, e.getMessage(), e);
                         backfillFetchErrors.add(1);
                         onDemandError = true;
                         onDemandBackfillStartBlock.set(-1); // Reset on error to allow new backfills
                     }
                 });
 
             } catch (RuntimeException e) {
-                LOGGER.log(TRACE, "Error scheduling backfill for gap {0}: {1}", gap, e.getMessage());
+                LOGGER.log(TRACE, "Error scheduling backfill for gap {0}: {1}", gap, e.getMessage(), e);
```
jsync-swirlds:
```suggestion
                LOGGER.log(TRACE, "Error scheduling backfill for gap {0}: {1}", gap, e);
```

Nana-EC:
> Some suggestions that might be best added as a batch.
> 
> Mostly this is just _replacing_ `e.getMessage()` with `e`. The reasoning:
> 
> * `Throwable.toString()` produces `ExceptionClass: Message`, which is better than, and a super set of, `Message` produced by `getMessage()`.
> * Logging an extra exception at the end of a message list does nothing, unfortunately, as the `System.Logger` just calls `toString` on every parameter then inserts them into the message format.
>   
>   * Including the stack trace _only_ happens if the `Throwable` is the only parameter after the message.

Aww man. Funny i had it as such before but modified. May be missing up log4j or other loggers.
I've just pushed it, just had to CTRL Z in a few places

jsync-swirlds (approved):

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillGrpcClient.java
```diff
@@ -170,10 +174,11 @@ public List<BlockUnparsed> fetchBlocks(LongRange blockRange) {
 
                         LOGGER.log(
                                 INFO,
-                                "Failed to fetch blocks from node address={0} port={1} error={2}",
+                                "Failed to fetch blocks from node address={0} port={1,number,#} error={2}",
                                 node.address(),
                                 node.port(),
-                                e.getMessage());
+                                e.getMessage(),
+                                e);
```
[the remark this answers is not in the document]
Nana-EC: Done

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
Nana-EC: Done

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -441,7 +443,7 @@ private boolean sendResponse(final PublishStreamResponse response) {
             long start = System.nanoTime();
             replies.onNext(response);
             long duration = System.nanoTime() - start;
-            LOGGER.log(DEBUG, "Handler {0} replies.onNext took {1} ns", handlerId, duration);
+            LOGGER.log(DEBUG, "Handler {0} replies.onNext took {1} ns to send {2}", handlerId, duration, response);
```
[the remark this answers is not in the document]
Nana-EC: Done

AlfredoG87 (approved):
LGTM

PR #1790, opened 2025-10-24
chore:  Update Observability Stack to latest versions

### LINKED ISSUE #1210: Update Observability Stack to latest versions
### Story Form

As a Block Node Operator
I want to Install the BN using the Observability stack that comes with the charts by default for convenience but want the latest versions with all the security updates
So that I can protect my deployments from known attacks on older versions.

### Technical Notes

We were notified of this [vulnerability](https://grafana.com/blog/2025/05/21/grafana-security-release-high-severity-security-fix-for-cve-2025-4123/) on grafana.

Make sure while updating the version that this vulnerability is patched in the latest version of grafana.

As part of this effort it would be good to document how to make this sort of a change in the future.

mustafauzunn:
## Reviewer Notes
- Update kube-prometheus-stack to version 78.4.0
- Add missing dashboards

## Related Issue(s)
Fixes #1210


jsync-swirlds (approved):
That's a big version jump.
I wonder why dependabot doesn't catch it...

AlfredoG87 (changes requested):
Provisioned Dashboards are showing **No Data**, there must be update to be compatible with the latest grafana version. Everything continues to work, is just about updating the json files.

PR #1795, opened 2025-10-24
feat: Add end-of-block messages to subscriber and publisher plugins.

### LINKED ISSUE #1626: Add end-of-block message to the BN Publisher API and internal messaging
### Story Form

As a Block Node
I want to know affirmatively when a block ends
So that I can support multiple block proofs and added-after-the-fact block proofs.

### Technical Notes

We intend to add a new message in the API for block node that indicates the end of a block so that we can add multiple block proof items in a block stream, and even add those items to a block _after_ it is stored.  This is part of supporting migration and enhancement of proof structures in the indeterminate future, and also supports private Hiero deployments which may need multiple assertion types, depending on use and purpose.

jsync-swirlds:
## Reviewer Notes
* Added sending an EndBlock message after each block in the Subscriber plugin.
* Added end of block support to publisher
* Removed block proof handling from publisher
    * We no longer need to look for block proofs; just the end of block message.
* Added `EndOfBlock` requests to `LiveStreamPublisherManagerTest`
* Added `EndOfBlock` requests to `StreamPublisherPluginTest`
* Updated `SubscriberServicePluginTest` to account for end of block messages.

## Related Issue(s)
 Resolves #1626

### DIFF
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
@@ -6,6 +6,7 @@ import com.hedera.pbj.grpc.helidon.config.PbjConfig;
 import io.helidon.webserver.ConnectionConfig;
 import io.helidon.webserver.WebServer;
 import io.helidon.webserver.WebServerConfig;
+import org.hiero.block.api.BlockEnd;
 import org.hiero.block.api.BlockItemSet;
 import org.hiero.block.api.BlockNodeServiceInterface;
 import org.hiero.block.api.BlockStreamSubscribeServiceInterface;
@@ -41,6 +42,10 @@ public class TestBlockNodeServer {
                                                     .block()
                                                     .items())
                                             .build())
+                                    .build());
+                            replies.onNext(SubscribeStreamResponse.newBuilder()
+                                    .endOfBlock(
+                                            BlockEnd.newBuilder().blockNumber(i).build())
                                     .build());
                         }
 
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -8,7 +8,6 @@ import static java.lang.System.Logger.Level.WARNING;
 import static org.hiero.block.node.spi.BlockNodePlugin.METRICS_CATEGORY;
 import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
 
-import com.hedera.hapi.block.stream.BlockProof;
 import com.hedera.pbj.runtime.grpc.Pipeline;
 import com.swirlds.metrics.api.Counter;
 import com.swirlds.metrics.api.LongGauge;
@@ -321,6 +320,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
      * when interrupt is used as a signal rather than signaling the `Condition`
      * variable.</blockquote>
      */
+    @SuppressWarnings("AwaitNotInLoop")
     private void waitForDataReady() {
         dataReadyLock.lock();
         try {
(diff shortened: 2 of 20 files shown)


jsync-swirlds:
> Looking good, just 1 question.
> 
> If I understand this PR correctly, this cannot be merged/published before the CN starts sending the endOfBlock Message right?
> 
> we won't be keeping backwards compatibility so to speak right?

Basically correct.
Once this PR is merged, the CN must start sending `EndOfBlock`.
&nbsp;&nbsp;&nbsp;The release with this PR included will carry a "must match" note, and applies (hopefully) to CN `0.68` or a pre-release build of `0.69`.
&nbsp;&nbsp;&nbsp;We've been _accepting_ the `EndOfBlock` for a while, and I believe that the CN is already sending that message (we had a fix in `0.21` for that exact item).
There is no expectation that this particular change is backward-compatible.

mustafauzunn (commented):
We will need logic handling the endOfBlock in `ConsumerStreamObserver` in simulator too

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -176,13 +177,22 @@ private void processNextRequestUnparsed(final PublishStreamRequestUnparsed reque
...
-            // @todo(1626) Do nothing for now.
+            handleEndOfBlock(request.endOfBlock());
         } else {
             // this should never happen
             sendEndAndResetState(Code.ERROR);
         }
     }
 
+    private void handleEndOfBlock(final BlockEnd endOfBlock) {
+        final long blockNumber = endOfBlock.blockNumber();
```
ata-nas: Might be beneficial if we verify the request, the block number has to be a whole number, otherwise that request should be invalid, we could close the connection.

Another thing would be, we should probably check if this block number matches the current streaming block number of this handler. If not, then that is problematic as well.

jsync-swirlds: I need to think on the matching part; there was a reason I did not do that here.

As to whole number; it's uint64 in protobuf, it won't parse if it's not a whole number. It might be above (2<sup>63</sup>-1), which looks like it's negative in Java, but that's a Java-ism we can mostly ignore for now.

block-node/stream-subscriber/src/test/java/org/hiero/block/node/stream/subscriber/SubscriberServicePluginTest.java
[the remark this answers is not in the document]
jsync-swirlds: If you're talking about the messages in the pipeline, we might (as a separate issue) consider making some improvements to check for correct end of block, but that's a bit beyond the scope of the current Issue.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -71,8 +71,9 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     private final long handlerId;
     /** The current streaming block number. This is used to track the current block being streamed. */
     private final AtomicLong currentStreamingBlockNumber;
-    /** The unacknowledged yet blocks that were streamed to completion by this handler. */
+    /** The unacknowledged blocks that were streamed to completion by this handler. */
     private final NavigableSet<Long> unacknowledgedStreamedBlocks;
+    // @todo() remove this (and its usage) and use telemetry or metrics queries instead
```
ata-nas: Do we have an issue number by any chance?

ata-nas (approved):

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: We can have multiple overlapping calls from the same publisher in some rare cases because the publisher may encounter failure modes (has been seen in testing).
To ensure we're always correct with how we handle end of block, we want to be sure that a close block closes the current open block, even if it's not necessarily "correct" so that the publisher plugin doesn't get stuck, and verification can determine if it was a timing bug in the publisher or invalid block.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
Nana-EC: nit
```suggestion
                        // publishes each block items and an EndOfBlock response for each block
                        for (long i = request.startBlockNumber(); i <= request.endBlockNumber(); i++) {
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -321,6 +320,7 @@ private void signalDataReady() {
      * when interrupt is used as a signal rather than signaling the `Condition`
      * variable.</blockquote>
      */
+    @SuppressWarnings("AwaitNotInLoop")
```
Nana-EC: Odd, what change required this?

AlfredoG87 (approved):
LGTM, great work! 💯

PR #1816, opened 2025-10-30
refactor: Files Historic Plugin Proceeds With Next Batch If Current is Unsuccessful

### LINKED ISSUE #1758: Files Historic Plugin Proceeds With Next Batch If Current is Unsuccessful
### Story Form

As a Block Node Operator
I want to ensure that my archive plugin continues to function if it fails to archive a batch and notifies me for the failure
So that I will not lose the next batches and be able to take an action for the failed one

### Technical Notes

- currently the files.historic plugin does not allow gaps in the zips and also does not proceed with the next batch until it finishes this one
- effectively this means that if we cannot supply a block for the current batch, the archive will "halt"
- a gap is not the only condition for a failed batch
- no matter why a batch may have failed, that should not halt the processing of the next batches
- an alert has to be sent (i.e. log error) when a batch has failed to archive
- we also need a way to ensure that we do not lose a batch we have failed to archive

We need to allow the files.historic plugin to continue with the next batch if a failure with the current one occurs. We also need to alert operators for the failure. Also, we need to ensure that we do not lose batches that we have failed to archive. Furthermore, the failed batch needs to be "set aside" to try again later (and in future updates we might send a notification to request high priority backfill of the missing blocks).

mustafauzunn:
## Reviewer Notes

- `files.historic` should be prompted by Verification notification
- `files.historic` should collect blocks in staging folder before zipping
- change log level to `ERROR` if we have a gap
- add unit test proving that we can zip after failed batch for zipping
- refactor tests to work with the new logic

## Related Issue(s)
Fixes #1758

### DIFF
--- a/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
+++ b/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
@@ -254,7 +254,7 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
 
         BlockItemUnparsed firstItem = block.blockItems().getFirst();
         if (!firstItem.hasBlockHeader()) {
-            LOGGER.log(INFO, "Block {0} has no block header, cannot write to staging path", blockNumber);
+            LOGGER.log(WARNING, "Block {0} has no block header, cannot write to staging path", blockNumber);
             return;
         }
 
@@ -292,8 +292,8 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
         } catch (final IOException e) {
             final String message = "Failed to create directories for path %s due to %s"
                     .formatted(verifiedBlockPath.toAbsolutePath(), e);
-            LOGGER.log(WARNING, message, e);
-            throw new UncheckedIOException(e);
+            LOGGER.log(INFO, message, e);
+            throw new UncheckedIOException(e.getMessage(), e);
         }
     }
 
@@ -436,7 +436,8 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
             for (long blockNumber = batchFirstBlockNumber; blockNumber <= batchLastBlockNumber; blockNumber++) {
                 Path path = BlockFile.nestedDirectoriesBlockFilePath(
                         stagingPath, blockNumber, config.compression(), config.maxFilesPerDir());
-                final BlockAccessor currentAccessor = new BlockFileAccessor(path, CompressionType.ZSTD, blockNumber);
+                final BlockAccessor currentAccessor =
+                        new BlockStagingFileAccessor(path, CompressionType.ZSTD, blockNumber);
                 if (currentAccessor.block() == null) {
                     break;
                 } else {
@@ -448,7 +449,7 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
                 // we have a gap in the batch, so we cannot zip it
                 final String message = "Batch of blocks [%d -> %d] has a gap, skipping zipping"
                         .formatted(batchFirstBlockNumber, batchLastBlockNumber);
-                LOGGER.log(ERROR, message);
+                LOGGER.log(INFO, message);
             } else {
                 // move the batch of blocks to a zip file
                 try {
--- a/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockStagingFileAccessor.java
+++ b/block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockStagingFileAccessor.java
@@ -18,11 +18,11 @@ import org.hiero.block.node.base.CompressionType;
 import org.hiero.block.node.spi.historicalblocks.BlockAccessor;
 
 /**
- * An implementation of the {@link BlockAccessor} interface that provides access to a block stored in a file with
+ * An implementation of the {@link BlockAccessor} interface that provides access to blocks stored in the staging folder with
  * optional compression types on that file. It aims to provide the most efficient transfer for each combination of
  * input and output formats.
  */
-final class BlockFileAccessor implements BlockAccessor {
+final class BlockStagingFileAccessor implements BlockAccessor {
     /** Message logged when the protobuf codec fails to parse data */
     private static final String FAILED_TO_PARSE_MESSAGE = "Failed to parse block from file %s.";
     /** Message logged when data cannot be read from a block file */
@@ -37,13 +37,13 @@ final class BlockFileAccessor implements BlockAccessor {
     private final long blockNumber;
 
     /**
-     * Constructs a BlockFileAccessor with the specified block file path and compression type.
+     * Constructs a BlockStagingFileAccessor with the specified block file path and compression type.
      *
      * @param blockFilePath   the path to the block file, must exist
      * @param compressionType the compression type used for the block file
      * @param blockNumber     the block number of the block
      */
-    BlockFileAccessor(
+    BlockStagingFileAccessor(
             @NonNull final Path blockFilePath, @NonNull final CompressionType compressionType, final long blockNumber) {
         this.blockFilePath = Preconditions.requireRegularFile(blockFilePath);
         this.compressionType = Objects.requireNonNull(compressionType);


block-node/block-providers/files.historic/src/test/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPluginTest.java
```diff
@@ -276,6 +276,39 @@ void testZipRangeHappyPathArchivalBatchAndAHalf() throws IOException {
...
+                final BlockItemUnparsed[] block = SimpleTestBlockItemBuilder.createSimpleBlockUnparsedWithNumber(i);
+                testHistoricalBlockFacility.handleBlockItemsReceived(new BlockItems(List.of(block), i), false);
+            }
+            // assert that none of the first 20 blocks are zipped yet
+            for (int i = 0; i < 20; i++) {
+                assertThat(BlockPath.computeExistingBlockPath(testConfig, i)).isNull();
+            }
+            // send a block persisted notification for the range we just created
+            blockMessaging.sendBlockPersisted(
+                    new PersistedNotification(19, true, toTest.defaultPriority() + 1, BlockSource.PUBLISHER));
```
jsync-swirlds: Why is this plugin relying on persisted notifications?
It should only rely on _Verification_ notification (that is, it shouldn't be reliant on another persistence plugin).

This might be something that needs to be addressed in a different PR, however.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -204,37 +229,93 @@ public BlockRangeSet availableBlocks() {
...
+        Bytes headerBytes = block.blockItems().getFirst().blockHeader();
+        try {
+            return BlockHeader.PROTOBUF.parse(headerBytes);
+        } catch (final ParseException e) {
+            LOGGER.log(INFO, "Failed to parse block header", e);
+            return null;
+        }
+    }
+
+    private void sendBlockNotification(final long number, final boolean succeeded, final BlockSource source) {
```
Nana-EC:
```suggestion
    private void sendBlockPersistedNotification(final long number, final boolean succeeded, final BlockSource source) {
```

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -204,37 +229,93 @@ public BlockRangeSet availableBlocks() {
...
+            cleanup();
         } catch (final RuntimeException e) {
             final String message = "Failed to handle persistence notification due to %s".formatted(e);
             LOGGER.log(WARNING, message, e);
         }
     }
 
+    private void writeBlockToLivePath(final BlockUnparsed block, final long blockNumber, final BlockSource source) {
+        final BlockSource effectiveSource = source == null ? UNKNOWN : source;
+        if (block != null && block.blockItems() != null && !block.blockItems().isEmpty()) {
```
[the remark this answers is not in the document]
ata-nas: I agree we should try to aim for simplicity here. The logic that is fitting for the other plugin may not necessarily be fitting here.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileBlockAccessor.java
```diff
@@ -0,0 +1,165 @@
...
+import org.hiero.block.internal.BlockUnparsed;
+import org.hiero.block.node.base.CompressionType;
+import org.hiero.block.node.spi.historicalblocks.BlockAccessor;
+
+/**
+ * An implementation of the {@link BlockAccessor} interface that provides access to a block stored in a file with
+ * optional compression types on that file. It aims to provide the most efficient transfer for each combination of
+ * input and output formats.
+ */
+final class BlockFileBlockAccessor implements BlockAccessor {
```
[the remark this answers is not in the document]
Nana-EC: Okay, the notion of a base block accessor was what I was going on based on if truly all the logic was useful and advisable to multiple plugins.
If this contains items that are not applicable then no need to share as we do want plugins independent.

We should strip this down to what is needed

jsync-swirlds:
> @AlfredoG87 we should also update charts with volume for the staging folder right?

@mustafauzunn: There shouldn't be a separate volume, the staging folder is placed where it is to avoid needing another volume.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -98,6 +111,17 @@ public void init(final BlockNodeContext context, final ServiceBuilder serviceBui
             LOGGER.log(ERROR, "Could not create root directory", e);
             context.serverHealth().shutdown(name(), "Could not create root directory");
         }
+        try {
+            Files.createDirectories(config.stagingPath());
+        } catch (IOException e) {
+            LOGGER.log(ERROR, "Could not create temp directory", e);
+            context.serverHealth().shutdown(name(), "Could not create temp directory");
```
Nana-EC:
```suggestion
            context.serverHealth().shutdown(name(), "Could not create staging directory");
```

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -49,6 +60,8 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
     private int numberOfBlocksPerZipFile;
     /** The set of available blocks. */
     private final ConcurrentLongRangeSet availableBlocks = new ConcurrentLongRangeSet();
+    /** The set of available temporary blocks (not yet zipped). */
+    private final ConcurrentLongRangeSet availableTemporaryBlocks = new ConcurrentLongRangeSet();
```
Nana-EC:
```suggestion
    private final ConcurrentLongRangeSet availableStagedBlocks = new ConcurrentLongRangeSet();
```

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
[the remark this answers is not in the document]
Nana-EC: I'm not sure how the complexity of the method relates to the number of files to be zipped in a batch.
The algorithm should be able to be simplified regardless of how large the batch is right?
We're highlighting that the complexity if the flow can hide bugs and considerations and make testing and maintenance harder

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -389,6 +455,16 @@ private void moveBatchOfBlocksToZipFile(final LongRange batchRange) {
                     totalBytesStored.addAndGet(zipFileSize);
                     // Increment the blocks written counter
                     blocksWrittenCounter.add(numberOfBlocksPerZipFile);
+
+                    for (long blockNumber = batchFirstBlockNumber; blockNumber <= batchLastBlockNumber; blockNumber++) {
+                        Path path = BlockFile.nestedDirectoriesBlockFilePath(
```
Nana-EC: Q: Will this mean every time zipping is attempted we walk the whole staging directory?
Isn't `availableTemporaryBlocks` available to avoid that?
I expected we'd walk the dir once during init, then form then the plugin add to the list as blocks are verified and clears the entry list based on successful zipping

Nana-EC (changes requested):
A few clarifications and potential improvements

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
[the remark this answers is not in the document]
mustafauzunn: I thought you were talking about logic for block file creation and structure.

For the logic itself, I agree we can do stuff like early failure return and so.

block-node/block-providers/files.historic/src/test/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPluginTest.java
mustafauzunn: The logic is refactored and we use Verification notification

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -204,38 +231,87 @@ public BlockRangeSet availableBlocks() {
...
 
     // ==== Private Methods ============================================================================================
+    private void writeBlockToStagingPath(final BlockUnparsed block, final long blockNumber) {
+        if (block == null || block.blockItems() == null || block.blockItems().isEmpty()) {
+            return;
+        }
+
+        BlockItemUnparsed firstItem = block.blockItems().getFirst();
+        if (!firstItem.hasBlockHeader()) {
+            LOGGER.log(INFO, "Block {0} has no block header, cannot write to staging path", blockNumber);
```
jsync-swirlds: This might be `WARNING` level, as we have a block that somehow passed verification with no block header (which should not be possible).

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileAccessor.java
```diff
@@ -0,0 +1,146 @@
...
+import org.hiero.block.common.utils.Preconditions;
+import org.hiero.block.node.base.CompressionType;
+import org.hiero.block.node.spi.historicalblocks.BlockAccessor;
+
+/**
+ * An implementation of the {@link BlockAccessor} interface that provides access to a block stored in a file with
+ * optional compression types on that file. It aims to provide the most efficient transfer for each combination of
+ * input and output formats.
+ */
+final class BlockFileAccessor implements BlockAccessor {
```
jsync-swirlds: Just to help keep things straight, perhaps rename to `BlockStagingFileAccessor` and adjust Javadoc to match?
Otherwise it might get a little confusing.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -375,7 +448,7 @@ private void moveBatchOfBlocksToZipFile(final LongRange batchRange) {
                 // we have a gap in the batch, so we cannot zip it
                 final String message = "Batch of blocks [%d -> %d] has a gap, skipping zipping"
                         .formatted(batchFirstBlockNumber, batchLastBlockNumber);
-                LOGGER.log(DEBUG, message);
+                LOGGER.log(ERROR, message);
```
jsync-swirlds: Why would we make this an `ERROR` level? It seems to me that this should still be DEBUG; INFO at most.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
jsync-swirlds: Note also, this just walks _the recently archived batch_ of files and deletes them all.
It doesn't scan the whole folder (that has to happen somewhere, but I'm not sure where).

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
mustafauzunn: Agreed and refactored

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -204,38 +231,87 @@ public BlockRangeSet availableBlocks() {
...
+
+    private void createDirectoryOrFail(final Path verifiedBlockPath) {
+        try {
+            // create parent directory if it does not exist
+            Files.createDirectories(verifiedBlockPath.getParent());
+        } catch (final IOException e) {
+            final String message = "Failed to create directories for path %s due to %s"
+                    .formatted(verifiedBlockPath.toAbsolutePath(), e);
+            LOGGER.log(WARNING, message, e);
+            throw new UncheckedIOException(e);
```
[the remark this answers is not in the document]
mustafauzunn: Refactored

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
mustafauzunn: Sure
This is the task https://github.com/hiero-ledger/hiero-block-node/issues/1858

Issue #1805, opened 2025-10-29
NPE in S3ArchivePlugin when HistoricalBlockFacility cannot retrieve a block


Nana-EC:
### Description

In recent previewnet testing we observed an NPE in the S3ArchivePlugin which was preventing tar uploads

```
java.lang.NullPointerException: Cannot invoke "org.hiero.block.node.spi.historicalblocks.BlockAccessor.blockUnparsed()" because the return value of "org.hiero.block.node.spi.historicalblocks.HistoricalBlockFacility.block(long)" is null
2025-10-28 00:46:01.806	at org.hiero.block.node.archive.s3cloud@0.21.1/org.hiero.block.node.archive.s3.S3ArchivePlugin.uploadBlocksTar(S3ArchivePlugin.java:250)
```

The issue seems to be because the block being accessed is not present on the BN.

### Steps to Reproduce

1. Enable S3ArchivePlugin with appropriate configs
2. Set earliest managed blocknode above the initial range of 0-99999
3. Start block node
4. Publish blocks with a starting block above 99999
5. observe error

### Additional context

This is a valid case for Tier 2 and testing scenarios but not expected to occur in Tier 1.

S3 plugin should have similar behaviour to the resolution for https://github.com/hiero-ledger/hiero-block-node/issues/1758. That is it should be resilient and progress forward

Issue #1629, opened 2025-09-16
bug(archive): Unable to upload archive block batch


AlfredoG87:
### Description

Did a GCP deployment for `previewnet-archive0` and connected to a local solo deployment and it is collecting blocks.

Since this is intended to test out the `archive` path flow, I set the Batch size of an archive batch to `1000` instead of the default `100_000` blocks.

when my CN solo deploy reached the 1000 blocks it failed with the following error:

```
2025-09-16 00:38:41.582+0000 INFO    [org.hiero.block.node.archive.s3.S3ArchivePlugin lambda$scheduleBatchArchiving$1] Uploading archive block batch: 0-999
2025-09-16 00:38:41.643+0000 SEVERE  [org.hiero.block.node.archive.s3.S3ArchivePlugin lambda$scheduleBatchArchiving$1] Failed to upload archive block batch: null
java.lang.NullPointerException
```

### Steps to Reproduce

reproduce k8s deployment with archive configuration as noted here:
https://www.notion.so/swirldslabs/Simulation-of-Deployment-of-Partial-History-Node-for-Previewnet-Testnet-2637c9ab259180d7ad8ee8bba87303be#2637c9ab259180a590c1d02621f31881

### Additional context

see complete logs

[previewnet-arvhive0-0.18.0.log](https://github.com/user-attachments/files/22353543/previewnet-arvhive0-0.18.0.log)

AlfredoG87:
Also there is an issue with block `53` of this run, maybe it became corrupted... bug still showing in `0.18.0` :(

PR #1827, opened 2025-11-04
fix: Unable to Upload Archive Block Batch

ata-nas:
* Refactored the plugin and extracted an upload task
* Tests now run async
* We no longer silence exceptions from an upload run
* Introduced an executor service to help us manage uploads
* Cleaned up message formatting
* Best effort attempt (for now) to mitigate NPEs when gaps are detected
* Added tests:
  * Test for not throwing exceptions when an upload task completed
    exceptionally

Explores #1629 
Fixes: #1805  

## Reviewer Notes

With this PR I aim to refactor the plugin so we no longer silence exceptions and we have a much cleaner plugin to run with.

Here is an artificially induced NPE (I hard-coded `null` as the value for the block iterator in the tarred iterator) just to showcase what we will start seeing from now on:

image

Since we do not know yet the actual cause of the NPE (my guess is that a block was missing so a block accessor was mapped to null), we cannot really solve the issue, we see the log for the missing block accessor, but also a simple null afterwards which has no stacktrace.

We should probably proceed with this intermediate improvement and run the deployments again where we expect to hit the actual exception.

### DIFF
--- a/block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
+++ b/block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
@@ -255,7 +255,7 @@ public class S3ArchivePlugin implements BlockNodePlugin, BlockNotificationHandle
         private static final String END_UPLOAD_MESSAGE = "Finished uploading archive block batch: {0} - {1}";
 
         @Override
-        public Void call() throws Exception {
+        public Void call() throws S3ClientInitializationException, S3ResponseException, IOException {
             // Check the nextBatchStartBlockNumber is less than lastArchivedBlockNumber, ie is this a duplicate
             // batch and
             // already been archived. That can happen as the archive job has not been completed for a batch and
@@ -266,16 +266,23 @@ public class S3ArchivePlugin implements BlockNodePlugin, BlockNotificationHandle
                 LOGGER.log(DEBUG, START_UPLOAD_MESSAGE, nextBatchStartBlockNumber, nextBatchEndBlockNumber);
                 try (final S3Client s3Client = createClient(archiveConfig)) {
                     // fetch the blocks from the persistence plugins and archive
-                    uploadBlocksTar(s3Client, nextBatchStartBlockNumber, nextBatchEndBlockNumber);
-                    // write the latest archived block number to the archive
-                    s3Client.uploadTextFile(
-                            LATEST_ARCHIVED_BLOCK_FILE,
-                            LATEST_FILE_STORAGE_CLASS,
-                            String.valueOf(nextBatchEndBlockNumber));
-                    // update the last archived block number
-                    lastArchivedBlockNumber.set(nextBatchEndBlockNumber);
-                    // log completed
-                    LOGGER.log(DEBUG, END_UPLOAD_MESSAGE, nextBatchStartBlockNumber, nextBatchEndBlockNumber);
+                    if (uploadBlocksTar(s3Client, nextBatchStartBlockNumber, nextBatchEndBlockNumber)) {
+                        // write the latest archived block number to the archive
+                        s3Client.uploadTextFile(
+                                LATEST_ARCHIVED_BLOCK_FILE,
+                                LATEST_FILE_STORAGE_CLASS,
+                                String.valueOf(nextBatchEndBlockNumber));
+                        // update the last archived block number
+                        lastArchivedBlockNumber.set(nextBatchEndBlockNumber);
+                        // log completed
+                        LOGGER.log(DEBUG, END_UPLOAD_MESSAGE, nextBatchStartBlockNumber, nextBatchEndBlockNumber);
+                    } else {
+                        LOGGER.log(
+                                INFO,
+                                "Could not upload archive block batch: {0} - {1}",
+                                nextBatchStartBlockNumber,
+                                nextBatchEndBlockNumber);
+                    }
                 }
             }
             return null;
@@ -291,9 +298,10 @@ public class S3ArchivePlugin implements BlockNodePlugin, BlockNotificationHandle
          * @param s3Client the S3 client to use for uploading the blocks
          * @param startBlockNumber the first block number to upload
          * @param endBlockNumber the last block number to upload, inclusive
+         * @return true if the upload was successful, false otherwise
          */
-        private void uploadBlocksTar(final S3Client s3Client, final long startBlockNumber, final long endBlockNumber)
-                throws IllegalStateException, S3ResponseException, IOException {
+        private boolean uploadBlocksTar(final S3Client s3Client, final long startBlockNumber, final long endBlockNumber)
+                throws S3ResponseException, IOException {
             // The HTTP client needs an Iterable of byte arrays, so create one from the blocks
             final List<BlockAccessor> accessors = new ArrayList<>();
             for (long i = startBlockNumber; i <= endBlockNumber; i++) {
@@ -302,15 +310,14 @@ public class S3ArchivePlugin implements BlockNodePlugin, BlockNotificationHandle
                     accessors.add(accessor);
                 } else {
                     LOGGER.log(WARNING, GAP_FOUND_MESSAGE, i, startBlockNumber, endBlockNumber);
-                    return;
+                    return false;
                 }
             }
             final Iterator<BlockAccessor> blocksIterator = accessors.iterator();
             final Iterator<byte[]> tarBlocks = new TaredBlockIterator(Format.ZSTD_PROTOBUF, blocksIterator);
             // fetch the first blocks consensus time so that we can place the file in a directory based on year and
             // month
-            final BlockUnparsed firstBlock =
-                    context.historicalBlockProvider().block(startBlockNumber).blockUnparsed();
+            final BlockUnparsed firstBlock = accessors.getFirst().blockUnparsed();
             final ZonedDateTime firstBlockConsensusTime;
             try {
                 final Bytes headerBytes = firstBlock.blockItems().getFirst().blockHeader();
@@ -340,6 +347,7 @@ public class S3ArchivePlugin implements BlockNodePlugin, BlockNotificationHandle
                     .formatted(basePath, yearMonthDate, filePrefix, startBlockNumberFormatted, endBlockNumberFormatted);
             // Upload the blocks to S3
             s3Client.uploadFile(objectKey, archiveConfig.storageClass(), tarBlocks, "application/x-tar");
+            return true;
         }
     }
 }
--- a/block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
+++ b/block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
@@ -141,11 +141,9 @@ public final class S3Client implements AutoCloseable {
 
     /**
      * Closes the HTTP client.
-     *
-     * @throws Exception if an error occurs while closing the client
      */
     @Override
-    public void close() throws Exception {
+    public void close() {
         this.httpClient.close();
     }
 
@@ -595,7 +593,8 @@ public final class S3Client implements AutoCloseable {
                 case PUT -> requestBuilder.PUT(HttpRequest.BodyPublishers.ofByteArray(requestBody));
                 case GET -> requestBuilder.GET();
                 case DELETE -> requestBuilder.DELETE();
-                default -> throw new IllegalArgumentException("Unsupported HTTP method: " + httpMethod);};
+                default -> throw new IllegalArgumentException("Unsupported HTTP method: " + httpMethod);
+            };
             requestBuilder = requestBuilder.headers(localHeaders.entrySet().stream()
                     .flatMap(entry -> Stream.of(entry.getKey(), entry.getValue()))
                     .toArray(String[]::new));
--- a/block-node/s3-archive/src/main/java/module-info.java
+++ b/block-node/s3-archive/src/main/java/module-info.java
@@ -18,7 +18,6 @@ module org.hiero.block.node.archive.s3cloud {
     requires org.hiero.block.node.base;
     requires org.hiero.block.protobuf.pbj;
     requires com.github.spotbugs.annotations;
-    requires java.logging;
     requires java.management;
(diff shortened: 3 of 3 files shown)


block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -177,104 +203,123 @@ public void handlePersisted(final PersistedNotification notification) {
...
+         * @param s3Client the S3 client to use for uploading the blocks
+         * @param startBlockNumber the first block number to upload
+         * @param endBlockNumber the last block number to upload, inclusive
+         */
+        private void uploadBlocksTar(final S3Client s3Client, final long startBlockNumber, final long endBlockNumber)
+                throws IllegalStateException, S3ResponseException, IOException {
+            // The HTTP client needs an Iterable of byte arrays, so create one from the blocks
+            final Iterator<BlockAccessor> blocksIterator = LongStream.range(startBlockNumber, endBlockNumber + 1)
+                    .mapToObj(blockNumber -> context.historicalBlockProvider().block(blockNumber))
+                    .iterator();
```
ata-nas: This is probably the culprit for the NPE but cannot be proven until we run the tests again with the improvements from this PR which will show much more information.

What I reckon is happening is that this for loop will map to null the block accessors for blocks which are missing (which is intended). So in order to ensure no gaps, we need to have all blocks present and cannot really rely on the accessors and/or others providing for us. We do have plans to change drastically the logic for the archiving and we will be uploading data as it comes available, so the approach here might be completely redundant in the near future.

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -177,104 +215,123 @@ public void handlePersisted(final PersistedNotification notification) {
...
+                            LATEST_ARCHIVED_BLOCK_FILE,
+                            LATEST_FILE_STORAGE_CLASS,
+                            String.valueOf(nextBatchEndBlockNumber));
+                    // update the last archived block number
+                    lastArchivedBlockNumber.set(nextBatchEndBlockNumber);
+                    // log completed
+                    LOGGER.log(
+                            INFO,
+                            "Uploaded archive block batch: " + nextBatchStartBlockNumber + "-"
+                                    + nextBatchEndBlockNumber);
```
jsync-swirlds: Also debug and placeholders.

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -177,104 +215,123 @@ public void handlePersisted(final PersistedNotification notification) {
...
+                final BlockNodeContext context) {
+            this.nextBatchStartBlockNumber = nextBatchStartBlockNumber;
+            this.nextBatchEndBlockNumber = nextBatchEndBlockNumber;
+            this.archiveConfig = archiveConfig;
+            this.lastArchivedBlockNumber = lastArchivedBlockNumber;
+            this.context = context;
+        }
+
+        @Override
+        public Void call() throws Exception {
```
jsync-swirlds: You can be more specific than the interface, and absolutely should be.

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
jsync-swirlds: Do you mean the one at the end?
That is within the lambda and most likely line 223 (before your updates).

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -177,104 +215,123 @@ public void handlePersisted(final PersistedNotification notification) {
...
+                throw new IllegalStateException("Failed to parse Block Header from first block", e);
+            }
+
+            // Upload the blocks to S3
+            s3Client.uploadFile(
+                    archiveConfig.basePath() + "/" + YEAR_MONTH_FORMATTER.format(firstBlockConsensusTime)
+                            + "/"
+                            + FILE_PREFIX_FORMATTER.format(firstBlockConsensusTime)
+                            + blockNumberFormated(startBlockNumber)
+                            + "-" + blockNumberFormated(endBlockNumber) + ".tar",
```
jsync-swirlds: Ugh, can we use a format string to make this more readable?

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -166,6 +170,28 @@ public void handlePersisted(final PersistedNotification notification) {
             nextStart = nextEnd + 1;
             nextEnd = nextStart + archiveConfig.blocksPerFile() - 1;
         }
+        handleFinishedUploads();
+    }
+
+    private void handleFinishedUploads() {
+        // Poll the active uploads to see if any are finished
+        Future<Void> upload;
```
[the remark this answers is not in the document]
ata-nas: Agreed.

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -166,6 +182,28 @@ public void handlePersisted(final PersistedNotification notification) {
...
+            // Handle the finished upload
+            try {
+                // Call get() to ensure the upload is complete.
+                upload.get();
+            } catch (final CancellationException e) {
+                // If the upload was canceled, log the that event.
+                LOGGER.log(Level.DEBUG, "S3 batch upload cancelled", e);
+            } catch (final ExecutionException e) {
+                // If the upload was canceled or failed, log the exception.
+                LOGGER.log(Level.INFO, "S3 batch upload failed", e);
```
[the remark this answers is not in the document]
ata-nas: I had them merged in the beginning. A cancelled batch is still a failed batch. Will merge again.

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -177,104 +215,123 @@ public void handlePersisted(final PersistedNotification notification) {
...
+            // Check the nextBatchStartBlockNumber is less than lastArchivedBlockNumber, ie is this a duplicate
+            // batch and
+            // already been archived. That can happen as the archive job has not been completed for a batch and
+            // a new
+            // batch is created.
+            if (nextBatchStartBlockNumber > lastArchivedBlockNumber.get()) {
+                // log started
+                LOGGER.log(
+                        INFO,
+                        "Uploading archive block batch: " + nextBatchStartBlockNumber + "-" + nextBatchEndBlockNumber);
```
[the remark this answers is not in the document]
ata-nas: Yeah, these were all copy-pasted. Will improve.

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
ata-nas: This is so because of the `try-with-resources` which forces me to handle an `Exception` because methods within the construct throw. I agree with your statement, however.

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -101,6 +107,7 @@ public void init(final BlockNodeContext context, final ServiceBuilder serviceBui
                 .createSingleThreadExecutor(
                         "S3ArchiveRunner",
                         (t, e) -> LOGGER.log(ERROR, "Uncaught exception in thread: " + t.getName(), e));
+        activeUploads = new ExecutorCompletionService<>(executorService);
```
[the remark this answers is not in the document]
jsync-swirlds: The future changes will likely be a new plugin so we don't disrupt testing.
Much as we replaced an old publish plugin with a new stream publisher plugin.

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
jsync-swirlds: That's because `S3Client.close`, _which does not throw checked exceptions_ declares `throws Exception`.  This is why it's important to be specific always for implementations of interfaces.
Remove that and it's cured.

```patch
diff --git a/block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java b/block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
index 26bd1196..26ee7689 100644
--- a/block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
+++ b/block-node/base/src/main/java/org/hiero/block/node/base/s3/S3Client.java
@@ -135,24 +135,23 @@ public final class S3Client implements AutoCloseable {
             this.documentBuilderFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
             this.documentBuilderFactory.setNamespaceAware(true);
         } catch (final Exception e) {
             throw new S3ClientInitializationException(e);
         }
     }
 
     /**
      * Closes the HTTP client.
      *
-     * @throws Exception if an error occurs while closing the client
      */
     @Override
-    public void close() throws Exception {
+    public void close() {
         this.httpClient.close();
     }
 
     /**
      * Lists objects in the S3 bucket with the specified prefix
      *
      * @param prefix The prefix to filter the objects.
      * Use {@link StringUtilities#EMPTY} for a wildcard prefix (list all).
      * @param maxResults The maximum number of results to return.
      * @return A list of object keys.
diff --git a/block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java b/block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
index cb490ffe..6de7bdf1 100644
--- a/block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
+++ b/block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
@@ -248,21 +248,21 @@ public class S3ArchivePlugin implements BlockNodePlugin, BlockNotificationHandle
             this.nextBatchEndBlockNumber = nextBatchEndBlockNumber;
             this.archiveConfig = archiveConfig;
             this.lastArchivedBlockNumber = lastArchivedBlockNumber;
             this.context = context;
         }
 
         private static final String START_UPLOAD_MESSAGE = "Uploading archive block batch: {0} - {1}";
         private static final String END_UPLOAD_MESSAGE = "Finished uploading archive block batch: {0} - {1}";
 
         @Override
-        public Void call() throws Exception {
+        public Void call() throws S3ClientInitializationException, S3ResponseException, IOException {
             // Check the nextBatchStartBlockNumber is less than lastArchivedBlockNumber, ie is this a duplicate
             // batch and
             // already been archived. That can happen as the archive job has not been completed for a batch and
             // a new
             // batch is created.
             if (nextBatchStartBlockNumber > lastArchivedBlockNumber.get()) {
                 // log started
                 LOGGER.log(DEBUG, START_UPLOAD_MESSAGE, nextBatchStartBlockNumber, nextBatchEndBlockNumber);
                 try (final S3Client s3Client = createClient(archiveConfig)) {
                     // fetch the blocks from the persistence plugins and archive
@@ -294,21 +294,21 @@ public class S3ArchivePlugin implements BlockNodePlugin, BlockNotificationHandle
         /**
          * Upload a batch of blocks to the tar archive in S3 bucket. This is called when a batch of blocks can
          * be archived.
          *
          * @param s3Client the S3 client to use for uploading the blocks
          * @param startBlockNumber the first block number to upload
          * @param endBlockNumber the last block number to upload, inclusive
          * @return true if the upload was successful, false otherwise
          */
         private boolean uploadBlocksTar(final S3Client s3Client, final long startBlockNumber, final long endBlockNumber)
-                throws IllegalStateException, S3ResponseException, IOException {
+                throws S3ResponseException, IOException {
             // The HTTP client needs an Iterable of byte arrays, so create one from the blocks
             final List<BlockAccessor> accessors = new ArrayList<>();
             for (long i = startBlockNumber; i <= endBlockNumber; i++) {
                 final BlockAccessor accessor = context.historicalBlockProvider().block(i);
                 if (accessor != null) {
                     accessors.add(accessor);
                 } else {
                     LOGGER.log(WARNING, GAP_FOUND_MESSAGE, i, startBlockNumber, endBlockNumber);
                     return false;
                 }
```

block-node/s3-archive/src/main/java/org/hiero/block/node/archive/s3/S3ArchivePlugin.java
```diff
@@ -177,104 +214,140 @@ public void handlePersisted(final PersistedNotification notification) {
...
+            this.lastArchivedBlockNumber = lastArchivedBlockNumber;
+            this.context = context;
+        }
+
+        private static final String START_UPLOAD_MESSAGE = "Uploading archive block batch: {0} - {1}";
+        private static final String END_UPLOAD_MESSAGE = "Finished uploading archive block batch: {0} - {1}";
+
+        @Override
+        public Void call() throws S3ClientInitializationException, S3ResponseException, IOException {
+            // Check the nextBatchStartBlockNumber is less than lastArchivedBlockNumber, ie is this a duplicate
```
Nana-EC: nit: comments read in the opposite from the code.
Could have said somehting like
```suggestion
            // Check the nextBatchStartBlockNumber is greater than lastArchivedBlockNumber before attempting batched upload. If not we assume this is a duplicate and skip attempting upload for this batch as it may already be archived. This can also happen when the archive job has not been completed for a batch and a new batch is created
```

Nana-EC (approved):
Nice refactor for visibility.
Left 2 non-blocking comments for future follow up consideration or backlog tickets

PR #1829, opened 2025-11-04
chore: Followup to Issue 1626

jsync-swirlds:
## Reviewer Notes
* Added explanatory comment
* Removed many cases of `throws Exception` in tests that could (should) be specific so we catch new exceptions or unhandled failures
* Removed `throws Exception` from `S3Client.close()` that was forcing all users to catch or declare `Exception`.
    * This was particularly pernicious because the method did not actually throw _any_ checked exceptions.
* Replaced `new String(stdout, charset)` with `stdout.toString(charset)`.

### DIFF
--- a/tools-and-tests/simulator/src/test/java/org/hiero/block/simulator/mode/PublisherClientModeHandlerTest.java
+++ b/tools-and-tests/simulator/src/test/java/org/hiero/block/simulator/mode/PublisherClientModeHandlerTest.java
@@ -16,6 +16,7 @@ import com.swirlds.config.api.Configuration;
 import java.io.IOException;
 import java.util.Map;
 import org.hiero.block.simulator.config.data.BlockStreamConfig;
+import org.hiero.block.simulator.exception.BlockSimulatorParsingException;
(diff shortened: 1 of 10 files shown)


tools-and-tests/simulator/src/test/java/org/hiero/block/simulator/mode/PublisherClientModeHandlerTest.java
```diff
@@ -53,7 +54,8 @@ void setUp() throws IOException {
     }
 
     @Test
-    void testStartWithMillisPerBlockStreaming_WithBlocks() throws Exception {
+    void testStartWithMillisPerBlockStreaming_WithBlocks()
```
Nana-EC:
```suggestion
    void testStartWithMillisPerBlockStreamingWithBlocks()
```

Nana-EC: Here and all the other test names with an underscore since you've touched them

PR #1833, opened 2025-11-05
chore: `BlockAccessor`s Hardening & Improved Error Handling

### LINKED ISSUE #1273: `BlockAccessor`s Hardening & Improved Error Handling
### Story Form

As a Block Node Developer/User/Operator
I want to have reliable `BlockAccessor`s
So that I can trust the reliability of my system

### Technical Notes

- see [this comment thread](https://github.com/hiero-ledger/hiero-block-node/pull/1262#discussion_r2128819135)
- currently `BlockAccessors` are generally created only when the actual block they target exists
  - this is correct, but that does not handle the case where during the lifetime of the accessor, the actual block may be rendered unaccessible (the reason does not matter, deleted, moved, filesystem issue etc.)
- in those cases we need to be absolutely sure that the system will not fail as this is a very much expected condition
- we must agree on API level what should be done in such cases
  - for now, what makes sense is that accessors never throw any exceptions, except for bugs and unrecoverable cases In such cases, the node itself cannot recover?
  - we should return `null` when we have an accessor, but attempts to read the bytes are not successful

ata-nas:
* BlockAccessors are now Autocloseable
* BlockAccessors now create hard links for the duration of their
  lifespan
  * This allows us to ensure data is accessible for the duration of the
    accessor's life
* Tests are updated and migrated
* Deleted redundant configuration tests
* Added tests:
  * Tests for subsequent reads to ensure that closing an accessor does
    not affect the data or the ability for a new accessor to access it

## Reviewer Notes

- migrating accessors to an autocloseable variant
- improving the ability to retrieve data by having hard links

## Related Issue(s)

Resolves #1273


block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
```diff
@@ -30,16 +30,7 @@
...
-        @Loggable @ConfigProperty(defaultValue = "0") long blockRetentionThreshold) {
-    /**
-     * Constructor.
-     */
-    public FilesHistoricConfig {
-        Objects.requireNonNull(rootPath);
-        Objects.requireNonNull(compression);
-        Preconditions.requireInRange(powersOfTenPerZipFileContents, 1, 6);
-        Preconditions.requireGreaterOrEqual(blockRetentionThreshold, 0L);
-    }
```
ata-nas: These constructors here and for the other config have been removed as validation must be done by the config dependency. I've also deleted the unit tests for the configs. I think they are redundant and maintaining them is a hassle.

ata-nas:
We could also add tests for:
1. Clearing the links root path on startup for recents
2. Clearing the links root path on startup for historic

mustafauzunn (commented):
We should update `docs/block-node/configuration.md` with the new configs

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -77,14 +74,21 @@ public BlockResponse getBlock(BlockRequest request) {
...
+                    responseCounterSuccess.increment();
+                    return response;
+                } else {
+                    return new BlockResponse(Code.NOT_FOUND, null);
+                }
+            }
+        } catch (final RuntimeException e) {
+            final String message = "Failed to retrieve block number %d.".formatted(request.blockNumber());
             LOGGER.log(ERROR, message, e);
             responseCounterNotFound.increment(); // Should this be a failure counter?
```
rockysingh: I agree with comment maybe we should have a failure counter. However, this could be done as part of another issue.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
```diff
@@ -30,16 +30,7 @@
 @ConfigData("files.historic")
 public record FilesHistoricConfig(
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/data/historic") Path rootPath,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/data/historic-links") Path linksRootPath,
```
jsync-swirlds: instead of totally new paths, why not use a subfolder of the existing path?
e.g. `rootPath + "/accessor-links"` would separate the folders but ensure it's on the same physical device (which is a requirement for hard links).

block-node/block-access/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
ata-nas: This comment may be redundant and we simply need to delete it.

@AlfredoG87 what do we mean here by failure counter? Is that something that is simply grafana UI?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessor.java
```diff
@@ -55,12 +76,11 @@ public long blockNumber() {
...
         Objects.requireNonNull(format);
-        CompressionType compressionType = blockPath.compressionType();
-        try (final FileSystem zipFs = FileSystems.newFileSystem(blockPath.zipFilePath())) {
-            final Path entry = zipFs.getPath(blockPath.blockFileName());
+        try (final FileSystem zipFs = FileSystems.newFileSystem(zipFileLink)) {
+            final Path entry = zipFs.getPath(zipEntryName);
             return getBytesFromPath(format, entry, compressionType);
         } catch (final UncheckedIOException | IOException e) {
-            LOGGER.log(WARNING, FAILED_TO_READ_MESSAGE.formatted(blockPath), e);
+            LOGGER.log(WARNING, FAILED_TO_READ_MESSAGE.formatted(zipFileLink), e);
```
jsync-swirlds: I think the link should be a hidden internal; pointing to the block path in logs seems much more useful.

block-node/base/src/main/java/org/hiero/block/node/base/tar/TaredBlockIterator.java
jsync-swirlds: IS this class used?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -91,9 +93,22 @@ public void init(final BlockNodeContext context, final ServiceBuilder serviceBui
         blockRetentionThreshold = config.blockRetentionThreshold();
         // Initialize metrics
         initMetrics(context.metrics());
-        // create plugin data root directory if it does not exist
+        final Path linksRoot = config.linksRootPath();
+        // attempt to clear any existing links root directory
+        if (Files.exists(linksRoot)) {
+            try (final Stream<Path> paths = Files.walk(linksRoot)) {
+                for (final Path path : paths.sorted(Comparator.reverseOrder()).toList()) {
```
jsync-swirlds: Why sort the hardlinks?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
```diff
@@ -86,6 +87,9 @@ long writeNewZipFile(List<BlockAccessor> batch) throws IOException {
                 // it is here possible that the accessor will no longer be able to access the block
                 // because it is possible that the block has been deleted or has been moved
                 final Bytes bytes = blockAccessor.blockBytes(format);
+                // close the accessor as we are done with it and we need to free
+                // resources
+                blockAccessor.close();
```
jsync-swirlds: Might be worth defining the batch as an `AutoCloseable` and `Iterable` object (instead of just a list) so that close can be handled automatically in try-with-resources by the _caller_ of this method instead of handling the individual close calls here.

In particular I'm concerned what happens to all the remaining not-yet-complete accessors if this method encounters an exception.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -228,8 +243,14 @@ public BlockAccessor block(final long blockNumber) {
                     config.liveRootPath(), blockNumber, config.compression(), config.maxFilesPerDir());
             if (Files.exists(verifiedBlockPath)) {
                 // we have the block so return it
-                blocksReadCounter.increment();
-                return new BlockFileBlockAccessor(verifiedBlockPath, config.compression(), blockNumber);
+                try {
+                    final BlockFileBlockAccessor accessor =
+                            new BlockFileBlockAccessor(verifiedBlockPath, config, blockNumber);
+                    blocksReadCounter.increment();
```
jsync-swirlds: Thought for a future PR:
What would happen if we added the metric to the accessor constructor here, and let the accessor increment the counter when the block is read? If an accessor might be read repeatedly, we could get more accurate metrics that way.

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/FilesRecentConfig.java
```diff
@@ -23,16 +24,7 @@
 @ConfigData("files.recent")
 public record FilesRecentConfig(
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/data/live") Path liveRootPath,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/data/live-links") Path linksRootPath,
```
jsync-swirlds: Similar to above, this should probably be a hard-coded subfolder of the liveRootPath rather than a separate configured item.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessor.java
ata-nas: Yes, you are right about that. Good catch.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
```diff
@@ -30,16 +30,7 @@
...
         @Loggable @ConfigProperty(defaultValue = "4") @Min(1) @Max(6) int powersOfTenPerZipFileContents,
-        @Loggable @ConfigProperty(defaultValue = "0") long blockRetentionThreshold) {
-    /**
-     * Constructor.
-     */
-    public FilesHistoricConfig {
-        Objects.requireNonNull(rootPath);
-        Objects.requireNonNull(compression);
-        Preconditions.requireInRange(powersOfTenPerZipFileContents, 1, 6);
-        Preconditions.requireGreaterOrEqual(blockRetentionThreshold, 0L);
```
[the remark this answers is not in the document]
ata-nas: ✅ Done.

block-node/base/src/main/java/org/hiero/block/node/base/tar/TaredBlockIterator.java
jsync-swirlds: Thanks.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -91,9 +96,16 @@ public void init(final BlockNodeContext context, final ServiceBuilder serviceBui
...
+        final Path linksRootPath = dataRootPath.resolve("links");
+        // create plugin data and links root directories if they do not exist
         try {
-            Files.createDirectories(config.rootPath());
+            // attempt to clear any existing links root directory
+            if (Files.isDirectory(linksRootPath, LinkOption.NOFOLLOW_LINKS)) {
+                Files.walkFileTree(linksRootPath, new RecursiveFileDeleteVisitor());
+            }
+            Files.createDirectories(linksRootPath);
+            Files.createDirectories(dataRootPath);
```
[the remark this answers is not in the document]
ata-nas: True, I would leave both, however, just in case something changes and the links root path is no longer a subdir. Unlikely to happen, but these invocations are made a single time at startup and do not cost us anything really.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -91,9 +93,23 @@ public void init(final BlockNodeContext context, final ServiceBuilder serviceBui
...
+        final Path linksRootPath = dataRootPath.resolve("links");
+        // attempt to clear any existing links root directory
+        if (Files.exists(linksRootPath)) {
+            try (final Stream<Path> paths = Files.walk(linksRootPath)) {
+                for (final Path path : paths.sorted(Comparator.reverseOrder()).toList()) {
+                    Files.delete(path);
+                }
+            } catch (final IOException e) {
+                LOGGER.log(ERROR, "Could not clear links root directory", e);
+                context.serverHealth().shutdown(name(), "Could not clear links root directory");
```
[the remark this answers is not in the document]
ata-nas: This was refactored and does not exist anymore.

jsync-swirlds (dismissed):
Just a note, Part of the changes here using links for files.historic and the already-merged-to-main change for files.historic to stage based on Verification notification are completely incompatible.
I am working on reimplementing the useful changes here in light of that change; but it will take a day or three to get it all reconciled and tested.

block-node/base/src/main/java/org/hiero/block/node/base/tar/TaredBlockIterator.java
```diff
@@ -98,9 +101,13 @@ public byte[] next() {
...
-                        final BlockAccessor currentBlock = blockIterator.next();
-                        currentBlockBytes = currentBlock.blockBytes(format).toByteArray();
-                        currentBlockName = BLOCK_NUMBER_FORMAT.format(currentBlock.blockNumber()) + extension;
+                        final BlockAccessor currentBlockAccessor = blockIterator.next();
+                        currentBlockName = BLOCK_NUMBER_FORMAT.format(currentBlockAccessor.blockNumber()) + extension;
+                        currentBlockBytes =
+                                currentBlockAccessor.blockBytes(format).toByteArray();
+                        // close the accessor as we are done with it and we need
+                        // to free resources
+                        currentBlockAccessor.close();
```
ata-nas: This `BlockAccessor#close` call could be ommitted. We should be using the autocloseable block accessor batch in the s3 plugin with a try-with-resources and this iterator should be within that scope.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -427,74 +443,85 @@ private void startMovingBatchOfBlocksToZipFile(final LongRange batchRange) {
...
+     */
+    private BlockAccessorBatch gatherAccessors(final long startBlockNumber, final long endBlockNumber) {
+        final BlockAccessorBatch accessors = new BlockAccessorBatch();
+        try {
+            for (long i = startBlockNumber; i <= endBlockNumber; i++) {
+                Path path = BlockFile.nestedDirectoriesBlockFilePath(
+                        stagingPath, i, config.compression(), config.maxFilesPerDir());
+                final BlockAccessor accessor = new BlockStagingFileAccessor(path, config.compression(), i);
+                if (accessor != null) {
+                    accessors.add(accessor);
```
ata-nas: I did consider to make the accessors batch mutable, thought that a read-only variant should be better. Not opposed to this.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessor.java
```diff
@@ -25,42 +25,51 @@ final class ZipBlockAccessor implements BlockAccessor {
...
-        this.blockPath = Objects.requireNonNull(blockPath);
+    ZipBlockAccessor(@NonNull final BlockPath blockPath) throws IOException {
+        blockPathData = blockPath;
+        blockNumber = blockPath.blockNumber();
+        final Path zipFilePath = blockPath.zipFilePath();
+        this.absoluteZipFilePath = zipFilePath.toAbsolutePath().toString();
+        if (!Files.isRegularFile(zipFilePath)) {
+            final String msg = INVALID_ZIP_FILE_PATH_MESSAGE.formatted(zipFilePath);
+            throw new IOException(msg);
+        }
```
ata-nas: Why the removal of the hardlink to the zip file? Is the reason that we do not expect the zip files to be removed?

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
```diff
@@ -89,6 +89,9 @@ long writeNewZipFile(List<BlockAccessor> batch) throws IOException {
...
```
ata-nas: We can omit this accessor close here. We are using accessor batch which is autocloseable. This call should be within the scope of the try-with-resources.

block-node/block-providers/files.historic/src/test/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPluginTest.java
```diff
@@ -64,19 +64,95 @@ class BlockFileHistoricPluginTest {
...
+            // Assert that the roots do not exist
+            assertThat(dataRoot).doesNotExist();
+            // Now start the plugin
+            start(toTest, testHistoricalBlockFacility, getConfigOverrides());
+            // Assert that the roots are created
+            assertThat(dataRoot).exists().isDirectory().isNotEmptyDirectory();
+            // assertThat(Files.list(dataRoot).toList()).hasSize(1).containsExactly(linksRoot);
+        }
+
+        /*
```
ata-nas: Broken Javadoc

block-node/base/src/main/java/org/hiero/block/node/base/tar/TaredBlockIterator.java
jsync-swirlds: Will fix.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessor.java
jsync-swirlds: I couldn't find that, but might have just missed it.

Also, we create a zipfs for the accessor, which will hold the file open until it's closed.

And I just noticed I forgot to move that to the constructor/close.

Will fix that part.

jsync-swirlds: The zipfs change (which should also cut down on duplicate file reads) will still handle preventing unexpected removal, but will look at cleanup to ensure it doesn't fail unexpectedly.

block-node/base/src/main/java/org/hiero/block/node/base/tar/TaredBlockIterator.java
jsync-swirlds: This is actually problematic because the TarredBlockIterator just takes an iterator; we have no guarantee the accessor is held by anything that will close it.
Will make some related changes to make this sensible.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
jsync-swirlds: Creating and adding a list was error prone.  Using a more build-as-we-go approach made a lot more sense, and kept close calls in one place.

block-node/block-providers/files.historic/src/test/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPluginTest.java
jsync-swirlds: That was on purpose, otherwise spotless complained about javadoc without a method below it.

Removed.

block-node/block-providers/files.historic/src/test/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessorTest.java
```diff
@@ -43,19 +43,23 @@ class ZipBlockAccessorTest {
     private FileSystem jimfs;
     /** The configuration for the test. */
     private FilesHistoricConfig defaultConfig;
-    /** The temporary directory used for the test. */
-    private Path tempDir;
+    /** The temporary data directory used for the test. */
+    private Path dataTempDir;
+    /** The temporary links directory used for the test. */
+    private Path linksTempDir;
```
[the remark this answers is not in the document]
jsync-swirlds: Removed.

block-node/base/src/main/java/org/hiero/block/node/base/tar/TaredBlockIterator.java
jsync-swirlds: Changed to check `isClosed`.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -427,74 +443,85 @@ private void startMovingBatchOfBlocksToZipFile(final LongRange batchRange) {
...
-                // we have a gap in the batch, so we cannot zip it
-                final String message = "Batch of blocks [%d -> %d] has a gap, skipping zipping"
-                        .formatted(batchFirstBlockNumber, batchLastBlockNumber);
-                LOGGER.log(INFO, message);
-            } else {
+        // first off, let's create our batch of blocks
+        final long batchFirstBlockNumber = batchRange.start();
+        final long batchLastBlockNumber = batchRange.end();
+        try (final BlockAccessorBatch blockAccessors = gatherAccessors(batchFirstBlockNumber, batchLastBlockNumber)) {
+            if (true) {
```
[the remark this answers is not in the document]
jsync-swirlds: Replaced with correct if test.

jsync-swirlds:
> Timeout for e2e tests. and also
> 
> ```
> java.lang.ClassCastException: class sun.nio.fs.UnixPath cannot be cast to class com.google.common.jimfs.JimfsPath (sun.nio.fs.UnixPath is in module java.base of loader 'bootstrap'; com.google.common.jimfs.JimfsPath is in module com.google.common.jimfs@1.3.1 of loader 'app')
>         at org.hiero.block.node.blocks.files.historic@0.24.0-SNAPSHOT/org.hiero.block.node.blocks.files.historic.ZipBlockArchiveTest$FunctionalTests.testMinStoredSingleZipFile(ZipBlockArchiveTest.java:168)
> ```
> 
> for `ZipBlockArchiveTest` tests

Yes, I am working on this error. It's quite strange as it only happens from command line; in IDE it does not happen (so it cannot be debugged).
It appears, so far, to be related to containers and Erasure.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessor.java
jsync-swirlds: This did not work out well.  JimFS fought against holding files open, even briefly.

For now I've added links back in, but with the earlier suggested tempfile structure rather than UUID.

block-node/base/src/main/java/org/hiero/block/node/base/tar/TaredBlockIterator.java
ata-nas: The tarred iterator should probably accept an iterable rather than an iterator. I did not make this change, however.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessor.java
```diff
@@ -112,12 +148,29 @@ private Bytes getJsonBytesFromProtobufBytes(final Bytes sourceData) {
...
+        } catch (final IOException e) {
+            final String message = FAILED_TO_DELETE_LINK_MESSAGE.formatted(
+                    blockNumber, absoluteZipFilePath, blockPathData.blockFileName());
+            LOGGER.log(INFO, message, e);
+        }
+    }
+
+    @Override
+    public boolean isClosed() {
+        return !Files.exists(zipFileLink);
```
ata-nas: Could this be a simple boolean that will be changed when close is called a single time? Or even a CAS on an atomic boolean? Or is it not as resilient as checking for the existing link?

block-node/block-providers/files.recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
```diff
@@ -154,12 +170,31 @@ private Bytes getJsonBytesFromProtobufBytes(final Bytes sourceData) {
...
+        } catch (final IOException e) {
+            final String message = "Failed to delete accessor link for block: %d, path: %s"
+                    .formatted(blockNumber, absolutePathToBlock);
+            LOGGER.log(INFO, message, e);
+        }
+    }
+
+    @Override
+    public boolean isClosed() {
+        return Files.exists(blockFileLink);
```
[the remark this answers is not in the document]
jsync-swirlds: 1. Checking the filesystem is important as managing an internal flag is both harder to do correctly and less resilient to interactions with other plugins or processes.
2. Agreed that this should be inverted; thanks for catching.

block-node/block-providers/files.recent/src/test/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPluginTest.java
```diff
@@ -55,19 +58,79 @@ class BlockFileRecentPluginTest {
...
         this.blockFileRecentPlugin = new BlockFileRecentPlugin(this.filesRecentConfig);
         this.historicalBlockFacility = new SimpleInMemoryHistoricalBlockFacility();
     }
 
+    /**
+     * Nested class for plugin startup tests.
+     */
+    @Nested
+    @DisplayName("Startup Tests")
+    final class StartupTest extends PluginTestBase<BlockFileRecentPlugin, BlockingExecutor> {
```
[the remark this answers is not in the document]
jsync-swirlds: Would like to add those as a separate PR; this one is already getting bloated and taking too long.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessor.java
```diff
@@ -22,45 +24,79 @@
  * The ZipBlockAccessor class provides access to a block stored in a zip file.
  */
 final class ZipBlockAccessor implements BlockAccessor {
+    private static final String FAILED_TO_DELETE_LINK_MESSAGE =
+            "Failed to delete accessor link for block: %s, zipFilePath: %s, entryName: %s";
```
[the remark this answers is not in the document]
jsync-swirlds: Would like to add that in a separate PR; this one is already getting bloated and taking too long.

block-node/block-providers/files.historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -102,61 +105,78 @@ public List<Class<? extends Record>> configDataTypes() {
...
-        initMetrics(context.metrics());
-        // create plugin data root directory if it does not exist
         try {
+            this.context = Objects.requireNonNull(context);
+            config = context.configuration().getConfigData(FilesHistoricConfig.class);
+            blockRetentionThreshold = config.blockRetentionThreshold();
+            stagingPath = config.rootPath().resolve("staging");
+            // Initialize metrics
+            initMetrics(context.metrics());
+            // create plugin data root directory if it does not exist
```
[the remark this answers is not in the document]
jsync-swirlds: Tests I would like to add separately (otherwise this change will be delayed another full release).
Creating the links root will be added shortly.

block-node/base/src/main/java/org/hiero/block/node/base/tar/TaredBlockIterator.java
```diff
@@ -98,9 +101,17 @@ public byte[] next() {
...
-                        currentBlockName = BLOCK_NUMBER_FORMAT.format(currentBlock.blockNumber()) + extension;
+                        final BlockAccessor currentBlockAccessor = blockIterator.next();
+                        currentBlockName = BLOCK_NUMBER_FORMAT.format(currentBlockAccessor.blockNumber()) + extension;
+                        currentBlockBytes =
+                                currentBlockAccessor.blockBytes(format).toByteArray();
+                        // close the accessor as we are done with it and we need
+                        // to free resources, but only if it isn't already closed.
+                        // @todo replace Iterator with BlockAccessorBatch for this
+                        //     class so we don't need to call close here.
+                        if (!currentBlockAccessor.isClosed()) {
```
Nana-EC: Should we move the `isClosed()` considerations into the implementations and allow callers to just call close()?
The would also allow using of try with patterns
```
try { final BlockAccessor currentBlockAccessor = blockIterator.next()}
....
}
```

block-node/base/src/main/java/org/hiero/block/node/base/BlockFile.java
```diff
@@ -213,21 +227,63 @@ public static Set<Long> nestedDirectoriesAllBlockNumbers(Path basePath, Compress
...
-     * @return the block number
+     * @return the block number, or `-1` if the file does not match the expected format
+     *     or the file base name does not parse as a `long` value.
      */
     public static long blockNumberFromFile(final Path file) {
-        return blockNumberFromFile(file.getFileName().toString());
+        try {
+            if (file != null && file.getFileName() != null) {
+                final String fileName = file.getFileName().toString();
+                if (fileName != null && fileName.indexOf('.') > 0) {
```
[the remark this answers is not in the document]
jsync-swirlds: fileName can still be null; it's not checked yet because `file.getFileName` can return non-null but not be able to return non-null `toString` due to external changes happening at the same time or printable character issues.

block-node/base/src/main/java/org/hiero/block/node/base/tar/TaredBlockIterator.java
Nana-EC: Sounds good, wasn't sure if this was just missing the simplification.
A close and delete is a bit sneaky though.
Could that result in unintended consequences from a dev.
I guess  this is where Java docs come in and making it clear that the given Accessor deletes on close also and then tests.

Nana-EC (approved):
Unblocking per comments.
There's a lot of follow ups to improve confidence but the threshold for now seems good.

AlfredoG87 (approved):
Lot of changes and improvements, LGTM 👍

PR #1834, opened 2025-11-05
Network topology tool

AlfredoG87:
## Reviewer Notes

Quick attempt at defining a hiero network topology schema that can be used to effectively communicate, and use on repeatable tests.


AlfredoG87:
Ideas and scripts were capture on https://github.com/hiero-ledger/hiero-block-node/pull/2030
So this is no longer relevant.

PR #1863, opened 2025-11-13
test:  Add e2e test for socket closing

### LINKED ISSUE #1862: Add e2e test for socket closing
We want to introduce an e2e test verifying that after the latest changes we successfully close sockets with publishers in the required cases

mustafauzunn:
## Reviewer Notes

Add e2e test verifying that after the latest changes we successfully close sockets with publishers in the required cases.

## Related Issue(s)
Fixes #1862

### DIFF
--- a/tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeAPITests.java
+++ b/tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeAPITests.java
@@ -546,6 +546,8 @@ public class BlockNodeAPITests {
 
         // This asserts that UncheckedIOException is thrown and its cause is a SocketException with "socket closed"
         UncheckedIOException ex = assertThrows(UncheckedIOException.class, () -> {
+            // Expected exception to be thrown on the onNext() due to closed socket
+            // from previous duplicate block publish
             requestStream.onNext(request1);
             publishCompleteCountDownLatch.await(); // wait
             endBlock(blockNumber1, requestStream);


ata-nas (approved):

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeAPITests.java
```diff
@@ -470,4 +472,85 @@ void swappedPublisherConnection() throws InterruptedException {
...
+
+        final long blockNumber1 = 1;
+        BlockItem[] blockItems1 = BlockItemBuilderUtils.createSimpleBlockWithNumber(blockNumber1);
+        PublishStreamRequest request1 = PublishStreamRequest.newBuilder()
+                .blockItems(BlockItemSet.newBuilder().blockItems(blockItems1).build())
+                .build();
+
+        // This asserts that UncheckedIOException is thrown and its cause is a SocketException with "socket closed"
+        UncheckedIOException ex = assertThrows(UncheckedIOException.class, () -> {
+            requestStream.onNext(request1);
```
Nana-EC: It's not clear where you expect the exception to be thrown? If it happened on any line the check would still pass right?

Nana-EC (approved):
Thanks for the follow up.
Added a note to make the test case more precise.
Non-blocking but it'd be good to fix it

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeAPITests.java
mustafauzunn: It will be thrown on the first `onNext()`. Added the note for clarity

PR #1865, opened 2025-11-14
feat(verification): Hashing refactor for 0.68.x HAPI version

### LINKED ISSUE #1385: feat(verification): Hashing Refactor to support Internal Nodes differentiation design.
### Story Form

As a BN Developer
I want to keep Hashing consistency with CN
So that we implemeent the latest changes to the way we calculate the BlockRoot Hash

### Technical Notes

There are on-going conversations in how to make the Block Root Hash resistant to a certain type of attack that allow a malicious entity to prove things that does not exist by replacing a leaf with another sub-tree, for this, once the changes are approved and implemented by the CN, we need to update BN hashing algorithm to use the same.

AlfredoG87:
## Reviewer Notes

- Updated Protobuf from CN to latest version, a commit from recent `main`
- Removed VerificationSession for Preview BlockStreams (version 0.64 to 0.67)
- Added new DummyVerificationSession that always returns true for older and newer versions
  - This session will always return true and does not parse the items, hence is backwards compatible with `0.64` to `0.67` HAPI versions
  - Also, it is selected for higher or equal to `0.69.0` since we expect more changes on HAPI `0.69` and while we implement those changes, we want to stay compatible.
  - This version should be removed from Production once no more changes to the protobuf API are present.
 - Finished implementation for VerificationSession: `ExtendedMerkleTreeSession`. 
   - Originally tried to implement the verification as is described on the Notion Document, however after talking with Matt Hess from CN team, we discover that there are some "bugs" or missing parts on the CN side, so I settle with only the changes needed to be able to "match" their implementation albeit flawed.
   - They will address this issues and fix them on HAPI version `0.69.0` along with more changes to come, but this represents good progress and next PR I expect to be a little lighter, since most of the BlockStream changes like `BlockFooter` are included in this version.
   - Added a sample taken from a recording using CN `0.68.0`
 -  Modified all UTs as needed
 - Modified all E2E Suites tests as needed
 - Modified API E2E as needed

## Related Issue(s)

Fixes #1385

### DIFF
--- a/block-node/base/build.gradle.kts
+++ b/block-node/base/build.gradle.kts
@@ -21,6 +21,5 @@ testModuleInfo {
     requires("org.mockito")
     requires("org.testcontainers")
     requires("io.minio")
-    requires("junit")
     requires("org.hiero.block.protobuf.pbj")
 }


protobuf-sources/build.gradle.kts
```diff
@@ -20,7 +20,7 @@ val generateBlockNodeProtoArtifact: TaskProvider<Exec> =
         // proto/internal/unparsed.proto matches structure for
         // block-node-protobuf/block/stream/block_item.proto after update otherwise Verification
         // might fail
-        val cnTagHash = "d7cb85ac7384f809a15489635122504661131af4" // v0.65.0
+        val cnTagHash = "2d7962205082b388db0187da602f2dde9d42f3ec" // main
```
mustafauzunn: Maybe it is better to use the hash of the v0.68.0 release here

mustafauzunn (approved):

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/DummyVerificationSession.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hiero.block.node.spi.blockmessaging.VerificationNotification;
+import org.hiero.block.node.verification.session.VerificationSession;
+
+// TODO, before going live we should remove the Dummy Implementation.
+/**
+ * Dummy implementation of VerificationSession for testing purposes.
+ * It simply collects block items and returns a successful verification notification
+ * if the last block item has a block proof.
+ */
+public class DummyVerificationSession implements VerificationSession {
```
Nana-EC: WHy did we create a new Session vs just simplify the Preview one like you have here?

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -30,24 +49,100 @@ public class ExtendedMerkleTreeSession implements VerificationSession {
...
+
+    /**
+     * Verifies the signature of a hash, for the dummy implementation this always returns true.
+     *
+     * @param hash the hash to verify
+     * @param signature the signature to verify
+     * @return true if the signature is valid, false otherwise
+     */
+    protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature) {
+        // TODO we are close to having real TTS signature verification, we maybe should have a config if we are work on
```
Nana-EC:
```suggestion
        // TODO we are close to having real TSS signature verification, we maybe should have a config if we are work on
```

protobuf-sources/build.gradle.kts
AlfredoG87: Yes, that would be good. also I need to update this so the commit matches the release tag instead of some "main" commit.

block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationServicePluginTest.java
```diff
@@ -172,10 +172,6 @@ void testHandleBackfilledNotification() throws IOException, ParseException {
                 blockNotification.blockNumber(),
                 "The block number should be the same as the one in the block header");
         assertTrue(blockNotification.success(), "The verification should be successful");
-        assertEquals(
```
Nana-EC: Need to capture this in a ticket so make sure to restore the coverage

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -30,24 +49,100 @@ public class ExtendedMerkleTreeSession implements VerificationSession {
...
+                    blockProofs.add(blockProof);
+                }
+            }
+        }
+
+        // since we are only expecting 1 block proof per block, we can finalize verification here
+        // however in the future, we might want to revisit this if we expect multiple proofs
+        // and use the EndOfBlock signal to finalize verification.
+        if (blockFooter != null && blockProofs.size() > 0) {
+            return finalizeVerification(blockProofs.get(0));
```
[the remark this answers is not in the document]
AlfredoG87: Totally agreed, we cannot rely on order alone, we might want to discuss a bit further this in a follow-up, for now we are only expecting `1` proof so this is good for now. But I see your point, we could look for the TSS based proof instead of just getting the first (an only) one.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
jsync-swirlds: Probably should move this class to the Verification Plugin.
It _really_ shouldn't be used anywhere else.

common/src/main/java/org/hiero/block/common/hasher/StreamingHasher.java
```diff
@@ -0,0 +1,120 @@
...
+    private final LinkedList<byte[]> hashList = new LinkedList<>();
+    /** The count of leaves in the tree. */
+    private long leafCount = 0;
+
+    /** Create a new StreamingHasher with an empty state. */
+    public StreamingHasher() {
+        try {
+            digest = MessageDigest.getInstance("SHA-384");
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
```
jsync-swirlds: I don't often agree with Codacy, but here I do.
We need to break out of this _terrible_ habit of using `RuntimeException` as a wildcard catch-all.
I know it's a little bit of effort to create our own exception types, but we should still do that.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/SimpleTestBlockItemBuilder.java
```diff
@@ -68,15 +69,16 @@ public static Bytes createRoundHeaderUnparsed(final long roundNumber) {
...
-                Bytes.wrap("block_signature".getBytes()),
-                Collections.emptyList(),
-                new OneOf<>(
-                        BlockProof.VerificationReferenceOneOfType.VERIFICATION_KEY,
-                        Bytes.wrap("verificationKey".getBytes())));
+
+        TssSignedBlockProof.Builder tssBuildder =
+                TssSignedBlockProof.newBuilder().blockSignature(Bytes.wrap("block_signature".getBytes()));
+
+        BlockProof blockProof = BlockProof.newBuilder()
```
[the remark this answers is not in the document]
AlfredoG87: `block_proof` should not have those fields, copy pasting how a end of a block looks for reference:

```
...
  }, {
    "blockFooter": {
      "previousBlockRootHash": "+OwkmGKsVW3vHcJBJC4Cf8IXPMxUT3yUnEbI3q1e3pIhH2mhemJszr9FKeDRqlTr",
      "rootHashOfAllBlockHashesTree": "jZDBka6rt6XTmfJtfxvvHjTuXiO2rXaZoglhIGhI4c5yFz8ZprU15bBhs063eesY",
      "startOfBlockStateRootHash": "nJ4PNCgeg+DCAHqvTCJGKSs1VBd/TiEojv0+Te0zpqPX0kjDMkecaGjfzY3fzaQS"
    }
  }, {
    "blockProof": {
      "block": "14",
      "signedBlockProof": {
        "blockSignature": "M+PjKNwnZCve0TD2DAcVr8uWQUk506bkzs/N53nFqfhsEQZHNlbLgcK+zfVpNvEy"
      }
    }
  }]
```

common/src/main/java/org/hiero/block/common/hasher/StreamingHasher.java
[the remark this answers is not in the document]
AlfredoG87: It is complete copy-paste, but when we do the final implementation for the final spec changes of `0.69.0` there a few things that we might want to improve.

protobuf-sources/src/main/proto/internal/unparsed.proto
```diff
@@ -66,5 +66,7 @@ message BlockItemUnparsed {
         bytes block_proof = 9;
         bytes record_file = 10;
         bytes trace_data = 11;
+        bytes block_footer = 12;
+        bytes redacted_item = 19;
```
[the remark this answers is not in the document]
AlfredoG87: Added it to the list of items to review on next follow-up
https://github.com/hiero-ledger/hiero-block-node/issues/1906

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -200,36 +214,47 @@ public static ByteBuffer getBlockItemHash(@NonNull BlockItemUnparsed blockItemUn
...
 
-        Bytes inputHash = inputTreeHasher.rootHash().join();
-        Bytes outputHash = outputTreeHasher.rootHash().join();
-        Bytes consensusHeaderHash = consensusHeaderHasher.rootHash().join();
-        Bytes stateChangesHash = stateChangesHasher.rootHash().join();
-        Bytes traceDataHash = traceDataHasher.rootHash().join();
-
-        Bytes lastBlockHash = blockProof.previousBlockRootHash();
-        Bytes blockStartStateHash = blockProof.startOfBlockStateRootHash();
+        final Bytes previousBlockHash = blockFooter.previousBlockRootHash();
```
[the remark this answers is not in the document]
AlfredoG87: ✅ Done!

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅ Done!

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/SimpleTestBlockItemBuilder.java
```diff
@@ -191,7 +193,8 @@ public static BlockItem[] createNumberOfVerySimpleBlocks(final long startBlockNu
             final int i = (blockNumber - (int) startBlockNumber) * 3;
             blockItems[i] = sampleBlockHeader(blockNumber);
             blockItems[i + 1] = sampleRoundHeader(blockNumber * 10L);
-            blockItems[i + 2] = sampleBlockProof(blockNumber);
+            // blockItems[i + 2] = sampleBlockFooter(blockNumber * 10L); //TODO add footer
```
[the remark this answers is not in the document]
AlfredoG87: Thanks! 

Copy pasted some helper methods from you other PR :)

✅  Done!

block-node/verification/src/test/java/org/hiero/block/node/verification/session/TestVerificationHapiVersions.java
```diff
@@ -65,9 +68,9 @@ void verifyHapiVersionCompatibility(final String sampleName, final BlockUtils.Sa
     /** Supply the concrete samples you want to cover. Add more here as they’re added to fixtures. */
     private static Stream<Arguments> sampleBlocks() throws IOException, ParseException {
         // Use readable case names to make failures obvious in the parameterized display
-        final var s1 = BlockUtils.getSampleBlockInfo(BlockUtils.SAMPLE_BLOCKS.HAPI_0_64_0_BLOCK_14);
-        final var s2 = BlockUtils.getSampleBlockInfo(BlockUtils.SAMPLE_BLOCKS.HAPI_0_66_0_BLOCK_10);
+        final BlockUtils.SampleBlockInfo s2 =
```
Nana-EC: Q: Did we want `HAPI_0_68_0_BLOCK_14` in here?

protobuf-sources/build.gradle.kts
```diff
@@ -20,7 +20,7 @@ val generateBlockNodeProtoArtifact: TaskProvider<Exec> =
         // proto/internal/unparsed.proto matches structure for
         // block-node-protobuf/block/stream/block_item.proto after update otherwise Verification
         // might fail
-        val cnTagHash = "d7cb85ac7384f809a15489635122504661131af4" // v0.65.0
+        val cnTagHash = "273aa291b7224ab4c22ff459ba666ec6fbd9930b"
```
Nana-EC:
```suggestion
        val cnTagHash = "273aa291b7224ab4c22ff459ba666ec6fbd9930b" // v0.68.2
```
BTW this means all the references should be corrected.
You should either update all the other references and the test file to 0.68.2 or go the easier route and do set this to the following so everything lines up.

```suggestion
        val cnTagHash = "96ee3efc7814668d3f0c90d7aa1e3dc174a97a35" // v0.68.0
```

build.gradle.kts
```diff
@@ -34,3 +34,13 @@ tasks.versionAsSpecified {
         }
     }
 }
+
+subprojects {
```
Nana-EC: Q: Was this supposed to be here?

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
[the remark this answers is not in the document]
AlfredoG87: Added this question to the follow-up issue. https://github.com/hiero-ledger/hiero-block-node/issues/1906

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
AlfredoG87: https://github.com/hiero-ledger/hiero-block-node/issues/1910

Nana-EC (dismissed):
LG for 0.68 support. We'll make some expansions in follow up PRs for 0.69/0.70 and for test coverage.
Thanks

Nana-EC (approved):

PR #1877, opened 2025-11-17
docs: 1701 document single node k8s deployment

### LINKED ISSUE #527: Design appropriate deployment support for chosen single-node Kubernetes

### LINKED ISSUE #1614: Verify single node helm deployment flow, update with docs
As a block node operator
I want to have easy to reference steps to deploy BN in a single node k8s environment
To ensure that node operator who want to deploy to bare metal can do so easily

**Technical Notes**
This requires 
- deploying the BN to a single node k8s environment capturing the step
- repeating them from a clean setup with no assumed knowledge
- creating or updating an in repo doc with the steps

### LINKED ISSUE #1701: Document Simple Single k8s deployment flow
### Story Form

As a block node operator
I want to capture easy to follow deployment steps for single node scenarios
So that other operators have an easy path to deployment

### Technical Notes

Create a deployment README capturing steps to take using only open source tooling

Nana-EC:
## Reviewer Notes

For production purposes block nodes are recommended to be installed in a single node k8s style.
It's important to make this process clear and easy to reference for node operators of all types.
With this operators can customize and add on their own business needs

- Add a step by steps doc for operators to follow, calling out prerequisites and recommendations
- Add a Taskfile to help automate many steps
- Add provisioner script(s) for reference

## Related Issue(s)
Fixes #1701 
Fixes #527 
Fixes #1614

### DIFF
--- a/docs/block-node/operations/single-node-k8s-deployment.md
+++ b/docs/block-node/operations/single-node-k8s-deployment.md
@@ -49,13 +49,14 @@ streamlined and automated fashion.
 To utilize the Solo Weaver experience
 
 1. Install Weaver on the server
+
 ```bash
 curl -sSL https://raw.githubusercontent.com/hashgraph/solo-weaver/main/install.sh | bash
 weaver --help
 ```
 
 2. Run the provisioning and install flow
-Follow the [Setup Block Node](https://github.com/hashgraph/solo-weaver/blob/main/docs/quickstart.md#setup-block-node) steps.
+   Follow the [Setup Block Node](https://github.com/hashgraph/solo-weaver/blob/main/docs/quickstart.md#setup-block-node) steps.
 
 ### Custom Provisioning Script
 


charts/block-node-server/single-node-k8s-deployment.md
```diff
@@ -0,0 +1,159 @@
+# Single Node Kubernetes Deployment
```
Nana-EC: Add a reference here from the charts README after the current PR is merged.

charts/block-node-server/scripts/Taskfile.yml
```diff
@@ -0,0 +1,68 @@
...
+            - tar -xzf block-node-protobuf-$VERSION.tgz -C ~/block-node-protobuf-$VERSION
+            - rm block-node-protobuf-$VERSION.tgz
+
+    setup-grpcurl:
+        cmds:
+            - curl -L https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz -o grpcurl.tar.gz
+            - sudo tar -xzf grpcurl.tar.gz -C /usr/local/bin grpcurl
+            - rm grpcurl.tar.gz
+
+    scale-bn:
```
AlfredoG87: why we call this job `scale-bn` is more of a restart-bn

charts/block-node-server/single-node-k8s-deployment.md
```diff
@@ -0,0 +1,159 @@
...
+   curl -L "${BASE_URL}/${ARCHIVE}" -o "${ARCHIVE}"
+
+   # Ensure unzip is available
+   command -v unzip >/dev/null 2>&1 || sudo apt-get install -y unzip
+
+   mkdir -p "${DEST_DIR}"
+   tar -xzf "${ARCHIVE}" -C "${DEST_DIR}"
+   # This is needed as tar doesnt support to overwrite the existing tar
+   tar -xzf "${DEST_DIR}"/block-node-protobuf-${VERSION}.tgz -C "${DEST_DIR}"
+   rm "${ARCHIVE}" "${DEST_DIR}/${TGZ}"
```
AlfredoG87: similar to the above comment, but I see that we might want to improve the `setup-bn-proto` to install `unzip` if missing.

charts/block-node-server/scripts/Taskfile.yml
```diff
@@ -0,0 +1,68 @@
...
+        cmds:
+            - helm upgrade $RELEASE oci://ghcr.io/hiero-ledger/hiero-block-node/block-node-server --version $VERSION -n $NAMESPACE  --install --values values-override/bare-metal-values.yaml
+            - kubectl get all -n $NAMESPACE
+
+    reset-file-store:
+        cmds:
+            - kubectl -n ${NAMESPACE} exec ${POD} -- sh -c 'rm -rf /opt/hiero/block-node/data/live/* /opt/hiero/block-node/data/historic/*'
+            - kubectl -n ${NAMESPACE} delete pod $POD
+
+    reset-upgrade:
```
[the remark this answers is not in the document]
Nana-EC: All resets should clear store to make BN be new. Only consideration is if you upgrade along with the reset.
I added the restart prior but comments would be good agreed

charts/block-node-server/scripts/Taskfile.yml
```diff
@@ -0,0 +1,68 @@
...
+        cmds:
+            - kubectl -n ${NAMESPACE} exec ${POD} -- sh -c 'rm -rf /opt/hiero/block-node/data/live/* /opt/hiero/block-node/data/historic/*'
+            - kubectl -n ${NAMESPACE} delete pod $POD
+            - helm upgrade $RELEASE oci://ghcr.io/hiero-ledger/hiero-block-node/block-node-server --version $VERSION -n $NAMESPACE  --install --values values-override/bare-metal-values.yaml
+            - sleep 90
+            - kubectl get all -n $NAMESPACE
```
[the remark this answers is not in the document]
Nana-EC: Yes, will update

charts/block-node-server/single-node-k8s-deployment.md
```diff
@@ -0,0 +1,159 @@
...
+   # [org.hiero.block.node.app.BlockNodeApp start] Started BlockNode Server : State = RUNNING, Historic blocks =
+   ```
+8. **Access Block Node Server**: Connect to the Block Node Server using the configured service endpoint.
+
+   Install `grpcurl` if not already installed:
+
+   ```bash
+   curl -L https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz -o grpcurl.tar.gz
+   sudo tar -xzf grpcurl.tar.gz -C /usr/local/bin grpcurl
+   rm grpcurl.tar.gz
```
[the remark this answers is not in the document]
Nana-EC: Good note, I wrote this before creating the Taskfile

charts/block-node-server/single-node-k8s-deployment.md
Nana-EC: i thought we don't need the unzip option anymore?

charts/block-node-server/scripts/linux-bare-metal-provisioner.sh
```diff
@@ -0,0 +1,518 @@
...
+readonly CILIUM_VERSION="1.18.1"
+readonly METALLB_VERSION="0.15.2"
+readonly DASEL_VERSION="2.8.1"
+
+# Update System Packages
+sudo apt update && sudo apt upgrade -y
+
+# Disable Swap (necessary for kubeadm)
+# WARNING: This failed in docker container
+sudo sed -i.bak 's/^\(.*\sswap\s.*\)$/#\1\n/' /etc/fstab
```
jsync-swirlds: This line prevents the next line from working correctly, and is strictly unnecessary if the goal is just to disable swap for this script.
If we need to disable swap for the host system permanently, then we may need a more sophisticated fix than a simple sed regex (at minimum the sed command should be prefixed with a check whether there exists a swap device in fstab).

charts/block-node-server/scripts/linux-bare-metal-provisioner.sh
```diff
@@ -0,0 +1,518 @@
...
+net.ipv4.tcp_tw_reuse = 1
+net.ipv4.tcp_rfc1337 = 1
+net.ipv4.tcp_syncookies = 1
+net.ipv4.tcp_fin_timeout = 15
+net.ipv4.tcp_keepalive_time = 300
+net.ipv4.tcp_keepalive_intvl = 15
+fs.file-max = 2097152
+vm.swappiness = 10
+vm.dirty_ratio = 60
+vm.dirty_background_ratio = 2
```
jsync-swirlds: If we are trying to disable swap, swappiness should be `0`.
If we want "bare minimum swap, only when absolutely needed" then it should be set to `1`.
This value is a "please don't swap, but if we're running out of disk cache, then go ahead.

The next two lines push for a bit more disk cache than normal, but aren't that aggressive.

Given we tried to completely remove swap devices on line 36; we might not have the exact correct values here.

charts/block-node-server/scripts/linux-bare-metal-provisioner.sh
```diff
@@ -0,0 +1,518 @@
...
+net.ipv4.udp_rmem_min = 16384
+net.ipv4.tcp_wmem = 8192 65536 33554432
+net.ipv4.udp_wmem_min = 16384
+net.ipv4.tcp_max_tw_buckets = 1440000
+net.ipv4.tcp_tw_reuse = 1
+net.ipv4.tcp_rfc1337 = 1
+net.ipv4.tcp_syncookies = 1
+net.ipv4.tcp_fin_timeout = 15
+net.ipv4.tcp_keepalive_time = 300
+net.ipv4.tcp_keepalive_intvl = 15
```
jsync-swirlds: We might want to do some performance tuning on these...

charts/block-node-server/scripts/linux-bare-metal-provisioner.sh
```diff
@@ -0,0 +1,518 @@
...
+fs.file-max = 2097152
```
jsync-swirlds: This is going to cause issues with block nodes.
We process a lot of files, and this sets a hard limit of 2 million file descriptors.
The kernel default is absurdly large (typically max_ulong), so setting it smaller is more likely to create problems for block node than to improve anything.

jsync-swirlds (approved):
Good first pass on this.

Issue #1072, opened 2025-04-29
Ranged Sets and Ranges Expected Behavior


ata-nas:
### Story Form

As a Block Node Developer
I want to have a determined and final decision on the behavior of the ranged sets and ranges that we use
So that I can develop based on the agreed upon expected behavior

### Technical Notes

- based on [this comment thread](https://github.com/hiero-ledger/hiero-block-node/pull/1064#discussion_r2058642221), let's discuss and explore the few mentioned topics:
  - should ranges be inclusive vs exclusive
  - should ranges support `-1 -> -1` (which is problematic and throws exceptions, and is indeed the thing that lead to this issues) and the notion of an "empty" range
  - can we consider that a ranged set is empty when it has no ranges inside of it (`null`) rather than it having an "empty" range
- based on discussion/agreement, some changes might need to be made in the code base

ata-nas:
After a discussion with the team, we reached the conclusion:

1. Ranges should be inclusive:
   - a range of `0 -> 0` means that `0` is the only valid contained value in the range
1. We can drop the notion of an empty `Range`
   - `RangedSet`s will be empty when they do not have any `Range` classes inside them (`null`)
   - we should update the existing logic to not allow any negative values within ranges and "clearing" a `RangedSet` means that we remove all `Range` instances inside the set

cc: @Nana-EC @jasperpotts @jsync-swirlds @AlfredoG87 @georgi-l95

PR #1887, opened 2025-11-21
feat: Add logic for greedy backfill

### LINKED ISSUE #1886: feat(backfill): Greedy Backfill proactively fills in new blocks
### Story Form

As a block node engineer
I want to ensure that block nodes with no publishers are able to keep up with the network 
So that a block node doesn't remain too far behind the network when there are blocks available on other BNs

### Technical Notes

Today Tier 1 BN Backfill has 2 modes
- **autonomous**: given a start block number in configs it looks for gaps from that start to the earliest block the BN has and attempts to retrieve missing blocks from other available BNs
- **on-demand**: given a `TOO_FAR_BEHIND` notification from the CN, the BN will try to retrieve blocks from it's latest persisted block up to the block noted in the `TOO_FAR_BEHIND`

Whiles good this leaves a gap of what happens when a BN has no publishers pushing blocks to it or has publishers not sending new blocks (dealing with duplicated etc) but the network is progressing and other BNs have the blocks.
In this case we want BNs where possible to proactively retrieve new blocks that exist on other block nodes.

We expect due to network latency it may happen that CNs switch around BNs to push blocks, in this case it's likely for a BN to fall behind and we want it to be close enough to live should a CN ever switch back to it.

This may be seen as a subset of the on-demand backfill mode caused by inactivity.

The main changes should be
- Update `autonomous` mode to utilize the `serverStatusResponse.lastAvailableBlock` as the threshold
- Update `on-demand` mode to utilize the `serverStatusResponse.lastAvailableBlock` as the threshold
- Add logic to `PublisherPlugin` that periodically issues a `NewestBlockKnownToNetworkNotification` every 10 secs

Considerations
- Backfill should not query for blocks it already has, to avoid wasting resources if we get greedy we should only ask for gaps
- For inactivity on-demand backfill if we ignore new notification we may not request enough blocks, this could be an issue in edge cases but initially it may be fine to ignore. In future do we need an optimization? Potentially backfill just updates it's higher threshold

Nana-EC:
## Reviewer Notes

Current backfill logic doesn't account for blocks that are available from other blocks when BN has no publishers pushing blocks to it or has publishers not sending new blocks (dealing with duplicated etc) but the network is progressing and other BNs have the blocks.
In this case we want BNs where possible to proactively retrieve new blocks that exist on other block nodes.

We expect due to network latency it may happen that CNs switch around BNs to push blocks, in this case it's likely for a BN to fall behind and we want it to be close enough to live should a CN ever switch back to it.

- Add `backfill.greedy` config to toggle enablement of greedy backfill
- Add a `getNewAvailableRange()` to `BackfillGrpcClient` to determine the available range from peer BNs
- Update autonomous mode to search for blocks up to the max of `serverStatusResponse.lastAvailableBlock` available from other BNs
- Update on-demand mode to skip run if greedy autonomous is running
- Update ThreadPoolManager with support to `createVirtualThreadScheduledExecutor`. Update all test references and impacts

**Notes to Reviewer**

## Related Issue(s)
Fixes #1886

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -2,7 +2,6 @@
 package org.hiero.block.node.backfill;
 
 import static java.lang.System.Logger.Level.DEBUG;
-import static java.lang.System.Logger.Level.ERROR;
 import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.TRACE;
 
@@ -228,7 +227,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
                 .createVirtualThreadScheduledExecutor(
                         2, // Two threads: one for autonomous backfill, one for on-demand backfill
                         "BackfillPluginRunner",
-                        (t, e) -> LOGGER.log(ERROR, "Uncaught exception in thread: " + t.getName(), e));
+                        (t, e) -> LOGGER.log(INFO, "Uncaught exception in thread: " + t.getName(), e));
 
         scheduler.scheduleAtFixedRate(
                 this::detectGaps,


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -96,6 +103,18 @@ public LiveStreamPublisherManager(
...
+        this.scheduledExecutorService = context.threadPoolManager()
+                .createSingleThreadScheduledExecutor(
+                        "StreamPublisherRunner",
+                        (t, e) -> LOGGER.log(ERROR, "Uncaught exception in thread: " + t.getName(), e));
+        // schedule a notification to prompt backfill to proactively look for newer blocks from other block nodes
+        this.scheduledExecutorService.scheduleWithFixedDelay(
+                () -> notifyTooFarBehind(UNKNOWN_BLOCK_NUMBER),
+                publisherConfiguration.noActivityNotificationIntervalSeconds(),
+                publisherConfiguration.noActivityNotificationIntervalSeconds(),
+                TimeUnit.SECONDS);
```
jsync-swirlds: I don't think we need this at all.
Backfill already has scheduled activity and could check for changes in latest available block (rather than publisher blind sending an event assuming it's too far behind)

AlfredoG87: I finished listening the recording and catching up with followup conversion offline.

I agree that if we are just sending this message from the publisher every X amount of time, and not using the `no publishers` condition we discussed initially. then it does not make sense to have  a new scheduled executor on the publisher.

however I think we can re-use the work you did for the scheduled executor within the backfill plugin to schedule the `autonomous` task that checks for gaps.

and as part of that check, if greedy mode is enabled we can check for newer blocks as well...

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -271,21 +271,39 @@ private void detectGaps() {
...
                     .toList();
 
-            // Calculate total missing blocks
-            long pendingBlocks = 0;
+            // add block range available from other BN sources
+            LongRange detectedRecentGapRange = backfillGrpcClientAutonomous.getNewAvailableRange(
+                    blockRanges.getLast().end());
+            if (detectedRecentGapRange.size() > 0 && detectedRecentGapRange.start() >= 0) {
+                detectedGaps = new ArrayList<>();
+                detectedGaps.add(detectedRecentGapRange);
```
AlfredoG87: Not sure if we want to wrap this in a feature flag to determine if we want autonomous backfill to be greedy.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
```diff
@@ -11,7 +12,8 @@
  */
 @ConfigData("producer")
 public record PublisherConfig(
-        @Loggable @ConfigProperty(defaultValue = "9_223_372_036_854_775_807") @Min(100_000L) long batchForwardLimit) {
+        @Loggable @ConfigProperty(defaultValue = "9_223_372_036_854_775_807") @Min(100_000L) long batchForwardLimit,
+        @Loggable @ConfigProperty(defaultValue = "10") @Min(5) @Max(60) int noActivityNotificationIntervalSeconds) {
```
AlfredoG87: If we use the same scheduled task of autonomous for greedy we don't even need this property.

AlfredoG87 (commented):
First pass is looking good.

I agree with Joseph recomendation to avoid triggering the process from the publisher Manager since we don't care if we have or not active publishers in order to backfill in greedy mode.

There is a method `getBackfillTypeForBlock` that determines if the PersistenceNotification is coming from a autonomous or on-demand flow based on the  block number, after this changes we cannot longer trust this method for determining the source. Here I would recommend switching for a map for each flow and store there the blocs that were submitted for verification+persistence and backfill is awaiting to complete before moving on to the next batch.

or even fire and forget and don't care what happens after blocks are submitted, this approach might greatly simplify logic, but might also potentially create some queue saturation if the BN is too slow.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
Nana-EC: Good note.
This is left over design from the first part where it was staged on no publisher.
I should have pulled this out after the design approach update.
Will do this and it should likely simplify all the test changes.
Will fix that up

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
Nana-EC: Removed

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
Nana-EC: Done

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/TestUtils.java
```diff
@@ -37,7 +37,8 @@ public static ConfigurationBuilder createTestConfiguration() {
                 .withConfigDataType(com.swirlds.common.metrics.config.MetricsConfig.class)
                 .withConfigDataType(com.swirlds.common.metrics.platform.prometheus.PrometheusConfig.class)
                 .withConfigDataType(org.hiero.block.node.app.config.ServerConfig.class)
-                .withConfigDataType(org.hiero.block.node.app.config.node.NodeConfig.class);
+                .withConfigDataType(org.hiero.block.node.app.config.node.NodeConfig.class)
+                .withValue("prometheus.endpointEnabled", "false");
```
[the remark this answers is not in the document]
Nana-EC: Thanks, i think it is needed in a few other places also

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -249,43 +257,106 @@ public void stop() {
...
+            LOGGER.log(TRACE, "Greedy backfilling recent blocks to stay close to network");
+
+            // greedy backfill newer blocks available from peer BN sources
+            detectedGaps = new ArrayList<>();
+            LongRange detectedRecentGapRange = backfillGrpcClientAutonomous.getNewAvailableRange(lastAcknowledgedBlock);
+
+            // to-do: consider on-demand backfill range and remove from detectedRecentGapRange if overlapping
+            if (isOnDemandBackfillRunning()) {
+                // if on-demand backfill is running, we need to adjust the detected recent gap range
+            }
```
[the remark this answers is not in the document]
Nana-EC: Meant to comeback but don't think I'll get to it with room to go into 0.24.
So might pull this out to a follow up PR for 0.25

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillGrpcClient.java
```diff
@@ -117,6 +117,58 @@ private LongRange getAvailableRangeInNode(BlockNodeClient node, LongRange blockR
...
+                INFO,
+                "Determined block range from peer blocks nodes earliestPeerBlock={0,number,#} to latestStoredBlockNumber={1,number,#}",
+                earliestPeerBlock,
+                latestPeerBlock);
+
+        // confirm next block is available if not we still can't backfill
+        if (latestStoredBlockNumber + 1 < earliestPeerBlock
+                || latestStoredBlockNumber > latestPeerBlock
+                || latestStoredBlockNumber + 1 > latestPeerBlock) {
+            return new LongRange(-1, -1);
```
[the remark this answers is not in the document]
Nana-EC: Fixed

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
```diff
@@ -95,4 +96,30 @@ public ExecutorService createSingleThreadExecutor(
...
+        if (threadName != null) {
+            factoryBuilder.name(threadName, 0);
+            return corePoolSize <= 1
+                    ? Executors.newSingleThreadScheduledExecutor(factoryBuilder.factory())
+                    : Executors.newScheduledThreadPool(corePoolSize, factoryBuilder.factory());
+        } else {
+            final ThreadFactory factory = factoryBuilder.factory();
+            return corePoolSize <= 1
+                    ? Executors.newSingleThreadScheduledExecutor(factory)
+                    : Executors.newScheduledThreadPool(corePoolSize, factory);
```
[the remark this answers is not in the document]
ata-nas: Or, you could simply add what we are currently using, i.e. a non-single threaded pool. We can always add the method to get a single threaded pool when we need to do so. No point in having dead code, I agree with you on that completely.

block-node/spi/src/main/java/org/hiero/block/node/spi/historicalblocks/LongRange.java
```diff
@@ -27,7 +27,9 @@ public record LongRange(long start, long end) implements Comparable<LongRange> {
      */
     public LongRange {
         // Special case: allow both start and end to be -1 for clean state initialization
-        if (!(start == -1 && end == -1)) {
+        if (start == -1 && end == -1) {
+            throw new IllegalArgumentException("Both start and end should not be -1 simultaneously.");
```
jsync-swirlds: Slightly concerned here as we have a somewhat large array of places that assume `{-1,-1}` is "empty".  Might be better to leave this change to be done under issue #1072.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillGrpcClient.java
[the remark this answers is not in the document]
Nana-EC: Not resolving 1072. Just the instantiation of -1,-1.
Th change was simple enough and test updated. There didn't seem to be other usages and all tests UTs and E2Es highlight expected operations.

block-node/spi/src/main/java/org/hiero/block/node/spi/historicalblocks/LongRange.java
Nana-EC: Hmm, I only found one instances and all tests cases are operating correctly after the update and test addition.
Do we not expect that all the UTs and E2E to have caught that?
It throws if anyone uses it so it shouldn't go unnoticed.

Nana-EC: Okay, if you feel uncomfortable I can backtrack this change, it's minor and has no impact on the PR.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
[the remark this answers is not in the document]
Nana-EC: Updated with intended logic

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeAPITests.java
```diff
@@ -549,8 +549,8 @@ void e2eSocketClosureTest() throws InterruptedException {
             // Expected exception to be thrown on the onNext() due to closed socket
             // from previous duplicate block publish
             requestStream.onNext(request1);
-            publishCompleteCountDownLatch.await(); // wait
-            endBlock(blockNumber1, requestStream);
+            //            publishCompleteCountDownLatch.await(); // wait
+            //            endBlock(blockNumber1, requestStream);
```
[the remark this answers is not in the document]
Nana-EC: Reverted

AlfredoG87 (approved):
Looking Good, can't wait to test it out in the wild 😅

block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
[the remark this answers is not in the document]
ata-nas: Just to note: for the future, shall we consider to reuse pools as we do with the other ones, with the map as exception handler key? Not asking for it now, just noting. If we do, we should be careful as to how it would affect the whole system post changing it in the future.

ata-nas (approved):

PR #1891, opened 2025-11-21
chore: Improvements to Block Footer and Proof Support

Nana-EC:
## Related Issue(s)

### DIFF
--- a/block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
+++ b/block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
@@ -7,6 +7,9 @@ import static org.hiero.block.common.hasher.HashingUtilities.getBlockItemHash;
 import static org.hiero.block.common.hasher.HashingUtilities.noThrowSha384HashOf;
 
 import com.hedera.hapi.block.stream.BlockProof;
+import com.hedera.hapi.block.stream.SignedRecordFileProof;
+import com.hedera.hapi.block.stream.StateProof;
+import com.hedera.hapi.block.stream.TssSignedBlockProof;
 import com.hedera.hapi.block.stream.output.BlockFooter;
 import com.hedera.hapi.block.stream.output.BlockHeader;
 import com.hedera.pbj.runtime.ParseException;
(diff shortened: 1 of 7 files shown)


protobuf-sources/src/main/proto/internal/unparsed.proto
```diff
@@ -66,5 +66,7 @@ message BlockItemUnparsed {
         bytes block_proof = 9;
         bytes record_file = 10;
         bytes trace_data = 11;
+        bytes block_footer = 12;
+        bytes redacted_item = 19;
```
jsync-swirlds: Just a minor note.
1. This is the only change strictly needed to fix the observed error; all the rest can be deferred to a later PR.
2. We probably should take care with `redacted_item` here, because a Publisher _must not ever_ send a redacted stream, and attempt to do so should be detected and rejected. That's more a task for later work that fills out the full block verification process, however.

Nana-EC:
Leaving in draft until @AlfredoG87 's PR #1891 goes in.
I can more easily rebase my PR to add some tests and other support I found useful here.
More important for verification logic to go in sooner

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -94,16 +100,31 @@ public VerificationNotification processBlockItems(List<BlockItemUnparsed> blockI
...
+                        }
+                    }
+
+                    blockProofsReceived++;
                 }
+                case REDACTED_ITEM -> LOGGER.log(WARNING, "Redacted item observed in block {0}", blockNumber);
+                case FILTERED_ITEM_HASH -> LOGGER.log(WARNING, "Filtered item hash observed in block {0}", blockNumber);
+                case RECORD_FILE -> LOGGER.log(WARNING, "Record file observed in block {0}", blockNumber);
+                case UNSET -> LOGGER.log(WARNING, "Unset block item observed in block {1}", kind, blockNumber);
+                default -> LOGGER.log(WARNING, "Unknown block item type {0} in block {1}", kind, blockNumber);
```
jsync-swirlds: We shouldn't have default; the enumerated value is comprehensive and we should ensure every known enum is handled (which the compiler will only do if there is no default).
Any enumerated type we don't yet handle should be in a different place (the unknown values map); not mixed in with existing block items.

jsync-swirlds (commented):
A few concerns we might need to address.

AlfredoG87:
@Nana-EC @jsync-swirlds should we close this PR? 

We've been doing lots of improvements and changes to the classes in question since November.

PR #1902, opened 2025-11-25
chore: add signature validation listing command + correct hour-based GCP lookup & local comparison logic

rockysingh:
### #Summary 
This PR introduces the new sig-validation-ls command and completes the design corrections requested during review. The command randomly samples (date, hour) ranges and compares GCP signature counts against locally derived counts from compressed .tar.zst day archives.

**Key Changes**
	•	Implemented sig-validation-ls command to perform hour-level signature validation across sampled days.
	•	Corrected GCP lookup logic:
(date + hour) is now converted to a UTC Instant and then to block-time nanos via
RecordFileDates.instantToBlockTimeLong, ensuring MainNetBucket.listHour receives the correct input.
	•	Reworked comparison logic (compareHour):
	•	Explicit union of GCP + local timestamp keys
	•	Missing timestamps are treated as mismatches
	•	Only equal counts on both sides increment the same counter
	•	Fixed PerTimestampDiff record and ensured consistent usage throughout the command.
	•	Updated and clarified Javadocs to match the new behaviour and better explain how timestamps and signature counts are derived.
	•	Added a clean, deterministic test that injects synthetic GCP and local summaries without relying on filesystem IO.

**Why This Matters**

This corrects the core design flaw where signature validation was based only on the day rather than the specific hour. It also aligns the logic with expected behaviour for detecting incomplete or mismatched signature sets across GCP and local compressed archives.

**Testing**
	•	Added unit test for compareHour using controlled synthetic input.
	•	Manually validated against 2019-09-13 and 2019-09-14 compressed days.

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/ValidateSignatureCounts.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/ValidateSignatureCounts.java
@@ -54,8 +54,8 @@ public class ValidateSignatureCounts implements Runnable {
     CommandSpec spec;
 
     @Option(
-        names = {"-d", "--downloaded-days-dir"},
-        description = "Directory where downloaded days are stored")
+            names = {"-d", "--downloaded-days-dir"},
+            description = "Directory where downloaded days are stored")
     private File compressedDaysDir = new File("compressedDays");
 
     @Parameters(index = "0", description = "From year")


tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/GcpLsCommand.java
```diff
@@ -0,0 +1,374 @@
...
+        description = "Randomly sample (day, hour) and compare GCP LS vs local downloadedDays signature counts")
+public class GcpLsCommand implements Runnable {
+
+    @Option(names = "--start-date", required = true, description = "Start date (inclusive), yyyy-MM-dd")
+    private LocalDate startDate;
+
+    @Option(names = "--end-date", required = true, description = "End date (inclusive), yyyy-MM-dd")
+    private LocalDate endDate;
+
+    @Option(names = "--samples", description = "Number of random (day, hour) samples", defaultValue = "10")
```
Nana-EC: Might be worth considering a max to not overwhelm a user.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/GcpLsCommand.java
```diff
@@ -0,0 +1,374 @@
...
+        for (final NodeResult r : results) {
+            String status = (r.diff == 0 && r.gcpSigCount >= 0) ? "OK" : "MISMATCH";
+            if (r.gcpSigCount < 0) {
+                status = "ERROR";
+            }
+            System.out.printf(
+                    "%-8s %-10d %-10d %-10d %-8s%n", "0.0." + r.node, r.gcpSigCount, r.localSigCount, r.diff, status);
+        }
+
+        printDifferenceView(results);
```
Nana-EC: Might be worth a reorder `printView(List<NodeResult> results)` which iterates over `results` and then calls `printDifferenceView(NodeResult result)` and `printLsView(NodeResult result)` based on appropriate flags.
This allows for a single loop of results which could be useful is samples are large

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/GcpLsCommand.java
```diff
@@ -0,0 +1,374 @@
...
+    /**
+     * Check one node for a given date+hour:
+     *
+     *  - GCP: uses MainNetBucket.listHour(date),
+     *         then filters blobs for this node and record file extensions.
+     *  - Local: counts *.rcd / *.rcd.gz files in downloadedDaysDir for date+hour+node.
+     */
+    private NodeResult checkNodeForHour(LocalDate date, int hour, int node) {
+        List<String> gcpSigFiles = List.of();
+        long gcpSigCount;
```
Nana-EC: For clarity I think you should default the counts and diff to 0, then set them to -1 on failures or skips (skips should be clear as the printouts should input)
Then in the else that sets `localRecordFiles` below it should be an `if else` conditions.
In this way you can have the logic fail fast and skip certain lines

Nana-EC (commented):
Nice updates.
Some suggestions where applicable

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/GcpLsCommand.java
```diff
@@ -0,0 +1,399 @@
...
+
+    @Override
+    public void run() {
+        try {
+            if (mainNetBucket == null) {
+                mainNetBucket = new MainNetBucket(
+                        cacheEnabled, cacheDir.toPath(), minNodeAccountId, maxNodeAccountId, userProject);
+            }
+            if (!skipLocalCompare && downloadedDaysDir == null) {
+                throw new NotFoundException("DownloadedDaysDir has not been set");
```
Nana-EC: Just use an `IllegalArgumentException` here also as this only happens when `downloadedDaysDir` is invalid.
Seems too much to add helidon `NotFoundException` just for this

Nana-EC (approved):
LG
Minor suggestion.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/GcpLsCommand.java
```diff
@@ -0,0 +1,398 @@
...
+        List<String> localRecordFiles = List.of();
+
+        try {
+            var blobs = mainNetBucket.listHour(date.toEpochDay());
+            final String hourToken = date + "T" + String.format("%02d", hour) + "_";
+
+            gcpSigFiles = blobs.stream()
+                    .map(b -> b.path())
+                    .filter(path -> path.contains(hourToken))
+                    .filter(this::isRecordName)
```
jasperpotts: this is wrong isRecordName looks for ".rcd" or ".rcd.gz". Signature files have names like "2019-09-13T21_53_51.396440Z.rcd_sig"

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/GcpLsCommand.java
```diff
@@ -0,0 +1,398 @@
...
+                    .filter(path -> matchesNode(path, node))
+                    .map(this::extractFileName)
+                    .toList();
+
+            gcpSigCount = gcpSigFiles.size();
```
jasperpotts: this will almost always be 1 and will tell you there is a single record file not that there is a signature file.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/GcpLsCommand.java
```diff
@@ -0,0 +1,398 @@
...
+    private void handleSample(
+            final int sampleIndex, final LocalDate sampleDate, final int hour, final ExecutorService executor)
+            throws ExecutionException, InterruptedException {
+
+        System.out.printf("%n=== Sample %d: date=%s hour=%02d ===%n", sampleIndex, sampleDate, hour);
+        System.out.printf("%-8s %-10s %-10s %-10s %-8s%n", "Node", "gcpFiles", "localFiles", "diff", "status");
+
+        // submit node tasks in parallel
+        final List<Future<NodeResult>> futures = new ArrayList<>();
+        for (int node = 0; node < nodeCount; node++) {
```
jasperpotts: Not sure why you look node's here. Should do all nodes at once. A single GCP ls using org.hiero.block.tools.utils.gcp.MainNetBucket.listHour will include all nodes. That is one of the features of  MainNetBucket class.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/GcpLsCommand.java
```diff
@@ -0,0 +1,398 @@
...
+            try {
+                for (int i = 0; i < samples; i++) {
+                    long dayOffset = ThreadLocalRandom.current().nextLong(daysRange);
+                    LocalDate sampleDate = startDate.plusDays(dayOffset);
+                    int hour = ThreadLocalRandom.current().nextInt(24);
+
+                    handleSample(i + 1, sampleDate, hour, executor);
+                }
+            } finally {
+                executor.shutdownNow();
```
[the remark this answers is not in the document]
jasperpotts: ok I see, handleSample blocks on future.get() So you have parallelism across nodes but not samples. Feel you can move threading up a level and do multiple samples in parallel.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/GcpLsCommand.java
```diff
@@ -0,0 +1,398 @@
...
+
+        List<String> gcpOnly = List.of();
+        List<String> localOnly = List.of();
+        List<String> localRecordFiles = List.of();
+
+        try {
+            var blobs = mainNetBucket.listHour(date.toEpochDay());
+            final String hourToken = date + "T" + String.format("%02d", hour) + "_";
+
+            gcpSigFiles = blobs.stream()
```
jasperpotts: change to filter to sig files only, grouopby timestamp, then count and produce map.

jasperpotts (changes requested):

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/SignatureValidationListingCommand.java
```diff
@@ -0,0 +1,335 @@
...
+
+        return new HourResult(date, hour, same, diffs);
+    }
+
+    private String extractFileName(String path) {
+        int idx = path.lastIndexOf('/');
+        return idx >= 0 ? path.substring(idx + 1) : path;
+    }
+
+    static final class SigSummary {
```
jasperpotts: this can be a "record"

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/SignatureValidationListingCommand.java
```diff
@@ -0,0 +1,335 @@
...
+        final java.util.Map<String, java.util.List<String>> filesByTimestamp;
+
+        protected SigSummary(
+                java.util.Map<String, Integer> counts, java.util.Map<String, java.util.List<String>> filesByTimestamp) {
+            this.counts = counts;
+            this.filesByTimestamp = filesByTimestamp;
+        }
+    }
+
+    static final class PerTimestampDiff {
```
jasperpotts: this can be a "record" too

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/SignatureValidationListingCommand.java
```diff
@@ -0,0 +1,335 @@
...
+                java.util.List<String> localFiles) {
+            this.timestamp = timestamp;
+            this.gcpCount = gcpCount;
+            this.localCount = localCount;
+            this.gcpFiles = gcpFiles;
+            this.localFiles = localFiles;
+        }
+    }
+
+    static final class HourResult {
```
jasperpotts: and this

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/SignatureValidationListingCommand.java
```diff
@@ -0,0 +1,335 @@
...
+                same++;
+            } else {
+                diffs.add(new PerTimestampDiff(ts, gcp, local, gcpFilesForTs, localFilesForTs));
+            }
+        }
+
+        return new HourResult(date, hour, same, diffs);
+    }
+
+    private String extractFileName(String path) {
```
jasperpotts: See org.hiero.block.tools.records.RecordFileDates.extractRecordFileTime, can just use that I expect

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/SignatureValidationListingCommand.java
```diff
@@ -0,0 +1,335 @@
...
+    protected SigSummary buildGcpSigSummary(LocalDate date, int hour) {
+        java.util.Map<String, Integer> counts = new java.util.HashMap<>();
+        java.util.Map<String, java.util.List<String>> filesByTimestamp = new java.util.HashMap<>();
+        final String hourToken = date + "T" + String.format("%02d", hour) + "_";
+
+        try {
+            // Scan the whole day and then filter to the hour, as requested in review.
+            var chainFilesList = mainNetBucket.listHour(date.toEpochDay(), false);
+
+            chainFilesList.stream()
```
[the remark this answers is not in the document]
rockysingh: done.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/SignatureValidationListingCommand.java
rockysingh: done

PR #1907, opened 2025-11-26
feat: Updated protocol definition for publishers with new changes

### LINKED ISSUE #1914: BN closes connection after sending BEHIND
### Description

It may occur that a publisher sends a BN a block that is ahead of the desired next block e.g. publisher sends block 100 but BN lack acknowledged block 75.
according to the protocol the BN should send a BEHIND status and be open to receiving follow up communication including a TOO_FAR)BEHIND or other block items if the publisher has the next set of blocks i.e. block 76 in this case.

Instead the BN closes the connection and publishers observe a `SocketException: Socket closed` error.

The BN should likely close the stream only but not the connection to allow the publisher to either follow up and or close the connection.

### Steps to Reproduce

1. Deploy a BN
2. As a publisher send a block ahead of BNs last acknowledge number
3. Observe BN BEHIND response
4. Attempt to send either TOO_FAR_BEHIND or other block items on the same connection
5. Observe socket closed exception

### Additional context

The protocol is still in beta and thus in the early stages and this is an instance of a gap in design highlighted through testing.
The long term fix is to update the protocol treating BEHIND not as an EndOfStream but at the same level as SkipBlock or ResendBlock. This will require protobuf changes and updates to CN v0.70 to handle the new breaking changes of the beta protocol

jsync-swirlds:
## Reviewer Notes
* Updated the API definition with the changes for `BehindPublisher`.
* Updated the node service API with the start of a new "server status details" API.
* Updated a distressing array of test classes, mostly due to tests that basically cross-checked the _compiler_, not our design.

Fixes #1914

### DIFF
--- a/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamObserver.java
+++ b/tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamObserver.java
@@ -14,7 +14,10 @@ import java.util.Deque;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import org.hiero.block.api.protoc.PublishStreamResponse;
+import org.hiero.block.api.protoc.PublishStreamResponse.BehindPublisher;
 import org.hiero.block.api.protoc.PublishStreamResponse.BlockAcknowledgement;
+import org.hiero.block.api.protoc.PublishStreamResponse.ResendBlock;
+import org.hiero.block.api.protoc.PublishStreamResponse.SkipBlock;
 import org.hiero.block.simulator.startup.SimulatorStartupData;
 
 /**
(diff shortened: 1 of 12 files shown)


protobuf-sources/src/main/proto/block-node/api/node_service.proto
```diff
@@ -14,6 +14,29 @@ import "services/basic_types.proto";
...
+/**
+ * A response to a server status detail request.
+ *
+ * This message SHALL provide a client with extended detail information needed
+ * to successfully query this Block-Node for a block, stream of blocks, or
+ * state snapshot.<br/>
+ * This API SHALL provide detailed range information that describes exactly
+ * which ranges of blocks are available and which are not. This may be a very
+ * large set of data (potentially millions of points in a degenerate case).<br/>
+ * A request for blocks within the `available_ranges` SHALL succeed. Any
```
ata-nas: `A request for blocks within the `available_ranges` SHALL succeed.`

Is that always true? It should be for full history nodes, but what about nodes that have retention policy which could cleanup blocks in the time we query for the ranges and then requesting the blocks themselves.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamObserver.java
```diff
@@ -62,15 +65,20 @@ public void onNext(final PublishStreamResponse publishStreamResponse) {
...
-            try {
-                startupData.updateLatestAckBlockStartupData(ack.getBlockNumber());
-            } catch (final IOException e) {
-                throw new UncheckedIOException(e);
-            }
-        } else if (publishStreamResponse.hasResendBlock() || publishStreamResponse.hasSkipBlock()) {
+            updateLastAcked(ack.getBlockNumber());
+        } else if (publishStreamResponse.hasResendBlock()) {
+            final ResendBlock resendValue = publishStreamResponse.getResendBlock();
+            updateLastAcked(resendValue.getBlockNumber());
```
ata-nas: Is this correct? I remember for some messages we send the block number last completed, so we have to send that +1, but in some cases, we send the block number currently expected, so we need to update with that directly.

ata-nas (commented):
Looks good, a few suggestions.

AlfredoG87 (approved):
LGTM

PR #1912, opened 2025-11-27
fix: api-tests timeout issue

AlfredoG87:
adding timeout for both threads and latches so test can fail/timeout faster instead of having to wait 30mins for the CI runner to cancel it.

Improved some tests to make it less likely to stall


Nana-EC:
Good idea.
Odd, the 2 tests that consistently fail are because they are waiting on the onComplete latch due to a duplicate block.
What's odd is the BN registers and handles the duplicate case fine but for some reason the latch countdown doesn't occur.
Can't reproduce locally but will see what other options we have

PR #1915, opened 2025-11-29
fix: Update PublishHandler to close only the stream on BEHIND

Nana-EC:
## Reviewer Notes
Currently the handler closes both the stream and connection when handling BEHIND code.
However, the expectation is the publisher would follow up either with more appropriate block items or a TOO_FAR_BEHIND.
When the publisher does it experiences a socket closed exception

- Update BatchHandleResult to separate shutdown, reset and close stream
- Update `handleBlockActionResult()` to manage 3 cases in a more coordinate fashion and not call either unnecessarily
- Add API Test scenario that shows publisher no longer observes socket exception

Partially address #1914 but does not attempt to fix the protocol definition as that requires CN coordination and larger changes

## Related Issue(s)

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
@@ -537,7 +537,7 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
      * the caller whether the handler should shut down and/or reset its current
      * block action and current streaming block number.
      */
-    private record BatchHandleResult(boolean shouldShutdown, boolean shouldReset) {}
+    private record BatchHandleResult(boolean shouldShutdown, boolean shouldReset, boolean shouldCloseStream) {}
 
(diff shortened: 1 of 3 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -537,19 +537,27 @@ public long handleFailedPersistence() {
...
-        if (handleResult.shouldReset()) {
-            resetState();
-        }
+        // close connection and clean up or handle partial clean-ups
         if (handleResult.shouldShutdown()) {
             shutdown();
+        } else {
+            // close stream leaving connection open for subsequent publishers calls
+            if (handleResult.shouldCloseStream()) {
+                replies.onComplete();
```
ata-nas: Not sure if that is the only thing we need to do to close the stream. I remember there were some changes recently that allow us to actually close the stream completely.

ata-nas (commented):
Just some preliminary notes to think about.

PR #1920, opened 2025-12-02
docs: change behind handling in publish protocol to not end stream

jsync-swirlds:
## Reviewer Notes
* Adjusted the documentation only.
    * Actual implementation is separate.
* Behind is no longer a code on `EndOfStream` but is a message on its own
    * In combination, the block node no longer closes the stream when behind, but
      waits for the Publisher to either stream an earlier block or
      send "too far behind".

### DIFF
--- a/docs/design/communication-protocol/publish-block-stream.md
+++ b/docs/design/communication-protocol/publish-block-stream.md
@@ -4,13 +4,13 @@
 
 1. [Abstract](#abstract)
    1. [Reliability note](#reliability-note)
-1. [Definitions](#definitions)
-1. [Base Protocol](#base-protocol)
+2. [Definitions](#definitions)
+3. [Base Protocol](#base-protocol)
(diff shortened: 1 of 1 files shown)


ata-nas (dismissed):
LG, as far as I understand it, the change is to not end the stream, as behind will no longer be an EndOfStream message code, but a standalone message itself?

mustafauzunn (dismissed):

Nana-EC (dismissed):
LG.

Nana-EC:
@jsync-swirlds how does it render when you follow the spotless suggestions.
If malformed then let's requst a github maintainers exception for this.
You can also rebase to get the api-tests passing

AlfredoG87 (dismissed):
LG

rockysingh (approved):
lgtm

mustafauzunn (approved):

PR #1935, opened 2025-12-04
test: Add CI Support for `k6`

### LINKED ISSUE #1878: Add CI Support for `k6`
### Story Form

As a block node engineer
I want to utilize k6 for average load smoke test, that need to run in CI
So that I can utilize a common testing framework on local and remote BNs that show cases expected load

### Technical Notes

* create a CI workflow to be able to run k6 tests
* run existing tests in CI
* update and cleanup existing tests
* use simulator currently with generated simple blocks so that we can run in CI

This issue will NOT solve:
- having sane tests
- having reproducible CI runs, we need a dedicated runner for that
- supply actual blocks that could be streamed, we need (most likely) a bucket with actual blocks, which then could be optionally streamed via k6 or simulator
- split tests so we do not have conflicting results
  - we've discovered that the grpc module of k6 is quite limited and we will not, as of today, be able to make custom metrics, since responses do not contain timings, like in the http module with the responses thereof
  - some tests need to be reworked and split, we cannot achieve true e2e latency measurement (e.g. time between publishing a block and a subscriber to receive it), but splitting these cases into separate tests will still yield valid results.

ata-nas:
## Reviewer Notes

* create a CI workflow to be able to run k6 tests
* run existing tests in CI
* update and cleanup existing tests
* use simulator currently with generated simple blocks so that we can run in CI

This issue will NOT solve:
- having sane tests
- having reproducible CI runs, we need a dedicated runner for that
- supply actual blocks that could be streamed, we need (most likely) a bucket with actual blocks, which then could be optionally streamed via k6 or simulator
- split tests so we do not have conflicting results
  - we've discovered that the grpc module of k6 is quite limited and we will not, as of today, be able to make custom metrics, since responses do not contain timings, like in the http module with the responses thereof
  - some tests need to be reworked and split, we cannot achieve true e2e latency measurement (e.g. time between publishing a block and a subscriber to receive it), but splitting these cases into separate tests will still yield valid results.

## Related Issue(s)

Resolves #1878


tools-and-tests/k6/test/bn-query-validation.js
```diff
@@ -1,63 +1,68 @@
 import { Client, StatusOK } from 'k6/net/grpc';
 import { check, sleep } from 'k6';
 import { SharedArray } from 'k6/data';
+import {GetBlockRequest, ServerStatusRequest} from "../lib/grpc.js";
```
Nana-EC: Q: why the rename on folder.
The gaol was to maintain a directory structure for each types of test. See https://github.com/hiero-ledger/hiero-block-node/tree/main/tools-and-tests/k6#test-types

Smoke tests are intentionally for simple sanity runs

tools-and-tests/k6/test/bn-server-status.js
```diff
@@ -1,37 +1,41 @@
 import { Client, StatusOK } from 'k6/net/grpc';
 import { check, sleep } from 'k6';
 import { SharedArray } from 'k6/data';
+import {ServerStatusRequest} from "../lib/grpc.js";
```
Nana-EC: Same question on types change.
Average-load also has a purpose.

tools-and-tests/k6/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "k6",
```
Nana-EC: We must apply convention here even if we don't publish.
Also see https://github.com/hiero-ledger/hiero-mirror-node/blob/main/rest/package.json for a template of what this file should look like.
```suggestion
  "name": "@hiero-ledger/block-node-k6",
```

tools-and-tests/k6/package-lock.json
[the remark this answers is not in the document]
Nana-EC: Whiles I understand the hesitation I'd actually be against this recommendation.
It's unconventional to do this for k6 tests and might make odd behaviour harder to catch and fix.
Even in Java projects most people write and run tests in JS/TS.
The current main team has sufficient JS/TS expertise to handle this.
Plus web3 industry is much more familiar with JS/TS so this could increase contributions.
It's in a silo and once it hits a certain maturity of test coverage it'll mostly be minor maintenance and copying existing logic.

tools-and-tests/k6/data.json
```diff
@@ -1,7 +1,7 @@
 {
     "configs": [{
         "blockNodeUrl": "localhost:40840",
-        "protobufPath": "./../block-node-protobuf-0.23.1",
+        "protobufPath": "./../k6-proto",
```
Nana-EC: Note, the goal here was to match the current release proto folder name and then we'd automate the version bump every release so it always matched latest

Nana-EC (commented):
Thanks for getting simulator pulled in.
I'm not sure about the current approach though so I've left some questions and suggestions.

This are blocking changes but I don't want to actually block due to scheduling

tools-and-tests/k6/package.json
ata-nas: If we are to keep these dev dependencies, we can create these conventions, otherwise if we delete these, the change is not needed. I see for and against opinions on deleting the package.json.

tools-and-tests/k6/run-k6-tests.sh
```diff
@@ -0,0 +1,103 @@
...
+    (
+    cd ../../
+    ./gradlew stopDockerContainer # stop all running containers
+    )
+}
+
+# todo add more tests here and also make sure to reset BN
+
+run_server_status_test() {
+    run_bn
```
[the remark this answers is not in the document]
ata-nas: So, as I mentioned in meetings, I will be utilizing a shell script for setups. here a couple of things to consider:

1. Yes, I agree that we best run a block node as a CI step and then test against it. In fact that was how I initially did it in one of the previous amends. However, we currently have an issue, as mentioned, that we cannot unfortunately create custom metrics, which we need in order to have correctness for the tests. As discussed, some tests make grpc calls for server status in order to setup the tests, but that skews the data as all requests get averaged together. That required a more elaborate setup. There is a possibility to somewhat work around this, by using tags, but we'd have to only looked at the tagged results. This means that certain calls are being placed a tag and at the end of the run, we have the total average (which again is always wrong) and then the averages for each tag. This could be an option for now.
2. Now, we also have to consider that tests leave side effects. We need to be careful about this, as this means that tests could interfere with each other. An example is publisher related tests. These will publish blocks, which will change the state of the block node. This means that we have to consider this when writing the tests. So it is not, in my opinion, ideal to always have a shared block node to run all tests agains.
3. Then, if we consider the above, we have to note that it means we will probably need some elaborate setup before running certain tests. If so, this means that we need to update the shell script. If we'd done everything in the CI workflow, this is way too cumbersome and requires a lot more to be merged. On another note, all our CI runs like this. Building the app calls upon a shell script. Running e2e tests invoke a gradle tasks, which are essentially a kotlin script. Is it going to be a shell or kotlin script, or any other type is all but the same. So in the same manner, when we need to do updates in the logic, we do it in the script, which is much less disruptive. Essentially, in my opinion, we should be able to setup any CI workflow once, really only with essentials, and hopefully never touch it again. Only changing more "local" things, like shell scripts, in order to be more productive and less disruptive.
4. I tried to make the tests be really easily executable locally, so that we can run these as needed with ease. That is why most of the setup, except basic tool provision, is in the shell script. The idea is to simply run the script and everything will be done for you. Prerequisite is that you have java, node, k6 etc. which once installed, you will have.

On another note, this PR does not have the goal to solve everything, but mainly to introduce the CI workflow. That being said, I could only run a single block node for now, not resetting, because the tests we currently have will not interfere with each other. And I do agree with that to an extent, but again, we need to think about side effects.

tools-and-tests/k6/package.json
Nana-EC: We have to do it as part of Hiero compliance and best practices. Similar to having the license information on files.
You should only need to see a few fields to be proper

tools-and-tests/k6/run-k6-tests.sh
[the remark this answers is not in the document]
Nana-EC: 🤔 I see.
The shell script is hiding some of the setup to avoid skewing the metrics.

tools-and-tests/k6/package.json
ata-nas: Noted about the Hiero compliance.

tools-and-tests/k6/package-lock.json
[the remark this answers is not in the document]
ata-nas: Let's reach agreement here. Do these remain or shall we remove these? I have no strong opinion.

tools-and-tests/k6/package.json
ata-nas: On the note of having the license headers in the files, should we add these as well to the new k6 tests? Also, is it possible to enable spotless to also run on the js/ts files? That would be great!

tools-and-tests/k6/lib/grpc.js
```diff
@@ -0,0 +1,39 @@
+import {Stream} from "k6/net/grpc";
```
Nana-EC: Please add the license headers. I think i missed them on the files also
```suggestion
// SPDX-License-Identifier: Apache-2.0
import {Stream} from "k6/net/grpc";
```

Nana-EC (approved):
No further blocking comments.
Just more context and request to add license.
We may need to plan out a bit more how the current setup applies for varied usage

tools-and-tests/k6/run-k6-tests.sh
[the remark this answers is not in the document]
ata-nas: Mind you, the shell script in it's current form is not what I envision as a final thing. Simply, I do the minimum amount of changes needed for us to have a CI workflow and also be able to locally run the tests. With future updates, I do hope to drastically clean up the shell script and rely more on the tests themselves to possibly do their own setup.

tools-and-tests/k6/lib/grpc.js
ata-nas: Agreed, what about [this comment](https://github.com/hiero-ledger/hiero-block-node/pull/1935#discussion_r2610245191)?

JeffreyDallas (approved):

rbarker-dev (approved):

PR #1952, opened 2025-12-10
fix(backfill): Backfill Logic refactor to reduce serverStatus requests to the absolute necessary and on-demand edge case fix

### LINKED ISSUE #1942: bug(backfill): When BN is new (empty), On-Demand backfill does not respect startBlock config.
### Description

When the BN receives a `TOO_FAR_BEHIND` message will attempt to backfill from: `LatestAvailableBlock` up to `LatestBlockKnownToNetwork`.

However for an empty block, this means from block 0 regardless of what was configured on config property `backfill.startBlock`

### Steps to Reproduce

1. Create a BN with blocks from 1000 to 2000.
2. Create a new BN with empty storage
3. Configure BN2 to backfill from BN1
4. Disable greedy backfill and do not stream blocks to BN2.
5. Manually send `TOO_FAR_BEHIND` message 
6. Look at the logs, it will start from 0 and not from backfill.startBlock configured property as it should in this edge case

### Additional context

On-Demand Backfill should choose the largest of the 2, something like this:

```
        // we should create  new Gap and a new task to backfill it
        long lastPersistedBlock =
                context.historicalBlockProvider().availableBlocks().max();
        // if lastPersistedBlock is less than backfill start block, we start from backfill start block
        long startBackfillFrom = Math.max(lastPersistedBlock, backfillConfiguration.startBlock());
        long newestBlockKnown = notification.blockNumber();
        LongRange gap = new LongRange(startBackfillFrom, newestBlockKnown);
```

### LINKED ISSUE #1948: bug(backfill): Misconfigured Backfill and serverStatus partial info is Generating too many logs.
### Description

In the last 24 hours backfill incurred in 7.7 Million logs that are only produced with each serverStatus request.

This is creating too much overload in servers.

Image

### Steps to Reproduce

Set up 2 BNs.

1. Fill block node source with blocks from 10 to 100 and from 100,000 to 110,000.
2. ServerStatus should read earliest_available_block = 10, and latestAvailableBlock=110,000
3. Even if the above is true, is really misleading since most of the blocks are not present.
4. Set-up Block node 2, to backfill from the first.
5. let greedy or autonomous backfill kick-in.
6. Once the gap is identified it proceeds to chunkify and then to attempt to fetchBlocks to that gap from all the BNs.

### Additional context

**Recommendation:** Refactor the method `backfillGap` to avoid hitting multiple times the block-node sources.

AlfredoG87:
## Reviewer Notes

- When choosing startBlock for On-Demand backfill, in case the BN is new and no `latest_available_block` exists, then take the **startBlock of backfill configuration** instead of trying to backfill from `0`.
- Refactored the BackfillPlugin and BackfillGrpcClient to greatly reduce the `serverStatus` requests footprint by a 99% by flipping logic, instead of the naive approach of `Chunkify` a gap first and then try to fetch the blocks batch by batch. Now we first calculate the available ranges per node for a given gap, and out of them we calculate the next batch and select the ideal block based on priority, and if a source BN does return NOT_AVAILABLE  (misleading) we mark it as Unavailable for the duration of this backfill run.
- Added a new test case with a Lying BN that says it has a range but it does not.
- More UTs to cover edge cases and verify expected behaviour of the helper methods and BackfillGrpcClient public methods.

## Related Issue(s)
Fixes #1942 
Fixes #1948

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -415,10 +415,11 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
      * @throws ParseException if there is an error parsing the block header
      */
     private void backfillGap(LongRange gap, BackfillType backfillType) throws InterruptedException, ParseException {
-        BackfillGrpcClient backfillGrpcClient = backfillType.equals(BackfillType.AUTONOMOUS)
-                ? backfillGrpcClientAutonomous
-                : backfillGrpcClientOnDemand;
-
+        BackfillGrpcClient backfillGrpcClient =
+                switch (backfillType) {
+                    case AUTONOMOUS -> backfillGrpcClientAutonomous;
+                    case ON_DEMAND -> backfillGrpcClientOnDemand;
+                };
         Map<BackfillSourceConfig, List<LongRange>> availability = planAvailabilityForGap(backfillGrpcClient, gap);
         long currentBlock = gap.start();
         long batchSize = backfillConfiguration.fetchBatchSize();


block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
```diff
@@ -30,31 +30,32 @@ public TestBlockNodeServer(int port, HistoricalBlockFacility historicalBlockFaci
...
-                                            .build())
-                                    .build());
-                            replies.onNext(SubscribeStreamResponse.newBuilder()
-                                    .endOfBlock(
-                                            BlockEnd.newBuilder().blockNumber(i).build())
+                                    .status(SubscribeStreamResponse.Code.NOT_AVAILABLE)
                                     .build());
+                            replies.onComplete();
+                            return;
                         }
```
jsync-swirlds: Unconditional return is no different, in this case, than an `else` block.
The latter, generally, is easier to reason about and understand for someone new to the codebase.

AlfredoG87: ✅  Done!

jsync-swirlds (approved):
Looks Good, just a few potential improvements.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -412,28 +415,38 @@ private void processDetectedGaps() throws ParseException, InterruptedException {
      * @throws ParseException if there is an error parsing the block header
      */
     private void backfillGap(LongRange gap, BackfillType backfillType) throws InterruptedException, ParseException {
-        // Reset client status to retry previously unavailable nodes
-        BackfillGrpcClient backfillGrpcClient;
+        BackfillGrpcClient backfillGrpcClient = backfillType.equals(BackfillType.AUTONOMOUS)
+                ? backfillGrpcClientAutonomous
+                : backfillGrpcClientOnDemand;
```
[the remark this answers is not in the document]
AlfredoG87: I like it, an inline switch expression 💯 

✅  Done!

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillGrpcClientTest.java
```diff
@@ -0,0 +1,143 @@
...
+                selection.get().nodeConfig().equals(nodeA)
+                        || selection.get().nodeConfig().equals(nodeB),
+                "Selection should be one of the tied nodes");
+    }
+
+    @Test
+    @DisplayName("Retries once, increments retry counter, then marks node unavailable on repeated failure")
+    void retriesAndMarksUnavailableOnFailure() throws Exception {
+        final BackfillSourceConfig node = node("localhost", 1, 1);
+        final Counter retryCounter = Mockito.mock(Counter.class);
```
ata-nas: Instead of using a mock, we can instantiate an actual counter, I've personally done so for other tests. See `PublisherHandlerTests` and the usage of `com.swirlds.metrics.impl.DefaultCounter`. Although there we should be utilizing the newer approach @jsync-swirlds added, i. e. TestUtils.createMetrics() which then can be used to resolve metrics. But for your approach, simply instantiating a counter and other metrics you need will be enough.

ata-nas (approved):
Good improvements @AlfredoG87! Optional, non-blocking comments.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
```diff
@@ -30,31 +30,32 @@ public TestBlockNodeServer(int port, HistoricalBlockFacility historicalBlockFaci
         final Builder pbjRoutingBuilder = PbjRouting.builder()
                 .service(new TrivialBlockNodeServerInterface(historicalBlockFacility))
                 .service((BlockStreamSubscribeServiceInterface) (request, replies) -> {
```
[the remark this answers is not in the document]
Nana-EC: comments too, I had to infer vs being told what it does

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillGrpcClient.java
```diff
@@ -280,6 +181,170 @@ public void resetStatus() {
...
+            }
+
+            List<LongRange> ranges = resolveAvailableRanges(currentNodeClient);
+
+            List<LongRange> intersections = new ArrayList<>();
+            for (LongRange range : ranges) {
+                long intersectionStart = Math.max(targetRange.start(), range.start());
+                long intersectionEnd = Math.min(targetRange.end(), range.end());
+                if (intersectionStart <= intersectionEnd) {
+                    intersections.add(new LongRange(intersectionStart, intersectionEnd));
```
Nana-EC: Q: Do we need to check for continguous ranges here?

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillGrpcClient.java
```diff
@@ -280,6 +181,170 @@ public void resetStatus() {
...
+
+        if (earliestAvailableStart == Long.MAX_VALUE) {
+            return OptionalLong.empty();
+        }
+
+        return OptionalLong.of(earliestAvailableStart);
+    }
+
+    /** Build the list of nodes that can serve the earliest available start. */
+    private List<NodeSelection> candidatesForEarliest(
```
Nana-EC: In the future we could optimize this by doing it as part of getting availability.
Right now it's not as costly but something to note as it would simplify the logic

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillGrpcClient.java
```diff
@@ -264,7 +165,7 @@ public List<BlockUnparsed> fetchBlocks(LongRange blockRange) {
      * @param node the BackfillSourceConfig to get the client for
      * @return a BlockNodeClient for the specified node
      */
-    private BlockNodeClient getNodeClient(BackfillSourceConfig node) {
+    protected BlockNodeClient getNodeClient(BackfillSourceConfig node) {
```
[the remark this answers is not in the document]
AlfredoG87: Is needed for UT testing purposes.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillGrpcClient.java
```diff
@@ -280,6 +181,170 @@ public void resetStatus() {
...
+    /** Choose the lowest-priority candidate (tie-breaking randomly) from the supplied list. */
+    private Optional<NodeSelection> chooseBestCandidate(@NonNull List<NodeSelection> candidates) {
+        if (candidates.isEmpty()) {
+            return Optional.empty();
+        }
+
+        int bestPriority = candidates.stream()
+                .mapToInt(selection -> selection.nodeConfig().priority())
+                .min()
+                .orElse(Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
AlfredoG87: added to the followup issues

Issue #1442, opened 2025-07-30
feat(charts): Add Mount Volume for Plugins Source


AlfredoG87:
### Story Form
As a Block-Node Operator
I want to be able to select only the plugins that I am interested in deploying
So that I can configure my BN as I see fit

### Technical Notes
- Currently the BN Image brings **ALL** core plugins, in the future we want to release a bare-bone image that contains NO plugins.
- Then adding them to a specific local mount where the BN container would pick them up and load them at runtime when starting up.

We intend to start by defining the necessary scripts and Helm chart changes to enable downloading a configurable list of plugins from a "companion" OCI image and/or the Maven Central repository.  This download _may_ include SHA-512 hash verification for the Maven source.
This effort includes defining the local hostPath volume for the block node container, adding Helm chart support to download a configured list of plugins and copy them to that local hostPath volume folder.
As part of this story we intend to define example Helm values files for Base/Minimal, Full History, S3 Archive, Partial History, and Testing (i.e. all known plugins) deployments.

PR #2011, opened 2026-01-07
feat: Modify the Block Node build for dynamic plugins

### LINKED ISSUE #1657: Modify the Block Node build for dynamic plugins
### Story Form

As a Release Engineer
I want to build a minimal OCI as part of the block node build
So that I can deploy different block node configurations dynamically

### Technical Notes

Adjust the block node build to run bare-bone OCI image build processes.
1. Build with only the minimum required plugins (perhaps none) for use by Helm chart deployments that add plugins dynamically based on chart configuration.
3. Modify the build to publish all current plugin modules to Maven Central as individual libraries so that they can be downloaded individually during Helm chart deployments.

mustafauzunn:
## Reviewer Notes

1. Adjust the block node build to run a bare-bone OCI image
2. Introduce a mechanism to add selected plugins to a bare-bones image

## Related Issue(s)
Fixes #1657 #1442


block-node/app/build.gradle.kts
```diff
@@ -49,20 +70,12 @@ mainModuleInfo {
...
-    runtimeOnly("org.hiero.block.node.archive.s3cloud")
     runtimeOnly("org.hiero.block.node.messaging")
     runtimeOnly("org.hiero.block.node.health")
-    runtimeOnly("org.hiero.block.node.stream.publisher")
-    runtimeOnly("org.hiero.block.node.stream.subscriber")
-    runtimeOnly("org.hiero.block.node.verification")
-    runtimeOnly("org.hiero.block.node.blocks.files.historic")
-    runtimeOnly("org.hiero.block.node.blocks.files.recent")
     runtimeOnly("org.hiero.block.node.access.service")
     runtimeOnly("org.hiero.block.node.server.status")
```
ata-nas: Are the remaining plugins in the `mainModuleInfo` considered "core"? I would think that server status and access service are probably optional? Really the only "core" plugins are messaging and health, as the app itself is an `HealthFacility` and messaging must be loaded in order to startup the app. Personally, I think if a plugin's absence does not in any way hinder the ability to startup the app, it should probably be considered "optional". That, however, is indeed only technically speaking. I do see an argument saying that health and access could be considered "core" design wise. Not sure about this.

jsync-swirlds: My opinion:
No plugins should be considered "core".  All plugins (including messaging, health, status, etc...) should be loaded at runtime.
We made these plugins so we could replace them if desired, but continuing to include them in the base (i.e. "barebones") image prevents that replacement.
The base image _really_ shouldn't have anything other than the `app` module and non-pluggable modules (e.g. helidon, config, metrics).

block-node/app/docker/Dockerfile
```diff
@@ -144,6 +144,9 @@ ENV BN_WORKDIR=/opt/hiero/block-node
 
 # Install Java
 COPY --from=java-builder ${JAVA_HOME}/ ${JAVA_HOME}/
+# COPY bash script to container
+COPY download-and-load-plugins.sh ${BN_WORKDIR}/download-and-load-plugins.sh
```
jsync-swirlds: I don't think this belongs in the application container.
Downloading and loading plugins should happen exclusively outside of the main application container.

block-node/app/docker/download-and-load-plugins.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Change 0.25.0-rc1 to ${VERSION} when versions align
```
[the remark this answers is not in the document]
jsync-swirlds: Rather than hardcode, perhaps accept the version as a script parameter and pass that in?

charts/block-node-server/values-overrides/minimal.yaml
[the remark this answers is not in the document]
jsync-swirlds: Correct, `Minimal` is about resources, not plugins.
All current plugins are needed for minimal as it's a configuration for development and testing.

charts/block-node-server/values-overrides/rfh.yaml
```diff
@@ -0,0 +1,3 @@
+blockNode:
+    config:
+        PLUGINS: "stream-publisher,stream-subscriber,verification,blocks-file-historic,blocks-file-recent,backfill,s3-archive"
```
jsync-swirlds: Missing: `messaging`, `status`, `health`, `block-access-service`.
Remove: `blocks-file-historic`, `stream-subscriber`
Open question: Do we require `blocks-file-recent`, or can we remove that as well?

block-node/app/docker/download-and-load-plugins.sh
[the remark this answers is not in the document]
AlfredoG87: When we do a release, we do the publish of the OCI images and plugins in parallel, after this change, we need to publish OCI image only if the plugins were sucessfully published to maven, so subsequent tasks don't fail, that will require changes in https://github.com/hiero-ledger/hiero-block-node/blob/main/.github/workflows/release-push-image.yaml to make the release consistent and that chart test won't fail:

image

block-node/app/build.gradle.kts
```diff
@@ -11,12 +11,33 @@ description = "Hiero Block Node Server App"
...
     environment("FILES_RECENT_UNVERIFIED_ROOT_PATH", "${serverDataDir}/files-unverified")
+    mainModuleInfo {
+        runtimeOnly("org.hiero.block.node.archive.s3cloud")
+        runtimeOnly("org.hiero.block.node.stream.publisher")
+        runtimeOnly("org.hiero.block.node.stream.subscriber")
+        runtimeOnly("org.hiero.block.node.verification")
+        runtimeOnly("org.hiero.block.node.blocks.files.historic")
+        runtimeOnly("org.hiero.block.node.blocks.files.recent")
+        runtimeOnly("org.hiero.block.node.backfill")
+    }
```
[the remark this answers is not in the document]
mustafauzunn: We need this for `gradlew run` for local usage with all of the plugins

charts/block-node-server/values-overrides/lfh.yaml
```diff
@@ -0,0 +1,3 @@
+blockNode:
+    config:
+        PLUGINS: "stream-publisher,stream-subscriber,verification,blocks-file-historic,blocks-file-recent,backfill"
```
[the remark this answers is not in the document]
mustafauzunn: Refactored after the change in the approach

block-node/app/docker/Dockerfile
mustafauzunn: Reverted the change after the change in the approach

charts/block-node-server/values-overrides/rfh.yaml
```diff
@@ -0,0 +1,7 @@
+plugins:
+  volumeName: "plugins"
+  size: 2Gi
+  enabled: true
+  baseUrl: ""
+  version: ""
+  names: "facility-messaging,block-access-service,health,server-status,stream-publisher,stream-subscriber,verification,blocks-file-historic,blocks-file-recent,backfill,s3-archive"
```
ata-nas: After s3 rework, local history should be redundant here. The s3 plugin can briefly store blocks until it is certain the full batch is on the cloud.

jsync-swirlds: A couple additional specific items:
`blocks-file-historic` should never be included in an RFH block node.
`stream-subscriber` _generally_ should not be included in an RFH block node.
`blocks-file-recent` might be needed at the moment (not sure), but at the latest, once we've fully reworked the S3 Archive plugin to use partial upload instead of local storage, we should be able to completely remove that as well.

ata-nas: I agree with your view that we should probably remove historic, but we have to be very careful because the default retention policy of recent is not hight enough by default to even gather a single s3 default batch! Moreover if we do increase the retention policy size, we need to make sure that a failure in s3 will be OK and we will still have the failed batch. Lastly, gaps should be backfilled in order to have remote history in conjunction with the retention policy.

And again, this is simply our situation until s3 rework.

charts/block-node-server/README.md
```diff
@@ -121,6 +121,119 @@ or passed at the command line:
...
+
+Minimal functional node for development and testing.
+* `facility-messaging`
+* `block-access-service`
+* `health`
+* `server-status`
+* `stream-publisher`
+* `stream-subscriber`
+* `verification`
+* `blocks-file-historic`
```
jsync-swirlds: No reason to have historic for a minimal development/testing install.
files-recent is plenty in that case.

block-node/app/build.gradle.kts
```diff
@@ -129,7 +176,7 @@ tasks.register<Exec>("startDockerContainer") {
         "Starts the docker production container of the Block Node Server for the current version"
     group = "docker"
 
-    dependsOn(createDockerImage)
+    dependsOn(createDockerImageCI)
```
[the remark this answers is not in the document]
AlfredoG87: This section has been fully refactored. The current startDockerContainer task depends on createDockerImage and prepareDockerPlugins, which is the correct dependency chain

We no longer have 2 different createDockerImage flows, both were adding all the plugins 😅 
Now the only one that exists is not adding any dependency of plugins.

block-node/app/build.gradle.kts
```diff
@@ -11,12 +11,60 @@ description = "Hiero Block Node Server App"
...
+
+distributions {
+    main {
+        contents {
+            val pluginsDir = layout.buildDirectory.dir("tmp/plugins")
+
+            from(pluginsDir) { into("plugins") }
+        }
+    }
+}
```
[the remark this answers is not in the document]
AlfredoG87: Added descriptive comments throughout the build script. 

✅

AlfredoG87 (approved):
Unblocking.

AlfredoG87:
Closing in favor of a new PR — the majority of this work has been reimplemented.

PR #2012, opened 2026-01-07
test: Refactor Existing `k6` Tests to Measure Accurately

### LINKED ISSUE #1957: Refactor Existing `k6` Tests to Measure Accurately
### Story Form

As a Block Node Engineer
I want to have sane k6 tests, which accurately reflect important for me metrics
So that I can know if performance suddenly changes

### Technical Notes

**DEPENDS ON**: #1878 

- currently, tests that are written for k6 include multiple method calls
  - this by itself is not problematic
- we have found out, however, that the grpc module of k6 is very limited, and unfortunately, as of today, it appears we will not be able to create custom metrics
  - custom metrics are needed so that we know for sure what we are measuring and if multiple calls of different methods are made, we can differentiate them and only look at what interest us
- in order to achieve an e2e test for example, we will need to make a test for publisher only and then for subscriber only, to work around the limitations
- the simulator could be utilized to feed data to the BN and then tests could only focus on a single method call for the tests to be accurate
- the main purpose of the k6 tests is to give us averages and not exact performance, we have to be able to quickly detect severe performance degradation based on current changes and know that our current changes have caused it

ata-nas:
* Refactored existing tests to give us more accurate average readings
* Improved test logs

## Related Issue(s)

Closes #1957

### DIFF
--- a/.github/workflows/k6-tests.yml
+++ b/.github/workflows/k6-tests.yml
@@ -70,6 +70,12 @@ jobs:
           trap "popd" EXIT
           npm run test
 
+      - name: Display Test Logs
+        run: |
+          pushd ${K6_ROOT} || exit 1
+          trap "popd" EXIT
+          npm run print-logs
+
       - name: Stop Containers
         if: ${{ always() }}
         run: ${GRADLE_EXEC} stopDockerContainer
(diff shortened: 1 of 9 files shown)


AlfredoG87 (approved):
LGTM, it took me some time, since the k6 version that I had installed already was too old and some things were just failing to compile 😅 

Finally did it and experienced the serverStatus failures that you mentioned and try out some ideas, with no luck :(

PR #2090, opened 2026-01-29
feat: Files Historic Plugin Zip Filename Style

### LINKED ISSUE #1027: Files Historic Plugin Zip Filename Style
### Story Form

As a Block Node Developer/Operator/User
I want to choose the best zip file name style
So that I can have the best possible structure/file naming convention

### Technical Notes

- based on [this comment thread](https://github.com/hiero-ledger/hiero-block-node/pull/1019#discussion_r2044575776) let's choose what the final version of the zip filename convention will be, with or without an `s` at the end of the digits
- some arguments for the `s` are that it would be more user friendly and will be easier to do manual searches for
- some arguments against the `s` are that this convention should not be human friendly as it should not stimulate manual intervention
- we are currently appending the `s`, so if that shall remain the chosen convention, we should do nothing and simply close this issue

ivannov:
## Reviewer Notes

This PR removes the unnecessary 's' suffix from block batch zip file names, simplifying the naming convention. The change updates:                   
  - Zip file name generation in BlockPath.java to use .zip instead of s.zip (e.g., 50000.zip instead of 50000s.zip)                                    
  - String parsing logic in ZipBlockArchive.java to use lastIndexOf('.') instead of indexOf('s') for more robust file extension handling               
  - All test cases and code comments to reflect the new naming convention      

## Related Issue(s)
Closes #1027

### DIFF
--- a/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
+++ b/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
@@ -145,6 +145,7 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
             numberOfBlocksPerZipFile = intPowerOfTen(config.powersOfTenPerZipFileContents());
             // create the executor service for moving blocks to zip files
             zipMoveExecutorService = context.threadPoolManager().createSingleThreadExecutor("FilesHistoricZipMove");
+            renameOldFormatArchives(config.rootPath());
             zipBlockArchive = new ZipBlockArchive(context, config);
             // get the first and last block numbers from the zipBlockArchive
             final long firstZippedBlock = zipBlockArchive.minStoredBlockNumber();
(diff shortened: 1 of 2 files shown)


tools-and-tests/simulator/docker/update-env.sh
```diff
@@ -13,6 +13,9 @@ PROMETHEUS_ENDPOINT_ENABLED=true
 # For publisher service
 PUBLISHER_BLOCK_STREAM_SIMULATOR_MODE=PUBLISHER_CLIENT
 PUBLISHER_PROMETHEUS_ENDPOINT_PORT_NUMBER=16008
+GENERATOR_START_BLOCK_NUMBER=0
```
Nana-EC: Q: Was this accidentally added?

ivannov: No, we noticed with @ata-nas that we had some missing env vars (introduced with previous PR). So we fixed that along the way

ivannov:
> Seems reasonable. Please check for reference for `s.zip` in doc files as they will be outdated with this update.

Found some references, updated them

ata-nas (approved):
LG, one minor detail.

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockArchive.java
```diff
@@ -151,7 +151,7 @@ long minStoredBlockNumber() {
                             .filter(path -> path.getFileName().toString().endsWith(".zip"))
                             .min(Comparator.comparingLong(filePath -> {
                                 String fileName = filePath.getFileName().toString();
-                                return Long.parseLong(fileName.substring(0, fileName.indexOf('s')));
+                                return Long.parseLong(fileName.substring(0, fileName.lastIndexOf('.')));
```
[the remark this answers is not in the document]
ata-nas: Good point. I missed that.

tools-and-tests/simulator/docker/update-env.sh
Nana-EC: Okay, unrelated to PR but desired

AlfredoG87:
@Nana-EC this should be unblocked now, there is renaming check and a test that verifies it works.

ivannov:
> > Thanks for addressing backwards compatibility. Also i'm assuming we already have coverage for an empty store and that added compatibility won't fail on that.
> 
> Not exactly, in fact it's entirely possible we _will_ fail on empty data because the added code (specifically the file visitor) checks inputs for null when it doesn't need to and when null inputs are entirely valid. Most of the `requireNonNull` calls in the FileVisitor implemented to do the renames should probably not be present.

@jsync-swirlds most of the tests (if not all besides the one that I added) are run on an empty store. None of them failed. The `requireNonNull` stuff I must confess I copied from the other visitor in the same class. It is run while walking the staging directories right before the new visitor is run. And it seems to work.

I would be happy to create a separate task and dig deeper into possible errors if I get a hint of a setup that this "migration" might fail and is not covered by current tests. Thanks!

PR #2092, opened 2026-01-29
fix(backfill): Fix Backfill HTTP/2 Connection Fails on Virtual Threads in Container Environments

### LINKED ISSUE #2091: Bug(backfill): HTTP/2 Connection Fails on Virtual Threads in Container Environments using K8S
### Description

Block Node backfill fails to establish HTTP/2 connections to peer Block Nodes when running in Kubernetes containers. The `PbjGrpcClient` constructor blocks indefinitely during HTTP/2 connection establishment.

And after timeout of 20 seconds (hardcoded by helidon) returns error: 

```
Failed to initialize gRPC client: Filed to send HTTP/2 preface within 20 seconds, this connection is broken                    
  java.lang.IllegalStateException: Filed to send HTTP/2 preface within 20 seconds, this connection is broken                     
      at io.helidon.webclient.http2.Http2ClientConnection.start(Http2ClientConnection.java:349)                      
```

### Steps to Reproduce

Use solo to deploy the following network:
```
# SPDX-License-Identifier: Apache-2.0
# Minimal backfill test topology: 1 CN, 2 BNs
#
# For testing backfill functionality with minimal resources.
# CN1 streams to both BNs, BN1 configured to backfill from BN2.
#
# CN1 → BN1
# BN1 ← BN2 (backfill)

name: minimal-backfill
description: "1 CN, 2 BNs for backfill testing"

block_nodes:
  block-node-1:
    address: block-node-1
    port: 40840
  block-node-2:
    address: block-node-2
    port: 40840
    peers: [ block-node-1 ]
    greedy: true

consensus_nodes:
  node1:
    block_nodes: [block-node-1]

# Skip MN/Relay/Explorer for faster deployment
mirror_nodes: {}
relay_nodes: {}
explorer_nodes: {}
```

The only way of getting blocks for BN2 is from BN1 and in greedy mode.

However its unable to connect to the host and hence not able to start backfilling.

## Severity

**Critical** - Blocks BN-to-BN backfill functionality entirely in containerized deployments in solo and potentially other environments.

## Environment

- Kubernetes (Kind cluster tested, likely affects all K8s deployments)
- Helidon 4.3.3
- PBJ 0.13.3
- Java 21 with virtual threads

## Symptoms

- Backfill scheduler hangs during gap detection with `unable to connect` issue.
- No blocks are fetched from peer Block Nodes

## Root Cause

**Helidon WebClient HTTP/2 implementation has issues when called from virtual threads** in container environments.

The `PbjGrpcClient` constructor synchronously establishes an HTTP/2 connection. When this code runs on a **virtual thread** (as created by `createVirtualThreadScheduledExecutor`), the HTTP/2 connection establishment blocks indefinitely.

When the same code runs on a **platform thread** (as created by `newSingleThreadScheduledExecutor`), it works correctly.

### Evidence

1. **CN (Consensus Node) works**: CN uses `Executors.newSingleThreadScheduledExecutor()` (platform thread)
2. **curl works**: Both HTTP/2 upgrade and prior-knowledge modes work perfectly from inside the container
3. **Local port-forward works**: The same BN code works when connecting to a port-forwarded service

AlfredoG87:
### Primary Fix: Use Platform Threads for All Backfill Executors

**File:** `block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java`

All three executors must use platform threads:
- `autonomousExecutor` - runs gap detection
- `historicalExecutor` - runs historical backfill (makes gRPC calls)
- `liveTailExecutor` - runs live tail backfill (makes gRPC calls)

### Secondary Fix: Allow Client Reconnection

**File:** `block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java`

Remove unreachable clients from cache to allow reconnection attempts, this helps clients recover faster.

### Minor Improvements

**File:** `block-node/backfill/src/main/java/org/hiero/block/node/backfill/client/BlockNodeClient.java`

- Change `priorKnowledge` default to `false` for better compatibility
- Add exception logging in `initializeClient()`

## Testing

1. Deploy topology with 2 Block Nodes where one backfills from another and only one CN is present and streams to a single bn1.
2. Verify blocks are successfully fetched and persisted on BN1
3. Verify blocks are backfilled successfully as expected on BN2

## Related

- Consensus Node uses platform threads: `BlockNodeConnectionManager.java:451`
- PBJ gRPC client establishes connection in constructor: `PbjGrpcClient.java`

## Related Issue(s)

Fixes #2091

### DIFF
--- a/block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
@@ -118,4 +118,18 @@ final class DefaultThreadPoolManager implements ThreadPoolManager {
             return Executors.newScheduledThreadPool(corePoolSize, factory);
         }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @NonNull
+    @Override
+    public ScheduledExecutorService createSingleThreadScheduledExecutor(
+            @NonNull final String threadName, @NonNull final Thread.UncaughtExceptionHandler uncaughtExceptionHandler) {
+        Preconditions.requireNotBlank(threadName);
+        final OfPlatform factoryBuilder = Thread.ofPlatform();
+        factoryBuilder.name(threadName);
+        factoryBuilder.uncaughtExceptionHandler(uncaughtExceptionHandler);
+        return Executors.newSingleThreadScheduledExecutor(factoryBuilder.factory());
+    }
 }
--- a/block-node/app/src/test/java/org/hiero/block/node/app/DefaultThreadPoolManagerTest.java
+++ b/block-node/app/src/test/java/org/hiero/block/node/app/DefaultThreadPoolManagerTest.java
@@ -5,6 +5,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import java.lang.Thread.UncaughtExceptionHandler;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import org.assertj.core.api.InstanceOfAssertFactories;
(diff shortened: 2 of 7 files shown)


block-node/app/src/main/java/org/hiero/block/node/app/DefaultThreadPoolManager.java
```diff
@@ -118,4 +118,18 @@ public ScheduledExecutorService createVirtualThreadScheduledExecutor(
...
+    @NonNull
+    @Override
+    public ScheduledExecutorService createSingleThreadScheduledExecutor(
+            @NonNull final String threadName, @NonNull final Thread.UncaughtExceptionHandler uncaughtExceptionHandler) {
+        Preconditions.requireNotBlank(threadName);
+        final OfPlatform factoryBuilder = Thread.ofPlatform();
+        factoryBuilder.name(threadName);
+        factoryBuilder.uncaughtExceptionHandler(uncaughtExceptionHandler);
+        return Executors.newSingleThreadScheduledExecutor(factoryBuilder.factory());
+    }
```
jsync-swirlds: I believe @ata-nas has much the same code already implemented in a different PR; please coordinate this change so we aren't accomplishing the same task in two different ways.

jsync-swirlds:
I would like us to verify this is still an issue with Java 25 before merging this change.
It's entirely possible that the improvements in Java 25 to Virtual Thread handling will make this more expensive (performance-wise) change unnecessary.

jsync-swirlds:
> Initial test indicates that JAVA 25 does indeed solves the issue too. 💯
> 
> I know using a platform thread is more expensive, but for backfill purposes I think it might make sense still.
> 
> How do we want to proceed here?

We can make this change _only_ on the release branch, but upgrade to Java 25 on `main`; or we can make this change on both branches, but include a revert of this change in the Java 25 PR on `main`.

AlfredoG87:
> > Initial test indicates that JAVA 25 does indeed solves the issue too. 💯
> > I know using a platform thread is more expensive, but for backfill purposes I think it might make sense still.
> > How do we want to proceed here?
> 
> We can make this change _only_ on the release branch, but upgrade to Java 25 on `main`; or we can make this change on both branches, but include a revert of this change in the Java 25 PR on `main`.

I've already rebased and forced pushed my changes again so we can target `release/26` branch only, and also, working on a Java 25 upgrade branch, but still ironing out some small nits: 
https://github.com/hiero-ledger/hiero-block-node/pull/2094

Nana-EC (approved):
Lg,
Good catch on the root cause

PR #2103, opened 2026-01-31
fix: 2095 block zero state

jasperpotts:
Fixes 2095

This adds the saved state at block zero as resources and adds code and tools for reading and validating that state. This will be used as part of the wrapping process to add the state of the network at beginning of block zero into the wrapped block zero.

### DIFF
--- a/tools-and-tests/tools/README.md
+++ b/tools-and-tests/tools/README.md
@@ -104,7 +104,7 @@ subcommands
 ├── networkCapacity           # Network capacity testing (gRPC/HTTP2)
 │
 └── states                   # Tools for saved state directories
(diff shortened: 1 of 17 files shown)


AlfredoG87 (dismissed):
Did not went trought with everything, but unblocking  since this is not production code but tooling.
Althought the CI is blocking it for Style Check 😅 

Let me know when you need re-approval.

AlfredoG87 (approved):
Unblocking

Nana-EC (approved):
Approval based on offline assessment.
Sits outside of main product code and is still seeing refinement so we should be good.

AlfredoG87 (approved):

PR #2107, opened 2026-01-31
fix(subscriber): Fix issue with streaming large blocks by chunking 

### LINKED ISSUE #2102: Bug(subscriber): Stream Subscriber should chunk large historical blocks for gRPC streaming
### Description

The current block streaming implementation sends entire blocks as single gRPC messages. When a block exceeds PBJ's `maxMessageSizeBytes` limit (**4MB**), the block becomes unstreamable and the block stream to downstream clients halts.

This is a **PBJ limitation**, not a gRPC protocol limitation. gRPC/HTTP2 automatically chunks messages into frames on the wire, but PBJ must allocate a buffer for the entire message.

**Live blocks are not affected** - they arrive in batches and are already chunked.

**Discovery:** This issue was found when Backfill operations failed to receive large historical blocks.

**We currently are unable to backfill large blocks, we can remediate this by increasing some configs on the backfill client, however the ultimate fix most be chunking from the source.**

### Steps to Reproduce

1. Start a block node with historical blocks stored
2. Ensure at least one block exceeds PBJ limit (4MB)
3. Request historical blocks via `subscribeBlockStream`
4. Client fails to parse the large block

## Proposed Solution

The Stream Subscriber should chunk historical blocks when sending, similar to how Consensus Nodes handle this:

**CN approach (for reference):**
- Soft limit: aim for batches of ~2MB
- Hard limit: a single item cannot exceed 4MB
- If an item exceeds the soft limit but is under the hard limit, it ships by itself

**BN implementation:**
1. Split large blocks into smaller chunks when streaming to subscribers
2. Each chunk sent as a separate `BlockItemSetUnparsed` message
3. Final chunk followed by `sendEndOfBlock()` as today

## Considerations

- Chunk size should be configurable but default to a safe value well under 4MB limit (e.g., 1MB)
- Must not break existing subscribers - multiple `BLOCK_ITEMS` messages per block is already valid

## Affected Components

- `stream-subscriber/BlockStreamSubscriberSession.java` - historical block streaming (sender side)
- `backfill/BlockNodeClient.java` - receives streamed blocks (where issue was discovered)
- `SubscriberConfig` - add chunk size configuration

AlfredoG87:
## Reviewer Notes

- Add maxChunkSizeBytes config (default 1MB) for historical block streaming
- Implement size-based chunking in BlockStreamSubscriberSession to stay within PBJ 2-4MB buffer limit

## Related Issue(s)

Fixes #2102

### DIFF
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
@@ -23,12 +23,12 @@ import org.hiero.block.node.base.Loggable;
 // spotless:off
 @ConfigData("server.http2")
 public record WebServerHttp2Config(
-        @Loggable @ConfigProperty(defaultValue = "2000") int flowControlTimeout,
-        @Loggable @ConfigProperty(defaultValue = "16_777_216") int initialWindowSize,
-        @Loggable @ConfigProperty(defaultValue = "50") long maxConcurrentStreams,
+        @Loggable @ConfigProperty(defaultValue = "1000") int flowControlTimeout,
+        @Loggable @ConfigProperty(defaultValue = "8_388_608") int initialWindowSize,
+        @Loggable @ConfigProperty(defaultValue = "8") long maxConcurrentStreams,
         @Loggable @ConfigProperty(defaultValue = "10") int maxEmptyFrames,
         @Loggable @ConfigProperty(defaultValue = "8_388_608") int maxFrameSize,
         @Loggable @ConfigProperty(defaultValue = "8192") long maxHeaderListSize,
-        @Loggable @ConfigProperty(defaultValue = "200") int maxRapidResets,
+        @Loggable @ConfigProperty(defaultValue = "50") int maxRapidResets,
         @Loggable @ConfigProperty(defaultValue = "10000") int rapidResetCheckPeriod) {}
 // spotless:on
--- a/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberConfig.java
+++ b/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberConfig.java
@@ -38,22 +38,13 @@ import org.hiero.block.node.base.Loggable;
  *     clients reconnect simultaneously. Excess requests are rejected with
  *     NOT_AVAILABLE status. Set to 0 to disable rate limiting.
  */
+// spotless:off
 @ConfigData("subscriber")
 public record SubscriberConfig(
-        @Loggable @ConfigProperty(defaultValue = "4000") @Min(100)
-        int liveQueueSize,
-
-        @Loggable @ConfigProperty(defaultValue = "4000") @Min(10)
-        long maximumFutureRequest,
-
-        @Loggable @ConfigProperty(defaultValue = "400") @Min(10)
-        int minimumLiveQueueCapacity,
-
-        @Loggable @ConfigProperty(defaultValue = "1_048_576") @Min(100_000)
-        int maxChunkSizeBytes,
-
-        @Loggable @ConfigProperty(defaultValue = "50") @Min(0)
-        int maxConcurrentSessions,
-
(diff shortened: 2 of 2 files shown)


block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
```diff
@@ -20,13 +20,15 @@
...
  * @param rapidResetCheckPeriod Period for counting rapid resets(stream RST sent by client before any data have been sent by server). Default value is 10,000 ms.
  */
+// spotless:off
 @ConfigData("server.http2")
 public record WebServerHttp2Config(
-        @Loggable @ConfigProperty(defaultValue = "1000") int flowControlTimeout,
-        @Loggable @ConfigProperty(defaultValue = "8_388_608") int initialWindowSize,
-        @Loggable @ConfigProperty(defaultValue = "8") long maxConcurrentStreams,
+        @Loggable @ConfigProperty(defaultValue = "2000") int flowControlTimeout,
+        @Loggable @ConfigProperty(defaultValue = "16_777_216") int initialWindowSize,
```
jsync-swirlds: This seems unnecessary; I suspect we are still fine with the earlier values.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java
```diff
@@ -20,13 +20,15 @@
...
+        @Loggable @ConfigProperty(defaultValue = "50") long maxConcurrentStreams,
         @Loggable @ConfigProperty(defaultValue = "10") int maxEmptyFrames,
         @Loggable @ConfigProperty(defaultValue = "8_388_608") int maxFrameSize,
         @Loggable @ConfigProperty(defaultValue = "8192") long maxHeaderListSize,
-        @Loggable @ConfigProperty(defaultValue = "50") int maxRapidResets,
+        @Loggable @ConfigProperty(defaultValue = "200") int maxRapidResets,
```
jsync-swirlds: I think we were fine with the previous value here as well.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
jsync-swirlds: Couldn't we have adjusted the Backfill client to allow for a larger parsing limit when calling the parse method?
We could accept a byte[] in the backfill client and then parse, when received, with a less limited parse size if we choose.

This might require the _client_ to hold more data in memory, but would relieve the server of the cost to parse the whole block (in theory we could choose not to parse the block at all here), split everything up, and construct multiple batch responses.

jsync-swirlds (approved):

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/SubscriberConfig.java
```diff
@@ -24,9 +24,27 @@
...
+// spotless:off
 @ConfigData("subscriber")
 public record SubscriberConfig(
         @Loggable @ConfigProperty(defaultValue = "4000") @Min(100) int liveQueueSize,
         @Loggable @ConfigProperty(defaultValue = "4000") @Min(10) long maximumFutureRequest,
-        @Loggable @ConfigProperty(defaultValue = "400") @Min(10) int minimumLiveQueueCapacity) {}
+        @Loggable @ConfigProperty(defaultValue = "400") @Min(10) int minimumLiveQueueCapacity,
+        @Loggable @ConfigProperty(defaultValue = "1_048_576") @Min(100_000) int maxChunkSizeBytes,
+        @Loggable @ConfigProperty(defaultValue = "50") @Min(0) int maxConcurrentSessions,
+        @Loggable @ConfigProperty(defaultValue = "10") @Min(0) int sessionRateLimitPerSecond) {}
```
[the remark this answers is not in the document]
AlfredoG87: https://github.com/hiero-ledger/hiero-block-node/issues/2136

PR #2119, opened 2026-02-03
fix(test): Fix ScheduledBlockingExecutor to resolve Backfill UT Flakyness

### LINKED ISSUE #2117: Flaky test: BackfillPluginTest - Greedy autonomous starts at peer earliest block when startBlock is default on an empty BN"
### Description

The test Greedy autonomous starts at peer earliest block when startBlock is default on an empty BN in BackfillPluginTest is flaky and intermittently fails in the CI pipeline.                                                                              
                                                                                                                                                                                                                                                              
  Test Details:                                                                                                                                                                                                                                               
  - Class: org.hiero.block.node.backfill.BackfillPluginTest                                                                                                                                                                                                   
  - Test: Greedy autonomous starts at peer earliest block when startBlock is default on an empty BN                                                                                                                                                           
  - Duration: ~12.8s when failing                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                              
  Observed Behavior:                                                                                                                                                                                                                                          
  Test passes locally but intermittently fails in CI pipeline.                                                                                                                                                                                                
                                                                                                                                                                                                                                                              
  Expected Behavior:                                                                                                                                                                                                                                          
  Test should consistently pass in both local and CI environments.   

  Possible Causes:                                                                                                                                                                                                                                            
  - Race condition in async/concurrent operations                                                                                                                                                                                                             
  - Timing-sensitive assertions without adequate wait/polling                                                                                                                                                                                                 
  - Test isolation issues (shared state between tests)                                                                                                                                                                                                        
  - CI environment resource constraints affecting timing                                                                                                                                                                                                      
                                                                   

### Steps to Reproduce

  Run the test multiple times in CI or with parallel test execution:                                                                                                                                                                                          
  `./gradlew :backfill:test --tests "BackfillPluginTest.Greedy autonomous starts at peer earliest block when startBlock is default on an empty BN"`

### Additional context

_No response_

AlfredoG87:
## Summary
Fix `ScheduledBlockingExecutor.execute()` to call `super.execute()` so tasks actually run.

## What was happening
The test fixture `ScheduledBlockingExecutor` overrides `execute()` to capture tasks, but it never called the parent's `execute()`. This meant any task submitted via `execute()` just sat in a queue forever.

The tricky part: `scheduleAtFixedRate()` bypasses `execute()` internally, so the periodic gap scanner worked fine. But when the scanner tried to kick off the actual backfill work with `executor.submit(drain)`, that task never ran. The `workerRunning` flag got set to `true` and stayed there since `drain()` never executed to reset it.

## Why it looked flaky
This test uses a short 500ms scan interval, so the scanner kept detecting the same gap over and over while the drain was stuck. Whether it hit the 30s timeout depended on CI timing - sometimes it squeaked by, sometimes it didn't.

## Test plan
- [x] Added regression test `ScheduledBlockingExecutorTest.executedTasksShouldRun()` - fails without fix, passes with fix
- [x] Ran `testHistoricalGapsSkippedWhenSchedulerRunning` 5 times, all passed (~6s each)
- [x] All 116 backfill tests pass
- [x] All 32 app tests pass

Fixes #2117


jsync-swirlds (commented):
I don't think we should be merging a test for test code.
That's like adding unit tests to improve test coverage of unit tests.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
```diff
@@ -30,5 +30,8 @@ public void execute(@NonNull final Runnable command) {
             Thread.currentThread().interrupt();
             throw new RuntimeException("Thread was interrupted while trying to put a task into the work queue", e);
         }
+        // Must invoke superclass version to preserve ScheduledThreadPoolExecutor functionality.
+        // Tasks submitted via execute() are treated as zero-delay scheduled tasks.
+        super.execute(command);
```
ata-nas: I would say we should absolutely not do that at all! We should take inspiration from the other blocking executor. We collect tasks and then, if we want, we can execute them serially, or async, but that should always be on demand in order to remove flakiness. But if we proceed with this change, we effectively lose all the benefits of this abstraction, we better simply use a regular pool. We do not solve flakiness this way.

If we do not care about what this class should promise and provide, let's just use a regular pool. Sometimes we do that for testing, and it is fine and acceptable.

AlfredoG87: ✅ Done!

Nana-EC (approved):

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
```diff
@@ -21,6 +21,15 @@ public ScheduledBlockingExecutor(int corePoolSize, @NonNull final BlockingQueue<
...
     }
 
+    /**
+     * Captures the task into the work queue without executing it.
+     * <p>
+     * We intentionally do NOT call {@code super.execute(command)} here. The
+     * purpose of this executor is to hold tasks until tests explicitly drain
+     * the queue, giving tests full control over execution ordering and timing.
+     * Calling the parent's execute would defeat that contract. Tests that need
+     * tasks to actually run should use a real {@link java.util.concurrent.ScheduledThreadPoolExecutor}.
```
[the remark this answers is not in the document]
Nana-EC: Can add in follow up

PR #2157, opened 2026-02-07
refactor(build): Reduce docker size to 271MB

berryware:
Used UBI10 Micro and RHEL JDK 25


berryware:
Clsoing pull request as E2E tests not running

PR #2167, opened 2026-02-09
feat: Files Recent Plugin Accessor IO Capabilities Improvements

### LINKED ISSUE #1023: Files Recent Plugin Accessor IO Capabilities Improvements
### Story Form

As a Block Node User/Operator
I want to have improved IO capabilities of the `Files Recent` Plugin
So that the application is much more flexible and comprehensive

### Technical Notes

- currently the `Files Recent` Plugin's IO capabilities are limited and based on runtime configuration
  - we rely on runtime configuration to determine input/output stream wrapping for compression/decompression
  - this limits us to only being able to set the configuration once and then on each restart if we do not preserve the configuration, we will not be able to do proper IO using the currently existing accessors
  - when we wish to get some bytes in any given supported format, we must make sure that we are doing this properly, decompressing and compressing again if needed when the source format is different than the desired one
- we should attempt to also do some performance improvements as needed
- we should also attempt to streamline the process of determining the compression type that was used to persist a block based on the actual one, not a runtime configuration assumption or other assumption
- we should also probably have a retry mechanism for IO operations

ivannov:
## Reviewer Notes
                                                                                                                                                     
This PR refactors the Files Recent Plugin to determine block compression types by reading magic bytes from file content rather than relying on the configuration or file extensions. This makes the system more robust and helps detect mismatches between file names and actual content.             
                                                                                                                                                       
  ### Key Changes

  1. **BlockFileBlockAccessor Enhancement**
     - Removed `compressionType` parameter from constructor - now determined automatically
     - Added `determineCompressionType(Path)` method that reads file headers and identifies compression by magic bytes
     - Added `compressionType()` package-private getter for testing
     - The method checks all known compression types' magic bytes and returns the matching type, or `NONE` if no match is found

  2. **Comprehensive Testing**
     - Added `BlockFileRecentPluginTest.testCompressionTypeIsDetermined()` that verifies compression detection for all combinations of:
       - Actual compression used in the file content (ZSTD, NONE, etc.)
       - File extension suggesting different compression types
     - This ensures that if files are misnamed or if someone adds a new `CompressionType` with incorrect magic bytes, the system will correctly detect the actual compression from file content

  3. **Test Cleanup**
     - Simplified `BlockFileBlockAccessorTest` by removing redundant test cases that are now covered by the more comprehensive test in `BlockFileRecentPluginTest`

## Related Issue(s)
Closes #1023

### DIFF
--- a/.github/workflows/helm-charts.yaml
+++ b/.github/workflows/helm-charts.yaml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Create kind cluster
         if: steps.install-needed.outputs.needed == 'true'
-        uses: helm/kind-action@92086f6be054225fa813e0a4b13787fc9088faab # v1.13.0
+        uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc # v1.14.0
 
       - name: Build Docker images
         if: steps.install-needed.outputs.needed == 'true'
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -184,7 +184,7 @@ jobs:
         uses: azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4 # v4.3.1
 
       - name: Setup Kind
-        uses: helm/kind-action@92086f6be054225fa813e0a4b13787fc9088faab # v1.13.0
+        uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc # v1.14.0
         with:
           install_only: true
           node_image: kindest/node:v1.31.4@sha256:2cb39f7295fe7eafee0842b1052a599a4fb0f8bcf3f83d96c7f4864c357c6c30
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -7,6 +7,7 @@ import static org.hiero.block.common.constants.StringsConstants.APPLICATION_PROP
 import static org.hiero.block.common.constants.StringsConstants.APPLICATION_TEST_PROPERTIES;
 import static org.hiero.block.node.spi.BlockNodePlugin.METRICS_CATEGORY;
 
+import com.hedera.hapi.block.stream.Block;
 import com.hedera.pbj.grpc.helidon.config.PbjConfig;
 import com.swirlds.common.metrics.platform.DefaultMetricsProvider;
 import com.swirlds.config.api.Configuration;
@@ -242,6 +243,7 @@ public class BlockNodeApp implements HealthFacility {
         return BlockNodeVersions.newBuilder()
                 .installedPluginVersions(pluginVersions)
                 .blockNodeVersion(SemanticVersionUtility.from(BlockNodeApp.class))
+                .streamProtoVersion(SemanticVersionUtility.from(Block.class))
                 .build();
     }
 
--- a/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
+++ b/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Stream;
+import org.hiero.block.api.BlockNodeVersions;
 import org.hiero.block.api.BlockNodeVersions.PluginVersion;
 import org.hiero.block.node.app.fixtures.plugintest.TestBlockMessagingFacility;
 import org.hiero.block.node.base.ranges.ConcurrentLongRangeSet;
@@ -234,16 +235,21 @@ class BlockNodeAppTest {
     void testBlockNodeVersions() throws IOException {
         final ServiceLoaderFunction serviceLoaderFunction = new ServiceLoaderFunction();
         final BlockNodeApp blockNodeApp = new BlockNodeApp(serviceLoaderFunction, false);
+        final BlockNodeVersions blockNodeVersions = blockNodeApp.blockNodeContext.blockNodeVersions();
 
-        final SemanticVersion blockNodeVersion =
-                blockNodeApp.blockNodeContext.blockNodeVersions().blockNodeVersion();
+        final SemanticVersion blockNodeVersion = blockNodeVersions.blockNodeVersion();
         assertNotNull(blockNodeVersion);
         // This will need to be changed to 1 at some point
         assertEquals(0, blockNodeVersion.major());
 
+        // test the stream protocol version
+        final SemanticVersion streamProtocolVersion = blockNodeVersions.streamProtoVersion();
+        assertNotNull(streamProtocolVersion);
+        assertEquals(0, streamProtocolVersion.major());
+        assertTrue(streamProtocolVersion.minor() > 70);
+
         // In dev, the plugins should have the same SemVer as the BlockNodeApp
-        final List<PluginVersion> pluginVersions =
-                blockNodeApp.blockNodeContext.blockNodeVersions().installedPluginVersions();
+        final List<PluginVersion> pluginVersions = blockNodeVersions.installedPluginVersions();
         for (PluginVersion pluginVersion : pluginVersions) {
             assertNotNull(pluginVersion.pluginId());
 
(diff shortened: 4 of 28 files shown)


block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
```diff
@@ -181,4 +187,50 @@ public void close() {
...
+        // Find the longest magic bytes sequence to determine how many bytes we need to read
+        final int max = Stream.of(compressionOpts)
+                .map(ct -> ct.magicBytes().length)
+                .max(Integer::compare)
+                .orElse(0);
+
+        // Read the file header once with enough bytes to check all compression types
+        final byte[] fileHeader = new byte[max];
+        try (final InputStream is = Files.newInputStream(verifiedBlockPath)) {
+            final int _ = is.read(fileHeader);
```
jsync-swirlds: Naming a variable `_` alone is not good or readability.
`ignored` is much clearer that you intend to ignore the number of bytes read.

You could also just leave out the assignment entirely:

```suggestion
            is.read(fileHeader);
```

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
```diff
@@ -181,4 +187,50 @@ public void close() {
...
+     * type is returned. If no magic bytes match, the file is assumed to be uncompressed.
+     *
+     * @param verifiedBlockPath the path to the block file that has been verified to exist
+     * @return the compression type of the file, or {@link CompressionType#NONE} if the file
+     *         is not compressed or uses a compression type without magic bytes
+     * @throws IOException if an I/O error occurs while reading the file header
+     */
+    CompressionType determineCompressionType(Path verifiedBlockPath) throws IOException {
+        // Get all available compression types to check
+        final CompressionType[] compressionOpts = CompressionType.class.getEnumConstants();
```
AlfredoG87:
```suggestion
        final CompressionType[] compressionOpts = CompressionType.values();
```

can we use standard Java convention for enums?

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
```diff
@@ -181,4 +187,50 @@ public void close() {
...
+        // Check each compression type's magic bytes against the file header
+        for (CompressionType currentOpt : compressionOpts) {
+            final byte[] magicBytes = currentOpt.magicBytes();
+            if (magicBytes != null && magicBytes.length > 0) {
+                // Extract the relevant portion of the header for this compression type
+                final byte[] headerChunk = Arrays.copyOf(fileHeader, magicBytes.length);
+                if (Arrays.equals(headerChunk, magicBytes)) {
+                    // Magic bytes match, return this compression type
+                    return currentOpt;
+                }
```
AlfredoG87: suggestion:  Consider using Arrays.compare with range or checking fileHeader.length >= magicBytes.length first
And avoid allocating a new array on every iteration:

```suggestion
                 // Extract the relevant portion of the header for this compression type
                 byte[] headerChunk = Arrays.copyOf(fileHeader, magicBytes.length);
                  if (magicBytes.length > 0 && fileHeader.length >= magicBytes.length) {
                      if (Arrays.compare(fileHeader, 0, magicBytes.length, magicBytes, 0, magicBytes.length) == 0) {
                          return currentOpt;
                      }
                  }
```

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -241,8 +241,8 @@ public BlockAccessor block(final long blockNumber) {
             if (Files.exists(verifiedBlockPath)) {
                 // we have the block so return it
                 try {
-                    final BlockFileBlockAccessor accessor = new BlockFileBlockAccessor(
-                            verifiedBlockPath, config.compression(), linksRootPath, blockNumber);
+                    final BlockFileBlockAccessor accessor =
+                            new BlockFileBlockAccessor(verifiedBlockPath, linksRootPath, blockNumber);
```
ata-nas: This is an example of breaking the accessor if we change something. If we pass a `BlockPath` and extend it, nothing will break here.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
```diff
@@ -47,22 +49,17 @@ final class BlockFileBlockAccessor implements BlockAccessor {
...
-     * @param compressionType the compression type of the block file, must not be null
      * @param linksRootPath the root path where hard links to block files will be created
      * @param blockNumber the block number of the block
      */
-    BlockFileBlockAccessor(
-            @NonNull final Path blockFilePath,
-            @NonNull final CompressionType compressionType,
-            @NonNull final Path linksRootPath,
-            final long blockNumber)
+    BlockFileBlockAccessor(@NonNull final Path blockFilePath, @NonNull final Path linksRootPath, final long blockNumber)
```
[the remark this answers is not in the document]
ivannov: I started from there: implementing the BlockPath logic in the _recent_ plugin. However, I didn't find the reason for that. And I tried hard to find the usecase :)

BlockPath in _historic_ plugin is focused more on encapsulating blocks in zips in the trie file system. Here we don't have that requirement. We only wanted to make sure that we always recognize the compression type and that's it.

If we want to remove the dependency to `BlockFile`, then I am happy to re-implement its `nestedDirectoriesBlockFilePath` static method here. That's the only dependency that I see in the plugin

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
ivannov: Changed to use values()

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
ivannov: Used Arrays.compare

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/RecentBlockPath.java
```diff
@@ -0,0 +1,144 @@
...
+            final Set<Long> blockNumbers = new HashSet<>();
+            final String fullExtension = BLOCK_FILE_EXTENSION + compressionType.extension();
+            Files.walkFileTree(
+                    basePath,
+                    Set.of(),
+                    MAX_FILE_SEARCH_DEPTH,
+                    new BlockNumberCollectionVisitor(blockNumbers, fullExtension));
+            return blockNumbers;
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
```
ata-nas: If we cannot do anything about the checked exception here, we should propagate it, rather than throwning unchecked. The caller _could_ have something to do about the exception. If not, they can also propagate. We are silencing a potential issue or an expected event that could be handled.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -208,6 +219,29 @@ private void initMetrics(final Metrics metrics) {
...
+            final String fullExtension = BLOCK_FILE_EXTENSION + compressionType.extension();
+            Files.walkFileTree(
+                    basePath,
+                    Set.of(),
+                    MAX_FILE_SEARCH_DEPTH,
+                    new BlockNumberCollectionVisitor(blockNumbers, fullExtension));
+            return blockNumbers;
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
```
ata-nas: Just noting for future improvements, based on recent talks we've had with the team, we better not throw or silence exceptions, they should be handled by someone who can do something about it. By extending and improving the server health (future effort), we will be able to do something about an exception occurring here, by utilizing the improved server health in the caller.

ata-nas (approved):
Great job on this one! Many great improvements here, mostly dynamic and correct compression type resolution for a stored block.

Issue #2176, opened 2026-02-09
`wrap` command fails on corrupted/empty signature files


rockysingh:
### Story Form

As a **block node operator**
  I want the **`wrap` command to skip corrupted signature files gracefully**
  So that **a single corrupted file doesn't stop the entire wrapping process**

## Description

  The `blocks wrap` command fails with an exception when encountering corrupted or empty signature files (0 bytes).

  ### Error

  java.lang.RuntimeException: Error processing signature file 2020-05-29T11_30_05.385184Z/node_0.0.11.rcd_sig because: Unrecognized signature file format version -1 in file ...

  ### Root Cause

  - Signature file is 0 bytes (empty/corrupted download)
  - Reading first byte returns -1 (EOF)
  - Code throws exception for unrecognized version -1

  ### Fix

  - Catch exceptions when parsing signature files in `ParsedRecordBlock.parse()`
  - Log warning with file path and size in bytes
  - Skip corrupted file and continue with remaining signatures
  - Block can still be validated with other node signatures

PR #2177, opened 2026-02-09
fix: Skip corrupted signature files gracefully in wrap command

rockysingh:
## Summary                     

  - Skip corrupted/empty signature files instead of failing the wrap command                                                                                                                                                                                                                                 
  - Log warning with file path and size in bytes for diagnosis
  - Add unit test for corrupted signature file handling                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                           
  ## Problem

  The `blocks wrap` command failed with an exception when encountering corrupted or empty (0 byte) signature files:

  java.lang.RuntimeException: Error processing signature file 2020-05-29T11_30_05.385184Z/node_0.0.11.rcd_sig because: Unrecognized signature file format version -1

  This was caused by incomplete downloads where signature files were 0 bytes. Reading the first byte returned -1 (EOF), which was treated as an invalid version.

  ## Solution

  - Wrap signature file parsing in try-catch in `ParsedRecordBlock.parse()`
  - Log warning with file path and size: `Warning: Skipping corrupted signature file: .../node_0.0.11.rcd_sig (0 bytes) - ...`
  - Filter out failed parses and continue with valid signatures
  - Blocks can still be validated with remaining node signatures

  ## Testing

  - Added `testCorruptedSignatureFilesAreSkipped` test covering:
    - Empty file (0 bytes)
    - Invalid version byte (0xFF)
    - Truncated file
    
    
    This closes: https://github.com/hiero-ledger/hiero-block-node/issues/2176

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/model/parsed/ParsedRecordBlock.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/model/parsed/ParsedRecordBlock.java
@@ -52,7 +52,7 @@ public record ParsedRecordBlock(
                         .map(sigFile -> {
                             try {
                                 return new ParsedSignatureFile(sigFile);
-                            } catch (Exception e) {
+                            } catch (RuntimeException e) {
                                 System.err.println("Warning: Skipping corrupted signature file: " + sigFile.path()
                                         + " (" + sigFile.data().length + " bytes) - " + e.getMessage());
                                 return null;


jsync-swirlds (dismissed):

Nana-EC (dismissed):
LG, but we should consider refining the catch

jsync-swirlds (approved):

Nana-EC (approved):

PR #2185, opened 2026-02-11
test: plugin startup tests

### LINKED ISSUE #2144: Add unit tests that prove that starting plugins in parallel works correctly
### Story Form

As a BlockNode engineer I want to know that starting plugins in parallel works and that adding plugins that expect an start order will fail.

### Technical Notes

Now that plugins are started in parallel tests need to be added to verify that this works. There are currently 13 plugins that get loaded. Not all permutations can be tested as that is 6.2B tests. 3 test cases come to mind, that should produce errors if there are any startup dependencies:
1. Test in parallel as that is currently what the code is doing.
2. Test in reverse order returned by the service loader. This guarantees the worst case that any dependencies on previously loaded plugins will not be there.
3. Use Collections.shuffle to test a few more permutations to introduce some controlled randomness

berryware:
-  Test in parallel.
-  Test in ServiceLoader Order
-  Test in reverse order returned by the service loader. This should break any dependencies on load order.
-  Use {@code Collections.shuffle()} to test a few more permutations to introduce some controlled randomness.
  -  Use just a as this greatly increases the unit test time.

Fixes #2144

### DIFF
--- a/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
+++ b/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
@@ -202,7 +202,7 @@ class BlockNodeAppTest {
     @Test
     @DisplayName("Test plugin startup in shuffled order")
     void testPluginStartupIndependence() throws IOException {
-        final int SHUFFLE_COUNT = 10;
+        final int SHUFFLE_COUNT = 100;
         final ServiceLoaderFunction serviceLoaderFunction = new ServiceLoaderFunction();
 
         // Case 4: Test in reverse order returned by the service loader.


block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
```diff
@@ -132,4 +134,98 @@ void testMain() throws IOException {
...
+        startBlockNode(blockNodeApp);
+    }
+
+    /**
+     * This test aims to insure the independence of plugins by starting them in varying order.
+     * Use {@code Collections.shuffle()} to test a few more permutations to introduce some controlled randomness.
+     * as this greatly increases the unit test time.
+     */
+    @Test
+    @DisplayName("Test plugin startup in shuffled order")
```
jsync-swirlds: We could, if we wish, test every possible ordering by generating the permutations instead of a random shuffle (which would be flakey if it detected any issues).

That would be a lot of test cases (hundreds or even thousands), but we do that for PBJ (which generates close to an hundred thousand test cases dynamically every build), so it's not unreasonable.

PR #2179, opened 2026-02-09
feat: Reduce docker size

berryware:
## Summary
- Reduced the docker image by 30%
- Changed the base image to eclipse-temurin:25.0.2_10-jre-ubi10-minimal
- Removed the application tar file from the image

PR #2190, opened 2026-02-12
fix: Empty .core-modules.txt breaks plugin deduplication at boot

### LINKED ISSUE #2188: Empty .core-modules.txt Breaks Plugin Deduplication when deploying BN using helm
## Error

```
java.lang.module.ResolutionException: Modules com.goterl.lazysodium and lazysodium.java
  export package com.goterl.lazysodium to module com.google.common
```

### Steps to Reproduce

Try to deploy the BN using helm charts.

## Root Cause

PR #2180 bakes `.core-modules.txt` into the Docker image using `/usr/local/java/bin/jar`. PR #2179 then switched the base image to `eclipse-temurin` where Java lives at a different path. Because the `jar` call redirects stderr to `/dev/null`, it fails silently and `.core-modules.txt` is written empty.

With an empty core modules list, the `resolve-plugins` init container can't deduplicate anything — all 116 transitive dependencies get copied to `/plugins/`, including jars that overlap with core `lib/`. Two lazysodium artifacts with conflicting JPMS packages end up on the module path, and the JVM crashes at boot layer initialization. The pod never becomes ready.

## Affected Run

[Solo E2E Scheduler #21895128376](https://github.com/hiero-ledger/hiero-block-node/actions/runs/21895128376)

## Contributing PRs

| PR | Change | Merged |
|----|--------|--------|
| #2158 | Introduced dynamic plugin loading and `resolve-plugins` init container | Feb 10 |
| #2180 | Baked `.core-modules.txt` into Docker image using `/usr/local/java/bin/jar` | Feb 10 17:00 UTC |
| #2179 | Switched base image to `eclipse-temurin` (Java no longer at `/usr/local/java/`) | Feb 10 23:00 UTC |

AlfredoG87:
### `.core-modules.txt` generation — Gradle `ModuleFinder` (package-level)

The previous approach generated `.core-modules.txt` with normalized jar names, which broke when Gradle-patched jars had different JPMS module names than Maven originals. The `distTar` task now uses Java's `ModuleFinder` API to introspect every core runtime jar and extract all Java **package names** (via `ModuleDescriptor.packages()`). The resulting `.core-modules.txt` is included in the distribution tar and baked into the Docker image — no JDK intermediate stage needed.

### Package-based deduplication in `resolve-plugins`

The original matching used JPMS module names, which breaks because Gradle patches jars with different module names than Maven originals (e.g. `resource-loader-2.0.2-module.jar` gets JPMS name `com.goterl.resourceloader`, while Maven's unpatched jar gets automatic name `resource.loader`). Jar-name matching was also fragile when Maven resolves a different version of the same library. Package-level matching is resilient to both — the init container now runs `jar --describe-module` on each Maven-resolved jar, and if any of its packages overlap with the core package list, the jar is skipped.

### `distTar` extraction in Gradle

UBI-minimal's `tar` fails under QEMU emulation on `linux/arm64` (`Cannot open: Invalid argument`). A new `extractDistForDocker` Gradle `Copy` task extracts the tar after `distTar`, so the Dockerfile just `COPY`s the pre-extracted directory.

### Helm CI — local image builds

`ct install` previously pulled images from `ghcr.io` and only ran when chart files changed, so Dockerfile-only PRs like #2179 were never install-tested. The workflow now builds images locally with Gradle, tags them to match Helm chart expectations, and loads them into the kind cluster. The install step is gated on changes to Dockerfiles, `build.gradle.kts`, or chart files — PRs touching only Java source or docs skip it entirely.

### Simulator startup data

Pod restarts before any blocks are processed leave both startup data files present but empty. Previously this threw `IllegalStateException`. Now the empty-files case is treated as initial state (defaults: block number `-1`, zeroed hash), matching the behavior when files don't exist at all.

Fixes #2188

Daily Scheduler was also fixed: 
https://github.com/hiero-ledger/hiero-block-node/actions/runs/21936143999
its using SNAPSHOT from this branch: https://github.com/hiero-ledger/hiero-block-node/actions/runs/21936222076

### DIFF
--- a/.github/workflows/helm-charts.yaml
+++ b/.github/workflows/helm-charts.yaml
@@ -43,7 +43,7 @@ jobs:
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
           distribution: temurin
-          java-version: 25.0.2
+          java-version: 25
 
       - name: Run chart-testing (list-changed)
         id: list-changed


.github/workflows/helm-charts.yaml
```diff
@@ -39,6 +39,12 @@ jobs:
       - name: Set up chart-testing
         uses: helm/chart-testing-action@6ec842c01de15ebb84c8627d2744a0c2f2755c9f # v2.8.0
 
+      - name: Set up Java
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
+        with:
+          distribution: temurin
+          java-version: 25.0.2
+
```
jsync-swirlds: Could we use the temurin base image instead of installing Java?

Also, (if we keep this step) perhaps use version `25` to reduce the rate at which this needs to be manually updated.

charts/block-node-server/templates/statefulset.yaml
```diff
@@ -160,13 +160,21 @@ spec:
...
 
-              echo "Filtering out jars whose JPMS module already exists in core..."
+              echo "Filtering out jars whose packages overlap with core..."
               copied=0
               skipped=0
               for jar in /tmp/deps/*.jar; do
                 [ -f "$jar" ] || continue
-                mod=$(jar --describe-module --file="$jar" 2>/dev/null | head -1 | sed 's/@.*//' | awk '{print $1}')
-                if [ -n "$mod" ] && [ "$mod" != "No" ] && grep -qx "$mod" /plugins/.core-modules.txt 2>/dev/null; then
+                packages=$(jar --describe-module --file="$jar" 2>/dev/null \
```
jsync-swirlds: Do we want to add an if test here to be sure `jar` exists?
```shell
jar_binary=$(which jar)
if [[ $jar_binary != "" && -x $jar_binary ]]
```

.github/workflows/helm-charts.yaml
AlfredoG87: I don't know how to set a base image on the runner, I only know we use: 
```
    runs-on: hiero-block-node-linux-medium

```

but now that you mention, this could be an idea for RE, so they can provide runners that already have most of the tooling, heck, they could provide runners thought for BN, that already have everything we need, and the only thing we do is checkout of the branch and pull (update) and then run tests, etc.

andrewb1269 (approved):

Issue #2204, opened 2026-02-13
Add a server detail status grpc call to the block node server


berryware:
### Story Form

As a block node user I want to be able to get detailed status information from the block node
So that I can get the block node version, a list of plugin information including name, version, and features.

### Technical Notes

- The block version information should get loaded in the constructor so that it is only processed once. 
- The block version information, including plugin version,  should default to the JPMS module information. 
- The plugin information can be overridden by the plugin itself

PR #2207, opened 2026-02-13
feat: Detail status api

berryware:
- Updated the node_service.proto to include the new service method as defined by @jsync-swirlds 
- Added a new `version()` method to `BlockNodePlugin`
  - The default functionality uses the module information from JPMS to gather all the versions and features.
- Added a helper class `SemanticVersionUtilies` to convert the JPMS version strings to `SemanticVersions`
- Added a helper class `ModuleInfo` to gather the module information for a module
  - module information is cached as all the classes in that module will share the same module information
- The `BlockVersions` class is build when the block node server starts and is store in the BlockNodeContext so that it is available to all plugins

Fixes: #2204

### DIFF
--- a/block-node/spi-plugins/src/main/java/module-info.java
+++ b/block-node/spi-plugins/src/main/java/module-info.java
@@ -16,7 +16,5 @@ module org.hiero.block.node.spi {
     requires transitive com.swirlds.metrics.api;
     requires transitive org.hiero.block.protobuf.pbj;
     requires transitive io.helidon.webserver;
-    requires org.hiero.block.common; // javax.annotation.processing.Generated
     requires static transitive com.github.spotbugs.annotations;
-    requires static java.compiler;
 }
--- a/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/module/SemanticVersionUtility.java
+++ b/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/module/SemanticVersionUtility.java
@@ -5,7 +5,6 @@ import com.hedera.hapi.node.base.SemanticVersion;
 import java.lang.module.ModuleDescriptor;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import org.hiero.block.common.utils.StringUtilities;
 
 /**
  * A class to facilitate constructing a SemanticVersion
@@ -65,8 +64,8 @@ public final class SemanticVersionUtility {
         // Build the SemanticVersion from the matcher values
         final SemanticVersion.Builder semanticVersionBuilder = new SemanticVersion.Builder();
         semanticVersionBuilder.major(major).minor(minor).patch(patch);
-        if (!StringUtilities.isBlank(preRelease)) semanticVersionBuilder.pre(preRelease);
-        if (!StringUtilities.isBlank(build)) semanticVersionBuilder.build(build);
+        if (preRelease != null) semanticVersionBuilder.pre(preRelease);
+        if (build != null) semanticVersionBuilder.build(build);
 
         return semanticVersionBuilder.build();
     }


common/src/main/java/org/hiero/block/common/utils/SemanticVersionUtilities.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A class to facilitate constructing a SemanticVersion
+ */
+public final class SemanticVersionUtilities {
+    private static final Pattern SEMVER_PATTERN = Pattern.compile(
+            "^(?<major>0|[1-9]\\d*)\\.(?<minor>0|[1-9]\\d*)\\.(?<patch>0|[1-9]\\d*)(?:-(?<preRelease>(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+(?<build>[0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$");
```
jsync-swirlds: We can probably simplify this if we just allow pre-release and build to be any string.
Most "semver" strings don't really follow proper rules for the pre-release and build sections (actually, even for version it's not uncommon to see alphanumeric values like `seven` or `v15a` in major, minor, or patch values).

We have to match our proto value in terms of restrictions (e.g. major, minor, and patch are numbers), but probably should be as relaxed as possible beyond that.

common/src/main/java/module-info.java
```diff
@@ -7,5 +7,6 @@
     requires transitive com.hedera.pbj.runtime;
     requires transitive org.hiero.block.protobuf.pbj;
     requires com.swirlds.common;
+    requires java.management;
```
jsync-swirlds: What do we need from this module?
I tried to find where it's used and couldn't.

berryware: removed.

block-node/spi-plugins/src/main/java/module-info.java
```diff
@@ -15,6 +15,7 @@
     requires transitive com.swirlds.metrics.api;
     requires transitive org.hiero.block.protobuf.pbj;
     requires transitive io.helidon.webserver;
+    requires org.hiero.block.common;
```
jsync-swirlds: Rather than adding this here, we're probably better off moving ModuleInfo to SPI.

common/src/main/java/org/hiero/block/common/utils/SemanticVersionUtilities.java
jsync-swirlds: Rather than adding this to common (which we'd prefer not to add to), it might be good to add to the status plugin.

protobuf-sources/src/main/proto/block-node/api/node_service.proto
```diff
@@ -95,7 +104,41 @@ message BlockNodeVersions {
...
+     * Version information for a registered plugin.
+     * <p>
+     * This MAY be empty.<br/>
+     * Providing version information is OPTIONAL for each plugin, but defaults
+     * to the module version information.
+     */
+    message PluginVersion {
+        /**
+         * The id of the plugin.
+         * defaults to the class name of the plugin
```
jsync-swirlds: These don't quite meet documented standards, but it's a bit tricky to get just right.
I've commented with recommended adjustments.

```suggestion
         * A plugin identifier.<br/>
         * This string is provided by the plugin as an identifier for that plugin.
         * <p>
         * This value SHALL be the canonical class name of the plugin implementation class
         * if not provided by the plugin.
```

protobuf-sources/src/main/proto/block-node/api/node_service.proto
```diff
@@ -95,7 +104,41 @@ message BlockNodeVersions {
...
+         */
+        string plugin_id = 1;
+        /**
+         * The `SemanticVersion` of the plugin.
+         * defaults to the `SemanticVersion` of the module the plugin is in.
```
jsync-swirlds:
```suggestion
         * A plugin version.<br/>
         * Plugin versions follow semantic version format.
         * <p>
         * If not set by the plugin this value SHALL be set to the `SemanticVersion` of the
         * module that provides the plugin.
```

protobuf-sources/src/main/proto/block-node/api/node_service.proto
```diff
@@ -95,7 +104,41 @@ message BlockNodeVersions {
...
+         */
+        proto.SemanticVersion plugin_software_version = 2;
+        /**
+         * The features of the plugin.
+         * defaults to the defined services provided by the plugin as defined in
+         * the module.
```
jsync-swirlds: Here, I recommend we adjust the default to be empty, rather than the module "services", because that will be the same for all plugin modules.
```suggestion
         * A set of plugin features.<br/>
         * This set of one or more string values describes the features provided by the plugin.
         * <p>
         * If not set by the plugin, this SHALL NOT be set.
```

common/src/main/java/org/hiero/block/common/utils/ModuleInfo.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A class that makes it easier to retrieve the information from the module. It will cache module-info so that it can
+ * be reused efficiently
+ */
+public final class ModuleInfo {
```
AlfredoG87: Not sure I like the name of this class its easy to confuse it with `module-info.java`

AlfredoG87 (approved):
LG, just 2 very minor things.

Please let me know if in need of re-approval.

block-node/spi-plugins/src/main/java/module-info.java
berryware: moved to spi and renamed to ModuleInfoAccessor

common/src/main/java/org/hiero/block/common/utils/ModuleInfo.java
berryware: renamed to ModuleInfoAccessor

block-node/spi-plugins/src/main/java/module-info.java
jsync-swirlds: Not totally sure all the added items here are necessary.
Perhaps just double-check all of them before merging.

berryware: I used a `StringUtilities` call from common. A null check suffices, so I removed this dependency. Removed the `requires static java.compiler;` as well

PR #2225, opened 2026-02-20
fix: Use 5-arg parse to correctly pass maxSize for sidecar files

### LINKED ISSUE #2218: `blocks wrap` command fails on block 48168427 (2023-05-17) due to PBJ protobuf parsing size limit.
### Story Form

  As a **block node operator**                                                                                                                                                                                                                         
  I want to **parse sidecar files larger than 2MB during block wrapping**
  So that **I can successfully wrap all mainnet blocks without parsing failures**              


### Technical Notes

  ## Error

  com.hedera.pbj.runtime.ParseException: sidecar_records size 4511896 is greater than max 2097152

  ## Description

  The sidecar file for this block is ~4.5MB, but PBJ's default max message size is 2MB (2097152 bytes). This causes parsing to fail when processing mainnet record files.

  The issue is in `ParsedRecordBlock.java:66` where `SidecarFile.PROTOBUF.parse()` is called without specifying a larger max size.

  ## Affected Block

  - Block: 48168427
  - Date: 2023-05-17
  - Sidecar size: 4,511,896 bytes (~4.3MB)

  ## Proposed Fix

  Use the PBJ `parse()` overload that accepts a custom `maxSize` parameter, or configure a larger default limit for sidecar file parsing.

rockysingh:
The 3-arg `Codec.parse(Bytes, boolean, int)` method incorrectly passes
the int parameter as `maxDepth`, then hardcodes the default 2MB `maxSize`.
This caused `ParseException` for blocks with large sidecars (>2MB).

Switch to the 5-arg `parse(ReadableSequentialData, boolean, boolean,
int, int)` to explicitly set both `maxDepth (512)` and `maxSize (8MB)`

this closes #2218

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/model/parsed/ParsedRecordBlock.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/model/parsed/ParsedRecordBlock.java
@@ -74,7 +74,7 @@ public record ParsedRecordBlock(
                                 return SidecarFile.PROTOBUF.parse(
                                         Bytes.wrap(ps.data()).toReadableSequentialData(),
                                         false, // strictMode
-                                        false, // parseUnknownFields
+                                        true, // parseUnknownFields
                                         MAX_DEPTH, // maxDepth
                                         MAX_SIDECAR_SIZE); // maxSize
                             } catch (ParseException e) {


jsync-swirlds (approved):
One possible improvement.

jsync-swirlds (approved):

Issue #2173, opened 2026-02-09
Add balance checking to wrapped block validator


jasperpotts:
The wrapped block validator command checks for 50B HBAR at the end of each block but that is not enough to know that all data and amendments are good.  To do that we need to validate with CSV balance files and or saved states.

So need to add to validator to download and signature check CSV balance files and validate balances against what is computed from differential changes in the block data.

Issue #2212, opened 2026-02-17
Add to wrapped block validator token tracking


jasperpotts:
Wrapped block validator should track token balances as well as accounts

PR #2265, opened 2026-02-25
feat: wrapped block validator token tracing for post october 2023 with saved states

rockysingh:
## Summary                                                                                                                                                                                                                                                  
  - Add token balance tracing to the wrapped block validator, extending beyond HBAR-only supply checks to per-account, per-token balance verification against signed protobuf balance files from GCP                                                        
  - Introduce `FetchBalanceCheckpointsCommand` (`blocks fetchBalanceCheckpoints`) to download and bundle weekly/monthly balance snapshots from the mainnet GCP bucket into length-prefixed protobuf checkpoint files                                          
  - Add `BalanceCheckpointValidator` and `BalanceProtobufValidator` to compare computed running account/token balances against checkpoint snapshots at configurable intervals (`--balance-check-interval-days`, default 30)                                   
  - Add `BalanceFileBucket` utility for downloading midnight-targeted `.pb.gz` balance files from GCP with signature verification support                                                                                                                     
  - Add `RunningAccountsState` to track per-account HBAR and token balances across blocks, processing amendments (crypto transfers, token mints/burns/wipes/transfers)                                                                                        
  - Add bundled checkpoint auto-loading (`accountBalances_91019204.pb.gz`) so `validate-wrapped` works out of the box without a separate fetch step                                                                                                           
  - Add new CLI options to `validate-wrapped`: `--validate-balances`, `--balance-checkpoints`, `--custom-balances-dir`, `--balance-check-interval-days`                                                                                                       
  - Add comprehensive unit tests for `BalanceCheckpointValidator`, `BalanceCheckpointsLoader`, `BalanceProtobufValidator`, and `BalanceFileBucket`                                                                                                            
  - Update documentation for `fetchBalanceCheckpoints` and `validate-wrapped` commands               
  
  ## Notes
  
  
  The bundled `accountBalances_91019204.pb.gz` checkpoint was generated using the standalone [hiero-state-to-balance-file]project, which extracts account/token balances from a consensus node
  saved state at block 91,019,204. This file provides a known-good starting point for balance validation. If the saved-state extraction tool or its protobuf format diverges in the future, this checkpoint may need to be regenerated. There is a potential diverge issue we need to raise. The "diverge issue" is that this is a separate standalone project — not part of the hiero-block-node repo — and the generated .pb.gz file was copied over. If that project's protobuf format, extraction logic, or Hiero dependency version changes, the bundled file could become stale or incompatible. Therefore, we will need to generate other files for any diverge.                                                                                                                                                      
                                                                                                                                                                                                                                                            
  ## Test plan
  - [ ] Verify `blocks validate-wrapped` runs with default balance validation enabled (uses bundled checkpoint)
  - [ ] Verify `blocks fetchBalanceCheckpoints` downloads and bundles balance files from GCP
  - [ ] Verify `--no-validate-balances` disables balance checking
  - [ ] Verify `--balance-check-interval-days` controls checkpoint frequency
  - [ ] Verify `--custom-balances-dir` loads balance files from a local directory
  - [ ] Run unit tests: `BalanceCheckpointValidatorTest`, `BalanceCheckpointsLoaderTest`, `BalanceProtobufValidatorTest`, `BalanceFileBucketTest`
  
  Also, rebased using this pr: https://github.com/hiero-ledger/hiero-block-node/pull/2184 as it was needed to work on  this issue etc.
  
  
  This closes:  #2173 and #2212

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/wrapped/ValidateWrappedBlocksCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/wrapped/ValidateWrappedBlocksCommand.java
@@ -6,6 +6,7 @@ import static org.hiero.block.tools.blocks.model.hashing.BlockStreamBlockHasher.
 import com.hedera.hapi.block.stream.Block;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.concurrent.Callable;
@@ -262,11 +263,12 @@ public class ValidateWrappedBlocksCommand implements Callable<Integer> {
         String[] bundledFiles = {"accountBalances_91019204.pb.gz"};
 
         for (String filename : bundledFiles) {
-            try (var stream = getClass().getResourceAsStream("/metadata/" + filename)) {
+            try (InputStream stream = getClass().getResourceAsStream("/metadata/" + filename)) {
                 if (stream != null) {
                     // Extract block number from filename
                     String blockStr = filename.replace("accountBalances_", "").replace(".pb.gz", "");
                     long blockNumber = Long.parseLong(blockStr);
+                    // Note: loadFromGzippedStream may close the stream, do not read from it after this call
                     validator.loadFromGzippedStream(stream, blockNumber);
                     System.out.println(Ansi.AUTO.string("@|yellow Loaded bundled checkpoint:|@ block " + blockNumber));
                 }


tools-and-tests/tools/src/main/resources/metadata/accountBalances_91019204.pb.gz
jsync-swirlds: Do we need these files in the git repo?

tools-and-tests/tools/src/main/resources/metadata/balance_checkpoints_monthly.zstd
jsync-swirlds: Same question.

tools-and-tests/tools/src/main/resources/metadata/balance_checkpoints_weekly.zstd
jsync-swirlds: Same question.

rockysingh: This is created using a command but it can be very flakey to run due to GCP but gets even worse as its at a weekly granularity. This is why I've pushed them to the repo as a source of truth.

tools-and-tests/tools/src/main/resources/metadata/accountBalances_91019204.pb.gz
rockysingh: This was created using another project etc so there is no command for it.

jsync-swirlds (approved):

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/wrapped/ValidateWrappedBlocksCommand.java
```diff
@@ -252,4 +252,25 @@ public Integer call() {
...
+        // List of bundled checkpoint files (block number extracted from filename)
+        String[] bundledFiles = {"accountBalances_91019204.pb.gz"};
+
+        for (String filename : bundledFiles) {
+            try (var stream = getClass().getResourceAsStream("/metadata/" + filename)) {
+                if (stream != null) {
+                    // Extract block number from filename
+                    String blockStr = filename.replace("accountBalances_", "").replace(".pb.gz", "");
+                    long blockNumber = Long.parseLong(blockStr);
+                    validator.loadFromGzippedStream(stream, blockNumber);
```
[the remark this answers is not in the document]
rockysingh: Added as requested.

jsync-swirlds (approved):

PR #2284, opened 2026-02-27
fix: 2280 cli wrapping performance part 2

### LINKED ISSUE #2280: CLI Wrapping performance part 2
Working on more performance improvements to record stream wrapping and validation.

jasperpotts:
fixes #2280
# perf: CLI wrap/validate pipeline optimisations (part 2)

Continues the performance work from #2258, addressing two separate bottlenecks in the
record-file → wrapped-block toolchain.

---

### `BlockStreamBlockHashRegistry` — batched writes

`addBlock()` previously issued one `RandomAccessFile.seek + write` syscall per block (48 bytes
each). With 90 million blocks this amounts to 90 million individual disk writes. The registry now
accumulates hashes in a 10,000-entry in-memory buffer (480 KB) and flushes in a single large write
when full or on `close()`. This reduces ~90 M syscalls to ~9,000 without impacting crash safety —
the existing monthly checkpoint and shutdown hook cover recovery.

### `BlockWriter.writeBlockEntry` — pre-computed CRC-32 overload

Added a second `writeBlockEntry(zip, blockPath, bytes, crc32)` overload that accepts a
pre-computed CRC-32. The pipeline's Stage 3 (compress) now computes the checksum alongside
compression; Stage 4 (zip write) only performs I/O. This eliminates a redundant full-byte-scan on
the hot zip-write thread.

### `ToWrappedBlocksCommand` — directory creation cache

`Files.createDirectories()` was called on every block. Since directories only change every 1,000
blocks, 999 of every 1,000 calls were no-ops. A `lastCreatedDir` variable now skips the call when
the directory hasn't changed.

### `TarZstdDayReaderUsingExec` — 256 KB read buffer on `zstd` stdout

Wrapped `process.getInputStream()` in a `BufferedInputStream(256 * 1024)`. The unbuffered stream
caused many small reads from the process pipe, creating unnecessary context-switch overhead. Also
fixed a latent bug where `streamTarZstd(path)` was not using parallel decompression (`-T0`).

---

### `BlockReader` — zip file-system cache

`readBlockFromZipWithPowersOfTen()` previously called `FileSystems.newFileSystem(zipPath)` for
every block. A 10,000-block zip triggered 10,000 open/close cycles on the same file, each
re-parsing the central directory. A static `ConcurrentHashMap<Path, FileSystem>` cache
(`ZIP_FS_CACHE`) now keeps each zip open across consecutive reads — one open per zip file instead
of one per block. `BlockReader.closeZipCache()` releases the file handles when done.

### `ValidateWrappedBlocksCommand` — 2-stage parallel pipeline

The validator was entirely sequential. It now uses a sliding-window prefetch pipeline:

| Stage | Thread(s) | Work |
|---|---|---|
| **Stage 1** | `readAndParsePool` (N threads) | `BlockReader.readBlock` (zip I/O + decompress + protobuf parse) + `validateRequiredItems` + `validateNoExtraItems` |
| **Stage 2** | Main thread (sequential) | `validateBlockChain`, `validateHistoricalBlockTreeRoot`, `validate50Billion`, `hashBlock`, `streamingHasher.addNodeByHash` |

Stage 2 stays sequential because each check requires state computed from all prior blocks
(previous block hash, streaming Merkle tree, running balance map).

### New CLI options

| Option | Default | Description |
|---|---|---|
| `--threads` | CPUs − 1 | Stage 1 thread pool size |
| `--prefetch` | threads × 2 | Sliding window size |
| `--start-block` | first stored | First block to validate |
| `--end-block` | last stored | Last block to validate |
| `--skip-50b` | false | Skip the 50-billion HBAR supply check |

Progress reporting changed from every-1,000-blocks to time-based (once per 5 seconds), matching
`ToWrappedBlocksCommand`.


rockysingh (approved):

jasperpotts:
This has been made redundent by https://github.com/hiero-ledger/hiero-block-node/pull/2307

Issue #2267, opened 2026-02-26
Testnet Support for Block Stream Tools


rockysingh:
## Background
  The tools project (`tools-and-tests/tools/`) currently assumes mainnet throughout — hardcoded GCP buckets, mirror node URLs, genesis dates, node ID ranges, address books, and balance constants. This epic tracks the work to make the tooling configurable for **testnet** (and eventually previewnet).

### Intention

We want to get the tools project working on testnet as a fast feedback for download, validate and wrap and wrap-validate feedback. 

 ### Inventory of Mainnet-Hardcoded Values        
                                                          
  | Category | Current Value | Testnet Equivalent | File(s) |
  |----------|--------------|-------------------|---------|                                                                                                                                                                                                   
  | GCS Bucket | `hedera-mainnet-streams` | `hedera-testnet-streams` | `DownloadConstants`, `MainNetBucket` |
  | Mirror URL | `mainnet-public.mirrornode.hedera.com` | `testnet.mirrornode.hedera.com` | `MirrorNodeUtils` |                                                                                                                                               
  | Genesis Date | `2019-09-13` | TBD | `TimeUtils`, `UpdateDayListingsCommand`, `ExtractDayBlocksFromApi` |                                                                                                                                                
  | Node ID Range | `3–37` | TBD | `MetadataUpdateCommand`, `UpdateDayListingsCommand`, `FixMissingSignatures` |                                                                                                                                              
  | HBAR Supply | 50B | 50B (confirm) | `WrappedBlockValidator` |                                                                                                                                                                                             
  | Address Book | `mainnet-genesis-address-book.proto.bin` | TBD | `AddressBookRegistry` |
  | Block 0 Timestamp | `1568411631.396440000` | TBD | `MainnetBlockZeroState` |
  | Record Path | `recordstreams/record0.0.X/` | Same pattern (confirm) | `MainNetBucket`, `DownloadDayImplV2` |

 ### Approach
  - Create a central `NetworkConfig` abstraction that provides all network-specific values based on a `--network` flag
  - Thread `NetworkConfig` through existing commands so they pick up the right defaults
  - Provide testnet genesis resources (address book, block zero state, balances) as bundled classpath resources
  - Rename `MainNetBucket.java` → `NetworkBucket.java` to reflect multi-network support

  ### Open Questions
  - [ ] What is testnet's genesis date and block 0 timestamp?
  - [ ] Does testnet have the same 50B HBAR total supply?
[...]

PR #2288, opened 2026-02-27
feat: add explicit testnet case to AmendmentProvider

### LINKED ISSUE #2270: Amendment Provider for Testnet
### Story Form

**As a** tools operator wrapping testnet blocks,
  **I want to** have testnet-specific amendments applied at the correct block numbers/dates,
  **So that** wrapped block conversion produces correct output matching testnet's actual transaction processing history.

### Technical Notes
  - `AmendmentProvider.createAmendmentProvider()` currently switches on `"mainnet"` / `"none"`
  - `MainnetAmendmentProvider.java` contains mainnet-specific amendment timelines (fee schedule changes, HIP activations)
  - Determine if testnet amendments differ from mainnet (likely different activation block numbers)
  - If they differ: create `TestnetAmendmentProvider.java` and add `case "testnet"` to the factory switch
  - If identical: confirm and document that `"mainnet"` provider works for both
  - The `--network` option on `ToWrappedBlocksCommand` already exists and passes through to the factory

There is no amendments but need to create the provider class for testnet.

rockysingh:
- Add an explicit `"testnet"` case to `AmendmentProvider.createAmendmentProvider()` so testnet no longer falls through to the default branch with a warning log                                                                                            
   - Testnet had a clean start in Feb 2024 and requires no amendments, so it uses a `NoOpAmendmentProvider`                                                                                                                                                   
   - Add unit tests for the `createAmendmentProvider` factory method covering testnet, mainnet, none, and unknown network cases.

   Closes #2270

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/AmendmentProvider.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/AmendmentProvider.java
@@ -75,6 +75,7 @@ public interface AmendmentProvider {
     static AmendmentProvider createAmendmentProvider(String network) {
         return switch (network.toLowerCase()) {
             case "mainnet" -> new MainnetAmendmentProvider();
+            case "testnet" -> new NoOpAmendmentProvider("testnet");
             case "none", "disabled" -> new NoOpAmendmentProvider();
             default -> {
                 System.out.println("No specific amendments for network: " + network + ", using no-op provider");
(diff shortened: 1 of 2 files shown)


AlfredoG87 (approved):

rockysingh:
@jasperpotts Thanks for the review! Created #2310 to track the minor test improvements as a follow-up subtask under the testnet epic #2267.

AlfredoG87 (approved):

Issue #2154, opened 2026-02-06
download-live2: Blocks written without signatures when GCS listing is stale


rockysingh:
```                                                                                                                                                                                                                                                                           
   Scenario: GCS listing is stale and missing signature files
       Given a block exists in GCS at timestamp "2026-02-03T00_00_00.261123000Z"
       And the GCS listing includes the .rcd file
       But the GCS listing does NOT include any .rcd_sig files
       When download-live2 processes this block
       Then it should fetch signature files directly from GCS for all nodes (0.0.3 to 0.0.37)
       And it should add the fetched signatures to the block
       And it should write the block with signatures to the archive
   ```

   ```
   Scenario: Signature files not yet available in GCS
       Given a block exists in GCS at timestamp "2026-02-03T00_00_00.261123000Z"
       And the GCS listing does NOT include any .rcd_sig files
       And direct GCS fetch also returns no signature files
       When download-live2 processes this block
       Then it should wait 5 minutes
       And retry fetching signatures from GCS
       And continue retrying indefinitely until signatures are found
       And never write a block without signatures
   ```

   ```
   Scenario: Corrupted download detected
       Given a block is being downloaded
       And the downloaded data is corrupted (e.g., "Unsupported record format version: 177996554")
       When download-live2 attempts to parse the block
       Then it should detect the corruption
       And wait with exponential backoff (30s, 60s, 120s, up to 5 minutes)
       And re-download all files for the block
       And retry indefinitely until successful
   ```

   ### Technical Notes

   The GCS listing API has eventual consistency - signature files uploaded by consensus nodes may not appear in listings immediately. The download process uses listings to determine which files to fetch, so missing listings = missing signatures.

   ## Related Bug: State File Hash Mismatch

   When archives are deleted and re-downloaded, the state file contains the hash from the END of the deleted day, not the BEGINNING. This causes hash validation issues.

   ## Affected Components

   - `DownloadLive2.java` - live mode processing
   - `DownloadDayImplV2.java` - bulk/catch-up mode processing

   ## Fix Status

   - [x] Signature fetching from GCS when listing is stale
   - [x] Infinite retry with 5-minute delay for missing signatures
   - [x] Corruption detection with exponential backoff retry
   - [ ] State file hash mismatch when re-downloading deleted archives"

PR #2300, opened 2026-02-28
feat(tools): add fix-missing-records command for repairing day archives

rockysingh:
## Summary

- Adds new `fix-missing-records` CLI command that repairs `tar.zstd` day archives where primary record files failed to download from GCP
- Promotes node-specific record file copies (e.g., `_node_0.0.3.rcd`) as the primary when the original `.rcd` is missing — fixes crashes in `validate-with-stats` and `blocks wrap` commands
- Adds `listRecordFilesForDay()` to `MainNetBucket` for querying which nodes have record files for each block in a given day

Fixes some of the issues described in #2154

## Test plan

- [x] `./gradlew :tools:compileJava` — compiles successfully
- [x] `./gradlew :tools:test --tests "org.hiero.block.tools.days.subcommands.FixMissingRecordsTest"` — all 5 tests pass
- [x] `./gradlew spotlessApply` — formatting clean
- [ ] Manual test: run against a broken day archive with `--single-day` flag

### DIFF
--- a/tools-and-tests/tools/src/test/java/org/hiero/block/tools/days/subcommands/FixMissingRecordsTest.java
+++ b/tools-and-tests/tools/src/test/java/org/hiero/block/tools/days/subcommands/FixMissingRecordsTest.java
@@ -1,0 +1,110 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.tools.days.subcommands;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import org.hiero.block.tools.records.model.unparsed.InMemoryFile;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import picocli.CommandLine;
+
+/**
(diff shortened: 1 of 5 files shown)


tools-and-tests/tools/src/test/java/org/hiero/block/tools/days/subcommands/FixMissingRecordsTest.java
```diff
@@ -0,0 +1,110 @@
...
+            bytes[0] = 0;
+            bytes[1] = 0;
+            bytes[2] = 0;
+            bytes[3] = 6; // version 6
+            return bytes;
+        }
+
+        @Test
+        @DisplayName("TarZstdDayReaderUsingExec should promote node copy when primary is missing")
+        void promotesNodeCopyWhenPrimaryMissing() {
```
jasperpotts: Weak test coverage for the main feature

  The promotesNodeCopyWhenPrimaryMissing test (the only test of the promotion behavior) doesn't test anything:

  void promotesNodeCopyWhenPrimaryMissing() {
      // ... builds InMemoryFile list ...
      FixMissingRecords command = new FixMissingRecords();
      assertNotNull(command, "FixMissingRecords command should be instantiable");
  }

  This just asserts the command can be instantiated. The actual promotion logic in TarZstdDayReaderUsingExec.processDirectoryFiles is completely untested by this PR. The test should use TarZstdDayReaderUsingExec.streamTarZstd() on a synthetic/fixture archive with a missing primary (similar to existing tests in the test suite using 2019-09-14.tar.zstd).

PR #2301, opened 2026-02-28
fix: plugins PVC now follows the same pattern as other PVCs

### LINKED ISSUE #2308: bug(charts): New version of BN (0.28.0) Introduced a new volume: `plugins-storage` but is not compatible with solo-provisioner
### Description

The new volume and PVC did not follow the same pattern of other PVC hence is failing and is not compatible when installed using solo-provisioner.

### Steps to Reproduce

Install the BN 0.28.0 using solo-provisioner

### Additional context

_No response_

AlfredoG87:
- Add existingClaim support for the plugins volume so Weaver and similar deployments can use pre-created PVCs
- Add storageClass support for the plugins volume
- Rename default volumeName from "plugins" to "plugins-storage" for naming consistency
- Move plugins storage config to blockNode.persistence.plugins to match the pattern of archive, live, logging, and verification volumes
- Add plugins.thirdParty field for third-party Maven plugins with full groupId:artifactId:version coordinates
- Move hardcoded Maven repositories from the init container script into configurable plugins.repositories in values.yaml
- Include thirdParty and repositories in the resolve-plugins cache hash so changes trigger re-resolution
- Use manifest-based cleanup in resolve-plugins so custom non-Maven JARs are preserved across config changes
- Rename plugin profile overrides for clarity: all-plugins.yaml -> plugin-profile-all.yaml, lfh.yaml -> plugin-profile-lfh.yaml, rfh.yaml -> plugin-profile-rfh.yaml, minimal.yaml -> plugin-profile-minimal.yaml
- Fix README to match actual plugin-profile-minimal.yaml content (4 plugins, not 9)

Fixes #2308


charts/block-node-server/values.yaml
```diff
@@ -95,11 +95,27 @@ tolerations: []
...
+  #   - "com.example:my-custom-plugin:1.0.0"
+  # Maven repositories used to resolve plugins.
+  # Override this list in your values to add custom or private repositories.
+  repositories:
+    - id: central
+      url: https://repo1.maven.org/maven2
+    - id: besu-maven
+      url: https://hyperledger.jfrog.io/artifactory/besu-maven
+    - id: consensys
+      url: https://artifacts.consensys.net/public/maven/maven/
```
jsync-swirlds: We probably should not include these two.

PR #2305, opened 2026-03-02
feat(verification): TSS Integration on verification path

### LINKED ISSUE #1659: Incorporate the TSS library into the Block Node Build
### Story Form

As a Block Node consumer
I want to have a self-contained verifiable proof
So that I can verify blocks without exetrnal references

### Technical Notes

The TSS library provided by the consensus-node foundation offers mechanisms to both create and verify TSS signatures.  We need to incorporate the TSS library into the block node build for verification purposes.

Note, This only covers adding the library to the build and ensuring it does not negatively impact other functionality.

### LINKED ISSUE #1660: Add TSS signature verification to the Verification plugin.
### Story Form

As a Block Node consumer
I want to have a self-contained verifiable proof
So that I can verify blocks without external references

### Technical Notes

With the TSS verification library incorporated in the build, the next step is to enable TSS signature verification using that library.

This relies on having a version of consensus node that generates TSS signatures and testing environment where TSS signatures can be generated and verified.

### LINKED ISSUE #2008: Implement TSS signature Verification
### Story Form

As a BlockNode Dev/Operator
I want to BN to actually verify block signatures
So that uses the TSS signature to verify the signatures.

### Technical Notes

This requires the CN to produce a TSS Signed blockstream.
This might require the Simulator to also produce TSS signed blocks. (ideally, but a stretch, maybe follow-up)

AlfredoG87:
## Context

Starting with HAPI 0.72.0, Consensus Nodes sign block proofs using TSS (Threshold Signature Scheme) instead of the legacy  hash of hash signature. Verifying these proofs requires three components: a **ledger ID**, an **address book** (node public keys, weights, and IDs), and a **WRAPS verification key**. All three are published together in block 0 via a `LedgerIdPublicationTransactionBody` embedded in a signed transaction.

This PR wires that bootstrap into the live verification path so that block nodes can verify TSS-signed blocks end-to-end.

## How It Works

**Block 0 bootstrap:** When the session processes block 0, it scans `SIGNED_TRANSACTION` items for a `LedgerIdPublicationTransactionBody`. When found, it calls `VerificationServicePlugin.initializeTssParameters()`, which sets the native TSS state (address book, WRAPS VK) and stores the publication on the plugin. If block 0 verification then succeeds, the full publication is persisted to disk as a serialized protobuf.

**Restart recovery:** On startup, if the TSS parameters file exists, the plugin loads and restores the full publication — ledger ID, address book, and WRAPS VK — before any blocks are processed.

**Retry semantics:** If block 0 fails verification (e.g., tampered items), TSS parameters are not locked in. Any subsequent block 0 can overwrite them and try again. Once parameters are persisted after a successful verification, they are locked — `initializeTssParameters()` becomes a no-op, preventing a corrupt block 0 from poisoning verified state.

**Session design:** Each session is short-lived (one per block) and receives the current `ledgerId` via constructor. The plugin — which owns process-level state — is the right home for TSS parameters.

## Key Changes

- `VerificationServicePlugin` owns TSS state as `public static` fields (`activeLedgerId`, `activeTssPublication`, `tssParametersPersisted`)
- `ExtendedMerkleTreeSession` receives `ledgerId` via constructor and calls `VerificationServicePlugin.initializeTssParameters()` when it discovers the publication in block 0
- Persist/restore the full `LedgerIdPublicationTransactionBody` protobuf instead of just raw ledger ID bytes (fixes incomplete state restoration on restart)
- Rename `verification.ledgerIdFilePath` → `verification.tssParametersFilePath` to reflect that it stores more than just the ledger ID
- Remove `verification.ledgerId` config property — a ledger ID alone is useless without the address book and WRAPS VK
- Fix backfill handler that was persisting TSS parameters for block 0 without checking `backfillNotification.success()`
- Add `hedera-cryptography-wraps` dependency for WRAPS VK initialization
- Add real TssWraps block fixtures (block 0 and block 50) with unit and integration tests covering the full bootstrap-to-verify flow

## Follow-up

TSS state currently lives as static fields on the plugin as a pragmatic starting point (the native TSS library is itself a singleton). Moving it into `BlockNodeContext` so other plugins can access it is tracked in #2321.

Fixes #1660
Fixes #1659
Fixes #2008

### DIFF
--- a/block-node/verification/src/main/java/module-info.java
+++ b/block-node/verification/src/main/java/module-info.java
@@ -14,11 +14,11 @@ module org.hiero.block.node.verification {
     requires transitive com.hedera.pbj.runtime;
     requires transitive com.swirlds.config.api;
     requires transitive org.hiero.block.node.spi;
+    requires transitive org.hiero.block.protobuf.pbj;
     requires com.hedera.cryptography.wraps;
     requires com.swirlds.metrics.api;
     requires org.hiero.block.common;
     requires org.hiero.block.node.base;
-    requires transitive org.hiero.block.protobuf.pbj;
     requires com.github.spotbugs.annotations;
     requires org.antlr.antlr4.runtime;
 


block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
```diff
@@ -98,6 +102,23 @@ public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
         // setting config and context
         this.context = context;
         verificationConfig = context.configuration().getConfigData(VerificationConfig.class);
+        // Bootstrap priority for pre-run startup: persisted file > ledgerId config string.
+        // Block 0 is always authoritative and overwrites both when received.
+        final var ledgerIdFile = verificationConfig.ledgerIdFilePath();
+        if (Files.exists(ledgerIdFile)) {
+            try {
+                ExtendedMerkleTreeSession.ACTIVE_LEDGER_ID.set(Bytes.wrap(Files.readAllBytes(ledgerIdFile)));
```
jsync-swirlds: We're storing and loading the ledger ID, but while we process the other TSS parameters we neither store nor load them.
We really should store and load all of the relevant parameters (which means we probably need to store those in the plugin, instead of _just_ setting them to the library when we first see them.

docs/design/tss-block-proof-verification.md
```diff
@@ -0,0 +1,48 @@
...
+dispatch between the genesis Schnorr path and the settled WRAPS path internally.
+
+## Ledger ID Bootstrap
+
+`ACTIVE_LEDGER_ID` is process-level state initialized from one of three sources:
+
+1. **Persisted file** (`verification.ledgerIdFilePath`) — written by block 0, loaded on restart.
+2. **Config string** (`verification.ledgerId`) — runtime-only seed for nodes joining a network
+   mid-stream after block 0 has already passed. Never persisted.
+3. **Block 0** — always authoritative. Overwrites both in-memory state and the persisted file.
```
jsync-swirlds: This should not overwrite anything; it should only be accepted if we don't have any values yet.  Block `0` is only authoritative the first time it's seen (otherwise it's possible a corrupted block could break things).

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -102,7 +123,15 @@ public VerificationNotification processBlockItems(BlockItems blockItemsMessage)
...
                 }
                 case ROUND_HEADER, EVENT_HEADER -> consensusHeaderHasher.addLeaf(getBlockItemHash(item));
-                case SIGNED_TRANSACTION -> inputTreeHasher.addLeaf(getBlockItemHash(item));
+                case SIGNED_TRANSACTION -> {
+                    inputTreeHasher.addLeaf(getBlockItemHash(item));
+                    // LedgerIdPublicationTransactionBody is only present in block 0; stop scanning
+                    // once found. Uses a per-session flag so pre-configured ACTIVE_LEDGER_ID does
+                    // not suppress block 0 from initializing the address book and WRAPS VK.
+                    if (blockNumber == 0 && !ledgerIdLoaded) {
+                        ledgerIdLoaded = loadLedgerId(item.signedTransaction());
```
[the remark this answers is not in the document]
AlfredoG87: I am positive, I already extracted it using a CN generated block with tss enabled.

block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationConfig.java
```diff
@@ -10,12 +10,21 @@
...
+ *     block 0 is processed. Takes priority over {@code ledgerId} on startup.
  */
 @ConfigData("verification")
 public record VerificationConfig(
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/verification/rootHashOfAllPreviousBlocks.bin") Path allBlocksHasherFilePath,
         @Loggable @ConfigProperty(defaultValue = "true") boolean allBlocksHasherEnabled,
-        @Loggable @ConfigProperty(defaultValue = "10") int allBlocksHasherPersistenceInterval) {}
+        @Loggable @ConfigProperty(defaultValue = "10") int allBlocksHasherPersistenceInterval,
+        @Loggable @ConfigProperty(defaultValue = "") String ledgerId,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/verification/ledger-id.bin") Path ledgerIdFilePath) {}
```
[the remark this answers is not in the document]
AlfredoG87: have created this follow-up issue so we can extract this out of verification and move it to the BN Context.

https://github.com/hiero-ledger/hiero-block-node/issues/2321

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -183,20 +212,64 @@ protected VerificationNotification getVerificationResult(
...
+    }
 
-        // Dummy implementation
-        // signature = is Hash384( BlockHash )
-        return signature.equals(noThrowSha384HashOf(hash));
+    // Parses a signed transaction looking for LedgerIdPublicationTransactionBody; when found,
+    // bootstraps TSS native state (address book + WRAPS VK) and sets ACTIVE_LEDGER_ID.
+    // Returns true if the publication was found and applied, false otherwise.
+    private static boolean loadLedgerId(Bytes signedTxBytes) throws ParseException {
+        if (signedTxBytes == null || signedTxBytes.length() == 0) {
```
Nana-EC: Q: Should we capture the block number and ensure that we're only ever setting assuming the block number is incrementing.
I'm curious about a backfill scenario that included the `LedgerIdPublicationTransactionBody` and overwrites the more recent values?

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -183,20 +212,64 @@ protected VerificationNotification getVerificationResult(
...
+        }
+    }
 
-        // Dummy implementation
-        // signature = is Hash384( BlockHash )
-        return signature.equals(noThrowSha384HashOf(hash));
+    // Parses a signed transaction looking for LedgerIdPublicationTransactionBody; when found,
+    // bootstraps TSS native state (address book + WRAPS VK) and sets ACTIVE_LEDGER_ID.
+    // Returns true if the publication was found and applied, false otherwise.
+    private static boolean loadLedgerId(Bytes signedTxBytes) throws ParseException {
```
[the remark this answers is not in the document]
AlfredoG87: ✅  Done. 
Renamed to initializeTssParameters() and moved it to VerificationServicePlugin as a public static method. The session calls VerificationServicePlugin.initializeTssParameters(publication) when it discovers the LedgerIdPublicationTransactionBody in block 0. The method sets the native TSS state       
  (address book, WRAPS VK) and updates the plugin's static fields (activeLedgerId, activeTssPublication).

block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
AlfredoG87: ✅  Done. 
We now persist and restore the full LedgerIdPublicationTransactionBody protobuf (ledger ID + address book + WRAPS VK) instead of raw ledger ID bytes. The config property was renamed from verification.ledgerIdFilePath to verification.tssParametersFilePath. On startup, the file is deserialized and
  all three components are restored via initializeTssParameters().

block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
```diff
@@ -275,6 +314,10 @@ public void handleBackfilled(BackfilledBlockNotification notification) {
                 if (backfillNotification != null) {
                     // Log the backfill verification result
                     LOGGER.log(TRACE, COMPLETED_MESSAGE, notification.blockNumber(), backfillNotification.success());
+                    // Block 0 is authoritative even when backfilled: persist its ledger ID to file.
+                    if (notification.blockNumber() == 0) {
```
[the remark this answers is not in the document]
AlfredoG87: ✅ Done. 
Added a tssParametersPersisted guard on initializeTssParameters(). TSS parameters can be overwritten by any block 0 until they are persisted (after successful verification or file bootstrap). Once persisted, initializeTssParameters() becomes a no-op. This allows retry when block 0 fails
  verification, while preventing a corrupted block 0 from overwriting already-verified parameters.

docs/design/tss-block-proof-verification.md
AlfredoG87: ✅ Done! 

The tssParametersPersisted flag prevents overwriting once TSS parameters have been persisted (either from file bootstrap or after successful block 0 verification). The design doc has been updated to describe this overwrite-until-persisted semantics.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: Good question. TSS initialization only triggers for block 0 (if (blockItemsMessage.blockNumber() == 0)), so backfill of other blocks won't attempt to overwrite. Additionally, the tssParametersPersisted guard on initializeTssParameters() prevents any overwrite once the parameters have been successfully
  verified and persisted. A backfilled block 0 would only initialize TSS parameters if they haven't been persisted yet, which is the correct behavior for a node that doesn't have block 0 locally.

PR #2319, opened 2026-03-04
feat: add in new k6 tests

### LINKED ISSUE #1963: `serverStatus` Calls do not Always Succeed
### Story Form

As a Block Node User
I want to reliably be able to call `serverStatus`
So that I can rely on the method

### Technical Notes

- in recent k6 test runs, we notice that 95 % of around 10k calls to serverStatus succeed
- we need to investigate why so, is that a k6 issue or not
- if not a k6 issue, we need to resolve the issue

Image

### LINKED ISSUE #2228: Add K6 Tests for serverStatusDetails
### Story Form

As a BlockNode Developer I would like to load test the serverStatusDetails BlockNod gRPC call so that I can tell that it works efficiently under load.

### Technical Notes

The serverStatus call no longer returns BlockNode versions information. The information has been moved to the newly created serverStatusDetails call. We should run similar k6 testing for serverStatusDetails.

### LINKED ISSUE #2338: Refactor K6 tests to use solo as the environment for testing Block Nodes
### Story Form

As a Block node engineer I want to be able to reliable test the block node server using an environment that simulates a production environment

### Technical Notes

Solo is the environment that we will be providing to community developers to allow them to test their code using real CN and BN servers. Refactoring the K6 Tests to use solo will allow us to test in a similar environment to community developers and hashgraph developers.

berryware:
## Add javascript spotless to the k6 tests
 - used the gradle plugin to accomplish this
 - added a gradle task to run the tests which depends on spotless
 - moved all the javascript to a src directory to make spotless configuration easier
 - added node setup to the github workflow as it is required for spotless
 - added the k6 tests project to gradle so that the spotless errors will get picked up on `./gradlew build`
##  Ignore but track StatusDeadlineExceeded errors
 - k6 reports StatusDeadlineExceeded errors indicating that the BlockNode api is not returning fast enough
 - The team has concluded that k6 is in error not the block node server
 - added a counter to count the number of times this occurs and report it in the report
## Fail the gradle task if the test metrics are not met
 - modified the bash scripts to track the errors and to return 1 if a failure is detected, or 0 if no error
## Remove duplicated code in the test scripts
 - the bash scripts had a lot of cut and paste code in it
 - created methods with the common code
 - pass test information in as parameters
## Add in a serverStatusDetail test
 - added in a test for the new api
## Use solo as the test environment for the k6 tests
 - remove calls to start the block node server and simulator
 - add calls to the solo-e2e-tests tasks (`task up`, `task down`)
 - use default config of 1 cn, 1bn, 1mn
 - make tests fail fast if solo `task up` fails

closes #1963 
closes #2228
closes #2338

### DIFF
--- a/.github/workflows/k6-tests.yml
+++ b/.github/workflows/k6-tests.yml
@@ -20,6 +20,7 @@ permissions:
 
 env:
   GRADLE_EXEC: ionice -c 2 -n 2 nice -n 19 ./gradlew --no-daemon
+  SOLO_VERSION: "0.63.0"
 
 jobs:
   k6-tests:
@@ -71,7 +72,7 @@ jobs:
       - name: Install Solo CLI via npm
         id: solo
         run: |
-          npm i @hashgraph/solo@latest -g
+          npm i @hashgraph/solo@${SOLO_VERSION} -g
           # Extract version number from solo output (handles various output formats)
           SOLO_INSTALLED=$(solo --version 2>/dev/null | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || echo "unknown")
           echo "Solo version: ${SOLO_INSTALLED}"
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -59,15 +59,17 @@ jobs:
     name: "Quality Checks and UTs"
     runs-on: hiero-block-node-linux-medium
     steps:
-      - name: Harden Runner
-        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+      - name: Setup Node, Java, checkout code and harden runner
+        uses: PandasWhoCode/initialize-github-job@a57cd6d8d768b2f3c95334bdd4fa8c21609fc651 # v1.0.5
         with:
-          egress-policy: audit
-
-      - name: Checkout code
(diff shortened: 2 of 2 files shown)


tools-and-tests/k6/average-load/bn-server-status-detail.js
```diff
@@ -0,0 +1,58 @@
...
+    const params = {
+        tags: {name: 'block_node_server_status_detail'}
+    };
+    let response = new ServerStatusRequest(client).invoke(params);
+    // ignore StatusDeadlineExceeded as it appears there is an issue with K6 grpc
+    if (response.status !== StatusDeadlineExceeded) {
+      check(response, {
+          'status is OK': (r) => r && r.status === StatusOK,
+      });
+      if (response.status !== StatusOK) {
```
Nana-EC: Q: why are we double checking the response?
It seems we expect `StatusOK` the first time but then not the second time but are curious of the failures the second time?

Nana-EC (commented):
Seems reasonable, but why the double error check?

tools-and-tests/k6/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: Apache-2.0
+import org.hiero.gradle.spotless.LicenseHeader
+
+plugins {
+    id("com.diffplug.spotless")
```
ata-nas: Is this the correct target for spotless? Not sure if we _should_ use one of the hiero plugins.

PR #2320, opened 2026-03-05
fix: add --day option to updateDayListings and remove duplicate --network flag

rockysingh:
## Summary
- Add `--day` option to `UpdateDayListingsCommand` to regenerate the listing for a specific day (e.g. `--day 2026-02-06`), deleting any existing stale listing first
- Remove duplicate `--network` option from `ToWrappedBlocksCommand` since it is now inherited from parent

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/UpdateDayListingsCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/UpdateDayListingsCommand.java
@@ -84,13 +84,56 @@ public class UpdateDayListingsCommand implements Runnable {
             description = "GCP project to bill for requester-pays bucket access (default: from GCP_PROJECT_ID env var)")
     private String userProject = DownloadConstants.GCP_PROJECT_ID;
 
+    @Option(
+            names = {"--day"},
+            description = "Regenerate listing for a specific day (YYYY-MM-DD). Deletes existing listing first.")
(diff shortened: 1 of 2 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/UpdateDayListingsCommand.java
```diff
@@ -84,13 +84,51 @@ public class UpdateDayListingsCommand implements Runnable {
...
+     *
+     * <p>When {@code --day} is specified, only that single day's listing is regenerated
+     * (any existing listing file is deleted first).
      */
     @Override
     public void run() {
-        updateDayListings(listingDir, cacheDir.toPath(), cacheEnabled, minNodeAccountId, maxNodeAccountId, userProject);
+        if (targetDay != null) {
+            // Delete existing listing file so it gets regenerated fresh
+            deleteExistingListing(listingDir, targetDay);
```
Nana-EC: If an error occurred while this ran would it be okay that the data was already deleted or would it be better for the existing data to remain?
You could flip the order by getting the new data, placing it in a staging area and replacing on success.

Nana-EC (approved):
LG
2 minor non blocking considerations

Nana-EC (approved):

jsync-swirlds (approved):
Question about the network option; Did you _intend_ to remove that?

PR #2342, opened 2026-03-10
fix(verification): allBlocksHasher not maintained correctly for first non-genesis block and backfill

### LINKED ISSUE #2341: Unexpected ignoring of BLOCK_NODE_EARLIEST_MANAGED_BLOCK
https://perf.analytics.eng.hashgraph.io/ephemeral/test_runs/main_AdHoc7_595/report/NftTransferLoadTest/podlog_solo-sdpt-n7/
Logs with evidence
@AlfredoG87 


Exact root cause of block 168 verification failure                                                                                         
                                                                                                                                             
  Initialization sequence (AllBlocksHasherHandler.init()):                                                                                   
                                                                                                                                           
  1. available.size() == 0 (fresh BN, no stored blocks) → initGenesis() is called
  2. initGenesis() sets lastBlockHash = ZERO_BLOCK_HASH (SHA384 of empty tree), leaves hasher.leafCount() == 0
  3. Back in initAllBlocksHasherIfEnabled(): isAvailable() == true AND lastBlockHash() != null → so previousBlockHash = ZERO_BLOCK_HASH is
  set on the plugin

  When block 168 arrives and a session is created:

  // VerificationServicePlugin.java:204-206
  currentSession = HapiVersionSessionFactory.createSession(
      currentBlockNumber,         // 168
      BlockSource.PUBLISHER,
      semanticVersion,
      previousBlockHash,          // ← ZERO_BLOCK_HASH (wrong! should be hash of block 167)
      getRootOfAllPreviousBlocks(), // ← ZERO_BLOCK_HASH (wrong! should be Merkle root of blocks 0–167)
      activeLedgerId);

  getRootOfAllPreviousBlocks() at line 264-270 just checks isAvailable() (which is true because hasher is non-null), then calls
  computeRootHash() which returns ZERO_BLOCK_HASH because leafCount == 0.

  In finalizeVerification() (lines 159-164):

  // both are non-null (ZERO_BLOCK_HASH), so footer values are completely bypassed
  Bytes rootOfAllPreviousBlockHashes = rootHashOfAllBlockHashesTree; // ZERO_BLOCK_HASH
  Bytes previousBlockHashToUse = previousBlockHash;                  // ZERO_BLOCK_HASH

  These wrong values are fed into computeFinalBlockHash() → the computed block hash doesn't match what CN signed → verifySignature() returns
  false.

AlfredoG87:
## Bug 1 — genesis-state ZERO_BLOCK_HASH overrides footer for mid-chain first block

On a fresh BN with `earliestManagedBlock > 0`, `AllBlocksHasherHandler` starts
in genesis state (`leafCount=0`, `lastBlockHash=ZERO_BLOCK_HASH`). Both values
were non-null so they bypassed the block footer's authoritative
`previousBlockRootHash` and `rootHashOfAllBlockHashesTree` in the session,
producing a wrong block root hash and a signature failure for every first
received block — triggering an unrecoverable `NODE_BEHIND_PUBLISHER` loop.

**Fix:** when `earliestManagedBlock > 0`:
- `previousBlockHash` is only seeded from the hasher when `leafCount > 0`
- `getRootOfAllPreviousBlocks` returns `null` (footer fallback) when `leafCount != currentBlockNumber`

When `earliestManagedBlock == 0` full genesis continuity is assumed; hasher
remains authoritative.

## Bug 2 — allBlocksHasher not updated on backfill

`handleBackfilled` never called `appendLatestHashToAllPreviousBlocksStreamingHasher`,
so live-tail backfill blocks were silently dropped from the hasher, breaking
continuity for all future verifications.

**Fix:** after successful backfill verification, append the block hash to the
hasher only when `leafCount == blockNumber` (sequential order). Out-of-order
historical backfill is skipped to preserve the contiguous chain from genesis.

## Tests added

- Fresh BN with `earliestManagedBlock=1` receives block 1 first → success (regression for bug 1)
- Sequential backfill of block 0 → `leafCount` increments to 1
- Out-of-order backfill of block 21 → `leafCount` stays 0

Fixes #2341

### DIFF
--- a/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
+++ b/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
@@ -63,7 +63,7 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
     /** The previous block hash, used for verification of the current block. */
     private Bytes previousBlockHash;
     /** Handler for root hash for all previous blocks hasher operations and lifecycle. */
-    private AllBlocksHasherHandler allBlocksHasherHandler;
+    AllBlocksHasherHandler allBlocksHasherHandler;
     /**
      * The earliest block number this node is configured to manage. When greater than zero the node
      * is not expected to have a continuous chain from genesis, so allBlocksHasher values must not
@@ -118,7 +118,8 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
         // setting config and context
         this.context = context;
         verificationConfig = context.configuration().getConfigData(VerificationConfig.class);
-        earliestManagedBlock = context.configuration().getConfigData(NodeConfig.class).earliestManagedBlock();
+        earliestManagedBlock =
+                context.configuration().getConfigData(NodeConfig.class).earliestManagedBlock();
         // Bootstrap TSS parameters from persisted file if available. The file contains a
         // serialized LedgerIdPublicationTransactionBody with ledger ID, address book, and WRAPS VK.
         final var tssParametersFile = verificationConfig.tssParametersFilePath();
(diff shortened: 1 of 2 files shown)


jsync-swirlds (approved):

Nana-EC (approved):

rockysingh (commented):
There's no test for the scenario where `earliestManagedBlock == 0` and the hasher has leafCount == 0 (fresh genesis start). The fix changes the condition, so verifying the genesis case still works with the new logic would be valuable. The existing testVerificationPlugin test covers block 0 + block 1 on a fresh start, but `earliestManagedBlock` defaults to 0, so it implicitly covers this, just worth calling out that it's relying on the default.

PR #2350, opened 2026-03-12
feat: Add testnet genesis state & address book

### LINKED ISSUE #2269: Testnet Genesis State & Address Book
### Story Form

**As a** tools operator running against testnet,
  **I want to** have testnet genesis resources (address book, block zero state, initial balances) bundled in the tools JAR,
  **So that** commands like `blocks wrap` and `blocks validate-wrapped` can initialize correctly from testnet's block 0.

### Technical Notes
  - Obtain testnet genesis address book and add as `testnet-genesis-address-book.proto.bin` in `src/main/resources/`
  - Create `TestnetBlockZeroState.java` mirroring `MainnetBlockZeroState.java` with testnet-specific:
    - Block 0 consensus timestamp
    - Transaction 1 ID
    - Genesis balances CSV timestamp
  - Update `AddressBookRegistry.java` to load the correct genesis address book based on network (currently hardcoded to `mainnet-genesis-address-book.proto.bin`)
  - Generate testnet genesis balances file (equivalent of `2019-09-13T22_00_00.000081Z_Balances.csv.gz`)
  - **Open question:** Need to determine testnet's genesis date, block 0 timestamp, and initial saved state snapshot

rockysingh:
Fixes #2269

## Summary

The tools JAR currently only bundles mainnet genesis resources. When running `blocks wrap` or `blocks validate` against testnet, `AddressBookRegistry` tries to load `testnet-genesis-address-book.proto.bin` from the classpath — which doesn't exist causing a crash. This PR adds testnet genesis resources so the tools work for testnet.

Testnet was reset in February 2024. Genesis date: `2024-02-01`, first block timestamp: `2024-02-01T18:35:20.644859297Z`, nodes: `0.0.3` through `0.0.9` (7 nodes).

### New files
- **`testnet-genesis-address-book.proto.bin`** — Binary protobuf `NodeAddressBook` with 7 testnet nodes, fetched from the testnet mirror node API
- **`TestnetBlockZeroState`** — Fetches testnet genesis account balances from the mirror node API at genesis timestamp and converts to `BlockItem` STATE_CHANGES
- **`TestnetAmendmentProvider`** — Wires testnet genesis STATE_CHANGES into the wrap pipeline for block 0
- **`GenerateTestnetGenesisAddressBook`** — CLI command (`mirror generateTestnetAddressBook`) to regenerate the address book proto binary from the testnet mirror node

### Modified files
- **`AmendmentProvider`** — Factory method now returns `TestnetAmendmentProvider` instead of `NoOpAmendmentProvider` for testnet
- **`MirrorNodeCommand`** — Registered new `GenerateTestnetGenesisAddressBook` subcommand
- **`AmendmentProviderTest`** — Updated tests to expect `TestnetAmendmentProvider` for testnet

## Verification

- `NetworkConfig.testnet()` already had correct values — no changes needed
- `AddressBookRegistry` already loads from `NetworkConfig` — no changes needed


tools-and-tests/tools/src/main/java/org/hiero/block/tools/states/TestnetBlockZeroState.java
```diff
@@ -0,0 +1,186 @@
...
+ * testnet mirror node API at genesis time.
+ *
+ * <p>Testnet was reset in February 2024. Genesis date: 2024-02-01, first block timestamp:
+ * 2024-02-01T18:35:20.644859297Z, nodes: 0.0.3 through 0.0.9 (7 nodes).
+ *
+ * <p>Unlike mainnet (which requires a saved state snapshot and transaction reversal), testnet
+ * genesis state can be fetched directly from the mirror node API since the network was freshly
+ * reset at that point.
+ */
+public final class TestnetBlockZeroState {
```
jasperpotts: Is this right, feel like it is wrong as will mirror node give you state at before block zero on testnet. In theory I think state should be empty at begining of block zero. Though there might be some entities that are created at genesis that are never recorded in record stream.  We should be able to get answer from mirror node code base

tools-and-tests/tools/src/main/java/org/hiero/block/tools/mirrornode/GenerateTestnetGenesisAddressBook.java
```diff
@@ -0,0 +1,75 @@
...
+ * it as a {@link NodeAddressBook} protobuf binary file suitable for use as a classpath resource.
+ *
+ * <p>Usage example:
+ * <pre>
+ *   subcommands mirror generateTestnetAddressBook -o testnet-genesis-address-book.proto.bin
+ * </pre>
+ */
+@Command(
+        name = "generateTestnetAddressBook",
+        description = "Generate testnet genesis address book proto binary from mirror node API")
```
jasperpotts: This is only correct if the nodes never changed between genesis and now

tools-and-tests/tools/src/main/java/org/hiero/block/tools/states/TestnetBlockZeroState.java
rockysingh: You're right  the mirror node returns empty results at the exact genesis timestamp because balance snapshots are periodic. I've updated to use the first available snapshot (~72 min after genesis), which shows the expected genesis state: treasury (0.0.2) holding all HBAR, system accounts existing with 0 balance. This captures the genesis entities that exist before block 0's transactions.

rockysingh: Also, confirmed that the mirror node does have the genesis entities. The first available balance snapshot shows system accounts (0.0.1 through at least 0.0.750) existing with 0 balance, and the treasury (0.0.2) with ~5 quintillion tinybars. Updated to use this earliest available snapshot instead of the exact genesis timestamp.

Nana-EC (dismissed):
Please increase test coverage when you get a chance

rockysingh:
> Please increase test coverage when you get a chance

I've increased coverage.

Nana-EC (approved):

PR #2351, opened 2026-03-12
refactor: remove swirlds.common from test fixtures (and any where else it's used) separately

### LINKED ISSUE #1220: remove `swirlds.common` from test fixtures (and any where else it's used) separately;
Thanks for putting these together and updating the plugin.

Please add an issue to remove `swirlds.common` from test fixtures (and any where else it's used) separately; we need to keep that out of the codebase.

_Originally posted by @jsync-swirlds in https://github.com/hiero-ledger/hiero-block-node/pull/1211#pullrequestreview-2876726347_

ivannov:
## Reviewer Notes

This PR removes the dependency on `com.swirlds:swirlds-common` and migrates the entire codebase to the new `org.hiero.metrics` API (`com.hedera.hashgraph:hiero-metrics`). It touches all plugins, the app, the simulator, and the SPI.

### New metrics API at a glance

| Old (`swirlds-metrics`) | New (`org.hiero.metrics`) |
|---|---|
| `Metrics.getOrCreate(new Counter.Config(...))` | `metricRegistry.register(LongCounter.builder(MetricKey.of(...)).setDescription(...)).getOrCreateNotLabeled()` |
| `LongGauge` / `DoubleGauge` — push: call `.set()` periodically | `ObservableGauge` — pull: supply a lambda once at registration |
| `metrics.addUpdater(this::updateMetrics)` — periodic callback | Removed entirely; lambdas are invoked by the exporter on demand |
| `DefaultMetricsProvider.start()` / `.stop()` — manual lifecycle | `MetricRegistry.builder().discoverMetricProviders().discoverMetricsExporter(config).build()` — SPI, no manual lifecycle |
| `metricsProvider.createGlobalMetrics()` returns `Metrics` | `MetricRegistry` is the single object passed around |

### `BlockNodeContext`

The `metrics` record component (`Metrics`) has been renamed to `metricRegistry` (`MetricRegistry`). Any plugin or test that constructs a `BlockNodeContext` directly was updated.

### `ObservableGauge` replaces push-based gauges

Wherever the old code stored a `LongGauge` field and called `.set(value)` inside an `addUpdater` callback, the new code registers an `ObservableGauge` with a lambda at startup and never touches it again

`ObservableGauge.observe()` accepts both `LongSupplier` and `DoubleSupplier`, so double-valued gauges (e.g. `queue-percent-used` in `BlockMessagingFacilityImpl`) retain full precision.

### Simulator: `MetricsServiceImpl` now owns the exporter

The simulator previously received a `Metrics` instance via Dagger injection (`MetricsInjectionModule` provided it). Now `MetricsServiceImpl` constructs and owns the `MetricRegistry` directly in its constructor, implementing `MetricsExporter` itself to intercept the snapshot supplier and forward it to the Prometheus HTTP exporter discovered via SPI.

### Configuration cleanup

`MetricsConfig` and `PrometheusConfig` no longer need to be explicitly registered with `ConfigurationBuilder` — they are auto-discovered via `.autoDiscoverExtensions()` in the block node, or resolved inside the new metrics library in the simulator.

### Test infrastructure

A shared `TestMetricsExporter` class was added to app testFixtures (`org.hiero.block.node.app.fixtures.TestMetricsExporter`). Pass it to `MetricRegistry.builder().setMetricsExporter(...)` in a test setup, then call `getMetricValue("category:metric_name")` to read current values. PluginTestBase exposes `getMetricValue()` directly, so existing plugin tests need no changes beyond the `BlockNodeContext` rename.

## Related Issue(s)

Closes #1220


block-node/app-config/src/main/resources/app.properties
```diff
@@ -1,3 +1,3 @@
 # SPDX-License-Identifier: Apache-2.0
-
-prometheus.endpointPortNumber=16007
+metrics.exporter.openmetrics.http.hostname=0.0.0.0
```
akugal: Why does default `metrics.exporter.openmetrics.http.hostname=localhost` not work? 

If there is a problem when running locally with Docker Compose, try adding the following to `prometheus` service:

```
    extra_hosts:
      - "host.docker.internal:host-gateway"
```

block-node/app/src/main/java/org/hiero/block/node/app/logging/ConfigLogger.java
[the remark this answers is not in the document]
jsync-swirlds: Perhaps.  That's more a Consensus Node concern than Block Node.
For Block Node (because plugins are not all in our codebase and might not use our config library) we need another mechanism.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/TestMetricsExporter.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * Returns the current long value of the named metric from the registry snapshot.
+     *
+     * @param metricName fully-qualified metric name
+     * @return the current value
+     * @throws IllegalArgumentException if the metric is not found
+     */
+    public long getMetricValue(@NonNull final String metricName) {
```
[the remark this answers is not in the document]
akugal: Also, see if you can use `MetricKey`, which already incorporates the category into the name.
You just need to have `public static final MetricKey` fields somewhere to reuse when creating a metric and when accessing it.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -501,41 +492,65 @@ public void accept(GapDetector.Gap gap) {
...
      */
     public record MetricsHolder(
-            Counter backfillGapsDetected,
-            Counter backfillFetchedBlocks,
-            Counter backfillBlocksBackfilled,
-            Counter backfillFetchErrors,
-            Counter backfillRetries,
-            LongGauge backfillStatus,
-            LongGauge backfillPendingBlocksGauge,
-            LongGauge backfillInFlightGauge) {
```
jsync-swirlds: What happened to these?

jsync-swirlds (commented):
Excellent work, I think we lost some gauge metrics in the switch, so we need to restore those.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
ivannov: On `ObservableGauve` vs `LongGauge`.

In the old metrics, we used to do:

```
metrics.addUpdater(this::updateAppMetrics);
   
   private void updateAppMetrics() {
        long minBlock = historicalBlockFacility.availableBlocks().min();
        long maxBlock = historicalBlockFacility.availableBlocks().max();
        long currentState = state.get().ordinal();

        historicalOldestBlockGauge.set(minBlock);
        historicalLatestBlockGauge.set(maxBlock);
        appStateStatus.set(currentState);
    }
```

Which is nearly the same: assigning a lambda with all the gauges. Now the difference is that we assign a lambda per gauge.

But your point on who controls the update of the gauges is not changed. We were not in control with the old library and are not with the new one.

gradle/modules.properties
```diff
@@ -1,4 +1,11 @@
 # SPDX-License-Identifier: Apache-2.0
+# Swirlds/Hiero modules - group ID migrated from com.swirlds to com.hedera.hashgraph at 0.63+
+com.swirlds.base=com.hedera.hashgraph:swirlds-base
+com.swirlds.config.api=com.hedera.hashgraph:swirlds-config-api
+com.swirlds.config.extensions=com.hedera.hashgraph:swirlds-config-extensions
+com.swirlds.config.impl=com.hedera.hashgraph:swirlds-config-impl
+org.hiero.metrics=com.hedera.hashgraph:hiero-metrics
+org.hiero.metrics.openmetrics.httpserver=com.hedera.hashgraph:openmetrics-httpserver
```
ata-nas: Interesting what does this mapping do? Will we have to revisit this?

jsync-swirlds: Ugh, it would be much easier if Consensus Node could migrate directly to hiero package and module names instead of jittering around non-hiero names.

hiero-dependency-versions/build.gradle.kts
```diff
@@ -35,8 +35,20 @@ dependencies.constraints {
...
+    api("com.hedera.hashgraph:swirlds-base:$hederaVersion") { because("com.swirlds.base") }
+    api("com.hedera.hashgraph:swirlds-config-api:$hederaVersion") {
+        because("com.swirlds.config.api")
+    }
+    api("com.hedera.hashgraph:swirlds-config-extensions:$hederaVersion") {
+        because("com.swirlds.config.extensions")
+    }
+    api("com.hedera.hashgraph:swirlds-config-impl:$hederaVersion") {
+        because("com.swirlds.config.impl")
+    }
```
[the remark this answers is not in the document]
jsync-swirlds: Eventually we'll move everything to `hiero`, but it's very hard to find time in schedules for a set of PRs that affect around 1,000 files in aggregate (not to mention community impact in SDKs and other projects).

Consensus Node repo hasn't made the change completely yet as a result, and some modules are jittering naming between the even older `hedera.hashgraph`, the old `swirlds`, the intermediate `hashgraph`, and the current `hiero` conventions.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -319,8 +306,6 @@ public void shutdown(String className, String reason) {
             LOGGER.log(INFO, "Stopping plugin: {0}", plugin.name());
             plugin.stop();
         }
-        // Stop metrics
-        if (metricsProvider != null) metricsProvider.stop();
```
AlfredoG87: does the new metrics server is autocloseable?

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/grpc/impl/PublishStreamGrpcClientImpl.java
```diff
@@ -132,11 +132,8 @@ public boolean streamBlock(
                                 .addAllBlockItems(streamingBatch)
                                 .build())
                         .build());
-                metricsService.get(LiveBlockItemsSent).add(streamingBatch.size());
-                LOGGER.log(
-                        DEBUG,
-                        "Number of block items sent: "
-                                + metricsService.get(LiveBlockItemsSent).get());
+                metricsService.get(LiveBlockItemsSent).increment(streamingBatch.size());
+                LOGGER.log(DEBUG, "Number of block items sent: " + metricsService.getValue(LiveBlockItemsSent));
```
AlfredoG87: We should avoid this string concatenation in the project, in favor of
```suggestion
  LOGGER.log(DEBUG, "Number of block items sent: {0}", metricsService.getValue(LiveBlockItemsSent));                                                                                                                                                                                                              

```

tools-and-tests/simulator/src/test/java/org/hiero/block/simulator/config/logging/SimulatorConfigurationLoggerTest.java
```diff
@@ -42,7 +43,7 @@ void testWithMockedSensitiveProperty() throws IOException {
         final SimulatorConfigurationLogger configurationLogging = new SimulatorConfigurationLogger(configuration);
         final Map<String, Object> config = configurationLogging.collectConfig(configuration);
         assertNotNull(config);
-        assertEquals(53, config.size());
+        assertFalse(config.isEmpty());
```
AlfredoG87: same as above:
why the removal of exact count? does this affect the test assertion purpose?

AlfredoG87 (approved):
Mostly nits and a couple of questions, nothing blocking. overall a great job!! 💯

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: It is autoclosable, but we can't use that feature. So I made a change now to close it explicitly. Thanks for noticing

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/metrics/MetricsServiceImpl.java
```diff
@@ -44,7 +117,31 @@ public MetricsServiceImpl(@NonNull final Metrics metrics) {
...
+     *
+     * @throws IllegalArgumentException if the metric is not found in the current snapshot
+     */
+    @Override
+    public long getValue(@NonNull SimulatorMetricTypes.Counter key) {
+        Objects.requireNonNull(snapshotSupplier, "MetricRegistry has not called setSnapshotSupplier yet");
+        final String metricName = CATEGORY + ":" + key.grafanaLabel();
+        for (MetricSnapshot snapshot : snapshotSupplier.get()) {
+            if (snapshot.name().equals(metricName)) {
+                for (var measurement : snapshot) {
```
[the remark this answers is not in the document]
ivannov: Changed

AlfredoG87 (approved):
Great Work! 💯 

Can't wait to start using labels on metrics 🙏

PR #2374, opened 2026-03-16
feat: upgrade PBJ to 0.14.2 and make maxIncomingBufferSize configurable

AlfredoG87:
## Summary

- Upgrade PBJ from 0.14.1 to 0.14.2, which fixes `PbjGrpcDatagramReader` to respect the configured `maxSize` instead of using a hardcoded 10 MB buffer limit
- Re-enable the TSS Wraps Block (1319) backfill test that was disabled pending the PBJ fix
- Makes configurable the max PBJ incoming datagram size, default to 100MB, with min of 10 and Max of 300.
  - made it 100, since even when the single largest item_block size is ~30MB, the block in questions is about 100MB, since it has 3 times the same data, what makes me believe that the client will need to support a buffer of at least 100MB.

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
@@ -186,7 +186,9 @@ public class BackfillFetcher implements PriorityHealthBasedStrategy.NodeHealthPr
             LOGGER.log(DEBUG, "Removed unreachable client for node [{0}], will attempt to recreate", node.address());
         }
         return nodeClientMap.computeIfAbsent(
-                node, n -> new BlockNodeClient(n, globalGrpcTimeoutMs, enableTls, maxIncomingBufferSize, n.grpcWebclientTuning()));
+                node,
+                n -> new BlockNodeClient(
+                        n, globalGrpcTimeoutMs, enableTls, maxIncomingBufferSize, n.grpcWebclientTuning()));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jsync-swirlds:
Have we looked into why the 30MB WRAPS transaction is in the block 3 times?

jsync-swirlds (approved):

AlfredoG87:
> Have we looked into why the 30MB WRAPS transaction is in the block 3 times?

YES!


## Block 488 Breakdown (131 items)

| Index | Item Type         | Size    | Description                                              |
|-------|-------------------|---------|----------------------------------------------------------|
| [47]  | signedTransaction | 30.1 MB | `HistoryProofKeyPublicationTransactionBody.wraps_message` |
| [49]  | stateChanges      | 30.1 MB | State change recording the transaction (stateId=47, mapUpdate) |
| [127] | stateChanges      | 30.1 MB | Round-end TSS state sync (stateId=44, singletonUpdate)   |
| other | (128 items)       | ~9.1 MB | Normal block items (headers, rounds, events, proof, etc.) |

The ~30 MB WRAPS key material appears **3 times** in the block:

1. **Transaction body** `[47]`: the original `HistoryProofKeyPublicationTransactionBody` system transaction containing the WRAPS key shares
2. **Transaction state change** `[49]`: `mapUpdate` on stateId 47 — the state written by this transaction's handler
3. **Round-end state sync** `[127]`: `singletonUpdate` on stateId 44 — the TSS service's round-end state snapshot

Items [49] and [127] have different timestamps (~2 seconds apart) and different state IDs, so they represent distinct state mutations — not a redundant copy. However, the underlying WRAPS key payload (~30 MB) is essentially duplicated across all three items.

PR #2380, opened 2026-03-17
docs: Add basic design documents for cloud storage plugins

### LINKED ISSUE #2035: S3 Plugin Rework Design
### Story Form

As a Block Node Engineer/User
I want to have a very detailed official design document
So that I know exactly how the plugin fuctions

### Technical Notes

- We plan a rework of how the S3 remote history plugin works
- We want to create an official design document of how the reworked plugin functions
- The rework includes:
   - Migrate to use the "Bucky" library instead of the local client.
   - Change to use multipart upload for blocks as they arrive instead of batching them in memory and chunking them and uploading them.
   - Change to recover to within the last multipart upload segment if the node shuts down and restarts.
   - Change to determine at the last block successfully stored by the active multipart upload.
     - We need to be able to get the _last_ block in the last successful multipart upload, which can be determined simply by reading back the last multipart item and scanning from the back (which tar supports) to read the last block in the segment. This will allow us to startup successfully after a partial upload has happened.
   - Add further support to achieve highly reliable and repeatable startup.
   - Work out a process to fill in gaps (i.e. signal backfill plugin) on startup.
   - Ensure the plugin has all necessary features to support a reliable, unbroken chain of all blocks uploaded to s3, i.e. full history.

jsync-swirlds:
## Reviewer Notes

* Added document for expanded cloud storage
   * Purpose, Goals, Terms, and Design
   * Other sections to be filled in following fine detail design
* Added document for archived cloud storage
   * Purpose, Goals, Terms, and Design
   * Other sections to be filled in following fine detail design

## Related Issue(s)

Resolves #2035.

### DIFF
--- a/docs/design/Archived-Cloud-Plugin.md
+++ b/docs/design/Archived-Cloud-Plugin.md
@@ -3,15 +3,15 @@
 ## Table of Contents
 
 1. [Purpose](#purpose)
-1. [Goals](#goals)
-1. [Terms](#terms)
-1. [Entities](#entities)
-1. [Design](#design)
-1. [Diagram](#diagram)
-1. [Configuration](#configuration)
-1. [Metrics](#metrics)
-1. [Exceptions](#exceptions)
-1. [Acceptance Tests](#acceptance-tests)
+2. [Goals](#goals)
+3. [Terms](#terms)
+4. [Entities](#entities)
+5. [Design](#design)
+6. [Diagram](#diagram)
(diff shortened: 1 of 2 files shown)


ivannov (commented):
Just some small questions. Maybe their answer is obvious :)

Nana-EC (approved):
Expanded looks fine to me and matches up with some thoughts.
I already took you skeleton and added to it in my draft PR

PR #2382, opened 2026-03-17
docs: Updated and expand BN hardware specs

### LINKED ISSUE #2375: Add a Hardware Requirements Document
### Story Form

As a  Block Node Engineers
I want to have a single doc that cleanly expresses hardware and performance requirments
So that node operators have an easy reference when selecting hardware

### Technical Notes

`docs/block-node/operations/single-node-k8s-deployment.md` currently has node specs in there but more details around benchmarks are needed and that document isn't the right place to add them
Let's move that content and add benchmarking and perf considerations

Nana-EC:
## Reviewer Notes

Add clear block node hardware spec details for node operators

- Add a new `block-node-hardware-specifications.md`
- Add LFH & RFH specs
- Add IOPS and CPY considerations
- Add ingress and egress consideration with simulation for scaling considerations

## Related Issue(s)
Fixes #2375

### DIFF
--- a/docs/block-node/operations/block-node-hardware-specifications.md
+++ b/docs/block-node/operations/block-node-hardware-specifications.md
@@ -213,6 +213,6 @@ All figures use the raw (uncompressed) wire size.
   NVMe and network maximum throughput targets above. PCIe 3.0 configurations may
   be bandwidth-limited.
 - **CPU vintage**: A 2024 or newer CPU is recommended to benefit from improved
-  single-thread IPC and PCIe 4+ native support.
+  single-thread IPC and PCIe 4+ support.
 - **OS disk**: Minimal sizing is sufficient; the OS disk sees negligible I/O after
   initial start-up. Allow at least 10 GB for OCI image storage.
--- a/docs/block-node/operations/single-node-k8s-deployment.md
+++ b/docs/block-node/operations/single-node-k8s-deployment.md
@@ -22,7 +22,7 @@ Quick reference for minimum mainnet deployments:
 
 Note: Servers may be acquired from bare metal or cloud providers that offer dedicated
 instances. LFH configurations require significant storage and are typically sourced from
-bare metal providers.
+bare metal providers, or purchased outright and self-hosted, or colocated.
 
 ## Server Provisioning
 


jasperpotts (approved):

docs/block-node/operations/block-node-hardware-specifications.md
```diff
@@ -0,0 +1,206 @@
...
+## Storage Benchmark Targets
+
+The Block Node is I/O-intensive. The following benchmarks define the sustained throughput,
+IOPS, and latency targets that storage must meet to avoid becoming a bottleneck.
+
+### Disk Performance Targets
+
+| Disk Type  | Sustained Write | Sustained Read | Required Write IOPS | Required Read IOPS | P99 Write Latency | P99 Read Latency |
+|------------|-----------------|----------------|---------------------|--------------------|-------------------|------------------|
+| Fast NVMe  | 4.8 Gbps        | 19.8 Gbps      | 250k                | 800k               | < 300 µs          | < 200 µs         |
```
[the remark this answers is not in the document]
Nana-EC: Units should be right
For instance taking todays 10k TPS block of around 21MB. Apply a 40% overhead and round up for nice numbers give you 30 MB.
That block today is across 2 sec, so double the amount to cover future scaling option of that same amount at 1 sec (I know we're aiming for more but this is a balance).
Then we're using a clean factor of 10 to go from MBps to Mbps to account for ethernet transfers and the extra bits.
Then not the consideration of unlikely but worst case scenario of around 8 CNs pushing to a single block node - assumes smallest distribution of 39 CNs and 5 BNs which shouldn't happen.

So 21 MB * 1.4 * 2 * 10 bits/byte * 8 ~= 4704 Gbps => rounded up is 4.8 Gbps

I didn't want to call out to much details here but I can if you think it'd be helpful

docs/block-node/operations/block-node-hardware-specifications.md
```diff
@@ -0,0 +1,206 @@
...
+
+- NICs: 20 Gbps or higher are recommended for better throughput and future-proofing,
+  even though 10 Gbps is the stated minimum.
+- Bulk storage: 500 TB is recommended for LFH to accommodate long-term block history
+  and state growth. 300 Tb would be the next preferred and 100 TB the minimum for Tier 1.
+- Servers may be sourced from bare metal providers or cloud providers offering dedicated
+  instances. LFH configurations require significant storage capacity and are typically
+  sourced from bare metal providers.
+- OS disk requirements are minimal; the OS disk sees little activity after start-up and
+  does not require explicit sizing beyond standard OS installation needs.
```
jsync-swirlds: Some policies truly minimize O/S disk, so we need to make sure there is room for OCI image files.
```suggestion
- OS disk requirements are minimal; the OS disk sees little activity after start-up and
  does not require explicit sizing beyond standard OS installation needs and at least 10GB for OCI image storage.
```

docs/block-node/operations/block-node-hardware-specifications.md
```diff
@@ -0,0 +1,206 @@
...
+> expected to operate at ≥ 10K TPS or to backfill aggressively.
+
+### Egress Bandwidth (Block Node → Subscribers)
+
+Each downstream subscriber (Mirror Node, Block Node, DApp) receives its own uncompressed stream.
+
+|    TPS | Per subscriber / day | Per subscriber / month | 33 subscribers / day | 33 subscribers / month |
+|-------:|---------------------:|-----------------------:|---------------------:|-----------------------:|
+|  2,000 |               6.5 TB |                 196 TB |               215 TB |                 6.5 PB |
+| 20,000 |                65 TB |                 1.9 PB |             2,145 TB |                64.4 PB |
```
jsync-swirlds: We can set HTTP (GZip) compression on the PBJ plugin, and probably should.
Also, I think you added a large multiplier for single streams somewhere that carried forward.  If we're delivering 6+ petabytes monthly at 2K TPS then block nodes are non-viable.

```suggestion
Each downstream subscriber (Mirror Node, Block Node, DApp) receives its own stream.

|    TPS | Per subscriber / day | Per subscriber / month | 33 subscribers / day | 33 subscribers / month |
|-------:|---------------------:|-----------------------:|---------------------:|-----------------------:|
|  2,000 |               83 GB |                 2.5 TB |               2.7 TB |                 82 TB |
| 20,000 |               747 GB |                22.4 TB |             24.6 TB |                738 TB |
```

docs/block-node/operations/block-node-hardware-specifications.md
```diff
@@ -0,0 +1,206 @@
...
+
+|    TPS | Worst-case egress (33 subscribers) |
+|-------:|-----------------------------------:|
+|  2,000 |                        ~2,000 Mbps |
+| 20,000 |                       ~20,000 Mbps |
+
+> At 20K TPS with 33 subscribers (1/3 of max BNs (13) backfilling, 10 MNs and 10 DApps subscribed),
+> egress at ~20 Gbps saturates the 10 Gbps minimum NIC.
+> Block Nodes serving many live subscribers at high TPS **require** at least a 20 Gbps NIC
+> and may need 40 Gbps or bonded links for headroom.
```
jsync-swirlds:
```suggestion
> Block Nodes serving many live subscribers at high TPS **require** at least a 25 Gbps NIC
> and may need 100 Gbps or multiple bonded 25 Gbps links for headroom.
```

PR #2386, opened 2026-03-17
docs: clarify --network flag can be placed anywhere on the command line

rockysingh:
## Summary
- Clarify in the Multi-Network Usage section that `--network` uses picocli `ScopeType.INHERIT` and can be placed before, between, or after subcommands
- Update all testnet examples across docs to use the more natural trailing position (e.g., `blocks validate /path --network testnet` instead of `--network testnet blocks validate /path`)
- Improves DevEx by matching the conventional command-then-flags ordering operators expect


AlfredoG87 (approved):
Looks good

PR #2389, opened 2026-03-18
feat: End Current Block and Start New One When New Header Received

### LINKED ISSUE #2390: End Current Block and Start New One When New Header Received
### Story Form

As a Publisher Handler
I want to end my current streaming block and start a new one if I receive a header mid-block
So that I will start supplying the newly received block

### Technical Notes

- We want to change the way we handle new headers received when we are currently streaming mid-block
- We want to stop streaming our current block, ending it in the middle and start streaming the newly received one

ata-nas:
## Related Issue(s)

Resolves #2390

### DIFF
--- a/block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/StreamPublisherPluginTest.java
+++ b/block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/StreamPublisherPluginTest.java
@@ -3,7 +3,6 @@ package org.hiero.block.node.stream.publisher;
 
 import static java.util.concurrent.locks.LockSupport.parkNanos;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.fail;
 import static org.hiero.block.node.app.fixtures.TestUtils.enableDebugLogging;
 import static org.hiero.block.node.stream.publisher.fixtures.PublishApiUtility.endThisBlock;
 
@@ -17,7 +16,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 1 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -744,6 +743,12 @@ private void shutdown() {
         }
     }
 
+    /// Check if we have a block in progress. The parameter should be used to supply the
+    /// [#currentStreamingBlockNumber] value.
+    private boolean isCurrentlyMidBlock(final long blockInProgress) {
+        return blockInProgress > UNKNOWN_BLOCK_NUMBER && currentBlockQueue.get() != null;
```
jsync-swirlds: We should _also_ check if the current block is `blockInProgress`.
Not sure if we do much different, but at minimum we should log that case.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -1574,6 +1575,137 @@ void testHandlerIsEndingWithIncompleteBlock(final EndStream.Code code) {
...
+            /// start streaming the new block, but will end mid-block the current one we are
+            /// streaming. Effectively, we will have to call
+            /// [LiveStreamPublisherManager#blockIsEnding(long, long)] for the block we will
+            /// stop streaming. We expect a resend to be sent if we have started streaming a different
+            /// block.
+            /// NOTE:
+            /// > In order for a handler to start streaming a block, it must pass through the manager.
+            /// The manager does not allow two or more publishers to stream the same block simultaneously.
+            /// That being said, this test aims to assert when we are streaming block X and then start
+            /// streaming block Y, we will end block X mid-block and ask for it to be resent.
```
jsync-swirlds: I don't think we always request resend; we only need to request resend if the block that is ended in the middle was already passed on to messaging; otherwise we drop it and _if it is lower than current streaming_ we add that to resend blocks.  Otherwise we just decrement current streaming.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -1574,6 +1575,137 @@ void testHandlerIsEndingWithIncompleteBlock(final EndStream.Code code) {
...
+            /// stop streaming. We do not expect a resend to be sent if we are re-starting the same block.
+            /// NOTE:
+            /// > In order for a handler to start streaming a block, it must pass through the manager.
+            /// The manager does not allow two or more publishers to stream the same block simultaneously.
+            /// That being said, this test aims to assert the only case where we _could_ possibly reach
+            /// a situation where we are currently streaming block X and we start streaming the same block X
+            /// while we are still currently streaming. In those cases, we do not need to ask for a resend
+            /// of block X. The message forwarder will be able to continue streaming this block, if it is
+            /// currently streaming it. Otherwise, it must expect it as a resent block, or a block that
+            /// is higher than the current streaming.
```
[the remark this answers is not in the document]
ata-nas: We have to improve semantics here as this is indeed confusing. Unfortunately we use "currently streaming" both for a block that is currently being published by a publisher (i.e. received by the publisher manager through queues), but also we use the same terminology to say a block is currently being forwarded (by the message forwarding task in the manager) to the live items ring buffer.

What I am saying here:
```
/// Otherwise, it must expect it as a resent block, or a block that
/// is higher than the current streaming.
```
is that the message forwarder task/loop has to expect the block we have just now started to publish as either one waiting to be resent, or higher than (strictly) the one that is currently being forwarded.
If the block we just started publishing is the same as the one we are currently forwarding to the RB, this is valid and will simply be propagated forward. 
The message forwarder, however, is able to detect lower values, this can be seen in the `org.hiero.block.node.stream.publisher.LiveStreamPublisherManager.MessagingForwarderTask#determineCurrentBlockNumber`

We can chat more about this, we have a planned revision of the forwarder loop soon.

What I would propose is that we change the semantics, so we have no confusion here (although I think you have correctly discerned things here). I think we can keep calling blocks that are currently being published as "streaming", but then stop calling it the same for the message forwarder loop. So in the message forwarder, we can change the "currently streaming block" to "current forwarded block" or similar.

PR #2403, opened 2026-03-20
chore: Bootstrap archived-cloud-plugin development

### LINKED ISSUE #2388: Bootstrap archived-cloud-plugin development
### Story Form

As a block node developer
I want to have a specific archived cloud plugin
So that I can implement the features of this epic

### Technical Notes

Create a new plugin module in Block Node.

It should be discoverable by the Block Node app and by Picachu.

As a beginning it should have a plugin with empty of functionality init and start methods. The start method should only log a message so that we can validate that the plugin is successfully integrated with the block node

ivannov:
## Reviewer Notes

This change bootstraps the archive-cloud-storage Gradle subproject. There is no business logic yet — the goal is a buildable, correctly wired module skeleton.                                                                                                                                            
                  
What was added:

- new subproject with build.gradle.kts, module-info.java, the plugin class, and an initial test                  
- com.hedera.bucky:bucky-client:0.1.0-rc1 not yet required in module-info.jav: the dependency is declared now so version alignment is established; the requires directive will be added once the implementation starts using it                                                                                                                                  
- registered the new plugin in the blockNodePlugins configuration and in testModuleInfo
 
## Related Issue(s)

Closes #2388

### DIFF
--- a/block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
+++ b/block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
@@ -20,7 +20,7 @@ import org.hiero.block.node.spi.blockmessaging.BlockNotificationHandler;
 public class ArchiveCloudStoragePlugin implements BlockNodePlugin, BlockNotificationHandler {
 
     /// The logger for this class.
-    private final System.Logger LOGGER = System.getLogger(getClass().getName());
+    private final System.Logger LOGGER = System.getLogger(ArchiveCloudStoragePlugin.class.getName());
 
     /// {@inheritDoc}
     ///
--- a/block-node/archive-cloud-storage/build.gradle.kts
+++ b/block-node/archive-cloud-storage/build.gradle.kts
@@ -16,5 +16,4 @@ testModuleInfo {
     requires("org.junit.jupiter.api")
     requires("org.assertj.core")
     requires("org.hiero.block.node.app.test.fixtures")
-    runtimeOnly("org.junit.jupiter.params")
 }


block-node/app/build.gradle.kts
```diff
@@ -203,6 +206,7 @@ testModuleInfo {
     runtimeOnly("org.hiero.block.node.access.service")
     runtimeOnly("org.hiero.block.node.backfill")
     runtimeOnly("org.hiero.block.node.archive.s3cloud")
+    runtimeOnly("org.hiero.block.node.archive.cloud.storage")
```
ata-nas: Suggestion: we could make this shorter. I think we should probably have a base package `cloud` and then the children could be `archive` or `expanded`. So I think this would look better:
`org.hiero.block.node.cloud.archive`
then later:
`org.hiero.block.node.cloud.expanded`

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -0,0 +1,32 @@
...
+/// order. No local staging occurs. A
+/// [org.hiero.block.node.spi.blockmessaging.PersistedNotification] is published per block
+/// only after durable remote storage is confirmed.
+///
+/// When blocks arrive out of order or an operation fails, the plugin initiates recovery via an
+/// `ArchiveRecoveryTask` that consolidates any temporary `tar` files and resumes normal archiving.
+public class ArchiveCloudStoragePlugin implements BlockNodePlugin, BlockNotificationHandler {
+
+    /// The logger for this class.
+    private final System.Logger LOGGER = System.getLogger(getClass().getName());
```
Nana-EC:
```suggestion
    private final System.Logger LOGGER = System.getLogger(ArchiveCloudStoragePlugin.class.getName());
```

Nana-EC (approved):

PR #2412, opened 2026-03-23
fix: Add address book auto-update to wrap and validate commands

### LINKED ISSUE #2404: Validate command fails at address book changes due to stale registry
## Bug Description

When validating testnet wrapped blocks, signature validation fails at block 29,535,352 because the `addressBookHistory.json` only contains the genesis address book (7 nodes). Testnet has had 7 address book changes (via file updates to `0.0.102`), but neither the `wrap` nor `validate` commands discover these changes from the block data.

The `wrap` command silently drops signatures that fail verification with stale keys, and `validate` enforces a BFT threshold (`n/3 + 1`) and fails when the address book has changed but the registry hasn't been updated.

## Root Cause

The validate command's prefetch pipeline runs signature validation in parallel worker threads using `addressBookRegistry.getAddressBookForBlock()`. When block N contains an address book update, blocks N+1 through N+prefetch are already in the queue with signature validation done against the OLD address book, causing them to fail.

## Fix

- Add `AddressBookUpdateValidation` as a sequential validation that parses each block's `RecordFileItem` to discover address book changes and updates the shared `AddressBookRegistry`
- Add retry logic: when a signature pre-validation error occurs from the parallel stage, defer it, run sequential validations first (which update the address book), then retry signature validation with the updated registry
- Make `AddressBookRegistry.addressBooks` use `CopyOnWriteArrayList` for thread-safe concurrent reads during parallel signature validation

## Parent Issue

Part of #2039

rockysingh:
The wrap command was silently dropping valid signatures when nodes
  rotated their RSA keys via file 0.0.102 updates because it verified
  signatures against a stale genesis address book. This caused validate
  to fail at block 29,535,352 with "2 unique nodes/7 verified, need 3/7"
  since only 2 of 7 signatures survived the wrap process.

  Both commands now discover address book changes from block transaction
  data (file 0.0.102 updates) as blocks are processed and re-verify
  signatures against the correct keys when the prefetch queue contains
  blocks verified with a stale address book.

  Additional fixes:
  - Handle empty/null hashes in PrettyPrint.simpleHash
  - Accept empty previousBlockRootHash in genesis block chain validation
  - Add diagnostic per-signature logging on signature validation failure
  - Make AddressBookRegistry thread-safe with CopyOnWriteArrayList
  - Fall back to genesis address book when no history file is provided

Tested on testnet box.

I ran wrap from the beginning

I then validated and the log below shows we are past block 29,535,352 and we are picking up changes in the address book correctly.

```
[==========================================        ] 85.00%  / Validated 28007919/~32880000 blocks speed 17149.6x ETA: 5m 14sAddress book updated at block 28023538: Address Book Changed, via file update:
   Node 3 key changed from 8a121c2f to 1591d23f

[===========================================       ] 86.00%  / Validated 28288830/~32880000 blocks speed 7937.5x ETA: 6m 47sAddress book updated at block 28325877: Address Book Changed, via file update:
   Node 4 key changed from af579fa8 to 0270015f
   Node 5 key changed from 57b73305 to 7fcebf80

[===========================================       ] 87.00%  / Validated 28756025/~32880000 blocks speed 13144.4x ETA: 6m 8s

The validation passes for all blocks until 17th March 2026:


════════════════════════════════════════════════════════════
  VALIDATION SUMMARY
════════════════════════════════════════════════════════════

Blocks validated: 32872172

VALIDATION PASSED
Time elapsed: 3355 seconds


```

closes #2404

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ToWrappedBlocksCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ToWrappedBlocksCommand.java
@@ -610,7 +610,8 @@ public class ToWrappedBlocksCommand implements Callable<Integer> {
                             effectiveBlock = updateAddressBookAndReverify(preVerified, blockNum, addressBookRegistry);
                         } catch (Exception e) {
                             // Don't fail wrapping for address book parse errors
-                            System.err.printf("Warning: address book auto-update failed at block %d: %s%n", blockNum, e);
+                            System.err.printf(
+                                    "Warning: address book auto-update failed at block %d: %s%n", blockNum, e);
                         }
 
                         // Monthly checkpoint: save state once per calendar month of blockchain data.
@@ -1043,9 +1044,13 @@ public class ToWrappedBlocksCommand implements Callable<Integer> {
                     .map(psf -> psf.toRecordFileSignature(currentBook))
                     .toList();
             PrettyPrint.clearProgress();
-            final String yellowMessage = Ansi.AUTO.string("@|yellow Block %d: re-verified signatures"
-                    + " with updated address book:|@ %d verified (was %d)")
-            System.out.printf(yellowMessage, blockNum, reverifiedSigs.size(), preVerified.verifiedSignatures().size());
+            final String yellowMessage = Ansi.AUTO.string(
+                    "@|yellow Block %d: re-verified signatures" + " with updated address book:|@ %d verified (was %d)");
+            System.out.printf(
+                    yellowMessage,
+                    blockNum,
+                    reverifiedSigs.size(),
+                    preVerified.verifiedSignatures().size());
             return new PreVerifiedBlock(preVerified.recordBlock(), currentBook, reverifiedSigs);
         }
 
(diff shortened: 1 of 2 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ToWrappedBlocksCommand.java
```diff
@@ -598,12 +602,24 @@ public Integer call() throws Exception {
...
+                        // ---- Address book auto-update from block transactions ----
+                        // Discover address book changes (file 0.0.102 updates) from the
+                        // block's record stream so signatures are verified with correct keys.
+                        PreVerifiedBlock effectiveBlock = preVerified;
+                        try {
+                            effectiveBlock = updateAddressBookAndReverify(preVerified, blockNum, addressBookRegistry);
+                        } catch (Exception e) {
+                            // Don't fail wrapping for address book parse errors
+                            System.err.println("Warning: address book auto-update failed at block " + blockNum + ": "
+                                    + e.getMessage());
```
jsync-swirlds: Nitpick, we could use `printf` here (and other similar places).

```suggestion
                            System.err.printf("Warning: address book auto-update failed at block %d: %s%n", blockNum, e);
```

PR #2457, opened 2026-03-28
chore: Design doc for the TssBootstrap

### LINKED ISSUE #2456: chore: Create design doc for TssBootstrap plugin
### Story Form

As a BlockNode engineer I would like to have a design document that describes how the TssBootstrapPlugin works.

### Technical Notes

Address concerns and comments from the weekly design session and the Skeleton PR

berryware:
- add in a design doc for the TssBootstrapPlugin
- document some open questions for the design

## Reviewer Notes
- contains open design questions.

## Related Issue(s)
closes #2456

### DIFF
--- a/docs/design/tss-bootstrap-plugin.md
+++ b/docs/design/tss-bootstrap-plugin.md
@@ -19,66 +19,93 @@ We intend to create a new Block Node plugin (TSSBootstrapPlugin) that queries an
 
 ## Goals
 
-- Deliver TSS booststrap information to a Block Node that does not have access to the network Block 0 transactions.
+- Delivers TSS booststrap information to a Block Node that does not have access to the network Block 0 transactions.
   - A new Block Node needs to learn the current TSS state (roster, verification key, ledger ID)
     - Plugin implements the existing `BlockNodePlugin` interfaces
-    - The planned plugin queries a peer BN's serverStatusDetail gRPC endpoint to retrieve TSS information
+    - The plugin queries a peer BN's serverStatusDetail gRPC endpoint to retrieve TSS information
       - gRPC peer communication follows the same WebClient pattern used elsewhere in the codebase
-    - An `ApplicationStateFacility` interface will be created.
-      - Responsible for handling requests to change `TssData`
-      - Responsible for persisting the `TssData`
-      - Responsible for notifying plugins when the `BlockNodeContext` changes
-      - Will initially be implemented by the `BlockNodeApp`
-    - The plugin will check to see if `TSSData` is stored in its config. `TssBootstrapPlugin` config can be used for both testing and temporary initialization for a Block Node
-    - The planned plugin will periodically query its peers for TSS data updates.
+    - The `ApplicationStateFacility` interface coordinates `BlockNodeContext` and `TssData` changes between plugins
+      - Handles requests to change `TssData`
+      - Persists the latest `TssData`
+      - Notifies plugins when the `BlockNodeContext` changes
+      - Implemented by the `BlockNodeApp`
+    - The plugin checks it's config at startup to see if `TSSData` is present.
+    - `TssBootstrapPluginConfig` can be used for both testing and temporary initialization for a Block Node
+    - The plugin will periodically query its peers for TSS data updates.
   - TSS data is exposed to other plugins
-    - The BlockNode plugin interface will be refactored to have an onContextUpdated method that will be called by the BlockNode application when the context is updated.
-    - The `StatusDetailPlugin` will implement `onContextUpdate()` and receive TSS data updates via the `ApplicationStateFacility`
+    - The BlockNode plugin interface contains an `onContextUpdate` method that is called by the `ApplicationStateFacility` when the context is updated.
+    - The `StatusDetailPlugin` implements `onContextUpdate()` and receives TSS data updates from the `ApplicationStateFacility`
 
 ## Terms
+
+<dl>
+  <dt>TSS</dt>
+  <dd>Threshold Signing Scheme, which consists of two main algorithms: WRAPS and HINTS.</dd>
+</dl>
+<dl>
+  <dt>WRAPS</dt>
+  <dd>Proves validity of the HINTS verification key  of the network with a given Ledger ID.</dd>
+</dl>
+<dl>
+  <dt>HINTS</dt>
+  <dd>Produces a HINTS signature attesting to the claim that majority of node by weight attest to block root hash.
+      This HINTS signature is verified using a daily HINTS verification key.</dd>
+</dl>
+<dl>
+  <dt>Verification Key</dt>
+  <dd>The HINTS verification key used to verify HINTS signature on a message, which in our case is the block root hash. This key is 616 bytes.</dd>
+</dl>
+<dl>
+  <dt>Ledger ID</dt>
+  <dd>SHA-256 hash of the genesis TSS Address Book. This is 32 bytes.</dd>
+</dl>
+<dl>
+  <dt>Schnorr public / private key</dt>
+  <dd>used by WRAPS for the purpose of signing the next day’s TSS Address Book and HINTS verification key</dd>
+</dl>
 
 ## Entities
 
 ### ApplicationStataFacility
 
-- A new entity responsible for updating the plugins when context changes.
-- It will call the `BlockNodePlugin.onContextUpdate()` on all plugins when the `BlockNodeContext` changes.
-- It will be passed directly to the plugins as a parameter to init()
+- Responsible for updating the plugins when context changes.
+- Calls the `BlockNodePlugin.onContextUpdate()` on all plugins when the `BlockNodeContext` changes.
+- Passed directly to the plugins as a parameter to the `init()` method.
 - Plugins can use the `updateTssData` method to update the `TssData` in the `BlockNodeContext`
-- It will process requests to change `TssData`
-- It Will persist `TssData`
-- It Will load persisted `TssData` prior to plugin initialization
-- Will initially be implemented by the `BlockNodeApp`
+- Processes requests to change `TssData`
+- Persists `TssData`
+- Loads persisted `TssData` prior to plugin initialization
+- Implemented by the `BlockNodeApp`
 
 ### BlockNodeContext
 
-- Will contain the `TssData` information and is passed to plugins in the `BlockNodePlugin.init()` call.
-- The BlockNodeContext will also be sent to the plugins via the `BlockNodePlugin.onContextUpdate()` callback function on the plugins
+- Carries `TssData` information and is passed to plugins in a `BlockNodePlugin.init()` call.
+- The `BlockNodeContext` will also be sent to plugins via a `BlockNodePlugin.onContextUpdate()` call to each plugin.
 
 ### ServiceStatusServicePlugin
 
 - The plugin responsible for the `serverStatus`  and `serverStatusDetail` gRPC calls.
-- Will implement the `BlockNodePlugin.onContextUpdate()` to receive `TssData` updates
-- Will respond to peer requests for `TssData` via the `serverStatusDetail` call.
+- Implements the `BlockNodePlugin.onContextUpdate()` to receive `TssData` updates.
+- Responds to peer requests for `TssData` via the `serverStatusDetail` call.
 
 ### TssBootstrapConfig
 
 - Used to configure the BN Peers that will be used to query `serverStatusDetail`
 - Used to bootstrap TSS data via the Config
-- Will override initial data provided by the `ApplicationStateFacility`
+- Overrides initial data provided by the `ApplicationStateFacility`
 
 ### TssBootstrapPlugin
 
-- Will contact peer BNs by querying the `serverStatusDetail` gRPC call.
-- `TssData` can also be configured using the `TssBootstrapConfig`
+- Queries peer BNs for `TssData` via the `serverStatusDetail` gRPC call.
+- `TssData` can be configured using the `TssBootstrapConfig`
 - `TssBootstrapConfig` data will override the `TssData` received during `init()`.
-- It will notify the `ApplicationStateFacility` when it has a `TssData` update.
+- Notifies the `ApplicationStateFacility` when it has a `TssData` update.
 
 ### VerificationPlugin
 
 - Verifies block 0
-- Will notify the `ApplicationStateFacility` when it detects changes to `TssData`
-- Will implement the `BlockNodePlugin.onContextUpdate()` to receive `TssData` updates
+- Notifies the `ApplicationStateFacility` when it detects changes to `TssData`
+- Implements the `BlockNodePlugin.onContextUpdate()` to receive `TssData` updates
 
 ### TssData
 
@@ -146,7 +173,7 @@ We intend to create a new Block Node plugin (TSSBootstrapPlugin) that queries an
 ## Design
 
 - Oustanding design decisions
-  - Should the TssData or TssRoster messages have the starting block number that the TssData is valid from. Either
+  - Should the TssData or TssRoster messages have the starting block number that the TssData is valid from.
 
 ## Diagram
 
@@ -182,7 +209,7 @@ TBD
 ## FAQ
 
 1. Plugin ordering — how does bootstrap init before verification? ServiceLoader doesn't guarantee order.
-   - Plugins will be updated using the ```BlockNodePlugin.onContextUpdate()```.
+   - Plugins will be updated using the `BlockNodePlugin.onContextUpdate()`.
    - Initial `TssData` will be provided in the BlockNodeContext.
 2. Will the `TssBootstrapPlugin` persist the .bin file into the Verification PVC?
    - Plugins should not share configuration information.
@@ -196,3 +223,5 @@ TBD
    - Plugins should not share configuration information.
    - The `TssBootstrapPlugin` will manage its own configuration.
    - Perhaps the `ApplicationStateFacility` should manage the BN Peer information
+     - BN peers for backfill are very likely not the same as BN peers for TSSData.
+       In general the two should not try to coordinate.
--- a/docs/design/tss-bootstrap-plugin.md
+++ b/docs/design/tss-bootstrap-plugin.md
@@ -70,7 +70,7 @@ We intend to create a new Block Node plugin (TSSBootstrapPlugin) that queries an
 
 - Responsible for updating the plugins when context changes.
 - Calls the `BlockNodePlugin.onContextUpdate()` on all plugins when the `BlockNodeContext` changes.
-- Passed directly to the plugins as a parameter to the `init()` method.
+- Passed directly to the plugins as a member of the `BlockNodeContext`.
 - Plugins can use the `updateTssData` method to update the `TssData` in the `BlockNodeContext`
 - Processes requests to change `TssData`
 - Persists `TssData`
@@ -174,6 +174,7 @@ We intend to create a new Block Node plugin (TSSBootstrapPlugin) that queries an
 
 - Oustanding design decisions
   - Should the TssData or TssRoster messages have the starting block number that the TssData is valid from.
+    - valid_from_block has been added to both TssData and TssRoster.
 
 ## Diagram
 
--- a/docs/design/tss-bootstrap-plugin.md
+++ b/docs/design/tss-bootstrap-plugin.md
@@ -1,0 +1,120 @@
+# Design Document Template
+
+## Table of Contents
+
+1. [Purpose](#purpose)
+2. [Goals](#goals)
+3. [Terms](#terms)
+4. [Entities](#entities)
+5. [Design](#design)
+6. [Diagram](#diagram)
+7. [Configuration](#configuration)
+8. [Metrics](#metrics)
+9. [Exceptions](#exceptions)
+10. [Acceptance Tests](#acceptance-tests)
+
+## Purpose
+We intend to create a new Block Node plugin (TSSBootstrapPlugin) that queries another Block Node's detailed status to pick up TSS configuration details needed to initiate TSS verification.
+
+## Goals
+ - Deliver TSS booststrap information to a block node that does not have access to the network Block 0 transactions.
+   - A new Block Node needs to learn the current TSS state (roster, verification key, ledger ID)
+     - Plugin implements the existing BlockNodePlugin interfaces
+     - The planned plugin queries a peer BN's serverStatusDetail gRPC endpoint to retrieve TSS information
+       - gRPC peer communication follows the same WebClient pattern used elsewhere in the codebase
+     - The planned plugin will persist the TSS data locally and load it during plugin init() allowing it to be ready when block proofs arrive with TSS Signatures.
+     - If TSS data is not present locally, the plugin will check to see if the TSS Data is stored in its config. TssBootstrapPlugin config can be used for both testing and temporary initialization for a block node
+     - The planned plugin will periodically query its peers for TSS data updates.
+   - TSS data is exposed to other plugins
+     - The BlockNode plugin interface will be refactored to have an onContextUpdated method that will be called by the BlockNode application when the context is updated.
+     - The StatusDetailPlugin will implement ```onContextUpdate()``` and receive TSS data updates via the  
+
+## Terms
+
+<dl>
+  <dt>Term</dt>
+  <dd>A Term is any word or short phrase used in this document that
+      requires clear definition and/or explanation.</dd>
+</dl>
+
+## Entities
+### BlockNodeContext
+ - Will contain the ```TssData``` information and is passed to plugins in the ```BlockNodePlugin.init()``` call.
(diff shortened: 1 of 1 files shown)


docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,200 @@
...
+    - The BlockNode plugin interface will be refactored to have an onContextUpdated method that will be called by the BlockNode application when the context is updated.
+    - The StatusDetailPlugin will implement ```onContextUpdate()``` and receive TSS data updates via the
+
+## Terms
+
+<dl>
+  <dt>Term</dt>
+  <dd>A Term is any word or short phrase used in this document that
+      requires clear definition and/or explanation.</dd>
+</dl>
```
jsync-swirlds: Leftover?
It would help to define any useful terms here.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Entities
+
+### BlockNodeContext
+
+- Will contain the ```TssData``` information and is passed to plugins in the ```BlockNodePlugin.init()``` call.
+- Could possibly contain the ```PluginManager``` so that plugins can pass context updates directly to the ```PluginManager```
+- The BlockNodeContext will also be sent to the plugins via the ```BlockNodePlugin.onContextUpdate()``` callback function on the plugins
```
jsync-swirlds: Slight rewording to clarify that this is separate from the plugin.

Note, also, that it's generally recommended to use a _single_ backquote(`) for code elements that are not multiple lines.

```suggestion
- Carries `TssData` information and is passed to plugins in a `BlockNodePlugin.init()` call.
- The `BlockNodeContext` will also be sent to plugins via a `BlockNodePlugin.onContextUpdate()` call to each plugin.
```

jsync-swirlds (commented):
One general "voice" concern.
Otherwise just a few suggestions and typos.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
+# Design Document Template
```
Nana-EC: Update title

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+- Persists `TssData`
+- Loads persisted `TssData` prior to plugin initialization
+- Implemented by the `BlockNodeApp`
+
+### BlockNodeContext
+
+- Carries `TssData` information and is passed to plugins in a `BlockNodePlugin.init()` call.
+- The `BlockNodeContext` will also be sent to plugins via a `BlockNodePlugin.onContextUpdate()` call to each plugin.
+
+### ServiceStatusServicePlugin
```
Nana-EC, line 79:
```suggestion
### ServerStatusServicePlugin
```

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+  <dd>SHA-256 hash of the genesis TSS Address Book. This is 32 bytes.</dd>
+</dl>
+<dl>
+  <dt>Schnorr public / private key</dt>
+  <dd>used by WRAPS for the purpose of signing the next day’s TSS Address Book and HINTS verification key</dd>
+</dl>
+
+## Entities
+
+### ApplicationStataFacility
```
Nana-EC, line 62:
```suggestion
### ApplicationStateFacility
```

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+
+- Responsible for updating the plugins when context changes.
+- Calls the `BlockNodePlugin.onContextUpdate()` on all plugins when the `BlockNodeContext` changes.
+- Passed directly to the plugins as a parameter to the `init()` method.
```
Nana-EC: Q: So `BlockNodePlugin.init()` would become `init(BlockNodeContext context, ServiceBuilder serviceBuilder, ApplicationStateFacility applicationStateFacility)`?

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+- Plugins can use the `updateTssData` method to update the `TssData` in the `BlockNodeContext`
```
Nana-EC, line 67: To be clear plugins would call `ApplicationStateFacility.updateTssData()` and this is intended to update the instances of `BlockNodeContext` in all other plugins?

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+
+TBD
+
+## Acceptance Tests
+
+TBD
+
+## FAQ
+
+1. Plugin ordering — how does bootstrap init before verification? ServiceLoader doesn't guarantee order.
```
Nana-EC, line 255: Q: Does this mean `TssBootstrapPlugin` must finish its initial logic before before VerificationPlugin needs the data?

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+    * The schnorr public key
+    */
+      bytes schnorr_public_key = 3;
+}
+```
+
+## Design
+
+- Oustanding design decisions
+  - Should the TssData or TssRoster messages have the starting block number that the TssData is valid from.
```
Nana-EC: I believe so.
It would support the ability to know if the roster is applicable or not

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+- The protobuf `TssData` used in the `serverStatusDetailsResponse`
+
+```markdown
+/**
+* TSS information for a Block-Node.<br/>
+* `TssData` contains the Ledger Id, wraps validation key, and TssRoster information
+*
+* All fields in this message SHALL be filled in, or none.
+*/
+  message TssData {
```
Nana-EC: If this is a snippet from CN proto (which I think it is) we should link back to it.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+
+TBD
+
+## Exceptions
+
+TBD
+
+## Acceptance Tests
+
+TBD
```
Nana-EC: Please fill this in

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+- [Class Diagram](https://mermaid.js.org/syntax/classDiagram.html)
+- [Block Diagram](https://mermaid.js.org/syntax/block.html)
+- [Architecture Diagram](https://mermaid.js.org/syntax/architecture.html)
+- [Flowchart](https://mermaid.js.org/syntax/flowchart.html)
+- [State Diagram](https://mermaid.js.org/syntax/stateDiagram.html)
+- [Entity Relationship Diagram](https://mermaid.js.org/syntax/entityRelationshipDiagram.html)**
+
+## Configuration
+
+TBD
```
Nana-EC: Please fill this in or note none

Nana-EC (changes requested):
Good thoughts.
Left some feedback

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+  <dd>Proves validity of the HINTS verification key  of the network with a given Ledger ID.</dd>
+</dl>
+<dl>
+  <dt>HINTS</dt>
+  <dd>Produces a HINTS signature attesting to the claim that majority of node by weight attest to block root hash.
+      This HINTS signature is verified using a daily HINTS verification key.</dd>
+</dl>
+<dl>
+  <dt>Verification Key</dt>
+  <dd>The HINTS verification key used to verify HINTS signature on a message, which in our case is the block root hash. This key is 616 bytes.</dd>
```
jsync-swirlds: This is actually the WRAPS verification key.  The HinTS verification key rotates daily and is part of what the WRAPS proves.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+## Terms
+
+<dl>
+  <dt>TSS</dt>
+  <dd>Threshold Signing Scheme, which consists of two main algorithms: WRAPS and HINTS.</dd>
+</dl>
+<dl>
+  <dt>WRAPS</dt>
+  <dd>Proves validity of the HINTS verification key  of the network with a given Ledger ID.</dd>
+</dl>
```
jsync-swirlds, line 40: WRAPS is an acronym of "Weighted Roster Attestation Proof System". We might include that here for completeness.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+<dl>
+  <dt>HINTS</dt>
+  <dd>Produces a HINTS signature attesting to the claim that majority of node by weight attest to block root hash.
+      This HINTS signature is verified using a daily HINTS verification key.</dd>
+</dl>
```
jsync-swirlds, line 45: HINTS is actually hinTS (**hin**ted **T**hreshold **S**ignature scheme), though it's a bit pedantic to harp on...

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+  <dd>The HINTS verification key used to verify HINTS signature on a message, which in our case is the block root hash. This key is 616 bytes.</dd>
+</dl>
+<dl>
+  <dt>Ledger ID</dt>
+  <dd>SHA-256 hash of the genesis TSS Address Book. This is 32 bytes.</dd>
+</dl>
+<dl>
+  <dt>Schnorr public / private key</dt>
+  <dd>used by WRAPS for the purpose of signing the next day’s TSS Address Book and HINTS verification key</dd>
+</dl>
```
jsync-swirlds, line 58: BTW, the Schnorr keys are not used by WRAPS, but by hinTS.
absent the WRAPS, we accept the Schnorr keys directly (and assume equal weight).

Once we have the WRAPS we can use the hinTS key generated/proved in the WRAPS and verify the _aggregate_ signature (which is generally much smaller than the concatenated Schnorr signatures).

The "magic" of BLS is that it takes a set of Schnorr signatures and produces a single signature that only verifies if a sufficient number of valid Schnorr signatures were combined.

What hinTS does is support _weight_ for each signature, so instead of just count the verification requires enough aggregate _weight_.

What WRAPS does is provide a SNARK that proves the roster, which includes the individual public keys, the daily verification key, and the weight for each signer.  That roster is what provides the "hints" needed to validate correctly with respect to _weight_ rather than just _count_.  The daily verification key is what we actually use to verify the BLS signature.

docs/design/tss-bootstrap-plugin.md
jsync-swirlds: Not quite.
`TssBootstrapPlugin` will call `ApplicationStateFacility.updateTssData()` and
the `ApplicationStateFacility` will call `onContextUpdate` for each plugin.

Later (with `VerificationPlugin` updates) the `VerificationPlugin` will call `ApplicationStateFacility.updateTssData()` after extracting TSS data from block `0`; and the same update will propagate that data to other plugins.

In both cases, the task of persisting the data will belong to the `ApplicationStateFacility`, so we have a single plugin managing both storing and loading the data (which avoids multiple plugins needing to manage individual storage and avoids multiple plugins getting too out of sync for the same data).

docs/design/tss-bootstrap-plugin.md
jsync-swirlds: The current method is `init(BlockNodeContext context, ServiceBuilder serviceBuilder)` (see [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/BlockNodePlugin.java#L55)).

This effort should add `ApplicationStateFacility` to the `BlockNodeContext` record, not as an added parameter to the `init` method.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+
+- The plugin responsible for the `serverStatus`  and `serverStatusDetail` gRPC calls.
+- Implements the `BlockNodePlugin.onContextUpdate()` to receive `TssData` updates.
+- Responds to peer requests for `TssData` via the `serverStatusDetail` call.
+
+### TssBootstrapConfig
+
+- Used to configure the BN Peers that will be used to query `serverStatusDetail`
+- Used to bootstrap TSS data via the Config
+- Overrides initial data provided by the `ApplicationStateFacility`
```
jsync-swirlds: I wonder if we want this to be the reverse, any data stored in the `ApplicationStateFacility` overrides configuration.
Not sure, both alternatives have downsides.

docs/design/tss-bootstrap-plugin.md
jsync-swirlds: Not CN; this is our own definition and should be in BN API.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+      uint64 node_id = 1;
+   /**
+    * The node weight
+    */
+      uint64 weight = 2;
+   /**
+    * The schnorr public key
+    */
+      bytes schnorr_public_key = 3;
+}
```
jsync-swirlds: General note that these should have 4 space indentation and the doc comments should align with the fields, no 1-space extra indent.

There should be a blank line after each field.

docs/design/tss-bootstrap-plugin.md
jsync-swirlds: I agree, the TSSData _and_ TSSRoster must _each_ include a `valid_from_block` field (and these might not be the same value).

docs/design/tss-bootstrap-plugin.md
jsync-swirlds: Not exactly.
All `init` calls must complete before the block node is "ready".
By the time all `init` calls are completed, the Verification Plugin will have either loaded with correct TSSData or received a Context update with correct TSSData.

So yes, the plugin must finish init before Verification needs the data; but that's already guaranteed by the app.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+
+1. Plugin ordering — how does bootstrap init before verification? ServiceLoader doesn't guarantee order.
+   - Plugins will be updated using the `BlockNodePlugin.onContextUpdate()`.
+   - Initial `TssData` will be provided in the BlockNodeContext.
+2. Will the `TssBootstrapPlugin` persist the .bin file into the Verification PVC?
+   - Plugins should not share configuration information.
+   - The verification plugin should manage its own configuration
+   - The `ApplicationStateFacility` will be responsible for persisting the `TssData`
+   - When the verification plugins detects new `TssData`, ie Block 0, it will notify the `ApplicationStateFacility`.
+   - The `ApplicationStateFacility` will update all plugins of BlockNodeContext changes via `BlockNodePlugin.onContextUpdate()`
```
jsync-swirlds, line 271: This is good, but perhaps add one more line at the top.
   - No plugin will write to storage managed by another plugin.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,227 @@
...
+3. Is this plugin intended to run only once at BN first run.
+   - The plugin will periodically query its peer BN servers to see if they have any `TssData` updates.
+4. Where and how it will get the BN source(s) for trusted peers to get the TSS data from? We already have the concept of backfill peers.
```
jsync-swirlds: Minor grammar fix:

```suggestion
4. Where and how will it get the BN source(s) for trusted peers to get the TSS data from? We already have the concept of backfill peers.
```

jsync-swirlds (commented):
Good progress.
A few additional notes.

docs/design/tss-bootstrap-plugin.md
jsync-swirlds:
```suggestion
# TSS Bootstrap Plugin Design
```

docs/design/tss-bootstrap-plugin.md
berryware: I did this for testing so that the config data provided by the test will win.

jsync-swirlds: The problem this raises is that _in production_ we could receive an update, then on restart that update gets wiped out because an older value is still set in config.

If nothing ever goes wrong this might be OK, but if the BN is restarted without clearing the config, it could be problematic.

docs/design/tss-bootstrap-plugin.md
berryware: I removed ApplicationStateFacility from `init` and added it to the BlockNodeContext. Updated the documentation to match

docs/design/tss-bootstrap-plugin.md
berryware: Joseph's description is accurate.

docs/design/tss-bootstrap-plugin.md
berryware: Some additional clarification. The `ApplicationStateFacility` loads the last persisted `TssData` prior to any `init()` calls. This allows the `TssBootstrapPlugin` to overwrite with `TssData` with TssData stored in the config. The VerificationPlugin could also override the `TssData` with what it thinks is the latest.

docs/design/tss-bootstrap-plugin.md
berryware: valid_from_block has been added to both TssData and TssRoster and the design doc has been updated to match.

docs/design/tss-bootstrap-plugin.md
berryware: Correct. This is part of the BN API.

docs/design/tss-bootstrap-plugin.md
berryware: corrected

docs/design/tss-bootstrap-plugin.md
berryware: Added

docs/design/tss-bootstrap-plugin.md
berryware: I got this from Notion: https://www.notion.so/swirldslabs/TSS-1f17c9ab2591801d83c5e16fb820df12?source=copy_link#2cd7c9ab2591803e92f6f116a64ae166

docs/design/tss-bootstrap-plugin.md
berryware: Updated.

docs/design/tss-bootstrap-plugin.md
berryware: Updated

docs/design/tss-bootstrap-plugin.md
berryware: Updated

docs/design/tss-bootstrap-plugin.md
berryware: Added.

docs/design/tss-bootstrap-plugin.md
berryware: Added

docs/design/tss-bootstrap-plugin.md
jsync-swirlds: It's used in a few cases. All short descriptions are simplifications, but perhaps a more _clear_ simplification might be as follows:
> A key produced privately by each node and used during the address book update process to adopt a new TSS roster. This key may also directly sign a block proof if the WRAPS process is not enabled.

docs/design/tss-bootstrap-plugin.md
jsync-swirlds: I still think that's risky.
I would expect ApplicationStateFacility to load data _during_ init, as plugins really aren't supposed to do much before then (even facility-type plugins).

Config overriding loaded data is difficult as it's risky (we could erase a critical update unexpectedly because config wasn't cleared), but also difficult to prevent (though there are some things we can do in that regard).

Something to consider, perhaps, in future updates.

Nana-EC: My concern was what if VerificationPlugin init runs first, how do we make sure it gets the latest data since the `TssBootstrapPlugin.init()` doesn't present the latest data till after?
In that case it seems the `ApplicationStateFacility` call in the `TssBootstrapPlugin` will ensure the `VerificationPlugin` `BlockNodeContext` gets updated?

docs/design/tss-bootstrap-plugin.md
Nana-EC: Okay, no-op then.

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,257 @@
...
+9. [Exceptions](#exceptions)
+10. [Acceptance Tests](#acceptance-tests)
+
+## Purpose
+
+We intend to create a new Block Node plugin (TSSBootstrapPlugin) that queries another Block Node's detailed status to pick up TSS configuration details needed to initiate TSS verification.
+
+## Goals
+
+- Delivers TSS booststrap information to a Block Node that does not have access to the network Block 0 transactions.
```
Nana-EC: nit: quite lengthy goals. 
This reads as a mixture of goals and design specs.

For this I might expect goals  as follows
1. BN detects missing TSS state and queries peer BNs for data
2. BN validate TSS Data on load
3. BN propagates TSS data load/changes to other plugins

The flow of logic and how the entities interact would then be in the Design section

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Design
+
+- Completed design decisions
+  - Should the TssData or TssRoster messages have the starting block number that the TssData is valid from.
+    - valid_from_block has been added to both TssData and TssRoster.
+
+## Diagram
+
+TBD
```
Nana-EC, line 206: One of these diagrams to show case the flow of the TSS data between entities would be a good illustration

Nana-EC (dismissed):
Some remaining nits, but unblocking so as to get 1st version in.
You can tweak on a related or follow up PR

docs/design/tss-bootstrap-plugin.md
jsync-swirlds: All init happens long before any plugin might have new data.
Data storage belongs to the ApplicationState plugin, and that plugin can (should?) ensure its data is updated to the context as part of init, but only if the stored data is newer than any previously set value.
After init, but before start, the context can be updated to all plugins (via `onContextUpdate`) _if_ it was updated during init.

Note, verification shouldn't be updating the TSS data in most cases; certainly not during init (we intend to remove any storage or configuration for TSS data in the Verification plugin, the ApplicationStateFacility and TSS Bootstrap will handle that instead). The only time Verification might update TSS data is based on Block `0` (and even that could migrate to another plugin if appropriate).

berryware: ApplicationStateFacility and all the plugins can now init without getting any context updates. Any TssData updates they submit will be added to a ConcurrentLinkedQueue. After plugin initialization, the ApplicationStateFacility takes the TssData with the greatest validFromBlock and calls the onContextUpdate method for each plugin.

docs/design/tss-bootstrap-plugin.md
berryware: Updated

docs/design/tss-bootstrap-plugin.md
berryware: ApplicationStateFacility and all the plugins can now init without getting any context updates. Any TssData updates they submit will be added to a ConcurrentLinkedQueue. After plugin initialization, the ApplicationStateFacility takes the TssData with the greatest validFromBlock and calls the onContextUpdate method for each plugin. So if the config is the greatest, it wins. If not, the persisted data wins.

docs/design/tss-bootstrap-plugin.md
berryware: Updated

jsync-swirlds (approved):

docs/design/tss-bootstrap-plugin.md
```diff
@@ -0,0 +1,279 @@
...
+
+- `tssDataFilePath`. The file path to use to persist and load `TssData`
+- `tssUpdatePollIntervalMillis`. The poll interval in milliseconds that the ApplicationStateFacility uses to check for
+  `TssData` updates
+
+### TssBootstrapPlugin (TssBootstrapConfig)
+
+- `ledgerId`. Base64 encoded `ledgerId`
+- `wrapsVerificationKey`. Base64 encoded `wrapsVerificationKey`
+- `nodeId`. The node id
```
Nana-EC, line 231: I saw this in the implementation. 
Why do we only have one entry in the roster covered here?
We can update in follow up

Nana-EC (approved):
Generally looks good.
Can get in and address updates based on implementation in follow up

Issue #1166, opened 2025-05-19
`S3ArchivePlugin` Cleanup and Error Handling Improvements


ata-nas:
### Story Form

As a Block Node Developer
I want to have a clean, well managed plugin with improved error handling capabilities
So that it is easily manageable and updatable if need be

### Technical Notes

- the `S3ArchivePlugin` is in need of some cleanup
- the plugin also needs to be extendable as we will have to introduce some improvements, like retries and cleanups of failed uploads, as well as local persistent key-value storage for plugin operations
- this issues is connected to #1118 as changes in that issue will affect this one

PR #2473, opened 2026-03-30
feat: S3ArchivePlugin Starts Uploads Sooner

### LINKED ISSUE #1052: `S3ArchivePlugin` Starts Uploads Sooner

ivannov:
## Reviewer Notes

This commit implements the core upload logic for the `archive-cloud-storage` plugin — verified blocks are now streamed to S3-compatible storage as tar archives using multipart upload on a virtual thread. 
 
### What was added 

**`ArchiveCloudStorageConfig`** — a config record for S3 settings: endpoint URL, region, bucket name, storage class, access/secret credentials, grouping level (controls how many blocks per archive; `10^groupingLevel`), and multipart part size.
 
**`BlockArchiveTask` (interface)** — defines the single-use lifecycle contract for a tar upload session: `init()` opens the S3 client and starts a multipart upload, `submit()` delivers a verified block to the running task (returns `SUCCESS`, `FINISHED`, or `BLOCK_OUT_OF_RANGE`), and `abort()` cancels any in-flight upload.

**`LiveBlockArchiveTask`** — the concrete implementation. Runs on a virtual thread, waits for each block in the range via `CompletableFuture`, appends it as a tar entry (zstd-compressed), and flushes multipart parts as they fill up. On range completion it sends a `PersistedNotification` for each archived block and completes the multipart upload.

**`BlockToTarEntry`** — stateless utility that serialises a `BlockUnparsed` into a ustar-format tar entry: 512-byte header (filename derived from zero-padded block number + `.blk.zstd` extension, octal fields, checksum) followed by zstd-compressed protobuf bytes padded to a 512-byte boundary.
 
**`ArchiveCloudStoragePlugin` (expanded)** — now implements `BlockNotificationHandler`. On each `handleVerification` call:

1. If no live task exists, a new `LiveBlockArchiveTask` is created for the block's group range, initialised, submitted to the virtual-thread executor, and any stashed blocks are replayed. 
2. The block is submitted to the task; `FINISHED` derefs the task (the executor still holds a handle until the upload completes), `BLOCK_OUT_OF_RANGE` stashes the block for the next task.

On `stop()`, any in-progress upload is aborted.

### Other changes

- `BlockSource.CLOUD_ARCHIVE` added to the SPI enum so archived blocks can be identified in `PersistedNotification`. 
- `bucky-client:0.1.0-rc2` activated in `hiero-dependency-versions` (was commented out pending this commit).
- `module-info.java` updated with all new `requires` and the config-package `exports`. 
 
### Tests
 
`ArchiveCloudStoragePluginTest` and `BlockToTarEntryTest` provide coverage for the plugin's orchestration logic (stash/replay, task lifecycle, abort 
on stop) and the tar-entry serialisation respectively.

## Related Issue(s)

Closes #1052


block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -20,13 +37,116 @@
...
+        context.blockMessaging().registerBlockNotificationHandler(this, false, "Archive Cloud Storage");
+    }
 
     /// {@inheritDoc}
-    ///
-    /// Starts the archive cloud storage plugin.
     @Override
     public void start() {
-        LOGGER.log(DEBUG, "Archive cloud storage plugin started");
+        virtualThreadExecutor = context.threadPoolManager().getVirtualThreadExecutor();
```
ata-nas: We have to make sure that we are not using a default shared pool here.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -20,13 +37,116 @@
...
+        }
+
+        // Submit the block to the archive task. Should be very quick, as the task is running in a different thread
+        // and actual processing happens in the background.
+        // liveBlockArchiveTask may be null here if the stash replay above already completed the task.
+        if (liveBlockArchiveTask == null) {
+            blocksStash.put(notification.blockNumber(), notification.block());
+            return;
+        }
+        final ArchiveResult result = liveBlockArchiveTask.submit(notification.block(), notification.blockNumber());
```
ata-nas: I think we can utilize a queue to share data between threads here. It seems wrong to continuously call a method on a task that is running.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/BlockArchiveTask.java
```diff
@@ -0,0 +1,66 @@
...
+///
+/// Typical call sequence:
+/// 1. **[#init]** — opens the S3 client and creates the multipart-upload session.
+/// 2. **[Runnable#run]** — submitted to a virtual-thread executor; blocks internally waiting for each block in range to
+/// arrive via [#submit].
+/// 3. **[#submit]** — called on the notification thread for every verified block; resolves the internal future that
+/// unblocks [Runnable#run].
+/// 4. **[#abort]** — called when the node is stopping or an unrecoverable error occurs; cancels any pending futures and
+// aborts the S3 multipart upload.
+interface BlockArchiveTask extends Runnable {
```
ata-nas: Runnable is a really bad idea here IMO. Especially because we are expecting a result.
This whole interface seems completely redundant to me. We should not really need to use these methods.

I question why we need an init, that should be in the constructor of a Callable. Submit (sharing data) should probably be a queue instead. Abort should be a simple cancel on a Callable.

ata-nas (commented):
First pass. Quite a few questions and my personal suggestions for improving this.

In my my mind, we can simplify this a lot. We should also really utilize a callable.

I think we can utilize an upload task that will live until the whole batch is uploaded, possibly start multiple ones for different batches, and then as blocks come in, they can be fed to the correct task, or compressed and tarred asyncrhonously even if we wish (I think this is by design as well) and then fed to the correct task. If we are smart about it, the archive task should receive ready to go, tarred blocks (byte array) and simply make the next multipart upload.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/LiveBlockArchiveTask.java
```diff
@@ -0,0 +1,293 @@
...
+    // block that is received. Once the buffer reaches a predefined size, that size is uploaded to S3 and the buffer is
+    // cut and what remains is put in the front of the buffer
+    byte[] taredBlocksBuffer = new byte[0];
+    // Keep track of the block numbers of the blocks that were buffered before the last upload. This is used to send
+    // persisted notifications for those blocks
+    private final List<Long> blocksInBuffer = new ArrayList<>();
+
+    // As there are two threads performing S3 client operations, we need to ensure thread safety when accessing it, so
+    // that abort doesn't happen mid-operation
+    private final ReentrantLock s3ClientLock = new ReentrantLock();
```
jsync-swirlds: Why would we have two threads sharing a client instead of just running a separate client for each task?

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -20,13 +37,116 @@
...
+            virtualThreadExecutor.submit(liveBlockArchiveTask);
+
+            tryReplayStash();
+        }
+
+        // Submit the block to the archive task. Should be very quick, as the task is running in a different thread
+        // and actual processing happens in the background.
+        // liveBlockArchiveTask may be null here if the stash replay above already completed the task.
+        if (liveBlockArchiveTask == null) {
+            blocksStash.put(notification.blockNumber(), notification.block());
```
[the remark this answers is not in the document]
ivannov: With the check that I introduced as per your advise, failed verification will not reach this point

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -20,13 +39,200 @@
...
+    @Override
+    public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
+        this.context = requireNonNull(context);
+        this.config = context.configuration().getConfigData(ArchiveCloudStorageConfig.class);
+        final List<String> violations = validateConfig();
+        if (!violations.isEmpty()) {
+            LOGGER.log(
+                    WARNING,
+                    "Archive cloud storage plugin is not active because of empty values for the following configurations: {0}",
+                    String.join(", ", violations));
```
ata-nas: Please note that this validation should one day be removed in favor of an updated health facility, where we will be able to notify the health facility that we are unhealthy. Appropriate decision and the action thereof shall be performed entirely by the health facility.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
ata-nas: We can revisit this.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -20,13 +39,200 @@
...
+        }
+
+        final long blockNumber = notification.blockNumber();
+
+        // If the active task has completed, surface any exception and clean up.
+        // Returns true when the task was cancelled (i.e. stop() was called): no new task should be
+        // started in that case since the plugin is shutting down.
+        if (handleCompletedUpload()) {
+            return;
+        }
```
ata-nas: Same as above, if we must return from this if, let's just put the rest of the code in an else clause. The compiler works in our favor. Here we can simply negate the reuslt and continue in the if clause
```suggestion
        if (!handleCompletedUpload()) {
            // your method continues here
        }
```

ata-nas (commented):
Good progress, I still have some questions and suggestions for improvements.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStorageConfig.java
```diff
@@ -0,0 +1,35 @@
...
+    @Loggable @ConfigProperty(defaultValue = "10") @Max(2047) @Min(5) int partSizeMb,
+    @Loggable @ConfigProperty(defaultValue = "") String endpointUrl,
+    @Loggable @ConfigProperty(defaultValue = "") String regionName,
+    @ConfigProperty(defaultValue = "") String accessKey,
+    @ConfigProperty(defaultValue = "") String secretKey,
+    @Loggable @ConfigProperty(defaultValue = "") String bucketName,
+    @Loggable @ConfigProperty(defaultValue = "STANDARD") S3StorageClass storageClass) {
+    // spotless:on
+
+    // Source for storage class values: https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-class-intro.html
```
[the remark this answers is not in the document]
ata-nas: For this one, I think we can put it as a note in the configs md. I have to see for other manuals.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/BlockUploadTask.java
```diff
@@ -0,0 +1,226 @@
...
+                final byte[] extended = new byte[buffer.length + tarEntry.length];
+                System.arraycopy(buffer, 0, extended, 0, buffer.length);
+                System.arraycopy(tarEntry, 0, extended, buffer.length, tarEntry.length);
+                buffer = extended;
+
+                // Flush a fixed-size part to S3 once the buffer reaches the threshold
+                if (buffer.length >= partSizeBytes) {
+                    try {
+                        buffer = uploadPart(buffer, s3, uploadId, etags);
+                    } catch (Exception _) {
```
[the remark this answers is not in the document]
ata-nas: I think there is a misunderstanding here. I am saying that we should not catch an `Exception` here at all. If there is a checked exception that we can do something about here, we should catch it and do something about it. Otherwise, we have to allow the exception to be thrown. That will result the future to end exceptionally and we will properly receive the thrown exception when we call get.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/BlockSource.java
```diff
@@ -9,5 +9,6 @@ public enum BlockSource {
     UNKNOWN,
     PUBLISHER,
     BACKFILL,
-    HISTORY
+    HISTORY,
+    CLOUD_ARCHIVE
```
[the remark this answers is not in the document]
ata-nas: We have to use the source that is coming from the `VerificationNotification` that you are handling. If the source of the notification is null, only then we can use UNKNOWN.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/BlockUploadTask.java
```diff
@@ -0,0 +1,226 @@
...
+
+            for (long blockNum = firstBlock; blockNum < firstBlock + groupSize; blockNum++) {
+                // Handle cancellation at the start of every iteration, in addition to the
+                // InterruptedException that blockQueue.take() throws on interruption. This additional
+                // prerequisite is necessary in case when the queue is filled (e.g. by stash replay) and
+                // at this point plugin is stopped. The blockQueue.take() call will not throw interrupted
+                // exception until the queue is drained.
+                if (Thread.interrupted()) {
+                    throw new InterruptedException();
+                }
```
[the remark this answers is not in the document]
ivannov: Yes, I handled this with the commit

```
if (blockNum == firstBlock) {
    s3.abortMultipartUpload(key, uploadId);
}
````

File: block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java (hunk 7 above)
Comment (ivannov) [reply block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java]: Applied your suggestion

File: block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/BlockSource.java (hunk 10 above)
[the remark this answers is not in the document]
Comment (ivannov) [reply block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/BlockSource.java]: Changed to look at the verification's block source

File: block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStorageConfig.java (hunk 8 above)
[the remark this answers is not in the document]
Comment (ata-nas) [reply block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStorageConfig.java]: We generally update the config md everytime we have a change that takes place. Now is definitely the right time.

**Review by ata-nas — COMMENTED** [review]
Great cleanup and improvements! I think we can address further discussions in follow ups, as applied, unless we do not have something hard blocking.

File: block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/BlockUploadTask.java [hunk 12]
```diff
@@ -0,0 +1,337 @@
...
+                        // and sent notifications — adding it again here would cause a duplicate notification.
+                        blocksInBuffer.put(blockNum, blockData.source());
+                    }
+                } catch (InterruptedException e) {
+                    LOGGER.log(TRACE, "Block upload task interrupted", e);
+                    try {
+                        cleanupEmptyUpload(s3, uploadId, etags);
+                    } catch (S3ResponseException | IOException ex) {
+                        LOGGER.log(INFO, "Failed to abort multipart upload after interruption", ex);
+                    }
```
jsync-swirlds: This should probably be in the `cleanupEmptyUpload` method.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -1,32 +1,278 @@
...
+            final long blockNumber = notification.blockNumber();
+            // Start a new upload task when there is none active
+            // TODO(1166) Is this null because of a new batch or because the previous block failed?
+            if (currentUploadFuture == null) {
+                startNewUploadTask(blockNumber);
+            }
+            // Stage the notification; if it belongs to the active task's range, attempt to drain
+            // consecutive blocks into the queue. Out-of-range notifications go to the cross-group stash.
+            routeNotification(blockNumber, notification);
+        }
```
jsync-swirlds: Rather than check for failure and log, perhaps invert this.

Also, the result of handCompletedUpload seems backwards, it returns true if stopped, but false if it succeeded?

```suggestion
        if (notification != null && notification.success() && notification.block() != null) {
            if (handleCompletedUpload()) {
                // What do we do if this succeeds?
                // handleCompletedUpload returns true when the task was cancelled (i.e. stop() was called):
                // no new task should be started in that case since the plugin is shutting down.
            } else {
                // If the active task has completed, surface any exception and clean up.
                final long blockNumber = notification.blockNumber();
                // Start a new upload task when there is none active
                // TODO(1166) Is this null because of a new batch or because the previous block failed?
                if (currentUploadFuture == null) {
                    startNewUploadTask(blockNumber);
                }
                // Stage the notification; if it belongs to the active task's range, attempt to drain
                // consecutive blocks into the queue. Out-of-range notifications go to the cross-group stash.
                routeNotification(blockNumber, notification);
            }
        } else {
            logInvalidOrFailedNotification(notification);
        }
```

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -1,32 +1,278 @@
...
+                    LOGGER.log(TRACE, "Block upload task was cancelled");
+                    cancelled = true;
+                } else {
+                    final UploadResult uploadResult = currentUploadFuture.get();
+                    if (uploadResult == UploadResult.FAILED) {
+                        LOGGER.log(WARNING, "Block upload task failed");
+                        // TODO(1166) Handle properly block upload task failure
+                    }
+                }
+            } catch (InterruptedException _) {
```
jsync-swirlds: Use `ignored` here if you intend to ignore the exception.  Do not use `_` for a variable name, anywhere.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/BlockToTarEntry.java
```diff
@@ -0,0 +1,101 @@
...
+    ///
+    /// @param block the block to convert
+    /// @param blockNumber the block number, used to derive the file name inside the tar
+    /// @return a byte array representing the complete tar entry for the block
+    static byte[] toTarEntry(@NonNull final BlockUnparsed block, final long blockNumber) {
+        Objects.requireNonNull(block);
+
+        final String currentBlockName = String.format("%019d", blockNumber) + EXTENSION;
+        final byte[] currentBlockBytes = CompressionType.ZSTD.compress(
+                BlockUnparsed.PROTOBUF.toBytes(block).toByteArray());
```
jsync-swirlds: Don't use the compress method for ZSTD, it's very slow and doesn't produce the same output as the stream option.  We left it in place for testing, but we must not use it for production code.

This should create a byte array stream and write using a wrapped stream
something similar to what we do [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java#L398) (but using a byte array output stream rather than a file output stream).

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/BlockUploadTask.java
```diff
@@ -0,0 +1,337 @@
...
+                } catch (S3ResponseException | IOException e) {
+                    blocksInBuffer.forEach((bn, src) ->
+                            blockMessaging.sendBlockPersisted(new PersistedNotification(bn, false, 1_000, src)));
+                    LOGGER.log(INFO, "Failed to upload final part for key {0}", key, e);
+                    result = UploadResult.FAILED;
+                }
+                if (result == UploadResult.SUCCESS) {
+                    blocksInBuffer.forEach((bn, src) ->
+                            blockMessaging.sendBlockPersisted(new PersistedNotification(bn, true, 1_000, src)));
+                }
```
jsync-swirlds: Don't send persistence notifications for every single block, we deliberately designed the structure to allow jumping forward.
For failure, only send one notification, for the _first_ block that failed.
For success, only send one notification, for the _last_ block that succeeded.

block-node/archive-cloud-storage/build.gradle.kts
```diff
@@ -14,6 +14,13 @@ mainModuleInfo {
 
 testModuleInfo {
     requires("org.junit.jupiter.api")
+    requires("org.junit.jupiter.params")
     requires("org.assertj.core")
     requires("org.hiero.block.node.app.test.fixtures")
+    requires("org.hiero.metrics")
+    requires("com.github.luben.zstd_jni")
+    requires("org.testcontainers")
+    requires("io.minio")
```
jsync-swirlds: Can we remove `minIO`?

jsync-swirlds (commented):
A few concerns, some awkward constructions, and a few possible alternatives.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/BlockUploadTask.java
```diff
@@ -0,0 +1,337 @@
...
+            // Upload any bytes that didn't fill a complete part during the loop.
+            if (result == UploadResult.SUCCESS && buffer.length > 0) {
+                try {
+                    doUploadPart(buffer, s3, uploadId, etags);
+                } catch (S3ResponseException | IOException e) {
+                    blocksInBuffer.forEach((bn, src) ->
+                            blockMessaging.sendBlockPersisted(new PersistedNotification(bn, false, 1_000, src)));
+                    LOGGER.log(INFO, "Failed to upload final part for key {0}", key, e);
+                    result = UploadResult.FAILED;
+                }
```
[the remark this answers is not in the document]
ata-nas: Generally, if we get an S3 exception, this should be because we really have a failure. The bucky client should handle retries and be smarter, something we discussed in the past.

About returning the failed result, I think this is acceptable, we indicate there has been a failure we cannot recover from. I think the next action here is to simply trigger the same mechanism we have during startup, i.e. check for active multipart upload and continue.

As far as the failed notification, as you make points elsewhere, yes, we should probably only fail the lowest block number that has failed to be persisted.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -1,32 +1,278 @@
...
-        LOGGER.log(DEBUG, "Archive cloud storage plugin started");
+        virtualThreadExecutor = context.threadPoolManager().getVirtualThreadExecutor();
+        LOGGER.log(TRACE, "Archive cloud storage plugin started");
+    }
+
+    /// {@inheritDoc}
+    @Override
+    public void handleVerification(VerificationNotification notification) {
+        if (notification == null || !notification.success() || notification.block() == null) {
+            logInvalidNotification(notification);
```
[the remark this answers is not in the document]
ivannov: Renamed to `logInvalidOrFailedNotification`

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
ivannov: Changed to `ignored`

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -20,13 +37,116 @@
...
+            try {
+                liveBlockArchiveTask.init();
+            } catch (S3ClientException | IOException e) {
+                LOGGER.log(DEBUG, "Failed to initialize live block archive task", e);
+                // Try to initialize the task again with next block
+                liveBlockArchiveTask = null;
+                context.blockMessaging()
+                        .sendBlockPersisted(new PersistedNotification(
+                                notification.blockNumber(), false, 1_000, BlockSource.CLOUD_ARCHIVE));
+                return;
```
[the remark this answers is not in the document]
jsync-swirlds: Agreed, here.
We should either send notification or add the block to a stash. We should not do both for the same block.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -1,32 +1,267 @@
...
+    /// when the notification is `null`, carries a failed verification, or has a `null` block.
+    private void logInvalidOrFailedNotification(VerificationNotification notification) {
+        if (notification == null) {
+            LOGGER.log(TRACE, "Received null verification notification, ignoring");
+        } else if (!notification.success()) {
+            LOGGER.log(TRACE, "Received failed verification notification, ignoring");
+        } else {
+            LOGGER.log(TRACE, "Received verification notification with null block, ignoring");
+        }
+    }
```
jsync-swirlds: All of these are probably `INFO` logs.
They represent situations that should not happen and may deserve human attention.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -20,13 +39,200 @@
...
+        drainPendingToQueue();
+    }
+
+    @Override
+    public void stop() {
+        // Cancelling with true interrupts the virtual thread, causing blockQueue.take() to throw
+        // InterruptedException inside BlockUploadTask.call()
+        if (currentUploadFuture != null) {
+            currentUploadFuture.cancel(true);
+        }
```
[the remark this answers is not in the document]
jsync-swirlds: After cancel, we _should not_ call `get()` as that may throw `CancellationException`.
We _should_, however, call `isDone` to see if it actually ended.
We _might_ retry a few times with a microsecond park in between if isDone isn't immediately true.

jsync-swirlds (commented):
Good progress, a few clarifications below.

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStorageConfig.java
```diff
@@ -0,0 +1,36 @@
...
+package org.hiero.block.node.cloud.archive;
+
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import com.swirlds.config.api.validation.annotation.Max;
+import com.swirlds.config.api.validation.annotation.Min;
+import org.hiero.block.node.base.Loggable;
+
+// spotless:off
+@ConfigData("cloud-archive")
```
ata-nas: We do not use `-` in config keys. We can use `.`
```suggestion
@ConfigData("cloud.archive")
```

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/BlockUploadTask.java
```diff
@@ -0,0 +1,337 @@
...
+                        blocksInBuffer.put(blockNum, blockData.source());
+                    }
+                } catch (InterruptedException e) {
+                    LOGGER.log(TRACE, "Block upload task interrupted", e);
+                    try {
+                        cleanupEmptyUpload(s3, uploadId, etags);
+                    } catch (S3ResponseException | IOException ex) {
+                        LOGGER.log(INFO, "Failed to abort multipart upload after interruption", ex);
+                    }
+                    throw e;
```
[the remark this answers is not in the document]
ata-nas: Yeah, as I mentioned before, if we have a valid reason to catch an exception and can/must do something about it in the task, we should absolutely do so. It is exceptions that we cannot do anything about or do not expect that we should not be catching.

Now, if catching is important so that we return the correct result, that is a valid reason, at least in my opinion. It is up to the plugin then to do something about the result.

ata-nas (dismissed):
Great start here. Lots of things to do, but I think we can start with this base. We have immediate follow ups, so we will be able to address issues.

A couple of general notes:
1. When using concurrent collections/maps, we have to be aware that they can at any point return `null` when querying for references or removing. We should harden with null checks.
2. Concurrent collections/maps do not allow `null` values, so this should also be something we have to assert will not happen.
3. We should never throw an exception while handling a notification. In general, plugins should not throw, but handle all types of failures, even ones we cannot recover from (here I would add that we are still missing some additions needed for health facility, which will also include the ability for plugins to notify unhealthy).

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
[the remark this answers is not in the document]
ata-nas: We can address in a follow up.
@ivannov can you take a note or a todo?

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
ata-nas: @ivannov can we add a todo here?

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/BlockUploadTask.java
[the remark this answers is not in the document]
ata-nas: We can improve in a follow up.

jsync-swirlds (dismissed):
This looks good to me as a foundation for further work.
Let us make sure @ata-nas concerns are fully addressed.

docs/block-node/configuration.md
```diff
@@ -80,6 +80,19 @@ Prometheus-format metrics. Its properties are prefixed with
 
 ## Configurations By Plugin
 
+### Archive Cloud Storage Plugin Configuration
+
+| ENV Variable                 | Description                                                                                                                                                                                              |    Default |
+|:-----------------------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------:|
+| CLOUD_ARCHIVE_GROUPING_LEVEL | Files per archive in powers of ten (1=10, 2=100, …, 6=1,000,000).                                                                                                                                        |          5 |
+| CLOUD_ARCHIVE_PART_SIZE_MB   | The size of each multi-part upload part in megabytes. Minimum value is 5, maximum value is 2047                                                                                                          |         10 |
```
Nana-EC, line 88: nit: to make config a bit more descriptive
```suggestion
| CLOUD_ARCHIVE_PART_SIZE_RANGE_MB   | The size of each multi-part upload part in megabytes. Minimum value is 5, maximum value is 2047                                                                                                          |         10 |
```

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStorageConfig.java
```diff
@@ -0,0 +1,36 @@
...
+    @Loggable @ConfigProperty(defaultValue = "") String regionName,
+    @ConfigProperty(defaultValue = "") String accessKey,
+    @ConfigProperty(defaultValue = "") String secretKey,
+    @Loggable @ConfigProperty(defaultValue = "") String bucketName,
+    @Loggable @ConfigProperty(defaultValue = "STANDARD") S3StorageClass storageClass) {
+    // spotless:on
+
+    // Source for storage class values: https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-class-intro.html
+    enum S3StorageClass {
+        STANDARD,
```
Nana-EC, line 26: Shrink this down to the ones we have high confidence and need for. If we haven't tested it we should have it here but the one we absolutely need should stay.
As we and others test we can expand

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -1,32 +1,268 @@
...
+
+    /// The block node context, set during [init].
+    private BlockNodeContext context;
+    /// The plugin configuration, set during [init].
+    private ArchiveCloudStorageConfig config;
+    /// Executor used to run each [BlockUploadTask] on a virtual thread.
+    private ExecutorService virtualThreadExecutor;
+    /// Whether this plugin successfully registered itself as a block notification handler.
+    /// Used in [stop] to avoid unregistering when registration never happened.
+    private boolean handlerRegistered = false;
```
Nana-EC, line 54: Other alternative are 
- catch the right exception
- registration handler being able to say if you are registered or not
- registration handler not throwing and unregistering if you are registered or handling it with no negative side effects if you aren't

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -1,32 +1,268 @@
...
+        } else {
+            // register to listen to block notifications if config is valid
+            context.blockMessaging().registerBlockNotificationHandler(this, false, "Archive Cloud Storage");
+            handlerRegistered = true;
+        }
+    }
+
+    /// Validates the plugin configuration and returns a list of human-readable violation messages
+    /// for any required fields that are empty.  An empty list means the configuration is valid.
+    private List<String> validateConfig() {
```
Nana-EC, line 112: Shouldn't this be handled in the Config class?

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -1,32 +1,268 @@
...
+                    }
+                } catch (InterruptedException ignore) {
+                    Thread.currentThread().interrupt();
+                } catch (ExecutionException e) {
+                    LOGGER.log(WARNING, "Block upload task failed", e);
+                    // TODO(1166) Handle properly block upload task failure
+                }
+            }
+            // If the task was cancelled (i.e. stop() was called), do not start a new task or process
+            // the current block since the plugin is shutting down.
```
Nana-EC, line 169: Why is the plugin shutting down as a result of a cancelled task?
Is the assumption that the only cause of a cancelled task is shutdown?

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
```diff
@@ -1,32 +1,268 @@
...
     @Override
     public void start() {
-        LOGGER.log(DEBUG, "Archive cloud storage plugin started");
+        virtualThreadExecutor = context.threadPoolManager().getVirtualThreadExecutor();
+        LOGGER.log(TRACE, "Archive cloud storage plugin started");
+    }
+
+    /// {@inheritDoc}
+    @Override
+    public void handleVerification(VerificationNotification notification) {
```
Nana-EC, line 141: I think let's wrap the method contents in a try catch with a catch for InterruptedException (from BlockingQueue.put()) and NoSuchElementException from ConcurrentSkipListMap navigations.
Some of the places that throw are in smaller methods so catching here is safe.

jsync-swirlds (dismissed):

Nana-EC (approved):
Nice start, I did a quick review and had some minor suggestions.
Atanas suggestions I agree with especially around NPEs

docs/block-node/configuration.md
ivannov: I think _range_ is misleading. It's exactly the size of a part

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStorageConfig.java
ivannov: Do we know which are these classes with highest confidence? If we know, I will leave only those

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
ivannov: I thought that I chose the simplest solution for now.

If you ask me, I would make the block messaging facility silently skip trying to unregister handler that was never registered.

Which looking at the code, it does. So maybe I can remove `handlerRegistered` even now (well, in the next PR)

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
ivannov: Good idea, I will move it there in the upcoming PR

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
ivannov: Yes, so far this is the assumption

block-node/archive-cloud-storage/src/main/java/org/hiero/block/node/cloud/archive/ArchiveCloudStoragePlugin.java
ivannov: The error handling will be handled with #1166

Nana-EC (dismissed):

rockysingh (approved):

Nana-EC (approved):

PR #2489, opened 2026-04-01
chore(verification): Upgrade protobuf sources to CN 0.73.0-rc.1 with TSS Wraps fixtures

AlfredoG87:
## Summary

- Upgrades protobuf sources to CN `0.73.0-rc.1` and TSS cryptography library to **`3.7.10`**
- Replaces all legacy test block fixtures (HAPI 0.66-0.72, old TSS) with a single set of CN 0.73 blocks that exercise the full TSS Wraps lifecycle: genesis (block 0), sequential pre-settled (1-4), Schnorr-to-Wraps transition (466), and post-settled WRAPS (467)
- Fixes oversized block handling: `TestBlockNodeServer` now streams block items in 2MB batches; `MAX_BLOCK_SIZE_BYTES` increased to 100MB
- Adds missing module export for `blockrecords.legacy` package introduced in CN 0.73

## Test plan

- [x] All verification tests pass with new CN 0.73 fixtures (TSS bootstrap, Schnorr and WRAPS signatures)
- [x] Backfill test handles oversized transition block 466 via batched streaming
- [x] Block access service retrieves and serves the ~90MB transition block
- [x] No regressions in existing tests

### DIFF
--- a/block-node/verification/src/test/java/org/hiero/block/node/verification/session/impl/TssBlockProofVerificationTest.java
+++ b/block-node/verification/src/test/java/org/hiero/block/node/verification/session/impl/TssBlockProofVerificationTest.java
@@ -63,7 +63,7 @@ class TssBlockProofVerificationTest {
     }
 
     @Test
-    void shouldVerifyTssWrapsBlock0_beforeSettled() throws ParseException {
+    void shouldVerifyTssWrapsBlock0BeforeSettled() throws ParseException {
         Bytes hash = computeBlockHash(wrapsBlock0, null);
         Bytes signature = extractSignature(wrapsBlock0);
         // genesis: vk (1096) + blsSig (1632) + aggregate Schnorr (192) = 2920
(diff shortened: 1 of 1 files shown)


block-node/app/src/testFixtures/resources/test-blocks/CN_0_73_TSS_WRAPS/466.blk.gz
jsync-swirlds: Just a reminder that large files (anything over 1M) have a significant impact on CI and every developer.  Removing them doesn't change that, either, because that data size remains in the git history forever, and every clone pulls the entire history of the codebase.
To the extent possible, we should remain diligent to avoid storing any file over 1M in the codebase.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
```diff
@@ -16,10 +16,11 @@
  */
 public interface BlockAccessor extends AutoCloseable {
     /**
-     * Maximum protobuf parse size in bytes (36 MB). This accommodates the largest known
-     * block items (TSS Wraps) which can exceed PBJ's default 2 MB limit.
+     * Maximum protobuf parse size in bytes (100 MB). This accommodates the largest known
+     * block items (TSS Wraps transition blocks) which can exceed PBJ's default 2 MB limit.
      */
-    int MAX_BLOCK_SIZE_BYTES = 37_748_736;
+    int MAX_BLOCK_SIZE_BYTES = 100 * 1024 * 1024;
```
jsync-swirlds: We definitely need to research a better alternative than just allowing for enormous blocks.

PR #2492, opened 2026-04-01
feat: Persist and expose TssData

### LINKED ISSUE #1947: Create a new "Detail Status" API and move version information there
### Story Form

As a Block Node user
I want to Query basic and detailed status separately
So that I have a fast/efficient query for frequently changing values and a separate query for deeper detail and rarely changing values.

### Technical Notes

* Version information should be extended to include plugin versions, features, and details.
* Basic status should be last known block, earliest known block, and a few global flags. Focus is to provide just enough for publishers and subscribers to determine if a block node is even close to needed ranges.
* Detail status should include all stored ranges, version information, plugin details, information from other services (like state management), and possibly plugin-specific status sections.
   * Detail status might be exposed via ReST for healthcheck or other purposes.

#### Initial Suggestion
Attached here an [initial suggested proto file change](https://github.com/user-attachments/files/24066604/node_service.proto.txt) for `node_service.proto` (note the attachment file name ends in `.txt` because github is nonsensical about file extensions and doesn't allow protobuf code attachments).

Note, after discussion we are of the opinion that the detail status request should include boolean flags to select whether larger data responses are or are not included in the response.

We may also want to include an integer limit to how many responses to return for (e.g.) ranges.

### LINKED ISSUE #2297: Persist and expose TSS bootstrap data locally
### Story Form
As a block node operator,
I want to adjust the plugin to publish retrieved TSS data in block node context and provide it to the `ServerStatusServicePlugin` to persist locally and publish to clients,
So that the data is durable across restarts and exposed to peer block nodes via `serverStatusDetail`.

### Technical Nodes
- Data flow -- `TSSBootstrapPlugin` → Informs `App` → tells `ServerStatusServicePlugin` via `BlockNodePlugin.updateContext` → persists locally → exposed in `serverStatusDetail`
   - Note: Verification can use the same pathway to update and persist TSS data when
     it reads block `0` in a Genesis network.
- This enables TSS data propagation: once a BN has bootstrapped, other BNs can bootstrap from it
- Consider using `BlockNodeContext` or a shared facility for the bootstrap → status data flow
- Data is persisted for future use by release 74 (TSS verification activation)

### Additional Items
- Provide TSS data to 
- Expose TSS data via shared context so that `ServerStatusServicePlugin` so it can expose the fields in `serverStatusDetail` and persist to local storage
   - `VerificationServicePlugin` will consume this data also, in a future release
- Status plugin persists TSS roster, verification key, ledger ID, and enablement block number to local storage
- TSS data is stored and exposed — **not used for block verification yet** (RSA remains active)
- Guard against applying stale data (validate enablement block is ahead of local head)
- Log successful bootstrap with details (enablement block, roster node count, key fingerprint)

### LINKED ISSUE #2298: Automatic TSS bootstrap on startup
### Story Form
As a block node operator,
I want to set up TSS bootstrap to run automatically on startup when configured
So that nodes self-bootstrap without manual intervention.

### Technical Notes
- Startup flow
   1. Check local storage for persisted TSS data
   2. If missing → query peer BN's `serverStatusDetail` → extract TSS fields → persist → provide to status plugin
   3. If present → load → validate against peer → provide to status plugin
   4. If peer has no TSS fields → skip (TSS not enabled on network)
- Once bootstrapped, this node's `serverStatusDetail` will include TSS fields, enabling other BNs to bootstrap from it

### Additional Items
- On `start()`, if TSS data is not already persisted locally, query peers to bootstrap
- After successful bootstrap, provide TSS data to `ServerStatusServicePlugin` so this node can serve as a bootstrap source for others
- If already persisted, load from local storage, validate against peer status (detect roster changes), and provide to `ServerStatusServicePlugin`
- Skip bootstrap entirely if TSS is not enabled on the network (peer status has no TSS fields)
- Emit health check status reflecting bootstrap state (not ready until bootstrap completes or is skipped)

berryware:
## Reviewer Notes
- add interface ApplicationStateFacility
- Add ApplicationStateFacility to the BlockNodeContext
- Add ApplicationStateConfig to NodeConfig
- modify BlockNodePlugin.init() to accept an ApplicationStateFacility
- modify BlockNodeApp to implement ApplicationStateFacility
- add TssBootstrapConfig
- have the BlockNodeApp call onContextUpdate for all plugins when it receives a call to updateTssData

## Related Issue(s)

Closes #2297
Closes #2298
Closes #1947


block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -73,7 +83,7 @@ public class BlockNodeApp implements HealthFacility {
     /** Should the shutdown() method exit the JVM. */
     private final boolean shouldExitJvmOnShutdown;
     /** The block node context. Package so accessible for testing. */
-    final BlockNodeContext blockNodeContext;
+    BlockNodeContext blockNodeContext;
```
Nana-EC, line 85: Should this be volatile.
As is the reference visibility I believe aren't guaranteed and is subject to race conditions, because you only guard the writes you're good there but not on on reads.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -48,7 +55,10 @@
 import org.hiero.metrics.core.MetricRegistry;
 
 /** Main class for the block node server */
-public class BlockNodeApp implements HealthFacility {
+public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
+    /** rentrant lock to allow plugins to update context data for this BlockNodeApp */
+    ReentrantLock lock = new ReentrantLock();
```
Nana-EC:
```suggestion
    private final ReentrantLock lock = new ReentrantLock();
```

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -48,7 +55,10 @@
 import org.hiero.metrics.core.MetricRegistry;
 
 /** Main class for the block node server */
-public class BlockNodeApp implements HealthFacility {
+public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
+    /** rentrant lock to allow plugins to update context data for this BlockNodeApp */
```
Nana-EC:
```suggestion
    /** reentrant lock to allow plugins to update context data for this BlockNodeApp */
```

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +358,83 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
     }
+
+    /**
+     * Allow plugins to update the TssData for this BlockNodeApp
+     *
+     * @param tssData - The TssData to be updated on the `BlockNodeContext`
+     */
+    @Override
+    public void updateTssData(TssData tssData) {
+        lock.lock();
```
Nana-EC: Q: Are we locking too many processes here?
I'm concerned about deadlock potential.
What if other plugins in their onContentUpdate call updateTssData()?

Should we instead lock only the `updateBlockNodeContext(tssData);`?

block-node/tss-bootstrap/src/main/java/org/hiero/block/node/tss/bootstrap/TssBootstrapPlugin.java
```diff
@@ -1,15 +1,121 @@
...
     @Override
     public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
-        // todo: see if the statusDetail TSS Data file exists
-        // todo: get the tss information and add to context
+        this.context = context;
+        this.applicationStateFacility = Objects.requireNonNull(this.context.applicationStateFacility());
+        TssBootstrapConfig tssBootstrapConfig = this.context.configuration().getConfigData(TssBootstrapConfig.class);
+
+        // process the config data
+        processTssDataConfiguration(tssBootstrapConfig);
```
Nana-EC, line 44: I still wonder here if it's okay to call this part way through and before some plugins have initialized.
Did we consider plugin start() to ensure that all plugins are fully initialized prior to this call?

block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
jsync-swirlds: This is a bit too early; it's best to keep the protected area small.
I'd recommend moving the spotless _off_ to just before the class annotation for `@ConfigData`.

Nana-EC (commented):
Nice adds, some questions and suggestions

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +358,83 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
+    private void persistTssData(TssData tssData) {
+        final var tssDataFilePath =
+                blockNodeContext.configuration().getConfigData(NodeConfig.class).tssDataFilePath();
+        try {
+            Files.createDirectories(tssDataFilePath.getParent());
+            Bytes serialized = TssData.PROTOBUF.toBytes(tssData);
+            Files.write(tssDataFilePath, serialized.toByteArray());
+            LOGGER.log(INFO, "Persisted TssData to file: {0}", tssDataFilePath);
+        } catch (IOException e) {
+            LOGGER.log(WARNING, "Failed to persist TssData to {0}: {1}".formatted(tssDataFilePath, e.getMessage()), e);
```
jsync-swirlds: We try to not use `Exception.getMessage()` because is isn't filled in in far too many case.
Allowing the `toString()` method to be called is generally more useful as it includes the type.
Also, you're mixing MessageFormat and StringFormat; which won't work correctly.

```suggestion
            LOGGER.log(WARNING, "Failed to persist TssData to %s: %s".formatted(tssDataFilePath, e), e);
```

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: If this is only accessed on a single thread, this should be fine. If that is not the case, then if reads happen when the lock is acquired, we should be fine again. If reads can happen outside the lock, this should be volatile.

Please take this with a grain of salt, I could be wrong, or somewhat inaccurate here.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: I share your concern here. We might need to think more carefully here.

block-node/tss-bootstrap/src/main/java/org/hiero/block/node/tss/bootstrap/TssBootstrapPlugin.java
berryware: ApplicationStateFacility and all the plugins can now init without getting any context updates. Any TssData updates they submit will be added to a ConcurrentLinkedQueue. After plugin initialization, the ApplicationStateFacility takes the TssData with the greatest validFromBlock and calls the onContextUpdate method for each plugin.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Any TssData updates submitted by plugins will be added to a ConcurrentLinkedQueue. ApplicationStateFacility is now running a virtual thread that periodically checks for TssData updates and processed them on that separate thread. The Reentrant lock is no longer in use..

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Removed

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Removed.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: BlockNodeContext is a record. Records Are Thread-Safe Without volatile.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
berryware: Done

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +365,125 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
+                    blockNodeContext.configuration(),
+                    blockNodeContext.metricRegistry(),
+                    blockNodeContext.serverHealth(),
+                    blockNodeContext.blockMessaging(),
+                    blockNodeContext.historicalBlockProvider(),
+                    blockNodeContext.applicationStateFacility(),
+                    blockNodeContext.serviceLoader(),
+                    blockNodeContext.threadPoolManager(),
+                    blockNodeContext.blockNodeVersions(),
+                    tssData);
```
jsync-swirlds, line 453: I wonder if we might need a `BlockNodeContextBuilder` with a copyFrom method so we can make updates like this a little cleaner?
Probably not worth doing that in this PR, but might be worth filing an issue to add.

Example of how this could then be simplified:
```
blockNodeContext = BlockNodeContextBuilder.copyFrom(blockNodeContext).withTss(tssData).build();
```

berryware: [#2611](https://github.com/hiero-ledger/hiero-block-node/issues/2611) has been created

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: Agreed. Specifically, records are immutable value objects, so the only write is in the constructor (which happens-before all reads) and all reads are thread-safe and consistent.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +368,136 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
+                this::checkForTssUpdates,
+                nodeConfig.tssUpdateInitialDelay(),
+                nodeConfig.tssUpdateScanInterval(),
+                TimeUnit.MILLISECONDS);
+    }
+
+    private void checkForTssUpdates() {
+        boolean updated = false;
+        while (!tssDataUpdates.isEmpty()) {
+            updated |= updateBlockNodeContext(tssDataUpdates.poll());
```
Nana-EC, line 409: Q: is polling a concurrent queue the right approach here? 
On one hand honoring the individual updates is good but on the other we're not using this anywhere but when loading the data so isn't it only the last instance that matters, and we aren't persisting each change.
Do we need to keep track of all the changes or just the latest and therefore a single update?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +368,136 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
+        // Create thread executors via threadPoolManager.
+        applicationStateExecutor = blockNodeContext
+                .threadPoolManager()
+                .createVirtualThreadScheduledExecutor(1, "ApplicationStateScanner", handler);
+
+        NodeConfig nodeConfig = blockNodeContext.configuration().getConfigData(NodeConfig.class);
+
+        // Schedule periodic gap detection task using autonomous executor
+        applicationStateExecutor.scheduleAtFixedRate(
+                this::checkForTssUpdates,
```
Nana-EC: This feels very TSS specific, should be more general and check if any tracked application state has been updated

```suggestion
                this::checkForApplicationStateUpdates,
```

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ApplicationStateFacility.java
```diff
@@ -0,0 +1,20 @@
...
+ * is passed to all block node plugins in the `init()` call.
+ * */
+public interface ApplicationStateFacility {
+    /**
+     * Used by plugins to update the TssData for this application. i.e. `TssBootstrapPlugin`, and `VerificationPlugin`
+     * The update will be forwarded to all plugins using the BlockNodePlugin.onContextUpdate() of the plugins
+     *
+     * @param tssData - The TssData to be updated on the `BlockNodeContext`
+     * */
+    default void updateTssData(TssData tssData) {
```
Nana-EC: Q: Did we consider if this is better as an abstract class with default implementations for updateTssData and check for updates here?

block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
```diff
@@ -17,8 +16,16 @@
...
+ *     checking to see if there are any {@code TssData} updates to process.
  */
+// spotless:off
 @ConfigData("block.node")
-public record NodeConfig(@Loggable @ConfigProperty(defaultValue = "0") @Min(0) long earliestManagedBlock) {}
-
+public record NodeConfig(
+        @Loggable @ConfigProperty(defaultValue = "0") @Min(0) long earliestManagedBlock,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/tss-data.bin") Path tssDataFilePath,
+        @Loggable @ConfigProperty(defaultValue = "500") @Min(100) long tssUpdateScanInterval,
```
Nana-EC: I think the intervals and delay should be generalized for applicationState and not just Tss

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/BlockNodePlugin.java
```diff
@@ -91,6 +91,8 @@ default PluginVersion version() {
     /**
      * Notify the plugin of an update to the BlockNodeContext. This method is called when the block node context has
      * been updated. It provides the plugin an opportunity to update its state when the BlockNodeContext changes.
+     * `onContextUpdate()` will be called from a different thread than the thread the plugin was started on. Care must be
+     * taken when updating internal state.
```
Nana-EC: Add a note to highlight that plugins should make their local blockNodeContext copy volatile.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssConfig.java
```diff
@@ -0,0 +1,25 @@
...
+import java.nio.file.Path;
+import org.hiero.block.node.base.Loggable;
+
+/// Configuration for the tss-bootstrap module.
+///
+@ConfigData("tss.bootstrap")
+public record TssBootstrapConfig(
+        @Loggable @ConfigProperty(defaultValue = "") String ledgerId,
+        @Loggable @ConfigProperty(defaultValue = "") String wrapsVerificationKey,
+        @Loggable @ConfigProperty(defaultValue = "0") long nodeId,
```
Nana-EC, line 16: It's unclear why the Tss data has one nodeId-weight-schnorrPublicKey tuple.
Usually it'd be a list forming the roster.
What's the significance of one node here?

Nana-EC (commented):
Some suggestions.
Non blocking as I'd like to get this in and build upon it with the `roster-bootsrap-rsa` plugin that will follow a similar style and bootstrap the address book.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +368,136 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
     }
+
+    /**
+     * Allow plugins to update the TssData for this BlockNodeApp
+     *
+     * @param tssData - The TssData to be updated on the `BlockNodeContext`
+     */
+    @Override
+    public void updateTssData(TssData tssData) {
+        if (tssData != null) tssDataUpdates.add(tssData);
```
ata-nas, line 379: Should we do something about nulls? Or are they expected?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +368,136 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
+    }
+
+    /**
+     * Starts the ApplicationStateFacility. The thread will be used to check if there are any TssData updates to
+     * process.
+     */
+    void startApplicationStateFacility() {
+        LOGGER.log(INFO, "ApplicationStateFacility start called");
+        Thread.UncaughtExceptionHandler handler =
+                (thread, e) -> LOGGER.log(INFO, "Uncaught exception in thread: " + thread.getName(), e);
```
ata-nas, line 389: We could think of a bit more non-ambiguous general message for the exception handler, something maybe related to the application state facility? Nit.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas, line 409: The poll could be null here, there is no guarantee that the queue is not empty after the check. I guess that is why we have a null check on my previous comment.

ata-nas, line 409: I'd argue that the bitwise operator is a bit misleading and ambiguous.

Just to understand, do we want to pass the `if (updated)` check below if we have at least one update, or pass based on the last attempt? My only concern is a possible race condition (even if we do not have it today, possibly) where we say the queue is not empty, but poll returns null.

If we aim to pass the below updated check in case where we have at least one update, then I think a more conventional approach will be safer and more readable.

berryware: I use the concurrent queue because multiple threads could call updateTssData.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssConfig.java
berryware: I just thought we would only need one roster entry for testing. How many should there be?

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/BlockNodePlugin.java
berryware: Added.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
berryware: Done.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ApplicationStateFacility.java
berryware: I removed the default completely. It was only there because I originally passed this into init() which touches all plugins, so a lot of test code was updated. Right now it is only included in 3 places and there is no common default.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: Gotcha, question was more about whether we want to have a queue at all to manage multiple or if it should be just the last.
Offline we agreed WRAPS verification only cares about the latest but aggregate Schnorr signatures does care for the specific roster as they care about weight changes.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssConfig.java
Nana-EC: The roster would contain all active nodes participating in consensus at the time, so it depends on network size e.g. mainnet is 32 I believe and other envs are 7

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: This is on initial load (during `init` processing), which is different from updates.
On `init`, we need to gather all of them and select the one with the latest block number after init is complete.  This is to handle that different sources/causes of `init` updates might be in random order.

jsync-swirlds: We do not currently have the ability to store multiple data sets over time.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssConfig.java
jsync-swirlds: Also, we want to support hundreds, eventually, so configuration should be able to load a single `Roster` entry from disk with potentially many nodes and the "as of block" value.
Configuration should just point to where the file is stored.
The file needs its own protobuf definition.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
```diff
@@ -17,8 +16,16 @@
...
  */
+// spotless:off
 @ConfigData("block.node")
-public record NodeConfig(@Loggable @ConfigProperty(defaultValue = "0") @Min(0) long earliestManagedBlock) {}
-
+public record NodeConfig(
+        @Loggable @ConfigProperty(defaultValue = "0") @Min(0) long earliestManagedBlock,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/app-state-data.bin") Path appStateDataFilePath,
+        @Loggable @ConfigProperty(defaultValue = "500") @Min(100) long appStateUpdateScanInterval,
+        @Loggable @ConfigProperty(defaultValue = "100") @Min(100) int appStateUpdateInitialDelay) {}
```
jsync-swirlds, line 30: (non-blocking) We might want a separate `ApplicationStateConfig` for these three values.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +368,139 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
```
jsync-swirlds, line 389: (non-blocking) We should add a proper uncaught exception handler (actual class, not just a lambda) that does more than just log exceptions.
This probably belongs with the upcoming application health/shutdown updates.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssConfig.java
```diff
@@ -0,0 +1,22 @@
...
+///
+@ConfigData("roster.bootstrap.tss")
+public record RosterBootstrapTssConfig(
+        @Loggable @ConfigProperty(defaultValue = "") String ledgerId,
+        @Loggable @ConfigProperty(defaultValue = "") String wrapsVerificationKey,
+        @Loggable @ConfigProperty(defaultValue = "0") long nodeId,
+        @Loggable @ConfigProperty(defaultValue = "0") long weight,
+        @Loggable @ConfigProperty(defaultValue = "") String schnorrPublicKey,
+        @Loggable @ConfigProperty(defaultValue = "0") long validFromBlock,
+        @Loggable @ConfigProperty(defaultValue = "0") long rosterValidFromBlock) {}
```
jsync-swirlds, line 20: Probably these should look like `TssData`, rather than being flattened and losing the repeated roster entries.

block-node/roster-bootstrap-tss/build.gradle.kts
```diff
@@ -3,6 +3,12 @@ plugins { id("org.hiero.gradle.module.library") }
 
 description = "Hiero Block Node TSS Bootstrap Service"
 
+dependencies {
+    implementation(project(":base"))
+    implementation(project(":common"))
+    implementation(project(":common"))
```
jsync-swirlds: duplicate?

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
jsync-swirlds, line 1: There might be a confusion item here in the original task description(s).
The "configured" data source should just be a serialized `TssData` loaded from local storage.
Rather than trying to add the fields as configuration (which isn't good at complex structure) the configuration should be a source location and this code can then just load the file and parse the proto data into a `TssData` object.

jsync-swirlds (commented):
Might want to make a change to the "configured" data source and load.
Otherwise just some small non-blocking improvements.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +368,139 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
+        }
+        applicationStateExecutor.shutdownNow();
+        try {
+            if (!applicationStateExecutor.awaitTermination(5, TimeUnit.SECONDS)) {
+                final String executorTerminationMsg = "applicationStateExecutor did not terminate in time";
+                LOGGER.log(INFO, executorTerminationMsg);
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
```
ata-nas: Optional nit:
We could skip a return and make this more readable if we negate the condition.
```suggestion
        if (applicationStateExecutor != null) {
            applicationStateExecutor.shutdownNow();
            try {
                if (!applicationStateExecutor.awaitTermination(5, TimeUnit.SECONDS)) {
                    final String executorTerminationMsg = "applicationStateExecutor did not terminate in time";
                    LOGGER.log(INFO, executorTerminationMsg);
                }
            } catch (InterruptedException e) {
                Thread.currentThread().interrupt();
            }
        }
```

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +368,139 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
+    /**
+     * Update the BlockNodeContext with the new TssData if the validFromBlock is greater than that of the context
+     *
+     * @param tssData The TssData to update
+     * @return A boolean indicating if the BlockNodeContext was updated.
+     */
+    private boolean updateBlockNodeContext(TssData tssData) {
+        boolean updated = false;
+        if (blockNodeContext.tssData() == null
+                || blockNodeContext.tssData().validFromBlock() < tssData.validFromBlock()) {
```
ata-nas, line 442: Based on the code I see, the `tssData` could be `null` as we blindly poll from a queue. We should handle that case here. As mentioned before, we cannot rely that checking for empty will be true by the time we poll as we asynchronously update the collection.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +368,139 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
+        return updated;
+    }
+
+    /**
+     * Persist the TssData
+     * Persists the TssData to the file path specified in the NodeConfig class.
+     *
+     * @param tssData The TssData to persist
+     */
+    private void persistTssData(TssData tssData) {
```
ata-nas, line 466: Same here, we assume `tssData` is not `null`.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas, line 1: We seem to thrown in a couple of places. Please note that throwing should not generally happen. If we have to stop all activity, then we can throw for now, until we have updates to the health facility. Otherwise, if we can move forward, we can reconsider throwing.

Not asking for a change, just noting.

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
```diff
@@ -93,6 +101,15 @@ public <C> Stream<? extends C> loadServices(Class<C> serviceClass) {
         blockNodeApp = spy(new BlockNodeApp(serviceLoaderFunction, false));
     }
 
+    @AfterEach
+    void cleanup() {
+        try {
+            Files.deleteIfExists(Path.of("build/tmp/data/block/node/app-state-data.bin"));
```
ata-nas, line 107: Something for the future, we should be using `jimfs` for tests. We can consider to switch to taht in the future.

Also, if we cannot use that for some reason, we can use a tempdir or a tempfile from junit, we will not need a cleanup here.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssConfig.java
```diff
@@ -0,0 +1,22 @@
...
+// spotless:off
+
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import org.hiero.block.node.base.Loggable;
+
+/// Configuration for the roster-bootstrap-tss module.
+///
+@ConfigData("roster.bootstrap.tss")
+public record RosterBootstrapTssConfig(
```
ata-nas, line 13: Shall we add some validation to these fields? My understanding is that some strings could be blank and that is ok, but I would imagine there are some other typed values that, if configured improperly, that is most likely an operator error.

protobuf-sources/src/main/proto/block-node/api/node_service.proto
```diff
@@ -197,6 +198,10 @@ message TssData {
      * The current TSS roster
      */
     TssRoster current_roster = 3;
+    /**
+     * The starting block number this TssData is valid from
+     */
+    uint64 valid_from_block = 4;
```
ata-nas, line 204: Why do we need the valid from here, is it not supposed to come from the current roster? I might be missing something?

ata-nas (commented):
LG, some cleanup suggestions.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: @jsync-swirlds is correct. The concurrent queue gathers up all the updates from plugins during `init()`. The queue is then used to update the latest `TssData` based on  'validFromBlock`.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: They are expected. pbj does not do a good job with default implementations that can be tested to determine if it is set or not. Using null here to indicate uninitialized.

protobuf-sources/src/main/proto/block-node/api/node_service.proto
berryware: This is for ledger_id changes and wraps_verification_key changes

block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssConfig.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssConfig.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

berryware: captured in [2645](https://github.com/hiero-ledger/hiero-block-node/issues/2645)

block-node/roster-bootstrap-tss/build.gradle.kts
berryware: removed

berryware: removed

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssConfig.java
berryware: captured in https://github.com/hiero-ledger/hiero-block-node/issues/2645

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -344,4 +368,138 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
+            try {
+                Bytes fileBytes = Bytes.wrap(Files.readAllBytes(tssDataFilePath));
+                TssData tssData = TssData.PROTOBUF.parse(fileBytes);
+                updateTssData(tssData);
+                LOGGER.log(INFO, "Loaded Application State Data from file: {0}", tssDataFilePath);
+            } catch (IOException e) {
+                throw new UncheckedIOException("Failed to read Application State Data file: " + tssDataFilePath, e);
+            } catch (ParseException e) {
+                throw new IllegalStateException("Failed to parse Application State Data file: " + tssDataFilePath, e);
+            }
```
jsync-swirlds, line 502: Instead of throwing another exception, perhaps just log the failure and don't update the Tss Data?
I'm not sure this needs to be handled much beyond that (until we have health updates, then we can report unhealthy until TSS data is available from somewhere else).

jsync-swirlds (approved):
One small concern about exceptions.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: captured in https://github.com/hiero-ledger/hiero-block-node/issues/2645

ata-nas (approved):
LG, I think we can continue, we've captured needed follow ups in an issue.

Nana-EC (approved):
LG. ApplicationStateFacility is getting good traction and promised used in the next sprint.

PR #2493, opened 2026-04-01
chore: Cleanup of Test-Only Logic & Utilize Config Converter Overrides for Testing

### LINKED ISSUE #1209: Cleanup of Test-Only Logic & Utilize Config Converter Overrides for Testing
### Story Form

As a Block Node Developer
I want to have clean production logic that has no "test-only" logic
So that my codebase is clean, maintainable and easily extendable

### Technical Notes

**DEPENDS ON**: https://github.com/hiero-ledger/hiero-consensus-node/pull/18617
- we currently have some sub-optimal workarounds due to lack of ability to inject configurations for tests
  - here we mean that we should be able to inject configuration in a way where we do not interfere with production logic, meaning we should be able to inject any and all configurations and they should be picked up by plugins without any extra logic added to production code
- an example place that we need to cleanup is `BlocksFilesRecentPlugin` where we have a "test-only" constructor which should be removed and configs should only (and always) be resolved by the init method
- the new converter override ability introduced in the issue mentioned above will allow us to construct a `JimfsPath` which then could be supplied via a lambda converter, in fact, this was why the feature request happened in the first place

ivannov:
## Reviewer Notes

A package-private constructor that accepted a FilesRecentConfig was being used in tests to inject config directly, bypassing the normal init() path. This commit removes it and uses the standard flow instead.                                                                                           
                        
To make that work, `PluginTestBase.start()` gains an optional `FileSystem` parameter. When provided, it registers a `Path` converter on the `ConfigurationBuilder` that resolves paths through that filesystem, so tests using an in-memory filesystem (Jimfs) can now pass config values as plain strings and have them resolved correctly.                                                                                                           
                        
Net effect: production code is cleaner (no test seam in the plugin constructor)

## Related Issue(s)

Closes #1209

### DIFF
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
@@ -13,6 +13,7 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.stream.Stream;
@@ -120,7 +121,12 @@ public abstract class PluginTestBase<
             @NonNull final HistoricalBlockFacility historicalBlockFacility,
             @Nullable final List<BlockNodePlugin> additionalPlugins,
             @Nullable final Map<String, String> configOverrides,
-            @Nullable final Map<Class<?>, ConfigConverter<?>> converters) {
+            @NonNull final Map<Class<?>, ConfigConverter<?>> converters) {
+
+        Objects.requireNonNull(plugin);
+        Objects.requireNonNull(historicalBlockFacility);
+        Objects.requireNonNull(converters);
+
         this.plugin = plugin;
         org.hiero.block.node.app.fixtures.logging.CleanColorfulFormatter.makeLoggingColorful();
         // Build the configuration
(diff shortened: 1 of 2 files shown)


block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
```diff
@@ -96,19 +98,30 @@ public void start(
...
      * @param configOverrides a map of configuration overrides to be applied to the loaded configuration
+     * @param fileSystem an optional {@link FileSystem} to use for resolving {@link Path} values in configuration
      */
     public void start(
             @NonNull final P plugin,
             @NonNull final HistoricalBlockFacility historicalBlockFacility,
             @Nullable final List<BlockNodePlugin> additionalPlugins,
-            @Nullable final Map<String, String> configOverrides) {
+            @Nullable final Map<String, String> configOverrides,
+            @Nullable final FileSystem fileSystem) {
```
ata-nas: I do not think this is the right place to do this, but could be wrong.
In tests, we should be able to override the configs and add a converter there. This was the initial idea at least.

In any case, we should not be passing a filesystem to this plugin base. We should be passing a map with converters. By passing an FS we limit ourselves to converter usage, but also it is really confusing because almost all plugins do not use an FS at all.

jsync-swirlds: It doesn't seem "bad" as such to do this.
Perhaps we should offer an alternative overload here that defaults the FileSystem to the standard value.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
```diff
@@ -122,6 +134,9 @@ public void start(
                 configurationBuilder = configurationBuilder.withValue(override.getKey(), override.getValue());
             }
         }
+        for (Entry<Class<?>, ConfigConverter<?>> entry : converters.entrySet()) {
```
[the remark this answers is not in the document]
ivannov: Did the other way round, as I prefer empty containers rather then null containers. Changed the annotation to `@NonNull` and introduces `requireNonNull`

PR #2505, opened 2026-04-02
feat: Improve `blockIsEnding` and `shutdown` of Publisher

### LINKED ISSUE #1949: Revisit `blockIsEnding` Method Usage
### Story Form

As a Publisher Handler
I want to use the `blockIsEnding` method
So that I can notify the manager that a block is ending in the middle

### Technical Notes

Currently `blockIsEnding` method is called 2 times from the same flow.
 ```
else if (request.hasEndStream()) {
            try {
                handleEndStreamRequest(Objects.requireNonNull(request.endStream()));
            } finally {
                shutdown();
            }
```
First call is from `handleEndStreamRequest` -> ... -> handleEndStream
Second call is from `shutdown`

1. We need to check why we have these duplicate calls and if we need them.
2. Research if we need to use `blockIsEnding` in other flows and apply these changes.

UPDATE:
We also need to examine the use of `shutdown` in handler.
Additionally, we need to ensure that blocks are discarded _only_ when necessary (i.e. when a block ends while incomplete).

### LINKED ISSUE #2339: Improve Sending of `BAD_BLOCK_PROOF` in Manager
### Story Form

As a Publisher Manager
I want to have an improved way of telling a handler to send the bad block proof code
So that I have this optimized

### Technical Notes

- Currently we loop over all active handlers and letting them handle the bad block proof code if they have supplied the bad block.
- As an improvement, we can first determine which handler has supplied that block and then telling it to handle the sending of the code.
- This is a response to [this comment](https://github.com/hiero-ledger/hiero-block-node/pull/2227#discussion_r2898905076).

### LINKED ISSUE #2344: Improve `blockIsEnding` and `shutdown` of Publisher
### Story Form

As a PublisherHandler & PublisherManager
I want to have improved ending mid-block and shutdown logic for a publisher
So that I will be able to introduce additional improvements related to these

### Technical Notes

- With #2227 we introduced a lot of improvements for the `ResendBlock` message handling.
- This opened up opportunities to add even more improvements to the `blockIsEnding` and `shutdown` of Publisher.
- We can explore way to improve these, especially when these could be triggered by threads different than the requests one.

### PR DESCRIPTION
**Pull request by ata-nas**
## Reviewer Notes

* Centralized publisher request results
* Improvements for shutdown
  * Manager schedules shutdowns instead of forcing them
* Send SKIP instead of END_ERROR in `getActionForCurrentlyStreaming`
* Receiving an EndOfBlock for block the manager no longer needs still allows a RESEND to possibly be sent as a response

## Related Issue(s)

Closes #1949
Closes #2344 
Closes #2339

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -4,7 +4,6 @@ package org.hiero.block.node.stream.publisher;
 import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.TRACE;
-import static java.lang.System.Logger.Level.WARNING;
 import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_BLOCKS_CLOSED_COMPLETE;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_BLOCK_BATCHES_MESSAGED;
@@ -69,7 +68,6 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     private final ConcurrentNavigableMap<Long, PublisherHandler> handlers;
     private final AtomicLong nextHandlerId;
     private final ConcurrentNavigableMap<Long, Deque<BlockItemSetUnparsed>> queueByBlockMap;
-    private final AtomicLong highestBlockNumber;
     private final AtomicLong blocksClosedComplete;
     private final Condition dataReadyLatch;
     private final ReentrantLock dataReadyLock;
@@ -104,7 +102,6 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         threadManager = serverContext.threadPoolManager();
         handlers = new ConcurrentSkipListMap<>();
         nextHandlerId = new AtomicLong(0);
-        highestBlockNumber = new AtomicLong(0);
         blocksClosedComplete = new AtomicLong(0);
         queueByBlockMap = new ConcurrentSkipListMap<>();
         lastForwardedBlockNumber = new AtomicLong(-1);
@@ -167,7 +164,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             final long blockNumber, final BlockAction previousAction, final long handlerId) {
         return switch (previousAction) {
             case null -> getActionForHeader(blockNumber);
-            case ACCEPT -> getActionForCurrentlyStreaming(blockNumber, handlerId);
+            case ACCEPT -> getActionForCurrentlyStreaming(blockNumber);
             case END_ERROR, END_DUPLICATE ->
                 // This should not happen because the Handler should have shut down.
                 BlockAction.END_ERROR;
@@ -197,17 +194,16 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     public ActionForBlock endOfBlock(final long blockNumber) {
         if (queueByBlockMap.containsKey(blockNumber)) {
             endBlocksReceived.add(blockNumber);
-            final long blockToResend = nextBlockToResend();
-            if (blockToResend != UNKNOWN_BLOCK_NUMBER) {
-                return new ActionForBlock(BlockAction.RESEND, blockToResend);
-            } else {
-                return new ActionForBlock(BlockAction.ACCEPT, blockNumber);
-            }
         } else {
             // If the queue is not in the map, this means that the manager no longer
             // needs the block
             // Remove a proof in case it is collected
             blockProofs.remove(blockNumber);
+        }
+        final long blockToResend = nextBlockToResend();
+        if (blockToResend != UNKNOWN_BLOCK_NUMBER) {
+            return new ActionForBlock(BlockAction.RESEND, blockToResend);
+        } else {
             return new ActionForBlock(BlockAction.ACCEPT, blockNumber);
         }
     }
@@ -303,8 +299,6 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             blocksToResend.add(blockNumber);
             // Iterate over all handlers and attempt to send the
             // bad block proof message.
-            // @todo(2339) improve the loop below, find the handler that should send the bad block proof code
-            //    and then let it handle the failed verification.
             for (final PublisherHandler handler : handlers.values()) {
                 if (handler.handleFailedVerification(blockNumber)) {
                     // There will always be only one handler that will send the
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java [hunk 1]
```diff
@@ -155,19 +161,28 @@ public void onSubscribe(@NonNull final Subscription subscription) {
...
-            LOGGER.log(INFO, "Error processing request: %s".formatted(e), e);
-            sendEndAndResetState(Code.ERROR);
+        if (!isActive.get()) {
+            checkMidBlockAndShutdown(currentStreamingBlockNumber.get());
+        } else {
+            try {
+                LOGGER.log(TRACE, "Handler {0} received request", handlerId);
+                final PublisherRequestResult result = processNextRequestUnparsed(request);
+                result.handle();
+                LOGGER.log(TRACE, "Handler {0} finished processing request", handlerId);
```
jsync-swirlds: We need to be careful about trace logging, we already have too much, and every added log has a real performance cost.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -771,6 +728,151 @@ private boolean isCurrentlyMidBlock(final long blockInProgress) {
         return blockInProgress > UNKNOWN_BLOCK_NUMBER && currentBlockQueue.get() != null;
     }
 
+    // ==== Publisher Request Results ==========================================
+
+    /// A simple [FunctionalInterface] that represents the result of handling
+    /// a [PublishStreamRequestUnparsed]. This result must be then handled
+    /// before continuing with the next request.
+    @FunctionalInterface
```
jsync-swirlds: Does this need to be a `FunctionalInterface`?

Also, we should extract this interface and it's implementations to a separate file (`PublisherRequestResult.java`).

jsync-swirlds (commented):
Apologies for the delay reviewing this.
There are a few areas where we can make some significant improvements.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -195,17 +193,16 @@ public long getLatestBlockNumber() {
...
         } else {
             // If the queue is not in the map, this means that the manager no longer
             // needs the block
             // Remove a proof in case it is collected
             blockProofs.remove(blockNumber);
+        }
+        final long blockToResend = nextBlockToResend();
+        if (blockToResend != UNKNOWN_BLOCK_NUMBER) {
+            return new ActionForBlock(BlockAction.RESEND, blockToResend);
+        } else {
```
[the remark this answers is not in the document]
ata-nas: Not sure what is meant here. This change aims to allow blocks to be resent, even when a publisher ends a block we no longer needs.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -600,17 +594,14 @@ private long ensureNextGreaterThanPersisted(final long lastPersisted) {
...
             // We're one of the handlers currently streaming, keep going.
             return BlockAction.ACCEPT;
         } else {
-            // This is not expected
-            final String message =
-                    "Handler {0} wants to continue streaming block {1}, but it has no registered queue for the block";
-            LOGGER.log(WARNING, message, handlerId, blockNumber);
-            return BlockAction.END_ERROR;
+            // The block is no longer needed by the manager, we can skip it
+            return BlockAction.SKIP;
```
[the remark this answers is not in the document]
ata-nas: Yes, absolutely. We should do that in #1420 I think.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
ata-nas: This is from the past, we can consider to remove this soon.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: Generally it's best to only apply the functional attribute when necessary, it can produce unexpected JIT results when applied unnecessarily.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -301,8 +298,6 @@ public void handleVerification(@NonNull final VerificationNotification notificat
             blocksToResend.add(blockNumber);
             // Iterate over all handlers and attempt to send the
             // bad block proof message.
-            // @todo(2339) improve the loop below, find the handler that should send the bad block proof code
-            //    and then let it handle the failed verification.
```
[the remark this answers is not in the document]
jsync-swirlds: I think the code was OK as it was.
I'm just thinking that we shouldn't have removed the comment...

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -386,26 +421,31 @@ private void handleEndStreamRequest(final EndStream endStream) {
...
             // the switch expression so that we are forced at compile time to
             // handle all possible end stream codes.
-            handleValidEndStreamRequest(code, earliestAndLatestBlockNumbers, endStreamLatestBlockNumber);
+            result = handleValidEndStreamRequest(code, earliestAndLatestBlockNumbers, endStreamLatestBlockNumber);
         } else {
             LOGGER.log(
                     INFO,
                     "Handler %d received an invalid EndStream request with code %s. %s"
                             .formatted(handlerId, code, earliestAndLatestBlockNumbers));
+            result = new ShutdownResult(this, currentStreamingBlockNumber.get());
```
jsync-swirlds: We shouldn't shut down in this case.
I'm not sure the correct response, but I don't think shutdown is what we need.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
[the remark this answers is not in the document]
ata-nas: I see what you mean, Happy to revert as I mentioned above. The improvement here is to not remove the handler from the map while we are iterating over it. We now schedule a shutdown instead.

One of the possibilities was to split the logic so that first we determine which publisher supplied the block and then to call handle on it. I initially did not do that, but I will revert the change after making it.

I thought that you meant that we still need an improvement, not that the comment is needed.

jsync-swirlds (dismissed):

AlfredoG87 (dismissed):
LGTM, just needs a rebase before merging.

Nana-EC (dismissed):

AlfredoG87 (approved):
LGTM

jsync-swirlds (approved):

PR #2508, opened 2026-04-02
feat: add stake weight histogram to validation stats

### LINKED ISSUE #2504: feat: add stake weight histogram to validation stats
## Summary

Enhance the existing validation-with-stats output to include stake weight information when running in stake-weighted signature validation mode. This gives visibility into the stake distribution of signing nodes per block.

## Details

Now that stake-weighted signature validation is implemented (#2491), we should surface the stake weight data in the validation stats output.

### Per-block (rolling/current)
- Signing node count vs total node count
- Validated stake vs total stake (and percentage)
- Mode (stake-weighted vs equal-weight)

### Per-day aggregates

**Weight distribution:**
- Min / Max / Mean / Median node stake weight
- Std deviation of weights (shows how uneven the distribution is)
- Top-N nodes by stake (concentration metric — e.g., "top 5 nodes hold X% of total stake")

**Signature participation:**
- Min / Max / Mean signing node count per block
- Min / Max / Mean validated stake percentage per block
- Nodes that never signed (missing signers)
- Nodes that signed every block (reliable signers)

**Threshold proximity:**
- Closest-to-threshold block (lowest validated stake margin) — highlights blocks that barely passed
- Blocks where threshold was met by fewest nodes (stake concentration risk)

### Across the full range
- Stake weight change events (when `NodeStakeUpdate` transactions were applied)
- Node count changes over time
- Trend of stake concentration (is it getting more or less concentrated?)

## Scope

- Append stake weight stats to the existing validation-with-stats headers and calculations
- Validation stats with wrap+validation as a separate file
- Download and its validation as a separate file
- Integrate into the wrap pipeline so stake weight stats are available during re-wrap + re-validate flows

## Related

- Part of #2039
- Depends on #2491 (stake-weighted signature validation)

rockysingh:
## Summary
- Extract `DayStatistics` and `SignatureStats` from `ValidateWithStats` into reusable `SignatureStatsCollector` class
- Add `SignatureBlockStats` record for per-block signature validation data with `NodeResult` enum and `fromPreVerifiedData()` factory
- Extend CSV output with 17 new stake weight columns: mode, total_stake, validated stake % (min/max/mean), closest threshold margin, signing node counts (min/max/mean), node stake distribution (min/max/mean/median/stddev), top-3 concentration %, missing signers, reliable signers
- Wire stats collection into `ValidateBlocksCommand`, `LiveSequential`, and `ToWrappedBlocksCommand` with command-specific CSV filenames
- Add `StakeMapWithTotal` to `NodeStakeRegistry` to avoid recomputing total stake per block
- Optimize `SignatureValidation` with lazy diagnostics — no string formatting on the happy path
- Stats collection enabled by default on the 2-arg constructor

closes: #2504

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ValidateBlocksCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ValidateBlocksCommand.java
@@ -44,6 +44,8 @@ import org.hiero.block.tools.blocks.validation.HistoricalBlockTreeValidation;
 import org.hiero.block.tools.blocks.validation.JumpstartValidation;
 import org.hiero.block.tools.blocks.validation.NodeStakeUpdateValidation;
 import org.hiero.block.tools.blocks.validation.RequiredItemsValidation;
+import org.hiero.block.tools.blocks.validation.SignatureBlockStats;
+import org.hiero.block.tools.blocks.validation.SignatureStatsCollector;
 import org.hiero.block.tools.blocks.validation.SignatureValidation;
 import org.hiero.block.tools.blocks.validation.StreamingMerkleTreeValidation;
 import org.hiero.block.tools.blocks.wrapped.BalanceCheckpointValidator;
@@ -546,6 +548,12 @@ public class ValidateBlocksCommand implements Runnable {
             }
         }
 
+        // Create signature stats collector
+        final Path statsOutputDir =
+                Arrays.stream(files).filter(Files::isDirectory).findFirst().orElse(files[0]);
+        final SignatureStatsCollector statsCollector =
+                new SignatureStatsCollector(statsOutputDir.resolve("signature_statistics_validate_block_command.csv"));
+
         // Track validation progress
         final long startNanos = System.nanoTime();
         long blocksValidated = checkpoint != null ? checkpoint.blocksValidated() : 0;
@@ -809,6 +817,14 @@ public class ValidateBlocksCommand implements Runnable {
                         blocksValidatedRef[0] = blocksValidated;
                     }
 
+                    // Collect signature stats from the validation
+                    if (signatureValidation != null) {
+                        SignatureBlockStats blockStats = signatureValidation.popBlockStats(blockNum);
+                        if (blockStats != null) {
+                            statsCollector.accept(blockStats);
+                        }
+                    }
+
                     // Periodic registry sync for crash safety (~every 100 blocks)
                     if (registry != null && blocksValidated % 100 == 0) {
                         registry.sync();
@@ -928,6 +944,10 @@ public class ValidateBlocksCommand implements Runnable {
             for (BlockValidation v : validations) {
                 v.close();
             }
+            // Finalize signature statistics
+            statsCollector.finalizeDayStats();
+            statsCollector.printFinalSummary();
+            statsCollector.close();
         }
 
         // Determine overall result
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ValidateBlocksCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ValidateBlocksCommand.java
@@ -397,7 +397,7 @@ public class ValidateBlocksCommand implements Runnable {
         parallelValidations.add(new BlockStructureValidation());
         SignatureValidation sigVal = null;
         if (!skipSignatures) {
-            sigVal = new SignatureValidation(abRegistry, nodeStakeRegistry);
+            sigVal = new SignatureValidation(abRegistry, nodeStakeRegistry, true);
             parallelValidations.add(sigVal);
         } else {
             System.out.println(Ansi.AUTO.string("@|yellow Skipping:|@ Signature validation (--skip-signatures)"));
@@ -548,11 +548,13 @@ public class ValidateBlocksCommand implements Runnable {
             }
         }
 
-        // Create signature stats collector
-        final Path statsOutputDir =
-                Arrays.stream(files).filter(Files::isDirectory).findFirst().orElse(files[0]);
-        final SignatureStatsCollector statsCollector =
-                new SignatureStatsCollector(statsOutputDir.resolve("signature_statistics_validate_block_command.csv"));
+        // Determine signature stats output directory. If files[0] is a regular file (e.g. a
+        // .blk.zip), fall back to its parent directory so the CSV is written alongside it.
+        final Path statsOutputDir = Arrays.stream(files)
+                .filter(Files::isDirectory)
+                .findFirst()
+                .orElseGet(() -> files[0].getParent() != null ? files[0].getParent() : Path.of("."));
+        final Path statsCsvPath = statsOutputDir.resolve("signature_statistics_validate_block_command.csv");
 
         // Track validation progress
         final long startNanos = System.nanoTime();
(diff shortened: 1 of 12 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SignatureStatsCollector.java
```diff
@@ -0,0 +1,524 @@
...
+                row.append(",").append(dayStats.getMissingSigners());
+                row.append(",").append(dayStats.getReliableSigners());
+            } else {
+                // Empty stake columns
+                for (int i = 0; i < STAKE_COLUMN_COUNT; i++) {
+                    row.append(",");
+                }
+            }
+
+            row.append("\n");
```
jsync-swirlds: It is a little strange to combine both format and string builder in this way.
Perhaps use a Formatter for somewhat cleaner interactions:

```suggestion
            Formatter row = new Formatter(new StringBuilder());
            // First parameter is a date, we're using string format, but that might not be the ideal choice.
            final String baseColumns = "%s,%.4f,%d,%d,%d,%.4f,%.4f";
            row.format(baseColumns,
                    dayStats.date,
                    dayStats.getAveragePercentage(),
                    dayStats.totalBlocks(),
                    dayStats.getAverageAddressBookNodeCount(),
                    dayStats.totalValidSignatures(),
                    dayStats.getMinPercentage(),
                    dayStats.getAveragePercentagePerBlock());

            // Histogram columns
            for (int i = 1; i <= MAX_SIG_COUNT; i++) {
                int blockCount = dayStats.blocksBySignatureCount.getOrDefault(i, 0);
                row.format(",%d").append(blockCount);
            }
            // Stake weight columns
            if (dayStats.hasStakeData()) {
                final String stakeColumns = ",%s,%d,%.4f,%.4f,%.4f,%d,%d,%d,%.1f";
                row.format(stakeColumns,
                        dayStats.mode,
                        dayStats.totalStake,
                        dayStats.minValidatedStakePct,
                        dayStats.maxValidatedStakePct,
                        dayStats.getMeanValidatedStakePct(),
                        dayStats.closestThresholdMargin,
                        dayStats.minSigningNodes,
                        dayStats.maxSigningNodes,
                        dayStats.getMeanSigningNodes());
                appendNodeStakeDistribution(row, dayStats);
                row.format(",%d,%d,,,,,,,,,,,,,,,,,",
                        dayStats.getMissingSigners(),
                        dayStats.getReliableSigners()
                    row.append(",").append(dayStats.mode);
            }
            row.format("%n");
            final String result = row.out().toString();
```

jsync-swirlds: Note, the above suggestion assumes `appendNodeStakeDistribution` is also modified to accept a `Formatter` instead of a `StringBuilder`.

jsync-swirlds (dismissed):
Just some possible improvements using a Formatter instead of combining StringBuilder and String.format.

rockysingh:
> Just some possible improvements using a Formatter instead of combining StringBuilder and String.format.

Thanks. I took the commit proposed and then made a few adjustments to it after.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ValidateBlocksCommand.java
```diff
@@ -546,6 +548,12 @@ public void run() {
             }
         }
 
+        // Create signature stats collector
+        final Path statsOutputDir =
+                Arrays.stream(files).filter(Files::isDirectory).findFirst().orElse(files[0]);
```
Nana-EC: This scenario fails in the `orElse` when `files[0]` is a normal file and not a directory.
Might want to consider using `getParent()` as a fallback

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ValidateBlocksCommand.java
```diff
@@ -546,6 +548,12 @@ public void run() {
...
+        final SignatureStatsCollector statsCollector =
```
Nana-EC: You might want to but here up to the close in a try-with resources, otherwise an error  in the middle prevents `statesCollector.close()` from being called.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SignatureValidation.java
```diff
@@ -39,25 +44,64 @@ public final class SignatureValidation implements BlockValidation {
...
     /**
      * Creates a new signature validation with optional stake-weighted consensus.
+     * Detailed stats collection is enabled by default.
      *
      * @param addressBookRegistry the address book registry for public key lookups
      * @param nodeStakeRegistry the node stake registry for stake weights (may be null for equal-weight)
      */
     public SignatureValidation(
             final AddressBookRegistry addressBookRegistry, final NodeStakeRegistry nodeStakeRegistry) {
+        this(addressBookRegistry, nodeStakeRegistry, true);
```
Nana-EC: I think this is not very intuitive and could result in misuse.
Instead anyone that was a states collection should as for it explciitly with a true param

```suggestion
        this(addressBookRegistry, nodeStakeRegistry, false);
```

Nana-EC (commented):
A few minor considerations

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ValidateBlocksCommand.java
rockysingh: Good catch fixed in 75aa3bc5. `files[0]` can be a regular file (e.g. a standalone `.blk.zip`) and we'd try to write the CSV into a non-directory path. Switched to `orElseGet(() -> files[0].getParent() != null ? files[0].getParent() : Path.of("."))` so we fall back to the containing directory, or the current working directory if the file has no parent.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ValidateBlocksCommand.java
rockysingh: Fixed in 75aa3bc5. Wrapped the validation loop in `try (final SignatureStatsCollector statsCollector = new SignatureStatsCollector(statsCsvPath))`. The existing try/finally is now nested inside so `finalizeDayStats()` and `printFinalSummary()` still run in the finally (where `statsCollector` is still in scope), and the explicit `close()` call is gone — try-with-resources handles it automatically, including on exception paths.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SignatureValidation.java
rockysingh: Agreed the hidden default is surprising. Rather than flip it to `false` (which would silently break the CSV stats output in `LiveSequential` and `ValidateBlocksCommand`), I've removed the 2-arg constructor entirely in 75aa3bc5. Callers now have to be explicit:
- `SignatureValidation(abRegistry)` for simple validation without stats
- `SignatureValidation(abRegistry, stakeRegistry, collectDetailedStats)` when stake-weighted consensus is needed

No hidden defaults, and intent is clear at every call site.

Nana-EC (approved):
LG

AlfredoG87 (approved):
LGTM

Issue #2055, opened 2026-01-26
Block Node not sending BlockAcknowledgements after unexpected ending of a stream


derektriley:
### Description

Tested against `block-node-server:0.27.0-SNAPSHOT` and `v0.25.0`.

When a consensus node is streaming to a block node, and the consensus node abruptly shuts down in the middle of streaming Block N, upon startup and streaming again to the block node the following behavior is observed:
- The consensus node receives a DuplicateBlock message for Block N - 1 (expected)
- The consensus node receives a SkipBlock message for Block N (unexpected)
- and then after streaming subsequent blocks, receives no acknowledgements (unexpected)

CN logs:
```
2026-01-26 10:10:12.390 <n0> INFO  980  BlockNodeConnection - [bn-conn-worker-STR.000002] [STR.000002/localhost:40840/ACTIVE] Worker thread started (messageSizeSoftLimit=2097152, messageSizeHardLimit=6292480, requestPadding=100, itemPadding=5)
2026-01-26 10:10:12.496 <n0> DEBUG 641  BlockNodeConnection - [bn-conn-worker-STR.000002] [STR.000002/localhost:40840/ACTIVE] [block=17, request=1] Sending request to block node (type=BLOCK_ITEMS)
2026-01-26 10:10:12.500 <n0> DEBUG 509  BlockNodeConnection - [helidon-client-3] [STR.000002/localhost:40840/ACTIVE] Received SkipBlock response; skipping to block 18
```

BN logs:
```
2026-01-26 15:06:55.111+0000 FINER   [org.hiero.block.node.stream.publisher.PublisherHandler onNext] Handler 0 finished processing request
2026-01-26 15:06:55.111+0000 FINER   [org.hiero.block.node.stream.publisher.LiveStreamPublisherManager$MessagingForwarderTask call] Forwarding batch for block=17 with blockItemSize=4
2026-01-26 15:06:55.111+0000 FINER   [org.hiero.block.node.messaging.BlockMessagingFacilityImpl sendBlockItems] Sending block items: size=4 isStartOfNewBlock=false isEndOfBlock=false, block=17
2026-01-26 15:06:55.111+0000 FINER   [org.hiero.block.node.verification.VerificationServicePlugin handleBlockItemsReceived] Appending 4 block items to the current session for block number: 17
[...]

Issue #1841, opened 2025-11-08
Add handling for publishers that stall in the middle of a block


jsync-swirlds:
### Story Form

We need to detect when a publisher won the "header race" but then streams the rest of the block slowly.
When this happens, and the block node detects the situation, the block node needs to disconnect that publisher (possibly placing that publisher in a "penalty box" for some time) and ask other connected publishers to resend that block.

### Technical Notes

Diagram showing how multiple publishers might detect this situation (Alice is the publisher that "stalls").
```mermaid
gantt
    title Block Publication "Pipeline" Diagram
    %% We just need order, not days, so (mis)use milliseconds for the times
    dateFormat SSS
    axisFormat %L
    tickInterval 5millisecond
    section Alice
      Block 9 (EndOfStream[TIMEOUT]) :active, 006, 022
      Recovery :active, 023, 030
      Connect :active, 031, 032
      Block 9 (Duplicate[10]) :done, 032, 033
      Update acknowledged blocks (10) :active, 033, 034
      Block 11 (SkipBlock) :done, 034, 035
      Block 12 (Completed) :active, 036, 041
    section Bob
      Block 9 (SkipBlock) :done, 006, 007
      Block 10 (Completed) :active, bb10c, 014, 022
      Block 9 (ResendBlock) :done, brb9, 022, 023
      Block 9 (Completed) :active, 022, 030
      Block 10 (SkipBlock) :done, 030, 031
      Block 11 (SkipBlock) :done, 032, 033
      Block 12 (SkipBlock) :done, 037, 038
    section Carol
      Block 9 (SkipBlock) :done, 006, 007
      Block 10 (SkipBlock) :done, 014, 015
      Block 11 (Interrupted) :active, 019, 022
      Block 9 (ResendBlock) :done, 022, 023 
      Block 9 (SkipBlock) :done, 023, 024
      Block 10 (SkipBlock) :done, 023, 024
      Block 11 (Completed) :active, 023, 038
      Block 12 (SkipBlock) :done, 039, 040
     section BlockNode
      Detect Block 9 timeout :milestone, after bb10c, before brb9
      Block 9 Acknowledge :milestone, 031, 031
      Block 10 Acknowledge :milestone, 031, 031
      Block 11 Acknowledge :milestone, 039, 039
      Block 12 Acknowledge :milestone, 042, 042
```

jsync-swirlds:
[...]

PR #2519, opened 2026-04-07
fix: off-by-one in clearObsoleteQueueItems

### LINKED ISSUE #2517: Bug: 0.31.0-rc1 failed upgrade and reset on previewnet
### Description

Previewnet was working fine with version `0.30.0` of BN, once we attempted to do the upgrade flow of blocks stopped, we then attempted a reset, but the flow of blocks was also lost.

we need to fix metrics, since 0.31.0 upgraded to new metrics framework from foundation. so we also need to fix flow of metrics to previewnet.

### Steps to Reproduce

No clear way to reproduce just yet, being investigated so we can create reproduceable TDD tests on BN repo.

### Additional context

_No response_

AlfredoG87:
Fixes #2517 

### Incident

On previewnet `0.31.0-rc1`, BN `lfh01` experienced a total pipeline stall when a CN stopped sending mid-block (sent header + items, never sent end-of-block). This can happen at any time — a CN crash, network partition, or process restart is sufficient to trigger it. Once triggered, the pipeline freezes indefinitely with no recovery:

- **776 blocks silently dropped** (1,175,214 -> 1,175,990) over 29 minutes
- **766 phantom "Completed blocks" increments** — counter climbed while zero blocks were verified, persisted, or ACKed
- **Stalled handler was never evicted** — no timeout, no EndStream, no removal

### Known limitation: stalled handler holds the pipeline

`LiveStreamPublisherManager` has no timeout for incomplete blocks (Bug 1 / #1841 and #2055). When a handler wins ACCEPT for a block and goes silent mid-stream, the forwarder's `queueByBlockMap` holds that block's incomplete queue indefinitely. `determineCurrentBlockNumber()` picks it via `firstEntry()` and the forwarder loops on an empty queue forever. All subsequent blocks are accepted by the decision layer but never forwarded.

### Expected mitigation: LIVE_TAIL greedy backfill

The LIVE_TAIL backfill mechanism was designed to mitigate this. When a peer BN has blocks that the local BN is missing, the backfill plugin fetches them and fires a `PersistedNotification`. The publisher manager's `handlePersisted()` then calls `clearObsoleteQueueItems()` to remove stale queues, unsticking the forwarder. This worked correctly in **v0.30.0**.

### Regression: PR #2419 / #2454 (2026-03-27)

PR #2454 (`chore: Revisit and Update BlockActions Returned by Manager`, commit `b11c8147`) introduced `clearObsoleteQueueItems()` with an off-by-one error:

```java
queueByBlockMap.headMap(blockNumber).keySet()
```

`ConcurrentNavigableMap.headMap(key)` returns entries **strictly less than** `key`. When backfill persists block N and `clearObsoleteQueueItems(N)` is called, block N's stalled queue is **excluded** from removal. The forwarder's `determineCurrentBlockNumber()` picks it via `firstEntry()` and remains stuck forever.

Additionally, if peer BNs are unreachable (restart, network partition, or simply no peers configured), the `autonomousFetcher` can't detect the gap and backfill never fires — leaving the stall completely unrecoverable. Rolling upgrades are recommended to ensure at least one peer is always available.

### Two compounding failures

1. **Backfill unavailable**: If peers are unreachable for any reason -> no LIVE_TAIL gap detected -> backfill never fires
2. **Backfill ineffective even when it fires**: `headMap(N)` off-by-one -> stalled block N's queue survives cleanup -> forwarder stays stuck

### Fix

Change `headMap(blockNumber)` to `headMap(blockNumber, true)` (inclusive), so that when `clearObsoleteQueueItems(N)` is called, block N's stalled queue is included in the removal set.

### Tests

Two new regression tests that reproduce the exact forwarder deadlock:

1. **`PublisherManagerRegressionTest.testForwarderAdvancesPastStalledBlockAfterBackfill`** — Manager-level: pre-loads blocks 0-4, stalls on block 5, backfill persists block 5, asserts block 6 reaches the messaging facility via the forwarder
2. **`StreamPublisherPluginRegressionTest.testForwarderAdvancesPastStalledBlockAfterBackfill`** — Plugin-level end-to-end: streams blocks 0-4 live with ACK verification, stalls on block 5, simulates LIVE_TAIL greedy backfill persisting block 5, streams blocks 6-7, asserts ACK for both

Both tests fail on `0.31.0-rc1` code and pass with the `headMap` fix.

### Remaining work

- **#1841**: Stalled handler timeout detection — the definitive fix for zombie handlers that eliminates the dependency on backfill as a recovery mechanism

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -388,13 +388,13 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     /// This will discard buffered data for blocks that will never be forwarded
     /// and prevent the forwarder becoming "stuck" on an incomplete block.
     ///
-    /// @param blockNumber the latest persisted block number. Only queue
-    ///     entries whose key is strictly less than this value are candidates
+    /// @param blockNumber the latest persisted block number. Queue entries
+    ///     whose key is less than or equal to this value are candidates
     ///     for removal.
     private void clearObsoleteQueueItems(final long blockNumber) {
         if (queueByBlockMap != null && !queueByBlockMap.isEmpty()) {
             final NavigableSet<Long> keysBeforeBlock = new TreeSet<>();
-            keysBeforeBlock.addAll(queueByBlockMap.headMap(blockNumber).keySet());
+            keysBeforeBlock.addAll(queueByBlockMap.headMap(blockNumber, true).keySet());
             for (final Long candidate : keysBeforeBlock) {
                 if (!(blockProofs.containsKey(candidate) || hasBlockProof(queueByBlockMap.get(candidate)))) {
                     queueByBlockMap.remove(candidate);
--- a/block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherManagerRegressionTest.java
+++ b/block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherManagerRegressionTest.java
@@ -17,6 +17,7 @@ import org.hiero.block.node.app.fixtures.TestUtils;
 import org.hiero.block.node.app.fixtures.async.BlockingExecutor;
 import org.hiero.block.node.app.fixtures.async.ScheduledBlockingExecutor;
 import org.hiero.block.node.app.fixtures.async.TestThreadPoolManager;
+import org.hiero.block.node.app.fixtures.blocks.TestBlock;
 import org.hiero.block.node.app.fixtures.blocks.TestBlockBuilder;
 import org.hiero.block.node.app.fixtures.pipeline.TestResponsePipeline;
 import org.hiero.block.node.app.fixtures.plugintest.SimpleBlockRangeSet;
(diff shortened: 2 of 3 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -388,13 +388,13 @@ public void handlePersisted(@NonNull final PersistedNotification notification) {
...
-    /// @param blockNumber the latest persisted block number. Only queue
-    ///     entries whose key is strictly less than this value are candidates
+    /// @param blockNumber the latest persisted block number. Queue entries
+    ///     whose key is less than or equal to this value are candidates
     ///     for removal.
     private void clearObsoleteQueueItems(final long blockNumber) {
         if (queueByBlockMap != null && !queueByBlockMap.isEmpty()) {
             final NavigableSet<Long> keysBeforeBlock = new TreeSet<>();
-            keysBeforeBlock.addAll(queueByBlockMap.headMap(blockNumber).keySet());
+            keysBeforeBlock.addAll(queueByBlockMap.headMap(blockNumber, true).keySet());
```
jsync-swirlds: This needs to be reevaluated as clearing queue items has a lot of side effects.
@ata-nas should consider the side effects here and incorporate any additional changes needed into his [current PR](https://github.com/hiero-ledger/hiero-block-node/pull/2505).

jsync-swirlds (approved):
We need to evaluate this change a bit more thoroughly.
The hotfix is OK, but Atanas needs to make sure we have all the possible side effects covered as well.

ata-nas (approved):

PR #2526, opened 2026-04-08
fix: add in gh workflow steps to copy plugins to the solo-dev docker image.

### LINKED ISSUE #2525: `hiero-block-node-solo-dev` CI Images Have Empty `plugins/` Directory
### Description

The `solo-dev` Docker image published to GHCR (`ghcr.io/hiero-ledger/hiero-block-node-solo-dev`) contains an empty `plugins/` directory. The image is intended to bundle all plugins so that Helm deployments can skip the Maven plugin-download init container by setting `plugins.names: null`. Because the plugins are absent, the block node crashes on startup:

```
No BlockMessagingFacility provided
```

The root cause is a two-part gap in the `publish-app` job of `.github/workflows/release-push-image.yaml`:

1. The Gradle task `prepareDockerPlugins` (defined in `block-node/app/build.gradle.kts`) is never called. This is the task responsible for collecting all plugin JARs and writing them to `build/docker/plugins/`.
2. The Docker build context points to `./block-node/app/docker` (the source directory, which has no `plugins/` subdirectory committed to git) instead of `./block-node/app/build/docker` (where `prepareDockerPlugins` writes its output).

As a result, `COPY ./plugins app-${VERSION}/plugins` in the `solo-dev` Dockerfile stage copies nothing, and the published image has no plugins.

**Why it works locally:** Running `./gradlew startDockerContainer` invokes `prepareDockerPlugins` and then `docker-build.sh` from `build/docker/`, so plugins are baked into the locally-built image. `docker-compose.yml` uses that local image and works correctly. The GHCR image skips both steps entirely.

### Steps to Reproduce

1. Pull the published `solo-dev` image and inspect its plugins directory:
   ```bash
   docker run --rm --entrypoint ls \
     ghcr.io/hiero-ledger/hiero-block-node-solo-dev:<version> \
     /opt/hiero/block-node/app-<version>/plugins
   ```
   Expected: plugin JARs. Actual: only `.keep`.

2. Deploy Solo with `ONE_SHOT_WITH_BLOCK_NODE=true` using a Helm values overlay that sets `plugins.names: null` (e.g., the `solo-dev.yaml` overlay from `charts/block-node-server/values-overrides/`).

3. Observe the block node pod enter `CrashLoopBackOff` with:
   ```
   No BlockMessagingFacility provided
   ```

### Additional context

**Proposed fix in `release-push-image.yaml`:**

Add a step to run `prepareDockerPlugins` after the build, and update the Docker build context:

```yaml
- name: Build
  run: ${GRADLE_EXEC} clean assemble

# Add this step:
- name: Prepare plugin JARs for solo-dev image
  run: ${GRADLE_EXEC} :block-node:app:prepareDockerPlugins

- name: Server - Build and push image
  uses: docker/build-push-action@...
  with:
    context: ./block-node/app/build/docker    # changed from ./block-node/app/docker
    build-contexts: |
      distributions=./block-node/app/build/distributions
```

berryware:
The local gradle build has an additional step for staging the plugins that is missing from the GH workflow. This missing step is causing the solo-dev image to not have any plugins. This pr adds the missing step to the github workflow and correctly sets the docker context so that docker build can find the plugins.
 
- add in steps to generate the plugins directory so that they are available

## Reviewer Notes
Steps to Verify plugins are in the image:

Pull the published solo-dev image and inspect its plugins directory:

docker run --rm --entrypoint ls \
  ghcr.io/hiero-ledger/hiero-block-node-solo-dev:<version> \
  /opt/hiero/block-node/app-<version>/plugins
Expected: plugin JARs.

## Related Issue(s)
Closes #2525

### DIFF
--- a/.github/workflows/release-push-image.yaml
+++ b/.github/workflows/release-push-image.yaml
@@ -124,7 +124,7 @@ jobs:
         run: ${GRADLE_EXEC} clean assemble
 
       - name: Prepare plugin JARs for solo-dev image
-        run: ${GRADLE_EXEC} :block-node:app:prepareDockerPlugins
+        run: ${GRADLE_EXEC} :app:prepareDockerPlugins
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
--- a/.github/workflows/release-push-image.yaml
+++ b/.github/workflows/release-push-image.yaml
@@ -123,6 +123,9 @@ jobs:
       - name: Build
         run: ${GRADLE_EXEC} clean assemble
 
+      - name: Prepare plugin JARs for solo-dev image
+        run: ${GRADLE_EXEC} :block-node:app:prepareDockerPlugins
+
       - name: Login to GitHub Container Registry
         uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
         with:
@@ -157,7 +160,7 @@ jobs:
       - name: Server - Build and push image
         uses: docker/build-push-action@d08e5c354a6adb9ed34480a06d141179aa583294 # v7.0.0
         with:
-          context: ./block-node/app/docker
+          context: ./block-node/app/build/docker
           file: ./block-node/app/docker/Dockerfile
           target: ${{ matrix.target }}
           cache-from: type=gha


AlfredoG87 (dismissed):
LGTM

AlfredoG87 (approved):
LG, tested! 👍

Nana-EC (approved):

PavelSBorisov (approved):

PR #2534, opened 2026-04-09
feat: (issue 2429) read gRPC hiero-correlation-id header and prefix publisher handler logs

a-saksena:
## Related Issue(s)

### DIFF
--- a/block-node/stream-publisher/src/main/java/module-info.java
+++ b/block-node/stream-publisher/src/main/java/module-info.java
@@ -19,6 +19,7 @@ module org.hiero.block.node.stream.publisher {
     requires org.hiero.block.node.app.config;
     requires org.hiero.block.node.base;
     requires com.github.spotbugs.annotations;
+    requires java.logging;
 
     provides org.hiero.block.node.spi.BlockNodePlugin with
             StreamPublisherPlugin;
--- a/block-node/stream-publisher/src/main/java/module-info.java
+++ b/block-node/stream-publisher/src/main/java/module-info.java
@@ -19,7 +19,6 @@ module org.hiero.block.node.stream.publisher {
     requires org.hiero.block.node.app.config;
     requires org.hiero.block.node.base;
     requires com.github.spotbugs.annotations;
-    requires java.logging;
 
     provides org.hiero.block.node.spi.BlockNodePlugin with
             StreamPublisherPlugin;
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
@@ -220,7 +220,8 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
 
             LOGGER.log(
                     TRACE,
-                    "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId=[{3}]",
+                    // Note: not surrounding correlationId={3} with [] intentionally as it will break Loki's parsing.
+                    "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId={3}",
                     newLastAcknowledgedBlockNumber,
                     System.nanoTime(),
                     handlerId,
@@ -361,7 +362,8 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
             currentStreamingBlockHeaderReceivedTime = System.nanoTime();
             LOGGER.log(
                     TRACE,
-                    "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId=[{3}]",
+                    // Note: not surrounding correlationId={3} with [] intentionally as it will break Loki's parsing.
+                    "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId={3}",
                     blockNumber,
                     currentStreamingBlockHeaderReceivedTime,
                     handlerId,
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
@@ -82,6 +82,10 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     private final StreamPublisherManager publisherManager;
     /// The ID of this handler. This is used to identify the handler within the publisher manager.
     private final long handlerId;
+    /// Pre-computed log prefix derived from the gRPC `hiero-correlation-id` header.
+    /// Computed once at construction time and reused on every log call to avoid per-call string allocation.
+    /// Value is `"[correlationId] "` when a correlation ID is present, or `""` when absent.
+    private final String correlationIdPrefix;
     /// The current streaming block number. This is used to track the current block being streamed.
     private final AtomicLong currentStreamingBlockNumber;
     /// The current queue of the block currently streaming
@@ -103,19 +107,31 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     /// @param replyPipeline the pipeline to send replies to
     /// @param handlerMetrics the metrics for this handler
     /// @param manager the publisher manager that manages this handler
+    /// @param correlationId the correlation ID from the gRPC `hiero-correlation-id` header, or empty string if absent
     public PublisherHandler(
             final long nextId,
             @NonNull final Pipeline<? super PublishStreamResponse> replyPipeline,
             @NonNull final MetricsHolder handlerMetrics,
-            @NonNull final StreamPublisherManager manager) {
+            @NonNull final StreamPublisherManager manager,
+            @NonNull final String correlationId) {
         handlerId = nextId;
         replies = Objects.requireNonNull(replyPipeline);
         metrics = Objects.requireNonNull(handlerMetrics);
         publisherManager = Objects.requireNonNull(manager);
+        final String validatedCorrelationId = Objects.requireNonNull(correlationId);
+        this.correlationIdPrefix = validatedCorrelationId.isEmpty() ? "" : "[" + validatedCorrelationId + "] ";
         currentStreamingBlockNumber = new AtomicLong(UNKNOWN_BLOCK_NUMBER);
         currentBlockQueue = new AtomicReference<>();
         blockAction = new AtomicReference<>();
         unacknowledgedStreamedBlocks = new ConcurrentSkipListSet<>();
+    }
+
+    /// Returns the pre-computed log prefix for this handler.
+    ///
+    /// Returns `"[correlationId] "` if a correlation ID was provided at connection time,
+    /// or `""` if none was present. Pre-computed in the constructor to avoid allocations on hot paths.
+    private String ctx() {
+        return correlationIdPrefix;
     }
 
     // ==== Flow Methods =======================================================
@@ -156,13 +172,13 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     @Override
     public void onNext(@NonNull final PublishStreamRequestUnparsed request) {
         try {
-            LOGGER.log(TRACE, "Handler {0} received request", handlerId);
+            LOGGER.log(TRACE, ctx() + "Handler {0} received request", handlerId);
             processNextRequestUnparsed(request);
-            LOGGER.log(TRACE, "Handler {0} finished processing request", handlerId);
+            LOGGER.log(TRACE, ctx() + "Handler {0} finished processing request", handlerId);
         } catch (final InterruptedException | RuntimeException e) {
             // If we reach here, it means that the handler was interrupted or
             // an unexpected error occurred. We should log the error and shut down.
-            LOGGER.log(INFO, "Error processing request: %s".formatted(e), e);
+            LOGGER.log(INFO, ctx() + "Error processing request: %s".formatted(e), e);
             sendEndAndResetState(Code.ERROR);
         }
         // @todo check the current backlog by calling a manager method to
@@ -188,7 +204,7 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     ///     verified and persisted.
     public void sendAcknowledgement(final long newLastAcknowledgedBlockNumber) {
         LOGGER.log(
-                TRACE, "Handler {0} sending acknowledgement for block {1}", handlerId, newLastAcknowledgedBlockNumber);
+                TRACE, ctx() + "Handler {0} sending acknowledgement for block {1}", handlerId, newLastAcknowledgedBlockNumber);
         // We only ever need to acknowledge once for a given block number, even
         // if there are several blocks "behind" that acknowledgement.
         // The publishers expect that acknowledgement for block N implicitly
@@ -207,9 +223,9 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
                     .clear();
             metrics.blockAcknowledgementsSent.increment(); // @todo(1415) add label
 
-            final String ackMessage = "Sent acknowledgement for block {0,number,#} from handler {1}";
-            final String traceMessage =
-                    "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
+            final String ackMessage = ctx() + "Sent acknowledgement for block {0,number,#} from handler {1}";
+            final String traceMessage = ctx()
+                    + "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
             LOGGER.log(TRACE, traceMessage, newLastAcknowledgedBlockNumber, System.nanoTime(), handlerId);
             LOGGER.log(TRACE, ackMessage, newLastAcknowledgedBlockNumber, handlerId);
         }
@@ -223,7 +239,7 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     /// @param blockNumber of the block that failed verification
     /// @return true if the handler has sent the [Code#BAD_BLOCK_PROOF] message
     boolean handleFailedVerification(final long blockNumber) {
-        LOGGER.log(DEBUG, "Handler {0} handling failed verification for block {1}", handlerId, blockNumber);
+        LOGGER.log(DEBUG, ctx() + "Handler {0} handling failed verification for block {1}", handlerId, blockNumber);
         if (unacknowledgedStreamedBlocks.remove(blockNumber)) {
             // If the block number that failed verification was sent by this
             // handler, we need to send an EndOfStream with BAD_BLOCK_PROOF code.
@@ -242,7 +258,7 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     /// We will attempt to send an [EndOfStream] with a [Code#PERSISTENCE_FAILED] and
     /// proceed to shut down the handler.
     void handleFailedPersistence() {
-        LOGGER.log(DEBUG, "Handler {0} handling failed persistence", handlerId);
+        LOGGER.log(DEBUG, ctx() + "Handler {0} handling failed persistence", handlerId);
         try {
             sendEndOfStream(Code.PERSISTENCE_FAILED);
         } finally {
@@ -298,7 +314,7 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
                 try {
                     header = BlockHeader.PROTOBUF.parse(headerBytes);
                 } catch (final ParseException e) {
-                    LOGGER.log(DEBUG, "Failed to parse BlockHeader due to {0}", e);
+                    LOGGER.log(DEBUG, ctx() + "Failed to parse BlockHeader due to {0}", e);
                     // if we have reached this block, this means that the
                     // request is invalid
                     if (isCurrentlyMidBlock(blockNumber)) {
@@ -308,7 +324,7 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
                     return;
                 }
             } else {
-                LOGGER.log(DEBUG, "Handler {0} received a BlockHeader with null bytes", handlerId);
+                LOGGER.log(DEBUG, ctx() + "Handler {0} received a BlockHeader with null bytes", handlerId);
                 // this should never happen
                 if (isCurrentlyMidBlock(blockNumber)) {
                     publisherManager.blockIsEnding(blockNumber, handlerId);
@@ -330,8 +346,8 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
             currentStreamingBlockNumber.set(blockNumber);
             // this means that we are starting a new block, so we can
             // update the current streaming block number
-            final String traceMessage =
-                    "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
+            final String traceMessage = ctx()
+                    + "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
             currentStreamingBlockHeaderReceivedTime = System.nanoTime();
             LOGGER.log(TRACE, traceMessage, blockNumber, currentStreamingBlockHeaderReceivedTime, handlerId);
         } else if (UNKNOWN_BLOCK_NUMBER == blockNumber) {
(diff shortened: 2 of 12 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -207,9 +227,9 @@ public void sendAcknowledgement(final long newLastAcknowledgedBlockNumber) {
...
             metrics.blockAcknowledgementsSent.increment(); // @todo(1415) add label
 
-            final String ackMessage = "Sent acknowledgement for block {0,number,#} from handler {1}";
-            final String traceMessage =
-                    "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
+            final String ackMessage = ctx() + "Sent acknowledgement for block {0,number,#} from handler {1}";
+            final String traceMessage = ctx()
+                    + "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
             LOGGER.log(TRACE, traceMessage, newLastAcknowledgedBlockNumber, System.nanoTime(), handlerId);
             LOGGER.log(TRACE, ackMessage, newLastAcknowledgedBlockNumber, handlerId);
```
jsync-swirlds: Another example; don't add the ID to the string, let the logger add it _when needed_.
Also, the way you added it to the trace message breaks log metrics, you have to add at the end with a key=value pattern.

```suggestion
            final String ackMessage = "{2} Sent acknowledgement for block {0,number,#} from handler {1}";
            final String traceMessage = "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId={3}";
LOGGER.log(TRACE, traceMessage, newLastAcknowledgedBlockNumber, System.nanoTime(), handlerId, correlationIdPrefix);
LOGGER.log(TRACE, ackMessage, newLastAcknowledgedBlockNumber, handlerId, correlationIdPrefix);
```

jsync-swirlds (changes requested):
The way this was added isn't good and breaks some important metrics.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherHandlerTest.java
```diff
@@ -125,11 +125,129 @@ void testNullMetricsHolder() {
...
+
+        /**
+         * This test aims to assert that the constructor of {@link PublisherHandler} throws a
+         * {@link NullPointerException} when the correlation ID is null.
+         */
+        @Test
+        @DisplayName("Test constructor with null correlation ID")
+        void testNullCorrelationId() {
+            assertThatNullPointerException().isThrownBy(() -> {
+                new PublisherHandler(validNextId, validReplyPipeline, validMetricsHodler, validPublisherManager, null);
```
jsync-swirlds: We generally want the _opposite_ of this.
Exceptions are a lot harder to handle than just not including a trace value, so it's better to just silently leave the ID value empty in the handler rather than throw a NPE for data that a publisher might not send.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -177,10 +192,26 @@ public List<Class<? extends Record>> configDataTypes() {
...
+    ///
+    /// @param correlationId the raw value from the gRPC header
+    /// @return the value unchanged if within the limit, or truncated to [#MAX_CORRELATION_ID_LENGTH]
+    @NonNull
+    static String truncateCorrelationId(@NonNull final String correlationId) {
+        if (correlationId.length() > MAX_CORRELATION_ID_LENGTH) {
+            return correlationId.substring(0, MAX_CORRELATION_ID_LENGTH);
+        }
+        return correlationId;
     }
```
jsync-swirlds: Rather than making this "NonNull", we should probably just allow and handle null correlation ID values.

```suggestion
    /// Truncates the given correlation ID to [#MAX_CORRELATION_ID_LENGTH] characters.
    ///
    /// Protects against oversized header values that could inflate every log line.
    /// If the value exceeds the limit, it is truncated and a warning is logged.
    ///
    /// @param correlationId the raw value from the gRPC header
    /// @return the value unchanged if within the limit, or truncated to [#MAX_CORRELATION_ID_LENGTH]
    static String truncateCorrelationId(final String correlationId) {
        if (correlationId != null && correlationId.length() > MAX_CORRELATION_ID_LENGTH) {
            return correlationId.substring(0, MAX_CORRELATION_ID_LENGTH);
        }
        return correlationId;
}
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -156,13 +173,13 @@ public void onSubscribe(@NonNull final Subscription subscription) {
     @Override
     public void onNext(@NonNull final PublishStreamRequestUnparsed request) {
         try {
-            LOGGER.log(TRACE, "Handler {0} received request", handlerId);
+            LOGGER.log(TRACE, ctx() + "Handler {0} received request", handlerId);
             processNextRequestUnparsed(request);
-            LOGGER.log(TRACE, "Handler {0} finished processing request", handlerId);
+            LOGGER.log(TRACE, ctx() + "Handler {0} finished processing request", handlerId);
```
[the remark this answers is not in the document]
a-saksena: Agree to use string interpolation instead of concatenation.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -103,21 +107,34 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
...
             @NonNull final MetricsHolder handlerMetrics,
-            @NonNull final StreamPublisherManager manager) {
+            @NonNull final StreamPublisherManager manager,
+            @NonNull final String correlationId) {
         handlerId = nextId;
         replies = Objects.requireNonNull(replyPipeline);
         metrics = Objects.requireNonNull(handlerMetrics);
         publisherManager = Objects.requireNonNull(manager);
+        final String validatedCorrelationId = Objects.requireNonNull(correlationId);
+        this.correlationIdPrefix = validatedCorrelationId.isEmpty() ? "" : "[" + validatedCorrelationId + "] ";
```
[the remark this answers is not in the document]
a-saksena: Agree that the id can come in as Null.
However, it makes the log much clearer with the [ ].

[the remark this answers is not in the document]
jsync-swirlds: There are some forms that cannot include the `[]`, so I've been adding that _when needed_ in the log message in these examples.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherHandlerTest.java
a-saksena: I will write another explicit test to test that Null is a valid value and NPE is not thrown because of it being null.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -183,12 +183,12 @@ void setup() {
             // Create a response pipeline to handle the responses from the first publisher handler.
             responsePipeline = new TestResponsePipeline();
             // Create the first publisher handler and add it to the manager.
-            publisherHandler = toTest.addHandler(responsePipeline, sharedHandlerMetrics);
+            publisherHandler = toTest.addHandler(responsePipeline, sharedHandlerMetrics, "");
```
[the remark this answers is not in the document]
Nana-EC: One approach is to use a `@ParameterizedTest` to cover `null`, `""` and non-empty for the same test

Nana-EC (changes requested):
Good start including passing through call chain and pre-computing correlationIdPrefix, character truncations and test additions.

@jsync-swirlds covered most of the cases I'm concerned about with respect to perf and breaking log layouts for other telemetry.
Also kindly add a PR description that makes it easier for reviewers to understand what they are reviewing and the impact

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -103,21 +107,34 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
...
     }
 
+    /// Returns the pre-computed log prefix for this handler.
+    ///
+    /// Returns `"[correlationId] "` if a correlation ID was provided at connection time,
+    /// or `""` if none was present. Pre-computed in the constructor to avoid allocations on hot paths.
+    private String ctx() {
+        return correlationIdPrefix;
+    }
+
```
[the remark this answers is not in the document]
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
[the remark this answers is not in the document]
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -330,8 +350,8 @@ private void handleBlockItemsRequest(
             currentStreamingBlockNumber.set(blockNumber);
             // this means that we are starting a new block, so we can
             // update the current streaming block number
-            final String traceMessage =
-                    "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
+            final String traceMessage = ctx()
+                    + "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
             currentStreamingBlockHeaderReceivedTime = System.nanoTime();
             LOGGER.log(TRACE, traceMessage, blockNumber, currentStreamingBlockHeaderReceivedTime, handlerId);
```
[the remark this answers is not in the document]
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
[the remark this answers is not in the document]
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -394,8 +415,9 @@ private void handleEndStreamRequest(final EndStream endStream) {
         } else {
             LOGGER.log(
                     INFO,
-                    "Handler %d received an invalid EndStream request with code %s. %s"
-                            .formatted(handlerId, code, earliestAndLatestBlockNumbers));
+                    ctx()
+                            + "Handler %d received an invalid EndStream request with code %s. %s"
+                                    .formatted(handlerId, code, earliestAndLatestBlockNumbers));
```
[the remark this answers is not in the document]
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -544,16 +567,17 @@ private boolean sendResponse(final PublishStreamResponse response) {
             // at debug rather than emitting noise in the logs.
             // Also, this confuses everyone, they all see this debug log and
             // assume the node crashed, so we must not print a stack trace.
-            final String messageFormat = "Publisher closed the connection unexpectedly for client %d: %s";
+            final String messageFormat = ctx() + "Publisher closed the connection unexpectedly for client %d: %s";
             final String exceptionMessage = e.getCause() != null ? e.getCause().getMessage() : e.getMessage();
             final String message = messageFormat.formatted(handlerId, exceptionMessage);
```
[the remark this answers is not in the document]
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -753,20 +784,21 @@ private void shutdown() {
...
+            LOGGER.log(TRACE, ctx() + "Handler {0} issued onComplete/closeConnection", handlerId);
         }
     }
 
     /// Check if we have a block in progress. The parameter should be used to supply the
     /// [#currentStreamingBlockNumber] value.
     private boolean isCurrentlyMidBlock(final long blockInProgress) {
         if (blockInProgress != currentStreamingBlockNumber.get()) {
-            LOGGER.log(DEBUG, "Ending mid block, but block number does not match.");
+            LOGGER.log(DEBUG, ctx() + "Ending mid block, but block number does not match.");
```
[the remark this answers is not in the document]
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/module-info.java
```diff
@@ -19,6 +19,7 @@
     requires org.hiero.block.node.app.config;
     requires org.hiero.block.node.base;
     requires com.github.spotbugs.annotations;
+    requires java.logging;
```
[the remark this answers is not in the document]
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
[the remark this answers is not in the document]
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherHandlerTest.java
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
a-saksena: Review comment incorporated in the latest code.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -207,11 +218,19 @@ public void sendAcknowledgement(final long newLastAcknowledgedBlockNumber) {
...
             metrics.blockAcknowledgementsSent.increment(); // @todo(1415) add label
 
-            final String ackMessage = "Sent acknowledgement for block {0,number,#} from handler {1}";
-            final String traceMessage =
-                    "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
-            LOGGER.log(TRACE, traceMessage, newLastAcknowledgedBlockNumber, System.nanoTime(), handlerId);
-            LOGGER.log(TRACE, ackMessage, newLastAcknowledgedBlockNumber, handlerId);
+            LOGGER.log(
+                    TRACE,
+                    "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId=[{3}]",
```
jsync-swirlds: Here we must not include the `[]` characters; it breaks parsing in Loki.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -330,18 +358,26 @@ private void handleBlockItemsRequest(
             currentStreamingBlockNumber.set(blockNumber);
             // this means that we are starting a new block, so we can
             // update the current streaming block number
-            final String traceMessage =
-                    "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2}";
             currentStreamingBlockHeaderReceivedTime = System.nanoTime();
-            LOGGER.log(TRACE, traceMessage, blockNumber, currentStreamingBlockHeaderReceivedTime, handlerId);
+            LOGGER.log(
+                    TRACE,
+                    "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId=[{3}]",
```
jsync-swirlds: Same here, this is a metric value, so we must not add extra characters that cause parsing issues.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -1392,6 +1392,60 @@ void testHandlePersistedNotificationFailedPersistence() {
                 assertThat(responsePipeline.getClientEndStreamCalls().get()).isEqualTo(0);
             }
 
+            /// Verifies that a failed [PersistedNotification] (succeeded = false)
```
jsync-swirlds, line 1395: These coverage tests will probably conflict with @ata-nas latest changes; so we should be careful about what order we merge the two PRs.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-tck-sdk-run.sh
```diff
@@ -8,7 +8,7 @@ set -euo pipefail
 
 SDK_DIR=""
 TCK_DIR=""
-TEST_FILE="src/tests/crypto-service/test-transfer-hbar-transaction.ts src/tests/crypto-service/test-account-create-transaction.ts src/tests/crypto-service/test-account-update-transaction.ts src/tests/token-service/test-token-create-transaction.ts src/tests/token-service/test-token-update-transaction.ts src/tests/topic-service/test-topic-create-transaction.ts src/tests/topic-service/test-topic-update-transaction.ts src/tests/contract-service/test-contract-delete-transaction.ts"
```
jsync-swirlds, line 11: Why did we remove these?

tools-and-tests/scripts/solo-e2e-test/scripts/solo-tck-sdk-run.sh
```diff
@@ -90,17 +90,12 @@ fi
...
 cp .env.custom_node .env
-if [[ "${TEST_FILE}" == "all" ]]; then
-  echo "Running all TCK tests (test:serial)"
-  npm run test:serial
-else
-  echo "Running TCK tests: $TEST_FILE"
-  # shellcheck disable=SC2086
-  # Word splitting is intentional — TEST_FILE may contain multiple space-separated file paths
-  npm run test:file $TEST_FILE
-fi
```
jsync-swirlds, line 103: Similar concern, this seems to remove the option to run "all" tests.

cc: @Nana-EC

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherHandlerTest.java
```diff
@@ -1946,6 +2067,111 @@ void testOnNextEndStreamTooFarBehindNoNotification() {
                         manager.getBlockMessagingFacility().getSentNewestBlockKnownToNetworkNotifications();
                 assertThat(sentNotifications).hasSize(0);
             }
+
```
jsync-swirlds, line 2070: These should also be examined by @ata-nas to ensure they aren't conflicting with his recent changes.

jsync-swirlds (dismissed):
A couple changes needed for the metric logs.
These are log statements we use Loki to parse into metrics because they're cross-plugin telemetry data we cannot easily turn into useable metrics otherwise.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-tck-sdk-run.sh
Nana-EC: Yeah, please revert all changes in this file for this PR.
If we want to streamline anything here let's do it separately so this PR doesn't get held up

tools-and-tests/scripts/solo-e2e-test/scripts/solo-tck-sdk-run.sh
Nana-EC: Please revert

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
Nana-EC: It's a gap in our documentation but basically anywhere you see the format <key>=<value> is an instance of logs that Loki can convert into telemetry and we can use to create adhoc telemetry.
As such we don't want to break those but you can add key-value pairs to them

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -122,6 +127,16 @@ public Pipeline<? super Bytes> open(
         final int maxMessageSize =
                 context.configuration().getConfigData(ServerConfig.class).maxMessageSizeBytes() - 16384;
 
+        final String rawCorrelationId = options.metadata().getOrDefault("hiero-correlation-id", "");
+        if (rawCorrelationId.length() > MAX_CORRELATION_ID_LENGTH) {
```
Nana-EC: I think this logic is repeated in truncateCorrelationId so you can get rid of it and just call the method

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -122,6 +127,16 @@ public Pipeline<? super Bytes> open(
...
+            LOGGER.log(
+                    System.Logger.Level.WARNING,
+                    "Received hiero-correlation-id header exceeds {0} characters and will be truncated: {1}",
```
Nana-EC: May also be worth logging the size that came in so it's clear how much over the limit we're observing

Nana-EC (changes requested):
Some mote items.
We should scope this to the necessary changes needed but also watch out for the value pairs in logs, those are logs that Loki can turn into telemetry that we specifcally formatted for

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
a-saksena: Incorporated. This should be fixed now.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
a-saksena: Incorporated. This should be fixed now.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
a-saksena: Will coordinate with Atanas.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-tck-sdk-run.sh
a-saksena: solo-tck-sdk-run.sh file changes got committed by mistake last time. Reverting.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherHandlerTest.java
a-saksena: Will coordinate with Atanas.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
a-saksena: Incorporated. This should be fixed now.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
a-saksena: Incorporated. This should be fixed now.

jsync-swirlds (approved):

Nana-EC (approved):

PR #2535, opened 2026-04-09
feat(verification): add StateProof (indirect proof) verification (#2019)

### LINKED ISSUE #2019: feat(verification): Add support for other BlockProof Types other than TSS
### Story Form

As a BlockNode Operator/Dev
I want to be able to verify blocks that don't have TSS proof but other types.
So that we can still verify those blocks

### Technical Notes

Wrapped Record Blocks and first few thousand blocks after cut-over or even genesis might contain different proof types. such as: `AggregatedNodeSignature`

This issue is to make sure that the Verification Plugin of the BlockNode supports it.

Make sure to address this comment:
https://github.com/hiero-ledger/hiero-block-node/pull/1993#discussion_r2692576763

AlfredoG87:
## Summary

Adds support for verifying blocks with indirect (state) proofs (`BlockProof.block_state_proof`), which occur when TSS signing is delayed after restarts, upgrades, or node additions (less than 1% of blocks in production).

- **Merkle path walking**: Iterative algorithm in `verifyStateProof()` reconstructs the directly-signed block's root hash by walking sibling hashes through gap blocks. Supports any gap length (formula: `4 * N_gap_blocks + 3` siblings).
- **Proof type fallback**: `finalizeVerification()` now tries direct TSS first, then falls back to state proof. Replaces the throwing `getSingle()` with `findFirst()` for graceful handling.
- **End-to-end tests**: Real TSS signature verification using block 0's `LedgerIdPublicationTransactionBody` for initialization. Tamper-rejection tests confirm that modified sibling hashes or timestamp leaves cause verification failure.
- **Test fixtures**: Real blocks from a hapiTestWraps capture (Schnorr TSS, every 4th block signed) covering 1-gap, 2-gap, and 3-gap indirect proofs, with `SAMPLE_BLOCKS_SCHNORR` enum in `BlockUtils`.

Resolves #2019

## Test plan

- [x] `StateProofVerificationTest` — end-to-end verification with real TSS for 1-gap (7 siblings), 2-gap (11 siblings), and 3-gap (15 siblings) proofs
- [x] Tampered sibling hash rejected by TSS verification
- [x] Tampered timestamp leaf rejected by TSS verification
- [x] Sibling count formula validation (`4 * gaps + 3`)
- [x] Direct TSS proof path still works (blocks 0 and 4)
- [x] All 59 verification tests pass


block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +240,137 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+        List<SiblingNode> siblings = siblingPath.siblings();
+        int totalSiblings = siblings.size();
+        // Must have at least 3 (signed block siblings) and remainder must be groups of 4 (gap blocks)
+        if (totalSiblings < 3 || (totalSiblings - 3) % 4 != 0) {
+            return false;
+        }
+
+        // Start from path1.hash = previousBlockHash of the target block (= rootHash of block T-1).
+        // Walk through siblings in groups, rebuilding each block's root hash in the chain.
+        byte[] current = siblingPath.hash().toByteArray();
```
Nana-EC, line 323: Should check for null or empty `siblingPath.hash()` to avoid incorrect hashes further down the line.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +240,137 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+
+    /**
+     * Walks the merkle path chain in a state proof to reconstruct the directly-signed block's
+     * root hash, then verifies the TSS signature on that root.
+     *
+     * <p>The state proof contains 3 paths:
+     * <ul>
+     *   <li>Path 0: timestamp leaf of the signed block</li>
+     *   <li>Path 1: sibling hashes from the target block through all gap blocks to the signed block</li>
+     *   <li>Path 2: terminal (empty)</li>
```
Nana-EC, line 285: Q: Do we care for this value and should we validate/log if it's not the expected empty?

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +240,137 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+     * </ul>
+     *
+     * <p>Path 1's siblings are laid out as groups:
+     * <ul>
+     *   <li>Per gap block (4 siblings): prevBlockRootsHash, depth5Node2, depth4Node2, timestamp</li>
```
Nana-EC, line 290: Can we rely on or do we need to verify the location of these siblings.
If there's ever a change or bug in their location we likely would find it hard to detect that it was due to some order bug.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -258,14 +401,15 @@ private LedgerIdPublicationTransactionBody findLedgerIdPublication(Bytes signedT
...
-    public static <T> T getSingle(List<T> list, Predicate<T> predicate) {
-        List<T> filtered = list.stream().filter(predicate).toList();
-
-        if (filtered.size() != 1) {
-            throw new IllegalStateException(String.format(
-                    "Expected exactly 1 element matching predicate [%s], but found %d.", predicate, filtered.size()));
+    /**
+     * Returns the first element matching the predicate, or null if none matches.
+     */
+    private static <T> T findFirst(List<T> list, Predicate<T> predicate) {
```
Nana-EC, line 407: So in replacing getSingle with findFirst() we've lost awareness and checks of additional items passed in where they shouldn't be.

block-node/verification/src/test/java/org/hiero/block/node/verification/session/impl/StateProofVerificationTest.java
```diff
@@ -0,0 +1,264 @@
...
+        block1 = BlockUtils.getSampleBlockInfo(SAMPLE_BLOCKS_STATE_PROOFS.BLOCK_1);
+        block2 = BlockUtils.getSampleBlockInfo(SAMPLE_BLOCKS_STATE_PROOFS.BLOCK_2);
+        block3 = BlockUtils.getSampleBlockInfo(SAMPLE_BLOCKS_STATE_PROOFS.BLOCK_3);
+        block4 = BlockUtils.getSampleBlockInfo(SAMPLE_BLOCKS_STATE_PROOFS.BLOCK_4);
+    }
+
+    @BeforeEach
+    void initializeLedgerState() throws ParseException {
+        // Reset static TSS state for test isolation
+        VerificationServicePlugin.activeLedgerId = null;
```
Nana-EC, line 67: Note, there are non volatile (non thread safe) fields. If we have any places where we run tests in parallel it could result in flaky tests or unexpected results.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +240,137 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+        }
+        if (!stateProof.hasSignedBlockProof()) {
+            return false;
+        }
+
+        List<SiblingNode> siblings = siblingPath.siblings();
+        int totalSiblings = siblings.size();
+        // Must have at least 3 (signed block siblings) and remainder must be groups of 4 (gap blocks)
+        if (totalSiblings < 3 || (totalSiblings - 3) % 4 != 0) {
+            return false;
```
Nana-EC, line 318: I think test cases are missing coverage for when the number of sibling sis not 3 e.g. 1, 2 or 4

Nana-EC (commented):
nice additions,  some suggestions for consideration

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/BlockUtils.java
```diff
@@ -53,15 +54,14 @@ public static Block toBlock(BlockUnparsed block) {
...
+     * Gets a SampleBlockInfo from any sample block enum.
      */
-    public static SampleBlockInfo getSampleBlockInfo(SAMPLE_BLOCKS sampleBlocks) throws IOException, ParseException {
+    public static SampleBlockInfo getSampleBlockInfo(SampleBlock sampleBlock) throws IOException, ParseException {
         BlockUnparsed blockUnparsed;
-        var stream = TestUtils.class.getModule().getResourceAsStream("test-blocks/" + sampleBlocks.blockName);
-        try (final var gzipInputStream = new GZIPInputStream(stream)) {
-            // Read the bytes from the GZIPInputStream
+        InputStream stream =
+                TestUtils.class.getModule().getResourceAsStream("test-blocks/" + sampleBlock.getBlockName());
```
ata-nas, line 62: Let's include the `stream` in the try-with-resources.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/BlockUtils.java
ata-nas, line 1: I think it would be a good idea to get rid of this class, as it is sometimes used. We should move these sample blocks to the `TestBlockBuilder` so we have a single utility to create test blocks.

This can of course be done in the future, but the more we put here, the harder it will be to get rid of this.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
ata-nas: We should generally be very defensive, failing immediately with a result whenever we cannot proceed.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -167,15 +173,21 @@ public VerificationNotification finalizeVerification(Bytes rootHashOfAllBlockHas
...
                     tssBasedProof, previousBlockHashToUse, rootOfAllPreviousBlockHashes, startOfBlockStateRootHash);
         }
 
-        // was not able to finalize verification yet
+        // Try indirect (state) proof — used when TSS signing was delayed
+        BlockProof stateBasedProof = findFirst(blockProofs, BlockProof::hasBlockStateProof);
+        if (stateBasedProof != null) {
+            return getStateProofVerificationResult(
+                    stateBasedProof, previousBlockHashToUse, rootOfAllPreviousBlockHashes, startOfBlockStateRootHash);
+        }
```
ata-nas, line 188: This seems like an if-else.
IF we have `tssBasedProof` or `stateBasedProof`, is that correct?
Do we always have one or the other? Just trying to understand the construct.

AlfredoG87: Not always will be one or the other, and we might have newer ones in the future, and it might even be the case that we have multiple block proofs per block, so is possible to have both, so in that case we should pick the most robust and easier to verify, out of the 2 we currently verify that would be the Direct TSS based proof.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: We do have test cases for 1 to 3 gaps blocks too:
  - shouldVerifyBlock3With1GapStateProof — 1-gap, 7 siblings                                                                                                                                                                                                                                                      
  - shouldVerifyBlock2With2GapStateProof — 2-gap, 11 siblings                                                                                                                                                                                                                                                     
  - shouldVerifyBlock1With3GapStateProof — 3-gap, 15 siblings                                                                                                                                                                                                                                                     
  - shouldVerifySiblingCountMatchesGapSize — explicitly asserts the 4*N+3 counts for all three cases

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅ Done, please re-review and resolve if OK.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: WARNING log added when terminalPath is unexpectedly non-empty ✅

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: findFirst was renamed back to getSingle (now instance method), logs WARNING and returns null on multiple matches ✅

block-node/verification/src/test/java/org/hiero/block/node/verification/session/impl/StateProofVerificationTest.java
AlfredoG87: Added @Execution(ExecutionMode.SAME_THREAD) to StateProofVerificationTest ✅

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/BlockUtils.java
AlfredoG87: Both stream and gzipInputStream now inside the try block ✅

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: I agree with Fredy.
At some point we might choose to have a switch on the oneOf type for proofs, parallel verify all of them, and have some rules for which types are required and which are just "nice to have".

A good example of this would be when the network begins using hybrid cryptography on the road to PQC support. We would probably initially require hinTS/TSS, but verify PQC/TSS if present (and if we have the required verification data), then later switch to require PQC/TSS and only verify hinTS/TSS if present as a confirmatory analytic.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: Note, StateProof type block proofs are _permitted_ to be more extensive, and we should verify every branch that we _can_ reasonably verify.

This should not happen _now_, but could be used in future scenarios with WRBs or when we start working on a hybrid PQC version of TSS (we _might_ need state proofs to prove more branches for PQC, depending on data sizes and other factors).

If the StateProof proves more than we expect, that isn't an error, it's just extra data that we should report in diagnostics because it might represent forward-compatible enhancements.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: An INFO log is appropriate, if we don't verify that additional path. It isn't an error or completely unexpected, however, so I don't recommend warning level logs or exceptions.

Note, we _should_ probably attempt to verify _every_ path that is provided in the StateProof (that is probably a future enhancement, however).

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: I don't think we have those problems, use cases right now. 
When we implement those, then we can adapt and change.

for every use-case, we add testing with sample blocks to support those cases, we can discover the critical path right now, instead of trying to solve all future "possible" scenarios right now.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: We have been avoiding to use WARN and ERROR only in exceptional cases, but as per owns Javadocs from oracle:

https://docs.oracle.com/javase/8/docs/api/java/util/logging/Level.html

> INFO is a message level for informational messages.

> WARNING is a message level indicating a potential problem.


In this case this is a potential problem, not an informational message. so we should stop being averse to correct logging practices, otherwise we are making it harder to debug and troubleshoot, both for ourselves and for other operators.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
ata-nas: Thanks for clarifying this for me, both of you.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +241,151 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+
+        return new VerificationNotification(
+                verified, blockNumber, blockRootHash, verified ? new BlockUnparsed(blockItems) : null, blockSource);
+    }
+
+    /**
+     * Walks the merkle path chain in a state proof to reconstruct the directly-signed block's
+     * root hash, then verifies the TSS signature on that root.
+     *
+     * <p>The state proof contains 3 paths:
```
mhess-swl:
```suggestion
     * <p>A valid block state proof contains 3 paths:
```

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +241,151 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+        // Walk through siblings in groups, rebuilding each block's root hash in the chain.
+        byte[] current = siblingPath.hash().toByteArray();
+        int index = 0;
+
+        // Process gap blocks — each contributes 4 siblings
+        while (totalSiblings - index > 3) {
+            SiblingNode prevBlockRootsHash = siblings.get(index);
+            SiblingNode depth5Node2Sibling = siblings.get(index + 1);
+            SiblingNode depth4Node2Sibling = siblings.get(index + 2);
+            SiblingNode timestampSibling = siblings.get(index + 3);
```
mhess-swl:
```suggestion
            SiblingNode hashedTimestampSibling = siblings.get(index + 3);
```

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +241,151 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+        }
+
+        // Process the signed block's 3 siblings
+        byte[] depth5Node1 = combineSibling(current, siblings.get(index));
+        byte[] depth4Node1 = combineSibling(depth5Node1, siblings.get(index + 1));
+        byte[] depth3Node1 = combineSibling(depth4Node1, siblings.get(index + 2));
+
+        // Reconstruct the signed block's root hash using Path 0's raw timestamp bytes
+        byte[] depth2Node2 = hashInternalNodeSingleChild(depth3Node1);
+        byte[] timestampLeaf = hashLeaf(timestampPath.timestampLeaf().toByteArray());
```
mhess-swl:
```suggestion
        byte[] hashedTimestampLeaf = hashLeaf(timestampPath.timestampLeaf().toByteArray());
```

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +241,151 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+            return false;
+        }
+
+        MerklePath timestampPath = paths.get(0);
+        MerklePath siblingPath = paths.get(1);
+        MerklePath terminalPath = paths.get(2);
+
+        if (!terminalPath.siblings().isEmpty() || terminalPath.hasTimestampLeaf()) {
+            LOGGER.log(WARNING, "Block {0} state proof path 2 (terminal) is unexpectedly non-empty", blockNumber);
+        }
```
mhess-swl, line 334: Just curious, why log a validation failure reason here, but not log a reason inside the other `if` blocks? 
e.g.
```
        if (!timestampPath.hasTimestampLeaf()) {
            LOGGER.log(WARNING, "Block {0} timestamp leaf is empty", blockNumber);
            return false;
        }
```

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +241,151 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+    }
+
+    /**
+     * Walks the merkle path chain in a state proof to reconstruct the directly-signed block's
+     * root hash, then verifies the TSS signature on that root.
+     *
+     * <p>The state proof contains 3 paths:
+     * <ul>
+     *   <li>Path 0: timestamp leaf of the signed block</li>
+     *   <li>Path 1: sibling hashes from the target block through all gap blocks to the signed block</li>
```
mhess-swl: Helpful to define 'gap' (if it's not already defined elsewhere in this file):

```suggestion
     *   <li>Path 1: sibling hashes from the target block through all 'gap' (i.e. not directly signed) blocks to the signed block</li>
```

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -228,6 +241,151 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+     *   <li>Path 2: terminal (empty)</li>
+     * </ul>
+     *
+     * <p>Path 1's siblings are laid out as groups:
+     * <ul>
+     *   <li>Per gap block (4 siblings): prevBlockRootsHash, depth5Node2, depth4Node2, timestamp</li>
+     *   <li>Final signed block (3 siblings): prevBlockRootsHash, depth5Node2, depth4Node2</li>
+     * </ul>
```
mhess-swl, line 314:
```suggestion
     * <p>Path 1's siblings are laid out as groups:
     * <ul>
     *   <li>Per gap block (4 siblings): prevBlockRootsHash (right), depth5Node2 (right), depth4Node2 (right), <i>already-hashed</i> timestamp (left)</li>
     *   <li>Final signed block (3 siblings): prevBlockRootsHash (right), depth5Node2 (right), depth4Node2 (right)</li>
     * </ul>
```

mhess-swl (approved):
Looks good to me! Well done @AlfredoG87! 

I left some comments that _may_ add clarity, but feel free to ignore them or postpone to a future PR if you'd like 🙂

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: I guess, there is another comment that asked for a comment and I did it without giving it much thought, but great catch, we should make it consistent 👍 💯

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/BlockUtils.java
AlfredoG87: yeah, we should revisit this in the future, I think we only need these blocks here for now.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: The `isLeft` field exists because sibling positions are not fixed — they depend on the tree structure of each block, so we can't safely hardcode expected positions.

That said, detectability has been covered in latest revision: the integrity check added in this PR verifies that the hash reconstructed from the state proof's siblings matches the block root hash computed independently from the actual block content. A wrong sibling position would cause those two hashes to diverge and fail immediately with a log message showing both values:

`Block N state proof integrity check failed: hash reconstructed from path 1 siblings [<hex>] does not match block root hash computed from block content [<hex>]`

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: @Nana-EC taking another look, I realized you were talking about the fail/negative test cases. I've added them now. ✅

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅  Done!
Nice comment.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅ Agreed and Done! 👍

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: good naming of variables is still one of the hardest software dev problems 😅 

Thanks! ✅

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅ Done!

By taking another look I actually also made consistent the failure cases and found another edge case so we can double check the blockRootHash with the BlockStateProof hash for the block is proving.

Thanks for making me take another look :)

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅  Done!

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅  Done!

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -258,14 +447,20 @@ private LedgerIdPublicationTransactionBody findLedgerIdPublication(Bytes signedT
...
 
-    public static <T> T getSingle(List<T> list, Predicate<T> predicate) {
+    /**
+     * Returns the single element matching the predicate, or null if none match.
+     * Logs a WARNING and returns null if multiple elements match.
+     */
+    private <T> T getSingle(List<T> list, Predicate<T> predicate) {
         List<T> filtered = list.stream().filter(predicate).toList();
-
-        if (filtered.size() != 1) {
```
ivannov, line 271: Not sure whether this deletion is a problem. But with this code, when `filtered.size()` was not 1, then the `IlleagalStateException` was caught and handled in the `VerificationServicePlugin.handleBlockItemsReceived` method:

```java
        catch (final RuntimeException | ParseException e) {
            LOGGER.log(WARNING, "Failed to verify BlockItems.", e);
            verificationBlocksFailed.increment();
            sendFailureNotification(currentBlockNumber, BlockSource.PUBLISHER);
        }
```

Now with returning null, this situation is no longer reported as notification failure (just logged) and the block is silently skipped.

Maybe this is the expected behavior, so feel free to resolve it

block-node/verification/src/test/java/org/hiero/block/node/verification/session/impl/StateProofVerificationTest.java
ivannov: Will this actually help here? I think that it ensures that all the tests within this class are run in the same thread. But it does not guarantee that other classes, that may mutate these same fields, aren't run in parallel.

I see that this approach is sticking around for quite some time and is used in other tests. So it seems to work. But maybe it deserves either a task to rework (in other plugins we just have these variables as instance and not static) or at least a general way (via config?) to tell junit platform to run the tests in a single thread

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
mhess-swl: No worries, agreed! 

The reason this distinction matters: the timestamp leaf in path ~~1~~ 0 is _not_ hashed, but any timestamps included in path ~~2~~ 1 _are_ hashed. I had to cross-check this--then had to cross-check this comment itself 😅 so IMO specificity is preferred in this case.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
mhess-swl: > The isLeft field exists because sibling positions are not fixed — they depend on the tree structure of each block, so we can't safely hardcode expected positions.

This is true for sibling hashes _in general_, yes. For clarity's sake, `prevBlockRootsHash, depth5Node2, depth4Node2` must _always_ be right (i.e. `!isLeft()`) siblings, and `timestamp` must _always_ be a left sibling.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
mhess-swl: 💯 🙌

mhess-swl (approved):

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅ 
Good catch — getSingle now throws IllegalStateException on multiple matches, caught in finalizeVerification to return a BAD_BLOCK_PROOF failure notification (same behavior as before), and a test covering this scenario was added.

AlfredoG87:
> Will this actually help here? I think that it ensures that all the tests within this class are run in the same thread. But it does not guarantee that other classes, that may mutate these same fields, aren't run in parallel.
> I see that this approach is sticking around for quite some time and is used in other tests. So it seems to work. But maybe it deserves either a task to rework (in other plugins we just have these variables as instance and not static) or at least a general way (via config?) to tell junit platform to run the tests in a single thread

I think this is a valid concern and we should circle back to this once we deliver the MVP and have some more time, in the mean time, I've created an issue to track and not forget the comment 💯 

https://github.com/hiero-ledger/hiero-block-node/issues/2663

Nana-EC (approved):
LG for initial implementation.
Can move non blocking considerations to new sprint.

rockysingh (commented):
Good work on the state proof verification. The core algorithm and test coverage are solid. A couple of minor items:

1. **`// --- Helpers ---` comment** in `StateProofVerificationTest` line 673: The triple dashes are a bit unusual. Consider just `// Helpers` for consistency with the rest of the codebase.

Otherwise this looks ready to go. I can approve but lets fix that maybe as a follow up?

rockysingh (approved):

PR #2541, opened 2026-04-10
fix: bind metrics server to all interfaces and expose chart values

AlfredoG87:
## Summary

Fixes #2537

The `hiero-metrics` `openmetrics-httpserver` module binds to `127.0.0.1` by default. In Kubernetes, Prometheus scrapes from the pod's cluster IP (not localhost), so metrics requests are refused with "connection refused". This PR fixes the default and exposes metrics server configuration through the Helm chart.

### What changed

**App-level default fix:**
- Added `metrics.exporter.openmetrics.http.hostname=0.0.0.0` to `app.properties` so the metrics server binds to all interfaces out of the box — fixes Docker, bare metal, and Kubernetes deployments alike.
- Removed the now-redundant `JAVA_TOOL_OPTIONS` workaround from `docker-compose.yml`.

**Helm chart — new `blockNode.metrics` section:**
- Exposes `hostname`, `port`, `path`, and `decimalFormat` as operator-configurable values.
- These are injected as JVM system properties (`-D` flags) via `JAVA_TOOL_OPTIONS`, which override the `app.properties` defaults.
- User-provided `JAVA_TOOL_OPTIONS` in `blockNode.config` is preserved and merged with the auto-generated flags (e.g., debug agents, profilers).

**Chart cleanup:**
- Removed `PROMETHEUS_ENDPOINT_ENABLED` and `PROMETHEUS_ENDPOINT_PORT_NUMBER` from `blockNode.config` — the block node app never reads these env vars (only the simulator has the `MappedConfigSource` mapping). They were dead config giving operators a false sense of control.
- Moved metrics port from `blockNode.health.metrics.port` to `blockNode.metrics.port` since metrics server config is not a health concern.

**Documentation:**
- Updated `configuration.md`, `metrics.md`, and `troubleshooting.md` with new property names, chart value references, and bind address troubleshooting guidance.

### Configuration override chain

1. `app.properties` (classpath, lowest priority) — sets defaults
2. `blockNode.metrics.*` chart values → auto-generated `-D` flags in `JAVA_TOOL_OPTIONS` (highest priority) — operator overrides

### Migration for existing deployments

| Before | After |
|--------|-------|
| `PROMETHEUS_ENDPOINT_ENABLED: true` | Removed (no effect on block node) |
| `PROMETHEUS_ENDPOINT_PORT_NUMBER: "16007"` | `blockNode.metrics.port: 16007` |
| `blockNode.health.metrics.port: 16007` | `blockNode.metrics.port: 16007` |

## Test plan

- [x] `helm template` renders `JAVA_TOOL_OPTIONS` with logging + metrics `-D` flags
- [x] `helm template --set blockNode.metrics.port=9090` correctly overrides the port in ConfigMap, Service, and StatefulSet
- [x] `helm template --set blockNode.config.JAVA_TOOL_OPTIONS="-agentlib:..."` merges custom flags with auto-generated ones
- [x] `./gradlew :app:assemble` builds successfully with new `app.properties`
- [ ] Deploy to test cluster and verify `curl http://<pod-ip>:16007/metrics` returns metrics (proving `0.0.0.0` binding)
- [ ] Verify ServiceMonitor scraping works when `kubepromstack.enabled: true`

### DIFF
--- a/charts/block-node-server/values-overrides/rfh-values.yaml
+++ b/charts/block-node-server/values-overrides/rfh-values.yaml
@@ -5,25 +5,24 @@ resources:
 
 blockNode:
     secretRef: "bn-<env>-s3-creds"
-        config:
-            JAVA_OPTS: '-Xms8G -Xmx8G -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="/tmp/dump.hprof"'
-            # Change the following Value with the next expected Block to be
-            # published by the Network.
-            # BLOCK_NODE_EARLIEST_MANAGED_BLOCK: "0"
-            # Archive Node Configuration
-            ARCHIVE_BLOCKS_PER_FILE: "10000" # Default is 100_000
-            ARCHIVE_BUCKET_NAME: "bn-eng-archive-bucket"
-            ARCHIVE_ENDPOINT_URL: "https://storage.googleapis.com"
-            ARCHIVE_BASE_PATH: "previewnet/archive1"
-            ARCHIVE_STORAGE_CLASS: "ARCHIVE"
-            ARCHIVE_REGION_NAME: "us-central1"
-
-        persistence:
-            archive:
-                size: 50Gi
-            live:
-                size: 10Gi
-            logging:
-                size: 2Gi
-            verification:
-                size: 500Mi
+    config:
+        JAVA_OPTS: '-Xms8G -Xmx8G -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="/tmp/dump.hprof"'
+        # Change the following Value with the next expected Block to be
+        # published by the Network.
+        # BLOCK_NODE_EARLIEST_MANAGED_BLOCK: "0"
+        # Archive Node Configuration
+        ARCHIVE_BLOCKS_PER_FILE: "10000" # Default is 100_000
+        ARCHIVE_BUCKET_NAME: "bn-eng-archive-bucket"
+        ARCHIVE_ENDPOINT_URL: "https://storage.googleapis.com"
+        ARCHIVE_BASE_PATH: "previewnet/archive1"
+        ARCHIVE_STORAGE_CLASS: "ARCHIVE"
+        ARCHIVE_REGION_NAME: "us-central1"
+    persistence:
+        archive:
+            size: 50Gi
+        live:
+            size: 10Gi
+        logging:
+            size: 2Gi
+        verification:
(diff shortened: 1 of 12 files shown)


charts/block-node-server/values-overrides/rfh-values.yaml
```diff
@@ -5,25 +5,24 @@ resources:
...
-        persistence:
-            archive:
-                size: 50Gi
-            live:
-                size: 10Gi
-            logging:
-                size: 2Gi
-            verification:
-                size: 500Mi
+    config:
```
Nana-EC, line 8: Good catch

Nana-EC (approved):
Nice add

docs/block-node/configuration.md
```diff
@@ -59,11 +59,24 @@ Each plugin has its own properties, but this focuses on core options and core pl
...
+| Chart Value                  | JVM Property                                 | Description                         |  Default |
+|:-----------------------------|:---------------------------------------------|:------------------------------------|---------:|
+| `blockNode.metrics.hostname` | `metrics.exporter.openmetrics.http.hostname` | Bind address for the metrics server |  0.0.0.0 |
+| `blockNode.metrics.port`     | `metrics.exporter.openmetrics.http.port`     | Prometheus endpoint port            |    16007 |
+| `blockNode.metrics.path`     | `metrics.exporter.openmetrics.http.path`     | HTTP path for metrics endpoint      | /metrics |
+
+> **Note:** These properties come from the `hiero-metrics` library, not from a Block
+> Node `@ConfigData` record. They cannot be set via environment variable
+> mapping (`AutomaticEnvironmentVariableConfigSource`). The chart injects them as
+> JVM system properties through `JAVA_TOOL_OPTIONS`.
```
jsync-swirlds, line 79: The environment variable source _should_ work for all config data, even the metrics library.
Have we tested this?

jsync-swirlds (approved):
Looks good; one small question on documenation.

Also, while the new PR descriptions are very thorough; they're somewhat redundant with the code diff.

It might be good to reduce the verbosity a bit (less about _what_ changed and more about _why_ perhaps?) or include a clear/clean (up to) 10-15 line _summary_ suitable for including as the commit message for the squash commit (so we don't need to track back through PRs when scanning the git history).

docs/block-node/configuration.md
AlfredoG87: Yes, these properties cannot be set from ENV source according to my tests. all others can and should.

PR #2543, opened 2026-04-10
fix: skip malformed GCS filenames with trailing hash suffixes (#2539)

### LINKED ISSUE #2539: Handle malformed GCS filenames with trailing hash suffixes gracefully
## Parent Issue
Part of #2039

## Problem
When network blips occur, corrupted or empty files can be uploaded to the GCS bucket with a trailing hash suffix appended to the file extension (e.g. `-d41d8cd98f00b204e9800998ecf8427e`, which is the MD5 of an empty file). Even after SRE deletes the object, GCS bucket versioning can keep the entry visible in listings.

This causes `ChainFile.Kind.fromFilePath()` to throw `IllegalArgumentException` because the filename no longer matches any known extension (`.rcd`, `.rcd.gz`, `.rcd_sig`):

```
java.lang.IllegalArgumentException: Unknown file type: recordstreams/record0.0.15/2026-03-21T06_42_18.024923000Z.rcd.gz-d41d8cd98f00b204e9800998ecf8427e
    at org.hiero.block.tools.utils.gcp.MainNetBucket.listWithFilePrefix(MainNetBucket.java:383)
```

This crashes the entire `updateDayListings` command for that day.

## Proposed Fix
Make `ChainFile.Kind.fromFilePath()` and/or `MainNetBucket.listWithFilePrefix()` skip unrecognized filenames with a warning instead of throwing. For example:
- Use `.contains()` instead of `.endsWith()` for extension matching, or
- Catch the `IllegalArgumentException` in `listWithFilePrefix()` and log a warning, filtering out the unrecognized blob

## Affected Files
- `tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/ChainFile.java` (`Kind.fromFilePath`)
- `tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/gcp/MainNetBucket.java` (`listWithFilePrefix`)

rockysingh:
## Summary

Fixes #2539.

`updateDayListings` (and any other `MainNetBucket.listWithFilePrefix` caller) was crashing with `IllegalArgumentException: Unknown file type: ...rcd_sig-d41d8cd98f00b204e9800998ecf8427e` when the GCS bucket returned blobs with trailing hash suffixes. This happens when a network blip during upload leaves a corrupt/versioned object visible in the listing — SRE can't always clean them up immediately, and a single bad object shouldn't kill the entire listing run.

## Changes

- **`ChainFile.java`**
  - `Kind.fromFilePath()` now returns `null` instead of throwing for unrecognized paths.
  - New `ChainFile.createOrNull(nodeAccountId, path, size, md5)` static factory that returns `null` for unrecognized paths. The existing throwing constructor is preserved for callers with known-good paths.

- **`MainNetBucket.java`**
  - `listWithFilePrefix()` now uses `ChainFile.createOrNull` and filters nulls with `Objects::nonNull`.
  - Logs a `[WARN] Skipping unrecognized GCS file: <name>` line for each skipped file so operators have visibility into what was skipped.

- **`ChainFileTest.java`** (new)
  - 9 tests covering all recognized file types, both malformed variants (`.rcd_sig-<hash>`, `.rcd.gz-<hash>`), unknown extensions, and `createOrNull` success/null paths.

## Design notes

- **Backward compatible**: The original `new ChainFile(...)` constructor still throws via the underlying `Kind.fromFilePath()` result — only `createOrNull` opts into the null behavior. Anywhere else that constructs a `ChainFile` from a known-good path retains fail-fast semantics.
- **Warning log, not silent skip**: Silent skipping would hide real issues. The warning lets operators see what was skipped.

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/ChainFile.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/ChainFile.java
@@ -58,10 +58,18 @@ public record ChainFile(
      * @return the chain file, or null if unrecognized
      */
     public static ChainFile createOrNull(int nodeAccountId, String path, int size, String md5) {
-        if (Kind.fromFilePath(path) == null) {
+        Kind kind = Kind.fromFilePath(path);
+        if (kind == null) {
             return null;
         }
-        return new ChainFile(nodeAccountId, path, size, md5);
+        return new ChainFile(
+                kind,
+                nodeAccountId,
+                path,
+                blockTimeInstantToLong(extractRecordFileTime(path.substring(path.lastIndexOf('/') + 1))),
+                size,
+                md5,
+                extractSidecarIndex(path));
     }
 
     /**
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/ChainFile.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/ChainFile.java
@@ -48,6 +48,23 @@ public record ChainFile(
     }
 
     /**
+     * Creates a new chain file, or returns {@code null} if the file path has an
+     * unrecognized extension (e.g. malformed GCS files with trailing hash suffixes).
+     *
+     * @param nodeAccountId the node account ID
+     * @param path the path to the file in bucket
+     * @param size the size of the file
+     * @param md5 the MD5 hash of the file
+     * @return the chain file, or null if unrecognized
+     */
+    public static ChainFile createOrNull(int nodeAccountId, String path, int size, String md5) {
+        if (Kind.fromFilePath(path) == null) {
+            return null;
+        }
+        return new ChainFile(nodeAccountId, path, size, md5);
+    }
+
+    /**
      * Extracts the sidecar index from the file path. If the file is not a sidecar file, returns -1.
      * <p>
      *     Example: <code>https://storage.googleapis.com/hedera-mainnet-streams/recordstreams/record0.0.34/sidecar/2024-04-04T18_03_26.007683847Z_01.rcd.gz</code>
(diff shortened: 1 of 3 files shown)


jsync-swirlds (dismissed):

tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/ChainFile.java
```diff
@@ -47,6 +47,23 @@ public ChainFile(int nodeAccountId, String path, int size, String md5) {
...
+     * unrecognized extension (e.g. malformed GCS files with trailing hash suffixes).
+     *
+     * @param nodeAccountId the node account ID
+     * @param path the path to the file in bucket
+     * @param size the size of the file
+     * @param md5 the MD5 hash of the file
+     * @return the chain file, or null if unrecognized
+     */
+    public static ChainFile createOrNull(int nodeAccountId, String path, int size, String md5) {
+        if (Kind.fromFilePath(path) == null) {
```
Nana-EC: Just to note you call this twice, once here to check then again in the constructor.

Nana-EC (dismissed):

tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/ChainFile.java
rockysingh: Good catch. `createOrNull` now computes `Kind.fromFilePath(path)` once and passes it directly to the canonical constructor, bypassing the convenience constructor that would call it a second time.

AlfredoG87 (approved):
LGTM

jsync-swirlds (approved):

Nana-EC (approved):

PR #2574, opened 2026-04-14
chore: audit TRACE logs in publisher, verification,  health, and backfill plugins

### LINKED ISSUE #2573: Promote non-verbose logs trace logs to debug
### Story Form

As a block node operator
I want to view block level life cycle events in logs
So that I can easily observe BN behaviour and discern issues without having to go to TRACE level

### Technical Notes

BN code base currently over utilizes trace logs.
As we harden it's good to audit our logs and distribute them to the right level.
In this case we plan to set default log level to debug and promote choice trace logs to debug

Nana-EC:
## Summary

  - Promotes 35 TRACE log statements to DEBUG or WARNING across stream-publisher, verification, health, and backfill plugins
  - All promoted logs are lifecycle events, per-block milestones, or error/misconfiguration conditions — not high-frequency per-item paths
  - High-frequency logs (Kubernetes probe responses, per-block-item backfill tracking, scan-cycle no-ops) intentionally kept at TRACE

  ## Changes by plugin

  **stream-publisher**
  - `PublisherHandler`: `onComplete/closeConnection` → DEBUG (stream lifecycle, once per publisher disconnect)
  - `LiveStreamPublisherManager`: handler add/remove → DEBUG (rare connection events)

  **verification**
  - `VerificationServicePlugin`: all 10 TRACE logs → DEBUG — init state, per-block session start/complete/notification, backfill verification path. The per-block verification outcome (`Completed block=[N] with result=[X]`) was previously
  invisible without TRACE.

  **health**
  - `HealthServicePlugin`: init completion → DEBUG. Liveness/readiness probe responses kept at TRACE (fire every few seconds from Kubernetes).

  **backfill**
  - `BackfillPlugin`: 11 logs → DEBUG (init guards, gap detection, scheduling decisions, shutdown), 3 → WARNING (path-not-found, parse-failed, peer-availability-failure — all silent no-ops that operators must see)
  - `BackfillFetcher`: backoff skip, available-range discovery → DEBUG; final fetch failure → WARNING
  - `BlockStreamSubscribeUnparsedClient`: onSubscribe/onComplete → DEBUG; onError → WARNING

## Related Issue(s)
Fixes #2573

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
@@ -3,7 +3,7 @@ package org.hiero.block.node.backfill;
 
 import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.INFO;
-import static java.lang.System.Logger.Level.TRACE;
+import static java.lang.System.Logger.Level.WARNING;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.ArrayList;
@@ -116,7 +116,7 @@ public class BackfillFetcher implements PriorityHealthBasedStrategy.NodeHealthPr
 
         for (BackfillSourceConfig node : blockNodeSource.nodes()) {
             if (isInBackoff(node)) {
-                LOGGER.log(TRACE, "Node [{0}] is in backoff, skipping range discovery", node.address());
+                LOGGER.log(DEBUG, "Node [{0}] is in backoff, skipping range discovery", node.address());
                 continue;
             }
             BlockNodeClient currentNodeClient = getNodeClient(node);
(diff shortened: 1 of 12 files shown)


block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
```diff
@@ -288,7 +288,7 @@ public List<BlockUnparsed> fetchBlocksFromNode(BackfillSourceConfig nodeConfig,
                 if (attempt == maxRetries) {
                     markFailure(nodeConfig);
                     // Log exception details on final failure for debugging
-                    LOGGER.log(TRACE, "Final failure stack trace.\n", e);
+                    LOGGER.log(WARNING, "Final failure stack trace.\n", e);
```
ata-nas, line 291: Judging by the comment, it seems that we are logging the final failure. Usually, the other way around is probably more correct. We can retry, but I think printing the initial failure should be more telling, as repeated attempts could produce a different error.

Then, not all exceptions are a warning. We should be careful to not diverge from our logging guidelines.

ata-nas (commented):
Interesting decision to generally bump up the logging levels. It is usually correct to have things in TRACE if they are short lived and used for fine tracing while debugging.

rockysingh (dismissed):

AlfredoG87 (dismissed):
LGTM

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
Nana-EC: Will come back to this

rockysingh (approved):
lgtm

AlfredoG87 (approved):

PR #2575, opened 2026-04-14
chore: audit TRACE logs in block-access-service, blocks-file-recent, blocks-file-historic, facility-messaging, and s3-archive plugins

### LINKED ISSUE #2573: Promote non-verbose logs trace logs to debug

Nana-EC:
## Summary

  - Promotes 16 TRACE log statements to DEBUG across block-access-service, blocks-file-recent, blocks-file-historic, facility-messaging, and s3-archive plugins
  - All changes are DEBUG-only (no WARN promotions in this group)
  - 3 high-frequency messaging paths kept at TRACE: per-item block send, backfilled-block notification, handler unregistration

  ## Changes by plugin

  **block-access-service**
  - `BlockAccessServicePlugin`: client block request log → DEBUG (per-client request, useful for access pattern diagnosis)

  **blocks-file-recent**
  - `BlockFileRecentPlugin`: persistence-started, wrote-block, file-deleted → DEBUG (per-block write confirmation with file path; key for diagnosing storage issues)

  **blocks-file-historic**
  - `BlockFileHistoricPlugin`: wrote-block, batch-move-start, batch-move-success → DEBUG (per-batch archival events; operators monitoring historic archiving need these at DEBUG)

  **facility-messaging**
  - `BlockMessagingFacilityImpl`: 11 logs → DEBUG — init, handler registration (3 variants), verification/persisted/status-update notification sends, start success, stop threads. The `isLoggable(TRACE)` guard on the start-success log updated
  to `isLoggable(DEBUG)`.
  - Kept at TRACE: `sendBlockItems` (highest-frequency path in the system), backfilled-block notification (burst-rate during backfill), handler unregistration (frequency unknown in production)

  **s3-archive**
  - `S3ArchivePlugin`: archive upload scheduling → DEBUG (per-batch trigger with block range; essential for diagnosing upload coverage gaps)

## Related Issue(s)
Fixes #2573

### DIFF
--- a/block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
+++ b/block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.messaging;
 
+import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.TRACE;
 import static org.hiero.block.node.spi.BlockNodePlugin.METRICS_CATEGORY;
 
@@ -249,7 +250,7 @@ public class BlockMessagingFacilityImpl implements BlockMessagingFacility {
 
         // log successful initialization
         LOGGER.log(
-                TRACE,
+                DEBUG,
                 "BlockMessagingFacility initialized with block item queue size: {0} and block notification queue size: {1}",
                 messagingConfig.blockItemQueueSize(),
                 messagingConfig.blockNotificationQueueSize());
@@ -373,7 +374,7 @@ public class BlockMessagingFacilityImpl implements BlockMessagingFacility {
 
         // log the registration of the handler
         LOGGER.log(
-                TRACE,
+                DEBUG,
                 "Registering block item handler: {0}, cpuIntensive: {1}, handlerName: {2}",
                 handler.getClass().getSimpleName(),
                 cpuIntensiveHandler,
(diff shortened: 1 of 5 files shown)


block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
```diff
@@ -461,7 +462,7 @@ public void sendBlockVerification(VerificationNotification notification) {
     public void sendBlockPersisted(PersistedNotification notification) {
         messageForwarder.submit(() -> {
             LOGGER.log(
-                    TRACE,
+                    DEBUG,
```
ata-nas, line 465: Verification and persistence are quite frequent. Are we sure we want to promote?

AlfredoG87 (approved):
LGTM

block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
Nana-EC: At a perf block level I think is fine especially with a block every 2 secs. However, indeed it could become concerning when we're doing 10s and 100's of blocks per second in the future.
This won't be the last audit. 
So should we try this out and modify in the next sprint based on what we're seeing?

rockysingh (approved):

PR #2589, opened 2026-04-15
fix(verification): detect forward block gaps via allBlocksHasher

### LINKED ISSUE #2587: Bug: Forward block gaps pass verification silently
### Description

PR #2471 (`fix: update verification state after live-tail backfill`) changed
`getRootOfAllPreviousBlocks()` in `VerificationServicePlugin`:

```java
// Before #2471 — only deferred to footer when node lacked full chain from genesis
if (earliestManagedBlock > 0 && allBlocksHasherHandler.getNumberOfBlocks() != currentBlockNumber) {
    return null;
}

// After #2471 — defers for ALL hasher/block-number mismatches, including forward gaps
if (allBlocksHasherHandler.getNumberOfBlocks() != currentBlockNumber) {
    return null;
}
```

Returning `null` causes the verification session to fall back to the block footer's own
`previousBlockRootHash`, which is self-consistent. The block passes even though the preceding
block was never verified or stored.

The intent of the change was to support backfill. The side effect is that the forward
gap path — where the hasher correctly returns the wrong root to force a PUBLISHER failure — is now
bypassed unconditionally. We should still support out-of-order backfill but only for Historical gaps not Live_Tail gaps, those should be backfilled in order to avoid any gaps.

**Note:** The publisher forwarder intentionally accepts out-of-order blocks via the RESEND
mechanism (added in #2389/#2419). Verification was the intended backstop for actual gaps. With
this regression, that backstop is gone.

### Steps to Reproduce

1. Start a block node with `verification.allBlocksHasherEnabled=true`.
2. Publish and fully verify blocks 0, 1, 2.
3. Publisher stalls after sending only the header for block 3 (no proof). Connection is severed.
   Block 3 is added to `blocksToResend`.
4. A new publisher connects and sends block 4 directly, ignoring the RESEND(3) signal.
5. **Expected:** Block 4 fails verification — hasher has 3 leaves, block 4 needs 4.
   **Actual:** Block 4 passes. Block 3 is never stored. Gap is silent.

This is the BNCE stalled-connection scenario: the stall triggers `blockIsEnding(3)` and the
RESEND path, but the reconnecting publisher does not fill the gap before resuming forward
streaming.

### Additional context

_No response_

AlfredoG87:
## Summary

- Fix `getRootOfAllPreviousBlocks()` in `VerificationServicePlugin`: narrow the "defer to footer" condition so forward block gaps fall through to `computeRootHash()`, returning the wrong root and triggering a `PUBLISHER` failure — restores gap detection regressed in #2471
- Add `VerificationRegressionTest.blockAfterGapFailsVerificationWhenPrecedingBlockMissing`: blocks 0–2 pass, block 4 (skipping 3) fails with `source=PUBLISHER`
- Add `BlockNodeApiRegressionTest`: E2E regression for the publisher RESEND mechanism — publisher 1 severs mid-block 3, publisher 2 fills the gap via RESEND; asserts blocks 3–5 in storage
- Remove `missingBlockGapWhenPublisherSeversConnectionBeforeEndOfBlock` from `BlockNodeAPITests` (replaced by the new regression class)

## Test plan

- [x] `./gradlew :verification:test`
- [x] `./gradlew :suites:runAPISuites`

Fixes #2587

### DIFF
--- a/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
+++ b/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
@@ -320,12 +320,11 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
 
     private Bytes getRootOfAllPreviousBlocks() {
         if (allBlocksHasherHandler != null && allBlocksHasherHandler.isAvailable()) {
-            // Only use the hasher's computed root when its leaf count matches currentBlockNumber
-            // exactly (i.e. it holds hashes for blocks 0 through currentBlockNumber-1). Any other
-            // count means the hasher is out of sync (e.g. backfill advanced it past the live-stream
-            // position, or the node doesn't have a continuous chain from genesis), so defer to the
-            // block footer's authoritative value.
-            if (allBlocksHasherHandler.getNumberOfBlocks() != currentBlockNumber) {
+            if (allBlocksHasherHandler.getNumberOfBlocks() != currentBlockNumber
+                    && (currentBlockNumber <= previousVerifiedBlockNumber || previousVerifiedBlockNumber == -1)) {
+                // Backfill, re-send, or startup: defer to the block footer's values.
+                // Forward gaps fall through to computeRootHash() instead — returning the wrong
+                // root so verification fails and forces the publisher to resend the missing block.
                 return null;
             }
             return Bytes.wrap(allBlocksHasherHandler.computeRootHash());
--- a/block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationRegressionTest.java
+++ b/block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationRegressionTest.java
@@ -2,6 +2,7 @@
 package org.hiero.block.node.verification;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.hedera.pbj.runtime.ParseException;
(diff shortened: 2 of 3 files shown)


block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
```diff
@@ -320,12 +320,11 @@ public void handleBlockItemsReceived(BlockItems blockItems) {
...
         if (allBlocksHasherHandler != null && allBlocksHasherHandler.isAvailable()) {
-            // Only use the hasher's computed root when its leaf count matches currentBlockNumber
-            // exactly (i.e. it holds hashes for blocks 0 through currentBlockNumber-1). Any other
-            // count means the hasher is out of sync (e.g. backfill advanced it past the live-stream
-            // position, or the node doesn't have a continuous chain from genesis), so defer to the
-            // block footer's authoritative value.
-            if (allBlocksHasherHandler.getNumberOfBlocks() != currentBlockNumber) {
+            if (allBlocksHasherHandler.getNumberOfBlocks() != currentBlockNumber
+                    && (currentBlockNumber <= previousVerifiedBlockNumber || previousVerifiedBlockNumber == -1)) {
+                // Backfill, re-send, or startup: defer to the block footer's values.
```
Nana-EC, line 325: nit: can you expand the comment here. The previous comment gave a sense of what the if condition represented.
This comment talks more about what will happen than what the true case implies and what it suggests.

block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationRegressionTest.java
```diff
@@ -145,6 +146,41 @@ void stalePreviousBlockHashAfterMultiBlockBackfillShouldNotCausePublisherFailure
                         + ", source=" + liveStreamBlock3.source());
     }
 
+    /**
+     * When a block arrives out of sequence (forward gap), the allBlocksHasher has fewer leaves
```
Nana-EC, line 150: An explanation similar to this is what is missing above

block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationRegressionTest.java
```diff
@@ -145,6 +146,41 @@ void stalePreviousBlockHashAfterMultiBlockBackfillShouldNotCausePublisherFailure
...
+     * than the arriving block number requires. The verification service must detect this mismatch
+     * and fail verification with PUBLISHER source, forcing the publisher to resend the missing block.
```
Nana-EC, line 152: Q: How do we ensure the verification failure has the PUBLISHER as source in the logic above?

Nana-EC (commented):
LG, just need a bit more clarity on the comments

jsync-swirlds (approved):

ata-nas (approved):
As we spoke, this should be good for now, we will work on top of this as follow up.

PR #2608, opened 2026-04-16
feat: Improved and Extended Verification Notification

ata-nas:
* Introduced a new `FailureType` for the verification notification
  * If verification has failed, the `FailureType` must not be `null`
  * If verification has succeeded, the `FailureType` must be `null`

## Related Issue(s)

Closes #2529

### DIFF
--- a/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
+++ b/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
@@ -3,18 +3,37 @@ package org.hiero.block.node.spi.blockmessaging;
 
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 
-/**
- * Simple record for block verification notifications.
- *
- * @param success     true if the block was verified successfully, false otherwise
- * @param blockNumber the block number this notification is for
- * @param blockHash   the hash of the block, if the type is BLOCK_VERIFIED
- * @param block       the block, if the type is BLOCK_VERIFIED
- * @param source      the source of the message
- */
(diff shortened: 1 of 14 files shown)


block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
```diff
@@ -3,18 +3,37 @@
...
+        if (success && failureType != null) {
+            throw new IllegalArgumentException("Verification is successful, but a failure reason is provided");
+        }
+        if (!success && failureType == null) {
+            throw new IllegalArgumentException("Verification failed, but no failure reason is provided");
+        }
+    }
+
+    /// The type of failure when verification fails.
+    public enum FailureType {
```
jsync-swirlds, line 31: Can we add `CANCELLED` here as well?  That will be used if verification passes for a much later block but this block is incomplete.
Example:
Block 3 begins verification
Blocks 4, 5, 6 begin _and complete_ verification (all succeed) but are held pending block 3.
Block 3 is cancelled because we have (in this example) 3 successful _later_ blocks, so something is stalling block 3 and we should have it resent.

jsync-swirlds (approved):
One small suggestion, but it can be done later.

Nana-EC (approved):

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
ata-nas: Thanks for the tip, will add in a follow up.

Issue #2289, opened 2026-02-27
Epic: CLI Logic to Pick Up TSS Enablement Transaction Values


rockysingh:
# Goal
Enable the Record Wrapping CLI to deliver TSS parameters to an associated Block Node for presentation via the Block Node _Detail Status_ API.

## Intention
We intend to add CLI logic to detect and extract TSS (Threshold Signature Scheme) enablement transaction values from the _record stream_.
We intend to execute this logic during `blocks wrap` and store that data persistently.
We intend to ensure that this data is available to all block nodes before the cutover release upgrade.

## Technical Details

- A TSS enablement transaction will appear in the record stream containing: new ledger ID, roster (node IDs, weights, TSS public key shares), and aggregate verification key
- The `blocks wrap` command (`ToWrappedBlocksCommand`) needs to detect and extract these values
- **No verification logic changes** — RSA remains the active verification path for record streams.
- Data is captured and delivered to an associated block node for dissemination to other block nodes

Nana-EC:
@rockysingh closing as you address all tickets. Let me know if there was any connected follow up items

PR #2615, opened 2026-04-17
feat: detect and persist TSS enablement transactions using `live-sequential`

### LINKED ISSUE #2098: Adjust the "live record wrapping" CLI to detect and extract TSS parameters from the Record Stream
### Story Form

As a Block Node Operator
I want to Have TSS values available at Cutover
So that my Block Node can verify TSS signatures in block proofs.

### Technical Notes

The TSS parameters for Block Proof verification include WRAPS verification key, Genesis TSS Roster (and hash thereof, a.k.a. Ledger ID), and current TSS Roster.

The Consensus Node will publish TSS system transactions when these values are set and updated. The "Record Wrapping CLI" should read these system transactions and write the latest parameter values to a designated output file.

### LINKED ISSUE #2291: Detect and parse TSS enablement transactions in block stream
### Story Form
As a tools operator running `blocks wrap`,
I want to enhance the CLI to detect TSS enablement transactions in the block stream and parse out the roster, ledger ID, and verification key,
So that these values are captured and available for future TSS verification in release 74.

### Technical Notes
- TSS is enabled but **not signing** — data is captured only
- The `ToWrappedBlocksCommand` 4-stage pipeline processes blocks; detection happens in Stage 1 (Parse + RSA-Verify)

### Additional Items
- Identify TSS enablement transaction type(s) during block parsing (Stage 1 of `ToWrappedBlocksCommand`)
- Extract from enablement transaction: ledger ID, roster (node IDs, weights, TSS public key shares), aggregate verification key
- Log enablement details (block number, consensus timestamp, roster size, key fingerprint)
- Existing RSA verification is unchanged — blocks continue to be verified with RSA signatures

### LINKED ISSUE #2292: Store TSS enablement state across the wrap pipeline
### Story Form
As a tools operator,
I want to store the extracted TSS values (roster, verification key, ledger ID) alongside wrapped block output
So that downstream consumers and future releases can reference the TSS configuration without re-scanning the block stream.

### Technical Notes
- Jumpstart data currently stores: block number (8 bytes), previous block root hash (48 bytes), streaming hasher state
- TSS enablement data should be appended to this checkpoint format

### Aditional Items
- Persist TSS enablement data (enablement block number, roster, verification key, ledger ID) as part of wrap checkpoint/jumpstart data
- Existing RSA verification in the wrap pipeline is unchanged
- TSS data is carried forward but **not used for verification**

### LINKED ISSUE #2293: Expose TSS enablement values in wrap output metadata
### Story Form
As a tools operator,
I want to extend the wrap command to report TSS enablement details in its output/logs when encountered,
So that operators have visibility into the TSS transition point without inspecting individual blocks.

### Additional Items
- Print summary when TSS enablement block is processed (block number, roster node count, ledger ID)
- Include TSS enablement block number in wrap completion summary
- No changes to block verification logic — RSA remains the active verification path

rockysingh:
## Summary
- Adjusts the live record wrapping CLI (`live-sequential`) to detect and extract TSS (Threshold Signature Scheme) enablement parameters from the record stream
- Adds `TssEnablementValidation` as a new block validation that scans blocks for `NodeTssTransactionBody.LEDGER_ID_PUBLICATION` system transactions, extracting the WRAPS verification key, Ledger ID, and TSS Roster
- Introduces `TssEnablementRegistry` to track publication history and serialize to the Block Node `TssData` protobuf format, writing the latest parameter values to a designated output file (`tss-enablement.bin`)
- Extends `TssData` and `TssRoster` protos with `valid_from_block` fields for block-level provenance
- Creates internal `tss_publication_history.proto` for checkpoint persistence across restarts

## Epic
Part of #2289 — Epic: CLI Logic to Pick Up TSS Enablement Transaction Values

Closes #2098
Closes #2291
Closes #2292
Closes #2293


tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/model/TssEnablementRegistry.java
```diff
@@ -0,0 +1,229 @@
...
+    public void writeTssParametersBin(final Path file) {
+        final TssData tssData = getLatestTssData();
+        if (tssData == null) {
+            return;
+        }
+        try {
+            final Bytes bytes = TssData.PROTOBUF.toBytes(tssData);
+            Files.write(file, bytes.toByteArray());
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
```
Nana-EC: File write errors are silently swallowed.
We should handle this better to expose the issue cleanly

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/model/TssEnablementRegistry.java
```diff
@@ -0,0 +1,229 @@
...
+            sb.append("No TSS publications in registry.\n");
+        }
+        return Ansi.AUTO.string(sb.toString());
+    }
+
+    /**
+     * Convert a HAPI {@link LedgerIdPublicationTransactionBody} to the Block Node's
+     * canonical {@link TssData} format.
+     */
+    private static TssData toTssData(final LedgerIdPublicationTransactionBody body, final long blockNumber) {
```
Nana-EC, line 197: Contains no validation for if the data is present and correct in format.
If there was ever a bug we'd just silently pass and not realize

Nana-EC (changes requested):
Quick check, we should validate and guard against bad data and not produce false positives in data

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/model/TssEnablementRegistry.java
rockysingh: Addressed. writeTssParametersBin now throws IOException instead of swallowing it. The caller in TssEnablementValidation catches it and logs a warning so live-sequential keeps running without interruption.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/model/TssEnablementRegistry.java
rockysingh: Addressed. toTssData now validates that ledgerId, historyProofVerificationKey, and nodeContributions are all present and non-empty before building. Throws IllegalArgumentException with a descriptive message including the block number if anything is missing. The outer validate() catch block ensures this won't crash live-sequential.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/TssEnablementValidation.java
```diff
@@ -0,0 +1,184 @@
...
+    public void validate(final BlockUnparsed block, final long blockNumber) throws ValidationException {
+        try {
+            final Instant blockInstant = extractBlockInstant(block);
+            final Bytes recordFileBytes = extractRecordFileBytes(block);
+            if (recordFileBytes == null || recordFileBytes.length() == 0 || blockInstant == null) {
+                return;
+            }
+            processLedgerIdPublications(recordFileBytes, blockInstant, blockNumber);
+        } catch (Exception e) {
+            System.err.println("[TssEnablement] Block " + blockNumber + " - Error extracting TSS data: "
```
Nana-EC, line 78: Q: How does the CLI recover or handle this case?

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/model/TssEnablementRegistry.java
```diff
@@ -0,0 +1,243 @@
...
+        final DatedTssPublication latest = getLatestPublication();
+        return latest == null ? null : latest.tssData();
+    }
+
+    /**
+     * Save the registry to a JSON file.
+     *
+     * @param file the path to the JSON file
+     */
+    public void saveToJsonFile(final Path file) {
```
Nana-EC, line 101: In other places of the CLI we use a save to temp then move logic. Should that be repeated here to avoid concerns over partially written or corrupt files e.g. HasherStateFiles.saveAtomically

Nana-EC (dismissed):
Non-blocking comments but suggestions to harden logic and expand coverage for those scenarios

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/model/TssEnablementRegistry.java
rockysingh: Good call. Updated to use HasherStateFiles.saveAtomically so it writes to a .tmp file first, rotates the existing file to .bak, then renames .tmp into place. This prevents partial/corrupt reads.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/TssEnablementValidation.java
rockysingh: The catch block on line 77-80 logs the error to stderr with the exception class name and message, then continues processing. This means if a single block has a malformed TSS transaction or unexpected data, live-sequential skips it and keeps going with the next block. The TSS registry state remains unchanged (nothing was recorded), so downstream consumers are unaffected. This is consistent with how other validations in live-sequential handle per-block errors: log and continue.

protobuf-sources/src/main/proto/block-node/api/node_service.proto
```diff
@@ -209,6 +213,10 @@ message TssRoster {
      * The list of `RosterEntry`
      */
     repeated RosterEntry roster_entries = 1;
+    /**
+     * The starting block number this TssRoster is valid from
+     */
+    uint64 valid_from_block = 2;
```
jsync-swirlds, line 219: Have we coordinated this change with @berryware?  I think his TSS plugin change adds the same values.

rockysingh: Yes - added it for now we'd need to one of the PR's to merge and then resolve the conflict.

jsync-swirlds (approved):

Nana-EC (approved):

PR #2616, opened 2026-04-18
feat: Query peer BN for TssData in TssBootstrapPlugin

### LINKED ISSUE #2296: Query peer BN for TSS bootstrap data
### Story Form
As a bootstrapping block node,
I want to update the `TSSBootstrapPlugin` to query a peer BN's detailed status via gRPC and extract TSS configuration,
So that a new or restarting node can learn the current TSS state from the network.

### Technical Notes
- The peer BN's `serverStatusDetail` includes TSS fields because its own `TSSBootstrapPlugin` or a `VerificationPlugin` provided them to its `ServerStatusServicePlugin`
- This is how TSS data propagates across the network: each BN bootstraps from a peer, then exposes the data for others
- Follow the same gRPC `WebClient` pattern used elsewhere in the codebase
- Configuration specifies peer addresses via `TSSBootstrapConfig`

### Additional Items
- Use gRPC `WebClient` to call `serverStatusDetail` on configured peer BN(s)
- Retry with backoff on transient failures (configurable max retries and intervals)
- Try multiple configured peers if the first is unavailable
- Extract and validate TSS data (roster, verification key, ledger ID, enablement block number) from the peer's status response

berryware:
## Reviewer Notes
- Adds the ability to query peer BNs for TssData

## Related Issue(s)
Fixes #2296


block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +72,124 @@ public List<Class<? extends Record>> configDataTypes() {
...
+        this.applicationStateFacility = Objects.requireNonNull(context.applicationStateFacility());
+        rosterBootstrapTssConfig = context.configuration().getConfigData(RosterBootstrapTssConfig.class);
+        this.blockNodeContext = context;
+
+        initMetrics();
+
+        // Validate block node sources configuration
+        final String sourcesPath = rosterBootstrapTssConfig.blockNodeSourcesPath();
+        if (sourcesPath == null || sourcesPath.isBlank()) {
+            LOGGER.log(DEBUG, "No block node sources path configured, TssBootstrapPlugin will not query any peers");
```
ivannov: Nit: Maybe this should be `WARNING`

block-node/roster-bootstrap-tss/build.gradle.kts
```diff
@@ -1,17 +1,38 @@
...
+plugins {
+    id("org.hiero.gradle.module.library")
+    id("com.hedera.pbj.pbj-compiler")
+}
 
 description = "Hiero Block Node Roster Bootstrap Tss Service"
 
 dependencies { implementation(project(":base")) }
 
+dependencies { implementation(project(":base")) }
```
ivannov: I think it is declared a couple of lines above. Maybe you can remove it

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
```diff
@@ -0,0 +1,210 @@
...
+                tls,
+                Optional.of(""),
+                "application/grpc",
+                GrpcCompression.IDENTITY,
+                GrpcCompression.getDecompressorNames(),
+                BlockAccessor.MAX_BLOCK_SIZE_BYTES,
+                maxIncomingBufferSize);
+
+        webClient = WebClient.builder()
+                .baseUri("http://" + blockNodeConfig.address() + ":" + blockNodeConfig.port())
```
ivannov: Is this always http? Even when tss is true?

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
```diff
@@ -0,0 +1,210 @@
...
+        final int initialBufferSize = INITIAL_BUFFER_SIZE.getValidOrDefault(tuning);
+
+        return GrpcClientProtocolConfig.builder()
+                .abortPollTimeExpired(false)
+                .pollWaitTime(Duration.ofMillis(pollWaitTimeMs))
+                .initBufferSize(initialBufferSize)
+                .build();
+    }
+
+    public void initializeClient() {
```
ivannov: NIT: Can we reduce the visibility of this method? Maybe private?

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
```diff
@@ -0,0 +1,210 @@
...
+import java.time.Duration;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.ToIntFunction;
+import org.hiero.block.api.BlockNodeServiceInterface;
+import org.hiero.block.node.roster.boostrap.tss.BlockNodeSourceConfig;
+import org.hiero.block.node.roster.boostrap.tss.GrpcWebClientTuning;
+import org.hiero.block.node.spi.historicalblocks.BlockAccessor;
+
+public class BlockNodeClient {
```
ivannov: I think here we manage a couple of clients, which should be closed? Maybe this should be auto closable and in its close metho attempt to close them?

block-node/roster-bootstrap-tss/src/main/proto/bn_source.proto
```diff
@@ -0,0 +1,115 @@
+// SPDX-License-Identifier: Apache-2.0
+syntax = "proto3";
+
+package org.hiero.block.node.roster.boostrap.tss;
```
ivannov: Not sure whether this is a problem, but it should be bootstrap, rather than boostrap

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
berryware: My understanding is that it will force a TLS connection even with just "http://". It is done this way in backfill, NetworkCapacityClient.

block-node/roster-bootstrap-tss/src/main/proto/bn_source.proto
berryware: fixed.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
berryware: Done.

block-node/roster-bootstrap-tss/build.gradle.kts
berryware: Removed.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: Good Catch. Done.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
berryware: I don't see any close() pattern for PbjGrpcClients. @jsync-swirlds @AlfredoG87  any suggestions. I tried to quickly make everything in the Closeable, but then the Helidon WebServer would not start.

ivannov (dismissed):
Looks good to me

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,124 @@ public List<Class<? extends Record>> configDataTypes() {
...
+    /// This method is called on a separate thread. Make sure this.context is marked as `volatile`
+    @Override
+    public void onContextUpdate(BlockNodeContext context) {
+        // save the context update
+        this.blockNodeContext = context;
+    }
+
+    /// queries peer BlockNodes for their TssData
+    private void queryPeerTssData() {
+        List<TssData> tssDataList = tssDataFetcher.getTssData();
```
Nana-EC, line 189: Q: Should we consider quickly checking if the TssData is new? If not then we should fail fast and not call `applicationStateFacility.updateTssData(tssData)`.
If we do that instead of n number of calls, we'd actually have the single latest roster to send.

This would make the logic in this plugin a bit more complex but it may be worth it instead of publishing to every plugin the update

Nana-EC, line 189: At minimum we should skip sending off TssData retrieved that is exactly the same as the current tssData from what's available in context

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,124 @@ public List<Class<? extends Record>> configDataTypes() {
...
+
+        /// Factory method to create a MetricsHolder with all metrics registered.
+        ///
+        /// @param metricRegistry the metrics registry instance to register metrics with
+        /// @return a new MetricsHolder with all metrics created
+        public static MetricsHolder createMetrics(
+                @NonNull final MetricRegistry metricRegistry, @NonNull final AtomicLong currentBlockNodePeers) {
+            metricRegistry.register(ObservableGauge.builder(METRIC_TSS_DATA_PEERS)
+                    .setDescription("Current number of block node peers")
+                    .observe(() -> Math.max(currentBlockNodePeers.get(), 0)));
```
Nana-EC, line 213: I think this always gets 0.
I don't see anywhere where we increment currentBlockNodePeers

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,124 @@ public List<Class<? extends Record>> configDataTypes() {
...
+                    "Failed to parse block node sources from path: [%s], TssBootstrapPlugin will not query any peers: %s"
+                            .formatted(rosterBootstrapTssConfig.blockNodeSourcesPath(), e.getMessage());
+            LOGGER.log(WARNING, parseFailedMsg, e);
+        }
+    }
+
+    /// {@inheritDoc}
+    @Override
+    public void stop() {
+        if (queryPeerExecutor != null) queryPeerExecutor.shutdown();
```
Nana-EC: We should `awaitTermination` and call shutdownNow() after some timeout.
Se the `BackfillPlugin` and `ExpandedCloudStoragePlugin` shutdowns as references

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/TssDataFetcher.java
```diff
@@ -0,0 +1,104 @@
...
+                node, n -> new BlockNodeClient(n, enableTls, maxIncomingBufferSize, n.grpcWebclientTuning()));
+    }
+
+    /// Perform a serverStatusDetail call per configured node and capture the TssData.
+    ///
+    /// @return List of the TssData from each of the Nodes
+    public List<TssData> getTssData() {
+        List<TssData> tssDataList = new ArrayList<>();
+
+        for (BlockNodeSourceConfig node : blockNodeSource.nodes()) {
```
Nana-EC, line 89: There's no consideration here or in the class for priority order of the BN targets.
Non blocking but we should add it in follow up

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
Nana-EC: `PbjGrpcClient` is autocloseable and has a close()

Nana-EC: I wonder if the Clients ( here and in backfill) need to have close() logic added to be able to expose this. 
@AlfredoG87 do you think the gap may be contributing to any resource considerations?

block-node/roster-bootstrap-tss/src/main/proto/bn_source.proto
```diff
@@ -0,0 +1,115 @@
+// SPDX-License-Identifier: Apache-2.0
+syntax = "proto3";
```
Nana-EC, line 2: Should we move this to proto internal then?
Or is there a reason to keep it within the plugin?

Nana-EC (changes requested):
Nice work,
Some improvement considerations.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: This just adds to the ConcurrentList. Only newer ones in the concurrent list will actually get updated.
PR #2756 also adds in making sure that only one update goes out for  all pending changes (TssData and Node Addressbook)

berryware: This just adds to the ConcurrentList. Only newer ones in the concurrent list will actually get updated.
PR #2756 also adds in making sure that only one update goes out for  all pending changes (TssData and Node Addressbook)

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/TssDataFetcher.java
berryware: The implicit ordering is they appear in the config file. PR [#2796](https://github.com/hiero-ledger/hiero-block-node/issues/2796) has been added to capture this for later.

block-node/roster-bootstrap-tss/src/main/proto/bn_source.proto
berryware: This follows how backfill does it. It seems to follow that plugins should be self-contained.
@jsync-swirldds advised that this was okay.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: done

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: Done

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
berryware: implemented Closeable for TssDataFetcher and BlockNodeClient.
call TssDataFetcher.close() in RosterBootstrapTssPlugin.stop()

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
```diff
@@ -209,7 +214,66 @@ public ServerStatusResponse serverStatus(@NonNull final ServerStatusRequest requ
...
+        ///
+        /// @param ledgerId The ledgerId Bytes
+        /// @param wrapsVerificationKey The wrapsVerificationKey Bytes
+        /// @param nodeId The node id
+        /// @param weight The weight
+        /// @param schnorrPublicKey The schnorrPublicKey Bytes
+        /// @param validFromBlock The block from which this TssData is valid
+        /// @param rosterValidFromBlock The block from which this TssRoster is valid
+        /// @return a `TssData` object
+        private TssData buildTssData(
```
Nana-EC, line 249: Q: Good start but is it okay that it's only 1 `RosterEntry`.
Should we add future support for n entries with a predictable variation of the inputs?
Maybe pass in a map of `nodeId -> weight`. Then each nodeId long is added to the `schnorrPublicKey` to make them unique entries.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
AlfredoG87: Last time I checked, there was no need to close Helidon webclients. that is what I remember from backfill refactor at least. but No harm in adding it if is already done.

block-node/roster-bootstrap-tss/src/main/proto/bn_source.proto
AlfredoG87: in the case of backfill we also place it at the plugin level, cause at the time o other plugin was using it, but if we are going to share maybe we can place it in a base plugin that is always present.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,142 @@ public List<Class<? extends Record>> configDataTypes() {
...
+                            .formatted(rosterBootstrapTssConfig.blockNodeSourcesPath(), e.getMessage());
+            LOGGER.log(WARNING, parseFailedMsg, e);
+        }
+    }
+
+    /// {@inheritDoc}
+    @Override
+    public void stop() {
+        if (queryPeerExecutor != null) {
+            queryPeerExecutor.shutdownNow();
```
Nana-EC: This might be overly aggressive.
It should be shutdown() then on timeout of awaitTermination() shutdownNow().
i realize the later we don't do in the other places and actually BackfillPlugin does this, so we need to align.
FYI @AlfredoG87  on backfill
```
queryPeerExecutor.shutdown();
            try {
                if (!queryPeerExecutor.awaitTermination(5, TimeUnit.SECONDS)) {
                    final String executorTerminationMsg = "queryPeerExecutor did not terminate in time.";
                    LOGGER.log(INFO, executorTerminationMsg);
                    queryPeerExecutor.shutdownNow();
                }
            }
```

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -14,6 +32,32 @@
...
 public class RosterBootstrapTssPlugin implements BlockNodePlugin {
+    public static final MetricKey<ObservableGauge> METRIC_TSS_DATA_PEERS =
+            MetricKey.of("tss_data_peers", ObservableGauge.class).addCategory(METRICS_CATEGORY);
+    public static final MetricKey<LongCounter> METRIC_TSS_DATA_ERRORS =
+            MetricKey.of("tss_data_errors", LongCounter.class).addCategory(METRICS_CATEGORY);
+    public static final MetricKey<LongCounter> METRIC_TSS_DATA_REQUESTS =
+            MetricKey.of("tss_data_requests", LongCounter.class).addCategory(METRICS_CATEGORY);
+
+    /// The logger for this class.
+    private final System.Logger LOGGER = System.getLogger(getClass().getName());
```
Nana-EC:
```suggestion
    private static final System.Logger LOGGER = System.getLogger(getClass().getName());
```

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,142 @@ public List<Class<? extends Record>> configDataTypes() {
...
+                tssDataFetcher.close();
+            } catch (IOException e) {
+                LOGGER.log(WARNING, "Unable to close tssDataFetcher: {0}", e);
+            }
+        }
+    }
+
+    /// UncaughtExceptionHandler for logging uncaught exceptions
+    private void uncaughtExceptionHandler(Thread thread, Throwable throwable) {
+        LOGGER.log(WARNING, "Uncaught exception in RosterBootstrapTssPlugin thread: " + thread.getName(), throwable);
```
Nana-EC:
```suggestion
        LOGGER.log(WARNING, "Uncaught exception in RosterBootstrapTssPlugin thread: {0}" + thread.getName(), throwable);
```

block-node/roster-bootstrap-tss/src/test/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPluginTest.java
```diff
@@ -1,39 +1,183 @@
...
+            public void onContextUpdate(BlockNodeContext context) {
+                contextUpdated[0]++;
+                tssData[0] = context.tssData();
+            }
+        };
+
+        start(plugin, new SimpleInMemoryHistoricalBlockFacility(), configOverride);
+
+        // allow some time for the queryingto run
+        parkNanos(3_000_000_000L);
```
Nana-EC: Not a fan of park or sleeps.
Should use a CountDownLatch is possible, we use it in a lot of other places

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
```diff
@@ -0,0 +1,219 @@
...
+                tls,
+                Optional.of(""),
+                "application/grpc",
+                GrpcCompression.IDENTITY,
+                GrpcCompression.getDecompressorNames(),
+                BlockAccessor.MAX_BLOCK_SIZE_BYTES,
+                maxIncomingBufferSize);
+
+        webClient = WebClient.builder()
+                .baseUri("http://" + blockNodeConfig.address() + ":" + blockNodeConfig.port())
```
Nana-EC: Indeed the http regardless of `tls` is consistent but I wonder if we should set it properly based on TLS boolean.
Not in this PR since it affects multiple clients, but something to follow up on

Nana-EC: If we're sure Helidon will negotiate TLS regardless of the scheme then it should be fine

Nana-EC (dismissed):
Good initial go.
Left some non-blocking comments

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: Changed

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: changed

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: you can't use getClass() with a static class.  Changing to:
    private static final System.Logger LOGGER = System.getLogger(RosterBootstrapTssPlugin.class.getName());

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
berryware: Set based on enabledTls

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
berryware: Changed to take a list of RosterEntry.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
jsync-swirlds: This is a common tension.  Do we make the logger static, or do we make it easy to mechanically add them to every class.
If it helps, loggers in most systems (including JUL and Log4J) are singletons; if you "get" the same logger repeatedly, you get the same instance each time.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
jsync-swirlds: Why are we adding the `{0}` here?
That will produce something like "stuff RuntimeException: messageSomeThreadName" because we have no space between the throwable insert and the thread name, and we're inserting the throwable before concatenating the thread name.

block-node/roster-bootstrap-tss/src/test/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPluginTest.java
berryware: changed to use a countdown latch

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
jsync-swirlds, line 1: Just a thought.
At some point we will need to have a Java SDK that has a block node client.
I wonder if we should have a single block node client that's separated out so it's easier to replace with the SDK client?

I know it seems circular, and we've tried to avoid sharing code between plugins so they're more easily severable; but when a block node plugin acts as a client to the block node, it somewhat makes sense to use the Block Node Java SDK as the client implementation....

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,145 @@ public List<Class<? extends Record>> configDataTypes() {
...
+        this.applicationStateFacility = Objects.requireNonNull(context.applicationStateFacility());
+        rosterBootstrapTssConfig = context.configuration().getConfigData(RosterBootstrapTssConfig.class);
+        this.blockNodeContext = context;
+
+        initMetrics();
+
+        // Validate block node sources configuration
+        final String sourcesPath = rosterBootstrapTssConfig.blockNodeSourcesPath();
+        if (sourcesPath == null || sourcesPath.isBlank()) {
+            LOGGER.log(WARNING, "No block node sources path configured, TssBootstrapPlugin will not query any peers");
```
jsync-swirlds: Logs for no configured path are probably informational; they do not indicate a major error, just a minor misconfiguration (which might be that the plugin is included but not needed).

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,145 @@ public List<Class<? extends Record>> configDataTypes() {
...
+            LOGGER.log(WARNING, blockNodeSourcesPathNotFoundMsg, rosterBootstrapTssConfig.blockNodeSourcesPath());
+            return;
+        }
+
+        try {
+            BlockNodeSource blockNodeSources =
+                    BlockNodeSource.JSON.parse(Bytes.wrap(Files.readAllBytes(blockNodeSourcesPath)));
+            // Let the logs know what we loaded.
+            for (BlockNodeSourceConfig node : blockNodeSources.nodes()) {
+                LOGGER.log(INFO, "Loaded peer BN source node: {0}", node);
```
jsync-swirlds: This should be `DEBUG`.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,145 @@ public List<Class<? extends Record>> configDataTypes() {
...
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            }
+        }
+        if (tssDataFetcher != null) {
+            try {
+                tssDataFetcher.close();
+            } catch (IOException e) {
+                LOGGER.log(WARNING, "Unable to close tssDataFetcher: {0}", e);
```
jsync-swirlds: Does this represent a pending failure of the node, risk data loss, or justify waking software engineers (i.e. you!) at 3AM local time to deal with it?
If not, then the log shouldn't be `WARNING` level.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,145 @@ public List<Class<? extends Record>> configDataTypes() {
...
+            }
+        }
```
jsync-swirlds: Multiple try/catch blocks in a single method is a code smell.
Either the method wants to be split into multiple methods, or you could probably move the try/catch to encompass the whole method (and properly treat exceptions as _exceptional_ events).

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,145 @@ public List<Class<? extends Record>> configDataTypes() {
...
+    @Override
+    public void start() {
+        // Don't start the querying thread if there are no peers defined.
+        if (!hasBNSourcesPath) {
+            LOGGER.log(WARNING, "RosterBootstrapTssPlugin: No BN Peers defined");
+            return;
+        }
+        // save the reference of this volatile object.
+        BlockNodeContext context = blockNodeContext;
+        LOGGER.log(INFO, "RosterBootstrapTssPlugin start called");
```
jsync-swirlds: Log levels are deceptively named.  "INFO" is not really "informational", but more "pay close attention".  For "things are fine, here's a progress indication" the appropriate level is either DEBUG or TRACE.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
```diff
@@ -25,7 +69,145 @@ public List<Class<? extends Record>> configDataTypes() {
...
+                currentBlockNodePeers.incrementAndGet();
+            }
+            hasBNSourcesPath = true;
+            tssDataFetcher = new TssDataFetcher(blockNodeSources, rosterBootstrapTssConfig, metricsHolder);
+        } catch (ParseException | IOException e) {
+            final String parseFailedMsg =
+                    "Failed to parse block node sources from path: [%s], TssBootstrapPlugin will not query any peers: %s"
+                            .formatted(rosterBootstrapTssConfig.blockNodeSourcesPath(), e.getMessage());
+            LOGGER.log(WARNING, parseFailedMsg, e);
+        }
```
jsync-swirlds: There are three things in this method (I know I only highlighted part).
1. When we start putting blank lines between blocks, we're implicitly saying "this should probably be a method, but I'm leaving it here instead".
1. When we include a try/catch, the whole point of exceptions is to move exceptional conditions out of the main code flow.  Adding try/catch around anything smaller than the entire method is usually a sign that we're misusing exceptions, catching exceptions we probably shouldn't catch, or otherwise missing something important.
1. This specific block is handling failed parsing.  If we don't want parse to result in actual failure (and hence want to catch its exceptions), then we should wrap _just_ the parse call in a method, with a try/catch, and return null (no log needed, or if included log at _DEBUG_ or _TRACE_). Then we're mostly dealing with return value and the exception handling is moved out of the main code flow (it's in an adapter method that adapts the library "exceptions are result codes" approach to the "results are results and exceptions are exceptional" approach we should prefer).

Note, this at least partly is my preference, exceptions should be treated as _exceptional_.  This is not universal, but I believe it results in _much_ more readable code and a more performant design.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
Nana-EC: Yeah,  in the future. SDK support of BN is on hold for now but open to community contributions as needed

jsync-swirlds: What I'm wondering is if we should, until the SDK is ready, have a single shared BlockNodeClient class (probably in base or common), with clear "replace this with the SDK when that is available" warnings?

block-node/roster-bootstrap-tss/src/main/proto/bn_source.proto
Nana-EC: Non-blocking but we have proto for the APIs in the protobuf-sources. If anything is exposed we definitely need it in there.
If this is internal plugin logic only then it's fair for now to leave it in the plugin as it would ensure related code isn't on the node if that plugin wasn't loaded.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
```diff
@@ -0,0 +1,219 @@
...
+    /// @param enableTls whether to enable TLS for connections
+    /// @param maxIncomingBufferSize the maximum incoming buffer size in bytes for gRPC message reception
+    /// @param tuning optional tuning for timeouts and HTTP/2 settings
+    public BlockNodeClient(
+            @NonNull BlockNodeSourceConfig blockNodeConfig,
+            boolean enableTls,
+            int maxIncomingBufferSize,
+            @Nullable GrpcWebClientTuning tuning) {
+
+        int connectTimeoutMs = CONNECT_TIMEOUT.getValidOrDefault(tuning);
```
Nana-EC: nit: assignments not needed here.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/client/BlockNodeClient.java
```diff
@@ -137,6 +137,8 @@ public BlockNodeClient(
         int pollWaitMs = POLL_WAIT_TIME.getValidOrDefault(tuning);
 
         Tls tls = Tls.builder().enabled(enableTls).build();
+        String protocol = enableTls ? "https://" : "http://";
```
Nana-EC, line 140: nice to do here but repeat in the TSS BlockNodeClient also

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
Nana-EC: Another consideration is should the `BlockNodeClient` be in `spi-plugins`?
Not for this PR though

Nana-EC (commented):
A few nits but nothing blocking.
Nice additions.
Push follow ups to 0.35

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
berryware: Captured: https://github.com/hiero-ledger/hiero-block-node/issues/2800

block-node/roster-bootstrap-tss/src/main/proto/bn_source.proto
berryware: Captured: https://github.com/hiero-ledger/hiero-block-node/issues/2800

block-node/backfill/src/main/java/org/hiero/block/node/backfill/client/BlockNodeClient.java
berryware: Done

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
berryware: removed. Moved the values inline to the grpcConfig and webclient builder calls

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
berryware: Done

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: changed to INFO

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: Changed to DEBUG

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: Changed to INFO.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: Split to multiple methods

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: Changed to DEBUG

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java
berryware: Refactored.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
```diff
@@ -209,7 +214,68 @@ public ServerStatusResponse serverStatus(@NonNull final ServerStatusRequest requ
...
+
+            for (LongRange longRange :
+                    historicalBlockFacility.availableBlocks().streamRanges().toList()) {
+                blockRanges.add(blockRangeBuilder
+                        .rangeStart(longRange.start())
+                        .rangeEnd(longRange.end())
+                        .build());
+            }
+
+            // Todo: add TssData flexibility to TestBlockNodeServer to allow tests to pass in TssData to hand back to
```
AlfredoG87, line 229: do we have a followup ticket captured for this comment?

AlfredoG87 (approved):
LGTM, I really think we should see the best way to share the same source files and proto that is using Backfill, so we reduce the configuration/maintenance effort 👍

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
berryware: Captured: [#2804](https://github.com/hiero-ledger/hiero-block-node/issues/2804)

Nana-EC (approved):

PR #2625, opened 2026-04-21
feat: S3ArchivePlugin Continue Current Upload on Restart

### LINKED ISSUE #2036: `S3ArchivePlugin` Continue Current Upload on Restart
### Story Form

As a S3 Archive Plugin
I want to be able to handle restarts properly
So that I can continue my normal operations

### Technical Notes

- Restarts could happen and we need to handle them.
- We to determine the last block successfully stored by the active multipart upload.
- We need to be able to get the _last_ block in the _last_ successful multipart upload, which can be determined simply by reading back the last multipart item and scanning from the back (which tar supports) to read the last block in the segment.
- This will allow us to startup successfully after a partial upload has happened.

ivannov:
## Reviewer Notes
                                                                                                                                                       
Startup recovery for the cloud-storage-archive plugin and rename the module from `archive-cloud-storage`.
                                                                                                                                                       
### Resume after restart  

Previously, if the node restarted mid-upload the in-progress multipart upload was abandoned on S3 and the plugin would re-archive blocks already covered by the hanging upload, or miss them entirely.

`StartupRecoveryTask` - runs once at plugin startup. It inspects S3 for multipart uploads and handles three cases:                                     
 
1. No hanging uploads - finds the last completed .tar key, parses its group-start via `ArchiveKey`, and returns the next group boundary so the plugin picks up exactly where it left off.
2. One hanging upload - completes the partially-uploaded parts into a temporary readable S3 object, creates a new multipart upload for the same key, then scans the parts backwards to find the last intact block-start boundary. Returns a `RecoveryResult` carrying the existing upload ID, collected ETags, the boundary block number, and any trailing bytes that overflowed the last flushed part.
3. Multiple hanging uploads - aborts all of them and falls back to Case 1 (should not happen in normal operation).                                   
                                                                                                                                                                               
`BlockUploadTask` - the main upload loop is unchanged, but the task now accepts an optional `RecoveryResult`. When one is present it reuses the existing upload ID, seeds the ETag list and accumulation buffer from the recovered boundary state, and starts the block loop from `nextBlockNumber` rather than `firstBlock`.

### Rename to `cloud-storage-archive`

The module, package, config record, and plugin class have been renamed from the `archive-cloud-storage / ArchiveCloudStorage*` naming to `cloud-storage-archive /  CloudStorageArchive*`. This is a mechanical search-and-replace with no behaviour changes in a separate commit.

## Related Issue(s)

Closes #2036


block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
ata-nas, line 1: Do we have tests for this class? I figure a couple of parameterized tests would go a long way here.

AlfredoG87 (commented):
Looking good first pass:

Question: why are we renaming the plugin? does this rename is going to affect published Jar for the Plugin, if so we need to update deployment charts.

does the reaming warrants more documentation changes?

I would argue against renaming plugins at this point to prevent unknown risk that are inherent to renames.

ivannov:
> Looking good first pass:
> 
> Question: why are we renaming the plugin? does this rename is going to affect published Jar for the Plugin, if so we need to update deployment charts.
> 
> does the reaming warrants more documentation changes?
> 
> I would argue against renaming plugins at this point to prevent unknown risk that are inherent to renames.

@AlfredoG87 this plugin is completely new. So it is not published or part of any deployment. It is still work in progress

Its documentation is still about to be written.

As for why the rename: we agreed on a new naming scheme a couple of weeks back and decided to apply it for this and the expanded cloud storage first

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchiveConfig.java
```diff
@@ -33,4 +35,26 @@ enum S3StorageClass {
...
+            violations.add("region name");
+        }
+        if (accessKey.isEmpty()) {
+            violations.add("access key");
+        }
+        if (secretKey.isEmpty()) {
+            violations.add("secret key");
+        }
+        if (bucketName.isEmpty()) {
+            violations.add("bucket name");
```
ata-nas: I believe the correct check is `isBlank`. `isEmpty` will return false for this example `" "`, but that is still invalid.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -85,100 +90,128 @@ public class ArchiveCloudStoragePlugin implements BlockNodePlugin, BlockNotifica
...
     /// {@inheritDoc}
     @Override
     public void start() {
         virtualThreadExecutor = context.threadPoolManager().getVirtualThreadExecutor();
-        LOGGER.log(TRACE, "Archive cloud storage plugin started");
+        // Skip recovery when config is invalid: the handler is not registered, so the result
+        // would never be consumed and the task would fail immediately with a bad-config S3 error.
+        if (configValid) {
+            recoveryFuture = virtualThreadExecutor.submit(new StartupRecoveryTask(config));
+        }
```
ata-nas, line 123: What do we do if config is invalid? We are not healthy. Here it seems that we continue, but we simply do not have a recovery task? Maybe I am missing something.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -85,100 +90,128 @@ public class ArchiveCloudStoragePlugin implements BlockNodePlugin, BlockNotifica
...
-                        }
+        try {
+            if (notification != null && notification.success() && notification.block() != null) {
+                // If the task was cancelled (i.e. stop() was called), do not start a new task or process
+                // the current block since the plugin is shutting down.
+                if (!checkCompletedUpload()) {
+                    // While recovery is running, stash every block.  currentGroupStart is still -1, so
+                    // routeNotification sends it directly to blocksStash without any extra logic.
+                    if (recoveryFuture != null && recoveryFuture.isDone()) {
+                        completeRecovery();
```
ata-nas, line 138: Odd, if the recovery is still running, should we continue forward? This seems strange to me, but maybe I am missing something?

I was imagining that we cannot move forward before we do the recovery task (while initializing).

My thought is that we cannot really continue if we are still determining where our starting point is. If so, not sure if we should first determine that and then run the plugin. It seems to me that we intend run this in the background while we still allow blocks to be collected, but I think we are simply starting a new upload task?

If that is so, we should be careful not to overfill the buffer! Mind you, if we keep ingesting blocks and not acknowledging them in a timely fashion, this will be problematic. Does this task possibly stall too much?

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -85,100 +90,128 @@ public class ArchiveCloudStoragePlugin implements BlockNodePlugin, BlockNotifica
...
+    /// `-1` and the next call to [handleVerification] will fall through to [startNewUploadTask]
+    /// as normal.
+    ///
+    /// When the result carries an upload ID, [BlockUploadTask] is created with the result so
+    /// that it resumes the existing upload and starts its block loop from
+    /// [RecoveryResult#nextBlockNumber()] rather than the group start.
+    private void completeRecovery() throws ExecutionException, InterruptedException {
+        try {
+            final RecoveryResult result = recoveryFuture.get();
+            if (result.currentGroupStart() != -1) {
```
ata-nas, line 196: Else?

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
```diff
@@ -0,0 +1,263 @@
...
+    public RecoveryResult call() throws S3ClientInitializationException, S3ResponseException, IOException {
+        try (final S3Client s3 = new S3Client(
+                config.regionName(),
+                config.endpointUrl(),
+                config.bucketName(),
+                config.accessKey(),
+                config.secretKey())) {
+            final Map<String, List<String>> uploads = s3.listMultipartUploads();
+            final int totalUploads =
+                    uploads.values().stream().mapToInt(List::size).sum();
```
ata-nas, line 63: Not sure about this. Keep in mind that the map contains all multipart uploads for all objects if they have such open. So if we have 2 tars in progress, you will see two map entries. If we have 1 tar in progress, you will have one entry. It seems like we are summing all multiparts of all active uploads, which is, I believe, wrong.

I think we need a different handling here. Possibly `totalUploads = uploads.size();`?

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
```diff
@@ -0,0 +1,263 @@
...
+        String lastPrefix = null;
+        String token = null;
+        do {
+            final S3Client.ListPage page = s3.listObjectsPage(prefix, token, "/", MAX_LIST_RESULTS);
+            if (!page.keys().isEmpty()) {
+                lastPrefix = page.keys().getLast();
+            }
+            token = page.continuationToken();
+        } while (token != null);
+        return lastPrefix;
```
ata-nas: Could we omit the do-while? I think it is less readable and somewhat error prone. Nit. Considered in other places as well. Not striclty against this.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
```diff
@@ -0,0 +1,263 @@
...
+                                newUploadId,
+                                newEtags.size() + 1));
+                    }
+                    blockNumber = partBlockNumber;
+                    // Bytes before the block-start marker; BlockUploadTask prepends these to its buffer on resume.
+                    trailingBytes = Arrays.copyOfRange(partBytes, 0, markerOffset);
+                    break;
+                } catch (NumberFormatException e) {
+                    // Should not happen, but if it does, log it and continue scanning
+                    LOGGER.log(DEBUG, "Invalid block number in part %d of key %s".formatted(i, key), e);
```
ata-nas: Is this the right action? If we have a number format exception, we should really find out why. Not sure if we can continue, but could be wrong on this one.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TarEntries.java
```diff
@@ -0,0 +1,214 @@
...
+    ///
+    /// @param block the block to convert
+    /// @param blockNumber the block number, used to derive the file name inside the tar
+    /// @return a byte array representing the complete tar entry for the block
+    static byte[] toTarEntry(@NonNull final BlockUnparsed block, final long blockNumber) throws IOException {
+        Objects.requireNonNull(block);
+
+        final String currentBlockName = String.format("%0" + BLOCK_NUMBER_WIDTH + "d", blockNumber) + EXTENSION;
+        final byte[] rawBytes = BlockUnparsed.PROTOBUF.toBytes(block).toByteArray();
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream(rawBytes.length);
```
ata-nas: We should probably include the baos in the try-with-resources. Considered everywhere.

On another note, this whole logic could actually be placed into the static compress method of the compression type enum. In that way we can safely start using the static method.

ata-nas (commented):
First pass, only on production logic.
Will have to come back and dig even deeper.
For now, some questions and cleanup suggestions.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TarEntries.java
ivannov: What is the point of caring for closing the byte array output stream? Its close method is a no-op

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: In the `init()` method we have this:

```java
        if (!violations.isEmpty()) {
            // Should be reported to a health facility once we have it
            LOGGER.log(
                    WARNING,
                    "Cloud storage archive plugin is not active because of empty values for the following configurations: {0}",
                    String.join(", ", violations));
        } else {
            // register to listen to block notifications if config is valid
            context.blockMessaging().registerBlockNotificationHandler(this, false, "Cloud Storage Archive");
            configValid = true;
        }
```

So if config is not valid:
* We log a warning (the only way ATM to signal unhealthy plugin)
* We don't register for new block notifications

So yes, we will not recover the plugin, but we will not receive new blocks either.

Probably I should move the recovery task start in the init method to avoid this confusion?

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Else we discovered that we start from 0. The method returns (null-ifying the `recoveryFuture`) and then the calling method (`handleVerification`) will spawn the upload task with "clean state" parameters

I added a comment explaining that

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
ivannov: The `s3.listMultipartUploads()` returns a mapping between an S3 key and the list of its currently active multipart upload IDs. Usually it is a key to a list of one upload ID. But as the API allows for more than one upload ID for a key, I decided to play defensively and sum the values. I don't know when it can happen though

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
ivannov: I do not like do-while's either. But the alternatives would be uglier in my eyes.

Anyway, I did it with a sentinel boolean, see whether you like it more or I should revert to do-while

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
ivannov: My favorite point: how do we handle errors :D

Here the intention is to find the last block start in this part and try to parse its number. If we can't parse the number (which should be very rare), we ignore the whole part and go back to the previous one.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
ivannov: It was tested implicitly by the other classes. But you are right, I added a parameterized test to be completely sure

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchiveConfig.java
ivannov: Good catch! I updated all the checks to `isBlank()`

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Thanks for looking into this! You found a flaw, which I have overlooked.

So, if a recovery is still running and a block arrivea, with the current state of the code we will fire the block upload task indefinitely.

Then when recovery finishes on a subsequent notification, `completeRecovery()` will create another block upload task with the recovered `currentGroupStart` and a fresh `currentBlockQueue`, thus orphaning the first task, which is now blocked forever on the old queue's `take()`.

A solution for this is to add another condition to the `if` below:

```java
        if (currentUploadFuture == null && recoveryFuture == null) {
        }
```

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TarEntries.java
ata-nas: Good practice, otherwise I agree with you that it is no-op.

AlfredoG87 (dismissed):
LGTM

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
jsync-swirlds: Log at WARNING level in this specific case.
This error means either we have a serious bug in our code, or something is writing to the bucket that we don't know about.
Either should result in an alert and investigation to find out what's going on.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TarEntries.java
jsync-swirlds: Atanas is correct, We should use try-with-resources as a matter of good craftsmanship. While baos `.close` does nothing _currently_ nothing guarantees it will never do anything, and nothing guarantees we won't change the implementation type at some point.
Including the try-with-resources now simply ensures that we're less likely to introduce a hard-to-find bug with a seemingly innocent future change.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
jsync-swirlds: I find do-while to be entirely reasonable when the pattern is "do this thing, and if &lt;condition&gt; is true, repeat it, until &lt;condition&gt; is not true.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: Not sure all of the context, but as a general guidance:
if a new block arrives during recovery, then that block is one of these cases:
* completely outside the range of the tar being recovered
   * This should start a new tar upload
* Within the range of the tar being recovered
   * _after_ the last completed block in the recovery tar
      * This should be retained and added to the tar when all preceding gaps filled.
   * _before_ the last completed block in the recovery tar
      * This can be dropped, or can replace the same block read from the recovery tar.

There is a _possibly_ simpler alternative, however.
If we are running recovery, then all new blocks are staged to a temporary tar file (a staging area, in essence), and that is read after the recovery completes as "new" blocks. The result is one or more new active tar tasks that start once recovery is finished.

Either approach might help with this corner case.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: Recovery needs to start in the `start()` method.  Otherwise it's starting before all configuration and context is setup.
In the event that our configuration is not valid, that's a little different.
1. We should have a basic "configuration check" in the init method.  This checks that config is sane, but doesn't start work.
1. _For the time being_ (until we have plugin health reporting). We should log warning, have a todo to mark the plugin unhealthy, and leave the plugin unhealthy after `start()`.

ivannov: If I read this correctly, @jsync-swirlds , I should leave the functionality as is at the moment:
* The plugin logs a warning in case of non-valid config
* The plugin will not receive validation notifications in case of non-valid config
* The startup recovery is running in the `start` method and in `init`

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: @jsync-swirlds there is a little difference between what you wrote, in the area of _completely outside the range..._. You suggest:

> This should start a new tar upload

The plugin doesn't work like that at the moment. We only have one running block upload task. If something is beyond its range, we put it in a block stash rather than handing it to an upload task.

When the upload task is completed (i.e. the current tar is created) and a new block arrives, a new task is spawned for the next tar and the block stash is replayed to check whether those blocks are in the new task's range.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
ivannov: Changed the level to WARNING

AlfredoG87 (approved):
LGTM

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: That's reasonable.
We can consider whether there is a better option in future work.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: We shouldn't attempt recovery during `init`; otherwise this looks good for the current PR.
We _really_ need to extract health to a proper plugin, and start building in plugin health monitoring soon.

Nana-EC (approved):
LG
Let's get this in and iterate, it's getting large and dragging on for non-blocking items

AlfredoG87 (approved):

PR #2631, opened 2026-04-23
feat: add roster-bootstrap-rsa plugin design doc and scaffold for WRB streaming

### LINKED ISSUE #2560: Create a design doc to support WRB Streaming
### Story Form

As a Block Node engineer,
I want a complete, self-contained design document describing how the Block Node ingests, verifies, stores, and serves Wrapped Record Blocks (WRBs),
So that there's a complete description of the approach and work needed for BN to support WRB streaming 

---

### Background

The Hiero network's original upgrade plan called for a single cutover that simultaneously adopted TSS, Block Streams, and Block Nodes. Following a risk assessment, the program was split into two staged cutovers 

- **Phase 2a (WRB Streaming Cutover):** Consensus Nodes (CNs) produce Wrapped Record Blocks (WRBs) and stream them to Block Nodes (BNs) over gRPC, while continuing to upload record files to cloud buckets in parallel. WRBs carry a `SignedRecordFileProof` — a set of gossiped RSA signatures from every node in the current roster — as their sole block proof. BNs verify these proofs using RSA verification logic comparable to what Mirror Nodes and the WRB CLI already use. The Hedera Mirror Node then cuts over to pull from BNs as its primary source, with a record file fallback.

- **Phase 2b (Full Block Stream + TSS):** Record file production ceases. CNs produce full Block Streams with TSS-BLS (hinTS) aggregate proofs. This phase is gated on Phase 2a passing its operational readiness checks.

This ticket covers the design document that must precede any Phase 2a implementation work in the BN repo. It is a **prerequisite** for all engineering tickets under epic [#2509](https://github.com/hiero-ledger/hiero-block-node/issues/2509).

---

### What is a WRB?

A Wrapped Record Block (WRB) is a block in the Hiero Block Stream format whose content is a wrapped record file rather than native block stream items. It is produced by a Consensus Node wrapping a record file it has already generated, hashing it, and emitting the result as a block with a single block proof of type `SignedRecordFileProof`.

The `SignedRecordFileProof` contains:
- The roster node IDs present at signing time
- One RSA signature per roster node, produced over a well-defined payload (the same payload used in record stream signature verification today)

The BN must treat a WRB as a first-class block: verify the proof, persist the block, and serve it via `getBlock` and `subscribeBlockStream` exactly as it would any other block.

### Why RSA and not TSS?

TSS (hinTS, [HIP-1200](https://hips.hedera.com/hip/hip-1200)) is intentionally deferred to Phase 2b. The RSA path reuses well-understood, production-validated cryptographic primitives already in use by Mirror Nodes and the WRB CLI, minimising novel surface area and eliminating the TSS ceremony as a cutover dependency.

### Scope exclusions confirmed by the epic

The roster-maintenance plugin sub-task was explicitly scoped out of epic [#2509](https://github.com/hiero-ledger/hiero-block-node/issues/2509). BN roster awareness for Phase 2a will be handled via configuration, not by a plugin that tracks on-chain roster changes.

---

## Design Doc Scope

The design document **must** address the following topics. Each section heading below is a required section in the output document.

### 1. WRB Block Format and Proof Structure

- Define what constitutes a valid WRB from the BN's perspective: required protobuf fields, expected block item sequence, and `SignedRecordFileProof` structure.
- Specify the payload that is signed and verified (must match the payload used by existing record stream verification in Mirror Nodes and the WRB CLI — confirm alignment).
- Clarify the distinction between a WRB block and a full block stream block at the protobuf level. Downstream consumers (e.g. MN) must be able to tell them apart if needed.
- Note any version or format constraints (e.g. record file versions v2, v5, v6 that may be wrapped).

### 2. Publisher Stream Protocol for WRBs

- Confirm whether the HIP-1081 `publishBlockStream` handshake protocol (`ACCEPT`, `DUPLICATE`, `SKIP`, `SEND_BEHIND`, `RESEND`) applies unchanged to WRBs, or whether any new response codes or states are needed.
- Specify the point in the handshake at which proof type detection occurs and where the RSA verification is triggered.
- Define behaviour when a publisher sends a non-WRB block during Phase 2a (e.g. a full block stream block): reject with `BAD_BLOCK_PROOF`, or a new response?

### 3. RSA Signature Verification

- Algorithm and verification steps, specified precisely enough to be implemented without reference to external code. This should be consistent with what Mirror Nodes and the WRB CLI currently do.
- Threshold requirements: what constitutes a passing verification? All roster nodes? A supermajority by stake weight? Partial valid sets?
- Failure modes and their handling:
  - Zeroed signature
  - Signature from a node not in the current roster
  - Fewer signatures than required by threshold
  - Roster mismatch (proof references an older or newer roster than the BN holds)
- Where verification sits in the block processing pipeline: is it in the existing `BlockVerificationService` / verification plugin, or does it require a new component?

### 4. Roster Configuration

- Define the configuration file format the BN will use to load the current roster and RSA public keys. This format must match the file format used by the TSS bootstrap plugin, per the epic.
- Minimum required fields per roster entry: node ID, RSA public key (and optionally certificate).
- BN startup behaviour: should the BN refuse to start if no roster config is found? Or start and fail verification on every block? Document the chosen approach and the rationale.
- Runtime roster updates: can the BN reload the roster without a restart? If so, what is the mechanism (file watch, admin API, etc.)? If not, document the upgrade procedure for operators.
- Operator tooling: specify the interface for a script that queries the Mirror Node and produces the roster config file. This is the primary upgrade path for decentralised BN operators when the network roster changes.

### 5. Block Storage and Persistence

- Confirm that WRBs are persisted using the existing block storage infrastructure (LFH and RFH paths), with no changes to the `.blk` file format or naming convention.
- Specify any metadata or tagging that distinguishes a stored WRB from a stored full block stream block, if needed.
- Describe how `getBlock` and `serverStatus` behave for WRBs (expected: no change from current behaviour).
- Address the preview block stream gap: when preview block streams halt and WRBs begin, block numbers may have a gap. Specify whether the BN storage and gap detection (`BackfillPlugin`) handles this gracefully, and whether preview blocks should be cleared before WRB ingestion begins.

### 6. Subscriber Compatibility

- Confirm that WRBs are served via `subscribeBlockStream` without modification to the subscription protocol.
- Specify how the Hedera Mirror Node subscription interacts with WRBs: are there any content-level differences it must handle when switching from record file consumption to WRB consumption?
- Address the live/historical boundary for a subscriber that starts subscribing before WRBs begin (i.e. straddles the record stream → WRB transition).

### 7. Optional Cloud Upload

- Specify the cloud bucket path format for WRB uploads (must be clearly distinct from both the existing record file bucket path and the future full block stream bucket path, as this path will be cleared before Phase 2b).
- Confirm that this feature uses existing RFH BN upload infrastructure.
- Note that this is **not required** for Phase 2a cutover — it is for operational validation only.

### 8. Phase 2a → Phase 2b Transition

- Describe how the BN transitions from WRB verification (RSA) to full block stream verification (TSS) at Phase 2b cutover, including whether this requires a BN restart or can be done via configuration change.
- Confirm if WRBs stored during Phase 2a remain fully queryable after the Phase 2b cutover. No migration of stored blocks is required.
- Note the relationship to epic [#1846](https://github.com/hiero-ledger/hiero-block-node/issues/1846) (TSS block proofs for historical wrapped record file blocks) — Phase 2b may retroactively attach TSS proofs to Phase 2a WRBs. The storage format must not preclude this.

### 9. Configuration Schema

- Provide a complete table of new configuration properties introduced by WRB support: property name, type, default, and description.
- Reference existing configuration patterns in the repo for consistency.

### 10. Observability and Metrics

- List all new metrics required for WRB ingestion and verification. At minimum:
  - RSA verification success and failure counters (with `source` label: `publisher`, `backfill`)
  - WRB ingestion rate
  - RSA verification latency (p50, p99)
  - Roster-not-found or roster-mismatch counter
- These must be consistent with the existing BN metric naming conventions.

### 11. Security Considerations

- Roster file integrity: what prevents a bad actor from swapping the on-disk roster config to redirect verification to attacker-controlled keys? Document the threat model and any mitigations (file permissions, checksum, signed roster, etc.).
- Replay attack handling: is this already covered by block number sequencing, or does the `SignedRecordFileProof` payload itself include anti-replay material?

---

## Architectural Considerations

The following are the most important design decision points. Each must either be resolved in the design doc or explicitly deferred with a rationale.

1. **Pluggable verification architecture.** Ensure the existing `BlockVerificationService` abstraction is sufficient to accommodate all proof types with considerations for both phases. A clean approach avoids rewrite and challenges in testing.

2. **Roster loading and validation at startup.** The BN must have the roster loaded and validated before it can accept any WRB. The design must specify where in the startup sequence this happens and what the failure mode is if the roster config is absent or malformed. Failing fast with a clear error is strongly preferred over silently accepting unverifiable blocks.

3. **Proof mode enforcement.** Should there be enforcement of proofs types supported. Does there need to be a "proof mode" configuration flag (e.g. `WRB`, `TSS`, or `ANY` for test environments) ? The design should evaluate this approach.

4. **Cloud upload path isolation.** The WRB cloud upload path must be named and provisioned in a way that makes it unambiguous to DevOps that it is ephemeral and will be cleared before Phase 2b. The design should propose a naming convention and confirm it with the DevOps team as part of the review process.

5. **Decentralised operator upgrade experience.** When the Hiero network roster changes (which happens during software upgrades), every BN operator must update their roster config file before the new roster takes effect. The design should specify the expected lead time, the tooling to generate the updated file, and what happens if an operator misses the window (e.g. whether blocks with the new roster's signatures fail verification until the file is updated).

8. **Test harness gaps.** The block stream simulator does not yet support generating valid or invalid `SignedRecordFileProof` blocks. The design should identify if these are acceptable and if there are other testing gaps.

---

## Acceptance Criteria

- [ ] Design document created and merged at `docs/design/wrb-streaming.md` (or at a path agreed during review)
- [ ] All topics in the **Design Doc Scope** section are addressed; none deferred without documented rationale
- [ ] All **Architectural Considerations** are either resolved with a decision or explicitly deferred with a rationale and a follow-on ticket number
- [ ] RSA verification algorithm is specified precisely enough that an engineer unfamiliar with record stream signing can implement it from the doc alone, without reading Mirror Node source code
- [ ] Roster config format is specified and confirmed to be compatible with the TSS bootstrap plugin file format
- [ ] New configuration properties are listed with names, types, and defaults
- [ ] New metrics are listed with names and label sets consistent with existing BN metric conventions
- [ ] Design reviewed and approved by at least one engineer from: BN team, Execution team (CN perspective), and MN team (subscriber perspective)
- [ ] Follow-on implementation tickets identified and created under epic [#2509](https://github.com/hiero-ledger/hiero-block-node/issues/2509)

---

## References

| Resource | Link |
|---|---|
| HIP-1056 | [Block Node](https://hips.hedera.com/hip/hip-1056) — block items |
| WRB Proof  | [SignedRecordFileProof](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hapi/hedera-protobuf-java-api/src/main/proto/block/stream/signed_record_file_proof.proto) — block items |
| HIP-1081 | [Block Node](https://hips.hedera.com/hip/hip-1081) — publisher handshake protocol |
| HIP-1193 | [Record Stream to Block Stream Cutover](https://hips.hedera.com/hip/hip-1200#hip-1193) — WRB wrapping algorithm |
| HIP-1200 | [hinTS / TSS](https://hips.hedera.com/hip/hip-1200) — Phase 2b proof format (for forward-compatibility context) |

Nana-EC:
## Reviewer Notes
This PR 
 - Introduces the org.hiero.block.node.roster.bootstrap.rsa module — the Block Node plugin responsible for loading the consensus node RSA address book at startup so that Wrapped Record Block (WRB) proofs can be verified during Phase 2a of the
   Hiero network upgrade.
  - Adds the full design document for the plugin, covering the startup sequence, ApplicationStateFacility integration, Mirror Node fetch, RSA signature verification algorithm, proof-type routing, and operator tooling.
  

### docs/design/wrb-streaming/bootstrap-roster-plugin.md
Full design document covering:
  - Plugin lifecycle: init() captures ApplicationStateFacility; start() performs the file-first check then Mirror Node fetch if no local file is present.
  - ApplicationStateFacility.updateAddressBook(): new method (analogous to updateTssData from PR #2492) that writes rsa-bootstrap-roster.pb and fans out onContextUpdate to all plugins.
  - Mirror Node fetch with pagination, 0x-prefix stripping, and retry on failure.
  - Fail-fast behaviour when Mirror Node is unreachable and no local file exists — with an open question on BN shutdown vs. degraded verification.
  - Proof-type routing table (SignedRecordFileProof → RSA, StateProof → state proof path, TssSignedBlockProof → TSS) — no operator-configured proof mode required.
  - Full RSA SHA384withRSA verification algorithm spec (supermajority threshold, payload computation per record file version, failure modes).
  - Phase 2b transition: BN auto-routes by proof type; no config change needed at cutover.
  - Configuration (6 properties under roster.bootstrap.*), metrics, security considerations, operator tooling script spec, acceptance tests.

### RSA Roster Bootstrap Plugin (block-node/roster-bootstrap-rsa)

Adds the new `org.hiero.block.node.roster.bootstrap.rsa` module — the Block Node plugin responsible for loading the consensus node RSA roster at startup so that Wrapped Record Block (WRB) proofs can be verified during Phase 2a of the Hiero network upgrade.

- RsaRosterBootstrapPlugin: BlockNodePlugin skeleton with ordered TODO stubs covering the file-first / Mirror Node fallback / fail-fast startup sequence documented in the design spec. Stores the ApplicationStateFacility reference in init()
  and performs roster loading in start().
  - module-info.java: Java SPI registration (provides BlockNodePlugin with RsaRosterBootstrapPlugin).
  - build.gradle.kts: module build config wired into block-node/app as a runtime plugin alongside tss-bootstrap.
  - README.md: operator reference — startup sequence, binary protobuf address book format (rsa-bootstrap-roster.pb), configuration table, relationship to TssBootstrapPlugin, follow-on ticket mapping.

## Related Issue(s)
Fixes #2560


ata-nas (dismissed):
LG, few of notes:

1. As we discussed, if we have to fetch data, we will then send it to the application state facility to handle, the plugin will not locally store any data (a bit of mismatch with the todo notes and docs)
1. We should migrate to markdown javadocs
1. We should do our logic on start because there could be an application state update after init. We will be able to catch changes on start, possibly seeing the file we are looking for.

We can address all of these in the follow up.

Nana-EC:
> LG, few of notes:
> 
> 1. As we discussed, if we have to fetch data, we will then send it to the application state facility to handle, the plugin will not locally store any data (a bit of mismatch with the todo notes and docs)
> 2. We should migrate to markdown javadocs
> 3. We should do our logic on start because there could be an application state update after init. We will be able to catch changes on start, possibly seeing the file we are looking for.
> 
> We can address all of these in the follow up.

Went ahead and addressed it for clarity reasons

ata-nas (dismissed):

ata-nas (dismissed):

block-node/roster-bootstrap-rsa/README.md
```diff
@@ -0,0 +1,99 @@
...
+| `roster.bootstrap.enabled`                          | `true`                                         | Enable/disable the plugin.                                |
+| `roster.bootstrap.filePath`                         | `data/config/rsa-bootstrap-rosterp.pb`         | Path to the local bootstrap file (binary protobuf).       |
+| `roster.bootstrap.mirrorNode.baseUrl`               | `https://testnet-public.mirrornode.hedera.com` | Mirror Node base URL.                                     |
+| `roster.bootstrap.mirrorNode.connectTimeoutSeconds` | `5`                                            | Connect timeout for Mirror Node calls.                    |
+| `roster.bootstrap.mirrorNode.readTimeoutSeconds`    | `10`                                           | Read timeout for Mirror Node calls.                       |
+| `roster.bootstrap.mirrorNode.pageSize`              | `100`                                          | Nodes per page for paginated Mirror Node calls (max 100). |
+
+---
+
+## Relationship to `TssBootstrapPlugin`
```
berryware, line 79: `TssBootstrapPlugin` has been renamed RosterBootstrapTssPlugin

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+happen in `start()`, after `init()` has stored the facility reference.
+
+```
+start() called
+   │
+   ├─ Bootstrap file exists at config path (rsa-bootstrap-roster.pb)?
+   │       │
+   │       YES ──► Read bytes → NodeAddressBook.PROTOBUF.parse() → validate (≥1 entry)
+   │       │        │
+   │       │        ├─ On parse error or empty book: LOG ERROR → FAIL FAST (throw)
```
ata-nas, line 345: Suggestion: on parse error, we could fall back and try a fetch and replace the file? In addition to logging of course.

Does that conflict with a possibly intentional provision of a file? If so, this behavior could be configurable: if parse fails - fail fast OR go fetch and replace. If we go fetch and replace, we are effectively moving to the next step below, which is the query and the handling there remains same.

ata-nas (approved):

berryware (approved):
TssBootstrapPlugin should be renamed RosterBootstrapTssPlugin, but otherwise looks good.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -0,0 +1,64 @@
...
+        // todo: store BootstrapRosterConfig from context.configuration()
+    }
+
+    /// {@inheritDoc}
+    @Override
+    public void start() {
+        // todo: check if local bootstrap file exists at config.filePath()
+        //         if yes  → parse binary protobuf → validate ≥1 entry with RSA_PubKey + nodeId
+        //                 → build NodeAddressBook
+        //                 → call applicationStateFacility.updateAddressBook(book)
```
jsync-swirlds, line 55: This is slightly off.
The _Application State Facility_ should manage the configured file completely.  This plugin should, in start, simply check if the context contains the NodeAddressBook, (the facility will have loaded the file data and updated context before or just after init).
If the context contains the data, it was loaded from disk.
If not, then we need to load from a remote source.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -0,0 +1,64 @@
...
+        //         if no   → build RosterSource (MirrorNodeRosterSource by default)
+        //                 → fetch roster entries (paginated)
+        //                 → build NodeAddressBook
+        //                 → call applicationStateFacility.updateAddressBook(book)
+        //                   (ApplicationStateFacility handles file persistence and onContextUpdate broadcast)
+        // todo: if roster is empty and failOnFetchError=true → throw to abort BN startup
```
jsync-swirlds, line 61: Instead of throwing, either log at WARNING or higher, or (if we have health support built) mark this plugin unhealthy.

We might keep retrying every so often until we succeed.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -0,0 +1,64 @@
...
+        // todo: emit roster_entries_loaded gauge metric (labelled by source: file | mirror_node)
```
jsync-swirlds, line 62: Do we need a roster_creation_failed metric?

block-node/roster-bootstrap-rsa/build.gradle.kts
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: Apache-2.0
+plugins { id("org.hiero.gradle.module.library") }
+
+description = "Hiero Block Node RSA Roster Bootstrap Plugin"
+
+// Remove the following line to enable all 'javac' lint checks that we have turned on by default
+// and then fix the reported issues.
+tasks.withType<JavaCompile>().configureEach { options.compilerArgs.add("-Xlint:-exports") }
```
jsync-swirlds, line 8: Why are we disabling lint checks?

block-node/roster-bootstrap-rsa/README.md
```diff
@@ -0,0 +1,99 @@
...
+set of gossiped RSA signatures from every node in the current roster as their block proof. Without
+a populated roster the Block Node cannot accept any WRB, so the plugin fails startup fast when
+the roster cannot be loaded.
+
+---
+
+## How it works
+
+1. **File-first:** On startup the plugin checks for a local bootstrap file at the configured path
+   (default `data/config/rsa-bootstrap-roster.pb`). If found, the roster is loaded from that file.
```
jsync-swirlds, line 27:
```suggestion
1. **File-first:** On startup the application state facility checks for a local bootstrap file at the configured path
   (default `data/config/rsa-bootstrap-roster.pb`). If found, the roster is loaded from that file.
```

block-node/roster-bootstrap-rsa/README.md
```diff
@@ -0,0 +1,99 @@
...
+2. **Mirror Node fallback:** If no local file is present, the roster is fetched from the Hedera
+   Mirror Node REST API (`GET /api/v1/network/nodes`, paginated). The result is written to the
+   bootstrap file for future restarts.
+3. **Fail fast:** If neither source succeeds startup is aborted with a clear error message.
```
jsync-swirlds, line 31: We might want block node detail status as an additional source.

block-node/roster-bootstrap-rsa/README.md
```diff
@@ -0,0 +1,99 @@
...
+4. **No runtime reload:** The roster is loaded once and does not change for the lifetime of the
+   BN instance. An address-book change requires a restart with a refreshed bootstrap file.
+
+---
+
+## Bootstrap file format
+
+The bootstrap file is a **binary protobuf serialization** of the `NodeAddressBook` message from
+`basic_types.proto` (`hiero-consensus-node`). This reuses the on-chain address book format stored
+at Hedera file `0.0.102`, avoiding a bespoke schema.
```
jsync-swirlds, line 41:
```suggestion
The bootstrap file is a **binary protobuf serialization** of the `NodeAddressBook` message from
`basic_types.proto` (`hiero-consensus-node`). This reuses the on-chain address book format used for
Hedera file `0.0.102`, avoiding a bespoke schema.
```

block-node/roster-bootstrap-rsa/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+---
+
+## Relationship to `TssBootstrapPlugin`
+
+This plugin parallels `TssBootstrapPlugin` in structure. Both plugins:
+
+- Implement `BlockNodePlugin` and perform all work in `init()`.
+- Use a binary protobuf bootstrap file in `data/config/`.
+- Populate a field on `BlockNodeContext` for downstream consumers.
```
jsync-swirlds, line 85:
```suggestion
- Implement `BlockNodePlugin` and perform all work in `init()`.
- Use a binary protobuf bootstrap file in `data/config/` managed by Application State Facility.
- Populate a field on `BlockNodeContext` for downstream consumers.
```

jsync-swirlds (approved):
Still going through this, but there are a few things that we could update initially.
Updates are design and direction, so they can be in a follow-up.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+- **Checksum (deferred):** An optional SHA-256 checksum of the raw file bytes could be stored alongside the file and
+verified at load time. This is not required for Phase 2a but is tracked as a follow-on hardening item.
+- **Mirror Node TLS:** The initial fetch uses HTTPS. Operators deploying a private Mirror Node should ensure TLS is configured.
+- **Binary format note:** The binary protobuf format is not human-editable, which reduces accidental modification risk
+compared to a JSON file. Intentional modification still requires a protobuf tool (e.g., `protoc --decode`, `prototext`).
+
+### 8.2 Replay attack handling
+
+The WRB has a `BlockHeader` with a block number that the Publisher and Verification logic utilize to ensure that blocks
+that have  already been processed are not replayed.
```
jsync-swirlds, line 639: We actually don't care, at all, about replay. If a publisher sends the same block 100 times then we will keep verifying it until it's persisted, then respond with "duplicate" for all the rest.
Even if we backfilled a block repeatedly, it still only gets stored, it's not like we're executing transactions.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+| `blocknode_rsa_roster_load_duration_seconds` | Gauge     | `source={file,mirror_node}` | Time taken to load the roster at startup (for alerting on slow Mirror Node calls).           |
+
+---
+
+## 8. Security
+
+### 8.1 Bootstrap file integrity
+
+The `rsa-bootstrap-roster.pb` file contains the public keys used for proof verification. A tampered file would allow an
+attacker to substitute keys and have the BN accept forged proofs.
```
jsync-swirlds, line 625: True, but tampering requires filesystem access, and at that point an attacker could inject any content they like without forging proofs.
If the BN is compromised to the point of arbitrary filesystem access (likely as root required to bypass filesystem namespace) then the system is already irrecoverable.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+
+**Mitigations:**
+- **File permissions:** `rsa-bootstrap-roster.pb` should be readable only by the BN process user (`chmod 600`). Document
+this in the operator guide.
+- **Checksum (deferred):** An optional SHA-256 checksum of the raw file bytes could be stored alongside the file and
+verified at load time. This is not required for Phase 2a but is tracked as a follow-on hardening item.
```
jsync-swirlds, line 631: This has very limited value, and is costly in terms of processing.
Also, anything that could replace the block file could replace the checksum, so it is highly questionable there is any value here.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+---
+
+## 9. Operator Tooling
+
+A script `tools-and-tests/scripts/node-operations/generate-roster-rsa-bootstrap.sh` should be provided. Its responsibilities:
+
+1. Accept `--network <mainnet|testnet|previewnet>` and optionally `--mirror-node-url <url>`.
+2. Query `GET /api/v1/network/nodes` (with pagination) to collect all nodes with non-empty `public_key`.
+3. Build a `NodeAddressBook` protobuf, setting only `nodeId` and `RSA_PubKey` (stripped of `0x` prefix) per `NodeAddress` entry.
+4. Serialize with `NodeAddressBook.PROTOBUF.toBytes()` and write to `rsa-bootstrap-roster.pb`.
```
jsync-swirlds, line 650: Is there a mirror-node API to get the full address book as a single JSON response?  That would be simpler; storing the whole address book is simpler and not much different for data size.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+
+---
+
+## 6. Configuration
+
+All properties are under the `roster.bootstrap.rsa` namespace.
+
+|                        Property                         |  Type   |                    Default                     |                                      Description                                       |
+|---------------------------------------------------------|---------|------------------------------------------------|----------------------------------------------------------------------------------------|
+| `roster.bootstrap.rsa.filePath`                         | string  | `data/config/rsa-bootstrap-roster.pb`          | Path to the local bootstrap file (binary protobuf). Relative to BN working directory.  |
```
jsync-swirlds, line 592: This should be in the application state facility config namespace.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+
+**Expected usage before a Phase 2a cutover:**
+
+```bash
+generate-rsa-roster-bootstrap.sh --network mainnet --output /opt/blocknode/data/config/rsa-bootstrap-roster.pb
+```
+
+Operators must regenerate and restart the BN if the address book changes (new nodes added, nodes removed).
+The `blocknode_roster_entries_loaded` gauge metric provides a quick sanity check that the expected number of nodes was
+loaded at startup.
```
jsync-swirlds, line 661: This shouldn't be necessary.  The address book (if managed by application state facility) can be updated dynamically and we probably want this plugin to do so (likely by querying MN daily or hourly to get the last update timestamp).

We _could_ watch the record stream ourselves, but there is a cost there we're trying to avoid.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+
+**Fallback — public NodeAddressBook snapshot:** If the Mirror Node API is unavailable at the time of the pre-cutover
+setup, operators can obtain a copy of the on-chain `NodeAddressBook` (file `0.0.102`) from a publicly available
+snapshot of Hedera state (e.g., a Mirror Node state snapshot or community-maintained archive). The binary contents of
+`0.0.102` are directly compatible with the binary protobuf format expected by the plugin.
```
jsync-swirlds, line 666: Is it worth noting here that this "file" is completely synthetic, is generated by the CN right _after_ upgrade, and might not be generated much longer?

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+| 2 | Should the bootstrap file include a separate SHA-256 checksum sidecar for integrity verification?                                                                                                                                                                                                                                                                                                                                                                                | Deferred — document the risk, implement in Phase 2a operational hardening if time allows.                                                                             |
```
jsync-swirlds, line 713: This isn't worth doing, in my opinion, it adds no security...

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+| 6 | Should the BN at startup check if there have been AddressBook changes that it can quickly retrieve from MN? This would ensure that upgrades are light on the node operator and a simple restart is needed.                                                                                                                                                                                                                                                                       | No, not initially. In future maybe the MAPI query logic can be optimized to update the address book at upgrade boundaries or upon request.                            |
```
jsync-swirlds, line 717: We should probably check address book roughly hourly (much as SDKs do).
Waiting for an upgrade is not easy for BN to manage, and there is little reason not to check more often.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+| 7 | Should `RsaRosterBootstrapPlugin` have a defined lifetime? Should we note a timeline after which it can be removed? Currently, it makes sense to leave it around so that a BN can be the recipient of all WRBs at any point in the future. Note, it will require the MAPI calls to consider time ranges of address books.                                                                                                                                                        | BN should be able to support processing WRBs from genesis going forward by any operator on their own schedule. As such the plugin should remain and be run as needed. |
```
jsync-swirlds, line 718: We can definitely remove this plugin once WRBs all have block proofs generated with a TSSSignature (StateProof type).
After that there is no generally applicable reason for a block node to try to manage the old address books.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+
+When the network transitions from Phase 2a to Phase 2b:
+
+1. Consensus nodes stop emitting WRBs and emit full Block Stream blocks with `TssSignedBlockProof`.
+2. The BN automatically routes `TssSignedBlockProof` blocks to the TSS verification path — no configuration change is
+   required. The `RsaRosterBootstrapPlugin` remains loaded and the `NodeAddressBook` remains in context but the
+   verification layer stops using it for new blocks.
+3. WRBs stored during Phase 2a remain fully queryable via `getBlock` and `subscribeBlockStream` — they are stored as
+   regular `.blk` files and no migration is required. Notably these will not contribute to future state contributions to
+   state services.
```
jsync-swirlds, line 503: Not just "will not contribute", WRBs have no state data at all.


```suggestion
3. WRBs stored during Phase 2a remain fully queryable via `getBlock` and `subscribeBlockStream` — they are stored as
   regular `.blk` files and no migration is required. Notably WRBs have no state data or event data and will not support
   state management services.
```

docs/design/wrb-streaming/bootstrap-roster-plugin.md
Nana-EC: You mean if file parsing fails?
Yeah I don't think we should replace the file in that case as we're unaware if the failure is a network related item that may be restored and we'd inadvertently change the node operators desired address book config

block-node/roster-bootstrap-rsa/build.gradle.kts
Nana-EC: This is in all (if not most plugins)

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+> | `SignedRecordFileProof` | RSA roster verification (this plugin)      |
+> | `StateProof`            | State proof verification (Phase 2a and 2b) |
+> | `TssSignedBlockProof`   | TSS verification (`TssBootstrapPlugin`)    |
+>
+> Both Phase 2a and Phase 2b support `StateProof`. No operator-configured proof mode is needed.
+>
+> **Placement in the pipeline:** RSA verification is performed by the block verification layer
+> (`BlockVerificationService`) immediately after a complete WRB is received from the publisher, before the block is
+> written to storage. The `RsaRosterBootstrapPlugin` supplies the `NodeAddressBook` via `onContextUpdate`; the
+> verification service reads it from `BlockNodeContext.nodeAddressBook()`.
```
jsync-swirlds, line 406:
```suggestion
> written to storage. The `ApplicationStateFacility` supplies the `NodeAddressBook` to plugins via `onContextUpdate`; the
> verification service reads it from `BlockNodeContext.nodeAddressBook()`.
```

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+---
+
+### 4.5 RSA Signature Verification
+
+This section specifies the verification algorithm precisely enough to be implemented without reference to external code.
+The algorithm is derived from the existing `SignatureValidation`, `SignatureDataExtractor`, and `SigFileUtils` classes
+in `tools-and-tests/tools`, which in turn match the verification logic used by Hedera Mirror Node.
+
+> **Proof type routing:** The BN determines which verification path to invoke based on the proof type present in each
+> block. No configuration flag is required:
```
jsync-swirlds, line 393: Perhaps note that Verification will attempt to validate multiple block proofs if multiple proofs are available?

We still need to work out the priority for that; I suspect it will need to change a few times as the network completes the shift from records to TSS in stages.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+either the bootstrap file is present or the Mirror Node is reachable before starting the BN.
+
+> **Open Question (see §12 #1):** Should the BN shut down immediately when the Mirror Node call fails, or should it
+> start but leave RSA verification failing until the address book is populated? The current recommendation is **fail
+> fast**. Additionally, publicly available snapshots of the on-chain `NodeAddressBook` (file `0.0.102`) can serve as
+> a fallback source for operators who cannot reach the Mirror Node API.
+
+**No mid-instance reload:** The roster is loaded once at `start()` and does not change for the lifetime of the BN
+instance. Roster updates (e.g., a node leaving the network) require a BN restart with a refreshed bootstrap file or
+by deleting the file so the next startup fetches a fresh copy from Mirror Node. Runtime reload is tracked in #2563.
```
jsync-swirlds, line 382: I'm not sure why we'd limit the roster to only loading on start.
We can update from mirror node while running without any difficulty; and even if we do not choose to do so initially, I don't think it's wise to design around a non-changing roster.  We should assume that it _might_ change and design to be data-safe if it does.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+
+**Mirror Node down — fail-fast rationale:** An absent or unreadable roster means the BN cannot verify any WRB proof.
+If the address book was not created the BN is in an unhealthy state. A clear error log at shutdown time is preferable
+to silently accepting or rejecting all blocks without a functioning verification layer. Operators are expected to ensure
+either the bootstrap file is present or the Mirror Node is reachable before starting the BN.
+
+> **Open Question (see §12 #1):** Should the BN shut down immediately when the Mirror Node call fails, or should it
+> start but leave RSA verification failing until the address book is populated? The current recommendation is **fail
+> fast**. Additionally, publicly available snapshots of the on-chain `NodeAddressBook` (file `0.0.102`) can serve as
+> a fallback source for operators who cannot reach the Mirror Node API.
```
jsync-swirlds, line 378: I would recommend retrying the mirror node every 5 minutes or so until it succeeds rather than failing the node entirely.

We might need to defer that until we have plugin health reporting, though.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+```
+
+**File-first rationale:** If the bootstrap file already exists there is no need to contact the Mirror Node. The file
+was written on a previous successful startup and represents a known-good address book. Calling Mirror Node on every
+restart would add latency and introduce a network dependency for what is otherwise a deterministic local load.
+
+**Mirror Node down — fail-fast rationale:** An absent or unreadable roster means the BN cannot verify any WRB proof.
+If the address book was not created the BN is in an unhealthy state. A clear error log at shutdown time is preferable
+to silently accepting or rejecting all blocks without a functioning verification layer. Operators are expected to ensure
+either the bootstrap file is present or the Mirror Node is reachable before starting the BN.
```
jsync-swirlds, line 373: Should we include a note that this should change when plugin health reporting is available?  I don't think we want hard failures if we can avoid them.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+   │                                 API unavailable. BN cannot verify WRB proofs."
+   │                     FAIL FAST → shutdown BN
+   │                     (see Open Question #1 below)
+   │
+   └─ Startup complete
+```
+
+**File-first rationale:** If the bootstrap file already exists there is no need to contact the Mirror Node. The file
+was written on a previous successful startup and represents a known-good address book. Calling Mirror Node on every
+restart would add latency and introduce a network dependency for what is otherwise a deterministic local load.
```
jsync-swirlds, line 368: I would still query mirror node, for updates in case the data was updated _after_ the file was written.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+
+**Field mapping:**
+
+| Mirror Node JSON field | `NodeAddress` protobuf field |               Notes               |
+|------------------------|------------------------------|-----------------------------------|
+| `node_id`              | `nodeId` (field 5)           | Direct mapping (int64)            |
+| `public_key`           | `RSA_PubKey` (field 4)       | Strip leading `0x` before storing |
+
+**Filtering:** Only include nodes where `public_key` is non-null and non-empty. Nodes without a public key are not
+eligible for RSA verification and must be excluded from the `NodeAddressBook`.
```
jsync-swirlds, line 293: Do we filter these, or do we count them as failed verifications?
Honestly, there should never be an entry with no public key.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -0,0 +1,719 @@
...
+        NodeAddressBook.PROTOBUF.parse(BufferedData.wrap(raw));
+```
+
+**Default file path:** `data/config/rsa-bootstrap-roster.pb`
+
+**Notes:**
+- The `RSA_PubKey` string in `NodeAddress` is the raw hex-encoded DER bytes without an `0x` prefix. Mirror Node
+returns the key with a leading `0x` which must be stripped before storing.
+- The file does not embed metadata such as network name or generation timestamp. Operators wishing to annotate the
+file should maintain a separate sidecar file or rely on filesystem timestamps.
```
jsync-swirlds, line 267: We should probably keep the generation timestamp, so we know when we get newer data.

jsync-swirlds (commented):
Additional comments (post merge) for followup.

PR #2634, opened 2026-04-23
Fix release missing chart package

### LINKED ISSUE #2633: release(chart): We need to redeploy the Block-Node-chart
### Description

Recently the `block-node-server` instead of `block-node-helm-chart` so now solo can't deploy, since is using the charts deployed at the `block-node-server` paakage.

For this we need to change a bit the release workflow github actions so it skips the jar publish and moves on to chart in case it already exists.

### Steps to Reproduce

_No response_

### Additional context

_No response_

AlfredoG87:
# Summary

Adds a `chart_only` manual trigger input to the release workflow so the `block-node-server` helm chart can be (re)published without rebuilding or re-pushing images, jars, or the simulator.

This unblocks recovery from situations where the chart package is missing on ghcr.io (e.g. accidentally deleted) but the images and jars for that version are already published — re-running the full release fails because Maven Central rejects the duplicate jar upload.

# Motivation

Release `v0.31.0` is missing the `block-node-server` helm chart on ghcr.io, but its images and jars are already published. Re-running the standard release workflow fails on `publishAggregationToCentralPortal` because the jars already exist, which blocks the chart publish from running. See failed run: https://github.com/hiero-ledger/hiero-block-node/actions/runs/24851113192

# Changes

- New `workflow_dispatch` input `chart_only` (boolean, default `false`).
- When `chart_only=true`, the following jobs are skipped:
  - `package-protobuf-source`
  - `publish-app`
  - `publish-jars`
  - `publish-simulator`
  - `helm-chart-release-simulator`
- `helm-chart-release-app` keeps `needs: [publish-app, publish-jars]` for ordering, but its `if` condition now accepts `skipped` dependency results so it still runs when upstream jobs are skipped via `chart_only`.
- Default behavior (tag push, push to `main`, or `workflow_dispatch` with `chart_only=false`) is unchanged.

# How to use (recovery path)

1. Actions → Release Workflow → Run workflow
2. Select the tag/ref for the version whose chart needs to be republished
3. Leave `publish` checked
4. Check `chart_only`
5. Run — only `helm-chart-release-app` executes and pushes `block-node-server-<version>.tgz` to `oci://ghcr.io/hiero-ledger/hiero-block-node`

Fixes #2633

### DIFF
--- a/.github/workflows/release-push-image.yaml
+++ b/.github/workflows/release-push-image.yaml
@@ -26,6 +26,11 @@ on:
         type: boolean
         required: false
         default: false
+      release_ref:
+        description: "Branch or tag to checkout for the release (e.g. v0.31.0). Leave empty to use the ref the workflow was triggered on."
+        type: string
+        required: false
+        default: ""
 
 defaults:
   run:
@@ -62,6 +67,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
+          ref: ${{ github.event.inputs.release_ref || github.ref }}
 
       - name: Extract version
         id: extract_version
@@ -120,6 +126,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
+          ref: ${{ github.event.inputs.release_ref || github.ref }}
 
       - name: Install JDK
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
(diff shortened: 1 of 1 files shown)


.github/workflows/release-push-image.yaml
```diff
@@ -21,6 +21,16 @@ on:
         type: boolean
         required: false
         default: true
+      chart_only:
```
Nana-EC, line 24: I wonder instead if we should make this a flag for which job we want to run.
Meaning it could take the value of any of the job names and if this flag is applicable it checks if the job is part of the list

Nana-EC (commented):
LG but adding a suggestion for opening this up to run any job or jobs

AlfredoG87:
Charts have been fixed using the Branch, no need to merge and create noise anymore.

PR #2637, opened 2026-04-27
feat: New `block-verification` Plugin Module

### LINKED ISSUE #2636: New `block-verification` Plugin Module
### Story Form

As a Block Node Engineer
I want to create a new plugin base for a new verification plugin
So that I can work on it without disrupting the current setup

### Technical Notes

- we have to create a new plugin base (module) for block verification
- we will be developing an improved version of the verification plugin
- we will use the existing plugin while developing the new one and switch over once done with the improved version

ata-nas:
* Created new plugin base for a new verification module
* Migrated to markdown javadoc 

## Related Issue(s)

Resolves #2636

### DIFF
--- a/block-node/block-verification/src/main/proto/root_hash_of_all_previous_blocks_hasher.proto
+++ b/block-node/block-verification/src/main/proto/root_hash_of_all_previous_blocks_hasher.proto
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+syntax = "proto3";
+
+package org.hiero.block.node.verification;
+
+message AllPreviousBlocksRootHashHasherSnapshot {
+    uint64 leaf_count = 1; // Number of leaves processed
+    repeated bytes intermediate_hashes = 2; // Intermediate hashes
+    bytes last_root_hash = 3; // Current root hash
+}
--- a/block-node/block-verification/src/test/java/org/hiero/block/node/verification/VerifiableBlockFactory.java
+++ b/block-node/block-verification/src/test/java/org/hiero/block/node/verification/VerifiableBlockFactory.java
@@ -1,0 +1,113 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.verification;
+
+import static org.hiero.block.common.hasher.HashingUtilities.getBlockItemHash;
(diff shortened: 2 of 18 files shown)


block-node/block-verification/src/main/proto/root_hash_of_all_previous_blocks_hasher.proto
jsync-swirlds, line 1: Just a note, part of future work should probably be to move this to the internal part of the shared proto module.

block-node/block-verification/src/test/java/org/hiero/block/node/verification/VerifiableBlockFactory.java
```diff
@@ -0,0 +1,113 @@
...
+                new NaiveStreamingTreeHasher());
+    }
+
+    private static BlockItemUnparsed toUnparsed(BlockItem item) {
+        try {
+            return BlockItemUnparsed.PROTOBUF.parse(BlockItem.PROTOBUF.toBytes(item));
+        } catch (ParseException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
jsync-swirlds, line 112: Note for future, if we use this, we are doing something wrong.
Once it's parsed, just keep it.
unparsed is only for parsing byte arrays.

jsync-swirlds (approved):

AlfredoG87 (approved):
Unblocking.

block-node/block-verification/src/main/proto/root_hash_of_all_previous_blocks_hasher.proto
ata-nas: I wanted to do that even now, and even tried it, but unfortunately the JPMS does not allow me to have the same generated classes in two places, it conflicts. Will have to do that after we switch.

Thanks for the note!

block-node/block-verification/src/test/java/org/hiero/block/node/verification/VerifiableBlockFactory.java
ata-nas: Thanks, noted! In any case, we should be receiving unparsed items from the RB, it seems that was lost somewhere.

PR #2659, opened 2026-04-29
docs: document AIO requirements

### LINKED ISSUE #2658: docs: Document AIO benchmarks
Per architecture team, though block node does not use the AIO API directly, it is a good approximation for how the block node will handle highly concurrent disk access.

We should document what acceptable AIO numbers are acceptable for the block node in https://github.com/hiero-ledger/hiero-block-node/blob/main/docs/block-node/operations/block-node-hardware-specifications.md

dalvizu:
## Reviewer Notes

This updates documentation to reflect required AIO numbers. This was borrowed from the upper figure for the consensus node 

Closes #2658


jsync-swirlds (dismissed):

Nana-EC (dismissed):
LG, thanks

Nana-EC:
Pinging @dalvizu to sign his commit so it can pass the verified commits CI rule.

Nana-EC (approved):

rockysingh (approved):

PR #2662, opened 2026-04-29
refactor: match PBJ parse size limit with consensus node

### LINKED ISSUE #2641: refactor(pbj-configuration): Match the same Size Limit for PBJ Max Parse Size as CN, from 100MB to 125MB
### Story Form

As a Block Node Operator
I want to Match PBJ Parse Limit as CN
So that BN and CN can parse and process the same size of Blocks.

### Technical Notes

Currently BN is using a 100MB hardcoded limit.

We can make it 125 (we need to confirm if this is the limit on CN)

**See here:**
https://github.com/hiero-ledger/hiero-block-node/blob/1d3c01d1725fb8d8b9263b8c6fb4c8cac76e5eda/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java#L22

but there might be other instances.

rockysingh:
## Summary
- Update `BlockAccessor.MAX_BLOCK_SIZE_BYTES` from 100 MB to **125 MB** to match the consensus node limit
- Consolidate all tools CLI parse sizes to **120 MB** via `ProtobufParsingConstants.MAX_PARSE_SIZE`, replacing stale 36 MB inline literals
- Centralise scattered inline constants to reference `ProtobufParsingConstants.MAX_PARSE_SIZE`

Fixes #2641

### DIFF
--- a/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
+++ b/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
@@ -16,10 +16,11 @@ import org.hiero.block.internal.BlockUnparsed;
  */
 public interface BlockAccessor extends AutoCloseable {
     /**
-     * Maximum protobuf parse size in bytes (100 MB). This accommodates the largest known
-     * block items (TSS Wraps transition blocks) which can exceed PBJ's default 2 MB limit.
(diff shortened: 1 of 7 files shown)


AlfredoG87 (approved):
QQ: why are we using 120MB on tools? but 125 on BN?

LGTM

rockysingh:
@AlfredoG87 its a fail save we see it earlier.

jsync-swirlds (approved):

Nana-EC (approved):

PR #2665, opened 2026-04-30
fix: Add detection and handling for stalled publishers

### LINKED ISSUE #1841: Add handling for publishers that stall in the middle of a block

jsync-swirlds:
## Reviewer Notes
* Added detection based on "future" completed block count
   * This is a constant difference for now, will replace with configuration soon
* Added handler reset/end when its publisher is detected as stalled
* Added tests for the new functionality
* Added a new metric to track stall events

## Related Issue(s)
 Resolves #1841

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,8 @@ protobuf-sources/hiero-consensus-node/
 node_modules/
 k6-proto/
 k6-out/
+
+# LLM agent files
+.agent/**
+agent/**
+
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -4,6 +4,7 @@ package org.hiero.block.node.stream.publisher;
 import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.TRACE;
+import static org.hiero.block.api.PublishStreamResponse.EndOfStream.Code.TIMEOUT;
 import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_BLOCKS_CLOSED_COMPLETE;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_BLOCK_BATCHES_MESSAGED;
@@ -12,6 +13,7 @@ import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_LATEST_BLOCK_NUMBER_ACKNOWLEDGED;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_LOWEST_BLOCK_NUMBER_INBOUND;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_OPEN_CONNECTIONS;
+import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_STALL_TIMEOUTS_SENT;
 
 import com.hedera.pbj.runtime.grpc.Pipeline;
 import edu.umd.cs.findbugs.annotations.NonNull;
@@ -19,6 +21,7 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.ArrayList;
 import java.util.Deque;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NavigableSet;
 import java.util.NoSuchElementException;
@@ -40,6 +43,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 import org.hiero.block.api.PublishStreamResponse;
+import org.hiero.block.api.PublishStreamResponse.EndOfStream.Code;
 import org.hiero.block.internal.BlockItemSetUnparsed;
 import org.hiero.block.internal.BlockItemUnparsed;
 import org.hiero.block.node.app.config.node.NodeConfig;
@@ -60,6 +64,9 @@ import org.hiero.metrics.core.MetricRegistry;
 /// todo(1420) add documentation
 public final class LiveStreamPublisherManager implements StreamPublisherManager {
     private static final int DATA_READY_WAIT_MICROSECONDS = 5000;
+    private static final int MAX_ADVANCE_FOR_STALL_DETECTION = 2;
+    private static final String STALL_DETECTED_LOG_MESSAGE =
+            "Stall detected: handler {0} has not completed block {1}, another handler completed block {2}";
     private final System.Logger LOGGER = System.getLogger(LiveStreamPublisherManager.class.getName());
     private final MetricsHolder metrics;
     private final BlockNodeContext serverContext;
@@ -93,6 +100,11 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     private final ConcurrentNavigableMap<Long, BlockItemUnparsed> blockProofs;
     private final NavigableSet<Long> endBlocksReceived;
     private final NavigableSet<Long> blocksToResend;
+    /// Maps block number to the ID of the handler that currently holds ACCEPT for that block.
+    /// An entry is added when a handler wins ACCEPT via registerQueueForBlock().
+    /// An entry is removed when endOfBlock() is called for that block, when blockIsEnding()
+    /// fires for a mid-block drop, when stall detection removes it, or on shutdown.
+    private final ConcurrentNavigableMap<Long, Long> activeStreamHandlerByBlock;
 
     /// todo(1420) add documentation
     public LiveStreamPublisherManager(
@@ -121,6 +133,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         blockProofs = new ConcurrentSkipListMap<>();
         endBlocksReceived = new ConcurrentSkipListSet<>();
         blocksToResend = new ConcurrentSkipListSet<>();
+        activeStreamHandlerByBlock = new ConcurrentSkipListMap<>();
         initializeBlockNumbers(serverContext);
     }
 
@@ -164,7 +177,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             final long blockNumber, final BlockAction previousAction, final long handlerId) {
         return switch (previousAction) {
             case null -> getActionForHeader(blockNumber);
-            case ACCEPT -> getActionForCurrentlyStreaming(blockNumber);
+            case ACCEPT -> getActionForCurrentlyStreaming(blockNumber, handlerId);
             case END_ERROR, END_DUPLICATE ->
                 // This should not happen because the Handler should have shut down.
                 BlockAction.END_ERROR;
@@ -183,6 +196,9 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             // re-starting it mid-block. We need to remove a potentially collected block proof.
             blockProofs.remove(blockNumber);
         }
+        // Record which handler won ACCEPT for this block so stall detection
+        // can identify and terminate a silent publisher.
+        activeStreamHandlerByBlock.put(blockNumber, handlerId);
     }
 
     @Override
@@ -200,6 +216,11 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             // Remove a proof in case it is collected
             blockProofs.remove(blockNumber);
         }
+        // The completing block is no longer an ACCEPT candidate for stall detection.
+        activeStreamHandlerByBlock.remove(blockNumber);
+        // After removing the completing block, check whether any remaining
+        // ACCEPT holders are stalled relative to this completion.
+        checkForStalledHandlers(blockNumber);
         final long blockToResend = nextBlockToResend();
         if (blockToResend != UNKNOWN_BLOCK_NUMBER) {
             return new ActionForBlock(BlockAction.RESEND, blockToResend);
@@ -213,6 +234,9 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         // @todo(2344) we can further improve this
         LOGGER.log(INFO, "Handler {0} is ending mid-block {1}", handlerId, blockNumber);
         final Deque<BlockItemSetUnparsed> deque = queueByBlockMap.remove(blockNumber);
+        // Remove this block from the ACCEPT winner tracking in all cases;
+        // if the handler dropped mid-block there is no stall to detect here.
+        activeStreamHandlerByBlock.remove(blockNumber);
         if (deque != null) {
             if (blockNumber > lastPersistedBlockNumber.get() && blockNumber < nextUnstreamedBlockNumber.get()) {
                 final String message = "Block {0} will be resent due to handler {1} ending mid block";
@@ -248,7 +272,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         for (final Long nextKey : handlers.keySet()) {
             final PublisherHandler value = handlers.remove(nextKey);
             if (value != null) {
-                value.closeCommunication();
+                value.endStreamWithCode(Code.SUCCESS, false);
                 value.onComplete();
             }
         }
@@ -259,6 +283,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             lastResult.cancel(true);
         }
         queueByBlockMap.clear();
+        activeStreamHandlerByBlock.clear();
         // We can safely shut down the scheduled executor abruptly. The timeout
         // tasks can be ignored completely as we shut down the manager.
         scheduledExecutor.shutdownNow();
@@ -299,6 +324,8 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             blocksToResend.add(blockNumber);
             // Iterate over all handlers and attempt to send the
             // bad block proof message.
+            // @todo(2339) improve the loop below, find the handler that should send the bad block proof code
+            //    and then let it handle the failed verification.
             for (final PublisherHandler handler : handlers.values()) {
                 if (handler.handleFailedVerification(blockNumber)) {
                     // There will always be only one handler that will send the
@@ -357,7 +384,9 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
                 final long blockNumber = notification.blockNumber();
                 // @todo(2198): We may have an extremely rare race condition here
                 nextUnstreamedBlockNumber.set(blockNumber);
-                handlers.values().parallelStream().unordered().forEach(PublisherHandler::handleFailedPersistence);
+                handlers.values().parallelStream()
+                        .unordered()
+                        .forEach((handler) -> handler.endStreamWithCode(Code.PERSISTENCE_FAILED, false));
             }
         }
     }
@@ -396,6 +425,8 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
                     queueByBlockMap.remove(candidate);
                     // possibly remove a just collected block proof
                     blockProofs.remove(candidate);
+                    // remove from stall-detection tracking; this block is now obsolete
+                    activeStreamHandlerByBlock.remove(candidate);
                 }
             }
         }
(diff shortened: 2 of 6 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -299,6 +324,8 @@ public void handleVerification(@NonNull final VerificationNotification notificat
             blocksToResend.add(blockNumber);
             // Iterate over all handlers and attempt to send the
             // bad block proof message.
+            // @todo(2339) improve the loop below, find the handler that should send the bad block proof code
+            //    and then let it handle the failed verification.
```
ata-nas: This feels like merge conflict resolution issue, #2339 is complete and the todo should not be applicable.

Nana-EC (changes requested):
## Review

Good mechanism overall — the `ConcurrentNavigableMap` + CAS `remove(key, value)` pattern for stall detection is correct and the test coverage for the detection logic is solid. Three issues need addressing before merge.

### P1 — Must Fix

**1. Double `onComplete` race when terminating a stalled handler from another thread**

`checkForStalledHandlers()` calls `stalledHandler.endStreamWithCode(TIMEOUT, true)` from the *completing* handler's thread. `immediate=true` resolves to `checkMidBlockAndShutdown()` → `shutdown()` → `replies.onComplete()`, executed on a foreign thread while the stalled handler's Helidon thread could still be live. `sendResponse()` and `onComplete()` have no mutual exclusion. If both threads reach `replies.onNext`/`onComplete` concurrently the gRPC pipeline receives two terminal signals.

Use `immediate=false` for cross-thread stall shutdowns (the scheduler path keeps the final shutdown on the stalled handler's own thread), or add a `compareAndSet(true, false)` guard in `shutdown()` so the second call is a no-op.

**2. `getActionForCurrentlyStreaming()` now accepts `handlerId` but the implementation doesn't appear to use it**

If the body doesn't reference `handlerId`, remove the parameter. If it's placeholder for future use, a `// @todo(N)` comment is required — a dead parameter here will mislead reviewers of the next change that touches this method.

**3. Missing E2E acceptance test for the full recovery path**

The `StallDetectionTests` unit tests confirm the detection mechanism fires correctly. But the full scenario — P1 stalls on block N, P2 completes N+2, stall fires, P2 is asked to resend N, resend flows through messaging — is not exercised end-to-end. `testResendDeliveredAtNextEndOfBlock` confirms `RESEND_BLOCK(0)` is returned from the manager but does not verify the block propagates to messaging after the resend. `BlockNodeApiRegressionTest` is the right place for this; the PR's own acceptance scenario should map 1:1 to a test there.

---

### P2 — Should Fix (can be follow-up tickets if each has a `@todo` reference)

**4. `MAX_ADVANCE_FOR_STALL_DETECTION = 2` needs a config ticket**
Large blocks (state snapshots) can legitimately take more than 2 block-times and would be falsely timed out. The PR notes say "will replace with configuration soon" — file the ticket now and add a `@todo(<number>)` so it doesn't get lost.

**5. `unacknowledgedStreamedBlocks.headSet().clear()` fix should be a separate commit**
This is a standalone correctness fix (previous code only removed the acked block, leaking all lower-numbered entries). Bundling it with stall detection makes both harder to isolate in a revert.

**6. Stall detection log level should be `INFO`, not `DEBUG`**
A stall firing in production is an operational event that ops need without enabling DEBUG. The metric counter alone is not sufficient for incident diagnosis.

**7. `handleEndStream()` mid-block path needs a test**
The new `blockIsEnding()` call when EndStream arrives mid-block is correct, but there's no test verifying the block ends up in `blocksToResend` via this path.

---

### Positives

- CAS on `activeStreamHandlerByBlock.remove(stalledBlock, handlerId)` is the right concurrent primitive.
- Six stall detection unit tests cover boundary, multi-stall, and idempotency cases well.
- `endStreamWithCode()` unification removes duplicated try/finally shutdown patterns.
- `handleEndStream()` mid-block fix closes a real gap where partial blocks were silently lost.
- `publisher_stall_timeouts_sent` metric is the right observability addition.

Full review notes in `agent/pr-reviews/2665.md`.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -200,6 +216,11 @@ public ActionForBlock endOfBlock(final long blockNumber) {
             // Remove a proof in case it is collected
             blockProofs.remove(blockNumber);
         }
+        // The completing block is no longer an ACCEPT candidate for stall detection.
+        activeStreamHandlerByBlock.remove(blockNumber);
```
Nana-EC, line 220: Q: There's 2 removal attempts, they are parallel right or does this result in 2 removal attempts?
The `ConcurrentNavigableMap` will handle but wondering if there's one location that would work instead.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -446,6 +477,50 @@ private <T> T getLastDequeItem(final Deque<T> deque) {
...
+    /// simultaneously.
+    ///
+    /// @param completedBlockNumber the block number that just finished via endOfBlock()
+    private void checkForStalledHandlers(final long completedBlockNumber) {
+        for (final Map.Entry<Long, Long> entry : activeStreamHandlerByBlock.entrySet()) {
+            final long stalledBlock = entry.getKey();
+            final long handlerId = entry.getValue();
+            if (completedBlockNumber > stalledBlock + MAX_ADVANCE_FOR_STALL_DETECTION
+                    && !endBlocksReceived.contains(stalledBlock)
+                    && !blocksToResend.contains(stalledBlock)
```
ata-nas, line 507: Maybe the check for blocks to resend here is redundant. This situation should not be possible theoretically. The manager should not allow two or more publishers to stream the same block simultaneously. If we have a resend for this block that is in the set, this publisher we are now checking for, is the one that would have picked that up. When a publisher starts a resend, the set gets cleared of the block to resend. Something to think about.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -228,9 +228,9 @@ public void sendAcknowledgement(final long newLastAcknowledgedBlockNumber) {
             // if response was sent successfully, we can remove
             // all unacknowledged blocks that are less than or equal to the
             // new last acknowledged block number.
-            // @todo(1582) we have to remove all history, i.e. get an inclusive head set up to
-            //    the new last ackd block and remove that together with lower ones.
-            unacknowledgedStreamedBlocks.remove(newLastAcknowledgedBlockNumber);
```
ata-nas, line 233: This change seems to be the result of a bad merge conflict. The todo was recently added and the switch to remove rather than headset clear was also added recently.

ata-nas (commented):
Some merge conflict resolution issues and a question

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -299,6 +324,8 @@ public void handleVerification(@NonNull final VerificationNotification notificat
             blocksToResend.add(blockNumber);
             // Iterate over all handlers and attempt to send the
             // bad block proof message.
+            // @todo(2339) improve the loop below, find the handler that should send the bad block proof code
```
Nana-EC: This issue is already closed.
Also does this current logic imply that we're currently having all handlers handle the failedVerification even if they didn't send the given block?
if so is this extra info to CN or is it disruptive as some CNs are taking action where they shouldn't?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -596,7 +671,7 @@ private long ensureNextGreaterThanPersisted(final long lastPersisted) {
     }
 
     /// todo(1420) add documentation
-    private BlockAction getActionForCurrentlyStreaming(final long blockNumber) {
+    private BlockAction getActionForCurrentlyStreaming(final long blockNumber, final long handlerId) {
```
ata-nas: `handlerId` param also seems to be a bad merge, it was recently removed due to lack of usage.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -60,6 +64,9 @@
 /// todo(1420) add documentation
 public final class LiveStreamPublisherManager implements StreamPublisherManager {
     private static final int DATA_READY_WAIT_MICROSECONDS = 5000;
+    private static final int MAX_ADVANCE_FOR_STALL_DETECTION = 2;
+    private static final String STALL_DETECTED_LOG_MESSAGE =
```
Nana-EC, line 68: Should we include correlationId here?
To do so you might have to add it to endOfBlock()

jsync-swirlds:
Note, these top-level comments generated from LLM are almost impossible to reply to; it would be good to take this and make individual comments _in code_ so we don't have massive unreadable wall-of-text replies.

> ### P1 — Must Fix
> **1. Double `onComplete` race when terminating a stalled handler from another thread**

This is a false positive.  The race is acceptable and TIMEOUT must be sent immediate; otherwise it will never be sent because the stalled publisher isn't sending data.

> **2. `getActionForCurrentlyStreaming()` now accepts `handlerId` but the implementation doesn't appear to use it**

Will look, this might be a merge issue

> **3. Missing E2E acceptance test for the full recovery path**

This will not be done in this PR.  E2E testing is complex and requires a lot of added code that deserves its own PR.

> ### P2 — Should Fix (can be follow-up tickets if each has a `@todo` reference)
> **4. `MAX_ADVANCE_FOR_STALL_DETECTION = 2` needs a config ticket**

1. This is based on a false premise (large blocks cannot take multiple block times; that's structurally impossible).
2. Config will be in follow-up, this is already addressed _in the PR description_.

> **5. `unacknowledgedStreamedBlocks.headSet().clear()` fix should be a separate commit** 

This is not helpful, it is saying don't make this small change to fix a test because it's not directly related.  Also, might be a merge issue; we'll see.

> **6. Stall detection log level should be `INFO`, not `DEBUG`**

There is a metric for alerting and detection, DEBUG is still appropriate.

> **7. `handleEndStream()` mid-block path needs a test**

This is not related to this PR, I did not add tests for previously untested functionality (handleEndStream is not new, and neither is the mid-block path).

.gitignore
```diff
@@ -68,3 +68,8 @@ protobuf-sources/hiero-consensus-node/
 node_modules/
 k6-proto/
 k6-out/
+
+# LLM agent files
```
Nana-EC, line 72: :)

Nana-EC (dismissed):
LG, non-blocking items to pick up in follow up.
Let's get this in an rc1 and iterate

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I'm not totally certain we've handled all of the relevant cases, but I'll remove this todo comment.

ata-nas:
A few bad merge conflict resolutions are seen, we should double-check for unintentional additions/deletions juxtaposed to main. Also, spotless run is needed.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: No, this is one of the unintentional re-introductions due to a merge resolution. Only a single publisher has supplied the block that failed and should handle the failed verification by closing.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Without a check for resend here, a handler that is receiving a resent block would get marked stalled and we'd immediately end that handler with timeout.

Scenario:
Blocks 2-5 accepted, received, ended.  Block 3 fails persistence and is scheduled to resend.
Handler 2 begins streaming block 5.
Handler 1 begins streaming block 3.
Handler 2 gets end of block for block 5.
(at this point, without the resend check, handler 1 would get endStream(TIMEOUT))


I think I see the concern, `blocksToResend` will not contain the block after it _starts_ to resend.
I need to add a slightly different check, otherwise _all_ resent blocks are likely to be marked stalled and ended prematurely.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: I _added_ the headset clear, based on the todo comment.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: There is only one removal from this map in this method, or did I miss something?
There are 2 _different maps_ that have remove calls, and those handle different things.

jsync-swirlds:
> A few bad merge conflict resolutions are seen, we should double-check for unintentional additions/deletions juxtaposed to main. Also, spotless run is needed.

Will do another conflict pass.

Spotless already passed this code, not sure what appears to be off...

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I'll add this for follow-up.
It cannot be passed in because it would be for the wrong handler; I'll have to get it from the handler that's stalled.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Removed.

Nana-EC (approved):

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
Nana-EC: 2 removals in the class not the method.
Iw as referring to the possibility of removals happening twice for the same block flow.
One in endOfBlock() and the other in blockIsEnding()

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Removed.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Ahh, understand now.
Both different removals are needed, they happen in different pathways.
In one case we're removing the block as a _candidate_ because it's done.
In the other case we're removing it because it failed, we know it failed, and we're in the process of moving past it in the calling handler.

We need both pathways.

rockysingh (approved):

rbarker-dev (approved):
Signing off on change to .gitignore...

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
Nana-EC: Yeah, I know we need both scenarios, I was checking if there was a shared path to handle both.
I'm okay with this.

Nana-EC:
> Note, these top-level comments generated from LLM are almost impossible to reply to; it would be good to take this and make individual comments _in code_ so we don't have massive unreadable wall-of-text replies.
> 
> > ### P1 — Must Fix
> > **1. Double `onComplete` race when terminating a stalled handler from another thread**
> 
> This is a false positive. The race is acceptable and TIMEOUT must be sent immediate; otherwise it will never be sent because the stalled publisher isn't sending data.
> 
> > **2. `getActionForCurrentlyStreaming()` now accepts `handlerId` but the implementation doesn't appear to use it**
> 
> Will look, this might be a merge issue
> 
> > **3. Missing E2E acceptance test for the full recovery path**
> 
> This will not be done in this PR. E2E testing is complex and requires a lot of added code that deserves its own PR.
> 
> > ### P2 — Should Fix (can be follow-up tickets if each has a `@todo` reference)
> > **4. `MAX_ADVANCE_FOR_STALL_DETECTION = 2` needs a config ticket**
> 
> 1. This is based on a false premise (large blocks cannot take multiple block times; that's structurally impossible).
> 2. Config will be in follow-up, this is already addressed _in the PR description_.
> 
> > **5. `unacknowledgedStreamedBlocks.headSet().clear()` fix should be a separate commit**
> 
> This is not helpful, it is saying don't make this small change to fix a test because it's not directly related. Also, might be a merge issue; we'll see.
> 
> > **6. Stall detection log level should be `INFO`, not `DEBUG`**
> 
> There is a metric for alerting and detection, DEBUG is still appropriate.
> 
> > **7. `handleEndStream()` mid-block path needs a test**
> 
> This is not related to this PR, I did not add tests for previously untested functionality (handleEndStream is not new, and neither is the mid-block path).

😲 Claude escaped it's sandbox 😅 , I asked it to do a local review, didn't realize it pushed upwards, i'm gonna go tell it off now 🚶🏾‍♂️ .

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -268,36 +268,26 @@ boolean handleFailedVerification(final long blockNumber) {
...
+            sendEndOfStream(codeToSend);
         } finally {
-            scheduleShutdown();
+            if (immediate) {
+                checkMidBlockAndShutdown(currentStreamingBlockNumber.get());
+            } else {
+                scheduleShutdown();
+            }
         }
     }
```
AlfredoG87, line 292: Suggestion — add isActive.set(false) in the immediate=true branch so the handler's own Helidon thread stops processing new requests before the foreign stall-detection thread calls shutdown() → replies.onComplete():


```suggestion
      void endStreamWithCode(final Code codeToSend, final boolean immediate) {
          try {                                                                                                                                                                                                                                                                                             
              LOGGER.log(DEBUG, "[{0}] Handler {1} ending with code {2}", correlationIdPrefix, handlerId, codeToSend);
              sendEndOfStream(codeToSend);                                                                                                                                                                                                                                                                  
          } finally {
              if (immediate) {                                                                                                                                                                                                                                                                              
                  isActive.set(false);
                  checkMidBlockAndShutdown(currentStreamingBlockNumber.get());                                                                                                                                                                                                                              
              } else {
                  scheduleShutdown();                                                                                                                                                                                                                                                                       
              }   
          }
      }

```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -731,6 +720,12 @@ private PublisherRequestResult handleEndError(
...
         LOGGER.log(logLevel, "[{0}] {1}", correlationIdPrefix, message);
+        final long blockInProgress = currentStreamingBlockNumber.get();
+        if (isCurrentlyMidBlock(blockInProgress)) {
+            // This should generally not happen, we expect an end stream request
+            // from a publisher after it has completely streamed a full block.
+            publisherManager.blockIsEnding(blockInProgress, handlerId);
+        }
         metrics.endStreamsReceived.increment();
         return new ShutdownResult(this, currentStreamingBlockNumber.get());
     }
```
AlfredoG87, line 731: double-calls: ShutdownResult.handle() → checkMidBlockAndShutdown(blockInProgress) → isCurrentlyMidBlock() checks currentBlockQueue.get() != null, which is still non-null (queue reference wasn't cleared) → calls blockIsEnding() a second time → logs INFO "Handler N is ending mid-block M"
  again. Second call is functionally harmless but the duplicate log is confusing.                                                                                                                                                                                                                           
   
  Suggestion — call resetState() after blockIsEnding() to clear currentBlockQueue, so isCurrentlyMidBlock() returns false on the second pass through checkMidBlockAndShutdown. Also use the already-captured blockInProgress in the return instead of a second currentStreamingBlockNumber.get() (which is  
  now UNKNOWN_BLOCK_NUMBER after reset):

```suggestion
private PublisherRequestResult handleEndStream(final Level logLevel, final String message) {
          LOGGER.log(logLevel, "[{0}] {1}", correlationIdPrefix, message);                                                                                                                                                                                                                                  
          final long blockInProgress = currentStreamingBlockNumber.get();
          if (isCurrentlyMidBlock(blockInProgress)) {                                                                                                                                                                                                                                                       
              // This should generally not happen, we expect an end stream request
              // from a publisher after it has completely streamed a full block.                                                                                                                                                                                                                            
              publisherManager.blockIsEnding(blockInProgress, handlerId);
              resetState();                                                                                                                                                                                                                                                                                 
          }
          metrics.endStreamsReceived.increment();                                                                                                                                                                                                                                                           
          return new ShutdownResult(this, blockInProgress);
      }

}
```

AlfredoG87 (approved):
Good Job, Nothing blocking.

Just 2 nit suggestions that can be thought and applied later.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: I'll add this in follow-up.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: Will address in follow-up.  I think there are some race conditions I'll have to handle to add this.

PR #2666, opened 2026-04-30
fix(verification): disable allBlocksHasher by default and add missing docs for VerificationPlugin.

### LINKED ISSUE #2628: Investigate Bad Block Proof Returned for Valid Blocks
### Technical Notes

- We are seeing cases where BAD_BLOCK_PROOF is returned by verification for blocks that successfully pass verification in Mirror Node.
- The signature and the block itself are known to be good.
- We need to find out why block node verification plugin fails.
- BNCE

AlfredoG87:
## Summary

- Change `allBlocksHasherEnabled` default from `true` to `false` — the hasher requires a strictly sequential block stream, and out-of-order or forward-arriving blocks cause it to fall out of sync and produce incorrect root hashes
- Add Verification Plugin Configuration section to `configuration.md` documenting all four `VERIFICATION_*` env vars with a note explaining why the hasher must remain disabled

Fixes #2628


Nana-EC (approved):

ata-nas (approved):
LG

PR #2673, opened 2026-05-01
fix: raise signature threshold to 1/2 of nodes for live-sequential

### LINKED ISSUE #2664: fix: live-sequential fails at live edge when signature stake weight is below threshold
### Story Form

As a Block Node Operator running live-sequential
I want the tool to wait for sufficient signatures at the live edge
So that validation does not fail when only a few nodes have published their signatures

### Problem

When live-sequential catches up to the current block (the live edge), only a small number of consensus nodes may have published their signature files. The tool downloads whatever signatures are available and queues the block for wrap+validate. The signature validation then fails because the stake from the available signatures is below the required 1/3 threshold.

**Example error:**
```
Validation 'Signatures' failed at block 94398862:
Block: 94398862 - Insufficient validated stake:
176209106600000000/1401578583100000000 (4 nodes), need 467192861033333334/1401578583100000000
blockTime=2026-04-30T01:08:26.113474Z, signatures=4, mode=stake-weighted
```

Only 4 out of 33 nodes had signatures available (~12.6% stake vs required ~33.3%).

### Technical Notes

- The prefetch window in `fillPrefetchWindow()` stops when `sigFiles < 3`, but 3 signatures is far below the 1/3 stake threshold
- The `waitForSufficientSignatures()` logic may not be retrying long enough at the live edge
- The download loop should either wait longer for signatures to appear or implement a backoff/retry when at the live edge
- This only happens when the tool has caught up to the current time and new blocks are being produced in near real-time

rockysingh:
## Summary
- Raised minimum signature threshold from hardcoded `3` to `(N/2)+1` in `fillPrefetchWindow()`, `waitForSufficientSignatures()`, and `checkLiveEdgeSignatures()`
- Added early-proceed check in `checkLiveEdgeSignatures()` — proceeds immediately when `sigCount >= minSigs` instead of waiting for ALL signatures
- Increased `MAX_SIG_WAIT` from 2 minutes to 5 minutes for live-edge signature collection

## Context
At the live edge, only a few nodes may have published signatures. With 33 mainnet nodes, 3 sigs is ~9% of nodes — far below the 1/3 stake threshold required by `SignatureValidation`. This caused the wrap thread to crash:
```
Validation 'Signatures' failed at block 94398862:
Insufficient validated stake: 176209106600000000/1401578583100000000 (4 nodes),
need 467192861033333334/1401578583100000000
```

closes #2664

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
@@ -1332,7 +1332,7 @@ public class LiveSequential implements Runnable {
                                         .getCurrentAddressBook()
                                         .nodeAddress()
                                         .size()
-                                / 3
+                                / 2
                         + 1;
                 if (!hasRecord || sigFiles < minSigs) {
                     break;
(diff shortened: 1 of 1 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
```diff
@@ -1327,7 +1328,13 @@ private void fillPrefetchWindow(
                 long sigFiles = group.stream()
                         .filter(f -> f.type() == ListingRecordFile.Type.RECORD_SIG)
                         .count();
-                if (!hasRecord || sigFiles < 3) {
+                long minSigs = addressBookRegistry
+                                        .getCurrentAddressBook()
+                                        .nodeAddress()
+                                        .size()
+                                / 3
+                        + 1;
```
jsync-swirlds, line 1336: IF we can raise this to `1/2`+1, that would be good.  `1/3` can still result in a network partition, while > `1/2` ensures we only have one possible valid result.

AlfredoG87 (dismissed):
LGTM

jsync-swirlds (approved):

Nana-EC (approved):
LG

PR #2674, opened 2026-05-01
fix(publisher): close MessagingForwarderTask gap-skip after stall detection

### LINKED ISSUE #2672: bug(publisher): Stalled connection for a block is causing gaps in BN.
[logs snippet.log](https://github.com/user-attachments/files/27282953/logs.snippet.log)

### Description

What Happened — Block 360/364 Analysis                                                                                                                                                                
                                                                                                                                                                                                        
  The log spans 03:49:20 → 03:49:34 UTC (14 seconds).                                                                                                                                                   
                                                                                                                                                                                                        
  Timeline                                                                                                                                                                                            
                                                                                                                                                                                                        
  03:49:20.417 — Block 359 persisted. nextUnstreamedBlockNumber = 360.                                                                                                                                  
   
  03:49:20.506 — Handler 2 (N6-STR1) ACCEPTs block 360. Handlers 0 (N3) and 1 (N0) immediately SKIP it. Block 360 starts streaming slowly — 24+22+22+1+1 items trickle in over 1 second, then complete  
  silence for 7 seconds (the 100MB block stalled the connection).                                                                                                                                     
                                                                                                                                                                                                        
  03:49:22 → 28 — While handler 2 sits stalled, handlers 0, 1, 3 pipeline blocks 361 → 362 → 363 in parallel. These blocks complete fast and are fully queued for dispatch.                             
   
  03:49:28.667 — endOfBlock(363) fires → checkForStalledHandlers(363) triggers:                                                                                                                         
  - 363 > 360 + 2 (MAX_ADVANCE_FOR_STALL_DETECTION = 2) → stall confirmed                                                                                                                             
  - queueByBlockMap.remove(360) — partial block 360 data discarded                                                                                                                                      
  - blocksToResend.add(360)                                                                                                                                                                           
  - stalledHandler.endStreamWithCode(TIMEOUT) — throws ClosedChannelException (N6 was already dead)                                                                                                     
  - Returns ActionForBlock(RESEND, 360) to handler 0 (who just completed block 363)                                                                                                                   
                                                                                                                                                                                                        
  03:49:28.671 — Handler 0 sends RESEND(360) to N3. ✓ RESEND mechanism worked.                                                                                                                          
                                                                                                                                                                                                        
  03:49:28.730 — Critical collision: At this exact millisecond, the queued pipeline for blocks 361-362-363 all flush to verification simultaneously. Block 361 is verified and persisted. Block 360's   
  new resend queue is registered (handler 0 won blocksToResend.remove(360) and registered a new queue). The MessagingForwarderTask forwards the first 95 items from the new block 360 queue. Also at    
  this millisecond:                                                                                                                                                                                     
                                                                                                                                                                                                      
  03:49:28.731 — handlePersisted(361) fires → clearObsoleteQueueItems(361) is called, which scans queueByBlockMap.headMap(361, inclusive) and removes block 360's freshly-created resend queue because  
  block 360 has no block proof yet (360 < 361 and incomplete).
                                                                                                                                                                                                        
  03:49:28.955 — Handler 0 continues receiving block 360 body items from N3 but calls getActionForCurrentlyStreaming(360) → queueByBlockMap.containsKey(360) = false (just cleared) → SKIP. The log also
   shows:
  Handler 0 dropping batch because first block item is not BlockHeader                                                                                                                                  
  The resend is effectively dead. N3 is told to stop.                                                                                                                                                 
                                                     
  03:49:28.949-32.734 — Block 364 streams and completes (closeBlock logged at 32.734) but can never be persisted because block 360 still isn't stored. Eventually block 364's buffered queue is also    
  evicted.

### Steps to Reproduce

Working on a TDD Test to reproduce the issue.

### Additional context
         clearObsoleteQueueItems kills the RESEND. When block 361 (already fully buffered from the pipeline) is persisted, the method removes all incomplete queues for blocks ≤ 361 — including block 360's 
  brand new resend queue that was just registered ~1ms earlier. The RESEND mechanism and the pipeline-persistence mechanism race against each other and the RESEND loses.                               
   
  There is even a TODO comment in the stall detection code acknowledging this area is fragile:                                                                                                          
  // TODO: Blocks to resend isn't enough here, need a better check...                                                                                                                                 
  && !blocksToResend.contains(stalledBlock)                                                                                                                                                             
                                                                                                                                                                                                        
  Secondary Issues                                                                                                                                                                                      
                                                                                                                                                                                                        
  1. MAX_ADVANCE_FOR_STALL_DETECTION = 2 is hardcoded — for a 100MB block the system gave ~8 seconds (3 block-times), enough to detect the stall but the fix window is tiny.                            
  2. RESEND queue has no protection from clearObsoleteQueueItems — there's no blocksBeingResent guard in the obsolete-queue sweep.                                                                    
  3. Block 364 is collateral damage — it completed successfully but can't persist because 360 is permanently missing.                                                                                   
                                                                                                                                                                                                        
  Fix Direction                                                                                                                                                                                         
                                                                                                                                                                                                        
  clearObsoleteQueueItems should skip any block that is actively being resent (i.e. present in activeStreamHandlerByBlock or a new "resend-in-progress" tracking set). Blocks that have been freshly    
  accepted after a RESEND should be immune from the obsolete sweep until they either complete or are explicitly abandoned.

AlfredoG87:
Fixes #2672

## Summary

Closes the block gap introduced when stall detection fires and the
`MessagingForwarderTask` skips the stalled block by jumping to `firstEntry()`.

### Root cause

PR #2675 introduced `activeResendBlocks` to track blocks being re-delivered
after a stall, but never called `.add()` on it — the set was always empty.
This left three related gaps:

1. `isResendingLive()` always returned false, making the set useless.
2. `clearObsoleteQueueItems` could evict a re-queued block before its proof arrived.
3. `determineCurrentBlockNumber` fell through to `firstEntry()` after the stall removed the
   block from `queueByBlockMap`, skipping over it entirely rather than waiting for
   re-delivery. ACKs for later blocks were emitted before the stalled block, and
   `handlePersisted` then suppressed ACK(N) because `N < lastPersisted`.

### Changes

**`LiveStreamPublisherManager`**

- `checkForStalledHandlers`: add `activeResendBlocks.add(candidateBlock)` alongside
  the existing `blocksToResend.add()` — makes `isResendingLive()` functional.
- `clearObsoleteQueueItems`: skip entries present in `activeResendBlocks` to protect
  a re-queued block from eviction during the re-delivery window.
- `determineCurrentBlockNumber` (`MessagingForwarderTask`): before falling through to
  `firstEntry()`, check `blocksToResend` and `activeResendBlocks`; if the last-forwarded
  block is in either set, hold position and wait for re-delivery instead of gap-skipping.
  The two sets cover the full re-delivery window:
  - `blocksToResend`: stall fired → publisher accepts the header
  - `activeResendBlocks`: header accepted → `endOfBlock` completes

**`BlockNodeApiRegressionTest`**

- Adds a 3-test regression suite covering:
  1. Publisher severs mid-block with RESET — gap filled by RESEND, all three blocks ACKed in order.
  2. Stall RESEND — stalled block is ACKed and persisted (primary bug reproducer).
  3. Stall RESEND ordering — no later block is ACKed before the stalled block.
- Updates test 1 to expect correct sequential ACK order (3 → 4 → 5) after the fix.
- Disables the original bug-repro test with `@Disabled` rather than deleting it; the
  assertions document the violation and will fail if the fix is ever reverted.

## Test plan

- [x] `./gradlew :suites:runAPISuites` — all 3 regression tests pass, bug-repro skipped
- [x] Verify `[BUG]` test is marked SKIPPED (not FAILED) in the test report
- [x] Confirm no regressions in existing `BlockNodeAPITests`

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -431,7 +431,9 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             final NavigableSet<Long> keysBeforeBlock = new TreeSet<>();
             keysBeforeBlock.addAll(queueByBlockMap.headMap(blockNumber, true).keySet());
             for (final Long candidate : keysBeforeBlock) {
-                if (!(blockProofs.containsKey(candidate) || hasBlockProof(queueByBlockMap.get(candidate)))) {
+                if (!(blockProofs.containsKey(candidate)
+                        || hasBlockProof(queueByBlockMap.get(candidate))
+                        || activeResendBlocks.contains(candidate))) {
                     queueByBlockMap.remove(candidate);
                     // possibly remove a just collected block proof
                     blockProofs.remove(candidate);
@@ -533,6 +535,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
                 queueByBlockMap.remove(candidateBlock);
                 blockProofs.remove(candidateBlock);
                 blocksToResend.add(candidateBlock);
+                activeResendBlocks.add(candidateBlock);
                 if (stalledHandler != null) {
                     stalledHandler.endStreamWithCode(TIMEOUT, true);
                 }
@@ -947,14 +950,22 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
                 // continue.
                 result = publisherLastForwardedBlockNumber;
             } else {
-                // Else, we either continue with the next one in line or proceed to
-                // supply a block that was resent.
-                final Entry<Long, Deque<BlockItemSetUnparsed>> firstEntry =
-                        publisherManager.queueByBlockMap.firstEntry();
-                if (firstEntry != null) {
-                    result = firstEntry.getKey();
+                // If the block the forwarder was on was stalled and is awaiting
+                // re-delivery, hold position rather than gap-skipping to a later
+                // block. blocksToResend covers the window before a publisher
+                // accepts the resend; activeResendBlocks covers the window from
+                // acceptance until endOfBlock completes.
+                if (publisherManager.blocksToResend.contains(publisherLastForwardedBlockNumber)
+                        || publisherManager.activeResendBlocks.contains(publisherLastForwardedBlockNumber)) {
+                    result = publisherLastForwardedBlockNumber;
                 } else {
-                    result = UNKNOWN_BLOCK_NUMBER;
+                    final Entry<Long, Deque<BlockItemSetUnparsed>> firstEntry =
+                            publisherManager.queueByBlockMap.firstEntry();
+                    if (firstEntry != null) {
+                        result = firstEntry.getKey();
+                    } else {
+                        result = UNKNOWN_BLOCK_NUMBER;
+                    }
                 }
             }
             // Always set the latest streaming block number to the result here
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -240,12 +240,18 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     public void blockIsEnding(final long blockNumber, final long handlerId) {
         // @todo(2344) we can further improve this
         LOGGER.log(INFO, "Handler {0} is ending mid-block {1}", handlerId, blockNumber);
-        final Deque<BlockItemSetUnparsed> deque = queueByBlockMap.remove(blockNumber);
         // Remove from the active resend set, if it's present
         activeResendBlocks.remove(blockNumber);
         // Remove this block from the ACCEPT winner tracking in all cases;
         // if the handler dropped mid-block there is no stall to detect here.
         activeStreamHandlerByBlock.remove(blockNumber);
+        // If stall detection already scheduled this block for resend its sentinel
+        // queue entry must remain in queueByBlockMap so the forwarder holds its
+        // position instead of gap-skipping to a later block.
+        if (blocksToResend.contains(blockNumber)) {
+            return;
+        }
+        final Deque<BlockItemSetUnparsed> deque = queueByBlockMap.remove(blockNumber);
         if (deque != null) {
             if (blockNumber > lastPersistedBlockNumber.get() && blockNumber < nextUnstreamedBlockNumber.get()) {
                 final String message = "Block {0} will be resent due to handler {1} ending mid block";
@@ -411,9 +417,12 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     /// - The last [BlockItemSetUnparsed] batch in the queue does **not** end
     ///   with a `BlockProof` item, which would indicate the proof arrived but
     ///   the block is not yet completed (this should be quite rare).
-    /// - The block number is **not** present in current blocks being resent.
-    ///   Blocks being actively resent would otherwise always be removed,
-    ///   resulting in broken blocks and handler failures.
+    /// - The block number is **not** present in either active or scheduled resend
+    ///   sets. Entries in either set are protected: `activeResendBlocks` covers
+    ///   the window while a publisher is actively streaming the resend, and
+    ///   `blocksToResend` covers the earlier window while the resend is scheduled
+    ///   but not yet accepted. Removing either would discard data or a sentinel
+    ///   queue entry that the forwarder depends on to hold its position.
     ///
     /// Entries that fail either check are left in the map so that in-flight
     /// complete blocks are not lost before they can be forwarded to the
@@ -433,7 +442,8 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             for (final Long candidate : keysBeforeBlock) {
                 if (!(blockProofs.containsKey(candidate)
                         || hasBlockProof(queueByBlockMap.get(candidate))
-                        || activeResendBlocks.contains(candidate))) {
+                        || activeResendBlocks.contains(candidate)
+                        || blocksToResend.contains(candidate))) {
                     queueByBlockMap.remove(candidate);
                     // possibly remove a just collected block proof
                     blockProofs.remove(candidate);
@@ -499,7 +509,9 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     /// signature gathering.
     ///
     /// For each stalled block found:
-    /// - Its queue and proof entries are removed so the forwarder can advance.
+    /// - Its queue is cleared (key retained as a sentinel) so the forwarder holds
+    ///   position rather than gap-skipping to a later block.
+    /// - Its proof entry is removed.
     /// - Its block number is added to blocksToResend.
     /// - The owning handler receives EndStream(TIMEOUT) and is shut down.
     /// - The stall-timeout metric is incremented.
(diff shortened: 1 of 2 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -533,6 +535,7 @@ private void checkForStalledHandlers(final long completedBlockNumber) {
                 queueByBlockMap.remove(candidateBlock);
                 blockProofs.remove(candidateBlock);
                 blocksToResend.add(candidateBlock);
+                activeResendBlocks.add(candidateBlock);
```
jsync-swirlds: We must not do this here, it's not active yet, it's just scheduled!
This add has to happen when the header is accepted and the value is _removed_ from blocksToResend (basically no block should be in both at the same time)

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -947,14 +950,22 @@ private long determineCurrentBlockNumber() {
...
+                    result = publisherLastForwardedBlockNumber;
                 } else {
-                    result = UNKNOWN_BLOCK_NUMBER;
+                    final Entry<Long, Deque<BlockItemSetUnparsed>> firstEntry =
+                            publisherManager.queueByBlockMap.firstEntry();
+                    if (firstEntry != null) {
+                        result = firstEntry.getKey();
+                    } else {
+                        result = UNKNOWN_BLOCK_NUMBER;
+                    }
```
jsync-swirlds: This will make stalls permanent, because blocks to resent is _scheduled_, not active and may never get sent, and putting this here will cause the handlers to start sending behind publisher messages, ending connections and breaking the network.

We are being deliberate about not updating the current block number for resends; resend is completely outside that path.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeApiRegressionTest.java
```diff
@@ -233,6 +238,361 @@ void missingBlockGapWhenPublisherSeversConnectionBeforeEndOfBlock() throws Inter
...
+        assertThat(ack2Position)
+                .as("publisher B must receive ACK for block 2 — stalled block must be acknowledged")
+                .isGreaterThanOrEqualTo(0);
+        assertThat(ackBlockOrder.subList(0, ack2Position))
+                .as("no block with number greater than 2 may be acknowledged before the stalled block 2")
+                .allSatisfy(blockNum -> assertThat(blockNum).isLessThanOrEqualTo(2L));
+
+        publisherAClient.close();
+        publisherBClient.close();
+    }
```
jsync-swirlds, line 488: This cannot be guaranteed in the publisher.
Scenario 1:
block 1 <- handler 1 sends, and stalls
block 2 <- handler 2 sends, completes
block 3 <- handler 2 sends, completes
block 4 <- handler 2 sends, completes
-- Stall detected, block 1 scheduled for resend, handler 1 disconnected
block 5 <- handler 2 sends, completes
-- handler 2 gets resend
block 1 <- handler 2 sends, completes

In this case, we can expect block 1 to be the next block sent, before 2, 3, or later.

Scenario 2:
block 7 <- handler 2 sends, completes
block 8 <- handler 1 sends, and stalls
block 9 <- handler 2 sends, completes
block 10 <- handler 2 sends, completes
-- Persistence fails for block 7, scheduled for resend
block 11 <- handler 2 sends, completes
-- Stall detected, block 8 scheduled for resend, handler 1 disconnected
block 12 <- handler 2 sends, completes
-- handler 2 gets resend for _block 7_
block 7 <- handler 3 sends, completes

The latter requires more flexibility, and I believe the code, as written, would fail in that case.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -431,7 +440,10 @@ private void clearObsoleteQueueItems(final long blockNumber) {
             final NavigableSet<Long> keysBeforeBlock = new TreeSet<>();
             keysBeforeBlock.addAll(queueByBlockMap.headMap(blockNumber, true).keySet());
             for (final Long candidate : keysBeforeBlock) {
-                if (!(blockProofs.containsKey(candidate) || hasBlockProof(queueByBlockMap.get(candidate)))) {
+                if (!(blockProofs.containsKey(candidate)
+                        || hasBlockProof(queueByBlockMap.get(candidate))
+                        || activeResendBlocks.contains(candidate)
+                        || blocksToResend.contains(candidate))) {
```
jsync-swirlds, line 446: There is a method for detecting active resend.  We should probably use that here, rather than the paired resend set checks.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
AlfredoG87: Yes, I was concerned about this too. 

Thanks for clarification on intent.

AlfredoG87: ✅  I've updated the PR with a more robust approach if you can take another look please

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
AlfredoG87: ✅  I've updated the PR with a more robust approach if you can take another look please

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -530,7 +544,10 @@ private void checkForStalledHandlers(final long completedBlockNumber) {
                         handlerId,
                         candidateBlock,
                         completedBlockNumber);
-                queueByBlockMap.remove(candidateBlock);
+                final Deque<BlockItemSetUnparsed> stalledQueue = queueByBlockMap.get(candidateBlock);
+                if (stalledQueue != null) {
+                    stalledQueue.clear();
+                }
```
jsync-swirlds, line 550: Why are we clearing a queue instead of removing it?
That breaks more things, and we have no active sender, so this will leave entries in the block map more-or-less forever.

We have no guarantee that any handler will ever successfully resend the block.  This isn't a minor concern, we're putting a block here that _may never resolve_.

That's why prior work deliberately allowed the other pending blocks to flow through.  The gap can be filled elsewhere, but we cannot guarantee we'll ever get the next block, so we handle "block out of order" in the _acknowledgement_ flow, rather than trying to prevent the gap here.

AlfredoG87: `determineCurrentBlockNumber()` uses `queueByBlockMap.containsKey(lastForwardedBlockNumber)` — not `isEmpty()` — to decide whether to hold position. With `remove()` the key is gone, the forwarder falls through to `firstEntry()` = block 3, and the gap-skip happens. With `clear()` the key stays, `containsKey` returns true, and the forwarder waits at block 2. The cleared deque only discards the stalled handler's partial (and now useless) data for that one block — other blocks' deques are not touched.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeApiRegressionTest.java
AlfredoG87: are you suggestion to make 2 more scnearios / regression tests?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds:
```suggestion
                if (!(blockProofs.containsKey(candidate)
                        || hasBlockProof(queueByBlockMap.get(candidate))
                        || isResendingLive(candidate))) {
```

AlfredoG87:
closing in favor of #2699  and #2705

PR #2679, opened 2026-05-02
Feature/issue 2660 rsa roster script

### LINKED ISSUE #2660: Implement `generate-rsa-roster-bootstrap.sh` operator script
### Story Form

As a block node operator
I want to have a script that can create a bootstrap address book roster
So that my BN doesn't have to query MN at deployment time and has a file ready to load

### Technical Notes

The bootstrap plugin has a fall back to query the MN API to get network nodes to be able to verify WRBs, however, preferably at deploy time we'd want node operators to already have a file.
Thus is would be good to have a secure and easy to run bash script that does the logic to build the bootstrap file in advance so the BN doesn't have to do it.

Additionally, this should also capture the runbook for this deployment

imshubham22apr-gif:
This PR introduces a standalone operator bash script (`generate-rsa-roster-bootstrap.sh`) to pre-fetch and build the RSA bootstrap address book roster prior to Block Node deployment.

Key implementations include:

1. Mirror Node Integration:Queries the `/api/v1/network/nodes` endpoint to dynamically build the roster.
2. Pagination Handling:Implements a `while` loop to follow `links.next` pagination until all nodes are retrieved, ensuring no nodes are missed on larger networks like Mainnet.
3. JSON Parsing: Utilizes `jq` to safely extract `node_id`, `node_account_id`, and `rsa_public_key`, formatting them into the expected `bootstrap-roster.json` structure.
4. Documentation: Adds a comprehensive runbook (`docs/runbooks/rsa-roster-bootstrap.md`) detailing prerequisites and execution steps for both Mainnet and Testnet environments.

Related Issue(s)
Closes #2660

### DIFF
--- a/docs/runbooks/rsa-roster-bootstrap.md
+++ b/docs/runbooks/rsa-roster-bootstrap.md
@@ -1,0 +1,27 @@
+# Runbook: Generating RSA Roster Bootstrap File
+
+## Overview
+As per Issue #2660, Block Nodes (BN) need an address book roster ready at deployment time so they don't have to query the Mirror Node (MN) API dynamically to verify Wrapped Record Blocks (WRBs). 
+
+This runbook explains how to use the `generate-rsa-roster-bootstrap.sh` script to pre-fetch this data.
+
+## Prerequisites
+- `curl` installed
+- `jq` installed (Command-line JSON processor)
+
+## Execution Steps
+
+1. Navigate to the root of the repository.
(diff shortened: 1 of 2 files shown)


AlfredoG87:
@imshubham22apr-gif commit needs to be signed.
like this:
image

imshubham22apr-gif:
@AlfredoG87 PTAL...

AlfredoG87:
This is no longer needed, there is an alternative script that does the same.

Issue #2682, opened 2026-05-02
Query peer BN for RSA roster bootstrap data


Nana-EC:
### Story Form
As a bootstrapping block node,
I want to update the `RsaRosterBootstrapPlugin` to query a peer BN's detailed status via gRPC and extract the consensus node RSA roster (NodeAddressBook),
So that a new or restarting node can learn the current RSA address book from the network without immediately falling back to the Mirror Node.

### Background
The current startup sequence in `RsaRosterBootstrapPlugin` is:
1. Check for a local `rsa-bootstrap-roster.pb` file (loaded by `BlockNodeApp` before plugins start).
2. If absent, query the Mirror Node REST API (`GET /api/v1/network/nodes`) with pagination and retries.

This mirrors the pre-peer-query state of `TSSBootstrapPlugin` before issue #2296. Adding a peer BN query step between the file check and the Mirror Node call reduces external dependency on Mirror Node availability at startup time and aligns the two bootstrap plugins.

### Desired Startup Sequence (post this change)
1. **File** — `BlockNodeApp` loads `rsa-bootstrap-roster.pb` if present; plugin detects `context.nodeAddressBook() != null` and returns.
2. **Peer BNs** — If no file, query one or more configured peer BNs via gRPC `serverStatusDetail` and extract the `NodeAddressBook` from the response.
3. **Mirror Node** — If no peer returns a usable roster, fall back to querying the Mirror Node REST API as today.

### Technical Notes
- The `NodeAddressBook` is already part of `BlockNodeContext` and can be surfaced in `serverStatusDetail` by the `ServerStatusServicePlugin`; each BN that has bootstrapped its roster exposes it for peers to consume.
- This is the same propagation model used by `TSSBootstrapPlugin` (see #2296): each BN bootstraps from a peer, then exposes the data for others.
- Follow the same gRPC `WebClient` pattern used elsewhere in the codebase.
- Peer addresses should be configurable via `BootstrapRosterConfig` (new field, analogous to the TSS peer config).

### Additional Items
- Use gRPC `WebClient` to call `serverStatusDetail` on configured peer BN(s).
- Retry with backoff on transient failures (configurable max retries and intervals).
- Try multiple configured peers if the first is unavailable.
- Validate the extracted `NodeAddressBook` (≥1 entry with a non-blank RSA public key) before accepting it.
- Only fall through to Mirror Node if all configured peers fail or return no usable roster.
[...]

PR #2683, opened 2026-05-02
feat: Address Book Loading support in RSA Roster Bootstrap Plugin

### LINKED ISSUE #2561: Add new plugin to acquire/build/load the network roster config
### Story Form

As a Block Node operator,

---

### Background

---

### Plugin structure

### Key implementation areas

### What this ticket is NOT

---

## Acceptance Criteria

---

## References

| Resource | Link |
|---|---|

Nana-EC:
## Summary
Implements the `RsaRosterBootstrapPlugin` for Phase 2a of the Hiero network upgrade. This plugin loads the
consensus node RSA address book (`node_id → RSA public key`) at Block Node startup so the WRB proof
verification pipeline has what it needs before the first Wrapped Record Block arrives.

Additionally:
- Extends `BlockNodeContext` and `ApplicationStateFacility` SPI with `nodeAddressBook` support
- Moves RSA file loading and persistence into `BlockNodeApp` (consistent with how TSS state is managed)
- Fixes a proto typo (`node_addess_book` → `node_address_book`) and exposes the address book in the `serverStatusDetail` gRPC endpoint
- Adopts Java 23 `///` markdown Javadoc style for all new code in the module

---

### New module — `block-node/roster-bootstrap-rsa`

**`RsaRosterBootstrapPlugin`**
- Implements `BlockNodePlugin`; registered via `ServiceLoader` in `module-info.java`
- `start()` checks `context.nodeAddressBook()` first — if non-null, `BlockNodeApp` already loaded the file; records metrics and returns
- If null, fetches from the Mirror Node REST API (`GET /api/v1/network/nodes`, paginated via `links.next`), strips optional `0x` prefix from keys, skips entries with blank `public_key`
- HTTP retries: up to 3 attempts with exponential backoff (1 s → 2 s → 4 s)
- Throws `IllegalStateException` (fail-fast) if Mirror Node is unreachable after all retries or returns zero usable entries
- Overrides `onContextUpdate()` to keep its cached `this.context` reference in sync after `BlockNodeApp` loads the file during `startApplicationStateFacility()` (runs before `startPlugins()`)
- Metrics: `blocknode:roster_entries_loaded` and `blocknode:roster_load_duration_ms` (both `ObservableGauge`)

**`BootstrapRosterConfig`**
- `@ConfigData("roster.bootstrap")` record with 5 properties:
  - `filePath` (default `data/config/rsa-bootstrap-roster.pb`)
  - `mirrorNodeBaseUrl` (default testnet mirror node)
  - `mirrorNodeConnectTimeoutSeconds` (default 5)
  - `mirrorNodeReadTimeoutSeconds` (default 10)
  - `mirrorNodePageSize` (default 100)

**`MirrorNodeNodesResponse`**
- Hand-rolled JSON parser to avoid adding a JSON library dependency to the module
- Extracts `node_id`, `public_key`, and `links.next` from Mirror Node responses
- Handles `null` public keys, `0x`-prefixed hex strings, empty node arrays, and multi-page pagination

### SPI extensions — `block-node/spi-plugins`

**`BlockNodeContext`**
- Added `nodeAddressBook` as the 11th component (`NodeAddressBook`, nullable)
- `context.nodeAddressBook()` returns null until the address book is loaded

**`ApplicationStateFacility`**
- Added `updateAddressBook(NodeAddressBook book)` method — persists the file atomically and fans `onContextUpdate` to all plugins

### App changes — `block-node/app`

**`BlockNodeApp`**
- `loadApplicationState()` (runs in `startApplicationStateFacility()`, before `startPlugins()`): loads `rsa-bootstrap-roster.pb` if present; on success rebuilds `blockNodeContext` with the book and calls `onContextUpdate` on all loaded plugins so their cached context is current before `start()` is called
- `updateAddressBook()` override: atomic write-to-tmp-then-rename of the binary protobuf file, then fans out `onContextUpdate`
- `validateAddressBook()`: static helper — requires ≥1 entry with a non-blank `rsaPubKey`
- `NodeConfig`: added `rsaBootstrapFilePath` (used by `BlockNodeApp` to locate the file without a hard dependency on the plugin module)

### Proto + endpoint changes

**`protobuf-sources/src/main/proto/block-node/api/node_service.proto`**
- Fixed typo: `node_addess_book = 4` → `node_address_book = 4` in `ServerStatusDetailResponse`
- PBJ sources regenerated — generated Java now uses `nodeAddressBook` throughout

**`ServerStatusServicePlugin.serverStatusDetail()`**
- Added `.nodeAddressBook(context.nodeAddressBook())` to the response builder
- When null (address book not yet loaded), PBJ omits the field from the wire response

### Test infrastructure — `block-node/app` (testFixtures)

**`PluginTestBase`**
- Added 6-arg `start()` overload with `@Nullable NodeAddressBook preloadedAddressBook`
- All existing 5-arg overloads chain to it with `null`
- Simulates `BlockNodeApp` pre-loading the file: builds the initial `BlockNodeContext` with the book already set

### Documentation

- All Javadocs in `block-node/roster-bootstrap-rsa/src` converted from `/** */` HTML style to Java 23 `///` markdown style

---

## Test coverage

|---|---|
| `RsaRosterBootstrapPluginTest` | Pre-loaded book reflected in context; metrics recorded; `onContextUpdate` delivered; unreachable Mirror Node throws; `configDataTypes` registration |
| `MirrorNodeNodesResponseTest` | Single node with `0x` prefix; multi-node with next-link; null `public_key`; empty nodes array |
| `ServerStatusDetailServicePluginTest` | `nodeAddressBook` absent when null; absent after context update with null; present with correct entries after context update with non-null; wire-level `hasNodeAddressBook()` checks |

---

## Startup sequence (as implemented)

```
BlockNodeApp.start()
  └─ startApplicationStateFacility()
       └─ loadApplicationState()
            ├─ Load TssData from JSON file (if present)
            └─ Load rsa-bootstrap-roster.pb (if present)
                 ├─ Parse NodeAddressBook.PROTOBUF
                 ├─ validateAddressBook (≥1 non-blank RSA key)
                 ├─ rebuild blockNodeContext with book
                 └─ call onContextUpdate on all loaded plugins
  └─ startPlugins()
       └─ RsaRosterBootstrapPlugin.start()
            ├─ context.nodeAddressBook() != null?
            │    YES → record metrics, return (file was loaded above)
            │    NO  → fetch Mirror Node (paginated, with retries)
            │           ├─ success → applicationStateFacility.updateAddressBook(book)
            │           │              └─ persists file + onContextUpdate all plugins
            │           └─ failure → throw IllegalStateException (fail-fast)
```

---

## Reviewer Notes

- The plugin itself has no file I/O — all load and persist is in `BlockNodeApp`, consistent with how `TssData` is managed. The plugin's `start()` only decides which path to take based on `context.nodeAddressBook()`.
- The Mirror Node JSON parser is intentionally minimal and hand-rolled to avoid adding a dependency. It handles the stable subset of the Mirror Node API response needed for bootstrapping.
- Peer BN querying (file → peer BNs → Mirror Node) is tracked separately in #2682.

## Related Issue(s)
Closes #2561


block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -376,6 +380,34 @@ public void updateTssData(TssData tssData) {
...
+                blockNodeContext.serviceLoader(),
+                blockNodeContext.threadPoolManager(),
+                blockNodeContext.blockNodeVersions(),
+                blockNodeContext.tssData(),
+                nodeAddressBook);
+        loadedPlugins.parallelStream().forEach(plugin -> plugin.onContextUpdate(blockNodeContext));
+        LOGGER.log(
+                INFO,
+                "ApplicationStateFacility called plugin.onContextUpdate for all plugins (nodeAddressBook update)");
+        persistNodeAddressBook(nodeAddressBook);
```
ata-nas: Quick note: I think this update should be in another place. There are multiple components that can issue any type of updates. These updates should be ordered and we must be sure that no update will discard another, but rather all updates are properly handled. There is a scheduled task that runs and checks for latest updates. Maybe we can utilize that.

Nana-EC: Yeah you're correct, will remove and update a local cache

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -24,6 +27,7 @@
 @ConfigData("app.state")
 public record ApplicationStateConfig(
        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/app-state-data.bin") Path dataFilePath,
+       @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/rsa-bootstrap-roster.pb") Path rsaBootstrapFilePath,
         @Loggable @ConfigProperty(defaultValue = "500") @Min(100) long updateScanInterval,
         @Loggable @ConfigProperty(defaultValue = "100") @Min(100) int updateInitialDelay) {}
```
ata-nas, line 32: Nit: spotless is off here, we could align all rows to start from the same indentation.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -378,6 +385,18 @@ public void updateTssData(TssData tssData) {
...
+    /**
+     * Allow plugins to update the NodeAddressBook for this BlockNodeApp. The book is queued for
+     * processing on the next {@code checkForApplicationStateUpdates} scan tick, which rebuilds the
+     * context, notifies all plugins via {@code onContextUpdate}, and persists the book to disk.
+     *
+     * @param nodeAddressBook the NodeAddressBook to store in BlockNodeContext
+     */
+    @Override
+    public void updateAddressBook(NodeAddressBook nodeAddressBook) {
+        if (nodeAddressBook != null) pendingAddressBook.set(nodeAddressBook);
```
ata-nas, line 397: Why would the atomic reference be null? It is final and we instantiate it every time.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -495,16 +545,38 @@ private void persistTssData(TssData tssData) {
         }
     }
 
+    private void persistNodeAddressBook(NodeAddressBook nodeAddressBook) {
+        final Path filePath = blockNodeContext
+                .configuration()
+                .getConfigData(ApplicationStateConfig.class)
+                .rsaBootstrapFilePath();
+        try {
+            Files.createDirectories(filePath.getParent());
```
ata-nas, line 554: I feel like the directory creation should simply happen once on startup of the app. Seems redundant here.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -495,16 +545,38 @@ private void persistTssData(TssData tssData) {
...
+            final Path tmp = filePath.resolveSibling(filePath.getFileName() + ".tmp");
+            final Bytes encoded = NodeAddressBook.PROTOBUF.toBytes(nodeAddressBook);
+            Files.write(tmp, encoded.toByteArray());
+            Files.move(tmp, filePath, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE);
```
ata-nas, line 558: Note, some FSs do not support atomic move. As a resiliency item, we could catch an atomic not supported exception (forgot the formal name) and try again.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/BootstrapRosterConfig.java
```diff
@@ -0,0 +1,24 @@
...
+/// @param mirrorNodeReadTimeoutSeconds per-request read timeout when calling the Mirror Node
+/// @param mirrorNodePageSize number of nodes requested per paginated Mirror Node call
+@ConfigData("roster.bootstrap")
+public record BootstrapRosterConfig(
+        @Loggable @ConfigProperty(defaultValue = "https://testnet-public.mirrornode.hedera.com")
+        String mirrorNodeBaseUrl,
+
+        @Loggable @ConfigProperty(defaultValue = "5") int mirrorNodeConnectTimeoutSeconds,
+        @Loggable @ConfigProperty(defaultValue = "10") int mirrorNodeReadTimeoutSeconds,
+        @Loggable @ConfigProperty(defaultValue = "100") int mirrorNodePageSize) {}
```
ata-nas, line 24: We should disable spotless for this config file and have each line be a different property.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/MirrorNodeNodesResponse.java
```diff
@@ -0,0 +1,96 @@
...
+    ///     { "node_id": 0, "public_key": "0x..." },
+    ///     ...
+    ///   ],
+    ///   "links": { "next": "/api/v1/network/nodes?..." }
+    /// }
+    /// ```
+    ///
+    /// @param json the raw JSON string from the Mirror Node API
+    /// @return parsed response containing node entries and the next-page link
+    static MirrorNodeNodesResponse parse(final String json) {
```
ata-nas, line 64: I feel like we might want a try-catch here. Are there quite a lot of possibilites to throw within this method? Otherwise the caller should expect a possible exception. Just something to think about.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,225 @@
...
-        // todo: if roster is empty and failOnFetchError=true → throw to abort BN startup
-        // todo: emit roster_entries_loaded gauge metric (labelled by source: file | mirror_node)
+        final long startMs = System.currentTimeMillis();
+        NodeAddressBook book = context.nodeAddressBook();
+        final String source;
+
+        if (book != null) {
+            // Bootstrap file was loaded by BlockNodeApp.loadApplicationState() which runs in
+            // startApplicationStateFacility() before startPlugins() and calls onContextUpdate().
+            source = "file";
```
ata-nas, line 109: This source could be an enum?

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,225 @@
...
+    // Mirror Node fetch
+    // -------------------------------------------------------------------------
+
+    /// Fetches the node address book from the Mirror Node REST API with pagination and retries.
+    ///
+    /// @return a non-empty `NodeAddressBook`
+    /// @throws IllegalStateException if the Mirror Node is unreachable after retries or returns
+    ///     no usable entries
+    private NodeAddressBook fetchFromMirrorNode() {
+        final HttpClient client = HttpClient.newBuilder()
```
ata-nas, line 138: The client must be closed. We should encapsulate with a try-with-resources.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,225 @@
...
+                        .GET()
+                        .build();
+                final HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+                if (response.statusCode() != 200) {
+                    throw new IOException("HTTP " + response.statusCode() + " from " + url);
+                }
+                return response.body();
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+                lastCause = ie;
```
ata-nas, line 212: I feel like the `InterruptedException` should probably be allowed to propagate to the caller. The caller should most likely be expecting this. If we are interrupted, this means that someone is shutting us down and we must simply stop our execution. I think in a timeout or similar, the exception would be different.

block-node/roster-bootstrap-rsa/src/main/java/module-info.java
```diff
@@ -1,22 +1,31 @@
...
+    exports org.hiero.block.node.roster.bootstrap.rsa to
+            com.swirlds.config.impl,
+            com.swirlds.config.extensions,
+            org.hiero.block.node.app;
+
     requires transitive org.hiero.block.node.spi;
+    requires transitive org.hiero.block.protobuf.pbj;
+    requires com.hedera.pbj.runtime;
+    requires org.hiero.block.node.base;
+    requires com.google.gson;
```
ata-nas, line 27: Are we (must we) use GSON?

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
```diff
@@ -0,0 +1,157 @@
...
+                new BlockingExecutor(new LinkedBlockingQueue<>()),
+                new ScheduledBlockingExecutor(new LinkedBlockingQueue<>()));
+    }
+
+    /// Override tearDown to gracefully handle cases where `start()` threw before
+    /// the base class could initialise `metricsRegistry` (e.g. exception tests).
+    @Override
+    @AfterEach
+    public void tearDown() throws IOException {
+        testThreadPoolManager.shutdownNow();
```
ata-nas, line 47: this should automatically happen in the plugin base. There should be an after each there that shuts down the thread pool maanger. At least that is what I remember.

block-node/roster-bootstrap-rsa/build.gradle.kts
```diff
@@ -10,6 +10,8 @@ tasks.withType<JavaCompile>().configureEach { options.compilerArgs.add("-Xlint:-
 mainModuleInfo { runtimeOnly("com.swirlds.config.impl") }
 
 testModuleInfo {
-    runtimeOnly("org.junit.jupiter.api")
-    runtimeOnly("org.hiero.block.node.app.test.fixtures")
+    requires("org.junit.jupiter.api")
+    requires("org.junit.jupiter.params")
+    requires("org.mockito")
```
ata-nas, line 15: Do we use mockito?

ata-nas (commented):
First pass here, some suggestions and questions.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -24,6 +27,7 @@
 @ConfigData("app.state")
 public record ApplicationStateConfig(
        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/app-state-data.bin") Path dataFilePath,
+       @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/rsa-bootstrap-roster.pb") Path rsaBootstrapFilePath,
```
berryware, line 30: For TSS this is a json file so that the values can be overwritten by testers or hoster

berryware (commented):
Do we need to allow editing of the rsa-bootstrap-raoster file like we do with tss. If so then it should be json.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
Nana-EC: I see, this is a claim for easy nod operator experience. As a protobuf file it's hard to manage, if in json it would be more friendly. Make sense

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: Defensive play as the null for AtomicReference with an empty constructor is null

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: By doing it here we make sure it's only created for a BN that configures.
If done outside we may create an empty dir for no reason.
Also this needs to happen before startup and after init() as part of `startApplicationStateFacility()`

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: Good call, added in PR 2718

block-node/roster-bootstrap-rsa/src/main/java/module-info.java
Nana-EC: Do you have an alternative suggestion?
I tried to avoid creating our own JSON parsing logic

ata-nas: I feel like we could be using our JSON parser? Maybe I am mistaken.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: My bad here, I've misread this.

Nana-EC:
Closing, logic was split across https://github.com/hiero-ledger/hiero-block-node/pull/2718 and https://github.com/hiero-ledger/hiero-block-node/pull/2719.
@ata-nas  and @berryware feedback above was address there also

Issue #2651, opened 2026-04-29
bug(backfill): We are seeing some invalid (corrupted) blocks on some production networks and timeouts


AlfredoG87:
### Description

On a previous investigation of a BN in Previewnet trying to backfill, there were some connections that timed-out, and also lots of blocks that were received unabled to be parsed, with different parsing issues.

When manually querying the source and the block ranges, I was able to get the blocks, some times after a single retry but most of the time at first. This manual test was done using Postman (not helidon) webClient.

### Steps to Reproduce

There is still not a reliable way to reproduce this, the current hypothesis is that HTTP2 WebClient of Helidon is not robust when network is choppy or has fluctuations, resulting in half-dead connections (we have seen this with the publisher as a form of stalled connections) and parsing issues.

The first part of this issue is to be able to find the root cause by introducing network fluctuations and chaos. 

One idea is to use Chaos Mesh to introduce some network conditions and see if we can reliably reproduce the error. finally create an isolated test that is able to identify the issue, and if possible to fix it.

### Additional context

These are some of the error messages found:

```
Failed to fetch blocks [151,290->151,309] from node [[lfh02.previewnet.blocknode.hashgraph-devops.com](http://lfh02.previewnet.blocknode.hashgraph-devops.com/)] (attempt 1/3) due to java.lang.RuntimeException: Error fetching blocks - , com.hedera.pbj.runtime.ParseException: java.io.IOException: Cannot understand wire_type of 6.
```

```
Failed to fetch blocks [0->19] from node [[lfh02.previewnet.blocknode.hashgraph-devops.com](http://lfh02.previewnet.blocknode.hashgraph-devops.com/)] (attempt 3/3) due to java.lang.RuntimeException: Error fetching blocks - , java.lang.RuntimeException: Timed out waiting for block stream response.
```

```
warn rfh01 2026-04-17 20:19:18.250+0000 WARNING [io.helidon.webclient.http2.Http2ClientConnection handle] Unsupported frame type!! UNKNOWN
```

```
[...]

PR #2684, opened 2026-05-04
fix(backfill): resolve Helidon timeouts and improve connection resilience

imshubham22apr-gif:
This PR fixes the backfill hangs we've been seeing and cleans up our configuration structure.

I noticed the backfill process was getting stuck on half-dead connections where the socket stays open but no data moves. To fix this, I added a read timeout to the Helidon client so it doesn't wait forever when a peer stops responding. I also updated the backfill fetcher to throw away the old client and start fresh if a connection fails, which makes the whole process much more reliable against network blips.

On the configuration side, I've moved the application state settings—like TSS data paths and scan intervals—out of the main NodeConfig and into a new ApplicationStateConfig class. This separates general node settings from app-specific state logic. Just a heads up that if you use local property files, you'll need to change any block.node.appState keys to use the new appState prefix instead.

This resolves issue 2651. Everything is tested and the config migration is working as expected.

### DIFF
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
@@ -4,7 +4,6 @@ package org.hiero.block.node.app.config.node;
 import com.swirlds.config.api.ConfigData;
 import com.swirlds.config.api.ConfigProperty;
 import com.swirlds.config.api.validation.annotation.Min;
-import java.nio.file.Path;
 import org.hiero.block.node.base.Loggable;
 
 /**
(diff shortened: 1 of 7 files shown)


AlfredoG87:
@imshubham22apr-gif, we need to keep the PRs focused on 1 change at the time only and link it to the Issue that is resolving.

Closing this since is not clear how we are making sure that the referenced issue is actually resolved, and since is trying to mix 2 to 3 separate issues on a single PR.

Thanks for the interest and contributions, if you feel strongly about the configuration needed to fix the issue, I would advise to show how the issue is replicable and then make a test for that, once you have a reproduceable test, you can go ahead and make changes to the configuration, and lets keep the PRs focused on a single issue at a time.

AlfredoG87:
I've added a comment to issue #2651, so we can circle back to your suggestions when picking up that issue.

Again, thank you for your interest and contribution.

PR #2696, opened 2026-05-04
fix: `CombinedBlockRangeSet#contains(long, long)` Checks Individual Internal Sets Without Combining Data

### LINKED ISSUE #2694: `CombinedBlockRangeSet#contains(long, long)` Checks Individual Internal Sets Without Combining Data
### Description

- The `CombinedBlockRangeSet#contains(long, long)` iterates over individual registered sets and calls `#contains(long, long)`, which does not cover the case where two sets - set1(0, 10); set2(11, 20) or overlapping, will not return `true` if we query the combined set for (0, 20).
- We have to be able to give a consolidated information for contiguous ranges available from all registered sets.

### Steps to Reproduce

_No response_

### Additional context

_No response_

ata-nas:
* Updated logic for the combined ranged set to consolidate all internal sets
* Additional tests to cover regressions

## Related Issue(s)

Fixes #2694

### DIFF
--- a/block-node/base/src/main/java/org/hiero/block/node/base/ranges/CombinedBlockRangeSet.java
+++ b/block-node/base/src/main/java/org/hiero/block/node/base/ranges/CombinedBlockRangeSet.java
@@ -3,6 +3,7 @@ package org.hiero.block.node.base.ranges;
 
 import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Arrays;
 import java.util.stream.LongStream;
 import java.util.stream.Stream;
@@ -48,12 +49,7 @@ public class CombinedBlockRangeSet implements BlockRangeSet {
      */
     @Override
     public boolean contains(long start, long end) {
-        for (BlockRangeSet set : blockRangeSets) {
-            if (set.contains(start, end)) {
-                return true;
-            }
-        }
-        return false;
+        return reduceRanges().contains(start, end);
     }
 
     /**
@@ -97,10 +93,7 @@ public class CombinedBlockRangeSet implements BlockRangeSet {
      */
     @Override
     public LongStream stream() {
-        return new ConcurrentLongRangeSet(Arrays.stream(blockRangeSets)
-                        .flatMap(BlockRangeSet::streamRanges)
-                        .toArray(LongRange[]::new))
-                .stream();
+        return reduceRanges().stream();
     }
 
     /**
@@ -108,9 +101,13 @@ public class CombinedBlockRangeSet implements BlockRangeSet {
      */
     @Override
     public Stream<LongRange> streamRanges() {
+        return reduceRanges().streamRanges();
+    }
+
+    @NonNull
+    private ConcurrentLongRangeSet reduceRanges() {
         return new ConcurrentLongRangeSet(Arrays.stream(blockRangeSets)
-                        .flatMap(BlockRangeSet::streamRanges)
-                        .toArray(LongRange[]::new))
-                .streamRanges();
+                .flatMap(BlockRangeSet::streamRanges)
+                .toArray(LongRange[]::new));
     }
 }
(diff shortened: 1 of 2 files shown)


block-node/base/src/main/java/org/hiero/block/node/base/ranges/CombinedBlockRangeSet.java
```diff
@@ -48,12 +49,7 @@ public boolean contains(long blockNumber) {
      */
     @Override
     public boolean contains(long start, long end) {
-        for (BlockRangeSet set : blockRangeSets) {
-            if (set.contains(start, end)) {
-                return true;
-            }
-        }
-        return false;
+        return reduceRanges().contains(start, end);
```
jsync-swirlds, line 52: This is a lot of processing for every contains, which is a high frequency action.
Why don't we reduce and consolidate ranges on add/update instead?
Changing ranges should be much less frequent, and adding anything other than a tail-increment (which does not require reduction) should be even rarer.

block-node/base/src/main/java/org/hiero/block/node/base/ranges/CombinedBlockRangeSet.java
```diff
@@ -3,6 +3,7 @@
 
 import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
```
jsync-swirlds, line 6: Is this used?

ata-nas: Can be removed, seems like an automatic addition to a method I extracted using IntelliJ. Used by the reduce method.

block-node/base/src/main/java/org/hiero/block/node/base/ranges/CombinedBlockRangeSet.java
jsync-swirlds, line 23: I think this is what gets us in trouble here.
If we keep reference to ranges owned elsewhere, then we cannot merge them...
I think we probably need a more thorough rethink/rework where we have callers submit new ranges/updates instead of just updating existing objects.

block-node/base/src/main/java/org/hiero/block/node/base/ranges/CombinedBlockRangeSet.java
ata-nas: I agree that this is somewhat an excessive work for this approach.
This is the easiest (quickest) fix that I could come up with. I think due to the nature of this combined set, it is a bit tricky to consolidate on add/update because this set never gets added to. Things get updated to the individual sets that this one contains. In fact, this set cannot be added to or updated. This is a simple "view" of a collection of registered sets (that get independently updated).

jsync-swirlds (approved):
This is a _costly_ change from a performance perspective.
Hopefully we can redo this with more efficient code in the future, but it works for now.

block-node/base/src/main/java/org/hiero/block/node/base/ranges/CombinedBlockRangeSet.java
jsync-swirlds: We need to change/fix that (it's based on a lot of now-flawed assumptions).
Not necessary for this particular PR, however.

block-node/base/src/main/java/org/hiero/block/node/base/ranges/CombinedBlockRangeSet.java
ata-nas: We can certainly have this as an item to improve upon.

block-node/base/src/main/java/org/hiero/block/node/base/ranges/CombinedBlockRangeSet.java
ata-nas: P.S. the method we fix here specifically is called only on the zipping and soon to be deprecated s3 plugin. I think we can handle the overhead for now, especially because we do not really expect to have too many ranges for now.

rockysingh (approved):
lgtm

PR #2699, opened 2026-05-04
fix: Add logic to reduce incorrect acknowledgements.

jsync-swirlds:
## Reviewer Notes
* Ensure no acknowledgement is _later_ than the earliest active resend.
* Ensure that no acknowledgement is _later_ than any known missing or in-flight block.
* Fix a small oversight in block queue cleanup.
* Removed an obsolete method.

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -67,6 +67,8 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     private static final int DATA_READY_WAIT_MICROSECONDS = 5000;
     private static final String STALL_DETECTED_LOG_MESSAGE =
             "[{0}] Stall detected: handler {1} has not completed block {2}, another handler completed block {3}";
+    private static final String BLOCK_ABANDONED_LOG_MESSAGE =
+            "[{0}] Replacement persistence detected: handler {1} has not completed block {2}, but a different source persisted the same block {3}";
     private final System.Logger LOGGER = System.getLogger(LiveStreamPublisherManager.class.getName());
     private final MetricsHolder metrics;
     private final BlockNodeContext serverContext;
@@ -75,6 +77,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     private final ConcurrentNavigableMap<Long, PublisherHandler> handlers;
     private final AtomicLong nextHandlerId;
     private final ConcurrentNavigableMap<Long, Deque<BlockItemSetUnparsed>> queueByBlockMap;
+    // @todo Replace this with a metric, once metric values can be read.
     private final AtomicLong blocksClosedComplete;
     private final Condition dataReadyLatch;
     private final ReentrantLock dataReadyLock;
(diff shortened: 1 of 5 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -368,36 +369,67 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+        // get the lowest block number resending or scheduled to resend
+        final long earliestInFlight = queueByBlockMap.isEmpty() ? blockNumber : queueByBlockMap.firstKey();
+        final long activeResend = activeResendBlocks.isEmpty() ? blockNumber : activeResendBlocks.first();
+        final long scheduledResend = blocksToResend.isEmpty() ? blockNumber : blocksToResend.first();
+        final long earliestResend = Math.min(activeResend, scheduledResend);
+        final long earliestKnownGap = Math.min(earliestInFlight, earliestResend);
+        // Ensure we don't acknowledge blocks at or after the earliest block we
+        // still have not persisted. Note, this is not perfect, it cannot account
+        // for pre-existing gaps earlier than the last known persisted block.
+        return (earliestKnownGap <= blockNumber) ? earliestKnownGap - 1 : blockNumber;
```
AlfredoG87, line 445: off-by-one issue.
```suggestion
        return (earliestKnownGap < blockNumber) ? earliestKnownGap - 1 : blockNumber;
```

jsync-swirlds: If we are trying to acknowledge block `5`, and we are expecting to resend block `5` (scheduled or active) or currently sending block `5` (not yet complete), then I would expect to require we acknowledge block `4` instead.

Not sure how this is off by one...

jsync-swirlds: OK, there is an off-by-one, just not on this line.
Fixing now.

jsync-swirlds: Fixed.

AlfredoG87: Nice! 💯

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -368,36 +369,80 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+    ///     This value is adjusted lower than any blocks we know are not yet
+    ///     stored successfully (including resends, active streams, etc...).
+    ///
+    private long correctForResendAndStreaming(final long blockNumber) {
+        // get the lowest block number resending or scheduled to resend
+        // Use block number + 1 as a default if something isn't set, because we'll reduce by one later.
+        final long defaultValue = blockNumber + 1;
+        final long earliestInFlight = queueByBlockMap.isEmpty() ? defaultValue : queueByBlockMap.firstKey();
+        final long activeResend = activeResendBlocks.isEmpty() ? defaultValue : activeResendBlocks.first();
+        final long scheduledResend = blocksToResend.isEmpty() ? defaultValue : blocksToResend.first();
```
ata-nas, line 439: Calling `first` or `firstKey` on the sorted collections/maps will, unfortunately, throw if there are no items available.

We have a generic `getLastDequeItem` that will prevent that. We should really make more such generic methods that will prevent throws from ever happening.

I would suggest we move to a method `getFirst` that we develop, which will return `null` if item not found, and then the ternary operator will switch based on `item == null ? ...` instead of checking for `isEmpty` which does not guarantee that a subsequent get call will succeed.

jsync-swirlds: I checked for isEmpty _in a ternary_ specifically to avoid that possible throw.

The chance of any throw here is extremely small (and I have to re-check, but it might actually be _prevented_ by the JMM because a ternary is a single expression).

I believe this is acceptable at the current time, but I'll look into other alternatives as well.  Adding a method has its own issues (not least that if we do throw-and-catch, the performance cost is hundreds or thousands of times larger), but we might look into that at some point.

AlfredoG87 (approved):

Nana-EC (approved):

PR #2701, opened 2026-05-05
feat: Verification Design Document Update

ata-nas:
## Reviewer Notes

- Updates to the verification design document

## Related Issue(s)

Closes #2530

### DIFF
--- a/docs/design/block-verification.md
+++ b/docs/design/block-verification.md
@@ -23,7 +23,7 @@ that a block, received from any source is valid and not tampered with, is the
 central point of the Block Node.
 We achieve this by having a Verification Service Plugin.
 
-- Upon successful verification, the Verificaiton Service Plugin is able to
+- Upon successful verification, the Verification Service Plugin is able to
   propagate the block downstream via the internal messaging system.
 - Upon verification failure - be that the block fails to verify, parse, has
   missing mandatory items or fields, or any other reason, the Verification
@@ -36,7 +36,7 @@ from different sources, strict ordering of block reception CANNOT be guaranteed.
 It is therefore a requirement of the Verification Service Plugin to ensure that
 the strong ordering of messages for blocks that pass verification is maintained.
 This guarantee ensures that it is not possible for any given block to be left
-unprocessed donwstream. This essentially achieves a contiguous, in order data
+unprocessed downstream. This essentially achieves a contiguous, in order data
 stream that downstream plugins can process safely.
 
 ## Goals
@@ -168,7 +168,11 @@ data (block number and block root hash) for each block that is received and
 verified successfully. The Application State Facility is responsible for
 delivering this data to the All Blocks Hasher instance. Based on a configurable
 interval (measured in number of blocks), the Application State Facility will
-persist the All Blocks Hasher state to disk, thus making it persistent.
+persist the All Blocks Hasher state to disk, thus making it persistent. The
+configurable persistence interval is an optimization to avoid excessive disk
+usage. IO failures must be reported but must not prevent the Block Node from
+functioning. Upon restart, an attempt to update the all blocks hasher will be
+made using local storage.
 
 If all blocks hasher data is not available to the verification plugin, the
 handlers and verifiers automatically use the hash data provided in the block
@@ -229,8 +233,8 @@ Failure types include:
   - at least one of the other types of items (different from the
     above-mentioned) in between the header and the footer.
 - **MISSING_MANDATORY_FIELD**: If the block has all the mandatory items, but
-  one or more of the mandatory fields are missing, the block is considered
-  unverified.
+  one or more of the mandatory fields are missing (no value provided), the block
+  is considered unverified.
 - **CANCELLED**: If a session takes to long to complete, measured in N number of
   subsequent sessions started, the session is canceled and the block that the
   session is trying to verify is considered unverified.
--- a/docs/design/block-verification.md
+++ b/docs/design/block-verification.md
@@ -163,15 +163,16 @@ The Application State Facility (a component that manages state that has to be
 persistent and accessible across registered plugins) manages the All Blocks
 Hasher instance and its state. On startup, the Verification Service Plugin
 queries the Application State Facility to get the last data stored. Then, the
-Verification Service Plugin will update the Application State Facility all new
-data (block number and block root hash) for all blocks that are received and
-verified. The Application State Facility is responsible for gathering these
-values and maintaining the All Blocks Hasher instance. Based on a configurable
-interval (measured in N number of blocks), the Application State Facility will
-persist the All Blocks Hasher instance to disk, thus making it persistent.
+Verification Service Plugin will update the Application State Facility with new
+data (block number and block root hash) for each block that is received and
+verified successfully. The Application State Facility is responsible for
+delivering this data to the All Blocks Hasher instance. Based on a configurable
+interval (measured in number of blocks), the Application State Facility will
+persist the All Blocks Hasher state to disk, thus making it persistent.
 
-If initialization fails on startup, the handler degrades gracefully and
-verification falls back to the root hash provided in the block footer.
+If all blocks hasher data is not available to the verification plugin, the
+handlers and verifiers automatically use the hash data provided in the block
+footer.
 
 ## Design
 
@@ -193,24 +194,25 @@ verification falls back to the root hash provided in the block footer.
 3. The session handler, which also runs asynchronously, will continuously check
    for completed sessions and then process the results thereof. The handler is
    responsible for ordering correctly the messages sent downstream, keeping a
-   configurable sized buffer of sessions running - cancelling sessions if need
+   configurable sized buffer of completed results - cancelling sessions if need
    be, keeping metrics up to date, making updates to the all blocks hasher. The
    handler must be resilient and handle failures gracefully, ensuring an
-   uninterrupted flow of messages.
+   uninterrupted flow of results.
 
 ### Verification Failure Types
 
-Each time a session fails, it will return a meaningful result. The session
-handler will then process this result and take appropriate action(s), and then
-send a message downstream, using the internal messaging system. Those messages
-will be of type `VerificationNotification`. In case of failure, a failure type
-will be provided. Subscribers to the internal messaging system will be able to
-handle these messages and take appropriate action(s).
+Each time a session completes, it will return a meaningful result, whether
+success or failure. The session handler will then process this result, take
+appropriate action(s), and then send a notification, using the internal
+messaging system. Those notifications will be of type
+`VerificationNotification`. In case of failure, a failure type will be provided.
+Subscribers to the internal messaging system will be able to handle these
+messages and take appropriate action(s).
 
 Failure types could be standard or informational. The purpose of the
-verification service plugin is to successfully pass verification on every block.
-If we take block N and pass verification on it, subsequent failures on the same
-block are considered informational.
+verification service plugin is to successfully complete verification on every
+block. If we complete verification on a block with a success result, subsequent
+failures on the same block are considered informational.
 
 Failure types include:
 
--- a/docs/design/block-verification.md
+++ b/docs/design/block-verification.md
@@ -13,31 +13,75 @@
 
 ## Purpose
 
-The purpose of the Block Verification feature is to ensure that blocks received
-from consensus nodes are valid and have not been tampered with. This is achieved
-by re-calculating the block hash and verifying it against the signature provided
-by the Consensus Node.
+The Block Node is able to receive blocks from different types of sources.
+Once received, blocks must pass verification in order to be safely propagated
+downstream. Other plugins have access to the verified blocks via the internal
+messaging system.
+
+Block verification, or in other words the process through which we can verify
+that a block, received from any source is valid and not tampered with, is the
+central point of the Block Node.
+We achieve this by having a Verification Service Plugin.
+
+- Upon successful verification, the Verificaiton Service Plugin is able to
+  propagate the block downstream via the internal messaging system.
+- Upon verification failure - be that the block fails to verify, parse, has
+  missing mandatory items or fields, or any other reason, the Verification
+  Service Plugin will convey the failure throughout the Block Node via the
+  internal messaging system.
+- In both cases, any plugin can take an appropriate action.
+
+Due to the asynchronous nature of the Block Node and the way blocks are received
+from different sources, strict ordering of block reception CANNOT be guaranteed.
+It is therefore a requirement of the Verification Service Plugin to ensure that
+the strict order of blocks that pass verification is maintained. This guarantee
+ensures that it is not possible for any given block to be left unprocessed
+donwstream. This essentially achieves a contiguous, in order data stream that
+downstream plugins can process safely.
 
 ## Goals
 
-1. The block-node must re-create the block hash from the block items and verify
-   that it matches the hash implied by the signature.
-2. If verification fails, the block should be considered invalid, and
-   appropriate error-handling procedures must be triggered.
+1. The Block Node MUST be able to receive blocks as raw data from any given
+   source.
+2. The Block node MUST be able to start a block verification process that
+   always returns a result.
+3. The Block Node MUST be able to accurately convey verification results to
+   plugins downstream, utilizing the internal messaging system.
+   - Successful verification of blocks MUST be propagated downstream strictly
+     in order.
 
 ## Terms
 
 <dl>
-<dt>Consensus Node (CN)</dt><dd>A node that produces and provides blocks.</dd>
+<dt>Publisher</dt><dd>A type of data source that is able to publish
+(i.e. stream) blocks as raw data to the Block Node.</dd>
+<dt>Consensus Node (CN)</dt><dd>A type of Publisher. A node that produces and
+streams blocks.</dd>
 <dt>Block Items</dt><dd>The block data pieces (header, events, transactions,
 transaction result, state changes, proof) that make up a block.</dd>
-<dt>Block Hash</dt><dd>A cryptographic hash representing the block's integrity.</dd>
-<dt>Signature</dt><dd>The cryptographic signature on the block hash, created by the
-network aggregation of "share" private keys, as described in the TSS design.</dd>
-<dt>Public Key</dt><dd>The public key (a.k.a. Ledger ID) of the network that signed the block.</dd>
-<dt>Empty-tree Hash</dt><dd>The hash returned by a streaming Merkle tree hasher when
-no leaves were added (i.e. an empty subtree). Since HAPI v0.72 this is
+<dt>Block Hash</dt><dd>A cryptographic hash representing the block's
+integrity.</dd>
+<dt>Signature</dt><dd>The cryptographic signature on the block hash, created by
+the network aggregation of "share" private keys, as described in the TSS
+design.</dd>
+<dt>Public Key</dt><dd>The public key (a.k.a. Ledger ID) of the network that
+signed the block.</dd>
+<dt>Empty-tree Hash</dt><dd>The hash returned by a streaming Merkle tree hasher
+when no leaves were added (i.e. an empty subtree). Since HAPI v0.72 this is
 SHA-384(0x00), matching the CN convention.</dd>
+<dt>Block Verification</dt><dd>The process of verifying the integrity of a block
+by computing its hash and verifying its signature.</dd>
+<dt>Verification Session</dt><dd>A session that deals with the
+verification of a single block.</dd>
+<dt>Verification Result</dt><dd>The result of a block verification
+session.</dd>
+<dt>Verification Notification</dt><dd>A message sent to the internal messaging
+system that conveys verification session results and other data.</dd>
+<dt>Session Handler</dt><dd>A component that handles completed verification
+sessions. It is responsible for the correct post-processing of a completed
+session. This includes making updates to internal state, preserving and ensuring
+strict and correct order of messages sent downstream, handle failures and
+more.</dd>
 </dl>
 
 ## Entities
@@ -46,15 +90,16 @@ SHA-384(0x00), matching the CN convention.</dd>
 
 The main plugin class. Implements `BlockNodePlugin`, `BlockItemHandler`, and
 `BlockNotificationHandler`. It:
-
-- Receives the stream of block items via the messaging facility.
-- On detecting a `BLOCK_HEADER`, creates a `VerificationSession` via
-  `HapiVersionSessionFactory` and feeds all subsequent block items into it.
-- On session completion, sends a `VerificationNotification` via
-  `blockMessaging.sendBlockVerification()`.
-- Handles backfilled blocks arriving via `handleBackfilled()`.
-- Initialises and owns the `AllBlocksHasherHandler`.
-- Tracks all verification metrics.
+- Receives blocks from multiple sources, either in the form of a stream of
+individual block items, or a whole block all at once.
+- When a block is received, a verification session is started. Sessions run
+asynchronously. Sessions always produce an end result.
+- Completed sessions are handled by a session handler and appropriate action(s)
+are taken to ensure we comply with the strict ordering guarantee, but also
+to preserve historic hashes, update in-memory state, metrics and more.
+- The session handler uses the internal messaging system to communicate the
+result of a verification session. It utilizes a detailed
+`VerificationNotification` to convey the result.
 
 ### HapiVersionSessionFactory
 
@@ -68,20 +113,18 @@ based on the block's HAPI proto version:
 
 ### VerificationSession
 
-Interface with a single method:
-
-```java
-VerificationNotification processBlockItems(BlockItems blockItems) throws ParseException;
-```
-
-Returns `null` while the block is incomplete, and a `VerificationNotification`
-when the final batch (`isEndOfBlock() == true`) has been processed.
+- A single session deals with a single block.
+- It runs asynchronously.
+- It produces a Verification Result when complete.
+- It is able to process items dynamically, i.e. it can keep receiving items
+  in multiple batches.
 
 ### ExtendedMerkleTreeSession
 
-The full block verification implementation, used for HAPI v0.72.0 and above.
+A type of VerificationSession, the full block verification implementation, used
+for HAPI v0.72.0 and above.
 
-Maintains five `NaiveStreamingTreeHasher` instances — one per block-item
+Maintains five `StreamingTreeHasher` instances — one per block-item
 category — that incrementally compute subtree roots as items arrive:
 
 |         Hasher          |                      Block item types                      |
@@ -102,9 +145,9 @@ convention introduced in HAPI v0.72.
 
 ### DummyVerificationSession
 
-A temporary placeholder that unconditionally returns a success notification with
-a zeroed hash. It will be removed; HAPI versions prior to 0.72 are not supported
-in production.
+A type of VerificationSession, a temporary placeholder that unconditionally
+returns a success notification with a zeroed hash. It will be removed; HAPI
+versions prior to 0.72 are not supported in production.
 
 ### AllBlocksHasherHandler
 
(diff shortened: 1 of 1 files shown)


berryware (dismissed):
Looks good.

docs/design/block-verification.md
```diff
@@ -46,15 +93,16 @@ SHA-384(0x00), matching the CN convention.</dd>
...
-- Tracks all verification metrics.
+- Receives blocks from multiple sources, either in the form of a stream of
+individual block items, or a whole block all at once.
+- When a block is received, a verification session is started. Sessions run
+asynchronously. Sessions always produce an end result.
+- Completed sessions are handled by a session handler and appropriate action(s)
+are taken to ensure we comply with a strong ordering requirement, but also
+to preserve historic hashes, update in-memory state, metrics and more.
+- The session handler uses the internal messaging system to communicate the
+result of a verification session. It utilizes a detailed
```
jsync-swirlds, line 104: I think we might need to consider a change here.
We need to be able to select paths based on _block proof_ type, not _HAPI_ version.

Here's a possible approach:
A session still matches based on HAPI version but only handles the hashing process, and is managed as a thread.
The session, after producing a Root Hash, then forks out a _verifier_ for each block proof present after the block.
The session joins the results, then forwards _one_ result for each proof.
The Session Handler just makes sure sessions are started, managed, and closed timely.
The results are gathered from the session and forwarded to a Result Handler, which decides what to notify based on the set of results.

This means the session can focus on the hashing algorithm, the session handler can focus entirely on selection and thread management, the verifiers focus on the cryptography, and the result handler focuses on how to handle multiple results.

docs/design/block-verification.md
```diff
@@ -102,43 +149,96 @@ convention introduced in HAPI v0.72.
...
+The Application State Facility (a component that manages state that has to be
+persistent and accessible across registered plugins) manages the All Blocks
+Hasher instance and its state. On startup, the Verification Service Plugin
+queries the Application State Facility to get the last data stored. Then, the
+Verification Service Plugin will update the Application State Facility all new
+data (block number and block root hash) for all blocks that are received and
+verified. The Application State Facility is responsible for gathering these
+values and maintaining the All Blocks Hasher instance. Based on a configurable
+interval (measured in N number of blocks), the Application State Facility will
+persist the All Blocks Hasher instance to disk, thus making it persistent.
```
jsync-swirlds: Minor corrections:

```suggestion
Verification Service Plugin will update the Application State Facility with new
data (block number and block root hash) for each block that is received and
verified successfully. The Application State Facility is responsible for
delivering this data to the All Blocks Hasher instance. Based on a configurable
interval (measured in number of blocks), the Application State Facility will
persist the All Blocks Hasher state to disk, thus making it persistent.
```

docs/design/block-verification.md
```diff
@@ -102,43 +149,96 @@ convention introduced in HAPI v0.72.
...
+
+If initialization fails on startup, the handler degrades gracefully and
+verification falls back to the root hash provided in the block footer.
```
jsync-swirlds:
```suggestion
If all blocks hasher data is not available to the verification plugin, the
handlers and verifiers automatically use the hash data provided in the block
footer.
```

docs/design/block-verification.md
```diff
@@ -102,43 +149,96 @@ convention introduced in HAPI v0.72.
...
+   stream of block items continuously until the whole block is received. A
+   session must always complete eventually and produce a result of the
+   verification process.
+3. The session handler, which also runs asynchronously, will continuously check
+   for completed sessions and then process the results thereof. The handler is
+   responsible for ordering correctly the messages sent downstream, keeping a
+   configurable sized buffer of sessions running - cancelling sessions if need
+   be, keeping metrics up to date, making updates to the all blocks hasher. The
+   handler must be resilient and handle failures gracefully, ensuring an
+   uninterrupted flow of messages.
```
jsync-swirlds:
```suggestion
   configurable sized buffer of completed results - cancelling sessions if need
   be, keeping metrics up to date, making updates to the all blocks hasher. The
   handler must be resilient and handle failures gracefully, ensuring an
   uninterrupted flow of results.
```

docs/design/block-verification.md
```diff
@@ -102,43 +149,96 @@ convention introduced in HAPI v0.72.
...
+
+### Verification Failure Types
+
+Each time a session fails, it will return a meaningful result. The session
```
jsync-swirlds:
```suggestion
Each time a session completes, it will return a meaningful result, whether success or failure. The session
```

docs/design/block-verification.md
```diff
@@ -102,43 +149,96 @@ convention introduced in HAPI v0.72.
...
+handler will then process this result and take appropriate action(s), and then
+send a message downstream, using the internal messaging system. Those messages
+will be of type `VerificationNotification`. In case of failure, a failure type
+will be provided. Subscribers to the internal messaging system will be able to
+handle these messages and take appropriate action(s).
```
jsync-swirlds:
```suggestion
handler will then process this result, take appropriate action(s), and then
send a notification, using the internal messaging system. Those notifications
will be of type `VerificationNotification`. In case of failure, a failure type
will be provided. Subscribers to the internal messaging system will be able to
handle these messages and take appropriate action(s).
```

docs/design/block-verification.md
```diff
@@ -102,43 +149,96 @@ convention introduced in HAPI v0.72.
...
+
+Failure types could be standard or informational. The purpose of the
+verification service plugin is to successfully pass verification on every block.
+If we take block N and pass verification on it, subsequent failures on the same
+block are considered informational.
```
jsync-swirlds:
```suggestion
Failure types could be standard or informational. The purpose of the
verification service plugin is to successfully complete verification on every block.
If we complete verification on a block with a success result, subsequent failures on the same
block are considered informational.
```

docs/design/block-verification.md
```diff
@@ -184,7 +284,7 @@ Configuration class: `VerificationConfig`
 |--------------------------------------|----------------------------------------------------------------------|--------------------------------------------------------|
 | `allBlocksHasherFilePath`            | `/opt/hiero/block-node/verification/rootHashOfAllPreviousBlocks.bin` | Path for the persistent hasher snapshot                |
 | `allBlocksHasherEnabled`             | `true`                                                               | Enable or disable the all-blocks root hash computation |
-| `allBlocksHasherPersistenceInterval` | `10` (seconds)                                                       | How often the hasher snapshot is written to disk       |
+| `allBlocksHasherPersistenceInterval` | `10` (blocks)                                                        | How often the hasher snapshot is written to disk       |
```
jsync-swirlds, line 293: This should probably be removed, it will move to the Application State Facility.

jsync-swirlds (commented):
Good work.
A few suggestions and corrections.

docs/design/block-verification.md
AlfredoG87: So VerificationSession stays the same but we extract all the verification logic to BlockProofVerifiers that are independent of the Session, we would have one for TSS, another for RSA. StateBlockProof can be a variant of each of these, TSS StateBlockProof and RSA StateBlockProof. ?

I like this approach.

docs/design/block-verification.md
ata-nas: ✅ Done, thanks for wording improvement suggestions!

docs/design/block-verification.md
ata-nas: ✅ Done, thanks for wording improvement suggestions!

docs/design/block-verification.md
ata-nas: ✅ Done, thanks for wording improvement suggestions!

docs/design/block-verification.md
ata-nas: ✅ Done, thanks for wording improvement suggestions!

docs/design/block-verification.md
ata-nas: ✅ Done, thanks for wording improvement suggestions!

docs/design/block-verification.md
ata-nas: ✅ Done, thanks for wording improvement suggestions!

docs/design/block-verification.md
ata-nas: All of these values are due to change soon. Let's update later, as these are still present and active, at least for the foreseeable future.

docs/design/block-verification.md
ata-nas: @AlfredoG87 as far as I also understand this, yes. It seems good that we decouple the hashing logic from the verification logic. In that way both are easily maintainable and extendable. We will be able to dynamically build what we need as we keep receiving items.

docs/design/block-verification.md
```diff
@@ -13,31 +13,78 @@
...
+Once received, blocks must pass verification in order to be safely propagated
+downstream. Other plugins have access to the verified blocks via the internal
+messaging system.
+
+Block verification, or in other words the process through which we can verify
+that a block, received from any source is valid and not tampered with, is the
+central point of the Block Node.
+We achieve this by having a Verification Service Plugin.
+
+- Upon successful verification, the Verificaiton Service Plugin is able to
```
Nana-EC:
```suggestion
- Upon successful verification, the Verification Service Plugin is able to
```

docs/design/block-verification.md
```diff
@@ -13,31 +13,78 @@
...
+  Service Plugin will convey the failure throughout the Block Node via the
+  internal messaging system.
+- In both cases, any plugin can take an appropriate action.
+
+Due to the asynchronous nature of the Block Node and the way blocks are received
+from different sources, strict ordering of block reception CANNOT be guaranteed.
+It is therefore a requirement of the Verification Service Plugin to ensure that
+the strong ordering of messages for blocks that pass verification is maintained.
+This guarantee ensures that it is not possible for any given block to be left
+unprocessed donwstream. This essentially achieves a contiguous, in order data
```
Nana-EC:
```suggestion
unprocessed downstream. This essentially achieves a contiguous, in order data
```

docs/design/block-verification.md
```diff
@@ -102,43 +149,98 @@ convention introduced in HAPI v0.72.
...
+  session is trying to verify is considered unverified.
+- **UNKNOWN_ERROR**: If an unexpected error occurs during the verification
+  process, the block is considered unverified.
+
+In case of failures that are informational, the prefix `INFO_` is added to the
+failure type.
+
+---
 
 Sequence Diagram:
```
Nana-EC, line 249: Q: DO we need to update this diagram to capture async nature of session handlers?

docs/design/block-verification.md
```diff
@@ -102,43 +149,98 @@ convention introduced in HAPI v0.72.
...
+- **MISSING_MANDATORY_ITEM**: If a mandatory item for a block is missing,
+  the block is considered unverified. Mandatory items are:
+  - `BLOCK_HEADER`
+  - `BLOCK_FOOTER`
+  - `BLOCK_PROOF` (at least one)
+  - at least one of the other types of items (different from the
+    above-mentioned) in between the header and the footer.
+- **MISSING_MANDATORY_FIELD**: If the block has all the mandatory items, but
+  one or more of the mandatory fields are missing, the block is considered
+  unverified.
```
Nana-EC: Might need an example here. I asked a similar question in your other PR

docs/design/block-verification.md
```diff
@@ -13,31 +13,78 @@
...
+- Upon verification failure - be that the block fails to verify, parse, has
+  missing mandatory items or fields, or any other reason, the Verification
+  Service Plugin will convey the failure throughout the Block Node via the
+  internal messaging system.
+- In both cases, any plugin can take an appropriate action.
+
+Due to the asynchronous nature of the Block Node and the way blocks are received
+from different sources, strict ordering of block reception CANNOT be guaranteed.
+It is therefore a requirement of the Verification Service Plugin to ensure that
+the strong ordering of messages for blocks that pass verification is maintained.
```
Nana-EC, line 37: I think it's more accurate to say that the output of the plugin and emitting verification notification in order is where the strong ordering must be ensured. The plugin itself must also handle out of order blocks

Nana-EC: You say this well at the end of Goal 3

docs/design/block-verification.md
```diff
@@ -102,43 +149,98 @@ convention introduced in HAPI v0.72.
...
-falls back to the root hash provided in the block footer.
+The Application State Facility (a component that manages state that has to be
+persistent and accessible across registered plugins) manages the All Blocks
+Hasher instance and its state. On startup, the Verification Service Plugin
+queries the Application State Facility to get the last data stored. Then, the
+Verification Service Plugin will update the Application State Facility with new
+data (block number and block root hash) for each block that is received and
+verified successfully. The Application State Facility is responsible for
+delivering this data to the All Blocks Hasher instance. Based on a configurable
+interval (measured in number of blocks), the Application State Facility will
```
Nana-EC, line 170: I imagine the interval is needed so that we're not updating every block. If so highlight the optimization and reason.
Also highlight how we'd handle failure to persist and BN failure, when we come back up would there need to be a recalculation? Is that in scope or out of scope?

Nana-EC (commented):
Good additions.
Provided some suggestions

docs/design/block-verification.md
ata-nas: Yes, this diagram will be completely reworked.

docs/design/block-verification.md
ata-nas: ✅ Done, thanks!

docs/design/block-verification.md
ata-nas: ✅ Done, thanks!

docs/design/block-verification.md
ata-nas: ✅ Done, thanks! Updated, value not provided.

docs/design/block-verification.md
ata-nas: ✅ Done, thanks! Updated.

docs/design/block-verification.md
jsync-swirlds: State Proof might still be singular; but might hand the _signature_ portion off to a shared signature verification element.

docs/design/block-verification.md
jsync-swirlds: Fair point.

docs/design/block-verification.md
```diff
@@ -13,31 +13,78 @@
...
-no leaves were added (i.e. an empty subtree). Since HAPI v0.72 this is
+<dt>Block Root Hash</dt><dd>A SHA2-384 hash at the root of the block merkle tree
+as defined in HIP-1424.</dd>
+<dt>Signature</dt><dd>A cryptographic operation on the Block Root Hash. The
+exact operation depends on signature type and algorithm, but may include a
+TSS/hinTS algorithm as defined in HIP-1200, a collection of RSA signatures, a
+collection of Schnorr signatures, or a (future) TSS signature defined for
+post-quantum cryptography.</dd>
+<dt>Public Key</dt><dd>The public key (which is algorithm-dependent) that a node
+or network will use to sign events and block proofs.</dd>
```
ivannov, line 71: Question: do we use the public keys to sign stuff or to verify signatures?

docs/design/block-verification.md
ivannov, line 287: If it is the Application State Facility that is responsible for managing (including persistence) of all back hasher data, maybe we don't need this section at all? Or at least not in this document

ivannov, line 304: Maybe these metrics need to be reworked as well? We either have one counter for all possible verification failure reasons (bad block proof, unable to parse, missing mandatory items, etc.), or one separate metric for each.

I would say the first option with a future option of assigning labels for the failure reason

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
+Failure types could be standard or informational. The purpose of the
+verification service plugin is to successfully complete verification on every
+block. If we complete verification on a block with a success result, subsequent
+failures on the same block are considered informational.
+
+Failure types include:
+
+- **BAD_BLOCK_PROOF**: If the computed hash does not match the hash signed by
+  the network, or the signature cannot be verified as correct, the block is
+  considered unverified.
```
ivannov, line 225: Nit: maybe it's just me, but _unverified_ sounds as if the block's verification did not start at all, so its state is unknown. A better wording would be: _the block's verification is considered failed._

Feel free to ignore this

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
+- **MISSING_MANDATORY_FIELD**: If the block has all the mandatory items, but
+  one or more of the mandatory fields are missing (no value provided), the block
+  is considered unverified.
+- **CANCELLED**: If a session takes to long to complete, measured in N number of
+  subsequent sessions started, the session is canceled and the block that the
+  session is trying to verify is considered unverified.
+- **UNKNOWN_ERROR**: If an unexpected error occurs during the verification
+  process, the block is considered unverified.
+
+In case of failures that are informational, the prefix `INFO_` is added to the
```
ivannov, line 244: Should this be a boolean field in the verification notification?

jsync-swirlds (approved):

docs/design/block-verification.md
```diff
@@ -13,31 +13,78 @@
...
-network aggregation of "share" private keys, as described in the TSS design.</dd>
-<dt>Public Key</dt><dd>The public key (a.k.a. Ledger ID) of the network that signed the block.</dd>
-<dt>Empty-tree Hash</dt><dd>The hash returned by a streaming Merkle tree hasher when
-no leaves were added (i.e. an empty subtree). Since HAPI v0.72 this is
+<dt>Block Root Hash</dt><dd>A SHA2-384 hash at the root of the block merkle tree
+as defined in HIP-1424.</dd>
+<dt>Signature</dt><dd>A cryptographic operation on the Block Root Hash. The
+exact operation depends on signature type and algorithm, but may include a
+TSS/hinTS algorithm as defined in HIP-1200, a collection of RSA signatures, a
+collection of Schnorr signatures, or a (future) TSS signature defined for
```
AlfredoG87, line 68:
```suggestion
Schnorr signature, or a (future) TSS signature defined for
```

or a `Combined Schnorr signature` we don't see a collection of schnorr signatures as we see for RSA case on WRBs

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
 
 ### DummyVerificationSession
 
-A temporary placeholder that unconditionally returns a success notification with
-a zeroed hash. It will be removed; HAPI versions prior to 0.72 are not supported
-in production.
+A type of VerificationSession, a temporary placeholder that unconditionally
+returns a success notification with a zeroed hash. It will be removed; HAPI
+versions prior to 0.76 are not supported in production.
```
AlfredoG87, line 155: Q: We need to start thinking on removing this Session to avoid pushing it to production.
@jsync-swirlds @ata-nas @Nana-EC

Do we still rely on this for any type of testing? Solo-e2e runs with TSS, BNCE runs with TSS, also we have the RSA for WRBs that will run on its own environment. I believe most of the usage for this is only in UTs and Simulator. We might want to address a deprecation path. Nothing to do right now on this PR, but noting that it is becoming increasingly a priority to remove it.

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
+interval (measured in number of blocks), the Application State Facility will
+persist the All Blocks Hasher state to disk, thus making it persistent. The
+configurable persistence interval is an optimization to avoid excessive disk
+usage. IO failures must be reported but must not prevent the Block Node from
+functioning. Upon restart, an attempt to update the all blocks hasher will be
+made using local storage.
+
+If all blocks hasher data is not available to the verification plugin, the
+handlers and verifiers automatically use the hash data provided in the block
+footer.
```
AlfredoG87, line 179: Is this still true? I thought we will not be using it at all for now or in the foreseeable future.

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
-9. On failure, the plugin sends a failure notification (`success = false`).
+### Core Design Workflow
+
+1. The Verification Service Plugin receives data from multiple sources.
+   The plugin is source-agnostic. It verifies every block according to the same
+   rules and only includes the source in the result notification. At the
+   moment, we have two distinct sources:
+   1. **Publisher** - blocks arrive as a stream of individual items
+   2. **Backfill** - blocks are received via backfill and are provided to the
+      plugin as a whole block.
```
AlfredoG87, line 191: should we mentioned that we are using a backfillNotification as opposed to the BlockItemsHandler.

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
+2. The plugin will start a verification session for each received block. The
+   session runs asynchronously, and it has the responsibility of running the
+   verification process. The sessions could accept a whole block, or be fed a
+   stream of block items continuously until the whole block is received. A
+   session must always complete eventually and produce a result of the
+   verification process.
```
AlfredoG87, line 197: what happens for incomplete blocks, lets say that a publisher starts streming block 10, wins the race condition and streams half the block. then goes silent/stalled and we receive block 11, block 12 and block 13, so we decide we need to kill block10 connection and ask for a RESEND(10) to other CNs.

Verification should overwrite a verification session for the same block, and if truly stalled, it should timeout eventually in order to comply with the `must always complete` rule (which is a good rule, otherwise we would end-up with orphaned half-way verification sessions that leak memory long term)

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
+3. The session handler, which also runs asynchronously, will continuously check
+   for completed sessions and then process the results thereof. The handler is
+   responsible for ordering correctly the messages sent downstream, keeping a
+   configurable sized buffer of completed results - cancelling sessions if need
+   be, keeping metrics up to date, making updates to the all blocks hasher. The
+   handler must be resilient and handle failures gracefully, ensuring an
+   uninterrupted flow of results.
```
AlfredoG87, line 204: Right now, I belive the publisher is using the Notifications of Persistence in order to detect stalled blocks. I need to verify, and is not sending ACKs out of order, so maybe we can relax the Verification sends the verifications results in order, so that we don't break current stalled/silent connections on publisher recover mechanism.

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
+
+### Verification Failure Types
+
+Each time a session completes, it will return a meaningful result, whether
+success or failure. The session handler will then process this result, take
+appropriate action(s), and then send a notification, using the internal
+messaging system. Those notifications will be of type
+`VerificationNotification`. In case of failure, a failure type will be provided.
```
AlfredoG87, line 212: should we also provide success details/type? like kind of signature was detected, amount of blockProofs present vs verified, and potentially other valuable success information to downstream consumers?

for some plugins this information might be useful in the future. for example, persistence might strip some `repetitive` data based on the signature. or store only 1 blockProof, or backfill might want to try to overwrite a block with stateProofs that might have an indirect proof on this block, but a direct proof on other BNs, etc.

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
+  - `BLOCK_FOOTER`
+  - `BLOCK_PROOF` (at least one)
+  - at least one of the other types of items (different from the
+    above-mentioned) in between the header and the footer.
+- **MISSING_MANDATORY_FIELD**: If the block has all the mandatory items, but
+  one or more of the mandatory fields are missing (no value provided), the block
+  is considered unverified.
+- **CANCELLED**: If a session takes to long to complete, measured in N number of
+  subsequent sessions started, the session is canceled and the block that the
+  session is trying to verify is considered unverified.
```
AlfredoG87, line 240: This matches my question above, for stalled blocks we will receive the same block again from beginning and we might need to cancel one verification session and start another, is this the same as cancelled, or we need a overwritten, and depending on that is CANCELLED enough information or should be said CANCELLED due to TIMEOUT.

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
+- **UNKNOWN_ERROR**: If an unexpected error occurs during the verification
+  process, the block is considered unverified.
+
```
AlfredoG87, line 243: Do we want to include the case we discuss today, we receive a TSS block but we lack the LEDGERID to verify it, or we receive a WRB/RSA block, but we lack the roster needed to verify it, etc..

docs/design/block-verification.md
```diff
@@ -102,43 +149,102 @@ convention introduced in HAPI v0.72.
...
+In case of failures that are informational, the prefix `INFO_` is added to the
+failure type.
```
AlfredoG87, line 245: do we want to depend on a prefix on the FailureType?

AlfredoG87 (approved):
Great Work, there are some questions and a couple nits.

Let me know if you want to address these as part of this PR or the next that touches and improves the documentation, as you know you have it on the to-do to further improve this.

Unblocking in case you want to capture the nits and suggestions for the next one, we can further meet to discuss the suggestions

docs/design/block-verification.md
jsync-swirlds: Agreed, removing that session type is priority.
We can remove from the _existing_ verification plugin as soon as we ensure it's not needed for testing.

For the new verification plugin (which has a different session/verification split), we're not going to implement the dummy verification option.

docs/design/block-verification.md
jsync-swirlds: I believe this document is for an interim stage; it doesn't yet describe the final form for verification.

docs/design/block-verification.md
jsync-swirlds: The new structure doesn't need to overwrite the session, and has stalled block detection.

If stalled (enough later blocks complete before it does) it will be cancelled by the manager (which completes the session) and the result handler produces a verification notification for failed validation with reason "cancelled".
This informs the source (backfill or publisher) that the block failed to complete within a reasonable time and needs to be re-requested.

For multiple validations of the same block, we can allow multiple simultaneous and if all succeed, good, if some succeed and some fail, we handle that by making fail-not-persisted a regular failed verification and failed-already persisted is an informational failed verification.

We probably need to document that better here.

docs/design/block-verification.md
jsync-swirlds: The Publisher plugin detects stalls via blocks not-yet-sent to messaging, so verification holding things in order won't affect that.

The ACK handling will only benefit from verification maintaining a modest effort to keep _successful_ results in order.

Ordered success also makes it easier for persistence to keep things in order (less staging usage for cloud plugins, for instance) and will make it _much_ easier for state management.

docs/design/block-verification.md
jsync-swirlds: 1. We should never allow any block to pass if _any_ block proof fails verification, regardless of how many there are.
1. We should always verify every block proof on the block (if we can; obsolescence may limit this someday).
1. We should never strip a block proof off a block, that's dishonest and potentially damaging.  Third-party plugins might someday decide to do so, but they can do so by examining the proofs on the block; they don't need the verification plugin to tell them what can or cannot be removed.

The comment about backfill is interesting. It think we might need to consider that more carefully for a future upgrade; but we probably shouldn't try to do anything along those lines in the next several releases, there is nothing inherently "lesser" about a StateProof type BlockProof that makes it worth trying to replace it (in fact we hope to add a StateProof type BlockProof to every single WRB soon after TSS cutover, that's most of the initial reason for the WRB process and "all blocks subtree").

docs/design/block-verification.md
jsync-swirlds: Cancelled is for stalled blocks; regardless of how we detect that it's stalled.

docs/design/block-verification.md
jsync-swirlds: Good point.  We might want to add a UNABLE_TO_VERIFY code (I don't think we need more detail than that, for a lot of good reasons).

docs/design/block-verification.md
jsync-swirlds: It's not really depending on a prefix, more that each failure type will have an `INFO` variant that other plugins can choose to handle differently.

We should make that clearer in this document.

ata-nas: So, this is something that came to my mind, but also @ivannov also mentioned.

We could have the `INFO` variant and that be a prefix to the name, or we could have a boolean field (isInformational). The reason why I initially went with the prefix is so that whenever someone switches the enum, they will be forced to handle informational cases. Whereas with a boolean, it might be overlooked.

It is true that if we have a boolean, we will have twice as less failure types, however. Not strongly convicted to any of the approaches.

docs/design/block-verification.md
ata-nas: Yes, I have already added these types. For now, I have a missing RSA keys type and missing TSS parameters type. I could merge them, but maybe making a distinction here will be beneficial for other plugins, for instance rsa bootstrap plugin could listen to verification failures and when it sees missing rsa keys can proactively try to supply them instead of waiting potentially for a very long time.

docs/design/block-verification.md
ata-nas: I also agree here, I was unsure if we still need this. There are a few todo's in the current version of the plugin which we have to address in the new one, but we have to confirm.

ata-nas:
@ivannov @Nana-EC @jsync-swirlds @AlfredoG87 feedback will be addressed in #2879 so we can close this.
Thanks for your comments!

PR #2704, opened 2026-05-05
feat: Revisit Hash On-Init for AllBlocksHasher

### LINKED ISSUE #2451: Revisit Hash On-Init for AllBlocksHasher
### Story Form

As a Verification Plugin
I want to be able to efficiently initialize the all blocks hasher on startup (init)
So that I can efficiently initialize myself

### Technical Notes

- We should revisit the current all blocks hasher
- We should improve the initialization of the all blocks hasher on startup

ata-nas:
* Move the `AllPreviousBlocksRootHashHasherSnapshot` to internal protos
* All blocks hasher general cleanup and hardening
* Configuration property for optional hashing from history
* Remove time-based `AllPreviousBlocksRootHashHasherSnapshot` persistence and make it block based (configurable)
* Consolidate verification additions from #2535 and #2666

## Related Issue(s)

Closes #2461


ivannov (commented):
Looks to me that this PR captures whatever it is expected to capture.

I give my official unofficial +1, but I would leave other folks confirm

block-node/block-verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -163,15 +170,30 @@ public VerificationNotification finalizeVerification(Bytes rootHashOfAllBlockHas
...
+            BlockProof stateBasedProof = getSingle(blockProofs, BlockProof::hasBlockStateProof);
+            if (stateBasedProof != null) {
+                return getStateProofVerificationResult(
+                        stateBasedProof,
+                        previousBlockHashToUse,
+                        rootOfAllPreviousBlockHashes,
+                        startOfBlockStateRootHash);
+            }
+        } catch (final IllegalStateException e) {
+            LOGGER.log(WARNING, "Block [{0}] has malformed proof structure: %s".formatted(e.getMessage()), e);
```
Nana-EC, line 191: Looks like you're missing logic to insert the block number into the log.
Also probably don't need both message and the exception itself

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
```diff
@@ -35,5 +35,11 @@ public enum FailureType {
         UNABLE_TO_PARSE,
         /// This type indicates that the block is missing a mandatory item
         MISSING_MANDATORY_ITEM,
+        /// This type indicates that the block is missing a mandatory field (null)
+        MISSING_MANDATORY_FIELD,
+        /// This type indicates that the session was cancelled
+        CANCELLED,
+        /// This type indicates that an unknow error occurred
```
Nana-EC, line 42:
```suggestion
        /// This type indicates that an unknown error occurred
```

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
```diff
@@ -35,5 +35,11 @@ public enum FailureType {
         UNABLE_TO_PARSE,
         /// This type indicates that the block is missing a mandatory item
         MISSING_MANDATORY_ITEM,
+        /// This type indicates that the block is missing a mandatory field (null)
+        MISSING_MANDATORY_FIELD,
```
Nana-EC, line 39: Q: Is this a field within a proto message?

block-node/block-verification/src/main/java/org/hiero/block/node/verification/VerificationConfig.java
```diff
@@ -12,15 +12,18 @@
...
 ///     are persisted across restarts as a serialized `LedgerIdPublicationTransactionBody`.
 ///     Written when block 0 is processed. Loaded on startup to restore full TSS state.
 @ConfigData("verification")
 public record VerificationConfig(
-        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/verification/rootHashOfAllPreviousBlocks.bin") Path allBlocksHasherFilePath,
-        @Loggable @ConfigProperty(defaultValue = "true") boolean allBlocksHasherEnabled,
-        @Loggable @ConfigProperty(defaultValue = "10") int allBlocksHasherPersistenceInterval,
-        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/verification/tss-parameters.bin") Path tssParametersFilePath) {}
+    @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/verification/rootHashOfAllPreviousBlocks.bin") Path allBlocksHasherFilePath,
+    @Loggable @ConfigProperty(defaultValue = "false") boolean allBlocksHasherEnabled,
```
Nana-EC, line 24: Q: is this the logic to not use the history of block hashes  tree in the block verification or is this just the calculation of the tree?

block-node/block-verification/src/main/java/org/hiero/block/node/verification/AllBlocksHasherHandler.java
```diff
@@ -48,12 +49,14 @@
 ///
 /// This class is available/functional only when a node has complete historical block store from genesis.
 public class AllBlocksHasherHandler {
-
     private static final System.Logger LOGGER = System.getLogger(AllBlocksHasherHandler.class.getName());
     /// Empty tree hash: SHA384(0x00), matching the CN's IncrementalStreamingHasher for an empty tree.
     public static final byte[] ZERO_BLOCK_HASH = EMPTY_TREE_HASH;
     /// Expected size of a block hash in bytes.
     public static final int BLOCK_HASH_LENGTH = ZERO_BLOCK_HASH.length;
+    /// The initial value of the [#persistenceThresholdCounter], while this value is
```
Nana-EC, line 57: Q: Do you mean where it's -1 persistence of the of allBlocksHash is disabled?

block-node/block-verification/src/main/java/org/hiero/block/node/verification/AllBlocksHasherHandler.java
```diff
@@ -48,12 +49,14 @@
...
+    /// present, persistence is inactive. This is during init, on start go to 0 as default value.
+    private static final int UNINITIALIZED_PERSISTENCE_THRESHOLD = -1;
```
Nana-EC, line 59: I don't see this getting set to 0 or anything else.
Also when used it checks if the counter is greater that it. 
The counter that is compared to it also only checks for > meaning you only need one increment, so the threshold is always 1.
Is that the intention?

Also is this variable needed if there's also a variable for whether it's enabled? Is there ever a case where it's enabled but persistence is disabled? If not then just use the config value

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
```diff
@@ -52,7 +52,7 @@ default BlockUnparsed blockUnparsed() {
             }
             return BlockUnparsed.PROTOBUF.parse(
                     rawData.toReadableSequentialData(), false, false, Codec.DEFAULT_MAX_DEPTH, MAX_BLOCK_SIZE_BYTES);
-        } catch (final ParseException e) {
+        } catch (final RuntimeException | ParseException e) {
```
Nana-EC, line 55: Hmm, this swallows a more concerning error no? Currently if the block can't be parsed we swallow and log, but Runtime would catch more severe considerations.
Any culprits that you're trying to catch right now?

Might be a larger problem but if such an issue occurs should we be trying to clean up so that the BN can backfill it?

block-node/block-verification/src/main/java/org/hiero/block/node/verification/AllBlocksHasherHandler.java
```diff
@@ -323,12 +338,26 @@ private void validateState() {
...
+    /// Append the latest block hash to the streaming hasher for all previous blocks.
+    /// We can only do so if the hasher [#isAvailable()] and the hash we provide
+    /// is the next one in line.
     /// @param blockHashBytes the latest block hash to append
-    public void appendLatestHashToAllPreviousBlocksStreamingHasher(@NonNull final byte[] blockHashBytes) {
-        if (hasher != null) {
+    public void appendLatestHashToAllPreviousBlocksStreamingHasher(
+            @NonNull final byte[] blockHashBytes, final long blockNumber) {
+        if (isAvailable()) {
+            // @todo(2461) use blockNumber
```
Nana-EC, line 349: yeah, unclear of why blockNumber isn't the only approach used here.
The the counter comparisons seem like proxies. I might need to circle back to this

Nana-EC (commented):
Good start, did a first pass with some suggestions and questions.
Will need a second pass.
@jsync-swirlds  and @AlfredoG87 should review next week

block-node/block-verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: This structure doesn't work correctly; the following should be better.

```suggestion
            LOGGER.log(WARNING, "Block %d has malformed proof structure".formatted(blockNumber), e);
```

block-node/block-verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -227,6 +249,184 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+        int totalSiblings = siblings.size();
+        // Must have at least 3 (signed block siblings) and remainder must be groups of 4 (gap blocks)
+        if (totalSiblings < 3 || (totalSiblings - 3) % 4 != 0) {
+            LOGGER.log(
+                    WARNING,
+                    "Block {0} state proof sibling count {1} is invalid (need >= 3, remainder must be multiple of 4)",
+                    blockNumber,
+                    totalSiblings);
+            return false;
+        }
```
jsync-swirlds, line 345: Actually, the groups of 4 part might not be true.
We can check `>= 3`, but probably shouldn't check past that, because there are different ways to structure things past the first 3.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
ata-nas: Thanks, I'll fix in the follow up.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
ata-nas: Yes, if a message is missing a mandatory field (no value provided).

block-node/block-verification/src/main/java/org/hiero/block/node/verification/VerificationConfig.java
ata-nas: So, with the next PR for the new plugin, where we will drastically rework things, the all blocks hasher will simply no longer be a part of the verification process. As far as the enable/disable configuration, currently if enabled, we will attempt to initialize it and if successful, we will use for verification process. In the next PR, we can either remove it, or it will simply mean "try to initialize it and use it only to store new hashes, but will no longer be part of the verification process".

block-node/block-verification/src/main/java/org/hiero/block/node/verification/AllBlocksHasherHandler.java
ata-nas: If we have successfully initialized the hasher, meaning it is enabled and working, on start we set this to 0. The sentinel value is used so that if -1, no persistence will be attempted.

This can and should probably change when we completely move the all blocks hasher out of verification.

block-node/block-verification/src/main/java/org/hiero/block/node/verification/AllBlocksHasherHandler.java
ata-nas: The value is changed on `start()` if we are available.

Here is how the counter is used:
On each successful addition of a new hash, the counter is incremented assuming the all blocks hasher is available. If the threshold is passed, then we roll it back and persist. Subsequent calls continue to increment the counter until the threshold is passed and we repeat the process. The reason for the counter is that we want to configure an interval for persistence to avoid excessive disk usage.

Note that with upcoming PR, there will be only a single place where these updates will happen, so we will not worry about things racing. And in the future when the all blocks hasher is completely removed from verification plugin, we will simply have to report the newest hash and forget about the rest, from the perspective of the verification plugin that is.

And finally, to answer your last question, no, there are no cases where the all blocks hasher is enabled and persistence disable. There are cases where the all blocks hasher is enabled, but unavailable for any given reason.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
ata-nas: So, the thing is that we have defined the block accessor to return `null` if something happens during reading. There was a fairly recent change that made the same additions to implementations, but not here. I've made this addition so that we do not end up in a situation where we throw while trying to get an accessor. The caller will receive `null` if the block is not available or something happened when trying to read it.

block-node/block-verification/src/main/java/org/hiero/block/node/verification/AllBlocksHasherHandler.java
ata-nas: This will be much improved once we move out the whole all blocks hasher out of verification. For now, (with an upcoming PR) the all blocks hasher will be extracted from the verification process and there will be a single point (thread) that will update this when appropriate. The counter is there to introduce an interval for persistence.

block-node/block-verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
ata-nas: Interesting, thanks for the note. This is a straight up copy of existing logic. Once we have a follow up of this PR, we will be able to decouple hashing logic from proof verification logic and we can, piece by piece address these.

block-node/block-verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
ata-nas: All of the new plugin was a straight up copy of the existing one. As things will change very drastically with the next PR, all of these will be addressed.

block-node/block-verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: Let's make sure to record that follow-up work in an issue, please.

block-node/block-verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: This is just a simple error in the code.  I think we can easily fix this defect immediately.

block-node/block-verification/src/main/java/org/hiero/block/node/verification/VerificationConfig.java
AlfredoG87: Yes @Nana-EC  this is the property that is currently set as false by default, and allowed us to verify blocks out of order.

I say we keep it but we change its description, if we are still going to calculate the SubTree, so we can enable or disable the calculation of the SubTree (even if we don't use it) so that we can have that as disabled by default too, until we have a use case for enabling it.

Now if we move out of verification to a new plugin or to another plugin, then we should remove the property from verification config.

jsync-swirlds (approved):

block-node/block-verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
ata-nas: Will do in the next PR that is already on top of this branch. A general revision of logging will also be done.

block-node/block-verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
ata-nas: I think this might have been possibly resolved, let me take another look in a different PR.

Nana-EC (approved):
LG. Items to be addressed in follow up

ata-nas:
@Nana-EC @jsync-swirlds thanks for your feedback. I will use it for the next PR we have for verification.

PR #2709, opened 2026-05-05
fix: Update server configuration for 125M block items

### LINKED ISSUE #2708: Update Server Config max message size to 125M
### Technical Notes
We already updated the overly widely used constant to 125M(131072000), we need to also update the configuration value that _should_ be used.
At the same time, increase the maximum value to 1.5G(1610612736) and the minimum to 1M(1048576) for that constant, to allow for reasonable configuration flexibility.

jsync-swirlds:
## Reviewer Notes
* Updated `ServerConfig.maxMessageSizeBytes` default, min, and max
* Updated `StreamPublisherPlugin` to remove the 16K reduction.

## Related Issue(s)
Fixes #2708

### DIFF
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
@@ -12,7 +12,7 @@ import org.hiero.block.node.base.Loggable;
  *
  * <p>ServerConfig will have settings for the server.
  *
- * @param maxMessageSizeBytes the http2 max message/frame size in bytes
+ * @param maxMessageSizeBytes the PBJ max message size in bytes
  * @param socketSendBufferSizeBytes the socket send buffer size in bytes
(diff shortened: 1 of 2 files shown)


AlfredoG87 (approved):
LGTM

Nana-EC:
Q: Why was 16k required previously and no longer anymore. I don't think we had it documented anywhere

Nana-EC (approved):
LG, I'm forgetting the context of the 16k

AlfredoG87:
> Q: Why was 16k required previously and no longer anymore. I don't think we had it documented anywhere

It was recommended by Anthony and Joseph on a Slack Thread, sending it offline.

jsync-swirlds:
> LG, I'm forgetting the context of the 16k

Originally we thought it would be a good idea to have a slight reduction on the BN side to account for overhead in the _network_ layer.
That turned out to not be needed, so it's obsolete.

PR #2716, opened 2026-05-06
fix(publisher): prune stale `blocksToResend` entries beyond a configurable buffer

### LINKED ISSUE #2714: Bug(publisher): Stale entry in `blocksToResend` causes BN to respond with RESEND for a long-persisted block to every new publisher
### Description

Once a block number is added to `LiveStreamPublisherManager.blocksToResend` (via failed verification, mid-block EndStream, or stall detection), it is only ever removed when a publisher sends a header for that exact block number. If the block is later persisted by backfill, the entry is never cleaned up.

After that, every fresh publisher connection ends in the same way:

1. Publisher connects, streams 1–2 current blocks successfully.
2. On `endOfBlock`, `nextBlockToResend()` returns the stale low block number (it does not filter by `lastPersistedBlockNumber`).
3. BN sends `RESEND(<stale block>)` to the publisher.
4. Publisher's earliest available block is far ahead, so it replies `EndStream(TOO_FAR_BEHIND)` and disconnects.
5. Backfill fills the resulting small gap.
6. Cycle repeats every ~15–30 s.

Net effect: publishers churn and disconnect continuously, while backfill keeps the chain moving — masking the issue from a "blocks still arriving" perspective.

## Reproduction

Prerequisites: a BN under test with a backfill peer source configured (any peer that can serve a small range of blocks beyond the gap).

1. Start a fresh BN. Connect **Publisher A** via `publishBlockStream`.
2. Publisher A streams blocks `0..N` to completion normally (`BlockHeader` → items → `BlockProof` → `BlockEnd`), receiving `Acknowledgement` for each.
3. For block `N+1`, Publisher A sends only the `BlockHeader`, then immediately sends `EndStream(RESET)` (or `TIMEOUT`) — i.e. ends mid-block. Disconnect.
   - This is what seeds `blocksToResend` with `N+1` server-side.
4. Wait for the BN's backfill to pull and persist blocks `N+1..N+M` from the peer (verify via `serverStatus` / metrics that `lastPersisted` advanced past `N+1`).
5. Connect **Publisher B**. Stream a complete block at `N+M+1` (header → items → proof → `BlockEnd`).
6. **Bug observed:** instead of `Acknowledgement`, the BN sends `ResendBlock(N+1)` — a block that has long since been persisted. Publisher B replies with `EndStream(TOO_FAR_BEHIND, earliest=…, latest=N+M+1)` and disconnects.
7. Reconnect any publisher and stream the next current block — the BN keeps responding with `ResendBlock(N+1)` indefinitely. Backfill keeps closing the small gaps each cycle, so block height continues to advance even though every publisher is being kicked.

Equivalent seeding alternative for step 3: have Publisher A stream a complete block at `N+1` with a deliberately invalid signature — failed verification of a `PUBLISHER`-sourced block also adds the block to `blocksToResend`.

## Root cause location

- Add sites: L256 (`blockIsEnding`), L336 (`handleVerification`), L580 (`checkForStalledHandlers`)
- Remove site: L726 only (`getActionForHeader` — requires a publisher to send that exact block)
- `nextBlockToResend()` (L706) returns `blocksToResend.first()` with no `lastPersistedBlockNumber` filter
- `handlePersisted` never prunes `blocksToResend`

## Suggested fix

- In `handlePersisted`, prune `blocksToResend` entries `<=` the newly acknowledged block.
- Defensive guard in `nextBlockToResend()` to skip/remove entries `<= lastPersistedBlockNumber`.

AlfredoG87:
## Summary

Recovers the publisher pipeline from the stale-resend stuck state observed on BNCE `s02`. An entry in `blocksToResend` at or below `lastPersistedBlockNumber` causes every fresh publisher's first `endOfBlock` to return `RESEND(<old block>)`; the publisher disconnects with `EndStream(TOO_FAR_BEHIND)` and the cycle repeats. `LiveStreamPublisherManager.handlePersisted` now drops entries beyond a configurable buffer of blocks behind the proposed last-persisted — past that depth, no live publisher can supply the block, so backfill owns the gap.

## Changes

- `PublisherConfig.staleResendPruneBuffer` — new property, default `100`, min `0`.
- `LiveStreamPublisherManager.pruneStaleResendEntries(proposedPersistedBlock)` — drops `blocksToResend` entries `<= proposedPersistedBlock - staleResendPruneBuffer`. Called from `handlePersisted` before `correctForResendAndStreaming`. Emits a `WARNING` log with the dropped block numbers when it actually fires.
- `docs/block-node/configuration.md` — new `PRODUCER_STALE_RESEND_PRUNE_BUFFER` row.
- `PublisherManagerRegressionTest.testStaleResendEntryPrunedWhenPersistenceMovesPastBuffer` — drives the entry into `blocksToResend` via the real publish API (`sendHeaderOnly` + `EndStream(RESET)` mid-block) and asserts `getActionForBlock(<stale block>)` returns `END_DUPLICATE` after `handlePersisted(<past buffer>, BACKFILL)`. Fails without the prune.

## Test Plan

- [x] Regression test reproduces the production symptom and fails without the prune.
- [x] `./gradlew :stream-publisher:test` — 460 pass, 1 unrelated skip.
- [x] `./gradlew :stream-publisher:spotlessApply` clean.

Closes #2714


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
```diff
@@ -14,11 +14,17 @@
...
+/// entries older than the buffer are dropped to avoid asking publishers to resend a
+/// block they cannot supply (which kills the connection with TOO_FAR_BEHIND).
 @ConfigData("producer")
 public record PublisherConfig(
         // spotless:off
         @Loggable @ConfigProperty(defaultValue = "9_223_372_036_854_775_807") @Min(100_000L) long batchForwardLimit,
         @Loggable @ConfigProperty(defaultValue = "300") @Min(0L) long publisherUnavailabilityTimeout,
-        @Loggable @ConfigProperty(defaultValue = "3") @Min(3) @Max(50) int MaxFutureBlocksBeforeStalled) {
+        @Loggable @ConfigProperty(defaultValue = "3") @Min(3) @Max(50) int MaxFutureBlocksBeforeStalled,
+        @Loggable @ConfigProperty(defaultValue = "100") @Min(0L) long staleResendPruneBuffer) {
```
jsync-swirlds: It might be good to add `@Max(200)` here, we shouldn't set terribly high.

jsync-swirlds (dismissed):

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
AlfredoG87: ✅ Good one, just added it!

jsync-swirlds (approved):

Nana-EC (approved):

PR #2718, opened 2026-05-07
feat: add NodeAddressBook state loading/persistence to block node

### LINKED ISSUE #2561: Add new plugin to acquire/build/load the network roster config

Nana-EC:
## Summary
Establishes `NodeAddressBook` (the RSA public-key roster used for WRB proof verification) as a first-class, persisted piece of application state in the Block Node, following the same two-phase load-and-drain lifecycle already used for `TssData`.

This PR is a **prerequisite for #2683** (`RsaRosterBootstrapPlugin`). It provides the SPI surface and the `BlockNodeApp` machinery that the plugin will call at startup; the plugin itself lives in the dependent PR.

Key additions:
- Extends `BlockNodeContext` and `ApplicationStateFacility` SPI with `nodeAddressBook` support
- Implements full lifecycle in `BlockNodeApp`: load from disk → validate → queue → drain → rebuild context → fan-out `onContextUpdate` → persist atomically
- Persists the address book as **JSON** (PBJ `NodeAddressBook.JSON` codec) so operators can inspect and edit the file without external tooling — consistent with how `TssData` is stored
- Fixes a proto typo (`node_addess_book` → `node_address_book`) and exposes the address book in the `serverStatusDetail` gRPC endpoint
- Adds `doInit`/`doStart` split and `simulatePreloadedAddressBook` helper to `PluginTestBase` for testing plugins that depend on a pre-loaded address book

---

### SPI extensions — `block-node/spi-plugins`

**`BlockNodeContext`**
- Added `nodeAddressBook` as the 11th component (`NodeAddressBook`, nullable)
- `context.nodeAddressBook()` returns `null` until the address book is loaded or pushed by a plugin
- All existing constructor call-sites across the codebase updated with the new trailing `null` argument

**`ApplicationStateFacility`**
- Added `updateAddressBook(NodeAddressBook book)` — allows plugins (e.g. `RsaRosterBootstrapPlugin`) to push a freshly-fetched address book into the context from their `start()` method
- Default implementation logs a WARNING rather than silently discarding the call, so non-`BlockNodeApp` implementations of the SPI fail visibly

### App changes — `block-node/app`

**`BlockNodeApp`**
- `ApplicationStateConfig`: added `rsaBootstrapFilePath` config property (`app.state.rsaBootstrapFilePath`, default `/opt/hiero/block-node/node/rsa-bootstrap-roster.json`)
- `loadApplicationState()` (runs synchronously inside `startApplicationStateFacility()`, before `startPlugins()`):
  - Reads `rsa-bootstrap-roster.json` if present and parses it with `NodeAddressBook.JSON.parse()`
  - Calls `validateAddressBook()` — requires ≥1 entry with a non-blank `rsaPubKey`; throws `IllegalStateException` with a "delete and restart" message on corrupt files
  - Stages the validated book in `pendingAddressBook` (an `AtomicReference`) for draining by the scanner
- `checkForApplicationStateUpdates()` (the periodic scanner, also called synchronously before plugins start):
  - Drains `pendingAddressBook` alongside the `TssData` queue in a single pass
  - Rebuilds `blockNodeContext` when either update is present, then fans `onContextUpdate` to all plugins
  - Triggers `persistNodeAddressBook()` when the address book was updated
- `updateAddressBook()` override (called by plugins at runtime):
  - Validates the book before queuing — silently rejects null, empty, or all-blank-key books with a WARNING log
  - Stages valid books in `pendingAddressBook` (last-write-wins semantics)
- `persistNodeAddressBook()`:
  - Serialises with `NodeAddressBook.JSON.toBytes()` for human-readable output
  - Atomic write: writes to a `.tmp` sibling, then `Files.move(ATOMIC_MOVE)`; catches `AtomicMoveNotSupportedException` and falls back to `REPLACE_EXISTING` with a WARNING log
- `blockNodeContext` field made `volatile` to ensure scheduler-thread writes are visible to all plugin threads without additional synchronisation
- `validateAddressBook()`: package-visible static helper, reusable by tests

**`block-node/app/build.gradle.kts`**
- Fixed module names in `testModuleInfo`: `archive-cloud-storage` → `cloud-storage-archive`, added `cloud-storage-expanded`
- Added `copyRsaTestFixture` Gradle task (copies the JSON fixture to `build/tmp/data/config/` for writable round-trip tests)

### Proto + endpoint changes

**`protobuf-sources/src/main/proto/block-node/api/node_service.proto`**
- Fixed typo: field `node_addess_book = 4` → `node_address_book = 4` in `ServerStatusDetailResponse`
- PBJ sources regenerated accordingly

**`ServerStatusServicePlugin.serverStatusDetail()`**
- Added `.nodeAddressBook(context.nodeAddressBook())` to the response builder
- When `null` (address book not yet loaded), PBJ omits the field from the wire response

### Test infrastructure — `block-node/app` (testFixtures)

**`PluginTestBase`**
- Added `doInit()` — initialises context and calls `plugin.init()` without starting the plugin
- Added `doStart()` — starts the historical block facility and the plugin; call after `doInit()`
- Added `simulatePreloadedAddressBook(NodeAddressBook)` — rebuilds context with the book and calls `plugin.onContextUpdate()`, simulating what `BlockNodeApp.loadApplicationState()` does between init and start
- All existing `start()` overloads chain to the new split methods; no call-site changes required

---

## Test coverage

|---|---|
| `BlockNodeAppTest.validateAddressBook*` | Empty book throws; all-blank-key book throws; valid book accepted |
| `BlockNodeAppTest.updateAddressBook*` | `null` input is a no-op; empty/all-blank-key books rejected with WARNING; valid book queued |
| `BlockNodeAppTest.loadApplicationState*` | Missing file → address book remains `null`; corrupt file → `IllegalStateException` |
| `BlockNodeAppTest.updateAddressBookQueuesForScanner` | Scanner picks up queued book, rebuilds context, calls `onContextUpdate` on plugins |
| `BlockNodeAppTest.addressBookPersistenceRoundTrip` | Book persisted by `updateAddressBook` is reloaded by a fresh `BlockNodeApp` |
| `ServerStatusDetailServicePluginTest` | `nodeAddressBook` absent when `null`; present with correct entries after context update |

---

## Startup sequence (as implemented)

```
BlockNodeApp.start()
  └─ startApplicationStateFacility()
       ├─ loadApplicationState()
       │    ├─ Load TssData from JSON file (if present) → tssDataUpdates queue
       │    └─ Load rsa-bootstrap-roster.json (if present)
       │         ├─ NodeAddressBook.JSON.parse()
       │         ├─ validateAddressBook() — ≥1 non-blank RSA key required
       │         └─ pendingAddressBook.set(book)
       │
       └─ checkForApplicationStateUpdates()  ← called synchronously here
            ├─ drain tssDataUpdates queue
            ├─ drain pendingAddressBook
            ├─ rebuild blockNodeContext (tssData + nodeAddressBook)
            ├─ plugin.onContextUpdate(ctx) for all loaded plugins
            └─ persistNodeAddressBook() / persistTssData()
       └─ schedule periodic checkForApplicationStateUpdates()
  └─ startPlugins()
       └─ RsaRosterBootstrapPlugin.start()  [lives in PR #2683]
            ├─ context.nodeAddressBook() != null?
            │    YES → book already loaded; record metrics, return
            │    NO  → fetch Mirror Node; call updateAddressBook(book)
            │              └─ queued, picked up on next scan tick
```

---

## Reviewer Notes
- **Relationship to #2683**: This PR contains only the infrastructure layer — `BlockNodeApp` state machinery, SPI extensions, and `PluginTestBase` helpers. The `RsaRosterBootstrapPlugin` that calls `updateAddressBook()` at runtime is in #2683, which must be rebased onto this branch.
- **File format**: The address book is stored as JSON (PBJ `NodeAddressBook.JSON`), not binary protobuf. The JSON uses PBJ's camelCase field names (`nodeAddress`, `RSAPubKey`, `nodeId`). The field name `nodeAddress` is dictated by PBJ code generation from the proto schema and cannot be changed without a proto rename.
- **Fail-fast on corrupt file**: A corrupt or unreadable RSA bootstrap file causes `startApplicationStateFacility()` to throw `IllegalStateException` at startup. The recommended operator action is to delete the file and restart; the dependent plugin (#2683) will re-fetch from the Mirror Node.
- **Last-write-wins `updateAddressBook`**: If a plugin calls `updateAddressBook()` multiple times before the scanner drains, only the most recent valid book is used. This is intentional and documented in the method Javadoc.
- **Peer BN querying** (file → peer BNs → Mirror Node) is tracked separately in #2682.

## Related Issue(s)
Closes #2561


block-node/app/build.gradle.kts
```diff
@@ -213,10 +213,22 @@ testModuleInfo {
...
 
     exportsTo("com.swirlds.config.impl")
 }
 
+// Copy the RSA bootstrap fixture into build/tmp so tests can read from (and write to) a
+// writable path that is consistent with other test data paths (appStateDataFilePath etc.).
+val copyRsaTestFixture =
+    tasks.register<Copy>("copyRsaTestFixture") {
+        description = "Copies the RSA bootstrap test fixture into build/tmp/data/config/"
+        from(layout.projectDirectory.file("src/test/resources/data/config/rsa-bootstrap-roster.pb"))
```
rockysingh, line 226: should this be the json file?

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -15,15 +15,24 @@
...
+ *     Configured via {@code app.state.rsaBootstrapFilePath}.
+ *     Defaults to {@code /opt/hiero/block-node/node/rsa-bootstrap-roster.json}.
  * @param updateScanInterval The amount of milliseconds that the {@code ApplicationStateFacility} waits between
  *     checking to see if there are any {@code TssData} updates to process.
  * @param updateInitialDelay The amount of milliseconds that the {@code ApplicationStateFacility} waits before
  *     starting the scan interval.
  */
-// spotless:off
 @ConfigData("app.state")
 public record ApplicationStateConfig(
```
rockysingh, line 27: I think its worth considering adding a startup check that the parent directory is writable, similar to how other file paths might be validated. This would catch permission issues early rather than at first persist attempt.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -378,6 +385,26 @@ public void updateTssData(TssData tssData) {
...
+     * @param nodeAddressBook the NodeAddressBook to store in BlockNodeContext
+     */
+    @Override
+    public void updateAddressBook(NodeAddressBook nodeAddressBook) {
+        if (nodeAddressBook == null) return;
+        try {
+            validateAddressBook(nodeAddressBook, "runtime update");
+        } catch (IllegalStateException e) {
+            LOGGER.log(WARNING, "Rejecting invalid address book update: {0}", e.getMessage());
+            return;
```
rockysingh, line 403: Either throw the exception or return a boolean/result object indicating success/failure so plugins can take corrective action. I'd prefer boolean tbf.

I think the issue with this is this could lead to plugins thinking they've successfully updated the address book when they haven't.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -82,14 +85,17 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     private final HistoricalBlockFacilityImpl historicalBlockFacility;
     /** Should the shutdown() method exit the JVM. */
     private final boolean shouldExitJvmOnShutdown;
-    /** The block node context. Package so accessible for testing. */
-    BlockNodeContext blockNodeContext;
+    /** The block node context. Volatile: written by the scheduled scanner thread, read by plugin threads. */
+    volatile BlockNodeContext blockNodeContext;
```
rockysingh, line 89: Do we need to document this is thread safety?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -514,5 +602,49 @@ private void loadApplicationState(Configuration configuration) {
...
+     * @param book the address book to validate
+     * @param source human-readable source name for error messages
+     * @throws IllegalStateException if the book is empty or has no usable entries
+     */
+    static void validateAddressBook(final NodeAddressBook book, final String source) {
+        if (book.nodeAddress().isEmpty()) {
+            throw new IllegalStateException(
+                    "RSA address book from " + source + " contains no entries — cannot verify WRB proofs");
+        }
+        final long usable = book.nodeAddress().stream()
```
rockysingh, line 642: Do we need to verify the key is valid format too? 

As right now a malformed key would pass validation but fail at proof verification time etc.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -413,20 +444,33 @@ void startApplicationStateFacility() {
...
 
-        if (updated) {
+        // Consume any address book cached at startup (one-shot).
+        boolean addressBookUpdated = false;
+        final NodeAddressBook addressBook = pendingAddressBook.getAndSet(null);
+        if (addressBook != null) {
+            addressBookUpdated = updateBlockNodeContext(null, addressBook);
+        }
+
+        if (tssUpdated || addressBookUpdated) {
```
rockysingh, line 463: The logic rebuilds context only if either is updated, but tracks them separately. However, `updateBlockNodeContext` is called twice with partial data, which is confusing.

jsync-swirlds: We should correct that in future work, but for now we don't have the structure needed to correctly defer the updates between multiple method calls for different updates.

block-node/app/build.gradle.kts
Nana-EC: Yes, good catch

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
Nana-EC: Yeah, I believe provisioner might do this but let me check.
We should add it as part of our docs.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: Yeah agreed, forgot to change this. Updating shortly

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: To what extent are you suggesting?
The key format is validated as part of verification, in that it'll fail.
However, I agree with a fail fast approach here if possible.

jsync-swirlds (approved):
Some of the logging statements could use adjustments to level and structure, and @rockysingh has some good points to address.

This can merge and those items be addressed in follow-up, however.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: Hmm, good note. 
I think we could make the call once and respond with an object that has a field for each type of state that could  be updated. Or we could try an enum with some bitwise logic. 🤔 
There's option but would prefer it be covered in follow up for improvement.

jsync-swirlds: I was thinking more along the lines of adding any update to a thread to gather updates for a short time (perhaps 100ms or so) then make a single context update when the time expires.

It's important that context remain a record to avoid needing to add multiple global locks.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
rockysingh: Just a fail safe check.

rockysingh (approved):

PR #2719, opened 2026-05-07
feat: implement RsaRosterBootstrapPlugin for WRB address book loading…

### LINKED ISSUE #2561: Add new plugin to acquire/build/load the network roster config

Nana-EC:
Implements `RsaRosterBootstrapPlugin` ([#2561](https://github.com/hiero-ledger/hiero-block-node/issues/2561)) — the plugin that loads the consensus-node RSA public-key address book at Block Node startup and publishes it to the proof-verification layer via `BlockNodeContext.nodeAddressBook()`.

> **Prerequisite context:** `BlockNodeContext.nodeAddressBook()`, `ApplicationStateFacility.updateAddressBook()`, and `BlockNodeApp.loadApplicationState()` were added in prior PRs and are already on `main`. This PR implements the plugin that consumes those extension points.

---

## What changed

**New module: `roster-bootstrap-rsa`**

- **`RsaRosterBootstrapPlugin`** — on `start()`, checks whether `BlockNodeApp` already loaded the address book from `rsa-bootstrap-roster.json` (via `context.nodeAddressBook()`). If not, queries `GET /api/v1/network/nodes` on the configured Mirror Node with pagination and exponential-backoff retry (up to 3 attempts), then calls `applicationStateFacility.updateAddressBook()` to persist the result and broadcast `onContextUpdate` to all plugins. If `mirrorNodeBaseUrl` is blank and no file is present, logs a WARNING and returns early without throwing.

- **`RsaRosterBootstrapConfig`** (`@ConfigData("roster.bootstrap.rsa")`) — configuration record for Mirror Node URL, timeouts, and page size. The `mirrorNodeBaseUrl` default is intentionally blank — operators must set it explicitly; no testnet URL is hard-coded.

- **`MirrorNodeNodesResponse`** — lightweight Gson-based JSON parser for the Mirror Node `/api/v1/network/nodes` response, including `timestamp.from` / `timestamp.to` fields and `links.next` for pagination.

**Active-entry filtering**

Mirror Node returns all historical node configuration records alongside current ones. The plugin queries with `order=desc` so active entries (`timestamp.to == null`) arrive first. Pagination stops on the first superseded entry (`timestamp.to != null`) — all remaining entries in descending order are also historical, so subsequent pages are never fetched.

**Async-delivery note**

When the Mirror Node fetch path is taken, `applicationStateFacility.updateAddressBook()` queues the book and returns before the context is rebuilt. The `NodeAddressBook` becomes visible to other plugins only after the `applicationStateExecutor` fires its next scan tick. Any plugin that relies on `context.nodeAddressBook()` must implement `onContextUpdate` and null-guard during its own `start()`.

---

## Tests

- **`RsaRosterBootstrapPluginTest`** — 11 tests:
  - Pre-loaded bootstrap file path (address book already in context)
  - Blank `mirrorNodeBaseUrl` early-return
  - Unreachable Mirror Node → `IllegalStateException` fail-fast after 3 retries
  - Embedded HTTP server: single-page response, multi-page pagination, blank/null key skipping, `0x` prefix stripping, HTTP 500 retry
  - Mixed active + historical entries — verifies pagination stops after first superseded entry (page 2 never fetched)

- **`MirrorNodeNodesResponseTest`** — 8 tests covering JSON parsing, timestamp field extraction, blank/null key handling, and relative `links.next` resolution.

---

## Configuration

| Key | Default | Notes |
|-----|---------|-------|
| `roster.bootstrap.rsa.mirrorNodeBaseUrl` | *(blank)* | Set to your Mirror Node base URL; blank disables Mirror Node fallback |
| `roster.bootstrap.rsa.mirrorNodeConnectTimeoutSeconds` | `5` | TCP connect timeout |
| `roster.bootstrap.rsa.mirrorNodeReadTimeoutSeconds` | `10` | Per-request read timeout |
| `roster.bootstrap.rsa.mirrorNodePageSize` | `100` | Nodes per paginated request (max 100) |

> **Note:** `docs/block-node/configuration.md` does not yet document the `roster.bootstrap.rsa.*` namespace — tracked as a follow-up from review (ivannov).

---

## Known issues / follow-up

| # | Issue | Tracking |
|---|-------|---------|
| 1 | `InterruptedException` in `fetchAndParseWithRetry` is caught and re-thrown as `IllegalStateException` — an anti-pattern flagged by jsync. The outer `try` block should either enclose the entire method or `InterruptedException` should propagate naturally without wrapping. | Open review comment ([jsync](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3227448370)) |
| 2 | The `fetchAndParseWithRetry` exception-as-flow-control structure should be refactored — either extracted into a dedicated method or the try block should span the full method scope. | Open review comment ([jsync](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3227541642)) |
| 3 | `MirrorNodeNodesResponse.parse()` uses manual Gson traversal — simplifiable to direct `gson.fromJson()` deserialization with a nested `Links` record. | [#2762](https://github.com/hiero-ledger/hiero-block-node/issues/2762) |

---
## Related Issue(s)
Closes #2561


block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/MirrorNodeNodesResponse.java
```diff
@@ -0,0 +1,124 @@
...
+    /// }
+    /// ```
+    ///
+    /// @param json the raw JSON string from the Mirror Node API
+    /// @return parsed response containing node entries and the next-page link
+    static MirrorNodeNodesResponse parse(final String json) {
+        final List<NodeEntry> entries = new ArrayList<>();
+        String nextLink = null;
+
+        final JsonObject root = JsonParser.parseString(json).getAsJsonObject();
```
berryware, line 86: Should we just let GSon handle mapping to the reduced fields. Only the fields in the POJO are included for deserialization. All other fields are ignored. If the fields change over time, we just update the MirrorNodeNodesResponse.

MirrorNodeNodesResponse response = gson.fromJson(json, MirrorNodeNodesResponse.class);

Nana-EC: Good note, let me look into it,

berryware: If you want, add this to the followup, and I will do it.

Nana-EC: @berryware good consideration. It is possible but it seems it may require quite a bit more code to support not less
- The nested `links` object would require a new record to map to it and an update to `MirrorNodeRepsonse` to use it
- We'd  like need additional gson attributes like @SerializedName to manage records
- potential modifications to the module-info directions to allow gson to utilize reflection to parse the inner details

It seems doable but could be overkill and add additional overhead to manage.
We can open a ticket to explore as a follow up item if desired?

Nana-EC: @berryware  captured in https://github.com/hiero-ledger/hiero-block-node/issues/2762, but ticket needs more details

berryware (approved):
Merge time.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapConfig.java
```diff
@@ -0,0 +1,24 @@
...
+/// `ApplicationStateConfig`; file loading and persistence are handled by `BlockNodeApp`.
+///
+/// @param mirrorNodeBaseUrl base URL of the Mirror Node REST API used when no local file is present;
+///     leave blank to disable Mirror Node fallback (a WARNING is logged at startup if blank and no
+///     bootstrap file is present)
+/// @param mirrorNodeConnectTimeoutSeconds TCP connect timeout when calling the Mirror Node
+/// @param mirrorNodeReadTimeoutSeconds per-request read timeout when calling the Mirror Node
+/// @param mirrorNodePageSize number of nodes requested per paginated Mirror Node call
+@ConfigData("roster.bootstrap.rsa")
+public record RsaRosterBootstrapConfig(
```
ivannov, line 20: Nit: describe this config in `docs/block-node/configuration.md`

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,251 @@
...
+        long delayMs = 1_000L;
+        Exception lastCause = null;
+        for (int attempt = 0; attempt < MAX_RETRIES; attempt++) {
+            if (attempt > 0) {
+                LOGGER.log(WARNING, "Mirror Node fetch retry {0}/{1}: {2}", attempt, MAX_RETRIES - 1, url);
+                try {
+                    Thread.sleep(delayMs);
+                } catch (InterruptedException ie) {
+                    Thread.currentThread().interrupt();
+                    throw new IllegalStateException("Interrupted while waiting for Mirror Node retry", ie);
```
ivannov, line 221: Other colleagues have more experience with interrupted exceptions handling. But I am not sure whether we should catch it and wrap it. If we want to log something (like "Interrupted while waiting for Mirror Node retry"), then maybe re-throw that

My understanding is that it is the upper-most code in the chain (in this case the start method) that should handle interruptions.

jsync-swirlds: Good point.
Catching interrupted exception like this is not advisable.  Interruption is not an illegal state, it's normal, and we should either cleanly (non-exceptionally) exit the current method, or not catch the `InterruptedException` in the first place (let it throw and only catch somewhere further up the call stack).

In this case, we should not even need this try block, suggestion offered separately.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,251 @@
...
+                }
```
jsync-swirlds, line 222: This (wrap sleep directly and throw a new exception) is an anti-pattern, we should not do this.
Either move the try block to enclose the entire method (which also fixes the poor exception handling in _another_ try block later in this same method), do not catch the exception, or use an alternative that does not throw unnecessarily.

This suggestion is for the third option.
Note, this suggestion is approximate.

```suggestion
                LockSupport.parkNanos(delayMs * 1_000_000L);
                if (Thread.currentThread().isInterrupted()) {
                    break;
                }
```

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,251 @@
...
+                if (response.statusCode() != 200) {
+                    throw new IOException("HTTP " + response.statusCode() + " from " + url);
+                }
+                return MirrorNodeNodesResponse.parse(response.body());
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+                lastCause = ie;
+            } catch (IOException | RuntimeException e) {
+                lastCause = e;
+            }
```
jsync-swirlds, line 241: This is a code smell.
Either this block wants to be a separate method, or the try block should enclose this entire method.
First, we're throwing an exception as flow-control rather than handling a status return; that's not ideal.
Second, at the moment, we're effectively catching exceptions (including the one we _just threw_!) just to record them for later, so I suspect this block really should be a method that returns a code (not an exception) indicating success or failure (and let's process return codes normally instead of transforming it into an exception, please).

The final throw below (of an illegal state exception, which is completely misused in this method) just wraps the exception generated here, which is another indication that we're using an exception in lieu of return codes and "normal" flow control.

block-node/roster-bootstrap-rsa/src/main/java/module-info.java
```diff
@@ -1,22 +1,32 @@
...
+            com.swirlds.config.impl,
+            com.swirlds.config.extensions,
+            org.hiero.block.node.app;
+
+    requires transitive com.swirlds.config.api;
     requires transitive org.hiero.block.node.spi;
+    requires org.hiero.block.node.base;
+    requires org.hiero.block.protobuf.pbj;
+    requires org.hiero.metrics;
+    requires com.google.gson;
```
jsync-swirlds, line 28: We shouldn't be using `GSON` at all.  We can use PBJ or we can look at other alternatives if necessary, but GSON shouldn't be used in production code.
This isn't _immediately_ blocking, but it's a high priority to replace.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,251 @@
...
+                        "roster.bootstrap.rsa.mirrorNodeBaseUrl is blank and no RSA bootstrap file is present."
+                                + " Provide rsa-bootstrap-roster.json or set roster.bootstrap.rsa.mirrorNodeBaseUrl.");
+                return;
+            }
+            LOGGER.log(INFO, "RSA bootstrap file not found, querying Mirror Node at {0}", config.mirrorNodeBaseUrl());
+            book = fetchFromMirrorNode();
+            // BlockNodeApp.updateAddressBook() persists the file and calls onContextUpdate.
+            applicationStateFacility.updateAddressBook(book);
+            addressBookSource = "Mirror Node";
+        }
```
jsync-swirlds, line 129: For follow-up, we should query mirror node even if we found a local file.
If mirror node is newer, then we should use the version that is valid for the current time (or the newest if multiple are valid at the current time).

jsync-swirlds (commented):
We can merge this if necessary to continue development, but there are concerning code quality issues that require follow-up.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,251 @@
...
+            String nextUrl = config.mirrorNodeBaseUrl() + "/api/v1/network/nodes?limit=" + config.mirrorNodePageSize()
+                    + "&order=desc";
+
+            outer:
+            while (nextUrl != null) {
+                final MirrorNodeNodesResponse response = fetchAndParseWithRetry(client, nextUrl);
+                for (final MirrorNodeNodesResponse.NodeEntry entry : response.nodes()) {
+                    // A non-null timestamp.to means this entry has been superseded — all subsequent
+                    // entries (in descending order) are also historical; stop here.
+                    if (entry.timestamp() != null && entry.timestamp().to() != null) {
```
rockysingh, line 168: entry.timestamp().to() could cause null pointer.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,251 @@
...
+    private NodeAddressBook fetchFromMirrorNode() {
+        try (final HttpClient client = HttpClient.newBuilder()
+                .connectTimeout(Duration.ofSeconds(config.mirrorNodeConnectTimeoutSeconds()))
+                .build()) {
+
+            final List<NodeAddress> addresses = new ArrayList<>();
+            // Use order=desc so the most-recently-active entries (timestamp.to == null) arrive first.
+            // We stop pagination as soon as we encounter a superseded entry (timestamp.to != null),
+            // since all remaining entries in descending order are also historical.
+            String nextUrl = config.mirrorNodeBaseUrl() + "/api/v1/network/nodes?limit=" + config.mirrorNodePageSize()
```
rockysingh, line 159: The code requests order=desc and assumes the first superseded entry means all remaining are also historical. But if Mirror Node has a bug or returns entries out-of-order, active nodes could be skipped.

I assume this is wont happen. I thought i'd call it out.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -1,64 +1,251 @@
...
+                        : rawNext.startsWith("http") ? rawNext : config.mirrorNodeBaseUrl() + rawNext;
+            }
+
+            if (addresses.isEmpty()) {
+                throw new IllegalStateException(
+                        "Mirror Node returned zero nodes with a non-blank public_key from " + config.mirrorNodeBaseUrl()
+                                + ". Provide rsa-bootstrap-roster.json or ensure Mirror Node is reachable.");
+            }
+
+            return NodeAddressBook.newBuilder().nodeAddress(addresses).build();
```
rockysingh, line 198: Do we need to validate adddressbook? Probably not.

rockysingh (approved):
lgtm - lets follow up with pr to address comments raised by the team.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapConfig.java
Nana-EC: To be address in https://github.com/hiero-ledger/hiero-block-node/issues/2797

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
Nana-EC: To be address in https://github.com/hiero-ledger/hiero-block-node/issues/2797

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
Nana-EC: To be address in https://github.com/hiero-ledger/hiero-block-node/issues/2797

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
Nana-EC: To be address in https://github.com/hiero-ledger/hiero-block-node/issues/2797

block-node/roster-bootstrap-rsa/src/main/java/module-info.java
Nana-EC: To be address in https://github.com/hiero-ledger/hiero-block-node/issues/2797

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
Nana-EC: To be addressed in https://github.com/hiero-ledger/hiero-block-node/issues/2797

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
Nana-EC: To be addressed in https://github.com/hiero-ledger/hiero-block-node/issues/2797

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
Nana-EC: To be addressed in https://github.com/hiero-ledger/hiero-block-node/issues/2797

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
Nana-EC: We should.
I have logic in the BlockNodeApp but maybe we should have it here also or rely on that for loading.
To be addressed in https://github.com/hiero-ledger/hiero-block-node/issues/2797

Nana-EC:
All comments to be addressed in follow up by https://github.com/hiero-ledger/hiero-block-node/issues/2797

PR #2720, opened 2026-05-07
feat: implement RSA SignedRecordFileProof verification for WRB blocks…

### LINKED ISSUE #2562: Add logic to verify WRBs using SignedRecordFileProof w 2/3 + 1 signature count
### Story Form

As a Block Node,
I want to verify incoming Wrapped Record Blocks by checking that at least 2/3 + 1 of the current roster nodes have produced valid RSA signatures in the `SignedRecordFileProof`,
So that only cryptographically authenticated WRBs are accepted and persisted.

---

### Background

In Phase 2a of the v3 rollout, every WRB streamed from a Consensus Node carries a `SignedRecordFileProof` as its sole block proof. This proof contains one RSA signature per roster node, produced over the record file payload. The BN must verify this proof before persisting the block — the same threshold check that Mirror Nodes and the WRB CLI perform today.

The threshold is **2/3 + 1 of the total roster node count** (count-based, not stake-weighted — per the issue title). A block with fewer valid signatures than this threshold must be rejected.

This work lives in the existing verification module (`block-node/verification`). The current pipeline handles TSS-BLS block proofs via `ExtendedMerkleTreeSession`. This ticket adds a parallel verification path for `SignedRecordFileProof` RSA proofs. The exact algorithm, failure behaviour, and session/dispatch design are to be specified in the design document ([#2560](https://github.com/hiero-ledger/hiero-block-node/issues/2560)). The notes below are orientation.

> **Important:** Do not begin implementation until [#2560](https://github.com/hiero-ledger/hiero-block-node/issues/2560) is approved and [#2561](https://github.com/hiero-ledger/hiero-block-node/issues/2561) is merged. The roster loaded by #2561 is a hard input to this work, however, if required the roster file structure can be relied on from [#2560](https://github.com/hiero-ledger/hiero-block-node/issues/2560) allowing this and [#2561](https://github.com/hiero-ledger/hiero-block-node/issues/2561) to be worked on in parallel.

---

### Where this fits in the existing pipeline

Verification is synchronous in the ingest path:

```
StreamPublisherPlugin
  → BlockMessagingFacility.sendBlockItems()
  → VerificationServicePlugin.handleBlockItemsReceived()   ← this is the entry point
    → VerificationSession.processBlockItems()              ← where proof type dispatching happens
      → [TSS path] ExtendedMerkleTreeSession.verifySignature()
      → [WRB path] ??? ← this ticket adds this path
    → context.blockMessaging().sendBlockVerification(VerificationNotification)
```

A `VerificationNotification(success=false, ...)` propagates upstream to close the publisher stream with `EndOfStream(BAD_BLOCK_PROOF)`.

### Key implementation areas

1. **Proof type detection and dispatch** — `ExtendedMerkleTreeSession.finalizeVerification()` currently extracts `BlockProof.signedBlockProof()` (TSS). It needs to also handle the case where the proof is a `SignedRecordFileProof`. The design doc will specify whether this is a branching path inside the existing session, a new `WrbVerificationSession`, or another approach. Either way, there should be a single, clearly located dispatch point.

2. **Session creation** — `HapiVersionSessionFactory` selects a `VerificationSession` implementation based on HAPI version. It may need a configuration flag or proof-mode signal to know when to produce a WRB-aware session rather than a TSS session. The design doc will specify this; the factory is the right place to make that decision.

3. **RSA signature verification** — The `SignedRecordFileProof` contains a list of `RecordFileSignature` entries (node ID + RSA signature bytes). For each entry, retrieve the RSA public key for that node ID from the roster loaded by [#2561](https://github.com/hiero-ledger/hiero-block-node/issues/2561) and verify the signature over the well-defined payload. The payload definition and the exact verification algorithm must match the Mirror Node and WRB CLI implementations — confirmed in [#2560](https://github.com/hiero-ledger/hiero-block-node/issues/2560).

4. **Threshold enforcement** — Count the number of valid signatures. If valid count < ⌈(2/3 × roster size) + 1⌉, produce `VerificationNotification(success=false, ...)`. Do not persist partial results. Handle edge cases: unknown node ID in a signature entry, duplicate node ID, signature over the wrong payload.

5. **Block hash continuity** — WRBs still participate in the block hash chain (`previousBlockHash`, `rootOfAllPreviousBlockHashes`). The existing Merkle tree hashing infrastructure in `ExtendedMerkleTreeSession` and `AllBlocksHasherHandler` should remain active for WRBs — only the proof signature step changes. Confirm this with the design doc.

### Key files to read first

| File | Why |
|---|---|
| `block-node/verification/src/main/java/.../VerificationServicePlugin.java` | Entry point; understand the item handling and notification flow |
| `block-node/verification/src/main/java/.../session/impl/ExtendedMerkleTreeSession.java` | `finalizeVerification()` and `verifySignature()` — this is where the new RSA path branches |
| `block-node/verification/src/main/java/.../session/HapiVersionSessionFactory.java` | Session selection logic — likely needs a new branch for WRB mode |
| `block-node/spi-plugins/src/main/java/.../VerificationNotification.java` | The result record passed downstream on success or failure |
| `protobuf-sources/.../signed_record_file_proof.proto` | The proof type being verified |

---

## Acceptance Criteria

- [ ] A WRB with valid RSA signatures from ≥ 2/3 + 1 roster nodes is accepted and produces `VerificationNotification(success=true, ...)`
- [ ] A WRB with fewer than 2/3 + 1 valid signatures produces `VerificationNotification(success=false, ...)` and the block is not persisted
- [ ] A WRB with a zeroed, malformed, or wrong-payload signature on any entry is counted as invalid (not panicked over)
- [ ] A WRB with a signature from a node ID not in the loaded roster is counted as invalid
- [ ] Block hash chain continuity (`previousBlockHash`, `AllBlocksHasherHandler`) is maintained correctly for verified WRBs
- [ ] `blocks_verification_failed_total` metric increments on every rejection; `blocks_verified_total` increments on every acceptance
- [ ] Unit tests cover: valid proof at threshold, one below threshold, zeroed signature, unknown node ID, wrong proof type
- [ ] Integration test confirms the publisher stream is closed with `BAD_BLOCK_PROOF` when verification fails

---

## References

| Resource | Link |
|---|---|
| Design doc (prerequisite) | [hiero-ledger/hiero-block-node#2560](https://github.com/hiero-ledger/hiero-block-node/issues/2560) |
| Roster plugin (dependency) | [hiero-ledger/hiero-block-node#2561](https://github.com/hiero-ledger/hiero-block-node/issues/2561) |
| `ExtendedMerkleTreeSession` | `block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java` |
| `HapiVersionSessionFactory` | `block-node/verification/src/main/java/org/hiero/block/node/verification/session/HapiVersionSessionFactory.java` |
| `VerificationServicePlugin` | `block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java` |
| `VerificationNotification` | `block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java` |
| HIP-1193 (cutover mechanics) | [Record Stream to Block Stream Cutover](https://hips.hedera.com/hip/hip-1200#hip-1193) |

Nana-EC:
## Details
Implements Phase 2a of WRB (Wrapped Record Block) streaming support: RSA-based verification of
`SignedRecordFileProof` block proofs. When a block carries a `SignedRecordFileProof` instead of a
TSS proof, the Block Node now verifies it by counting RSA signatures from the consensus node roster
and checking that a supermajority signed the correct V6 payload.

This PR **depends on #2718** (`NodeAddressBook` state loading/persistence). It consumes the
`NodeAddressBook` delivered by that PR's infrastructure and applies it inside the verification
plugin. The `RsaRosterBootstrapPlugin` that populates the address book at startup lives in a
separate PR (#2683).

Key additions:
- New `RsaKeyDecoder` utility: decodes hex-DER RSA public keys from a `NodeAddressBook` into a
  `Map<Long, PublicKey>` keyed by `node_id`
- `VerificationServicePlugin` extended with `onContextUpdate()` to rebuild the key map when the
  address book changes, and three new per-session RSA metrics
- `ExtendedMerkleTreeSession` extended with the full RSA proof dispatch path: extract
  `record_file_contents` bytes from the `RecordFileItem` proto using direct wire-format navigation,
  compute `SHA-384(int32(6) || rawRecordStreamFileBytes)`, verify each `RecordFileSignature` with
  `SHA384withRSA`, deduplicate by `node_id`, and accept if `validCount ≥ 2 * ceil(N/3) + 1`
- 12 new tests in `RsaWrbVerificationTest` covering threshold boundary, deduplication,
  blank/unknown key handling, wrong payload, V5 rejection, and wire-format field extraction

---

### New class — `block-node/verification`

**`RsaKeyDecoder`**
- Decodes a `NodeAddressBook` into an immutable `Map<Long, PublicKey>`
- Keys expected as hex-encoded X.509 DER bytes (no `0x` prefix) — normalisation was done
  upstream by `RsaRosterBootstrapPlugin`
- Fail-soft: entries with a blank key or an invalid DER encoding are skipped with a WARN log so
  one bad key does not prevent the rest of the roster from being loaded
- Mirrors `SigFileUtils.decodePublicKey` in `tools-and-tests/tools`, inlined to avoid a
  cross-module import

### `VerificationServicePlugin`

- Added `volatile Map<Long, PublicKey> keyByNodeId` — written once at startup by `onContextUpdate`,
  read on the block-handling thread; `volatile` ensures cross-thread visibility without a lock
- `onContextUpdate(BlockNodeContext)`: rebuilds `keyByNodeId` via `RsaKeyDecoder.buildKeyMap()`
  when the delivered `NodeAddressBook` is non-null and non-empty; emits a WARNING when an empty
  book arrives so operators can detect misconfiguration
- Three new metrics registered in `init()` and passed through to each session:
  - `blocknode:rsa_verification_success_total` — WRB blocks whose RSA proof was accepted
  - `blocknode:rsa_verification_failure_total` — WRB blocks whose RSA proof was rejected
  - `blocknode:rsa_roster_mismatch_total` — signatures from `node_id` values absent from the
    loaded address book
- Both `handleBlockItemsReceived` and `handleBackfilled` pass `keyByNodeId` and the three metric
  measurements to `HapiVersionSessionFactory.createSession()`

### `HapiVersionSessionFactory`

- Added a 10-parameter overload of `createSession()` that accepts the RSA key map and three metric
  measurements and threads them into `ExtendedMerkleTreeSession`
- The original 6-parameter overload delegates to the new one with `Map.of()` and `null` for the
  RSA parameters — backward-compatible; no existing call sites broken

### `ExtendedMerkleTreeSession`

**New fields:**
- `rawRecordFileItemProtoBytes` (`Bytes`, nullable) — captures the raw bytes of the
  `RecordFileItem` block item during `processBlockItems`; field 2 of this proto is the
  `record_file_contents` required for hash computation
- `rsaKeyByNodeId`, `rsaVerificationSuccessTotal`, `rsaVerificationFailureTotal`,
  `rsaRosterMismatchTotal` — passed from the plugin, non-null `Map.of()` when no book is loaded

**New `ThreadLocal` engines:**
- `SHA384_WITH_RSA` (existing) — reusable `Signature` engine per thread
- `SHA_384` (new) — reusable `MessageDigest` per thread, avoids per-block `getInstance("SHA-384")`
  allocation on the verification hot path

**`finalizeVerification`:**
- RSA proof dispatch added before the existing TSS path; `getFirst()` is used instead of
  `getSingle()` because duplicates in the proof list are handled inside `verifyRsaProof()` rather
  than at the dispatch layer

**`verifyRsaProof()`:**
1. Computes the block root hash for chain continuity (identical computation as the TSS path)
2. Guards: address book must be non-empty; `RECORD_FILE` item must have been seen; version must be 6
3. Calls `extractRecordStreamFileBytes()` to isolate field-2 bytes from the raw proto
4. Computes `SHA-384(int32(6) || rawRecordStreamFileBytes)` using the `ThreadLocal` `MessageDigest`
5. Iterates `proof.recordFileSignatures()` with a `Set<Long> validatedNodes` for deduplication —
   a repeated `node_id` entry cannot inflate `validCount`
6. Skips signatures where `node_id` is absent from the key map or the signature bytes are all zeros
7. Accepts if `validCount >= 2 * ceil(rosterSize / 3) + 1`

**`extractRecordStreamFileBytes()`:**
- Walks the protobuf wire format of the `RecordFileItem` message directly to extract field-2 bytes
- Full `PROTOBUF.parse()` is intentionally avoided: re-serialising a parsed object can produce
  subtly different bytes (omitted default fields, different varint encodings) which would break
  the hash comparison against the consensus node's signature
- Each field is identified by `(tag >>> 3, tag & 0x7)` and either returned (field 2, LEN) or
  skipped by wire type (VARINT/I64/LEN/I32); an unknown wire type or any `RuntimeException` returns
  `Bytes.EMPTY` so the caller fails verification cleanly

**Other:**
- `getFirst()` made `private` (was mistakenly `public static`; no external callers)
- `rawRecordFileItemBytes` renamed to `rawRecordFileItemProtoBytes` to clarify that the field holds
  the serialized outer `RecordFileItem` proto, not the inner record stream file

### Existing test updates

- `AllBlocksHasherHandlerTest`: restored `mock(ApplicationStateFacility.class)` for the
  `applicationStateFacility` argument (had been changed to `null`, risking NPE)
- `ExtendedMerkleTreeSessionTest`, `StateProofVerificationTest`, `TssBlockProofVerificationTest`,
  `TestVerificationHapiVersions`: updated `ExtendedMerkleTreeSession` constructor calls with the
  four new RSA parameters (`Map.of(), null, null, null`)

---

## Test coverage

|---|---|
| `RsaWrbVerificationTest.validProofAtExactThreshold_accepted` | Exactly supermajority (5/6) signatures accepted |
| `RsaWrbVerificationTest.validProofAllNodesSigning_accepted` | All 6 nodes signing accepted |
| `RsaWrbVerificationTest.validProofOneBelowThreshold_rejected` | 4/6 signatures rejected |
| `RsaWrbVerificationTest.allZeroSignatures_rejected` | All-zero signature bytes skipped; proof fails |
| `RsaWrbVerificationTest.emptySignatureList_rejected` | Empty signature list rejected |
| `RsaWrbVerificationTest.unknownNodeId_sigSkipped_othersStillCounted` | Unknown node_id skipped; 5 valid from known nodes still accepted |
| `RsaWrbVerificationTest.wrongPayloadSignature_countedInvalid` | Signatures over wrong data rejected |
| `RsaWrbVerificationTest.emptyAddressBook_rejected` | Empty key map causes immediate ERROR + rejection |
| `RsaWrbVerificationTest.malformedKeyExcludedFromMap_restStillCounted` | Reduced 5-node roster; 5 valid sigs accepted |
| `RsaWrbVerificationTest.duplicateNodeId_countedOnlyOnce` | Duplicate node_id entry does not inflate validCount |
| `RsaWrbVerificationTest.unsupportedVersion_rejected` | V5 proof rejected; only V6 supported |
| `RsaWrbVerificationTest.recordStreamFileBytesExtractedCorrectly_signatureVerifies` | Wire-format extraction with field-1 preceding field-2; end-to-end signature still verifies |
| `VerificationServicePluginTest.onContextUpdate_nonEmptyBook_rebuildsKeyMap` | `onContextUpdate` with a real RSA key updates the map |
| `VerificationServicePluginTest.onContextUpdate_emptyBook_retainsKeyMap` | Empty book does not crash or wipe existing map |

---

## Verification algorithm (V6, equal-weight)

```
Input:  SignedRecordFileProof { version=6, recordFileSignatures: [{node_id, sig_bytes}, ...] }
        rawRecordFileItemProtoBytes (captured from the RECORD_FILE block item)

1. Extract field-2 bytes from rawRecordFileItemProtoBytes via wire-format navigation
   → rawRecordStreamFileBytes

2. signedPayload = SHA-384(int32(6) big-endian || rawRecordStreamFileBytes)

3. validCount = 0 ; validatedNodes = {}
   for each (node_id, sig_bytes) in recordFileSignatures:
     - skip if node_id not in keyByNodeId           (roster mismatch)
     - skip if node_id in validatedNodes             (duplicate)
     - skip if sig_bytes are all zeros
     - SHA384withRSA.initVerify(keyByNodeId[node_id])
       SHA384withRSA.update(signedPayload)
       if SHA384withRSA.verify(sig_bytes): validCount++; validatedNodes.add(node_id)

4. threshold = 2 * ceil(rosterSize / 3) + 1
   accept if validCount >= threshold
```

> **Note — `SHA384withRSA` vs raw digest**: `signedPayload` is the 48-byte SHA-384 digest, not the
> raw data. `SHA384withRSA.update(signedPayload)` will hash it again internally. Both the test
> signing helper and the production verification code follow this same pattern, so they are
> internally consistent. Whether this matches the consensus node's signing contract (signing the
> raw bytes vs the pre-hashed bytes) should be confirmed before enabling mainnet verification.

---

## Reviewer Notes

- **Depends on #2718**: `BlockNodeContext.nodeAddressBook()` and `ApplicationStateFacility.updateAddressBook()` are defined there. This PR cannot be merged before #2718.
- **`RsaRosterBootstrapPlugin` (#2683)**: populates `keyByNodeId` at startup. Without it, `keyByNodeId` stays `Map.of()` and every WRB block fails with an ERROR log. This is expected during the transition period while #2683 is pending.
- **Phase 2a scope**: only record file format version 6 is supported. A block with `version != 6` is rejected with a WARNING.
- **Stake-weighted threshold**: the threshold formula `2 * ceil(N/3) + 1` uses equal-weight (one node, one vote). Stake-weighted verification is out of scope for Phase 2a and tracked as a follow-up.
- **Test key size**: `RsaWrbVerificationTest` uses 1024-bit keys for speed; production network nodes use 4096-bit keys.

---

## Related Issue(s)
Depends on #2718 (NodeAddressBook state — must be merged first)
Related: #2683 (RsaRosterBootstrapPlugin — provides the address book at startup)
Closes #2562

### DIFF
--- a/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
+++ b/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
@@ -9,6 +9,7 @@ import static java.lang.System.Logger.Level.WARNING;
 import com.hedera.cryptography.tss.TSS;
 import com.hedera.cryptography.wraps.WRAPSVerificationKey;
 import com.hedera.hapi.block.stream.output.BlockHeader;
+import com.hedera.hapi.node.base.NodeAddressBook;
 import com.hedera.hapi.node.base.SemanticVersion;
 import com.hedera.hapi.node.tss.LedgerIdNodeContribution;
 import com.hedera.hapi.node.tss.LedgerIdPublicationTransactionBody;
@@ -18,7 +19,9 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.file.Files;
+import java.security.PublicKey;
 import java.util.List;
+import java.util.Map;
 import org.hiero.block.node.app.config.node.NodeConfig;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
@@ -57,6 +60,15 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
     /** Metric key for block hashing time */
     public static final MetricKey<LongCounter> METRIC_HASHING_BLOCK_TIME =
             MetricKey.of("hashing_block_time", LongCounter.class).addCategory(METRICS_CATEGORY);
+    /** Metric key for WRB blocks whose RSA proof was accepted */
+    public static final MetricKey<LongCounter> METRIC_RSA_VERIFICATION_SUCCESS =
+            MetricKey.of("rsa_verification_success_total", LongCounter.class).addCategory(METRICS_CATEGORY);
+    /** Metric key for WRB blocks whose RSA proof was rejected */
+    public static final MetricKey<LongCounter> METRIC_RSA_VERIFICATION_FAILURE =
+            MetricKey.of("rsa_verification_failure_total", LongCounter.class).addCategory(METRICS_CATEGORY);
+    /** Metric key for signatures from `node_id` values not present in the loaded address book */
+    public static final MetricKey<LongCounter> METRIC_RSA_ROSTER_MISMATCH =
+            MetricKey.of("rsa_roster_mismatch_total", LongCounter.class).addCategory(METRICS_CATEGORY);
 
     private static final String COMPLETED_MESSAGE = "Verified backfill block items for block={0} with success={1}";
     /** The logger for this class. */
@@ -82,6 +94,18 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
     private LongCounter.Measurement verificationBlockTime;
     /** Metric for block hashing time. ignores time to receive block */
     private LongCounter.Measurement hashingBlockTimeNs;
+    /** Metric for accepted RSA WRB proof verifications. */
+    private LongCounter.Measurement rsaVerificationSuccessTotal;
+    /** Metric for rejected RSA WRB proof verifications. */
+    private LongCounter.Measurement rsaVerificationFailureTotal;
+    /** Metric for signatures from `node_id` values absent from the loaded address book. */
+    private LongCounter.Measurement rsaRosterMismatchTotal;
+    /**
+     * Most recent `node_id → PublicKey` map built from the `NodeAddressBook` delivered by
+     * `RsaRosterBootstrapPlugin`. Volatile so that `onContextUpdate` writes are visible to the
+     * block-handling thread without a lock.
+     */
+    private volatile Map<Long, PublicKey> keyByNodeId = Map.of();
     /** The previous block hash, used for verification of the current block. */
     private Bytes previousBlockHash;
     /** The block number of the last successfully verified block (live-stream or sequential backfill). */
@@ -140,6 +164,19 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
                 .getOrCreateNotLabeled();
         hashingBlockTimeNs = metricRegistry
                 .register(LongCounter.builder(METRIC_HASHING_BLOCK_TIME).setDescription("Hashing time per block (ms)"))
+                .getOrCreateNotLabeled();
+        rsaVerificationSuccessTotal = metricRegistry
+                .register(LongCounter.builder(METRIC_RSA_VERIFICATION_SUCCESS)
+                        .setDescription("WRB blocks whose RSA SignedRecordFileProof was accepted"))
+                .getOrCreateNotLabeled();
+        rsaVerificationFailureTotal = metricRegistry
+                .register(LongCounter.builder(METRIC_RSA_VERIFICATION_FAILURE)
+                        .setDescription("WRB blocks whose RSA SignedRecordFileProof was rejected"))
+                .getOrCreateNotLabeled();
+        rsaRosterMismatchTotal = metricRegistry
+                .register(LongCounter.builder(METRIC_RSA_ROSTER_MISMATCH)
+                        .setDescription(
+                                "RSA signatures from node_id values absent from the loaded address book"))
                 .getOrCreateNotLabeled();
     }
 
@@ -202,6 +239,22 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
 
     /**
      * {@inheritDoc}
+     *
+     * <p>Called when the `NodeAddressBook` is updated by `RsaRosterBootstrapPlugin`. Rebuilds the
+     * `node_id → PublicKey` map used by subsequent WRB verification sessions. A volatile write
+     * ensures visibility to the block-handler thread without a lock.
+     */
+    @Override
+    public void onContextUpdate(final BlockNodeContext updatedContext) {
+        final NodeAddressBook book = updatedContext.nodeAddressBook();
+        if (!book.nodeAddress().isEmpty()) {
+            keyByNodeId = RsaKeyDecoder.buildKeyMap(book);
+            LOGGER.log(INFO, "RSA key map updated: {0} nodes loaded from address book", keyByNodeId.size());
+        }
+    }
+
+    /**
+     * {@inheritDoc}
      */
     @Override
     public void start() {
@@ -258,7 +311,11 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
                         semanticVersion,
                         previousHash,
                         getRootOfAllPreviousBlocks(),
-                        activeLedgerId);
+                        activeLedgerId,
+                        keyByNodeId,
+                        rsaVerificationSuccessTotal,
+                        rsaVerificationFailureTotal,
+                        rsaRosterMismatchTotal);
                 LOGGER.log(DEBUG, "Started new block verification session for block number {0}", currentBlockNumber);
             } else {
                 headerValid = true; // header not present, assume it was valid
@@ -414,7 +471,11 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
                         blockHeader.hapiProtoVersionOrThrow(),
                         null,
                         null,
-                        activeLedgerId);
+                        activeLedgerId,
+                        keyByNodeId,
+                        rsaVerificationSuccessTotal,
+                        rsaVerificationFailureTotal,
+                        rsaRosterMismatchTotal);
                 // process the block items in the backfilled notification
                 // For backfill, we wrap items in BlockItems with isEndOfBlock=true (last item should be block proof)
                 BlockItems backfillBlockItems =
--- a/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
+++ b/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
@@ -254,8 +254,28 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
                             + " WRB blocks will fail verification until a valid address book is delivered.");
             return;
         }
+        // Count non-blank address book entries — these are the nodes that should be signable.
+        final int declaredCount =
+                (int) book.nodeAddress().stream().filter(a -> !a.rsaPubKey().isBlank()).count();
         keyByNodeId = RsaKeyDecoder.buildKeyMap(book);
-        LOGGER.log(INFO, "RSA key map updated: {0} nodes loaded from address book", keyByNodeId.size());
+        final int effectiveCount = keyByNodeId.size();
+        if (effectiveCount < declaredCount) {
+            // Malformed DER keys were skipped; threshold is calculated against effectiveCount.
+            // Fix the address book so all declared nodes can contribute signatures.
+            LOGGER.log(
+                    WARNING,
+                    "RSA key map: {0}/{1} keys decoded successfully; {2} node(s) had malformed"
+                            + " hex-DER bytes and cannot contribute signatures. Verification"
+                            + " threshold is calculated against the {0} decodable keys.",
+                    effectiveCount,
+                    declaredCount,
+                    declaredCount - effectiveCount);
+        }
+        LOGGER.log(
+                INFO,
+                "RSA key map updated: {0}/{1} nodes loaded from address book",
+                effectiveCount,
+                declaredCount);
     }
 
     /**
--- a/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
+++ b/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
@@ -255,8 +255,9 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
             return;
         }
         // Count non-blank address book entries — these are the nodes that should be signable.
-        final int declaredCount =
-                (int) book.nodeAddress().stream().filter(a -> !a.rsaPubKey().isBlank()).count();
+        final int declaredCount = (int) book.nodeAddress().stream()
+                .filter(a -> !a.rsaPubKey().isBlank())
+                .count();
         keyByNodeId = RsaKeyDecoder.buildKeyMap(book);
         final int effectiveCount = keyByNodeId.size();
         if (effectiveCount < declaredCount) {
@@ -271,11 +272,7 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
                     declaredCount,
                     declaredCount - effectiveCount);
         }
-        LOGGER.log(
-                INFO,
-                "RSA key map updated: {0}/{1} nodes loaded from address book",
-                effectiveCount,
-                declaredCount);
+        LOGGER.log(INFO, "RSA key map updated: {0}/{1} nodes loaded from address book", effectiveCount, declaredCount);
     }
 
     /**
--- a/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
+++ b/block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
@@ -175,8 +175,7 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
                 .getOrCreateNotLabeled();
         rsaRosterMismatchTotal = metricRegistry
                 .register(LongCounter.builder(METRIC_RSA_ROSTER_MISMATCH)
-                        .setDescription(
-                                "RSA signatures from node_id values absent from the loaded address book"))
+                        .setDescription("RSA signatures from node_id values absent from the loaded address book"))
                 .getOrCreateNotLabeled();
     }
 
@@ -240,17 +239,23 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
     /**
      * {@inheritDoc}
      *
-     * <p>Called when the `NodeAddressBook` is updated by `RsaRosterBootstrapPlugin`. Rebuilds the
-     * `node_id → PublicKey` map used by subsequent WRB verification sessions. A volatile write
-     * ensures visibility to the block-handler thread without a lock.
+     * <p>Called by the framework whenever `BlockNodeApp.updateAddressBook()` fires — typically once
+     * at startup by `RsaRosterBootstrapPlugin`. Rebuilds the `node_id → PublicKey` map used by
+     * subsequent WRB verification sessions. A volatile write ensures visibility to the
+     * block-handler thread without a lock.
      */
     @Override
     public void onContextUpdate(final BlockNodeContext updatedContext) {
         final NodeAddressBook book = updatedContext.nodeAddressBook();
-        if (!book.nodeAddress().isEmpty()) {
-            keyByNodeId = RsaKeyDecoder.buildKeyMap(book);
-            LOGGER.log(INFO, "RSA key map updated: {0} nodes loaded from address book", keyByNodeId.size());
+        if (book == null || book.nodeAddress().isEmpty()) {
+            LOGGER.log(
+                    WARNING,
+                    "onContextUpdate called with a null or empty NodeAddressBook — RSA key map not updated."
+                            + " WRB blocks will fail verification until a valid address book is delivered.");
+            return;
         }
+        keyByNodeId = RsaKeyDecoder.buildKeyMap(book);
+        LOGGER.log(INFO, "RSA key map updated: {0} nodes loaded from address book", keyByNodeId.size());
     }
 
     /**
--- a/block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
+++ b/block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
@@ -561,49 +561,85 @@ public class ExtendedMerkleTreeSession implements VerificationSession {
     }
 
     /**
-     * Extracts the raw `record_file_contents` bytes (proto field 2) from a serialized
-     * `RecordFileItem` message by navigating the protobuf wire format without full parsing.
+     * Extracts the raw {@code record_file_contents} bytes from a serialized {@code RecordFileItem}
+     * proto message by walking the protobuf wire format directly, without deserializing the message.
      *
-     * <p>The verbatim bytes are required for V6 signed hash computation so that the hash
-     * is identical to the one computed by the consensus node — re-serializing from a parsed
-     * representation would risk byte-order or encoding differences.
+     * <p>{@code record_file_contents} is proto field 2 of {@code RecordFileItem}. These bytes are
+     * the verbatim content of the {@code .rcd} record stream file exactly as the consensus node
+     * read it from disk when it computed the V6 signed hash. They must be returned byte-for-byte
+     * identical to what the consensus node used; full deserialization via
+     * {@code RecordFileItem.PROTOBUF.parse()} is deliberately avoided because re-serializing a
+     * parsed object can produce subtly different bytes (e.g. omitting default-value fields, different
+     * varint encoding choices), which would cause the recomputed hash to diverge from the one the
+     * consensus node signed.
      *
-     * @param recordFileItemBytes raw serialized bytes of a `RecordFileItem` proto message
-     * @return raw bytes of the `record_file_contents` field, or `Bytes.EMPTY` if not found
+     * <p><b>Protobuf wire format:</b> every field on the wire is encoded as a tag varint followed
+     * by its value. The tag packs two things:
+     * <ul>
+     *   <li>{@code fieldNumber = tag >>> 3}
+     *   <li>{@code wireType   = tag & 0x7}
+     * </ul>
+     * Wire type 2 ({@code LEN}) means the value is length-prefixed bytes, used for {@code bytes},
+     * {@code string}, and embedded messages. It is encoded as:
+     * {@code [tag varint] [length varint] [raw bytes...]}.
+     *
+     * <p><b>Algorithm:</b>
+     * <ol>
+     *   <li>Read the next field tag varint and decode its field number and wire type.</li>
+     *   <li>If {@code fieldNumber == 2} and {@code wireType == LEN}: read the length prefix varint,
+     *       read exactly that many bytes, and return them — these are the
+     *       {@code record_file_contents}.</li>
+     *   <li>Otherwise skip the field using the wire type to know how many bytes to consume:
+     *       <ul>
+     *         <li>VARINT (wire 0): read and discard one varint</li>
+     *         <li>I64 (wire 1): skip 8 bytes fixed</li>
+     *         <li>LEN (wire 2): read the length prefix, skip that many bytes</li>
+     *         <li>I32 (wire 5): skip 4 bytes fixed</li>
+     *       </ul>
+     *   </li>
+     *   <li>Repeat until field 2 is found or input is exhausted.</li>
+     * </ol>
+     *
+     * @param recordFileItemBytes raw serialized bytes of a {@code RecordFileItem} proto message
+     * @return verbatim bytes of the {@code record_file_contents} field (proto field 2), or
+     *         {@code Bytes.EMPTY} if field 2 is not present or if any parse error occurs
      */
     private static Bytes extractRecordStreamFileBytes(final Bytes recordFileItemBytes) {
         try {
             final ReadableSequentialData input = recordFileItemBytes.toReadableSequentialData();
             while (input.hasRemaining()) {
+                // Each field starts with a tag varint: high bits = field number, low 3 bits = wire type
                 final int tag = input.readVarInt(false);
                 final int wireType = tag & 0x7;
                 final int fieldNumber = tag >>> 3;
                 if (fieldNumber == 2 && wireType == 2) {
-                    // LEN field: read the length prefix then copy the raw bytes
+                    // Found record_file_contents (field 2, LEN wire type).
+                    // Read the length-prefix varint then copy the raw payload bytes verbatim.
                     final int len = input.readVarInt(false);
                     final byte[] raw = new byte[len];
                     input.readBytes(raw);
                     return Bytes.wrap(raw);
                 }
-                // Skip any other field by wire type
+                // Not field 2 — skip this field using its wire type to advance the cursor correctly
                 switch (wireType) {
-                    case 0 -> input.readVarLong(false); // VARINT: consume value
-                    case 1 -> input.skip(8); // I64: skip 8 bytes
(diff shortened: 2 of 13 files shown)


block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationServicePluginTest.java
```diff
@@ -397,6 +411,170 @@ void shouldNotOverwriteFileLoadedTssParameters() throws IOException, ParseExcept
...
+                blockNodeContext.blockNodeVersions(),
+                blockNodeContext.tssData(),
+                book);
+
+        plugin.onContextUpdate(updatedContext);
+
+        // Verify the key map now contains node 7's key (indirect: the plugin won't reject a WRB
+        // from node 7 as "unknown node"). The simplest assertion is that the method did not throw.
+        // A deeper assertion would require exposing `keyByNodeId`, which we avoid.
+        // Coverage is provided by the unit tests in `RsaWrbVerificationTest`.
```
AlfredoG87, line 454: Pending Assertions

block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationServicePluginTest.java
```diff
@@ -397,6 +411,170 @@ void shouldNotOverwriteFileLoadedTssParameters() throws IOException, ParseExcept
...
+                blockNodeContext.historicalBlockProvider(),
+                blockNodeContext.applicationStateFacility(),
+                blockNodeContext.serviceLoader(),
+                blockNodeContext.threadPoolManager(),
+                blockNodeContext.blockNodeVersions(),
+                blockNodeContext.tssData(),
+                emptyBook);
+
+        // Must not throw; a WARNING is expected but not asserted here (logged to system logger).
+        plugin.onContextUpdate(updatedContext);
```
AlfredoG87, line 477: Pending Assertions

block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationServicePluginTest.java
```diff
@@ -397,6 +411,170 @@ void shouldNotOverwriteFileLoadedTssParameters() throws IOException, ParseExcept
...
+                blockNodeContext.historicalBlockProvider(),
+                blockNodeContext.applicationStateFacility(),
+                blockNodeContext.serviceLoader(),
+                blockNodeContext.threadPoolManager(),
+                blockNodeContext.blockNodeVersions(),
+                blockNodeContext.tssData(),
+                null);
+
+        // Must not throw; a WARNING is expected but not asserted here (logged to system logger).
+        plugin.onContextUpdate(updatedContext);
```
AlfredoG87, line 498: Pending assertions

block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationServicePluginTest.java
```diff
@@ -397,6 +411,170 @@ void shouldNotOverwriteFileLoadedTssParameters() throws IOException, ParseExcept
...
     }
 
+    // ==== RSA Address Book / onContextUpdate Tests ==================================================================
+
+    @Test
+    @DisplayName("onContextUpdate with non-empty address book rebuilds keyByNodeId")
+    void onContextUpdate_nonEmptyBook_rebuildsKeyMap() throws Exception {
+        // Build a minimal NodeAddress with a real RSA public key encoded as hex-DER.
+        // 1024-bit RSA keys are used for test speed only — production network uses 4096-bit keys.
+        final java.security.KeyPairGenerator kpg = java.security.KeyPairGenerator.getInstance("RSA");
```
AlfredoG87, line 421: why using the fully qualified name? we should add imports

block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationServicePluginTest.java
```diff
@@ -397,6 +411,170 @@ void shouldNotOverwriteFileLoadedTssParameters() throws IOException, ParseExcept
...
+        kpg.initialize(1024);
+        final java.security.KeyPair kp = kpg.generateKeyPair();
+        final String hexKey = java.util.HexFormat.of().formatHex(kp.getPublic().getEncoded());
```
AlfredoG87, line 424: same as above, we should import: `java.util.HexFormat` 

Won't make more comments like this, but there are a lot, we need to make sure that we only use full names when strictly necessary.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -176,16 +257,21 @@ public VerificationNotification finalizeVerification(Bytes rootHashOfAllBlockHas
         // while we don't have state management, use the start of block state root hash from the footer
         Bytes startOfBlockStateRootHash = this.blockFooter.startOfBlockStateRootHash();
 
+        // Route by proof type. RSA WRB proofs take priority (Phase 2a); TSS proofs follow.
+        final BlockProof rsaProof = getFirst(blockProofs, BlockProof::hasSignedRecordFileProof);
```
AlfredoG87, line 261: do we expect more than 1 SignedRecordProof per Block?

as per my understanding we expect only one `signedRecordFileProof ` with multiple `recordFileSignatures`

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -258,15 +345,364 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+     * @return a `VerificationNotification` with `success=true` when the proof passes, `false` otherwise
+     */
+    private VerificationNotification verifyRsaProof(
+            final BlockProof blockProof,
+            final Bytes previousBlockHash,
+            final Bytes rootOfAllPreviousBlockHashes,
+            final Bytes startOfBlockStateRootHash) {
+
+        // Compute block root hash for chain continuity — identical computation for all proof types
+        final Bytes blockRootHash = HashingUtilities.computeFinalBlockHash(
```
AlfredoG87, line 424: we are never using this ever.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -141,6 +217,11 @@ public VerificationNotification processBlockItems(BlockItems blockItemsMessage)
                 case TRANSACTION_RESULT, TRANSACTION_OUTPUT -> outputTreeHasher.addLeaf(getBlockItemHash(item));
                 case STATE_CHANGES -> stateChangesHasher.addLeaf(getBlockItemHash(item));
                 case TRACE_DATA -> traceDataHasher.addLeaf(getBlockItemHash(item));
+                // WRB record file — captured for RSA payload computation and hashed as output
+                case RECORD_FILE -> {
+                    this.rawRecordFileItemProtoBytes = item.recordFileOrThrow();
+                    outputTreeHasher.addLeaf(getBlockItemHash(item));
+                }
```
AlfredoG87, line 224: Is this specified in some HIP or documentation?

block-node/verification/src/test/java/org/hiero/block/node/verification/AllBlocksHasherHandlerTest.java
```diff
@@ -317,8 +318,8 @@ private BlockNodeContext buildContext(final HistoricalBlockFacility facility) {
                 mock(ServiceLoaderFunction.class),
                 mock(ThreadPoolManager.class),
                 BlockNodeVersions.DEFAULT,
-                null,
-                null);
+                TssData.DEFAULT,
+                com.hedera.hapi.node.base.NodeAddressBook.DEFAULT);
```
AlfredoG87, line 322: do we have to use full name here?

block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
```diff
@@ -57,6 +60,15 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
...
             MetricKey.of("hashing_block_time", LongCounter.class).addCategory(METRICS_CATEGORY);
+    /** Metric key for WRB blocks whose RSA proof was accepted */
+    public static final MetricKey<LongCounter> METRIC_RSA_VERIFICATION_SUCCESS =
+            MetricKey.of("rsa_verification_success_total", LongCounter.class).addCategory(METRICS_CATEGORY);
+    /** Metric key for WRB blocks whose RSA proof was rejected */
+    public static final MetricKey<LongCounter> METRIC_RSA_VERIFICATION_FAILURE =
+            MetricKey.of("rsa_verification_failure_total", LongCounter.class).addCategory(METRICS_CATEGORY);
+    /** Metric key for signatures from `node_id` values not present in the loaded address book */
+    public static final MetricKey<LongCounter> METRIC_RSA_ROSTER_MISMATCH =
+            MetricKey.of("rsa_roster_mismatch_total", LongCounter.class).addCategory(METRICS_CATEGORY);
```
AlfredoG87, line 71: We need to add these metrics to the docs, also usually we add some widgets to the default dashboards, but this can be done in a follow-up.

block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
```diff
@@ -82,6 +94,18 @@ public class VerificationServicePlugin implements BlockNodePlugin, BlockItemHand
...
+    /** Metric for rejected RSA WRB proof verifications. */
+    private LongCounter.Measurement rsaVerificationFailureTotal;
+    /** Metric for signatures from `node_id` values absent from the loaded address book. */
+    private LongCounter.Measurement rsaRosterMismatchTotal;
+    /**
+     * Most recent `node_id → PublicKey` map built from the `NodeAddressBook` delivered by
+     * `RsaRosterBootstrapPlugin`. Volatile so that `onContextUpdate` writes are visible to the
+     * block-handling thread without a lock.
+     */
+    private volatile Map<Long, PublicKey> keyByNodeId = Map.of();
```
AlfredoG87, line 108: should we have a record of NodeID, PublicKey and Weight/Stake?

AlfredoG87 (commented):
Looking good, mostly just Nits and suggestions. also some questions...

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -258,15 +345,364 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+                inputTreeHasher,
+                outputTreeHasher,
+                consensusHeaderHasher,
+                stateChangesHasher,
+                traceDataHasher);
+
+        // Guard: address book must be loaded before any WRB arrives
+        if (rsaKeyByNodeId.isEmpty()) {
+            LOGGER.log(
+                    ERROR,
```
AlfredoG87, line 438: I would argue this should be a WARN instead of an ERROR. 
But just my opinion.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -258,15 +345,364 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+                    blockNumber);
+            if (rsaVerificationFailureTotal != null) rsaVerificationFailureTotal.increment();
+            return new VerificationNotification(
+                    false, FailureType.BAD_BLOCK_PROOF, blockNumber, null, null, blockSource);
+        }
+
+        // Guard: RECORD_FILE item must be present in the block
+        if (rawRecordFileItemProtoBytes == null) {
+            LOGGER.log(
+                    ERROR, "No RECORD_FILE item found in WRB block {0} — cannot compute signed payload", blockNumber);
```
AlfredoG87, line 450: I would argue this should be a WARN instead of an ERROR. 
But just my opinion.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -258,15 +345,364 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+                    blockNumber);
+            if (rsaVerificationFailureTotal != null) rsaVerificationFailureTotal.increment();
+            return new VerificationNotification(
+                    false, FailureType.BAD_BLOCK_PROOF, blockNumber, null, null, blockSource);
+        }
+
+        // Extract the raw RecordStreamFile bytes from the RecordFileItem proto (field 2)
+        final Bytes rawRecordStreamFileBytes = extractRecordStreamFileBytes(rawRecordFileItemProtoBytes);
+        if (rawRecordStreamFileBytes.length() == 0) {
+            LOGGER.log(ERROR, "Failed to extract record_file_contents from RECORD_FILE item in block {0}", blockNumber);
```
AlfredoG87, line 474: I would argue this should be a WARN instead of an ERROR. 
But just my opinion.

AlfredoG87 (dismissed):
Good to be merged as is, everything can be done on a follow-up.

Nana-EC:
All comments to be address in follow up ticket https://github.com/hiero-ledger/hiero-block-node/issues/2773

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
Nana-EC: Agreed

block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
Nana-EC: We don't need it in this revision but we will need it when we store all addressBook versions along with weights.
Then the map could be changed to something like this
```suggestion
    private volatile Map<Long, NodeEntry> keyByNodeId = Map.of();
```

block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
Nana-EC: missed the docs, good catch.
Agree with the dashboard mention

block-node/verification/src/test/java/org/hiero/block/node/verification/AllBlocksHasherHandlerTest.java
Nana-EC: Yep, agreed here and other places where I left full names - import them.
It's helpful early on when running since it doesn't pollute the imports with unused imports if you change the approach.

AlfredoG87 (approved):
Unblocking

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
Nana-EC: I'm currently adding it to the HIP.
I pulled the latest from the proto though - https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hapi/hedera-protobuf-java-api/src/main/proto/block/stream/block_item.proto#L130

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
Nana-EC: Do you mean the `blockRootHash`? 
We use it in the VerificationNotification return object

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
Nana-EC: Only 1 per block correct. However, in the future we expect a WRB could have multiple proof when a TSSSignedProof is added

block-node/verification/src/test/java/org/hiero/block/node/verification/VerificationServicePluginTest.java
Nana-EC: Yeah, missed coming to clean the import up.

block-node/verification/src/test/java/org/hiero/block/node/verification/RsaKeyDecoderTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        assertEquals(1, result.size());
+        assertNotNull(result.get(7L), "Node 7's PublicKey must be present");
+    }
+
+    @Test
+    @DisplayName("blank rsaPubKey is skipped; malformed hex-DER is skipped; valid keys are retained")
+    void blankAndMalformedKeys_skipped_validKeysRetained() throws Exception {
+        final KeyPairGenerator kpg = KeyPairGenerator.getInstance("RSA");
+        kpg.initialize(1024);
```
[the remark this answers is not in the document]
Nana-EC: In test. setup, not production code

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: We expect one, but should handle it correctly if there is more.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: Yes, in HIP-1424 the RecordFileItem is part of the `Output` tree.
Listed [here](https://github.com/Nana-EC/hiero-improvement-proposal/blob/2e01b1a295fcb26a83b71a2bdc2d29185d855f46/HIP/hip-1424.md#fixed-16-leaf-positions).

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: ?
Are you saying the block root hash is unused, or that this method call should not be used?

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -258,15 +345,364 @@ protected Boolean verifySignature(@NonNull Bytes hash, @NonNull Bytes signature)
...
+
+        if (rsaRosterMismatchTotal != null && mismatchCount > 0) {
+            rsaRosterMismatchTotal.increment(mismatchCount);
+        }
+
+        // Strict majority threshold: validCount must be strictly greater than half the roster.
+        // The CN only sends signatures from nodes whose combined consensus weight exceeded 50%,
+        // so we require validCount * 2 > rosterSize (equivalent to validCount > rosterSize / 2.0).
+        // For example: 6-node roster requires > 3, i.e. ≥ 4 valid signatures.
+        final boolean accepted = validCount * 2 > rosterSize;
```
jsync-swirlds, line 552: We probably need to use greater-than-or-equal here; to match what Consensus is including.

Also, we should reject any block proof if even `1` signature failed signature verification.  All signatures present in the proof must be valid; we can also save time if we check the count of signatures before validating them.

Nana-EC: Ahh yeah, should be # count of valid sigs >= N/2, where N is roster size.

Regarding the second point there's already logic present to fail if even one fails. See `verifyRsaProof()`, we fail fast and return a `VerificationNotification` with a `FailureType.BAD_BLOCK_PROOF`

jsync-swirlds: Count of signatures (valid or otherwise) is the faster check; we don't care to count valid, we just count total, because if that's not enough, we fail, and if it is, then all of them will be valid (or we fail).

jsync-swirlds: Note, this is for future enhancement.

jsync-swirlds (approved):

PR #2738, opened 2026-05-07
feat: API method for adding stored block

### LINKED ISSUE #2702: API method for adding stored block
### Story Form

As a Block Node
I want to provide interface to add stored blocks to the Application State Facility
So that plugins that store blocks are able to report that

### Technical Notes

* Add a specific API method to the Application State Facility that adds stored block(s)
* Keep that that data in a ranged set
* Periodically store the data on disk
* Load the data (if available) upon init

ivannov:
## Reviewer Notes

Adds `addBlockRange(LongRange, BlockRangeType)` to `ApplicationStateFacility` and implements it in `BlockNodeApp`, giving plugins a standard way to report which blocks they have stored or can serve.
                                                                                                                                                                                                                                    
Two `ConcurrentLongRangeSet` fields (`storedBlocks`, `availableBlocks`) are maintained in the implementation class to track the two types of sets: stored and available. Every `BLOCK_RANGE_PERSIST_INTERVAL` (1000) blocks accumulated across all calls, both sets are written to disk in a compact binary format. The sets are also persisted unconditionally on `stopApplicationStateFacility()` so no ranges are lost at shutdown regardless of where the counter is. On startup, `loadApplicationState` reads both files back via the same binary format.

Three new config properties are added to the `ApplicationStateConfig`: `tssDataFilePath` (actually renamed from dataFilePath), `storedBlocksFilePath`, and `availableBlocksFilePath`, all with defaults under `/opt/hiero/block-node/node/`.

_Application State Configuration_ section was added to configuration.md covering all five properties.     

## Related Issue(s)

Closes #2702


block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -378,6 +392,21 @@ public void updateTssData(TssData tssData) {
...
+        storedBlocks.add(blockRange);
+        if (blockRangeType == BlockRangeType.AVAILABLE) {
+            availableBlocks.add(blockRange);
+        }
+        // Trigger a persist whenever the running total crosses a BLOCK_RANGE_PERSIST_INTERVAL boundary.
+        final long blocksToAdd = blockRange.size();
+        final long before = storedBlockCount.getAndAdd(blocksToAdd);
+        if (before / BLOCK_RANGE_PERSIST_INTERVAL < (before + blocksToAdd) / BLOCK_RANGE_PERSIST_INTERVAL) {
+            persistBlockRanges();
+        }
```
ata-nas: This, being called from multiple threads will conflict on write (persist to disk), I believe.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -495,6 +525,34 @@ private void persistTssData(TssData tssData) {
...
+    private void persistBlockRanges() {
+        final ApplicationStateConfig appStateConfig =
+                blockNodeContext.configuration().getConfigData(ApplicationStateConfig.class);
+        persistRangeSet(storedBlocks, appStateConfig.storedBlocksFilePath());
+        persistRangeSet(availableBlocks, appStateConfig.availableBlocksFilePath());
+    }
+
+    private void persistRangeSet(ConcurrentLongRangeSet rangeSet, Path filePath) {
+        try {
+            Files.createDirectories(filePath.getParent());
```
ata-nas: The dir creation feels like something that should happen only once on startup.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -406,13 +435,13 @@ void startApplicationStateFacility() {
...
         applicationStateExecutor.scheduleAtFixedRate(
-                this::checkForApplicationStateUpdates,
+                this::checkForTssDataStateUpdates,
                 appStateConfig.updateInitialDelay(),
                 appStateConfig.updateScanInterval(),
                 TimeUnit.MILLISECONDS);
     }
 
-    private void checkForApplicationStateUpdates() {
+    private void checkForTssDataStateUpdates() {
```
ata-nas: Why the rename?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -378,6 +392,21 @@ public void updateTssData(TssData tssData) {
         if (tssData != null) tssDataUpdates.add(tssData);
     }
 
+    @Override
+    public void addBlockRange(LongRange blockRange, BlockRangeType blockRangeType) {
```
ata-nas: How do we later get this data?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: This will conflict with RSA changes. RSA update checks were added to  checkForApplicationStateUpdates()

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: In the task (that I created, LOL), the requirement was only to persist this data. I can add another sub-task for accessing the data

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: I made the method synchronized

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: I renamed some methods and variables that looked very general to me. But now, as RSA update was added, I renamed back the method as it was

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: I blindly copied this pattern from the other persist methods. But you are right, we don't need this called every time. Moved it to initialization

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: The "stored" ranged set should generally be always available at runtime to everyone. It must also be dynamically updated at runtime so we have all the latest.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: I do not believe this is the solution. The application state facility receive all kinds of updates. All updates must be made known to plugins. We can only ever do that, reasonably, if the next update comes after everyone has been notified of the previous one. There is a scheduled task, currently, that publishes all updates. I believe that is the correct place to do any update. As far as persistence, it must also be done in a similar fashion, making sure that no conflicts are possible. Making this method synchronized is quite the performance hit.

The way I see it conceptually, others could disagree, is that updates are being queued up. Then, each update is being published to everyone in the order things were queued up, by a single thread. In this way, we ensure that no update will be left unprocessed, but also that the end result, which is a block node context, will be in a state that has all the updates and none have been overwritten.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -96,6 +101,15 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     /** Pending address book loaded at startup; consumed by the first checkForApplicationStateUpdates run. */
     private final AtomicReference<NodeAddressBook> pendingAddressBook = new AtomicReference<>();
 
+    /** Blocks reported as stored by plugins that do not serve them for retrieval */
+    final ConcurrentLongRangeSet storedBlocks = new ConcurrentLongRangeSet();
+
+    /** Blocks reported as available by BlockProviderPlugin implementations */
+    final ConcurrentLongRangeSet availableBlocks = new ConcurrentLongRangeSet();
```
ata-nas, line 119: Not sure about the addition of available blocks here. Maybe this is the right choice, but availability is per provider and not a consolidated result.

If 2 providers have block 5, then one deletes block 5, how do we update this? We cannot remove block 5 from here, cause it is still available, but also, we cannot find who has the block available. I think the historical block facility is the right place to keep this data. It holds a combined ranged set of all available blocks, valid for all providers, but we are also able to do non-destructive updates for individual ranged set there. If we chose to consolidate and persist this data, we can easily do so, but I think we can only ever reasonably load this data in memory by doing the scans we do for each provider. I feel like availability is more of a runtime thing.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Persisting the block ranges is a different thing than updating other plugins with the block ranges. It is there just to store the current state on a regular basis, so that the next time the block node starts, it is retrieved.

As I wrote in another comment, how plugins get the information is beyond the scope of this task

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: As discussed with you earlier, we don't aim to replace the historical block facility. So we don't keep per-block provider data.

I am also not sure about the reason why we keep track of available blocks anyway. I just got this from discussion with @jsync-swirlds:

> The block providers should be updating available blocks exclusively. The ASF should add available blocks to both range sets (stored and available), and stored blocks only to the stored blocks set.
>
> The ASF should be storing both available and stored ranges to disk every so often (configurable based on blocks-as-timer, perhaps 1000 blocks).

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: This task (and its parent) were focused on obtaining the information about the stored and available blocks.

As far as I can understand from other comments here, we are not sure how the data should be made available to the plugins: is it via push updates by ASF, or it is an on demand pattern: a method that exposes the data.

So, as I commented elsewhere, this seems to be beyond the scope of this task.

berryware: We should probably use this in serverDetailStatus API call. Currently we have just what the historical block facility says as available. But perhaps we want stored and available.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -13,26 +13,35 @@
...
+ * @param availableBlocksFilePath path where the set of available block ranges is persisted. Available blocks
+ *     are the subset of stored blocks that can also be served to clients (reported by a
+ *     {@code BlockProviderPlugin}). The file is written and loaded on the same schedule as
+ *     {@code storedBlocksFilePath}.
  * @param updateScanInterval The amount of milliseconds that the {@code ApplicationStateFacility} waits between
  *     checking to see if there are any {@code TssData} updates to process.
  * @param updateInitialDelay The amount of milliseconds that the {@code ApplicationStateFacility} waits before
  *     starting the scan interval.
  */
+// spotless:off
```
Nana-EC: Let's start dropping these spotless flips

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -13,26 +13,35 @@
...
 @ConfigData("app.state")
 public record ApplicationStateConfig(
-        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/app-state-data.bin")
-        Path dataFilePath,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/app-state-data.bin") Path tssDataFilePath,
```
Nana-EC: If we're renaming we should completely rename and add consistency with the rsa case
```suggestion
        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/tss-bootstrap-roster.json") Path tssBootstrapFilePath,
```

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -13,26 +13,35 @@
...
 
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/rsa-bootstrap-roster.json")
         Path rsaBootstrapFilePath,
 
-        @Loggable @ConfigProperty(defaultValue = "500") @Min(100)
-        long updateScanInterval,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/node/stored-blocks-data.bin")
```
Nana-EC: @berryware suggested we try to conform to JSON to make it easier for node operators to edit as needed

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -625,6 +682,37 @@ private void loadApplicationState(final Configuration configuration) {
                         e);
             }
         }
+
+        try {
```
Nana-EC: Seems odd place to create the dir.
Flows above check for file first and then I believe create the dir either at init() or before persistence.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -385,6 +399,21 @@ public void updateTssData(TssData tssData) {
...
+        // Every range is stored; only AVAILABLE ranges are both stored and available.
+        storedBlocks.add(blockRange);
+        if (blockRangeType == BlockRangeType.AVAILABLE) {
+            availableBlocks.add(blockRange);
+        }
+        // Trigger a persist whenever the running total crosses a BLOCK_RANGE_PERSIST_INTERVAL boundary.
+        final long blocksToAdd = blockRange.size();
+        final long before = storedBlockCount.getAndAdd(blocksToAdd);
+        if (before / BLOCK_RANGE_PERSIST_INTERVAL < (before + blocksToAdd) / BLOCK_RANGE_PERSIST_INTERVAL) {
+            persistBlockRanges();
```
Nana-EC: I believe the convention is to update a cached value and leave it to `checkForApplicationStateUpdates()` to handle persistence.
What you could to is have it call `persistBlockRanges()` and in `persistBlockRanges()` have the check for the interval of blocks stored if needed

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -486,6 +515,7 @@ void stopApplicationStateFacility() {
                 Thread.currentThread().interrupt();
             }
         }
+        persistBlockRanges();
```
Nana-EC, line 524: Interesting, wondering if this should be done for all state data?
@berryware 🤔

Nana-EC (commented):
A few minor items to correct, hopefully shouldn't take to long as I know this is blocking another item.
Feel free to triage and note which can be addressed in a follow up.


Another consideration is pull out the ApplicationStateFacility interface update and the BlockNodeContext update (which is missing from here) to it's own PR. That will allow other plugins to see it and you could work on this and the cloud storage improvement in parallel

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
ivannov: I thought that the agreement was to have them?

Anyway, I removed them.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
ivannov: Changed them

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
ivannov: Changed to JSON, also added to protobuf. With JSON I can use one structure for both ranged sets

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Moved the persistence to `checkForApplicationStateUpdates()`.

I left the check where it is (around the method call and not inside it) for two reasons:
* The behavior is the same in the other two values which we persist
* We call the persist method also in the stop method of the block node app and there we want to store it unconditionally

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: The persist call is now moved to the thread, which persist the other two values managed by the application state. The method is no longer synchronized.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: As discussed with @ata-nas , I added a getter method to the app state facility interface that returns the store block ranges

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: @nana, so only persist in stopApplicationStateFacility()? Currently we store when the data changes. In the event of an unplanned system exit, we will have the latest changes.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: I recv'd the same feedback and moved it to the load method. So if the file does not exist on load, make sure the directory is created.

ivannov: Followed the same pattern as yours, @berryware

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: @berryware I mean in addition.
So persistence would happen on changes and then on stop.
The need for stop is to to ensure that any in memory changes are captured.
In some cases we may find changes are fine in memory for some time and we periodically persist.
Something to consider in the design and apply if and when applicable

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
jsync-swirlds: We require the spotless disables in config or spotless mangles the file into something unreadable.
The team agreed that we'd prefer a clean and readable config files over slavish spotless adherence.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: Stored == Blocks that are known to be securely stored by this block node; mostly used to inform CN and backfill.
Available == Blocks that can be provided to a subscriber (which should generally be a subset of Stored).

If a block is available, it is assumed to be stored, while a stored block may, or may not, be available.

The reasoning for storing both available and stored on disk is twofold:
1. We should be able to report stored blocks at startup immediately, even if we update that shortly thereafter.
1. We don't currently track "deleted" blocks in a meaningful sense, we should probably work on making that sane and rational when we migrate from "special facilities" to plugin-to-plugin interactions.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: I think that adding a `get` method might be risky.
We've very strongly discouraged query/accessor methods in the App and facility interfaces because they are nightmarish when we remove these special facility interfaces from the App.
I would suggest an update method similar to how we update Context, but I also agree that making the data available to other plugins is _not_ properly part of this Issue/PR.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: The following is informational - to be used in future work.
As to how to manage many updates, I would go a step further than Atanas approach.
I would recommend gathering updates until some criteria (time, specific update, volume of updates, etc...) is met, then push a _single_ update representing the current state of the application data.
This reduces update frequency while also allowing for immediate response when needed.
We can workshop further in future work (probably when we extract all of this out of the app and move it into a proper plugin).

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -398,6 +417,21 @@ public void updateTssData(TssData tssData) {
...
+        if (blockRangeType == BlockRangeType.AVAILABLE) {
+            availableBlocks.add(blockRange);
+        }
+        storedBlockCount.addAndGet(blockRange.size());
+    }
+
+    @Override
+    public BlockRangeSet storedBlocks() {
+        return storedBlocks;
+    }
```
jsync-swirlds, line 424: I'm not sure adding this is wise.
I'd prefer we continue to use an event-driven approach rather than enable polling.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -398,6 +417,21 @@ public void updateTssData(TssData tssData) {
         if (tssData != null) tssDataUpdates.add(tssData);
     }
 
+    @Override
+    public void addBlockRange(LongRange blockRange, BlockRangeType blockRangeType) {
```
jsync-swirlds: FYI, I'm not sure why we chose to add an enum here rather than just having a second method for stored versus available.
It's not a _bad_ choice, but seems slightly over-engineered.  I do not believe we'll ever add more block "range type"s beyond the current two...

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ApplicationStateFacility.java
```diff
@@ -26,4 +34,28 @@ public interface ApplicationStateFacility {
...
+    void addBlockRange(LongRange blockRange, BlockRangeType blockRangeType);
+
+    /**
+     * Returns the set of block ranges that have been reported as stored.
+     *
+     * @return a {@link BlockRangeSet} of all stored block ranges; never null
+     */
+    default BlockRangeSet storedBlocks() {
+        return BlockRangeSet.EMPTY;
+    }
```
jsync-swirlds: I do not think defaulting this is ideal.
It would be better to have the compiler feedback when we forget to implement the method.

protobuf-sources/src/main/proto/block-node/api/node_service.proto
```diff
@@ -184,6 +184,22 @@ message BlockRange {
...
+    /**
+     * All block ranges that have been stored by this node.
+     */
+    repeated BlockRange stored_blocks = 1;
+    /**
+     * The subset of stored blocks that are currently available for retrieval.
+     */
+    repeated BlockRange available_blocks = 2;
+}
+
```
jsync-swirlds: This _does not_ belong in the block node API.
The BlockRangesState should be a new file in the _internal_ protos so we do not pollute public API and are not constrained to the public API rules for this internal data structure.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Removed the accessor method

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Introduced two `add` methods and got rid of the enum

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Removed the accessor

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ApplicationStateFacility.java
ivannov: The method is removed altogether

protobuf-sources/src/main/proto/block-node/api/node_service.proto
ivannov: Introduced a new internal proto file

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
ivannov: I restored the spotless off and on

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Created a task to track how data reaches interested plugins: #2852

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -452,7 +483,10 @@ void startApplicationStateFacility() {
 
         // Schedule periodic check for live updates from running plugins.
         applicationStateExecutor.scheduleAtFixedRate(
-                this::checkForApplicationStateUpdates, 0, appStateConfig.updateScanInterval(), TimeUnit.MILLISECONDS);
+                this::checkForApplicationStateUpdates,
+                appStateConfig.updateInitialDelay(),
```
Nana-EC, line 481: This is intentionally 0 to ensure there's an immediate check following this
```suggestion
                0,
```

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -398,6 +416,19 @@ public void updateTssData(TssData tssData) {
         if (tssData != null) tssDataUpdates.add(tssData);
     }
 
+    @Override
+    public void addStoredBlockRange(LongRange blockRange) {
+        storedBlocks.add(blockRange);
+        storedBlockCount.addAndGet(blockRange.size());
```
ata-nas: I think the `storedBlockCount` must be the same as the `storedBlocks.size()`

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: The thing is that the available blocks, as data already exists in the Historical Block Facility. It's there where block provider plugins make updates to their available blocks ranged sets and the facility combines these. For available blocks, that is our source of truth. So I think it is there, where we must get our data for available blocks from. We can persist it, sure, but also, we must think about how each plugin, having an individual set, will notify the application state facility of that update. That is the tricky part. When a plugin adds/removes from its ranged set, that update is invisible to the application state facility as of today. The historical block facility is not reactive, but when called, it will look for registered sets in the combined ranged set.

So my concern is that we have to make these updates visible.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -107,6 +113,18 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
...
+    final ConcurrentLongRangeSet storedBlocks = new ConcurrentLongRangeSet();
+
+    /** Blocks reported as available by BlockProviderPlugin implementations */
+    final ConcurrentLongRangeSet availableBlocks = new ConcurrentLongRangeSet();
+
+    /** Running total of stored block numbers, used to trigger periodic persistence */
+    private final AtomicLong storedBlockCount = new AtomicLong(0);
+
+    /** Block count at the time of the last scheduled persist; only read/written by the scanner thread */
+    private long lastPersistedBlockCount = 0;
```
ata-nas, line 122: Does this need to be volatile? If really only read by the same thread, then we should be fine.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -398,6 +416,19 @@ public void updateTssData(TssData tssData) {
...
+    public void addStoredBlockRange(LongRange blockRange) {
+        storedBlocks.add(blockRange);
+        storedBlockCount.addAndGet(blockRange.size());
+    }
+
+    @Override
+    public void addAvailableBlockRange(LongRange blockRange) {
+        storedBlocks.add(blockRange);
+        availableBlocks.add(blockRange);
+        storedBlockCount.addAndGet(blockRange.size());
```
ata-nas: This method could call the addStoredBlockRange.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -584,6 +626,39 @@ private void persistNodeAddressBook(NodeAddressBook nodeAddressBook) {
...
+                .configuration()
+                .getConfigData(ApplicationStateConfig.class)
+                .blockRangesFilePath();
+        try {
+            final Path tmp = filePath.resolveSibling(filePath.getFileName() + ".tmp");
+            final Bytes json = BlockRangesState.JSON.toBytes(toBlockRangesState());
+            Files.write(tmp, json.toByteArray());
+            Files.deleteIfExists(filePath);
+            Files.createLink(filePath, tmp);
+            Files.deleteIfExists(tmp);
```
ata-nas, line 637: This seems a bit fragile. We want to overwrite the data, not delete it and then create it anew. I think we can be more careful about this. Possibly making a new file with the latest stored block as name and then deleting the old one. On startup, we look for the most recent one in case 2 or more and deleting the rest.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -584,6 +626,39 @@ private void persistNodeAddressBook(NodeAddressBook nodeAddressBook) {
...
+            LOGGER.log(INFO, "Persisted block ranges to file: {0}", filePath);
```
ata-nas, line 638: Not sure of the value of this log message. But it is also infrequent, so should not be disruptive.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -643,6 +718,28 @@ private void loadApplicationState(final Configuration configuration) {
...
+                rangeSet.availableBlocks()
+                        .forEach(r -> availableBlocks.add(new LongRange(r.rangeStart(), r.rangeEnd())));
+                LOGGER.log(INFO, "Loaded block ranges from file: {0}", blockRangesPath);
+            } catch (ParseException | IOException | IllegalArgumentException e) {
+                LOGGER.log(ERROR, "Failed to read block ranges file: " + blockRangesPath, e);
+            }
+        } else {
+            Path parent = blockRangesPath.getParent();
+            try {
+                Files.createDirectories(parent);
```
ata-nas, line 732: createDirectories could be executed regardless on startup.

ata-nas (commented):
Cleanup suggestions and a few clarifications.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: This is only read and written inside the `checkForApplicationStateUpdates` method, which in turns is called inside the `scheduleAtFixedRate` callback, which runs on the single-threaded `applicationStateExecutor`.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: I think there is a slight misunderstanding, that might be on me (maybe I did not ask the question good enough). I mean how do we get the ranged sets themselves, not the block (or accessors for them). An update to the context will be consistent with the current setup, but thinking more about the future, adding a getter for the ranged sets (what's inside them) could also be an option.

Now, if we rely on the onContextUpdate call, we will be able to easily detect changes, so that might be the correct way to go.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: That is ok, not opposed to it, just making a note in case it is of some value.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: I copied this approach from the block files historic plugin, because we spent there a lot of time discussing what is the safest way to overwrite a file with another one

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: I followed the pattern from other persistence methods in the ASF

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: As suggested on previous pass comment, I followed the pattern from other files' handling

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Right, I removed the `storedBlockCount` field

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: You are right, changed that

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: There's already a task to dig deeper into this: https://github.com/hiero-ledger/hiero-block-node/issues/2852

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: Just worried about some fragility here. If we are ok, we can proceed. Not sure where we copy from, seem to be  missing something.

ivannov: A few lines above we create the tmp file (a pattern from other persist methods in this class):

```java
final Path tmp = filePath.resolveSibling(filePath.getFileName() + ".tmp");
```

Then we write the ranges in the tmp file:

```java
final Bytes json = BlockRangesState.JSON.toBytes(toBlockRangesState());
Files.write(tmp, json.toByteArray());
```

And finally we delete the original file, create a link to it from the tmp file and delete the tmp file (a pattern that I took from the historic file plugin, where we discussed it back and forth before we decided that this is the way):

```java
Files.deleteIfExists(filePath);
Files.createLink(filePath, tmp);
Files.deleteIfExists(tmp);

```

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: That is OK, can tackle this in another PR.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: captured in [#2872](https://github.com/hiero-ledger/hiero-block-node/issues/2872)

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -452,7 +477,10 @@ void startApplicationStateFacility() {
 
         // Schedule periodic check for live updates from running plugins.
         applicationStateExecutor.scheduleAtFixedRate(
-                this::checkForApplicationStateUpdates, 0, appStateConfig.updateScanInterval(), TimeUnit.MILLISECONDS);
+                this::checkForApplicationStateUpdates,
+                appStateConfig.updateInitialDelay(),
```
berryware, line 481: Update InitialDelay has been removed. The initial state check should be immediate with 0.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: We need to move the "available" and "stored" blocks to another location (perhaps application state).
The provider was created early on without thinking through how we would manage similar needs across multiple plugins.

Something to discuss further in the future, certainly.

jsync-swirlds (approved):
Nothing blocking here.
Definite opportunities for follow-up.

berryware (approved):

ata-nas (approved):
@ivannov please resolve comments at your discretion. 

I still have some comments that want to discuss as to how we will actually integrate this, but we can proceed as this is now not used and should not be blocking.

Nana-EC (approved):
LG
Opened ticket with follows ups https://github.com/hiero-ledger/hiero-block-node/issues/2875.
You can resolved comments and merge.

PR #2740, opened 2026-05-07
test: fix flaky notificationsPublishedInAscendingBlockOrder test

### LINKED ISSUE #2739: Flaky Test ExpandedCloudStoragePluginTest notificationsPublishedInAscendingBlockOrder
### Description

The test cases confirms that "PersistedNotifications are published in ascending block-number order when results are drained together"
Test case seems to periodically fails on the check for "Block 2 must be published first even though it was submitted second" with a value not 2. In a recent case it was 7

### Steps to Reproduce

1. Run test in CI

### Additional context

_No response_

Nana-EC:
## Reviewer Notes
Test notificationsPublishedInAscendingBlockOrder has 2 race conditions:

1. (Primary — the reported failure) handleVerification() calls drainCompletedTasks() before submitting each new task. Because virtual threads are eager, block 7's task could complete its full call() path and be queued in the CompletionService before handleVerification(2) was called. The internal drain inside handleVerification(2) would then publish block 7 immediately — before block 2 was submitted — producing notifications in [7, 2] order instead of [2, 7].

2. (Secondary) countDown() fired inside uploadFile(), which is mid-way through SingleBlockStoreTask.call(). On slow CI, Thread.sleep(50) was not long enough for both virtual threads to finish building UploadResult and have their futures queued in the CompletionService before drainCompletedTasks() was called.

Fix:
- replace the single CountDownLatch + Thread.sleep(50) with a two-latch barrier. uploadsStarted counts down when each task enters uploadFile();
- releaseUploads holds both tasks blocked inside uploadFile() until the test thread signals. This guarantees block 7's task cannot complete before block 2 is submitted, eliminating the primary race. After releasing both tasks simultaneously they queue their futures within nanoseconds of each other;
- awaitNotifications(2) retries with a 10ms sleep so a single drain pass reliably sees both and publishes them in ascending block-number order.
- add a timeout for the test case since it's async

## Related Issue(s)
Closes #2739

### DIFF
--- a/block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
+++ b/block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
@@ -634,7 +634,6 @@ class ExpandedCloudStoragePluginTest
     @Test
     @DisplayName(
             "PersistedNotifications are published in ascending block-number order when results are drained together")
-    @Timeout(value = 10, unit = TimeUnit.SECONDS)
     void notificationsPublishedInAscendingBlockOrder() throws InterruptedException {
         // Two-latch barrier: hold each upload INSIDE uploadFile until both tasks are
         // simultaneously in-flight. This prevents block 7's task from completing before
@@ -651,9 +650,12 @@ class ExpandedCloudStoragePluginTest
                     final String contentType) {
                 uploadsStarted.countDown();
                 try {
-                    releaseUploads.await(5, TimeUnit.SECONDS);
+                    if (!releaseUploads.await(5, TimeUnit.SECONDS)) {
+                        throw new AssertionError("Timed out waiting for releaseUploads barrier");
+                    }
                 } catch (final InterruptedException e) {
                     Thread.currentThread().interrupt();
+                    throw new AssertionError("Interrupted while awaiting releaseUploads barrier", e);
                 }
             }
 
--- a/block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
+++ b/block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
@@ -635,22 +635,33 @@ class ExpandedCloudStoragePluginTest
     @DisplayName(
             "PersistedNotifications are published in ascending block-number order when results are drained together")
     void notificationsPublishedInAscendingBlockOrder() throws InterruptedException {
-        final CountDownLatch bothUploaded = new CountDownLatch(2);
-        final S3UploadClient countingClient = new S3UploadClient() {
+        // Two-latch barrier: hold each upload INSIDE uploadFile until both tasks are
+        // simultaneously in-flight. This prevents block 7's task from completing before
+        // block 2 is submitted — if block 7 finishes first, handleVerification(2)'s
+        // internal drainCompletedTasks() would publish block 7 ahead of block 2.
+        final CountDownLatch uploadsStarted = new CountDownLatch(2);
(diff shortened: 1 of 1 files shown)


block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
```diff
@@ -634,23 +634,34 @@ public void close() {}
     @Test
     @DisplayName(
             "PersistedNotifications are published in ascending block-number order when results are drained together")
+    @Timeout(value = 10, unit = TimeUnit.SECONDS)
```
ata-nas: We could consider to put this timeout on class level. This way, all tests will inherit it.

block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
```diff
@@ -634,23 +634,34 @@ public void close() {}
...
                     final String storageClass,
                     final Iterator<byte[]> contentIterable,
                     final String contentType) {
-                bothUploaded.countDown();
+                uploadsStarted.countDown();
+                try {
+                    releaseUploads.await(5, TimeUnit.SECONDS);
+                } catch (final InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                }
```
ata-nas, line 659: You could also throw here, junit should be able to detect it and kill the test. A simple declaration might be cleaner and we do not allow the test to continue.

block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
```diff
@@ -634,23 +634,34 @@ public void close() {}
...
             @Override
             public void uploadFile(
                     final String objectKey,
                     final String storageClass,
                     final Iterator<byte[]> contentIterable,
                     final String contentType) {
-                bothUploaded.countDown();
+                uploadsStarted.countDown();
+                try {
+                    releaseUploads.await(5, TimeUnit.SECONDS);
```
ata-nas: If await fails, I think it returns false. On false, maybe we should simply `fail()`.

ata-nas (dismissed):
LG, a couple of improvement suggestions that should improve overall usage.

block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
Nana-EC: The class indeed already has @Timeout(value = 30, unit = TimeUnit.SECONDS).
The per-test @Timeout(10s) was a stricter override but I removed it since the class level should be sufficient

block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
Nana-EC: I throw an assertion re-interrupt the thread now which will encourage a failure and JUnit will kill the test

block-node/cloud-storage-expanded/src/test/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePluginTest.java
Nana-EC: I added a `throw AssertionError("Timed out waiting for releaseUploads barrier")` so the test fails

ata-nas (approved):

berryware (approved):
Looks Good

AlfredoG87 (approved):

PR #2744, opened 2026-05-07
fix: Ensure the handler shuts down if sendResponse fails

### LINKED ISSUE #2743: Bug(publisher): Close inactive / closed publisher handlers so we stop trying to send ACKs to them
### Description

`LiveStreamPublisherManager.handlers` accumulates entries for publishers whose HTTP/2 channel has already closed. Every subsequent `handlePersisted` fans out to those dead entries, throws `ClosedChannelException`, logs a stack trace, and leaves the entry in the map. Result: `activePublishers` is inflated (s03 reported 13 in a 7-CN network during the 2026-05-07 incident), every persistence event burns work writing to dead sockets, and operator-facing health metrics lie.

The actual code in `PublisherHandler.sendResponse` catches the channel-closed exception, increments a metric, and returns false — it neither flips `isActive` nor calls `publisherManager.removeHandler`. So removal only happens via paths that need a live remote peer (an inbound request that finds `isActive == false`, or a peer-sent `END_STREAM` round-trip), neither of which ever arrives once the channel is dead.

## Evidence (2026-05-07 incident)

In `logs/s01.log` at 02:14:50.482, a single `handlePersisted(38267)` produced:

- 3 successful `ACKNOWLEDGEMENT` sends
- 6+ `Failed to send response 'ACKNOWLEDGEMENT' for handler N: java.io.UncheckedIOException: java.nio.channels.ClosedChannelException` stack traces

On s03, peak `activePublishers=13` was reported in a 7-CN network. `activePublishers` is computed as `handlers.size()` in `sendPublisherStatusUpdate` (`LiveStreamPublisherManager.java:677-683`), so 13 handlers in `handlers` for 7 CNs means 6+ are dead entries.

## Root cause

`PublisherHandler.sendResponse` (`PublisherHandler.java:604-636`):

```java
private boolean sendResponse(final PublishStreamResponse response) {
    try {
        long start = System.nanoTime();
        replies.onNext(response);
        ...
        return true;
    } catch (final UncheckedIOException e) {
        // "Publisher closed the connection unexpectedly..."
        LOGGER.log(DEBUG, message, e);
        metrics.sendResponseFailed.increment();
        return false;            // ← no removeHandler, no isActive flip
        ...
        return false;            // ← same
    }
}
```

The `UncheckedIOException` catch is exactly the channel-closed path (`io.helidon.common.socket.NioSocket.write` wraps `ClosedChannelException` as `UncheckedIOException`). Today it logs and returns false — the handler entry stays in `LiveStreamPublisherManager.handlers`.

Lifecycle of a ghost handler today:

```
peer disappears → next outbound write fails (UncheckedIOException) → catch returns false
               → no removeHandler call
               → no isActive flip
               → next handlePersisted re-attempts the write → repeat forever
```

## Affected code

- `block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java`
  - L604-636 — `sendResponse` catch blocks (no cleanup today)
- `block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java`
  - L170-181 — `removeHandler` (the only public removal path; nothing currently calls it from a write-failure context)
  - L398 — `handlePersisted` fan-out site that re-iterates dead handlers

### A. Eager removal on channel-closed (the core fix)

In `PublisherHandler.sendResponse`'s `UncheckedIOException` catch, call `publisherManager.removeHandler(handlerId)` and flip `isActive` to false before returning. The handler is dead — don't leave it in the map waiting for an inbound message that won't come.

Sketch:

```java
} catch (final UncheckedIOException e) {
    LOGGER.log(DEBUG, message, e);
    metrics.sendResponseFailed.increment();
    isActive.set(false);                          // ← bail any in-flight handler thread
    publisherManager.removeHandler(handlerId);    // ← eager cleanup
    return false;
}
```

### B. Defensive filter in fan-out

Add `.filter(PublisherHandler::isActive)` to every `handlers.values().parallelStream()` site in `LiveStreamPublisherManager`. Belt-and-suspenders — even if removal hasn't reached the map yet, an inactive handler shouldn't get sent ACKs. This also reduces interaction with issue 4: a handler whose channel closed mid-block can't accidentally become an "in-flight resend" target if it's filtered out before fan-out.

### C. Reduce log noise

When eager-remove fires from `sendResponse`, log once at INFO with handler id and correlation id (e.g. `"handler N (Nx-STRy) removed: peer closed connection"`) so operators can see disconnects. The DEBUG line already exists; add a single INFO summarizing the removal so log volume goes from one stack trace per future fan-out attempt down to one INFO per disconnect.

### D. (Optional, longer-term) periodic reaper

A scheduled task that scans `handlers` every N seconds and removes entries where `isActive == false` and last successful write was older than threshold. Catches handlers that somehow escape A. Not required for this fix; mention as future work.

## Reproduction (regression test)

Unit-testable in `LiveStreamPublisherManagerTest`:

1. Construct manager.
2. Add a handler whose injected `Pipeline` throws `UncheckedIOException(new ClosedChannelException())` on the next `onNext`.
3. Fire `handlePersisted` for a higher block.
4. **Assert**: handler is removed from `handlers` after the first failing send; `handlers.size()` decremented; `currentPublisherCount` reflects the removal; `sendPublisherStatusUpdate(PUBLISHER_DISCONNECTED, ...)` was sent once.
5. **Assert**: a second `handlePersisted` does not iterate the now-removed handler (no second failing send).

Add a second test for the defensive filter (B): construct a handler with `isActive=false` (without removing it from the map), call `handlePersisted`, assert the handler did not receive a send attempt.

## Severity

**P1 / correctness-adjacent.** Does not directly corrupt persisted blocks, but:

- Operators have an incorrect view of cluster health (`activePublishers` is `handlers.size()` and includes ghosts).
- Widens the race window for issue 4 (spurious ACK after RESEND): a ghost handler that "ACCEPTs" a header before the writer notices the dead channel removes a block from `blocksToResend` and leaves the manager exposed.
- Wastes work and produces large volumes of stack-trace log noise on every persistence event.

Should land alongside issue 4 since both touch `handlePersisted` and both interact with handler lifecycle.

## Open questions for the implementer

- **Reentrance.** Calling `publisherManager.removeHandler` from `sendResponse` reaches `synchronized sendPublisherStatusUpdate` and a notification through the messaging facility. The `handlePersisted` fan-out is a `parallelStream` — is it OK for a worker thread to make this call from inside the fan-out? Should be, since `removeHandler` doesn't take any lock the fan-out depends on, but worth checking once on a real run.
- **Idempotence.** Two write paths could fail concurrently for the same handler. `handlers.remove(handlerId)` is idempotent, but `metrics.currentPublisherCount.set` and `sendPublisherStatusUpdate` fire unconditionally inside `removeHandler`. Double-firing is benign for both. If you want strictness, gate them on `handlerRemoved != null`.
- **Metrics**. Worth adding a separate counter for handler removals attributed to `ClosedChannelException` vs. clean shutdown vs. timeout? Distinguishes peer-crash rate from graceful-disconnect rate in production. Not blocking for this fix.

jsync-swirlds:
## Reviewer Notes
* Fixed the "else" handling for sendAcknowledgement to shut down immediately instead of scheduling a shutdown for a future onNext (which will never be called).
* Fixed some spotless issues
* Fixed some tests that were testing the protocol wrong
   * No idea why these haven't been failing before this PR...

## Related Issue(s)
Fixes: #2743

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
@@ -46,6 +46,7 @@ import org.hiero.block.api.PublishStreamResponse.BlockAcknowledgement;
 import org.hiero.block.api.PublishStreamResponse.EndOfStream;
 import org.hiero.block.api.PublishStreamResponse.EndOfStream.Code;
 import org.hiero.block.api.PublishStreamResponse.ResendBlock;
+import org.hiero.block.api.PublishStreamResponse.ResponseOneOfType;
 import org.hiero.block.api.PublishStreamResponse.SkipBlock;
 import org.hiero.block.internal.BlockItemSetUnparsed;
 import org.hiero.block.internal.BlockItemUnparsed;
@@ -74,6 +75,12 @@ import org.hiero.metrics.core.MetricRegistry;
 /// block, the action for it might be different that what is usually expected,
 /// especially true in a multi-publisher environment.
 public final class PublisherHandler implements Pipeline<PublishStreamRequestUnparsed> {
+    // Note: not surrounding correlationId={3} with [] intentionally as it will break Loki's parsing.
+    private static final String LATENCY_END_METRIC_MESSAGE =
+            "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId={3}";
+    private static final String LATENCY_START_MESSAGE =
+            "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId={3}";
+
     private final Logger LOGGER = System.getLogger(getClass().getName());
     /// The replies pipeline, used for sending responses to the publisher.
     private final Pipeline<? super PublishStreamResponse> replies;
@@ -211,21 +218,17 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
     /// is sent as a response to the publisher, indicating that all blocks up to
     /// and including the given block number are safely stored in this block node.
     ///
-    /// @param newLastAcknowledgedBlockNumber the last block number that was
+    /// @param blockToAcknowledge the last block number that was
     ///     verified and persisted.
-    public void sendAcknowledgement(final long newLastAcknowledgedBlockNumber) {
-        LOGGER.log(
-                TRACE,
-                "[{0}] Handler {1} sending acknowledgement for block {2}",
-                correlationIdPrefix,
-                handlerId,
-                newLastAcknowledgedBlockNumber);
+    public void sendAcknowledgement(final long blockToAcknowledge) {
+        final String ackTraceStartMessage = "[{0}] Handler {1} sending acknowledgement for block {2}";
+        LOGGER.log(TRACE, ackTraceStartMessage, correlationIdPrefix, handlerId, blockToAcknowledge);
         // We only ever need to acknowledge once for a given block number, even
         // if there are several blocks "behind" that acknowledgement.
         // The publishers expect that acknowledgement for block N implicitly
         // acknowledges all blocks up to and including N.
         final BlockAcknowledgement ack = BlockAcknowledgement.newBuilder()
-                .blockNumber(newLastAcknowledgedBlockNumber)
+                .blockNumber(blockToAcknowledge)
                 .build();
         final PublishStreamResponse response =
                 PublishStreamResponse.newBuilder().acknowledgement(ack).build();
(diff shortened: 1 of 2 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -594,7 +597,7 @@ private void sendEndOfStream(final Code codeToSend) {
                 PublishStreamResponse.newBuilder().endStream(endOfStream).build();
         if (sendResponse(response)) {
             metrics.endOfStreamsSent.increment(); // @todo(1415) add label
-        }
+        } // no else here, we expect to be shut down regardless...
```
ata-nas, line 577: Yes, this method must always be followed by a shutdown, so no else here, as described also by the javadoc.

ata-nas (dismissed):
LG

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherHandlerTest.java
```diff
@@ -749,24 +749,23 @@ void testOnNextResponseRESEND() {
...
                 assertThat(repliesPipeline.getOnNextCalls())
                         .hasSize(1)
                         .first()
-                        .returns(ResponseOneOfType.END_STREAM, responseKindExtractor)
-                        .returns(Code.ERROR, endStreamResponseCodeExtractor)
-                        .returns(latestBlockNumber, endStreamBlockNumberExtractor);
-                assertThat(repliesPipeline.getOnCompleteCalls().get()).isEqualTo(1);
+                        .returns(ResponseOneOfType.RESEND_BLOCK, responseKindExtractor)
+                        .returns(block.number(), resendBlockNumberExtractor);
+                assertThat(repliesPipeline.getOnCompleteCalls().get()).isEqualTo(0);
```
ata-nas, line 768: The test was correct before. RESEND is not expected as a block action on a BlockItems request. This is why, if it is received as an action (should never happen), the response should be to end the stream.

ata-nas: There should be other tests that expect the resend response.

AlfredoG87 (approved):
LGTM

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublisherHandlerTest.java
jsync-swirlds: This test was setting up the code to send a resend _before_ end of block, which we don't do, we schedule the resend back to CN for after the block is closed.
Basically, either the test send end-of-block on a closed stream (not possible) or tried to validate the wrong response sequence, or both.

It was not correct before either.

I'll adjust in a separate PR after we've had a chance to discuss what it _should_ be actually doing.

jsync-swirlds: This test bypasses the actual action handling with a fixture, so it doesn't actually test some of the correct handling code.

rockysingh (approved):
lgtm

PR #2751, opened 2026-05-07
docs: add comprehensive WRB CLI operations runbook

### LINKED ISSUE #2710: docs: add runbook for block tools CLI (WRB) setup and operational workflows
## Description

Create a comprehensive runbook documenting the block tools CLI setup and operational workflows. This will be used by other teams (e.g. when running processes on HashSphere) who need to work with wrapped block streams.

### Setup
- Server requirements (CPU, memory, disk, ZFS recommendations)
- Building the shadow jar from source
- Required metadata files and how to generate them (address book history, block times, day listings)
- Network configuration (mainnet vs testnet)

### Core Workflows

**1. Full Mainnet Validation**
- How to run `blocks validate` end to end
- Recommended JVM flags (`-Xms`, `-Xmx`, `-XX:+UseZGC`, etc.)
- Checkpoint/resume behaviour
- Available flags (`--skip-supply`, `--skip-signatures`, `--verbose`, `--threads`, `--prefetch`)
- Interpreting the timing breakdown and speed multiplier
- Troubleshooting common errors

**2. Downloading Block Data**
- `days download-days` workflow for bulk historical download
- `days live-sequential` for live streaming with inline validation
- `days updateDayListings` for refreshing GCS metadata
- Handling corrupt/missing files and the `blocks repair-zips` command

**3. Wrapping Record Files**
- `blocks wrap` to convert record file blocks into wrapped block streams
- Expected directory structure and naming conventions

**4. Mirror Node Data**
- Generating address book history from mirror node
- Fetching block times and day block mappings
- `metadata update` for refreshing all metadata files

### Operational Notes
- How checkpoint save/resume works
- Graceful shutdown (`kill` vs `kill -9`)
- Monitoring progress and interpreting output
- Disk space requirements and cleanup
- Running validation with `nohup` for long runs

## Acceptance Criteria

- [ ] Runbook is clear enough for someone unfamiliar with the tooling to set up and run a full validation
- [ ] Includes example commands with real paths/flags
- [ ] Documents known gotchas (e.g. unknown file types in GCS, HBAR supply state save failures)

rockysingh:
## Summary

Adds a comprehensive operations runbook for the Wrapped Record Block (WRB) CLI tools, documenting setup and operational workflows for running block stream validation, migration, and archival processes.

### New File
- `docs/block-node/operations/wrb-cli-runbook.md` - 700+ line runbook covering all WRB CLI operations

## Contents

**Setup**:
- Server requirements (CPU, memory, disk, ZFS)
- Building the shadow jar
- Directory structure and network configuration

**Core Workflows**:
1. **Metadata Generation** - address book, block times, day listings
2. **Downloading** - bulk historical (`download-days-v3`) and live streaming (`live-sequential`)
3. **Wrapping** - converting record files to wrapped block format
4. **Validation** - full integrity checks with various flags
5. **Repair** - fixing corrupt or missing files

**Operational Notes**:
- Monitoring progress and interpreting output
- Disk space requirements and cleanup
- Graceful shutdown and checkpoint behavior
- Running with `nohup` for long operations

**Troubleshooting**:
- Unknown file types in GCS
- HBAR supply mismatches (testnet known issue documented)
- Block proof signature failures
- OOM errors
- Download timeouts

**Real-World Examples**:
- Mainnet production setup with cron jobs and systemd
- Testnet CI/CD validation pipeline
- Complete command reference

Closes #2710

### DIFF
--- a/docs/block-node/operations/wrb-cli-runbook.md
+++ b/docs/block-node/operations/wrb-cli-runbook.md
@@ -533,10 +533,10 @@ ls -d wrappedBlocks/*/ | wc -l
 
 ### Graceful Shutdown
 
-**SIGTERM (recommended)**:
+**SIGINT (recommended)**:
 
 ```bash
-kill <PID>
+kill -INT <PID>
 ```
 
 - Allows checkpoint save
--- a/docs/block-node/operations/wrb-cli-runbook.md
+++ b/docs/block-node/operations/wrb-cli-runbook.md
@@ -668,7 +668,7 @@ java -Xms20g -Xmx20g -XX:+UseZGC -jar tools.jar --network testnet blocks validat
 ### Mainnet Production Setup
 
 **Hardware**:
-- 32-core Xeon server
+- 32-core server
 - 128GB RAM
 - 20TB ZFS RAID10 SSD array
 
--- a/docs/block-node/operations/wrb-cli-runbook.md
+++ b/docs/block-node/operations/wrb-cli-runbook.md
@@ -708,10 +708,3 @@ fi
 ```
 
 ---
-
-## Support
-
-For issues or questions:
-- **GitHub Issues**: https://github.com/hiero-ledger/hiero-block-node/issues
-- **Discord**: Hiero developer community
-- **Logs**: Always include relevant log excerpts and command arguments when reporting issues
--- a/docs/block-node/operations/wrb-cli-runbook.md
+++ b/docs/block-node/operations/wrb-cli-runbook.md
@@ -1,0 +1,717 @@
+# WRB CLI Operations Runbook
+
+## Overview
+
+This runbook documents the setup and operational workflows for the Wrapped Record Block (WRB) CLI tools. These tools are used to download, wrap, validate, and maintain block streams for Hiero networks (mainnet and testnet).
+
+**Audience**: Operators running block stream validation, migration, or archival processes.
+
+---
+
+## Prerequisites
+
+### Server Requirements
+
+**Minimum Specifications:**
+- **CPU**: 16+ cores (validation benefits from parallelization)
+- **Memory**: 32GB+ RAM (recommend 64GB for large-scale operations)
+- **Disk**: 10TB+ SSD/NVMe storage
+  - Recommend ZFS for data integrity and compression
+  - Separate volumes for compressed days, wrapped blocks, and metadata
+- **Network**: High bandwidth (100Mbps+) for GCS downloads
+
+**Operating System**: Linux (Ubuntu 20.04+ or RHEL 8+ recommended)
+
+### Required Tools
+
+```bash
+# Install Java 21+
+sudo apt-get update
+sudo apt-get install openjdk-21-jdk
+
+# Install zstd (required for decompression)
+sudo apt-get install zstd
+
+# Verify installations
+java -version
+zstd --version
+```
+
+### Building the Shadow JAR
+
+```bash
+# Clone the repository
+git clone https://github.com/hiero-ledger/hiero-block-node.git
+cd hiero-block-node
+
+# Build the tools shadow jar
+./gradlew :tools:shadowJar
+
+# Locate the jar
+ls tools-and-tests/tools/build/libs/tools-*-all.jar
+```
+
+The shadow jar is self-contained and can be copied to your deployment server.
+
+---
+
+## Directory Structure
+
+Recommended layout for operations:
+
+```
+/mnt/wrb-operations/
+├── tools-<version>-all.jar          # WRB CLI jar
+├── metadata/                         # Metadata files
+│   ├── block_times.bin               # Block number → timestamp mapping
+│   ├── day_blocks.json               # Day → block range mapping
+│   └── listingsByDay/                # GCS file listings per day
+│       ├── 2024-01-01.json
+│       └── ...
+├── compressedDays/                   # Downloaded record files (compressed)
+│   ├── 2024-01-01/
+│   └── ...
+└── wrappedBlocks/                    # Wrapped block stream output
+    ├── addressBookHistory.json       # Address book changes over time
+    ├── 0/                            # Block 0
+    ├── 1/                            # Block 1
+    └── ...
+```
+
+---
+
+## Network Configuration
+
+### GCP Authentication (for GCS access)
+
+Export GCP credentials before running any commands:
+
+```bash
+export GOOGLE_APPLICATION_CREDENTIALS="/path/to/gcp-key.json"
+export GOOGLE_CLOUD_PROJECT="your-project-id"
+export GCP_PROJECT_ID="your-project-id"
+```
+
+### Network Selection
+
+All commands accept `--network <mainnet|testnet>`:
+
+- **Mainnet**: `--network mainnet` (default if omitted)
+- **Testnet**: `--network testnet`
+
+---
+
+## Core Workflows
+
+### 1. Generate Address Book History
+
+**Purpose**: Fetch address book changes from the mirror node API to enable block proof verification.
+
+**Command**:
+
+```bash
+nohup java -jar tools-<version>-all.jar \
+  --network testnet \
+  mirror generateAddressBook \
+  -o wrappedBlocks/addressBookHistory.json \
+  --show-changes \
+  > addressbook.log 2>&1 &
+```
+
+**Output**: `wrappedBlocks/addressBookHistory.json` contains all address book updates.
+
+**Notes**:
+- Run this before starting wrapping or validation
+- Re-run periodically to capture new address book changes
+
+---
+
+### 2. Update Metadata Files
+
+Metadata files are essential for mapping blocks to days and timestamps.
+
+#### Generate Block Times and Day Blocks
+
+```bash
+nohup java -jar tools-<version>-all.jar \
+  --network testnet \
+  metadata update \
+  --listing-dir metadata/listingsByDay \
+  > metadata-update.log 2>&1 &
+```
+
+**Outputs**:
+- `metadata/block_times.bin` - binary map of block numbers to consensus timestamps
+- `metadata/day_blocks.json` - JSON map of dates to block ranges
+
+**For bounded updates** (e.g., stopping at a specific date):
+
+```bash
+java -jar tools-<version>-all.jar \
+  --network testnet \
+  metadata update \
+  --block-times metadata/block_times.bin \
+  --day-blocks metadata/day_blocks.json \
+  --end-date 2026-03-13
+```
+
+#### Update Day Listings
+
+```bash
+nohup java -jar tools-<version>-all.jar \
+  --network testnet \
+  days updateDayListings \
+  --listing-dir metadata/listingsByDay \
+  > update-day-listings.log 2>&1 &
+```
+
+**What it does**: Queries GCS for all record files per day and caches the listings locally.
+
+**Fix corrupt/missing day**:
+
+```bash
+java -jar tools-<version>-all.jar \
+  --network testnet \
+  days updateDayListings \
+  --day 2026-04-03
+```
+
+---
+
+### 3. Download Block Data
+
+#### Bulk Historical Download (`download-days-v3`)
+
+**Purpose**: Download compressed record files for a specific date range.
+
+**Command**:
+
+```bash
+nohup java -jar tools-<version>-all.jar \
+  --network testnet \
+  days download-days-v3 \
+  --listing-dir metadata/listingsByDay \
+  --downloaded-days-dir compressedDays \
+  --threads 600 \
+  2024 02 01 \
+  2026 03 13 \
+  > download-days.log 2>&1 &
+```
+
+**Arguments**:
+- `--listing-dir`: Path to day listings metadata
+- `--downloaded-days-dir` (or `-d`): Output directory
+- `--threads` (or `-t`): Parallelism (recommend 100-600 based on server capacity)
+- Date range: `YYYY MM DD YYYY MM DD` (start and end, inclusive)
+
+**Example - Single Day**:
+
+```bash
+nohup java -jar tools-<version>-all.jar \
+  --network testnet \
+  days download-days-v3 \
+  --listing-dir metadata/listingsByDay \
+  -d compressedDays \
+  -t 100 \
+  2026 04 03 \
+  2026 04 03 \
+  > download-single-day.log 2>&1 &
+```
+
+**Monitoring Progress**:
+
+```bash
+tail -f download-days.log
+```
+
+**Resuming**: The command automatically resumes from where it left off if interrupted.
+
+---
+
+#### Live Streaming Download (`live-sequential`)
+
+**Purpose**: Continuously download, wrap, and validate blocks in near real-time.
+
+**Command**:
+
+```bash
+nohup java -jar tools-<version>-all.jar \
+  --network testnet \
+  days live-sequential \
+  -l metadata/listingsByDay \
+  -o compressedDays \
+  --wrap-output-dir wrappedBlocks \
+  --address-book wrappedBlocks/addressBookHistory.json \
+  > live-sequential.log 2>&1 &
+```
+
+**Arguments**:
+- `-l`: Listing directory
+- `-o`: Downloaded days output directory
+- `--wrap-output-dir`: Wrapped blocks output directory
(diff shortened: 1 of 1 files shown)


docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,690 @@
...
+**Audience**: Operators running Block Stream validation, migration, or archival processes.
+
+---
+
+## Prerequisites
+
+### Required Tools
+
+```bash
+# Install Java 21+
```
ata-nas: Should we use java25?

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,690 @@
...
+│       └── ...
+├── compressedDays/                   # Downloaded record files (compressed)
+│   ├── 2024-01-01/
+│   └── ...
+└── wrappedBlocks/                    # Wrapped Block Stream output
+    ├── addressBookHistory.json       # Address book changes over time
+    ├── 0/                            # Block 0
+    ├── 1/                            # Block 1
+    └── ...
+```
```
ata-nas, line 85: Nit, not asking for a change, but these trees look great as mermaid.

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,690 @@
...
+## Core Workflows
+
+### 1. Generate Address Book History
+
+**Purpose**: Fetch address book changes from the Mirror Node API to enable block proof verification.
+
+**Command**:
+
+```bash
+nohup java -jar tools-<version>-all.jar \
```
ata-nas, line 120: Nit, we can abbreviate `java -jar tools-<version>-all.jar ` usages as an alias, as that will be much easier for users. Maybe something like `bntools` or similar? I think we can also generally use gradle to start these? But passing args was cumbersome.

Moreover, we could generally, if applicable, provide aliases for standard commands with the flags preconfigured.

ata-nas, line 120: Nit, do we want `nohup` always? It could be an optional thing?

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,690 @@
...
+
+```bash
+nohup java \
+  -Xms20g \
+  -Xmx20g \
+  -XX:+UseZGC \
+  -XX:+AlwaysPreTouch \
+  -XX:+UseNUMA \
+  -XX:ConcGCThreads=12 \
+  -jar tools-<version>-all.jar \
```
ata-nas, line 344: Nit, this whole thing (and others similar) could be aliased as well. I know there is value to see the jvm args, but if these are the common use case, from user perspective alias would be less stressful.

ata-nas (dismissed):
LG, quite extensive document.

I leave non-blocking cleanup suggestions in case any of these are useful.

AlfredoG87 (dismissed):
LGTM

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,716 @@
...
+**Audience**: Operators running Block Stream validation, migration, or archival processes.
+
+---
+
+## Prerequisites
+
+### Required Tools
+
+```bash
+# Install Java 25+ (or Java 21+ minimum)
```
jsync-swirlds: Java 25 is minimum.  Some of our library dependencies require Java 25 now.

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,716 @@
...
+
+The shadow jar is self-contained and can be copied to your deployment server.
+
+### Optional: Command Aliases
+
+For convenience, you can set up aliases to simplify command invocations:
+
+```bash
+# Add to ~/.bashrc or ~/.bash_profile
+alias bntools='java -jar /mnt/wrb-operations/tools-0.35.0-SNAPSHOT-all.jar'
```
jsync-swirlds: When creating an alias, it is recommended to always include a single space at the end; some shells detect that and better support using the alias as a command.

```suggestion
alias bntools='java -jar /mnt/wrb-operations/tools-0.35.0-SNAPSHOT-all.jar '
```

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,716 @@
...
+
+### GCP Authentication (for GCS access)
+
+Export GCP credentials before running any commands:
+
+```bash
+export GOOGLE_APPLICATION_CREDENTIALS="/path/to/gcp-key.json"
+export GOOGLE_CLOUD_PROJECT="your-project-id"
+export GCP_PROJECT_ID="your-project-id"
+```
```
jsync-swirlds, line 99: Just a question, can this be run with AWS?
We might want to (in future work) detail how to use AWS as well.

Spheres are in GCP, AWS, and Azure (Azure, of course, is its own unique brand of headaches due to not supporting most widely used APIs, including S3).

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,716 @@
...
+
+**Notes**:
+- Run this before starting wrapping or validation
+- Re-run periodically to capture new address book changes
+
+---
+
+### 2. Update Metadata Files
+
+Metadata files are essential for mapping blocks to days and timestamps.
```
jsync-swirlds, line 145: Do these address book and metadata updates rely on having a mirror node available?
If so, how do we refer to that mirror node?

The critical driver here is spheres, which will have their own private mirror nodes, not the public ones.

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,716 @@
...
+  --network testnet \
+  days updateDayListings \
+  --day 2026-04-03
+```
+
+---
+
+### 3. Download Block Data
+
+#### Bulk Historical Download (`download-days-v3`)
```
jsync-swirlds, line 197: Do we have instructions for how to specify the S3 data source other than specifying a "network"?

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,716 @@
...
+- Separate volumes for compressed days and wrapped blocks
+- Monitor disk usage during bulk downloads
+
+### Graceful Shutdown
+
+**SIGTERM (recommended)**:
+
+```bash
+kill <PID>
+```
```
jsync-swirlds: Why not use SIGINT (same as CTRL+C from keyboard)?
That's even more mild and simpler to manage (granted, a lot of Java code swallows interrupted exception, which makes this less useful than we'd like...)

````suggestion
**SIGINT (recommended)**:

```bash
kill -INT <PID>
```
````

File: docs/block-node/operations/wrb-cli-runbook.md [hunk 11]
```diff
@@ -0,0 +1,716 @@
...
+
+---
+
+## Appendix: Real-World Examples
+
+### Mainnet Production Setup
+
+**Hardware**:
+- 32-core Xeon server
```
jsync-swirlds: Perhaps don't specify Xeon?

```suggestion
- 32-core server
```

docs/block-node/operations/wrb-cli-runbook.md
```diff
@@ -0,0 +1,716 @@
...
+- 128GB RAM
+- 20TB ZFS RAID10 SSD array
```
jsync-swirlds, line 673: Are we running RAID10?
As far as I know we originally setup the filesystem as striped with no duplication...

jsync-swirlds (commented):
A few questions.

Also, do we have an issue (or issues) to add support for Spheres in the future? After reading this, I believe we'll need to make a few non-trivial updates to be able to support Spheres wrapping their records (which is required for a few of them _very shortly after_ mainnet).

jsync-swirlds (dismissed):

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: I will address this now.

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Yes, the address book generation requires mirror node API access. The `--network` flag determines the mirror node URL (mainnet uses public mirror, testnet uses testnet mirror). For private spheres with custom mirror nodes, use the `--mirror-node-url` flag to specify the sphere's mirror node endpoint.

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: I will double check.

jsync-swirlds (dismissed):

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: The tools currently only support GCS (Google Cloud Storage), not S3. The gcsBucketName field in the network config points to GCS buckets. Supporting S3 would require adding a new storage abstraction layer to handle both GCS and S3 APIs.

We'd have to make a decision if we want to support this. We probably do want to do this tbf.

docs/block-node/operations/wrb-cli-runbook.md
jsync-swirlds: We should probably explain how to do that in this document, but that can be in a future PR.

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Currently the tools are GCP only, using Google Cloud Storage libraries. However, since spheres are deployed across GCP, AWS, and Azure, we should add support for multiple cloud providers.                                                             
                                                                                                                                                                                                                                                           
  This would require:
  1. Abstracting the storage layer (currently hardcoded to GCS via com.google.cloud.storage)
  2. Adding S3 SDK for AWS support
  3. Potentially using S3-compatible APIs for Azure (or native Azure SDK)

  Given that this is already a large documentation PR, I'd suggest:
  1. Keeping this runbook focused on the current GCP implementation
  2.Filing a follow-up issue to track multi-cloud storage support
  3.Once implemented, we can add cloud-specific sections to this runbook

  Does this approach work? I will add issues for it to follow up if required etc.

rockysingh: The only change would be downloas files tbf nothing else so the uplift wouldn't be that bad.

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Updatded, well noted.

docs/block-node/operations/wrb-cli-runbook.md
jsync-swirlds: Sounds like a larger consideration; I thought we were using Bucky and S3 APIs.

Definitely worth filing an issue to work out what we need to do for broader support in the future; particularly as Spheres don't all use GCS (I'm not sure if any do, actually).

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Good idea for future enhancement\! Keeping as plain text for now since it's easier to maintain and works in all viewers, but we can consider mermaid diagrams in a follow-up if it improves readability.

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Added an 'Optional: Command Aliases' section after 'Building the Shadow JAR' that shows:
- Basic alias: `alias bntools='java -jar /mnt/wrb-operations/tools-0.35.0-SNAPSHOT-all.jar '`
- Alias with JVM args: `alias bntools-large='java -Xmx32g -Xms16g -jar ...'`
- Usage examples

The aliases include a trailing space (as jsync-swirlds suggested) to better support shell command expansion.

For pre-configured command aliases with common flags, I can add examples like `alias wrb-validate='bntools blocks validate --skip-signatures --skip-supply'` if that would be helpful.

rockysingh: - `nohup` option for long-running commands that continue after logout
- Alternative foreground execution without `nohup`

This makes it clear that `nohup` is optional and operators can choose based on their needs.

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Good point! The 'Optional: Command Aliases' section (added after Building the Shadow JAR) includes an example for this:

```bash
alias bntools-large='java -Xmx32g -Xms16g -jar /mnt/wrb-operations/tools-0.35.0-SNAPSHOT-all.jar'
```

This allows users to run heavy operations like wrapping without typing JVM args each time:
```bash
bntools-large --network mainnet blocks wrap --input-dir compressedDays
```

The alias approach strikes a balance - the runbook shows full commands for clarity/documentation, but operators can use aliases for convenience.

docs/block-node/operations/wrb-cli-runbook.md
jsync-swirlds: I agree with creating issues to make the future updates.

jsync-swirlds (approved):

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Currently the tools are GCP-only (using `com.google.cloud.storage` libraries). However, since spheres are deployed across GCP, AWS, and Azure, we should add multi-cloud support.

I've created issue #2847 to track implementing:
- AWS S3 support via AWS SDK
- Azure Blob Storage support
- Storage provider abstraction layer

For now, this runbook documents the current GCP implementation. Once multi-cloud support is added, we can expand the runbook with cloud-specific sections.

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Yes, address book generation and metadata updates do rely on mirror node API access. Currently the mirror node URL is determined by the `--network` flag (mainnet uses public mirror, testnet uses testnet mirror).

For private spheres with custom mirror nodes, I've created issue #2848 to track adding a `--mirror-node-url` flag that would allow operators to specify their sphere's mirror node endpoint:

```bash
java -jar tools.jar --network other --mirror-node-url https://sphere-mirror.example.com/api/v1/ \\
  mirror generateAddressBook
```

This would give spheres full control over which mirror node to query.

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Currently the tools only support GCS (not S3), so the `--network` flag determines the GCS bucket to use. Custom S3 sources aren't supported yet.

For custom network configurations, you can use `--network other` with a custom network config JSON file that specifies the `gcsBucketName` (which despite the name is GCS-specific).

However, since spheres need multi-cloud support, issue #2847 tracks adding AWS S3 and Azure storage backends. Once implemented, we'll add documentation for:
- Custom S3 bucket configuration
- Storage provider selection
- Cloud-specific authentication

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Great point! SIGINT is more standard and consistent with keyboard behavior. I've created issue #2849 to track updating the runbook to:
- Recommend SIGINT (`kill -INT <PID>`) as the primary graceful shutdown method
- Explain it's the same as CTRL+C
- Keep SIGTERM as a fallback option

This will make the documentation more intuitive for operators.

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: Agreed - generic is better. I've created issue #2850 to track removing vendor-specific hardware references and verifying the filesystem configuration (RAID10 vs striped ZFS).

docs/block-node/operations/wrb-cli-runbook.md
rockysingh: I've added an issue for this to be followed up:

https://github.com/hiero-ledger/hiero-block-node/issues/2850

Nana-EC (approved):
LG, approving noting a future PR will address comments

PR #2755, opened 2026-05-08
fix(publisher): prevent ACK leak during active block resends

### LINKED ISSUE #2748: bug(publisher): Ack Leak happened during bad block proof.
### Description

In the case of the BNCE case when due to a BadBlockProof we had to send a RESEND to the publisher, there was a leak in which the BN replied ACK for a later block than the block that failed verification, hence producing Block Gaps in the BN, something that never has to happen in production

### Steps to Reproduce

Part of the Intent of this issue is to figure out the steps to reproduce and create a regression test to validate the fix.

### Additional context

`LiveStreamPublisherManager.activeResendBlocks` is queried by gap detection but **never `.add()`'d anywhere**. So between "publisher accepted RESEND header" (which removes the block from `blocksToResend`) and "verification completes", neither set contains the block. If a later block's persistence notification fires in that window, gap detection sees no gap and ACKs the later block — implicitly claiming the still-unverified block is durable.

Populate `activeResendBlocks` at the resend-accept site, and remove on failed-verification:

The existing `endOfBlock` and `blockIsEnding` removals stay as-is. Order on the failure path: re-add to `blocksToResend` first, then clear `activeResendBlocks`, so there's no sub-µs hole where neither set contains the block.

## Regression test

In `LiveStreamPublisherManagerTest`:

1. Seed `blocksToResend = {N+1}`, drive a header for N+1 through `getActionForBlock` → `ACCEPT`.
2. Without delivering items, fire `handlePersisted(N+3, success=true)`.
3. Assert: `lastPersistedBlockNumber` did not advance, `sendAcknowledgement` was not called.

Second test: after step 1, fire `handleVerification(N+1, success=false)` then `handlePersisted(N+3)`. Assert: still no advance.

AlfredoG87:
## Summary

`LiveStreamPublisherManager.activeResendBlocks` is queried by gap detection (`correctForResendAndStreaming`) and stall detection (`isResendingLive`) but is never `.add()`'d anywhere. Once a publisher's RESEND header is ACCEPTed and the block is removed from `blocksToResend`, no set tracks the in-flight resend. A `handlePersisted` for any later block in that window misses the gap and fans out a spurious ACK, telling publishers an unverified block is durable.

This PR populates `activeResendBlocks` at the resend ACCEPT site so gap detection sees the in-flight block.

## Changes

- `LiveStreamPublisherManager.getActionForHeader` — when a block transitions from `blocksToResend` to in-flight via the resend ACCEPT path, also `activeResendBlocks.add(blockNumber)`. Existing removals in `endOfBlock` and `blockIsEnding` handle cleanup on completion.
- `LiveStreamPublisherManagerTest` — two new tests under `handlePersisted() Tests`:
  - `testNoAckLeakWhileSingleResendHeaderAccepted` — single resend in flight when a later block persists.
  - `testNoAckLeakWhileTwoConcurrentResendsHeaderAccepted` — two concurrent resends accepted before either completes verification.

  Both fail on `main`; both pass with this change.
  
  Fixes #2748

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -768,7 +768,11 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         if (blocksToResend.contains(blockNumber)) {
             // If we expect a block to be resent, we must check if this publisher is currently publishing that block
             if (blocksToResend.remove(blockNumber)) {
-                // Only one publisher can enter here and start publishing the expected block to be resent
+                // Only one publisher can enter here and start publishing the expected block to be resent.
+                // Track the block as actively resending so gap detection in correctForResendAndStreaming
+                // and stall detection in isResendingLive can see it. The set is cleared by endOfBlock /
+                // blockIsEnding on completion, and by handleVerification on failed re-verification.
+                activeResendBlocks.add(blockNumber);
                 return BlockAction.ACCEPT;
             } else {
                 return BlockAction.SKIP;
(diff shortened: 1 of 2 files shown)


Nana-EC (approved):

berryware (approved):
Looks Good

jsync-swirlds (approved):
No idea how the original call here got lost, thanks for fixing.

PR #2758, opened 2026-05-08
feat(verification): Temporary bypass for 1-path StateBlock Proofs in order to troubleshoot and potentially unblock.

AlfredoG87:
In the 2026-05-07 incident the CN emitted state proofs with 1 path while the BN expects 3, stalling the live stream. As a temporary debug workaround, `ExtendedMerkleTreeSession.verifyStateProof` now treats a 1-path state proof as verified (instead of failing) and logs a WARN with the block number. The block is persisted via the normal storage path and can be retrieved from disk for offline analysis.

Independent improvement: the verification-failure log is upgraded from INFO to WARN and now includes `failureType`.

## TODO before production

This is **not a fix**. Once the CN/BN proof-shape mismatch is resolved, delete the bypass branch in `ExtendedMerkleTreeSession.verifyStateProof`. The site is marked `TODO: TEMPORARY DEBUG WORKAROUND`.


AlfredoG87:
Not needed anymore. issue came up to be a CN configuration mismatch.enableStateProofs=true

missing

PR #2764, opened 2026-05-08
fix: update runner size for solo E2E test deployment

### LINKED ISSUE #2790: Bug(solo-e2e-test): Runs are Failing due to TCK-SDK install issues with latest version
### Description

We need to pin the version of the SDK to a stable one and allow to override using an specific version, same as solo version. input should accept `latest`

As part of this fix, also fix the large topologies runs by using the newly provided runners.

### Steps to Reproduce

Hit the daily schedule on Actions

Image

### Additional context

_No response_

AlfredoG87:
## Reviewer Notes

This PR upgrades the Solo-e2e-test runner to use a Large Box so it can successfully run larger topologies and in the future we can also use it for longer testing with more load.

Also fixes an issue with the TCK-SDK installation that is happening on daily and weekly runs.

Finally it makes the TCK-SDK Version parametrized and pinned by default to "v2.83.0" and pnpm to version 9.

Main:
https://github.com/hiero-ledger/hiero-block-node/actions/runs/25541303143/job/74967658444

This branch:
https://github.com/hiero-ledger/hiero-block-node/actions/runs/25685466309

Fixes #2790

### DIFF
--- a/.github/workflows/solo-e2e-scheduler.yml
+++ b/.github/workflows/solo-e2e-scheduler.yml
@@ -52,9 +52,9 @@ on:
         type: string
         default: ""
       js-sdk-version:
-        description: "JS-SDK version (leave empty to use workflow default; 'latest' to use newest tag)"
+        description: "JS-SDK version ('latest' or tag like 'v2.83.0')"
         type: string
-        default: ""
+        default: "v2.83.0"
       solo-version:
         description: "Solo CLI version (leave empty to use run-type default)"
         type: string
--- a/.github/workflows/solo-e2e-scheduler.yml
+++ b/.github/workflows/solo-e2e-scheduler.yml
@@ -229,9 +229,8 @@ jobs:
           TCK_VERSION="${{ inputs.tck-version }}"
           TCK_VERSION="${TCK_VERSION:-latest}"
 
-          # JS-SDK version: use input if provided, otherwise default to last-known-good pinned version
+          # JS-SDK version: use input value (input has default v2.83.0)
           JS_SDK_VERSION="${{ inputs.js-sdk-version }}"
-          JS_SDK_VERSION="${JS_SDK_VERSION:-v2.83.0}"
 
           # TSS: use input value (defaults to true)
           TSS_ENABLED="${{ inputs.tss-enabled }}"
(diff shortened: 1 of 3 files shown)


jsync-swirlds (dismissed):

Nana-EC (dismissed):

.github/workflows/solo-e2e-scheduler.yml
```diff
@@ -51,6 +51,10 @@ on:
         description: "TCK-SDK version (leave empty for 'latest')"
         type: string
         default: ""
+      js-sdk-version:
+        description: "JS-SDK version (leave empty to use workflow default; 'latest' to use newest tag)"
+        type: string
+        default: ""
```
andrewb1269: Can we make the default `latest` or another specific version?

.github/workflows/solo-e2e-scheduler.yml
```diff
@@ -224,6 +229,10 @@ jobs:
           TCK_VERSION="${{ inputs.tck-version }}"
           TCK_VERSION="${TCK_VERSION:-latest}"
 
+          # JS-SDK version: use input if provided, otherwise default to last-known-good pinned version
+          JS_SDK_VERSION="${{ inputs.js-sdk-version }}"
+          JS_SDK_VERSION="${JS_SDK_VERSION:-v2.83.0}"
```
andrewb1269: Use this as a default version in the inputs.

andrewb1269 (changes requested):

.github/workflows/solo-e2e-scheduler.yml
AlfredoG87: ✅ Done!

.github/workflows/solo-e2e-scheduler.yml
AlfredoG87: ✅ Done!

andrewb1269:
```suggestion
```

andrewb1269 (changes requested):

.github/workflows/solo-e2e-scheduler.yml
andrewb1269: I don't think you need this line since the default version will be set on input.

AlfredoG87: makes sense, now is no longer needed, is redundant 👍 
✅ Done!

andrewb1269 (approved):

jsync-swirlds (approved):

PR #2788, opened 2026-05-12
test: enable WRB integration tests with network fixtures and various improvements

### LINKED ISSUE #2668: improvement(verification): Follow-up improvements for StateBlockProof.
### Story Form

Original PR is functional al ready to be merge. however there are  some things that can be improved:

https://github.com/hiero-ledger/hiero-block-node/pull/2535#discussion_r3161358433

https://github.com/hiero-ledger/hiero-block-node/pull/2535#discussion_r3080271476

https://github.com/hiero-ledger/hiero-block-node/pull/2535#pullrequestreview-4207518471

### Technical Notes

Also there is a valid concern that Ivan found with Gemini regarding using BlockProof that warrants take a closer look.

### LINKED ISSUE #2773: SignedRecordFileProof verification Improvements
## Story Form

As a Block Node maintainer,
I want to address the open review comments from PR #2720 (SignedRecordFileProof RSA verification),
So that the verification implementation is clean, well-documented, and production-ready.

## Technical Notes

The following comments were left on [PR #2720](https://github.com/hiero-ledger/hiero-block-node/pull/2720) and could not be addressed before merge. Each item below links directly to the original comment.

**CodeQL / Security**
- [x] [RsaWrbVerificationTest.java:107 — CodeQL: RSA test key size below 2048 bits](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3205948355)
- [x] [RsaKeyDecoderTest.java:41 — CodeQL: RSA test key size below 2048 bits](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3206310292)
- [x] [RsaKeyDecoderTest.java:60 — CodeQL: RSA test key size below 2048 bits](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3206310298)

**Missing assertions in tests**
- [x] [VerificationServicePluginTest.java:454 — Pending assertions](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211370113)
- [x] [VerificationServicePluginTest.java:477 — Pending assertions](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211372087)
- [x] [VerificationServicePluginTest.java:498 — Pending assertions](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211372968)

**Code style — fully qualified class names should be replaced with imports**
- [x] [VerificationServicePluginTest.java:421 — use import instead of fully qualified name](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211376524)
- [x] [VerificationServicePluginTest.java:424 — import java.util.HexFormat instead of full name](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211381175)
- [x] [AllBlocksHasherHandlerTest.java:322 — use import instead of fully qualified name](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211565855)

**Design / correctness questions**
- [x] [ExtendedMerkleTreeSession.java:261 — Do we expect more than one SignedRecordFileProof per block? Clarify whether the loop is correct or should be a single-item check](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211505830)
- [x] [ExtendedMerkleTreeSession.java:424 — Unused variable; remove or use it](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211515942)
- [x] [ExtendedMerkleTreeSession.java:224 — Is the signed payload computation (SHA-384 over int32(6) || rawBytes) specified in a HIP or design doc? Add a reference](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211563096)
- [ ] [VerificationServicePlugin.java:108 — Should the keyByNodeId map also store node weight/stake for weighted-majority checks?](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211593897)

**Documentation / observability**
- [x] [VerificationServicePlugin.java:71 — Add RSA verification metrics (blocknode_rsa_verification_total, blocknode_rsa_verification_latency_seconds, blocknode_roster_mismatch_total) to the metrics documentation and default Grafana dashboards](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211591410)

**Log level — WARN vs ERROR on RSA verification failure**
- [x] [ExtendedMerkleTreeSession.java:438 — Consider WARN instead of ERROR for RSA verification failure](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211632521)
- [x] [ExtendedMerkleTreeSession.java:450 — Consider WARN instead of ERROR for RSA verification failure](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211632817)
- [x] [ExtendedMerkleTreeSession.java:474 — Consider WARN instead of ERROR for RSA verification failure](https://github.com/hiero-ledger/hiero-block-node/pull/2720#discussion_r3211633926)

AlfredoG87:
**WRB integration tests**

- real solo-network fixtures (5 blocks + roster) replace the placeholder
- `VerificationServicePluginIntegrationTest` un-disabled and parametrised over every fixture
- negative case: empty address book
- negative case: address book carrying the wrong RSA keys (asserts `BAD_BLOCK_PROOF`)

**Verification improvements** (from PR #2720 review threads)

- CodeQL key-size bumps (1024 → 2048) in every RSA test keypair
- missing `onContextUpdate_*` assertions filled
- fully qualified class names replaced with imports across the module
- RSA proof selection: `getFirst` → `getSingle`
- `computeFinalBlockHash` deferred to the success branch in `verifyRsaProof`
- RSA failure logs demoted ERROR → WARNING

**State-proof observability**

- new `state_proof_verification_success_total` / `_failure_total` counters
- DEBUG / WARNING log pair on accept / reject
- new counters documented in `docs/block-node/metrics.md`
- Grafana verification dashboard: new RSA WRB row, new State-Proof row, Avg Hashing Time pair, panel-21 unit fix (`ms` → `ns`)

## Test plan

- [x] `./gradlew :verification:test :app:test` — 101 verification + 48 app, all passing
- [x] `./gradlew :verification:qualityCheck`
- [ ] manual sanity check of the rebuilt dashboard in Grafana

## Related Issue(s)

Fixes #2773
Fixes #2668


block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -258,13 +269,16 @@ public VerificationNotification finalizeVerification(Bytes rootHashOfAllBlockHas
         Bytes startOfBlockStateRootHash = this.blockFooter.startOfBlockStateRootHash();
 
         // Route by proof type. RSA WRB proofs take priority (Phase 2a); TSS proofs follow.
-        final BlockProof rsaProof = getFirst(blockProofs, BlockProof::hasSignedRecordFileProof);
-        if (rsaProof != null) {
-            return verifyRsaProof(
-                    rsaProof, previousBlockHashToUse, rootOfAllPreviousBlockHashes, startOfBlockStateRootHash);
-        }
-
+        // A block carries exactly one BlockProof item, so `getSingle` is correct for every type;
```
Nana-EC: I don't think this is true long term.
Post TSS cutover we will need to support a block with 2 SignRecordFileProof and TssSignedBlock Proof.

Why did we need to change from getFirst to getSingle?

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -723,6 +735,14 @@ protected VerificationNotification getStateProofVerificationResult(
 
         final boolean verified = verifyStateProof(blockRootHash, blockProof.blockStateProof());
 
+        if (verified) {
+            LOGGER.log(DEBUG, "State-proof verification accepted block {0}", blockNumber);
+            if (stateProofVerificationSuccessTotal != null) stateProofVerificationSuccessTotal.increment();
```
Nana-EC: Q: Are we able to start using labels?
Should we rather have a single success and failure metrics with the label being block proof type?

block-node/verification/src/test/java/org/hiero/block/node/verification/fixtures/WrbAddressBookFixtureGeneratorTest.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.hiero.block.node.app.fixtures.TestUtils;
+import org.hiero.block.node.verification.RsaKeyDecoder;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+/**
```
Nana-EC: nit: for new file might as well go with the markdown styled comments so reduce the future commenting revamp size

Nana-EC (commented):
Nice additions, did a quick pass. Will circle back again tomorrow

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: There is no harm if a block has multiple proof of different type, what this is trying to prevent is multiple proofs of the same type.

what makes me wonder, why would we expect 2 proofs of `SignedRecordFileProof` per block?

If we were to make the change from getSingle to getFirst, should we do it for all proof types or only `SignedRecordFileProof` ?

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: Yes!! I almost forgot we already have that capability, let me refactor the whole thing so we can use a single metric for all verifications results, we can have a single metric with multiple labels such as "result"-> true,false.
ProofType --> SignedRecordFileProof, BlockStateProof, SignedBlockProof

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
Nana-EC: Ah that distinction is important. 
Yeah, 1 of each type should be fine.
So in this case the getSingle would fail if we observed multiple of the same type?

Nana-EC: I'm okay with the above but let's confirm with CN and architects that we don't ever expect to se multiple of the same type and that it's safe to throw/block if so.
for example if an amendment was needed how would it be provided in this case

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅ Done!

image

block-node/verification/src/test/java/org/hiero/block/node/verification/fixtures/WrbAddressBookFixtureGeneratorTest.java
AlfredoG87: ✅ Done!

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: Changes according to offline disccusion have been made ✅

Nana-EC (dismissed):
No blocking comments

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -665,9 +659,20 @@ private static Bytes extractRecordStreamFileBytes(final Bytes recordFileItemByte
     /**
      * Computes the V6 RSA signed payload: `SHA-384(int32(6) || rawRecordStreamFileBytes)`.
      *
-     * <p>This matches `SignatureDataExtractor.computeSignedHash(6, ...)` in
-     * `tools-and-tests/tools`. Inlined here because that module cannot be imported from
-     * `block-node/verification`.
+     * <p>This is the Hedera Record Stream V6 signing convention: a big-endian 32-bit version
```
Nana-EC:
```suggestion
     * <p>This is the Hiero Record Stream V6 signing convention: a big-endian 32-bit version
```

block-node/verification/src/test/java/org/hiero/block/node/verification/fixtures/WrbAddressBookFixtureGeneratorTest.java
```diff
@@ -0,0 +1,183 @@
...
+        return is;
+    }
+
+    private static String readResourceUtf8(final String path) throws Exception {
+        try (InputStream is = openResource(path)) {
+            return new String(is.readAllBytes(), StandardCharsets.UTF_8);
+        }
+    }
+
+    private static int findMatchingBrace(final String s, final int openBraceIndex) {
```
Nana-EC, line 151: Hmm, wondering why we need this and had to code it ourselves.
might be worth seeing is there's an improvement here to more reliable parse.
Is this broken by any proto updates?

Nana-EC (commented):
2 nits I saw. Approval stands

block-node/verification/src/test/java/org/hiero/block/node/verification/fixtures/WrbAddressBookFixtureGeneratorTest.java
AlfredoG87: I did not thought it was necessary to create a new `.proto` file for parsing the `genesis-network.json` file.

That said, I can see that this code is brittle and we should not do it either. This is only used for Testing in order to convert the `genesis-network.json` file into a `address-book.json` originally I did not consider to check-in the source file and the helper to create the file we actually needed for the fixtures `address-book.json` however later I thought it might add some value, so we can replicate the fixtures by providing the source json file that we get from the CN.

2 options to improve here:

1. Remove the source file and the helper method.
2. Keep the source file and make a minimal proto that allow us to Parse the necessary fields only.

In any case we can revisit this on a follow-up.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: ✅ Improved.

Nana-EC (approved):

block-node/verification/src/main/java/org/hiero/block/node/verification/session/VerificationProofMetrics.java
```diff
@@ -0,0 +1,86 @@
...
+ * @param stateProofFailure  {@code verification_proof_total{proof_type="state_proof",result="failure"}}
+ * @param tssSuccess         {@code verification_proof_total{proof_type="tss",result="success"}}
+ * @param tssFailure         {@code verification_proof_total{proof_type="tss",result="failure"}}
+ * @param rsaRosterMismatch  {@code rsa_roster_mismatch_total} — RSA signatures from node IDs
+ *                           absent from the loaded address book
+ */
+public record VerificationProofMetrics(
+        @Nullable LongCounter.Measurement rsaSuccess,
+        @Nullable LongCounter.Measurement rsaFailure,
+        @Nullable LongCounter.Measurement stateProofSuccess,
```
berryware, line 28: Will we ever have more proof types? Seems like there might be a better Object Oriented way to implement this

berryware (approved):
Looks good

PR #2789, opened 2026-05-12
feat: AddressBook Roster Load/Persist Improvements

### LINKED ISSUE #2756: AddressBook Roster Load/Persist Improvements
## Context

Follow-up work from PR #2718 (_feat: add NodeAddressBook state loading/persistence to block node_). The following reviewer comments were not addressed before merge and require a dedicated follow-up.

## Outstanding items

- [x] **[File path should reference JSON, not binary](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3204988762)** — `block-node/app/build.gradle.kts`: resource path referenced in config may need to point to the `.json` file rather than the current value.

- [x] **[Add startup check that parent directory is writable](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3204994417)** — `ApplicationStateConfig`: validate that the parent directory of the persistence file path is writable at startup, so permission issues surface early rather than at first persist attempt.

- [x] **[`updateAddressBook` should signal success/failure to callers](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205007062)** — `BlockNodeApp`: method currently swallows errors silently. Return a `boolean` (or a result object) so plugins know whether the update succeeded.

- [x] **[Document thread-safety contract of `updateAddressBook`](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205010718)** — `BlockNodeApp`: add Javadoc stating the thread-safety guarantees (or lack thereof) for this method.

- [x] **[Fail-fast on malformed RSA key format at address-book update time](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205015626)** — `BlockNodeApp`: validate that each `rsaPubKey` value is parseable DER when the address book is ingested, rather than waiting for proof-verification time.

- [x] **[Avoid calling `updateBlockNodeContext` twice with partial state](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205024477)** — `BlockNodeApp`: the current logic triggers a context rebuild once per updated field, which is confusing and potentially races. Batch state updates and trigger a single `updateBlockNodeContext` call.

- [x] **[Consider coalescing context updates with a short debounce window](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205430786)** — `BlockNodeApp` (jsync suggestion): queue updates, let them accumulate for ~100 ms, then fire a single context rebuild. Keep `BlockNodeContext` as a record to avoid needing multiple global locks.

## References

- Source PR: https://github.com/hiero-ledger/hiero-block-node/pull/2718

### LINKED ISSUE #2757: Run checkForApplicationStateUpdates before startup
### Story Form

As a block node operator
I want to know that all state data has been loaded into context prior to plugin startup
So that I can be assured all plugins have latest state for consideration on startup

### Technical Notes

Currently there's a [scheduled run of checkForApplicationStateUpdates](https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java#L439) which will trigger a blockNodeContext update for any loaded state.
We should not expose a `appStateConfig.updateInitialDelay` value but rather have it hardcoded to 0 so that it's immediate and done before plugin `start()`.

berryware:
## Reviewer Notes
- [x] **[File path should reference JSON, not binary](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3204988762)** — `block-node/app/build.gradle.kts`: resource path referenced in config may need to point to the `.json` file rather than the current value.

- [x] **[Add startup check that parent directory is writable](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3204994417)** — `ApplicationStateConfig`: validate that the parent directory of the persistence file path is writable at startup, so permission issues surface early rather than at first persist attempt.

- [x] **[`updateAddressBook` should signal success/failure to callers](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205007062)** — `BlockNodeApp`: method currently swallows errors silently. Return a `boolean` (or a result object) so plugins know whether the update succeeded.

- [x] **[Document thread-safety contract of `updateAddressBook`](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205010718)** — `BlockNodeApp`: add Javadoc stating the thread-safety guarantees (or lack thereof) for this method.

- [x] **[Fail-fast on malformed RSA key format at address-book update time](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205015626)** — `BlockNodeApp`: validate that each `rsaPubKey` value is parseable DER when the address book is ingested, rather than waiting for proof-verification time.

- [x] **[Avoid calling `updateBlockNodeContext` twice with partial state](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205024477)** — `BlockNodeApp`: the current logic triggers a context rebuild once per updated field, which is confusing and potentially races. Batch state updates and trigger a single `updateBlockNodeContext` call.

- [x] **[Consider coalescing context updates with a short debounce window](https://github.com/hiero-ledger/hiero-block-node/pull/2718#discussion_r3205430786)** — `BlockNodeApp` (jsync suggestion): queue updates, let them accumulate for ~100 ms, then fire a single context rebuild. Keep `BlockNodeContext` as a record to avoid needing multiple global locks.

## Related Issue(s)
 Closes: #2756
 Closes: #2757

### DIFF
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -452,10 +452,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
 
         // Schedule periodic check for live updates from running plugins.
         applicationStateExecutor.scheduleAtFixedRate(
-                this::checkForApplicationStateUpdates,
-                appStateConfig.updateInitialDelay(),
-                appStateConfig.updateScanInterval(),
-                TimeUnit.MILLISECONDS);
+                this::checkForApplicationStateUpdates, 0, appStateConfig.updateScanInterval(), TimeUnit.MILLISECONDS);
     }
 
     private void checkForApplicationStateUpdates() {
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -31,7 +31,6 @@ import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.security.KeyFactory;
 import java.security.NoSuchAlgorithmException;
-import java.security.PublicKey;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.X509EncodedKeySpec;
 import java.time.Duration;
@@ -689,7 +688,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
             }
             try {
                 final byte[] keyBytes = hex.parseHex(addr.rsaPubKey());
-                final PublicKey key = kf.generatePublic(new X509EncodedKeySpec(keyBytes));
+                kf.generatePublic(new X509EncodedKeySpec(keyBytes));
                 usable++;
             } catch (InvalidKeySpecException | IllegalArgumentException e) {
                 LOGGER.log(WARNING, "Malformed RSA_PubKey for node {0} — skipped: {1}", addr.nodeId(), e.getMessage());
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -49,6 +49,7 @@ import org.hiero.block.node.app.logging.CleanColorfulFormatter;
 import org.hiero.block.node.app.logging.ConfigLogger;
 import org.hiero.block.node.spi.ApplicationStateFacility;
 import org.hiero.block.node.spi.BlockNodeContext;
+import org.hiero.block.node.spi.BlockNodeContext.Builder;
 import org.hiero.block.node.spi.BlockNodePlugin;
 import org.hiero.block.node.spi.ServiceLoaderFunction;
 import org.hiero.block.node.spi.blockmessaging.BlockMessagingFacility;
@@ -85,7 +86,11 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     private final HistoricalBlockFacilityImpl historicalBlockFacility;
     /** Should the shutdown() method exit the JVM. */
     private final boolean shouldExitJvmOnShutdown;
-    /** The block node context. Volatile: written by the scheduled scanner thread, read by plugin threads. */
+
+    /** The block node context. It is marked as volatile for thread safety.
+     * It is written by the scheduled scanner thread, read by plugin threads.
+     * Plugins should take care to make a copy of the BlockNodeContext before they use it so that they get a consistent
+     * BlockNodeContext */
     volatile BlockNodeContext blockNodeContext;
     /** list of all loaded plugins. Package so accessible for testing. */
     final List<BlockNodePlugin> loadedPlugins = new ArrayList<>();
@@ -376,7 +381,9 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     }
 
     /**
-     * Allow plugins to update the TssData for this BlockNodeApp
+     * Allow plugins to update the TssData for this BlockNodeApp. Uses a concurrentList to capture all TssData
+     * updates between scans. The ApplicationStateFacility scans for updates on a separate thread and will process any
+     * TssData updates that are newer than the current TssData.
      *
      * @param tssData - The TssData to be updated on the `BlockNodeContext`
      */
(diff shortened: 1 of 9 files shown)


block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -624,6 +639,14 @@ private void loadApplicationState(final Configuration configuration) {
                                 + " — delete and restart to re-fetch from Mirror Node",
                         e);
             }
+        } else {
+            // make sure the directory is created for the writes
+            Path parent = rsaFilePath.getParent();
+            try {
+                Files.createDirectories(tssDataJsonPath.getParent());
```
AlfredoG87: should it be:

  Files.createDirectories(rsaFilePath.getParent());

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -437,40 +452,27 @@ void startApplicationStateFacility() {
...
-            // Because we only update TssData for the most recent blockNumber,
-            // |= will let us know if any TssData were updated.
-            tssUpdated |= updateBlockNodeContext(tssData, null);
-            tssData = tssDataUpdates.poll();
+
+        // get any TssData update
+        TssData tssData = getPendingTssData();
+        if (tssData != null) {
+            persistTssData(tssData);
+            LOGGER.log(INFO, "ApplicationStateFacility persisted TssData");
```
AlfredoG87: should this message only be present if actual persistence is made?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -437,40 +452,27 @@ void startApplicationStateFacility() {
...
         }
 
-        // Consume any address book cached at startup (one-shot).
-        boolean addressBookUpdated = false;
         final NodeAddressBook addressBook = pendingAddressBook.getAndSet(null);
         if (addressBook != null) {
-            addressBookUpdated = updateBlockNodeContext(null, addressBook);
+            persistNodeAddressBook(addressBook);
```
AlfredoG87, line 468: similar as above, the log might be misleading.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -389,20 +402,22 @@ public void updateTssData(TssData tssData) {
...
+     * Null, empty, or all-blank-key books will ```return false;```
      *
      * @param nodeAddressBook the NodeAddressBook to store in BlockNodeContext
+     * @return true if the address book is queued for update, false if it was not
      */
     @Override
-    public void updateAddressBook(NodeAddressBook nodeAddressBook) {
-        if (nodeAddressBook == null) return;
+    public boolean updateAddressBook(NodeAddressBook nodeAddressBook) {
+        if (nodeAddressBook == null) return false;
```
AlfredoG87: should we skip if the old and new are the same, no changes.?

```
if (nodeAddressBook.equals(current)) {
          return false; // no-op update, nothing to do
      }   

```

AlfredoG87 (dismissed):
Only 1 potential blocking review. the others are nits
might need to do another pass later.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -601,6 +608,14 @@ private void loadApplicationState(final Configuration configuration) {
             } catch (ParseException | IOException e) {
                 LOGGER.log(ERROR, "Failed to read Application State Data file: " + tssDataJsonPath, e);
             }
+        } else {
+            // make sure the directory is created for the writes
+            Path parent = tssDataJsonPath.getParent();
+            try {
+                Files.createDirectories(tssDataJsonPath.getParent());
+            } catch (IOException e) {
+                LOGGER.log(ERROR, "Failed to create Application State Data directory: " + parent, e);
```
Nana-EC: nit: we should start saying TSS roster vs application state. At this point application state contains multiple items - tss roster, rsa roster, block ranges

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -624,6 +639,14 @@ private void loadApplicationState(final Configuration configuration) {
                                 + " — delete and restart to re-fetch from Mirror Node",
                         e);
             }
+        } else {
+            // make sure the directory is created for the writes
+            Path parent = rsaFilePath.getParent();
+            try {
+                Files.createDirectories(tssDataJsonPath.getParent());
+            } catch (IOException e) {
+                LOGGER.log(ERROR, "Failed to create Application State Data directory: " + parent, e);
```
Nana-EC: Same comment on log

Nana-EC (dismissed):
Thanks for the improvements.
Assuming resolution of other comments I only had non blocking nits

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Added

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Removed log message as it is already present in the persist method

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Removed log message as it is already present in the persist method

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Changed

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Good Catch. Fixed

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Changed.

AlfredoG87 (approved):
LG! 💯

Nana-EC (approved):

PR #2794, opened 2026-05-12
fix(solo-e2e): Fix the JS_SDK_VERSION Default on Scheduled runs

AlfredoG87:
## Reviewer Notes

fix: set default JS_SDK_VERSION to v2.83.0 in E2E test configurations

#2790 

Test run passed https://github.com/hiero-ledger/hiero-block-node/actions/runs/25760769815/job/75661195891 ✅

### DIFF
--- a/.github/workflows/solo-e2e-scheduler.yml
+++ b/.github/workflows/solo-e2e-scheduler.yml
@@ -230,7 +230,7 @@ jobs:
           TCK_VERSION="${TCK_VERSION:-latest}"
 
           # JS-SDK version: use input value (input has default v2.83.0)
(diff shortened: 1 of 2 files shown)


jsync-swirlds (dismissed):
Didn't we fix this in another PR recently?

jsync-swirlds (approved):

AlfredoG87:
> Didn't we fix this in another PR recently?

Yes, but when the scheduler was triggered by time, I found out that the `default` field on the input does not apply there. hence the sdk version was being blank and failing, this was apparent only once the scheduled job was triggered, no way to test it out manually.

rbarker-dev (approved):

PR #2798, opened 2026-05-12
test: Add solo-chaos latency considerations to e2e test framework

### LINKED ISSUE #2125: Chaos/Slow Traffic Test
# Chaos/Slow Traffic Test
As a block node engineer
I want to test BN behavior under degraded network conditions
So that I can validate resilience to real-world network issues

## Description

Create chaos engineering tests that simulate slow or degraded network traffic between components (latency, packet loss, bandwidth throttling).

## Goals

- BN degrades gracefully under network latency (slower but correct)
- BN recovers to normal throughput when network conditions improve
- No data corruption or out-of-order blocks under degraded conditions
- MN continues receiving blocks throughout

## Scenarios to Cover

| Scenario | Description |
|----------|-------------|
| Latency injection | Add 100-500ms delay to CN→BN traffic |
| Packet loss | Introduce 5-10% packet loss |
| Bandwidth throttling | Limit bandwidth between components |

## Implementation Notes

Extends existing test framework:
- New event types needed: `inject-latency`, `clear-faults` (or similar)
- Implementation options: `tc qdisc` via kubectl exec, Chaos Mesh, network policies
- Use existing assertions (`no-errors`, `blocks-increasing`, `node-healthy`)
- Test definition follows existing YAML structure

## Acceptance Criteria

- [ ] Test injects network latency/degradation
- [ ] BN continues processing under degraded conditions
- [ ] Recovery verified when faults cleared
- [ ] No verification errors or data corruption

Nana-EC:
## Summary

Adds the ability to inject **real network latency** between Consensus Nodes and Block Nodes in the solo e2e
kind cluster, and to assert that the network **degrades gracefully and recovers**. It layers a Chaos Mesh
`NetworkChaos` foundation onto the existing `solo-test-runner`, adds latency-aware assertions, ships
baseline / stress / severe latency profiles across all three CN/BN dimensions, and wires a conditional
Chaos-Mesh install into the solo-e2e workflow.

All changes are under `tools-and-tests/scripts/solo-e2e-test/` plus one conditional step in
`.github/workflows/solo-e2e-test.yml`. **No product code, chart, or proto changes.**

## Why

Block-node and consensus-node behaviour under realistic cross-region network
conditions wasn't testable locally. Production runs cross-region with 50–200 ms RTTs;
local testing ran at zero in-cluster latency. This lands the foundation to catch
backfill-timeout, publisher-reconnect, and consensus-round issues that only surface
when the network actually feels distant.

## The latency is real — how it's injected (not a mock)

Each `inject-latency` event renders a Chaos Mesh **`NetworkChaos`** resource with `action: netem` and applies
it to the cluster. Chaos Mesh programs Linux **`tc`/`netem`** on the *actual pod network namespaces* selected
by label, so packets between the targeted pods carry the configured delay ± jitter (and optional correlation
/ loss) at the kernel level — this is genuine egress shaping, not an application-level stub.

Two guardrails ensure a run can't silently "pass" without latency actually being applied:

- **Selector dry-run:** before applying, `chaos-dryrun.sh` resolves the source/target label selectors and
  **fails the test if either matches zero pods** — so a mislabeled selector can't produce a no-op rule.
- **Deterministic apply + cleanup:** each rule gets a stable name and is recorded; an `EXIT` trap tears down
  any active `NetworkChaos` even if the test crashes mid-window, so latency never leaks across runs.

Latency is injected on three independent paths, each stressing a different part of the system:

| Dimension | Selectors (source → target) | What it stresses |
|-----------|------------------------------|------------------|
| **CN ↔ CN** | `network-node` → `network-node` | Gossip / consensus round-trips (slower rounds) |
| **CN ↔ BN** | `network-node` → `block-node` | The live block-publish gRPC stream (delayed ACKs, publisher flow-control) |
| **BN ↔ BN** | `block-node` → `block-node` | The peer backfill mesh (slower catch-up between BNs) |

---

## Profiles — increasing stress on the system

Same event shape (`load-start → inject → clear → load-stop`); the tiers differ in **magnitude** and in what
the assertions expect. Select a tier by its test name (all require `paired-3` = 3 CN + 3 BN, TSS on).

| Profile | Test(s) | Latency (CN↔CN / CN↔BN / BN↔BN) | What it proves |
|---------|---------|----------------------------------|----------------|
| **baseline** | `latency-cn-to-cn`, `-cn-to-bn`, `-bn-to-bn`, `-all-three` | 100±20 / 150±30 / 200±40 ms | System tolerates normal cross-AZ/region latency with **no visible impact** (steady block rate). |
| **stress** | `latency-stress` | 300±80 / 450±120 / 600±150 ms, corr 50% | System **degrades then recovers**: BNs fall behind and catch up via **backfill**. |
| **severe** | `latency-severe` | 600±150 / 800±200 / 1000±250 ms, corr 75% | **Survival / recovery near the breaking point** — recovers cleanly once latency clears. |

---

## The assertion logic — what each check actually proves

| Assertion | How it works | What it proves |
|-----------|--------------|----------------|
| `node-healthy` | pod phase == Running | BNs stay up under chaos |
| `block-available` | `serverStatus` block range | BNs still hold their blocks |
| `blocks-increasing` | Δblocks > 0 over a window (polled, retried) | Production is *moving* (not stalled) |
| `block-rate-floor` | Δ`publisher_highest_block_number_inbound` ÷ Δt ≥ floor | Post-clear throughput **recovered** to a floor (0.000/s == a real stall) |
| `metric-threshold` | generic Prometheus comparator (label-tolerant scrape) | e.g. `publisher_stream_errors_total == 0` — no stream errors across the run |
| `backfill-triggered` | greps BN logs for the `backfill` marker in the window | **The node actually fell behind and recovered via backfill** |
| `signature-transition` | detects the block-proof size jump 2920→3432 bytes | TSS still transitions **Schnorr → WRAPS** under latency |

**Why `backfill-triggered` is the key signal that latency bit.** A Block Node only backfills when it has
fallen behind its source/peers. On a healthy, low-latency cluster it never needs to. So observing real
`backfill` activity during the stress/severe window is *direct evidence that the injected latency had a
material effect* — and the subsequent `blocks-increasing` + `block-rate-floor` prove the node then **caught
up and resumed normal production** after the latency was cleared.

---

## Confirming presence and recovery — CI evidence

Every tier has been run on `paired-3` with TSS on. The stress/severe runs show the full **inject → degrade →
backfill → recover** cycle:

| Run | Test | Latency injected | Backfill (presence + impact) | Recovery (post-clear rate vs floor) | Result |
|-----|------|------------------|------------------------------|--------------------------------------|--------|
| `28335141168` | chaos-foundation-smoke | install→inject→clear→cleanup | — | — | ✅ |
| `28335794369` | latency-cn-to-cn (+TSS) | 100±20 ms | — | steady; **Schnorr→WRAPS at block 434** | ✅ 2/2 |
| `28311376970` | latency-all-three | 100/150/200 ms | — | steady | ✅ |
| `28545982415` | **latency-stress** | 300/450/600 ms | **real backfill — 199 log hits ×3 BNs** | 0.467–0.500 blk/s vs **0.2** floor (~2.4×) | ✅ 5/5 |
| `28547498881` | **latency-severe** | 600/800/1000 ms | **real backfill — 225 log hits ×3 BNs** | 0.433–0.500 blk/s vs **0.1** floor (~4.5×) | ✅ 5/5 |

Read the stress/severe rows as: latency was applied → all three BNs measurably fell behind and **triggered
backfill** (proof the latency was real and impactful) → after `clear-latency` they **recovered** to normal
throughput and TSS signatures still verified. That is the behaviour a node operator cares about under a
degraded WAN.

> **Measurement note:** assertions run *after* the chaos window, so `block-rate-floor` measures **recovery**,
> not the (intentionally degraded) under-chaos rate. The under-chaos impact is captured by `backfill-triggered`.

---

## Running it

TSS runs **on** by default (`CN_VERSION=latest` is min-enforced to a TSS-capable tag that clears Solo's
`--wraps` floor). Run **one tier at a time** — `paired-3` shares a `cancel-in-progress` concurrency lane, so
a second same-topology dispatch cancels the first.

```bash
# smoke first, then a dimension / tier (one at a time)
gh workflow run solo-e2e-test.yml --ref feature/solo-e2e-latency -f topology=paired-3 -f tss-enabled=true -f test-definition=chaos-foundation-smoke
gh workflow run solo-e2e-test.yml --ref feature/solo-e2e-latency -f topology=paired-3 -f tss-enabled=true -f test-definition=latency-stress
gh workflow run solo-e2e-test.yml --ref feature/solo-e2e-latency -f topology=paired-3 -f tss-enabled=true -f test-definition=latency-severe
```
UI equivalent: Actions → Solo E2E Test → Run workflow → branch `feature/solo-e2e-latency`, `topology=paired-3`,
`tss-enabled=true`, `test-definition=<name>`. The workflow installs Chaos Mesh automatically for
`chaos`/`latency` tests. Full profile guide: `tools-and-tests/scripts/solo-e2e-test/docs/latency-scenarios.md`.

---

## Scope / impact

- Product code, Helm chart, proto: **untouched.**
- Latency injection is opt-in (only when a `chaos`/`latency` test is selected); the assertion library ships
  with unit tests (20 cases over fixtures).
- Adds two docs (`README` chaos section + `docs/latency-scenarios.md`) that are self-contained (no external
  links) for team handoff.

## For reviewers
- The heavy tiers (`stress`, `severe`) are CI-validated with comfortable margins (above), so the shipped
  thresholds are calibrated, not guesses.
- `loss` (netem packet loss) is plumbed end-to-end but no shipped scenario exercises it yet — noted in the
  docs as a future knob.

## Related Issue(s)
Resolves #2125

### DIFF
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/chaos-cleanup.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/chaos-cleanup.sh
@@ -42,9 +42,6 @@ if ! kubectl api-resources --api-group=chaos-mesh.org -o name >/dev/null 2>&1; t
   exit 0
 fi
 
-selector_arg=()
-[[ -n "${LABEL_SELECTOR}" ]] && selector_arg=(-l "${LABEL_SELECTOR}")
-
 failed=0
 IFS=',' read -ra KIND_LIST <<< "${KINDS}"
 for kind in "${KIND_LIST[@]}"; do
@@ -56,7 +53,14 @@ for kind in "${KIND_LIST[@]}"; do
     continue
   fi
   [[ "${QUIET}" == "true" ]] || echo "Deleting ${kind_trimmed} resources${LABEL_SELECTOR:+ (selector: ${LABEL_SELECTOR})}..."
-  if ! kubectl delete "${kind_trimmed}" --all-namespaces --all "${selector_arg[@]}" --ignore-not-found 2>&1 | sed 's/^/  /'; then
+  # --all and -l (label selector) are mutually exclusive in kubectl delete;
+  # use the selector path when one is given, the --all path otherwise.
+  if [[ -n "${LABEL_SELECTOR}" ]]; then
+    delete_args=(--all-namespaces -l "${LABEL_SELECTOR}" --ignore-not-found)
+  else
+    delete_args=(--all-namespaces --all --ignore-not-found)
+  fi
+  if ! kubectl delete "${kind_trimmed}" "${delete_args[@]}" 2>&1 | sed 's/^/  /'; then
     failed=1
   fi
 done
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/chaos-cleanup.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/chaos-cleanup.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#
+# Deletes chaos resources (NetworkChaos, PodChaos) without uninstalling
+# Chaos Mesh itself. Tolerates the case where Chaos Mesh CRDs are not
+# installed (returns success silently) so it is safe to wire into 'task down'
+# even when chaos was never enabled.
+#
+# Usage:
+#   ./chaos-cleanup.sh [options]
+#
+# Options:
+#   --label-selector SEL   Only delete resources matching this label selector
+#                          (e.g. solo-e2e-test/test-id=my-test). If omitted,
+#                          deletes ALL chaos resources in the cluster.
+#   --kinds KINDS          Comma-separated chaos kinds to delete
+#                          (default: networkchaos,podchaos)
+#   --quiet                Suppress per-kind logging
+#   --help                 Show this help message
+
+set -o pipefail
+
+LABEL_SELECTOR=""
+KINDS="networkchaos,podchaos"
+QUIET=false
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --label-selector) LABEL_SELECTOR="$2"; shift 2 ;;
+    --kinds)          KINDS="$2"; shift 2 ;;
+    --quiet)          QUIET=true; shift ;;
+    --help|-h)        sed -n '2,/^$/p' "$0" | sed 's/^# \{0,1\}//'; exit 0 ;;
+    *) echo "Unknown option: $1"; exit 1 ;;
+  esac
+done
+
+command -v kubectl >/dev/null 2>&1 || { echo "ERROR: kubectl not found"; exit 1; }
+
+# If chaos-mesh CRDs are not installed, nothing to do.
+if ! kubectl api-resources --api-group=chaos-mesh.org -o name >/dev/null 2>&1; then
+  [[ "${QUIET}" == "true" ]] || echo "Chaos Mesh CRDs not installed; nothing to clean."
+  exit 0
+fi
+
+selector_arg=()
+[[ -n "${LABEL_SELECTOR}" ]] && selector_arg=(-l "${LABEL_SELECTOR}")
+
+failed=0
+IFS=',' read -ra KIND_LIST <<< "${KINDS}"
+for kind in "${KIND_LIST[@]}"; do
+  kind_trimmed="${kind// /}"
+  [[ -z "${kind_trimmed}" ]] && continue
+  if ! kubectl api-resources --api-group=chaos-mesh.org -o name 2>/dev/null \
+       | grep -qx "${kind_trimmed}.chaos-mesh.org"; then
+    [[ "${QUIET}" == "true" ]] || echo "Kind '${kind_trimmed}' not present; skipping."
+    continue
+  fi
+  [[ "${QUIET}" == "true" ]] || echo "Deleting ${kind_trimmed} resources${LABEL_SELECTOR:+ (selector: ${LABEL_SELECTOR})}..."
+  if ! kubectl delete "${kind_trimmed}" --all-namespaces --all "${selector_arg[@]}" --ignore-not-found 2>&1 | sed 's/^/  /'; then
+    failed=1
+  fi
+done
+
+exit "${failed}"
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
@@ -57,6 +57,36 @@ ASSERTIONS_FAILED=0
 ASSERTION_RESULTS_FILE="/tmp/solo-test-assert-results-$$"
 : > "${ASSERTION_RESULTS_FILE}"
 
+# Chaos: track applied NetworkChaos resource names so we can clean up on exit
+# even if the test fails between inject-latency and clear-latency.
+#
+# Cleanup contract:
+#   - execute_inject_latency appends the resource name to CHAOS_ACTIVE_FILE.
+#   - execute_clear_latency removes it on graceful clear.
+#   - cleanup_chaos_on_exit (registered on EXIT below) drains any remaining
+#     names so a crash between inject and clear doesn't leak NetworkChaos
+#     resources into the cluster.
+# The file is PID-scoped (`$$`), so two concurrent runner invocations in the
+# same shell session don't share state. Running two test runners in parallel
+# is not a supported workflow anyway.
+CHAOS_TEMPLATE_DIR="${CHAOS_TEMPLATE_DIR:-${SCRIPT_DIR}/chaos-templates}"
+CHAOS_ACTIVE_FILE="/tmp/solo-test-chaos-active-$$"
+: > "${CHAOS_ACTIVE_FILE}"
+
+function cleanup_chaos_on_exit {
+    if [[ -s "${CHAOS_ACTIVE_FILE}" ]]; then
+        echo ""
+        echo "Cleaning up active NetworkChaos resources..."
+        while IFS= read -r name; do
+            [[ -z "$name" ]] && continue
+            kctl delete networkchaos -n chaos-mesh "$name" --ignore-not-found 2>/dev/null \
+                | sed 's/^/  /' || true
+        done < "${CHAOS_ACTIVE_FILE}"
+    fi
+    rm -f "${CHAOS_ACTIVE_FILE}" 2>/dev/null
+}
+trap cleanup_chaos_on_exit EXIT
+
 
 # ============================================================================
 # Argument Parsing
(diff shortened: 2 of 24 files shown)


tools-and-tests/scripts/solo-e2e-test/scripts/chaos-cleanup.sh
```diff
@@ -0,0 +1,64 @@
...
+for kind in "${KIND_LIST[@]}"; do
+  kind_trimmed="${kind// /}"
+  [[ -z "${kind_trimmed}" ]] && continue
+  if ! kubectl api-resources --api-group=chaos-mesh.org -o name 2>/dev/null \
+       | grep -qx "${kind_trimmed}.chaos-mesh.org"; then
+    [[ "${QUIET}" == "true" ]] || echo "Kind '${kind_trimmed}' not present; skipping."
+    continue
+  fi
+  [[ "${QUIET}" == "true" ]] || echo "Deleting ${kind_trimmed} resources${LABEL_SELECTOR:+ (selector: ${LABEL_SELECTOR})}..."
+  if ! kubectl delete "${kind_trimmed}" --all-namespaces --all "${selector_arg[@]}" --ignore-not-found 2>&1 | sed 's/^/  /'; then
```
berryware: Label selector and -all cannot be used together. It is either or.  Returns error "error: setting 'all' parameter but found a non empty selector."

AlfredoG87 (dismissed):
LGTM 💯

AlfredoG87 (approved):

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
```diff
@@ -482,6 +512,145 @@ function execute_reconfigure_cn_streaming {
...
+function execute_inject_latency {
+    local args="$1"
+    local name source_kind target_kind latency jitter correlation bidirectional loss
+    name=$(echo "$args" | yq '.name // ""')
+    source_kind=$(echo "$args" | yq '.source.kind // ""')
+    target_kind=$(echo "$args" | yq '.target.kind // ""')
+    latency=$(echo "$args" | yq '.latency // "0ms"')
+    jitter=$(echo "$args" | yq '.jitter // "0ms"')
+    correlation=$(echo "$args" | yq '.correlation // "0"')
+    bidirectional=$(echo "$args" | yq '.bidirectional // true')
```
berryware, line 553: This always evaluates to `true`:
```echo '{"bidirectional": false}' | yq '.bidirectional // true'   # => true```. 

All scenarios currently set it to true so it is masked right now.

File: tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh [hunk 3]
```diff
@@ -482,6 +512,145 @@ function execute_reconfigure_cn_streaming {
...
+    export SOURCE_LABEL_KEY="${source_key}"
+    export SOURCE_LABEL_VALUE="${source_val}"
+    export TARGET_LABEL_KEY="${target_key}"
+    export TARGET_LABEL_VALUE="${target_val}"
+    export LATENCY="${latency}"
+    export JITTER="${jitter}"
+    export CORRELATION="${correlation}"
+    export LOSS_BLOCK="${loss_block}"
+
+    envsubst < "${CHAOS_TEMPLATE_DIR}/network-latency.yaml.tmpl" > "${manifest}"
```
berryware, line 622: Does envsubst need to be verified in `Taskfile.yml's check` task?

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
```diff
@@ -482,6 +512,145 @@ function execute_reconfigure_cn_streaming {
...
+
+# Slug a string into a Kubernetes-name-safe lowercase token.
+function chaos_slug {
+    echo "$1" | tr 'A-Z' 'a-z' | sed -E 's/[^a-z0-9-]+/-/g; s/^-+//; s/-+$//'
+}
+
+# Deterministic NetworkChaos resource name for a (test, scenario) pair.
+# Used by both inject-latency and clear-latency so they cannot disagree on the
+# resource name a future clear must target.
+function chaos_resource_name {
```
berryware, line 538: Nit: Does not check for a trailing '-' which would be rejected by `kubectl apply`.

tools-and-tests/scripts/solo-e2e-test/scripts/test/test-chaos-assertions.sh
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#
+# Fixture-based unit tests for the Phase 3 latency-aware assertions in
```
berryware, line 4: Are these to be run manually? I did not see a task for running them.

tools-and-tests/scripts/solo-e2e-test/scripts/test/fixtures/sample-logs-with-backfill.txt
```diff
@@ -0,0 +1,6 @@
+2026-05-11T20:01:00Z INFO  org.hiero.block.node.backfill.BackfillFetcher - Loaded backfill source node: block-node-2.solo-network.svc.cluster.local:40840
+2026-05-11T20:01:05Z INFO  org.hiero.block.node.NetworkApp - Block Node started successfully
+2026-05-11T20:01:10Z DEBUG org.hiero.block.node.verification.VerificationServicePlugin - Received backfill notification for block=412
+2026-05-11T20:01:11Z DEBUG org.hiero.block.node.verification.VerificationServicePlugin - Received backfill notification for block=413
+2026-05-11T20:01:12Z DEBUG org.hiero.block.node.verification.VerificationServicePlugin - Received backfill notification for block=414
+2026-05-11T20:01:15Z INFO  org.hiero.block.node.publisher.PublisherHandler - New publisher connection from network-node1
```
jsync-swirlds, line 6: Not sure we want to include sample logs; particularly given this is already obsolete and includes logs from a plugin we replaced recently.

jsync-swirlds (approved):

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
Nana-EC: Good catch, will address in 3312

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
Nana-EC: Will follow up in 3312

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
Nana-EC: Will follow up in 3312

tools-and-tests/scripts/solo-e2e-test/scripts/test/test-chaos-assertions.sh
Nana-EC: Will follow up in 3312

tools-and-tests/scripts/solo-e2e-test/scripts/test/fixtures/sample-logs-with-backfill.txt
Nana-EC: Will follow up in 3312

PR #2803, opened 2026-05-13
refactor: relax RSA WRB acceptance to at-least-one-valid signature

### LINKED ISSUE #2801: improvement(verification): For WRB SignedRecordFileProof needs only the presence of 1 signature
### Story Form

As a BlockNode Operator
I want to make sure that the WRB Verification pass
So that Verification of Blocks that have `SignedRecordFileProof` pass when at least 1 signature is present

### Technical Notes

N/A

AlfredoG87:
Replace the strict-majority RSA WRB acceptance rule (`validCount * 2 > rosterSize`) with at-least-one-valid (`validCount >= 1`) in `ExtendedMerkleTreeSession#verifyRsaProof`. The fail-fast rule "any signature present that fails cryptographic verification rejects the block" is unchanged and still covers tampering detection.

`RsaWrbVerificationTest` updated to match the new contract: flipped assertions on the now-accepting cases (3-of-6, duplicate-nodeId), added a single-signature acceptance test, replaced the parametrised strict-majority test with a roster-size-parametrised `oneValidSignatureSuffices_zeroRejected`, and reworked `multipleProofs_oneFailing_rejected` to fail via a wrong-payload signature instead of sub-quorum.

Fixes #2801

### DIFF
--- a/block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
+++ b/block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
@@ -397,9 +397,10 @@ public class ExtendedMerkleTreeSession implements VerificationSession {
      *    - Verify with `SHA384withRSA`. If verification fails or throws, **reject the block
      *      immediately** — the CN only includes signatures from nodes that contributed to
      *      consensus, so any included signature must be cryptographically valid.
-     * 5. Accept if `validCount * 2 > rosterSize` (strict majority: more than half of the
-     *      address-book nodes must have signed). The CN sends exactly the signatures from
-     *      nodes whose combined consensus weight exceeded 50%.
+     * 5. Accept if at least one signature in the proof verified. Step 4 already rejects
+     *      the block on any failed verification, so reaching this step means every
+     *      signature we were able to verify did pass; we only need to confirm at least
+     *      one such verification occurred.
      *
      * <p>This method never throws; all error conditions return a `success=false` notification.
      *
@@ -529,20 +530,18 @@ public class ExtendedMerkleTreeSession implements VerificationSession {
             metrics.incrementRsaRosterMismatch(mismatchCount);
         }
 
-        // Strict majority threshold: validCount must be strictly greater than half the roster.
-        // The CN only sends signatures from nodes whose combined consensus weight exceeded 50%,
-        // so we require validCount * 2 > rosterSize (equivalent to validCount > rosterSize / 2.0).
-        // For example: 6-node roster requires > 3, i.e. ≥ 4 valid signatures.
-        final boolean accepted = validCount * 2 > rosterSize;
+        // Acceptance threshold: at least one valid signature from a roster node.
+        // Every signature present in the proof that failed cryptographic verification already
+        // rejected the block in the loop above, so reaching this point with validCount >= 1
+        // means all signatures we attempted to verify did pass.
+        final boolean accepted = validCount >= 1;
 
         if (!accepted) {
             LOGGER.log(
                     WARNING,
-                    "RSA WRB proof rejected for block {0}: {1}/{2} valid sigs, need > {3}",
+                    "RSA WRB proof rejected for block {0}: 0 valid signatures (roster size {1})",
                     blockNumber,
-                    validCount,
-                    rosterSize,
-                    rosterSize / 2);
+                    rosterSize);
             metrics.incrementRsaFailure();
             return new VerificationNotification(
                     false, FailureType.BAD_BLOCK_PROOF, blockNumber, null, null, blockSource);
@@ -564,11 +563,10 @@ public class ExtendedMerkleTreeSession implements VerificationSession {
 
         LOGGER.log(
                 DEBUG,
-                "RSA WRB proof accepted for block {0}: {1}/{2} valid sigs (need > {3})",
+                "RSA WRB proof accepted for block {0}: {1} valid signature(s) (roster size {2})",
                 blockNumber,
                 validCount,
-                rosterSize,
-                rosterSize / 2);
+                rosterSize);
         metrics.incrementRsaSuccess();
         return new VerificationNotification(
                 true, null, blockNumber, blockRootHash, new BlockUnparsed(blockItems), blockSource);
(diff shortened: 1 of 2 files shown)


block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -529,20 +530,18 @@ private VerificationNotification verifyRsaProof(
...
-        // Strict majority threshold: validCount must be strictly greater than half the roster.
-        // The CN only sends signatures from nodes whose combined consensus weight exceeded 50%,
-        // so we require validCount * 2 > rosterSize (equivalent to validCount > rosterSize / 2.0).
-        // For example: 6-node roster requires > 3, i.e. ≥ 4 valid signatures.
-        final boolean accepted = validCount * 2 > rosterSize;
+        // Acceptance threshold: at least one valid signature from a roster node.
+        // Every signature present in the proof that failed cryptographic verification already
+        // rejected the block in the loop above, so reaching this point with validCount >= 1
+        // means all signatures we attempted to verify did pass.
+        final boolean accepted = validCount >= 1;
```
Nana-EC: nit: greater than 0 i.e. at least 1 also works
```suggestion
        final boolean accepted = validCount > 0;
```

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -529,20 +530,18 @@ private VerificationNotification verifyRsaProof(
             metrics.incrementRsaRosterMismatch(mismatchCount);
         }
 
-        // Strict majority threshold: validCount must be strictly greater than half the roster.
-        // The CN only sends signatures from nodes whose combined consensus weight exceeded 50%,
-        // so we require validCount * 2 > rosterSize (equivalent to validCount > rosterSize / 2.0).
-        // For example: 6-node roster requires > 3, i.e. ≥ 4 valid signatures.
-        final boolean accepted = validCount * 2 > rosterSize;
+        // Acceptance threshold: at least one valid signature from a roster node.
+        // Every signature present in the proof that failed cryptographic verification already
```
Nana-EC: Is it more appropriate to say verification fails on the first bad signature so reaching here means all signatures were good?
I reach the sentence and was unclear if you means in a set of signatures the bad ones are rejected and the good ones kept.
However, from tests I see you fail fast on the first encountered bad signature.
This comment should line up there

Nana-EC (dismissed):
LG, nit suggestions for consideration if applicable

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: `> 0` is arithmetically the same as `>= 1` (both accept 1 or more) if we're using purely integer calculation (i.e. there are no numbers between 0 and 1).

I think the latter is easier to read, but that's just my preference.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: Perhaps reword as follows?
```suggestion
        // Acceptance threshold: Every signature present in the proof passes
        // validation, and the proof contains at least 1 signature.
        // Because we fail fast if any signatures fail, reaching this point means
        // all signatures passed verification, so we only need to check that the
        // count is one or more (>= 1).
```

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -397,9 +397,10 @@ private LedgerIdPublicationTransactionBody findLedgerIdPublication(Bytes signedT
      *    - Verify with `SHA384withRSA`. If verification fails or throws, **reject the block
      *      immediately** — the CN only includes signatures from nodes that contributed to
      *      consensus, so any included signature must be cryptographically valid.
-     * 5. Accept if `validCount * 2 > rosterSize` (strict majority: more than half of the
-     *      address-book nodes must have signed). The CN sends exactly the signatures from
-     *      nodes whose combined consensus weight exceeded 50%.
+     * 5. Accept if at least one signature in the proof verified. Step 4 already rejects
+     *      the block on any failed verification, so reaching this step means every
+     *      signature we were able to verify did pass; we only need to confirm at least
+     *      one such verification occurred.
```
jsync-swirlds: Possible reword for clarity:
```suggestion
     * 5. Accept if all signatures in the proof verified. Step 4 already rejects
     *      the block on any failed signature, so reaching this step means every
     *      signature that was present did pass; we only need to confirm at least
     *      one such signature exists.
```

block-node/verification/src/test/java/org/hiero/block/node/verification/session/impl/RsaWrbVerificationTest.java
```diff
@@ -44,28 +44,24 @@
  * `ExtendedMerkleTreeSession`.
  *
  * <p>Tests use synthetic blocks and in-process RSA key pairs — no real testnet blocks are required.
- * Six nodes are used so the strict majority threshold (`> N/2 = > 3` for 6 nodes, i.e. ≥ 4) is
- * testable with both passing (≥ 4 valid sigs) and failing (≤ 3 valid sigs) scenarios.
- * All signatures present must pass RSA verification — any single failed sig causes immediate
- * block rejection, regardless of how many other sigs are valid.
+ * The acceptance rule is: at least one signature in the proof must verify, AND every signature
+ * present that we are able to verify must succeed — any single failed cryptographic verification
```
jsync-swirlds:
```suggestion
 * present must succeed — any single failed cryptographic verification
```

jsync-swirlds (dismissed):
Looks good, some possible rewords of comments for clarity.
Also, looks like spotless is sad again.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
AlfredoG87: I will improve the wording, I also had a second thought on this 😅

jsync-swirlds (approved):
Issue 2808 is critical to implement, but this will unblock testing for now.

Nana-EC (approved):
LG

Issue #2236, opened 2026-02-23
Enhancements to support plugin ready/live status


jsync-swirlds:
### Goal

Ensure that the block node health plugin has access to ready/live status for each plugin, and each plugin updates its own ready/live status whenever needed.

### Intention

We intend to add methods to the block node SPI that a plugin may use to set its own ready and live status whenever needed.
We intend to adjust existing init/start methods to allow plugins to set their ready and live status for init and start process.
We intend to adjust each plugin to use the newly added SPI capabilities to set and manage ready/live status appropriately.
We intend to remove the existing `shutdown` method entirely; the health plugin will report block node health status and the orchestration system will manage shutdown and restart if needed.

### Technical Details

* Individual plugins will only report their own status, they will not be able to directly affect the status of the entire block node.
   * Even the health plugin cannot shut down the app, it simply reports status to the orchestration (typically a Kubernetes instance), and that external orchestration layer manages starting, ending, and/or restarting the application.
* With this epic we will adjust the health plugin to report not ready if _any_ plugin is not ready, and not live if _any_ plugin is not live.
   * In future work we will add configuration and support in the health plugin to determine application ready/live status based on plugin priority (in health plugin configuration), the number of plugins in each status, and additional factors as appropriate.

Issue #2038, opened 2026-01-23
`HistoricalBlockFacility` as a Separate Plugin


ata-nas:
### Story Form

As a Block Node Operator
I want to have each and every plugin discoverable via SPI, even facilities
So that I can control what functionality and services are deployed for my node.

### Technical Notes

- Currently the historical block facility is implemented directly in the app module
- It is also hard coded and not a standalone plugin
- We need to extract that implementation as a separate plugin and load it via SPI just like messaging
- We should most certainly consider renaming this facility as it conflicts with the "historical" plugins semantically
- This should be done after the pick and choose logic is merged (project pikachu)

AlfredoG87:
@ata-nas @berryware @Nana-EC @jsync-swirlds 
This seems like is not necessary after all, I was able to run a BN without the HistoricalPlugin and it worked just fine... 
Maybe for long terms of periods that is not the case, but the App does not fail to start or to work as this issue suggests.

[noHistoricalPluginLogs.log](https://github.com/user-attachments/files/25194765/noHistoricalPluginLogs.log)
[minimalBlockNodeDeployment.log](https://github.com/user-attachments/files/25194764/minimalBlockNodeDeployment.log)


I recommend we should close this task together with the PR #2158

Issue #1426, opened 2025-07-24
Make adding new notifications Generic, so plugins can dynamically add new notifications


AlfredoG87:
general note, It seems we've stumbled into a situation where every notification requires modifying this file.  That's not good.  We need to have _generic_ notification so that third party plugins can add their own notifications without having to modify the Hiero codebase.

Something to bring up in design and probably would be good to file an issue (I have a note to do so if you prefer not to).

_Originally posted by @jsync-swirlds in https://github.com/hiero-ledger/hiero-block-node/pull/1384#discussion_r2227082397_

jsync-swirlds:
Note: This is a bit larger than a single issue, so I created an epic [here](https://github.com/hiero-ledger/hiero-block-node/issues/1431).

PR #2830, opened 2026-05-15
docs: Document the plugin architecture as it is today, and then document the changes

### LINKED ISSUE #2818: Design doc for enhancing the plugin architecture to include plugin-to-plugin dependencies
### Story Form

As a BlockNode engineer I would like to be able to specify a plugins dependency on other plugins and to know that the dependent plugin is "ready" to provide its services.

### Technical Notes

[x] - Document the plugin architecture as-is.
[x] - Refactor the plugin architecture to allow plugin-to-plugin dependencies
 - JPMS 
   - will be used to identify dependencies between plugins
   - provider plugins will implement the `BlockNodePlugin` interface as usual
   - provider plugins will implement their own interface, ie `ApplicationStateFacility,` so that a dependency can be made
   - dependent plugins will add a `uses` provider plugin interface in their `module-info.java` file.
 - PROVIDER Plugins
   - provider plugins should provide a method for signaling that it is ready to provide its services.
   - provider plugins will implement their own interface, ie `ApplicationStateFacility,` so that a dependency can be made
   - Special entities, (ApplicationStateFacility, HealthService, HistoricalBlockFacility, etc), will be moved from the BlockNodeContext and made into Provider plugins.
   - dependencies to  these plugins will be specified in the respective dependent `module-info.java` file.
 - PLUGIN Method usages
   - `init()` Checks, using service loader,  to see if the provider plugins it needs exist. Should log an error message if a dependency is not met.
     -  If required plugins are not available, a plugin should mark itself unhealthy/failed.
   - `start()` Load data and resources. Start any worker threads.
   - `stop()` Close resources, stop worker threads.

berryware:

## Reviewer Notes
- Documented the plugin architecture as it is today to use as a basis for the refactoring of the plugin architecture.
  - current-plugin-architecture.md
- Documented the changes to the plugin architecture to remove Facility plugins from the BlockNodeContext and use JPMS to define dependencies between plugins.
  - new-plugin-architecture.md 

## Related Issue(s)
Closes #2818


docs/design/architecture/current-plugin-architecture.md
```diff
@@ -0,0 +1,415 @@
...
+```mermaid
+flowchart LR
+    PUB[Publisher Plugin] -->|sendBlockItems| BI[Block Items\nDisruptor]
+    BI --> SUB[Subscriber Plugin]
+    BI --> VER[Verification Plugin]
+    BI --> N1[... Plugin N]
+
+    VER -->|sendBlockVerification| BN[Block Notification\nDisruptor]
+    STOR[Storage Plugin] -->|sendBlockPersisted| BN
+    BN --> SUB2[Subscriber Plugin]
```
jsync-swirlds, line 347: Should add `BN --> STOR[Storage Plugin]` here, as storage plugins depend on the verification notification for the block to be stored.

docs/design/architecture/current-plugin-architecture.md
```diff
@@ -0,0 +1,415 @@
...
+    BN --> PUB2[Publisher Plugin]
+    BN --> N2[... Plugin N]
+```
+
+### Module Dependency Structure
+
+```mermaid
+flowchart TD
```
jsync-swirlds, line 358: Not entirely convinced we need/want to list every plugin as that list will change frequently over time and (if we are fully successful) these will only ever represent a relatively small fraction of available plugins.
Perhaps just list general categories (i.e. "core" plugins, "community" plugins, and "private" plugins)?

docs/design/architecture/current-plugin-architecture.md
```diff
@@ -0,0 +1,415 @@
...
+- **No-back-pressure skip**: A `NoBackPressureBlockItemHandler` that processes slowly receives `onTooFarBehindError()` calls under sustained load without causing the publisher to block.
+- **Configuration scoping**: A plugin's `@ConfigData` record is populated from its declared property keys; unrelated keys from other plugins do not affect its values.
+- **Graceful shutdown**: Calling `serverHealth.shutdown(...)` results in `stop()` being called on every plugin, the web server closing, and the JVM exiting with code 0.
+- **HTTP route registration**: A plugin that calls `serviceBuilder.registerHttpService("/foo", ...)` in `init()` has its handler reachable at `GET /foo` once `start()` completes.
+- **gRPC route registration**: A plugin that calls `serviceBuilder.registerGrpcService(...)` has its service reachable via HTTP/2 gRPC once `start()` completes.
+- **Block provider priority**: When two `BlockProviderPlugin` implementations both have a block at number N, the provider with the higher `defaultPriority()` value is the one whose `BlockAccessor` is returned by `HistoricalBlockFacility.block(N)`.
```
jsync-swirlds: These sections make sense for individual code modules, but not so much for an architecture document, as they get into details of the implementation.
They perhaps should be moved to the design document for the App module.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,549 @@
...
+  <dd>A slimmed-down immutable Java record passed to every plugin during <code>init()</code>. It carries only infrastructure concerns: <code>Configuration</code>, <code>MetricRegistry</code>, <code>ServiceLoaderFunction</code>, <code>ThreadPoolManager</code>, and <code>BlockNodeVersions</code>. Facilities are no longer fields on this record.</dd>
+
+  <dt>ServiceBuilder</dt>
+  <dd>An interface passed alongside <code>BlockNodeContext</code> during <code>init()</code> that allows plugins to register HTTP and gRPC service routes on the Helidon web server.</dd>
+
+  <dt>Disruptor</dt>
+  <dd>The LMAX Disruptor ring buffer used by the <code>BlockMessagingFacility</code> plugin as the underlying transport for both block-item distribution and block-notification events between plugins.</dd>
+
+  <dt>BlockProviderPlugin</dt>
+  <dd>A specialization of <code>BlockNodePlugin</code> that contributes a source of historical blocks (e.g., file system, S3, RAM cache). The <code>HistoricalBlockFacility</code> plugin composes all registered providers in priority order.</dd>
```
jsync-swirlds: This should probably not be present in the redesigned architecture.
Plugins that need blocks can add `uses org.hiero.block.node.spi.BlockSource` and plugins that can provide blocks add `provides org.hiero.block.node.spi.BlockSource with Implementation` to module info.  Then the block reader and block sources connect directly (and priority is by reader instead of a globally determined order).

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,549 @@
...
+
+|        Field        |          Type           |                                       Description                                        |
+|---------------------|-------------------------|------------------------------------------------------------------------------------------|
+| `configuration`     | `Configuration`         | Swirlds typed configuration. Plugins call `configuration.getConfigData(MyConfig.class)`. |
+| `metricRegistry`    | `MetricRegistry`        | Register counters, gauges, and histograms.                                               |
+| `serviceLoader`     | `ServiceLoaderFunction` | Load facility plugins and other SPI extensions at runtime.                               |
+| `threadPoolManager` | `ThreadPoolManager`     | Create managed virtual-thread or platform-thread executors.                              |
+| `blockNodeVersions` | `BlockNodeVersions`     | Version information for all loaded plugins.                                              |
```
jsync-swirlds: I wonder if this could be moved to the status plugin instead of context?

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,549 @@
...
+Passed to plugins during `init()`. Allows plugins to register routes without a direct dependency on Helidon.
+
+```java
+void registerHttpService(String path, HttpService... service);
+void registerGrpcService(ServiceInterface service);
+```
+
+### Facility Plugin Interfaces
+
+Each facility is defined by its own interface in the `spi-plugins` module. A facility plugin implements both `BlockNodePlugin` and its facility interface. The facility interface is registered as a JPMS service independently of `BlockNodePlugin`.
```
jsync-swirlds: Rather than adding special interfaces in SPI, why not just remove the "facility" concept entirely and just have plugins?
A plugin can declare it's own interfaces (perhaps in a &lt;plugin&gt;-api module) that it "provides", and a dependent plugin just "uses" that API.
Then a "facility" isn't a special category at all.

The most important item there (in my opinion) is that we don't want to be adding plugin-specific interfaces to the `spi-plugins` because that continues a roadblock in the way of individuals or groups adding their own "facility type" plugins.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,549 @@
...
+
+**Block Notifications** — five typed notification events for block lifecycle milestones:
+
+|              Notification               |       Sender        |                          Meaning                           |
+|-----------------------------------------|---------------------|------------------------------------------------------------|
+| `VerificationNotification`              | Verification plugin | Block passed or failed cryptographic verification.         |
+| `PersistedNotification`                 | Storage plugins     | Block has been durably written by a provider.              |
+| `BackfilledBlockNotification`           | Backfill plugin     | A previously missing block was retrieved and stored.       |
+| `NewestBlockKnownToNetworkNotification` | Publisher plugin    | Consensus Node has reported the latest known block header. |
+| `PublisherStatusUpdateNotification`     | Publisher plugin    | Publisher connection state changed.                        |
```
jsync-swirlds, line 122: Eventually we need to have a generic Notification mechanism; this strongly typed mechanism makes it impossible to add notifications for a new plugin without adding code to the Hiero project, which isn't ideal.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,549 @@
...
+void registerGrpcService(ServiceInterface service);
+```
+
+### Facility Plugin Interfaces
+
+Each facility is defined by its own interface in the `spi-plugins` module. A facility plugin implements both `BlockNodePlugin` and its facility interface. The facility interface is registered as a JPMS service independently of `BlockNodePlugin`.
+
+#### `BlockMessagingFacility` (facility interface)
+
+The inter-plugin event bus. Built on LMAX Disruptor ring buffers for high-throughput, low-latency delivery.
```
jsync-swirlds: Perhaps note that _the current implementation_ uses LMAX, but that's not guaranteed to remain true.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,549 @@
...
+| `PublisherStatusUpdateNotification`     | Publisher plugin    | Publisher connection state changed.                        |
+
+Any plugin may subscribe to notifications via `registerBlockNotificationHandler(BlockNotificationHandler, ...)`.
+
+#### `HistoricalBlockFacility` (facility interface)
+
+Aggregates all registered `BlockProviderPlugin` implementations and presents a unified read interface:
+
+- `block(long blockNumber)` → `BlockAccessor` — queries providers in descending priority order; returns the first non-null result.
+- `availableBlocks()` → `BlockRangeSet` — the union of all providers' available ranges.
```
jsync-swirlds: I would like to remove this "facility" entirely, without replacement.
I do not think it's necessary if plugins that want blocks can just directly reference block sources with a `uses` clause.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,549 @@
...
+
+#### `HealthFacility` (facility interface)
+
+```java
+enum State { STARTING, RUNNING, SHUTTING_DOWN }
+State blockNodeState();
+void shutdown(String className, String reason);
+```
+
+The `HealthServicePlugin` registers `/healthz/livez` and `/healthz/readyz` HTTP endpoints that return `200 OK` while the state is `RUNNING` and `503 Service Unavailable` otherwise.
```
jsync-swirlds: Here also, there should just be a HealthPlugin with no special status (i.e. it's not listed here, it just has its own design doc and plugins directly reference it.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,549 @@
...
+// Change listener registration
+void registerTssDataListener(Consumer<TssData> listener);
+void registerAddressBookListener(Consumer<NodeAddressBook> listener);
+```
+
+Plugins register listeners in `init()` so they receive updates as soon as the facility becomes ready. The facility delivers the current value immediately upon listener registration if state is already available, and delivers each subsequent update as it occurs.
+
+### `BlockProviderPlugin` (interface)
+
+Specialization of `BlockNodePlugin` for block storage backends. Each provider declares a `defaultPriority()` (higher = preferred). The `HistoricalBlockFacility` plugin loads all providers and sorts them by priority.
```
jsync-swirlds: These, also, should disappear.
Application state should declare its own interface.

BlockProviderPlugin is _perhaps_ worth including in the SPI, but it might also be worth having a separate `-api` module just for that interface (which keeps the pattern and approach clean).

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,549 @@
...
+
+## Design
+
+### Facility Plugin Pattern
```
jsync-swirlds: In my opinion, we really shouldn't have a "facilty plugin pattern".
Perhaps a better option might be to define an interface _module_ (blah-plugin-api) in addition to the implementation (blah-plugin) _if the plugin offers an API_, and that's up to other plugins to decide to add a `uses` clause for that interface, if they need to use that API.

docs/design/architecture/new-plugin-architecture.md
ata-nas: Was going to make this note since we have talked about this in the past. I think it is a good idea to describe generic notifications and their dynamic registration on startup. Anyone who needs a given plugin will also be able to see the notifications it also provides and be able to listen for them.

docs/design/architecture/new-plugin-architecture.md
ata-nas: The only benefit of this facility as of today is that it consolidates all available providers. So when we want to get a block, we will loop through all providers until we find what we are looking for. Same with availability, we consolidate a combined ranged set that is able to give us the needed information. There might be some benefit to keeping this, as it might become cumbersome to repeat this process for each plugin that needs blocks. Otherwise I agree that we can generally remove this.

docs/design/architecture/new-plugin-architecture.md
ata-nas: We should also define clearly what health plugin means. We currently have server health (facility) and also this plugin. If we unify those, how does the final product look like, what can it do? Certainly these endpoints are important for k8s, but then the plugins must also be able to report state so that the health plugin, based on configuration, can take certain actions. The data this provides is also necessary for server status I believe?

docs/design/architecture/new-plugin-architecture.md
ata-nas: I agree about application state facility. It should be something that is discoverable and used only by those who need it. We can achieve that via SPI on init. The state plugin should not depend on anyone.

docs/design/architecture/new-plugin-architecture.md
ata-nas: I agree with this. APIs specifically related to a certain plugin should be part of a specifically defined interface, but we should be very careful not to tangle the message bus and the way we use it with service API calls, as that couples things.

ata-nas (commented):
Docs look great! Left comments over what @jsync-swirlds has noted.

I share the opinion of @jsync-swirlds that we must completely get rid of the "facility" idea.

We have to generally be careful with service definition and not over-do APIs that can be called because that couples us, but we also tend to move away from the message bus we have defined and the way we use it.

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: The benefit of removing this is that instead of a global priority (no matter what requests the data, the ordering is the same) each plugin can decide what criteria it uses to order multiple sources.
That, I think, is worth considering.

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: Health plugin should, in my opinion, have responsibility for several things:
1. Receive updates from the core App and all plugins regarding current health status.
1. Determine, based on plugin-specific criteria, whether the full node is ready, live, and/or offline based on all available information.
1. Provide data to the status plugin for exposure on the status API
1. (optional) Provide the K8S `healthz` endpoints.
1. (optional) provide a `statusz` endpoint for monitoring tools that offers a detailed view of node health status.

This further supports diverse deployments as an operator might (for example) choose to deploy _without_ kunernetes, and also with a modified health plugin that does not provide Kubernetes endpoints.

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: I would prefer to not have any plugins in the SPI; we gain better deployment flexibility if the plugin provides its own API instead.
Example:
```
ApplicationState
   plugin-application-state-api
   plugin-application-state
HealthPlugin
   plugin-health-monitor-service
   plugin-health-monitor-api
   plugin-health-monitor
StatusPlugin
   plugin-status-monitor-service
   plugin-status-monitor-api
   plugin-status-monitor
```

API modules define interfaces for use by other plugins or the app
Service modules implement gRPC or ReST services
unadorned modules implement and _provide_ the interface defined in API.

Just a suggestion...

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: I agree, the message bus should be very carefully guarded.
We might choose to move toward more message-oriented communication (which is generally positive) over time, but adding a few carefully chosen direct APIs is reasonable for early releases.

docs/design/architecture/current-plugin-architecture.md
berryware: updated

docs/design/architecture/current-plugin-architecture.md
berryware: Updated

docs/design/architecture/current-plugin-architecture.md
berryware: Removed.

docs/design/architecture/new-plugin-architecture.md
berryware: updated

docs/design/architecture/new-plugin-architecture.md
berryware: Could we create an abstract class or class, eg, AbstractBlockNodeSink that could provide the default functionality of the HistoricalBlockFacility?
The stream subscriber plugin could extend this class and get the default functionality or override as necessary

docs/design/architecture/new-plugin-architecture.md
berryware: Captured this in the document. We can break this out later when we are ready to work on health.

docs/design/architecture/new-plugin-architecture.md
berryware: Updated

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: HistoricalBlockFacility is a block _source_, not a sink.

Basically, the blocks-files-recent and blocks-files-historic are _different plugins_ than all the rest and are loaded by this facility instead of the App they even implement a different plugin interface (which isn't ideal, I'd prefer they provide two; the baseline BlockNodePlugin, and an _additional_ BlockProviderPlugin).

What I'm advocating (as a design concept) is that we remove the facility entirely, and let any plugin that needs blocks can have `uses BlockProviderPlugin` to find block providers; and that interface will offer some additional detail _instead of_ the priority. Plugins that need to read blocks can then _independently_ choose which providers they query and the order they use based on that.

docs/design/architecture/current-plugin-architecture.md
```diff
@@ -0,0 +1,379 @@
...
+    STOR[Storage Plugin] -->|sendBlockPersisted| BN
+    BN --> SUB2[Subscriber Plugin]
+    BN --> PUB2[Publisher Plugin]
+    BN --> STOR[Storage Plugin]
+    BN --> N2[... Plugin N]
+```
+
+### Module Dependency Structure
+
+Plugins can be built internally, byt the community, or by private thirdParties.
```
jsync-swirlds: Minor typos
```suggestion
Plugins can be built internally, by the community, or by private third parties.
```

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: This item can be addressed in future updates.

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: Another item we'll probably need to address in future work.

jsync-swirlds (dismissed):

docs/design/architecture/new-plugin-architecture.md
berryware: Moved to Status Plugin

docs/design/architecture/new-plugin-architecture.md
berryware: These are moved out. Facility interfaces should be defined in their own module.

docs/design/architecture/new-plugin-architecture.md
berryware: Added that the LMX Disruptor may be changed at a future date

docs/design/architecture/new-plugin-architecture.md
berryware: Captured in: [#1426](https://github.com/hiero-ledger/hiero-block-node/issues/1426)

docs/design/architecture/new-plugin-architecture.md
berryware: My naming aside, I meant that maybe we can provide the aggregation, and sorting of the BlockProviderPlugins as an extendable class that Plugins that read blocks can extend. Basically making the current functionality available to them when we remove `HistoricalBlockFacility`

berryware: A ticket exists for this. [#2038](https://github.com/hiero-ledger/hiero-block-node/issues/2038)

docs/design/architecture/current-plugin-architecture.md
berryware: Corrected.

jsync-swirlds (dismissed):

jasperpotts:
I happened across this is daily status channel. I worry this adds complexity and want to push back on the proposed direction though, because I think a few properties of the original design that the new design quietly gives up are load-bearing, and the motivation as written may not justify the cost. 

**The single-threaded-per-handler property is doing more work than it looks like it is.**

The current design gives every `BlockItemHandler` and `BlockNotificationHandler` registration an actor-style invariant: each handler is invoked by exactly one thread, one event at a time. That falls out of how Disruptor sequence barriers work — it's structural, not conventional. Plugin authors get to write handler logic with no locks, no reentrancy concerns, no interleaving, and the mental model is essentially NodeJS-style: get a message, process it, get the next one. That's why the bus-routed code in Block Node reads as cleanly as it does.

Today the thread-safe-required bucket is small and well-known: `MetricRegistry`, `HealthFacility`, `ApplicationStateFacility`, `BlockMessagingFacility.sendBlockItems()`. Finite, framework-team-owned, reviewed for concurrency, all behind the SPI module.

Under the proposal, any plugin can `provides SomeInterface` and any other plugin can `uses` it and call it directly from whatever thread it happens to be on — a handler thread, a `start()`-spawned worker, a Helidon request thread, the state scanner thread. The provider's interface implementation is now in the thread-safe-required bucket whether the author realized it or not. The compiler won't tell them, JPMS won't tell them, and as far as I can see from the design doc, nothing tells them. They'll find out when a race shows up in production. A property that was *architecturally enforced* becomes *author-enforced*, and that's a real regression in the bug-prevention story that isn't called out in the goals.

This is the same concern @ata-nas raised about "moving away from the message bus" — same argument in two forms. The bus discipline isn't stylistic; it's what gives you the threading guarantee for free.

**JPMS `uses`/`provides` doesn't actually solve runtime ordering.**

JPMS resolves module readability at module-layer construction time. It guarantees a provider exists on the module path. It does *not* tell `BlockNodeApp` to call `init()` on a provider before its consumer, and `ServiceLoader.load(X)` returns providers in unspecified order. The current `configDataTypes → init → start` two-phase lifecycle is what gives ordering today: facilities load first, everything is guaranteed live by the time `start()` runs, and plugins never have to reason about "is my dependency ready yet."

The proposal pushes readiness and dependency-checking into each plugin (check deps in `init()`, expose a "ready" signal, mark self unhealthy if a dep is missing). If we keep two-phase init *and* add per-plugin readiness checks, we have two ordering mechanisms doing the same job, and the per-plugin one is the more error-prone of the two. The static-documentation value of `uses`/`provides` in `module-info.java` is real and worth keeping, but it's build-time hygiene, not a runtime ordering primitive.

**What's the concrete pain motivating this?**

Issue #2818 is framed as a capability — "I would like to be able to specify a plugin's dependency on other plugins" — but doesn't cite an existing plugin that the current model can't express cleanly. The original architecture was intentionally minimal, and that minimality has been a feature: there are two patterns plugin authors learn (message bus for data flow, BlockNodeContext for shared infra), and once you know them you know the system. Generalizing to "any plugin can declare a sync dependency on any other plugin" expands the surface area considerably, and it would help to anchor that on a specific case where the current SPI is blocking real work, vs. solving for hypothetical third-party plugins.

**A smaller change captures most of the legitimate benefit.**

The real awkwardness in the current design — and it is a real one — is that adding a new shared synchronous service requires modifying the `spi-plugins` module and the `BlockNodeContext` record. @ata-nas is right that the facility/plugin bifurcation is a wart. But you can fix that without the dependency-graph overhaul:

- Keep `BlockNodeContext` as the single injection surface.
- Have `BlockNodeApp` populate the context by `ServiceLoader`-loading each facility interface, instead of `BlockNodeContext` knowing the concrete impls.
- Keep the "facilities load before plugins" invariant — that *is* the dependency ordering, and it's a few lines of bootstrap.
- Keep the message bus as the path for inter-plugin coordination.

That gives third parties the ability to swap a facility implementation (the real extensibility win), removes the facility-vs-plugin duality at the SPI level, doesn't disturb the lifecycle, doesn't disturb the threading guarantee, and doesn't ask every plugin author to learn a new dependency-declaration pattern.

**Two questions I'd love to see addressed in the doc before this lands:**

1. What is the concrete current-day plugin (or near-term planned plugin) that the existing model can't express cleanly? If the answer is "future flexibility," I'd argue YAGNI is doing useful work in this codebase and we should be cautious about preemptive generality.
2. What's the threading contract for plugin-exposed interfaces under the new design? If it's "interface methods must be thread-safe; caller's thread invokes them," the doc should state that explicitly so authors know what they're signing up for. If there's a mechanism to preserve the actor-style guarantee for sync calls, I'd love to see it spelled out — though I suspect any such mechanism ends up looking a lot like "use the bus."

None of this is to say the current design is perfect — the SPI-module-must-change-for-new-facilities friction is real, and worth fixing. I'm just not yet convinced the proposed shape is the right fix, given the threading guarantee and messaging-discipline costs that come with it.

jsync-swirlds:
TL;DR:
It sounds like we need to work more on this document to make the various options and pathways clearer, and to incorporate answers to the concerns that @jasperpotts has raised so that we can be more certain that we have a strong design that retains the benefits of existing structure while also supporting the ground-truth realities we have encountered over the past several months.

---
A few notes (and I'll try to be brief).

### Architectural single-thread
We already have multi-threading and concurrency issues in all current plugins because the main messaging queue and the notification queue are separate threads, which means that _live published_ blocks and _backfilled_ blocks arrive on different threads.  Also, notification for successes and failures arrive on different threads and API requests arrive on different threads as well. This means Publisher, Verification, and Subscriber all deal with extensive concurrency issues already, and we have needed to add significant concurrency handling and deferred processing to these plugins in the past few months as a result.  The only place we see limited concurrency impacts _so far_ is the Persistence plugins, and even there we often need to launch our own threads because writing data can take a very long time and some necessary tasks require extensive deferred processing.

The single-thread-via-messaging was a great idea, but it didn't survive contact with testing and the addition of several requirements added during the end-to-end testing process.

### Facilities versus Plugins
The problem we're seeing with this is that we have plugins that depend on facilities that also depend on plugins, because we need to have data from both local disk and remote updates and all of this is creating a timing nightmare when combined with the extremely short startup time required for Block Nodes. We're hoping to take this all out of the critical path by allowing data-loading and plugin initialization to happen in parallel and manage the asynchronous load via independent data updates.
We also have the problem that the facilities were not originally implemented as plugins, they're (with one exception) all implemented entirely within the App module.
This makes facilities hard-bound to the App, difficult to modify or replace, and tightly coupled.
Facilities are also "special snowflakes" that only the Hiero team can create, which is less than ideal
Some existing facilities (e.g. BlockProvider) are also embedding assumptions in the architecture that turn out to not be reliably true, which is making it difficult to create the deployments we need for tier-1 use, and it's even worse for tier-2 use cases.
ApplicationState as a facility is driving much of this rework because it was turning out to be exceptionally difficult to manage small amounts of local data for multiple plugins without an N<sup>2</sup> complexity explosion. We believe this can be simple, but either we bake the interface into the application module (which means it's extremely difficult to change), or we need to make it something that is not bound to the application or SPI and accept less certainty versus future changes.

### Plugins declaring API
First, and most important, we do not want this to be the primary interaction, and the primary _block stream_ data should not flow this way.  The existing Messaging queue from Publisher to Verification and Subscriber (the only direct readers of the block _stream_ queue) and the Notification queue for all the other plugins should continue to be used.

We _could_ use messaging for all plugin-to-plugin interaction, and this might be better architecturally, but it's also a significant performance cost that seems to be excessive in certain use cases. Sending data to ApplicationState is the most immediately obvious; receiving data from ApplicationState might still be better as a Notification.

We need a solution for custom filter plugins, block ordering, failure handling, and plugins, such as state query, that provide external APIs not defined in the Block Node.  Some of these plugins need a mechanism to be called from other plugins in a direct-call manner (call with result) rather than the asynchronous fire-and-forget Notification approach.
**These should be rare**, but we want to have design and recommendation for them documented.

Acknowledging that plugins can declare their own APIs (we cannot prevent this in third-party plugins, even if we wanted to) and providing _guidance_ for that approach seems better, to me, than just ignoring the possibility.  Adding an example or two, where it makes our own plugins significantly better, seems a wise concurrent action.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,554 @@
...
+8. [Metrics](#metrics)
+9. [Exceptions](#exceptions)
+10. [Acceptance Tests](#acceptance-tests)
+
+## Purpose
+
+The Block Node is composed of many independently-developed functional components: block publishing, subscribing, verification, persistence, health checking, cloud archival, roster bootstrapping, and more. Rather than wiring these together statically at compile time, the Block Node uses a **plugin architecture** that discovers, configures, and manages the lifecycle of each component at runtime.
+
+This document describes a revised design of that architecture. The previous design ([plugin-architecture.md](plugin-architecture.md)) injected facilities into a shared `BlockNodeContext` record, which created an implicit coupling between the application bootstrap code and every facility type. This revision treats **facilities as plugins** — each facility implements `BlockNodePlugin`, exposes its own typed interface as a JPMS service, and is discovered by dependent plugins via `ServiceLoader` in their own `init()` method. The application bootstrap code becomes facility-agnostic. Dependency resolution and readiness coordination move into the plugins themselves.
```
Nana-EC:
```suggestion
This document describes a revised design of that architecture. The previous design ([plugin-architecture.md](./current-plugin-architecture.md)) injected facilities into a shared `BlockNodeContext` record, which created an implicit coupling between the application bootstrap code and every facility type. This revision treats **facilities as plugins** — each facility implements `BlockNodePlugin`, exposes its own typed interface as a JPMS service, and is discovered by dependent plugins via `ServiceLoader` in their own `init()` method. The application bootstrap code becomes facility-agnostic. Dependency resolution and readiness coordination move into the plugins themselves.
```

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,554 @@
...
+| `start()`                       | After all `init()` calls, on a virtual thread | Loads data and resources. Waits on required facility plugins via `isReady()` before proceeding. Starts background worker threads.                                                                                                                                         |
+| `stop()`                        | During graceful shutdown                      | Closes resources, stops worker threads.                                                                                                                                                                                                                                   |
+
+#### `ServerStatusPlugin`
+
+Owns and persists the mutable node state (`BlockNodeVersions`). These fields have been removed from `BlockNodeContext`. Plugins that need this state discover the facility via `ServiceLoader`                                         |
```
Nana-EC: To be clear is this how 1 plugin can call functions and details from another plugin?
Is this inter-service communication?

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,554 @@
...
+Passed to plugins during `init()`. Allows plugins to register routes without a direct dependency on Helidon.
+
+```java
+void registerHttpService(String path, HttpService... service);
+void registerGrpcService(ServiceInterface service);
+```
+
+### Facility Plugins
+
+Each facility is defined by its own interface in its own module. A facility plugin implements both `BlockNodePlugin` and its facility interface. The facility interface is registered as a JPMS service in its module info file.
```
Nana-EC: Is there a shared Facility concept or behaviour.
What differentiates a Facility plugin from a plugin that just implements interfaces.
Must the interface be centrally located somewhere?

jasperpotts:
Thanks @jsync-swirlds it is a hard balance to get right. I would be nice to keep things as simple as possible. We want to have cased like LiveState plugin provide API to other plugins to use. I wish there was a easy threading model but the trade off is having to switch threads at every plugin boundary which I worry would be expensive.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,554 @@
...
+State blockNodeState();
+void shutdown(String className, String reason);
+```
+
+The `HealthServicePlugin` registers `/healthz/livez` and `/healthz/readyz` HTTP endpoints that return `200 OK` while the state is `RUNNING` and `503 Service Unavailable` otherwise.
+
+The Health Facility has responsibility for the following:
+
+1. Receive updates from the core App and all plugins regarding current health status.
+2. Determine, based on plugin-specific criteria, whether the full node is ready, live, and/or offline based on all available information.
```
Nana-EC, line 139: Are we making consideration for scenarios where we want the BN container to be reachable but not in perfect state i.e. ready but not live?
Or is that yet to be defined?

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,554 @@
...
+boolean updateAddressBook(NodeAddressBook nodeAddressBook);
+
+// Change listener registration
+void registerTssDataListener(Consumer<TssData> listener);
+void registerAddressBookListener(Consumer<NodeAddressBook> listener);
+```
+
+Plugins register listeners in `init()` so they receive updates as soon as the facility becomes ready. The facility delivers the current value immediately upon listener registration if state is already available, and delivers each subsequent update as it occurs.
+
+### `BlockProviderPlugin` (interface)
```
Nana-EC, line 164: How does this reconcile with the new approach, is this out of place?
How does this compare to a facility?

Nana-EC (dismissed):
LG, just some questions for consideration

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: (opinion) Mostly correct.  A plugin that wishes to query the mutable node state (which is a lot more than the versions) can discover the status plugin via the service loader, and call a query method on the Status Plugin API.

The alternative would be for the Status Plugin to send notifications for every update, which might work well (and would better fit the design from early 2024), but which also adds more noise in the notification queue.

Either approach is acceptable, and we might change the primary/preferred path over time.

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: (opinion) We really shouldn't retain the "facility" concept.  It was a compromise from the start and is turning out to be more headache than help.

The only real differentiation in this redesign would be when the "facility" is loaded versus the "regular" plugins.  I'm not sure that's worth retaining, however, so we are probably better off just erasing that facility/plugin distinction.

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: BN container ready but not live means it's running and still loading data or otherwise preparing. Live means the BN is ready to serve requests.

If we want something more nuanced (or if we want reachable but not healthy) then we need to implement a different lifecycle and mechanism for reporting node health other than the Kubernetes health endpoints.

If/When we extend the Health Plugin beyond the current kubernetes-focused subset, we can consider these and other related questions as part of that expansion.

docs/design/architecture/new-plugin-architecture.md
jsync-swirlds: (opinion)
This interface is OK for now, the Historical Block Facility that currently `uses` this interface is problematic, but the interface itself is probably OK.

The one caveat there is that we want this to be a "mix-in" interface, not a replacement for `BlockNodePlugin`.

docs/design/architecture/new-plugin-architecture.md
berryware: I changed the definition of Facility to be a BlockNodePlugin that exposes its own unique interface. I can get more generic than that.

jsync-swirlds: My thought would be to remove the "facility" name.
A plugin is just a plugin, and a plugin that exposes additional interfaces is still just a plugin.

That's cleaner (a plugin is just a plugin), but does remove the differentiation in function or role...

docs/design/architecture/new-plugin-architecture.md
berryware: The "facility" pattern has been deprecated. We now have `BlockNodePlugin` components that can publish additional interfaces (optional) that other `BlockNodePlugin` components can consume.

docs/design/architecture/new-plugin-architecture.md
berryware: The "facility" pattern has been deprecated. We now have `BlockNodePlugin` components (plugins) that can publish additional interfaces (optional) that other plugins can consume.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,550 @@
...
+  <dt>ServiceBuilder</dt>
+  <dd>An interface passed alongside <code>BlockNodeContext</code> during <code>init()</code> that allows plugins to register HTTP and gRPC service routes on the Helidon web server.</dd>
+
+  <dt>Disruptor</dt>
+  <dd>The LMAX Disruptor ring buffer used by the <code>BlockMessagingFacility</code> plugin as the underlying transport for both block-item distribution and block-notification events between plugins. This messaging provider may be changed at a future date.</dd>
+
+  <dt>BlockProviderPlugin</dt>
+  <dd>A plugin that also implements the BlockSource interface and registers that interface as a JPMS service. Plugins that require a BlockSource will add `uses org.hiero.block.node.spi.BlockSource` to their module info file. There can be many BlockSources. It is up to the plugin to decide how to sort/choose the BlockSources they need to use.
+
+  <dt>isReady()</dt>
```
jsync-swirlds: One small suggestion (which may be addressed in follow-up):
Instead of `isReady`, perhaps name this `awaitReady` and assume the method will not return until it is ready. Still returns a boolean, however, because the method is permitted to return immediately and is not guaranteed to pause.

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,550 @@
...
+  <dd>A method on <code>BlockNodePlugin</code> that returns <code>true</code> when the plugin is fully initialized and ready to serve its callers. Plugins expose this so the `BlockNodeApp` and other plugins ca inspect the readiness of any given plugin.</dd>
```
jsync-swirlds:
```suggestion
  <dd>A method on <code>BlockNodePlugin</code> that returns <code>true</code> when the plugin is fully initialized and ready to serve its callers. Plugins expose this so the `BlockNodeApp` and other plugins can inspect the readiness of any given plugin.</dd>
```

docs/design/architecture/new-plugin-architecture.md
```diff
@@ -0,0 +1,550 @@
...
+| `start()`                       | After all `init()` calls, on a virtual thread | Loads data and resources. Waits on any required plugins via `isReady()` before proceeding. Starts background worker threads.                                                                                                                                     |
+| `stop()`                        | During graceful shutdown                      | Closes resources, stops worker threads.                                                                                                                                                                                                                          |
+
+#### `ServerStatusPlugin`
+
+Owns and persists the mutable node state (`BlockNodeVersions`). These fields have been removed from `BlockNodeContext`. Plugins that need this state discover the `ServerStatusPlugin` via `ServiceLoader`                                         |
```
jsync-swirlds: Leftover table marker:
```suggestion
Owns and persists the mutable node state (`BlockNodeVersions`). These fields have been removed from `BlockNodeContext`. Plugins that need this state discover the `ServerStatusPlugin` via `ServiceLoader`.
```

jsync-swirlds (dismissed):
I think we can merge this, and work to further improve in a separate PR.

The largest gap remaining is to ensure we make the preferred interactions via messaging clear and provide some guidelines for when to use direct interfaces and when to use messaging.

docs/design/architecture/new-plugin-architecture.md
berryware: Addressed

docs/design/architecture/new-plugin-architecture.md
berryware: Updated

docs/design/architecture/new-plugin-architecture.md
berryware: Updated

docs/design/architecture/new-plugin-architecture.md
berryware: Captured in #2236

jsync-swirlds (approved):

rockysingh (approved):

PR #2833, opened 2026-05-16
fix: Delay the actual closure of a connection after end stream

### LINKED ISSUE #2172: Finish Implementation for Blocking Scheduled Executor
### Story Form

As a Block Node Engineer
I want to have complete and working implementation for the blocking scheduled executor
So that I can use it properly in tests

### Technical Notes

- We have recently introduced an blocking scheduled executor.
- We have not fully implemented it.
- We must fully implement it, taking inspiration from the existing `BlockingExecutor`
  - The scheduled one should completely copy (or possibly internally use a `BlockingExecutor`)
  - The scheduled executor should add additional handling (if necessary) to be able to actually schedule tasks

### LINKED ISSUE #2805: Investigate EndStream behavior in Block Stream publish service
Currently when either a publisher or consumer is receiving an EndStream, there exists the chance that the message is not received and either side closes the connection immediately after sending the message. This has been observed in previewnet ({"from":"2026-05-13T19:19:06.933Z","to":"2026-05-13T19:24:46.783Z"}) where the BN tries to send a DUPLICATE_BLOCK message, immediately ends the stream, the CN side never receives the message but gets a socket closed exception and switches to another BN.

We may want to explore changing the protocol to have acknowledgments for these or on either side when sending an EndStream, expect the other side to close the connection, and if it's not closed within a timeout then close the connection.

jsync-swirlds:
## Reviewer Notes
* Add configuration for pause delay
* Add delay before shutdown
* Add pause support and pause the handler when a shutdown is scheduled

This is a quick fix to help reduce thrashing for publishers.  Further improvement and testing is required in follow-up.

## Related Issue(s)
Resolves #2805
Resolves #2172

### DIFF
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/TestConfigurationBuilder.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/TestConfigurationBuilder.java
@@ -1,0 +1,203 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.app.fixtures;
+
+import com.swirlds.config.api.Configuration;
+import com.swirlds.config.api.ConfigurationBuilder;
+import com.swirlds.config.api.converter.ConfigConverter;
+import com.swirlds.config.api.source.ConfigSource;
+import com.swirlds.config.api.validation.ConfigValidator;
+import com.swirlds.config.extensions.sources.SimpleConfigSource;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.locks.ReentrantLock;
+
+/// Helper for use in tests and to change the config for specific tests.
+/// Instance can be used per class or per test.
+public class TestConfigurationBuilder {
+    private final ReentrantLock configLock = new ReentrantLock();
+    private Configuration configuration = null;
+    private final ConfigurationBuilder builder;
+
+    /// Creates a new instance and automatically registers all config data records
+    /// (see [ConfigData]) on classpath/modulepath that are part of the packages
+    /// `com.swirlds` and `com.hedera`.
+    public TestConfigurationBuilder() {
+        this(true);
+    }
+
+    /// Creates a new instance and automatically registers all given config data
+    /// records. This call will not do a class scan for config data records on
+    /// classpath/modulepath like some of the other constructors do.
+    ///
+    /// @param dataTypes
+    public TestConfigurationBuilder(@Nullable final Class<? extends Record> dataTypes) {
+        this(false);
+        if (dataTypes != null) {
+            builder.withConfigDataTypes(dataTypes);
+        }
+    }
+
+    /// Creates a new instance and automatically registers all config data records
+    /// (see [ConfigData]) on classpath/modulepath that are part of the packages
+    /// `com.swirlds` and `com.hedera` if the `registerAllTypes` param is true.
+    ///
+    /// @param registerAllTypes if true all config data records on classpath will automatically be registered
+    public TestConfigurationBuilder(final boolean registerAllTypes) {
+        if (registerAllTypes) {
+            builder = ConfigurationBuilder.create().autoDiscoverExtensions();
+        } else {
+            builder = ConfigurationBuilder.create();
+        }
+    }
+
+    /// Sets the value for the config.
+    ///
+    /// @param propertyName name of the property
+    /// @param values list of values
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public TestConfigurationBuilder withValues(@NonNull final String propertyName, @Nullable final List<?> values) {
+        return withSource(new SimpleConfigSource(propertyName, values));
+    }
+
+    /// Sets the value for the config.
+    ///
+    /// @param propertyName name of the property
+    /// @param value the value
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public TestConfigurationBuilder withValue(@NonNull final String propertyName, @Nullable final String value) {
+        return withSource(new SimpleConfigSource(propertyName, value));
+    }
+
+    /// Sets the value for the config.
+    ///
+    /// @param propertyName name of the property
+    /// @param value the value
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public TestConfigurationBuilder withValue(@NonNull final String propertyName, final int value) {
+        return withSource(new SimpleConfigSource(propertyName, value));
+    }
+
+    /// Sets the value for the config.
+    ///
+    /// @param propertyName name of the property
+    /// @param value the value
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public TestConfigurationBuilder withValue(@NonNull final String propertyName, final double value) {
+        return withSource(new SimpleConfigSource(propertyName, value));
+    }
+
+    /// Sets the value for the config.
+    ///
+    /// @param propertyName name of the property
+    /// @param value the value
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public TestConfigurationBuilder withValue(@NonNull final String propertyName, final long value) {
+        return withSource(new SimpleConfigSource(propertyName, value));
+    }
+
+    /// Sets the value for the config.
+    ///
+    /// @param propertyName name of the property
+    /// @param value the value
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public TestConfigurationBuilder withValue(@NonNull final String propertyName, final boolean value) {
+        return withSource(new SimpleConfigSource(propertyName, value));
+    }
+
+    /// Sets the value for the config.
+    ///
+    /// @param propertyName name of the property
+    /// @param value the value
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public TestConfigurationBuilder withValue(@NonNull final String propertyName, @NonNull final Object value) {
+        Objects.requireNonNull(value, "value must not be null");
+        return withSource(new SimpleConfigSource(propertyName, value.toString()));
+    }
+
+    /// This method returns the [Configuration] instance. If the method is called
+    /// for the first time the [Configuration] instance will be created. All
+    /// values that have been set (see [#withValue(String, int)]) methods will
+    /// be part of the config. Next to this the config will support all config
+    /// data record types (see [ConfigData]) that are on the classpath.
+    ///
+    /// @return the created configuration
+    @NonNull
+    public Configuration getOrCreateConfig() {
+        configLock.lock();
+        try {
+            if (configuration == null) {
+                configuration = builder.build();
+            }
+            return configuration;
+        } finally {
+            configLock.unlock();
+        }
+    }
+
+    private void checkConfigState() {
+        configLock.lock();
+        try {
+            if (configuration != null) {
+                throw new IllegalStateException("Configuration already created!");
+            }
+        } finally {
+            configLock.unlock();
+        }
+    }
+
+    /// Adds the given config source to the builder
+    ///
+    /// @param configSource the config source that will be added
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public TestConfigurationBuilder withSource(@NonNull final ConfigSource configSource) {
+        checkConfigState();
+        builder.withSource(configSource);
+        return this;
+    }
+
+    /// Adds the given config converter to the builder
+    ///
+    /// @param converterType the type of the config converter
+    /// @param converter the config converter that will be added
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public <T> TestConfigurationBuilder withConverter(
+            @NonNull final Class<T> converterType, @NonNull final ConfigConverter<T> converter) {
+        checkConfigState();
+        builder.withConverter(converterType, converter);
+        return this;
+    }
+
+    /// Adds the given config validator to the builder
+    ///
+    /// @param validator the config validator that will be added
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public TestConfigurationBuilder withValidator(@NonNull final ConfigValidator validator) {
+        checkConfigState();
+        builder.withValidator(validator);
+        return this;
+    }
+
+    /// Adds the given config data type to the builder
+    ///
+    /// @param type the config data type that will be added
+    /// @param <T> the type of the config data type
+    /// @return the [TestConfigurationBuilder] instance (for fluent API)
+    @NonNull
+    public <T extends Record> TestConfigurationBuilder withConfigDataType(@NonNull final Class<T> type) {
+        checkConfigState();
+        builder.withConfigDataType(type);
+        return this;
+    }
+}
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
@@ -2,42 +2,359 @@
 package org.hiero.block.node.app.fixtures.async;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.lang.System.Logger;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Queue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.RunnableFuture;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Supplier;
 
 public class ScheduledBlockingExecutor extends ScheduledThreadPoolExecutor {
-    /** The work queue that will be used to hold the tasks. */
+    private final Logger LOGGER = System.getLogger(getClass().getCanonicalName());
+    private static final long TASK_TIMEOUT_MILLIS = 5_000L;
+    private static final long RUNNABLE_FUTURE_MAX_TIMEOUT_SECONDS = 60L;
+    /// The Map that will be used to hold the tasks.
+    private final ConcurrentLinkedQueue<Schedule> scheduleQueue;
+    /// Queue used to execute tasks
     private final BlockingQueue<Runnable> workQueue;
+    /// Counter to indicate total submitted tasks.
+    private int tasksSubmitted;
+    /// All async executors created when running tasks async
+    private final Queue<ExecutorService> asyncExecutors;
 
     public ScheduledBlockingExecutor(@NonNull final BlockingQueue<Runnable> workQueue) {
         super(1, Thread.ofVirtual().factory(), new AbortPolicy());
-
         this.workQueue = workQueue; // actual work queue
+        asyncExecutors = new ConcurrentLinkedQueue<>();
+        scheduleQueue = new ConcurrentLinkedQueue<>();
     }
 
     public ScheduledBlockingExecutor(int corePoolSize, @NonNull final BlockingQueue<Runnable> workQueue) {
         super(corePoolSize, Thread.ofVirtual().factory(), new AbortPolicy());
-
         this.workQueue = workQueue; // actual work queue
+        asyncExecutors = new ConcurrentLinkedQueue<>();
+        scheduleQueue = new ConcurrentLinkedQueue<>();
     }
 
-    /**
-     * Captures the task into the work queue without executing it.
-     * <p>
-     * We intentionally do NOT call {@code super.execute(command)} here. The
-     * purpose of this executor is to hold tasks until tests explicitly drain
-     * the queue, giving tests full control over execution ordering and timing.
-     * Calling the parent's execute would defeat that contract. Tests that need
-     * tasks to actually run should use a real {@link java.util.concurrent.ScheduledThreadPoolExecutor}.
-     */
+    @Override
+    public long getTaskCount() {
+        return workQueue.size() + scheduleQueue.size();
+    }
+
+    /// Overriding this method in order to make sure this logic will be called
+    /// every time we will submit a task to the pool.
     @Override
     @SuppressWarnings("all")
     public void execute(@NonNull final Runnable command) {
-        try {
-            workQueue.put(command);
-        } catch (final InterruptedException e) {
-            Thread.currentThread().interrupt();
-            throw new RuntimeException("Thread was interrupted while trying to put a task into the work queue", e);
+        if (command instanceof RunnableFuture<?> ft) {
+            // Wrap the FutureTask in a Runnable that will call get() on it
+            // otherwise when executing async below, we will not get an exception
+            // since the FutureTask will capture it. Super's submit() methods
+            // wrap the task as a FutureTask.
+            workQueue.offer(new FutureRunnable(ft));
+        } else {
+            workQueue.offer(command);
+        }
+        tasksSubmitted++;
+    }
+
+    /// Invoke this method once you have submitted all the tasks that need to be
+    /// run by this executor. This method will then proceed to poll each task
+    /// (in the same order that they have been submitted) and will execute the
+    /// tasks serially on the same thread that called this method. Once this
+    /// method returns (or throws), we are certain that all tasks have run and
+    /// can safely assert based on that. This method will throw an
+    /// [IllegalStateException] to indicate a broken state, when the queue
+    /// is empty.
+    public void executeSerially() {
+        moveScheduledToWorkQueue();
+        if (!workQueue.isEmpty()) {
+            while (!workQueue.isEmpty()) {
+                workQueue.poll().run();
+            }
+        }
+    }
+
+    /// This method executes all tasks that were submitted to this executor
+    /// asynchronously.
+    ///
+    /// Timeout is {@value TASK_TIMEOUT_MILLIS} milliseconds, blocking is enabled,
+    /// and exceptions will be thrown on exceptional completion.
+    /// Logging on exceptional completion is disabled.
+    /// @see #executeAsync(boolean, long, boolean, boolean, Supplier)
+    public List<CompletableFuture<Void>> executeAsync() {
+        return executeAsync(true);
+    }
+
+    /// This method executes all tasks that were submitted to this executor
+    /// asynchronously.
+    ///
+    /// Timeout is {@value TASK_TIMEOUT_MILLIS} milliseconds and blocking is
+    /// enabled. Logging on exceptional completion is disabled.
+    /// @see #executeAsync(boolean, long, boolean, boolean, Supplier)
+    public List<CompletableFuture<Void>> executeAsync(final boolean throwOnExceptionalCompletion) {
+        return executeAsync(TASK_TIMEOUT_MILLIS, throwOnExceptionalCompletion);
+    }
+
+    /// This method executes all tasks that were submitted to this executor
+    /// asynchronously.
+    ///
+    /// Blocking is enabled and exceptions will be thrown on exceptional
+    /// completion. Logging on exceptional completion is disabled.
+    /// @see #executeAsync(boolean, long, boolean, boolean, Supplier)
+    public List<CompletableFuture<Void>> executeAsync(final long timeoutMillis) {
+        return executeAsync(timeoutMillis, true);
+    }
+
+    /// This method executes all tasks that were submitted to this executor
+    /// asynchronously.
+    ///
+    /// Blocking is enabled. Logging on exceptional completion is disabled.
+    /// @see #executeAsync(boolean, long, boolean, boolean, Supplier)
+    public List<CompletableFuture<Void>> executeAsync(
+            final long blockTimeoutMillis, final boolean throwOnExceptionalCompletion) {
+        return executeAsync(blockTimeoutMillis, throwOnExceptionalCompletion, false);
+    }
+
+    /// This method executes all tasks that were submitted to this executor
+    /// asynchronously.
+    ///
+    /// Blocking is enabled.
+    /// @see #executeAsync(boolean, long, boolean, boolean, Supplier)
+    public List<CompletableFuture<Void>> executeAsync(
+            final long blockTimeoutMillis,
+            final boolean throwOnExceptionalCompletion,
+            final boolean logOnExceptionalCompletion) {
+        return executeAsync(
+                true,
+                blockTimeoutMillis,
+                throwOnExceptionalCompletion,
+                logOnExceptionalCompletion,
+                () -> Executors.newThreadPerTaskExecutor(Executors.defaultThreadFactory()));
+    }
+
+    /// This method executes all tasks that were submitted to this executor
+    /// asynchronously.
+    ///
+    /// Default pool implementation used.
+    /// @see #executeAsync(boolean, long, boolean, boolean, Supplier)
+    public List<CompletableFuture<Void>> executeAsync(
+            final long blockTimeoutMillis,
+            final boolean blockUntilDone,
+            final boolean throwOnExceptionalCompletion,
+            final boolean logOnExceptionalCompletion) {
+        return executeAsync(
+                blockUntilDone,
+                blockTimeoutMillis,
+                throwOnExceptionalCompletion,
+                logOnExceptionalCompletion,
+                () -> Executors.newThreadPerTaskExecutor(Executors.defaultThreadFactory()));
+    }
+
+    /// This method executes all tasks that were submitted to this executor
+    /// asynchronously.
+    ///
+    /// Internally, a [java.util.concurrent.ThreadPerTaskExecutor] is used
+    /// to submit each task. All tasks are started in the order they were
+    /// submitted to the executor. The way tasks are submitted is by
+    /// calling [CompletableFuture#runAsync(Runnable, java.util.concurrent.Executor)]
+    /// on each task in the internal worker queue. Depending on the parameters
+    /// described below, this method will either block or not, throw an
+    /// exception or not, log an exceptional execution or not, and will use a
+    /// timeout for each task or not.
+    ///
+    /// @param blockUntilDone boolean value indicating if the method should
+    /// block until all tasks are done or not. If true, the method will
+    /// call [CompletableFuture#join()] on each submitted task.
+    /// @param blockTimeoutMillis long value indicating the timeout in
+    /// milliseconds for each task to complete. Internally,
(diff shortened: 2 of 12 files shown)


jsync-swirlds:
This change is not well tested.
We should run some testing with this change on a branch to verify that it improves the required behavior.

ata-nas:
It appears the delay interferes with a lot of tests. I wonder, do we need to actually schedule a task, or we can simply briefly wait and then shutdown on the caller thread instead of scheduling that. Also, there seems to be a possibility for multiple tasks to be scheduled. I wonder, can't we simply `parkNanos` on the shutdown method? We will also be able to get rid of the paused variable.

jsync-swirlds:
Ugh, Claude created a worktree, tried to "help" and mangled a few things, then wiped the worktree (which removed changes to 3 files, fortunately I had a separate commit with the changes), and I missed it.
I'll recover those changes and re-push.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -313,6 +313,16 @@ public void signalDataReady() {
         }
     }
 
+    @Override
+    public ScheduledFuture<?> scheduleAfterDelay(final Runnable taskToSchedule, final long delayInNanoseconds) {
+        return scheduledExecutor.schedule(taskToSchedule, delayInNanoseconds, TimeUnit.NANOSECONDS);
```
ata-nas, line 322: Note, the scheduled executor has a pool size of 1. We need more I believe. Until now, we used it to schedule the inactivity timeout, but now we will have more need of threads.

We should also consider to have a thred-per-task pool, maybe that is a better option. We do have to add that to the pool manager.

Another thing we can do (_future_) is to simply be able to use the pool manager to schedule a task directly and it will give us an identifier back. We can use the identifier to cancel the task or do other things to it and in this way we allow the manager to completely handle the task, it's execution in some internal managed pool and all operations related to the task. Sort of like a "schedule me this task" API with the ability to stop.

_For now_, we can either increase the size of a fixed size pool or implement in the thread pool manager a simple thread-per-task virtual threads pool to use here.

Up to you.

Blocking so we do not miss this.

ata-nas (changes requested):

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -2586,14 +2606,16 @@ void testMultipleStalledHandlersAllTerminated() {
...
-                for (long block = 2L; block <= 5L; block++) {
+                for (long block = 2L; block <= 6L; block++) {
                     final PublishStreamRequestUnparsed req = PublishStreamRequestUnparsed.newBuilder()
                             .blockItems(TestBlockBuilder.generateBlockWithNumber(block)
                                     .asItemSetUnparsed())
                             .build();
                     publisherHandler3.onNext(req);
                     endThisBlock(publisherHandler3, block);
                 }
+                assertThat(threadPoolManager.scheduledExecutor().getTaskCount()).isEqualTo(2);
```
ata-nas: Note: this line is incorrect. Whenever we start a manager, an inactivity timeout task is scheduled at the beginning. It will be cancelled when the first publisher connects. So we are seeing 1 task for that. We must see 3 tasks if we have submitted delayed shutdowns for 2 publishers.

Something to think about, we can consider to run the inactivity timeout task prior to making this check. But note that the inactivity timeout task is present on _all_ tests!

Now, the reason this test fails to see the 3rd task is because of a conflict with the key. Simply, the first delayed shutdown task uses a `Schedule` for key in the `workMap` of the blocking scheduled executor. Then, when the second one comes, it uses a `Schedule` key with the same data. It will then overwrite the existing map entry and we end up losing the first one. The serial execution of the tasks is, I believe in the correct place. Once the scheduled executor is updated, we should see this test passing.

ata-nas:
@jsync-swirlds I've managed to push fixes for all failing tests here: 62ffaf2a4c9d215f72d471bdc5947011363fd376

The only test that still fails is:
`org.hiero.block.node.stream.publisher.LiveStreamPublisherManagerTest.FunctionalityTests.StallDetectionTests#testMultipleStalledHandlersAllTerminated`
but that is due to a fix needed in the scheduled executor (see [this comment](https://github.com/hiero-ledger/hiero-block-node/pull/2833#discussion_r3272241047)). It should pass automatically once a fix there is introduced.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
jsync-swirlds: Probably the use of a map was incorrect, it just seemed the simplest mechanism to use, but it is turning out to be likely incorrect.
I will work on this today, I think I'll change `Schedule` to include the task, and store that by itself in a queue rather than using a Map.

ata-nas: This will also comply with the javadoc, which says that tasks will be submitted in the order received.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Scheduling tasks offers a Future back (though it's not completely wired in yet, cancel _should_ work on that Future), but I agree that further improvements will likely be needed in the future.

I'll look at the pool size issue, we don't need more threads to handle more tasks (we have effectively unlimited queue depth), but we might want more _in parallel_.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/TestConfigurationBuilder.java
```diff
@@ -0,0 +1,236 @@
...
+     * {@link Configuration} instance will be created. All values that have been set (see
+     * {@link #withValue(String, int)}) methods will be part of the config. Next to this the config will support all
+     * config data record types (see {@link ConfigData}) that are on the classpath.
+     *
+     * @return the created configuration
+     */
+    @NonNull
+    public Configuration getOrCreateConfig() {
+        try {
+            configLock.lock();
```
ata-nas: I think the calling of the `lock` method should be outside of the try-catch. If that throws, we do not want to execute the finally method because of a silenced exception.
Probably not so critical for test fixtures.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: We could use the future that is returned, but it might be good to also leave that to the thread pool if we want to let it completely manage everything. Not convicted to any of those approaches.

For the pool size, yes, we simply need a bit more parallel runs to be able to happen. Yes, the queue is unbounded for receiving tasks.

ata-nas (commented):
Tests now seem fixed.

ata-nas (dismissed):

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/TestConfigurationBuilder.java
jsync-swirlds: Good catch, should have been just above the try.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I added dynamic sizing with a minimum of 2.

ata-nas (dismissed):

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
```diff
@@ -2,42 +2,359 @@
...
+                        // Log if configured to do so
+                        LOGGER.log(Logger.Level.ERROR, message, e);
+                    }
+                }
+            }
+            return futures;
+        }
+    }
+
+    /// Move all current schduled tasks to the work queue for execution.
```
Nana-EC:
```suggestion
    /// Move all current scheduled tasks to the work queue for execution.
```

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
```diff
@@ -2,42 +2,359 @@
...
+                // run the task
+                innerRunnable.run();
+                // get, but with a timeout to ensure we do not block forever
+                innerRunnable.get(RUNNABLE_FUTURE_MAX_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+            } catch (final InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
+            } catch (final ExecutionException e) {
+                // this is expected to happen, if we throw here, this means
+                // that the task has throw an exception during execution.
```
Nana-EC:
```suggestion
                // that the task has thrown an exception during execution.
```

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/TestConfigurationBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+import com.swirlds.config.extensions.sources.SimpleConfigSource;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.locks.ReentrantLock;
+
+/// Helper for use in tests and to change the config for specific tests.
+/// Instance can be used per class or per test.
+public class TestConfigurationBuilder {
```
Nana-EC, line 18: This is just a dupe of CN. [TestConfigBuilder](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/platform-sdk/swirlds-config-extensions/src/testFixtures/java/com/swirlds/config/extensions/test/fixtures/TestConfigBuilder.java) right?
Not a fan of more code to manage CN specific way of development.
We should be a able to use the ConfigBuilder from the library easily and without constant overhead.
For Plugin specific tests why is this needed when the plugin supports passing in Context with desired configs?

block-node/roster-bootstrap-tss/src/test/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPluginTest.java
```diff
@@ -96,18 +96,10 @@ void requestTssDataFromPeerBN() throws IOException, InterruptedException {
...
-        RosterBootstrapTssPlugin plugin = new RosterBootstrapTssPlugin() {
-
-            @Override
-            public void onContextUpdate(BlockNodeContext context) {
-                contextUpdated[0]++;
-                tssData[0] = context.tssData();
-                latch.countDown();
-            }
-        };
+        RosterBootstrapTssPlugin plugin = new TestBootstrapPlugin(contextUpdated, tssData, latch);
```
Nana-EC, line 99: Q: Why was this change needed?
I don't see additional usage or differences in the tests

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -131,8 +131,10 @@ public LiveStreamPublisherManager(
         dataReadyLatch = dataReadyLock.newCondition();
         NodeConfig nodeConfiguration = serverContext.configuration().getConfigData(NodeConfig.class);
         earliestManagedBlock = nodeConfiguration.earliestManagedBlock();
-        scheduledExecutor =
-                threadManager.createVirtualThreadScheduledExecutor(1, null, this::uncaughtScheduledExecutorException);
+        int threadCount = Runtime.getRuntime().availableProcessors();
+        threadCount = threadCount < 2 ? 2 : threadCount;
```
Nana-EC, line 137: Nice catch, worth a comment to highlight the need for at least 2 threads

Nana-EC (commented):
Initial review, just got to block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherManager.java, will circle back

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/TestConfigurationBuilder.java
jsync-swirlds: Using a test configuration builder is easier than a real configuration builder sometimes in unit tests, it's mostly useful when _creating_ the Configuration to be assigned to the context.

I added it because I find it useful at times, but we can certainly discuss with the team and decide not to have it in fixtures (I'll just have to put a subset of this code in the unit tests that actually need it).

block-node/roster-bootstrap-tss/src/test/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPluginTest.java
jsync-swirlds: Here, it just cleaned up the code so I could reason properly while working out how to fix the test.
It's not strictly necessary, but it's a lot easier to read.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I wish I could take the credit, but @ata-nas noticed this and commented above.

I'll add the comment.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
jsync-swirlds: Fixed.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
jsync-swirlds: Fixed.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Fixed.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
```diff
@@ -25,6 +25,7 @@ public record PublisherConfig(
         @Loggable @ConfigProperty(defaultValue = "9_223_372_036_854_775_807") @Min(100_000L) long batchForwardLimit,
         @Loggable @ConfigProperty(defaultValue = "300") @Min(0L) long publisherUnavailabilityTimeout,
         @Loggable @ConfigProperty(defaultValue = "3") @Min(3) @Max(50) int MaxFutureBlocksBeforeStalled,
-        @Loggable @ConfigProperty(defaultValue = "100") @Min(0L) @Max(200L) long staleResendPruneBuffer) {
+        @Loggable @ConfigProperty(defaultValue = "100") @Min(0L) @Max(200L) long staleResendPruneBuffer,
+        @Loggable @ConfigProperty(defaultValue = "100_000_000") @Min(100_000L) @Max(5_000_000_000L) long flowControlPauseDelay) {
```
Nana-EC: The Min has to be at least the value of the hardcoded SHUTDOWN_DELAY_TIME. Otherwise the LockSupport.parkNanos will complete before the SHUTDOWN_DELAY_TIME.
Either that or you need a `pauseDelay > shutdownDelay` check somewhere.

Nana-EC: nit: the unit is unclear here. Either say nanos in the field name or variable name

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -305,7 +317,10 @@ private void checkMidBlockAndShutdown(final long blockNumber) {
                 publisherManager.blockIsEnding(currentStreamingBlockNumber.get(), handlerId);
             }
         } finally {
-            shutdown();
+            // pause the handler while we delay shutdown
+            isPaused.set(true);
+            // Do not shutdown immediately, allow for final messages to send first.
+            publisherManager.scheduleAfterDelay(this::shutdown, SHUTDOWN_DELAY_TIME);
```
Nana-EC, line 323: The result of this is that resetState() will run before shutdown.
Might this be problematic. Is anything in shutdown reading state and do we have to this about preventing future additions to shutdown that might read state and make the wrong conclusions?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -180,6 +188,10 @@ public void onSubscribe(@NonNull final Subscription subscription) {
 
     @Override
     public void onNext(@NonNull final PublishStreamRequestUnparsed request) {
+        if (isPaused.compareAndSet(true, false)) {
+            // If we need to pause, park for the configured delay.
+            LockSupport.parkNanos(configuration.flowControlPauseDelay());
```
Nana-EC: nit: FYI name could be misleading, it gives the impression is managing flowControl at the gRPC level.

Nana-EC (commented):
Made it through `PublishHandler`.
There's some open items than need validation to ensure we're not opening a set of bugs.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
jsync-swirlds: The two do not need to be correlated.  Flow Control pause delay is pausing _incoming_ messages only (it's basically locking the Helidon inbound thread, not ideal but PBJ left no other option).
We delay shutdown to give CN a chance to receive a message, we pause the incoming thread for _inbound flow control_ and need that to be configurable across a fairly wide range because we don't know in advance what is a good value.

Edit:
I see the confusion, the pause is not intended to lock the thread for the entire duration, after the thread lock expires, the onNext will continue and reach the isActive check, which is't the long-term intent, but _is_ the correct behavior at the moment.
It's worth noting that this is a _very partial_ pause implementation, it's not complete and I had to put some limits in (only one pause) to prevent permanent deadlock because I don't have the _rest_ of the flow control implementation yet.
This works for now, provided we have reasonably sane settings, which the defaults are.
In future PRs (for Issue 1931) I'll extend this and add the remaining bits (including repeated pause if flow control still needs to pause inbound messaging) that are needed to cover every corner case.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: There is nothing worrisome about the order here.
Reset State affects only incoming requests (which is why we pause inbound data) and does not interact with shutdown.
Even if we had an inbound request arrive, the reset is what we do before accepting a new header anyway, so it still leaves the Handler in a reasonable state.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: That's exactly what it's doing.  We just don't have direct control there so we lock the Helidon calling thread because _that's the only option PBJ gave us_ to pause HTTP flow control (gRPC doesn't have flow control; it relies on HTTP for that).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
jsync-swirlds: Fixed.

Nana-EC (approved):

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
AlfredoG87, line 315: **[BLOCKING]**

Without a guard, every call to checkMidBlockAndShutdown schedules another shutdown() and re-arms isPaused. Twelve call sites can reach this method, so post-EndStream traffic produces duplicate PUBLISHER_DISCONNECTED notifications, double-counted currentPublisherCount updates, and replies.onComplete() called more than once
```suggestion
if (!isActive.compareAndSet(true, false)) {
    // Another path has already initiated shutdown for this handler.
    return;
}

try {
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -80,6 +81,7 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
             "metric-end-to-end-latency-by-block-end block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId={3}";
     private static final String LATENCY_START_MESSAGE =
             "metric-end-to-end-latency-by-block-start block={0,number,#} nsTimestamp={1,number,#} handlerId={2} correlationId={3}";
+    private static final long SHUTDOWN_DELAY_TIME = 10_000_000; // 10 milliseconds
```
AlfredoG87, line 84: Suggested to rename, also why not configurable?
Also I would argue that `10ms` is too little, for RTT on high congestion and/or high latency networks might need more.  I would recommend a default of `150ms`, at the end of the day, we are not accepting any more messages cause the isPaused and isActive guards, and if the client closes it first, there is no harm (is actually expected). and the config, I would set a min of 50 and a max of 500ms

```suggestion
    private static final long SHUTDOWN_DELAY_NANOS = 10_000_000L; // 10 milliseconds
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -131,8 +131,12 @@ public LiveStreamPublisherManager(
...
         NodeConfig nodeConfiguration = serverContext.configuration().getConfigData(NodeConfig.class);
         earliestManagedBlock = nodeConfiguration.earliestManagedBlock();
-        scheduledExecutor =
-                threadManager.createVirtualThreadScheduledExecutor(1, null, this::uncaughtScheduledExecutorException);
+        int threadCount = Runtime.getRuntime().availableProcessors();
+        // Ensure at least 2 threads for scheduled tasks so we do not have
+        // excessive contention with the idle detector task.
+        threadCount = threadCount < 2 ? 2 : threadCount;
+        scheduledExecutor = threadManager.createVirtualThreadScheduledExecutor(
+                threadCount, null, this::uncaughtScheduledExecutorException);
```
AlfredoG87, line 139: is not excessive to cap threadCount at available processors? at most we need 4.
for a 64 Cores CPU, we will have a waste of threads in the pool, even if they are virtual, is still a waste.

AlfredoG87 (changes requested):
**There is just one blocking request**. Otherwise looks good.

 The PR feels more complex than the problem needs. Atanas's suggestion — inline parkNanos then shutdown() — closes #2805 in a few lines. The extra machinery is groundwork for flow control, but I think that belongs in its own PR; reviewers can't really evaluate scaffolding for a feature that isn't here yet.

There's a cleaner design too. PublisherHandler.clientEndStreamReceived() already fires when CN closes its half of the HTTP/2 stream, and I believe CN closes the stream on receiving EndStream (worth confirming in consensus-node). If so, the fix is purely BN-side:

  - Send EndStream
  - Schedule fallback shutdown() at T ms
  - On clientEndStreamReceived() → cancel the future, run shutdown() immediately
  - On timeout → shutdown() anyway

Same scheduled-executor this PR already adds, just driven by an actual signal. No parkNanos, no Helidon thread parked during the wait.

10 ms is also too tight. That's best-case same-DC RTT, with no margin at all for cross-region setups where one-way latency alone is 50-150 ms — a healthy peer can legitimately get cut off before EndStream arrives. With await-or-timeout, the value is just a safety net for misbehaving CNs; 500 ms or 1 s costs nothing in the happy path.

The isPaused machinery is also a workaround that should not be nedded, adds complexity that does not belong here. The real long-term fix for flow control is PBJ honoring subscription.request(0) (Helidon already exposes the inbound HTTP/2 hook; PBJ's request(long) is literally return today). Worth opening against PBJ.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: There is no need to configure this, and it cannot be too short (within reason).
The actual disconnect still must flow across the wire, which means it takes roughly as long to propagate as the end stream message.

All this delay does is give the _client_ time to process the message. It does not need to handle network delays; only client processing; and for that 10ms ample.

Given it's a shutdown situation, it doesn't harm anything to wait 10ms, and there's no reasonable need to make it configurable.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: This is _half_ the processor count, and they're virtual threads, so it's extremely low cost to keep extras, and it is also very difficult to predict in advance how many might be useful across multiple handlers (it could easily be 10 or more).

jsync-swirlds:
> **There is just one blocking request**. Otherwise looks good.
> 
> The PR feels more complex than the problem needs. Atanas's suggestion — inline parkNanos then shutdown() — closes #2805 in a few lines. The extra machinery is groundwork for flow control, but I think that belongs in its own PR; reviewers can't really evaluate scaffolding for a feature that isn't here yet.
> 
> There's a cleaner design too. PublisherHandler.clientEndStreamReceived() already fires when CN closes its half of the HTTP/2 stream, and I believe CN closes the stream on receiving EndStream (worth confirming in consensus-node). If so, the fix is purely BN-side:
> 
> * Send EndStream
> * Schedule fallback shutdown() at T ms
> * On clientEndStreamReceived() → cancel the future, run shutdown() immediately
> * On timeout → shutdown() anyway
> 
> Same scheduled-executor this PR already adds, just driven by an actual signal. No parkNanos, no Helidon thread parked during the wait.
> 
> 10 ms is also too tight. That's best-case same-DC RTT, with no margin at all for cross-region setups where one-way latency alone is 50-150 ms — a healthy peer can legitimately get cut off before EndStream arrives. With await-or-timeout, the value is just a safety net for misbehaving CNs; 500 ms or 1 s costs nothing in the happy path.
> 
> The isPaused machinery is also a workaround that should not be nedded, adds complexity that does not belong here. The real long-term fix for flow control is PBJ honoring subscription.request(0) (Helidon already exposes the inbound HTTP/2 hook; PBJ's request(long) is literally return today). Worth opening against PBJ.

I disagree on several points.  This solution is not overcomplicated and is the simplest that actually works.
1. subscription.request does absolutely nothing.  PBJ does not implement flow control in that manner.  The PBJ plugin team's official position is that the only "correct" mechanism for flow control is blocking the `onNext` thread.
   - I already wasted considerable hours implementing flow control based on `Subscription` only to find it did nothing whatsoever.
   - We already opened an issue against PBJ, the response is not positive.
   - I need _something_ to block incoming messages during the pause; this is the smallest additional change that works for now. Adding proper flow control is coming soon and involves quite a lot more complexity due to the limitations in PBJ.
1. 10ms is ample, because we do not need to consider network latency (the end stream and shutdown will experience roughly the same latency), we only need to consider client processing time. Client processing within 10ms is extremely reasonable, generous even, but not so long as to be worth fine-tuning either.
1. If we block the thread calling shutdown, we will be blocking the Notification thread (among others), and that blocks sending messages to other handlers.  We cannot use a simple timeout (I tested that, it locks up the messaging unless I add even more complex code).
1. Adding a cancel in `clientEndStreamReceived` is an option.  It is _highly unreliable_ (we generally do not receive end streams for the same reason the client wasn't receiving ours, they close the connection too fast); but we can consider that in a future PR.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: 1. Rearming isPaused is fine; it won't finish until after the shutdown regardless, so the Handler will already be closed. (there is more to do there, but it belongs in the flow control PR, which is in progress).
1. There is a reasonable concern around multiple shutdown calls; though it is a very unlikely setup. It also will not break anything; we'll have more than one attempt to shutdown, but all after the first will just throw when calling onComplete and that is caught cleanly.

If you do not mind, I can add this check (or something similar) in a follow-up to sooner unblock current testing efforts.

AlfredoG87 (approved):
Unblocking due to strongs reasons presented.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
AlfredoG87: The snippet shows Runtime.getRuntime().availableProcessors(), no /2 — so on a 64-core host this is a 64-thread pool, which is what I was flagging.
  
  I get the "virtual threads are cheap" point, but the workload doesn't really benefit from extras. Each delayed shutdown is sub-millisecond (removeHandler + onComplete + closeConnection), so even under a mass-disconnect burst the queue drains way faster than tasks arrive. Parallelism only helps if a task blocks on slow socket close, and a handful of workers covers that. 4 (or even fixed at 2) is plenty for what we know about the workload; sizing it to the host's CPU count doesn't  buy anything concrete.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
AlfredoG87: Fair point about TCP ordering — EndStream and the close travel in order on the same connection, so the client receives EndStream first regardless of RTT. The 10 ms is local flush + client callback dispatch, not network propagation. That weakens the cross-region concern I had.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
AlfredoG87: shutdown() calls removeHandler before replies.onComplete(), and removeHandler always fires sendPublisherStatusUpdate(PUBLISHER_DISCONNECTED, ...) and updates currentPublisherCount. So duplicate  shutdowns produce duplicate disconnect notifications and metric updates, not just swallowed exceptions.
  
That said, fair that the setup is unlikely in practice. Happy with a follow-up if it gets tracked as an issue before merge.

AlfredoG87:
> > **There is just one blocking request**. Otherwise looks good.
> > The PR feels more complex than the problem needs. Atanas's suggestion — inline parkNanos then shutdown() — closes #2805 in a few lines. The extra machinery is groundwork for flow control, but I think that belongs in its own PR; reviewers can't really evaluate scaffolding for a feature that isn't here yet.
> > There's a cleaner design too. PublisherHandler.clientEndStreamReceived() already fires when CN closes its half of the HTTP/2 stream, and I believe CN closes the stream on receiving EndStream (worth confirming in consensus-node). If so, the fix is purely BN-side:
> > 
> > * Send EndStream
> > * Schedule fallback shutdown() at T ms
> > * On clientEndStreamReceived() → cancel the future, run shutdown() immediately
> > * On timeout → shutdown() anyway
> > 
> > Same scheduled-executor this PR already adds, just driven by an actual signal. No parkNanos, no Helidon thread parked during the wait.
> > 10 ms is also too tight. That's best-case same-DC RTT, with no margin at all for cross-region setups where one-way latency alone is 50-150 ms — a healthy peer can legitimately get cut off before EndStream arrives. With await-or-timeout, the value is just a safety net for misbehaving CNs; 500 ms or 1 s costs nothing in the happy path.
> > The isPaused machinery is also a workaround that should not be nedded, adds complexity that does not belong here. The real long-term fix for flow control is PBJ honoring subscription.request(0) (Helidon already exposes the inbound HTTP/2 hook; PBJ's request(long) is literally return today). Worth opening against PBJ.
> 
> I disagree on several points. This solution is not overcomplicated and is the simplest that actually works.
> 
> 1. subscription.request does absolutely nothing.  PBJ does not implement flow control in that manner.  The PBJ plugin team's official position is that the only "correct" mechanism for flow control is blocking the `onNext` thread.
>    
>    * I already wasted considerable hours implementing flow control based on `Subscription` only to find it did nothing whatsoever.
>    * We already opened an issue against PBJ, the response is not positive.
>    * I need _something_ to block incoming messages during the pause; this is the smallest additional change that works for now. Adding proper flow control is coming soon and involves quite a lot more complexity due to the limitations in PBJ.
> 2. 10ms is ample, because we do not need to consider network latency (the end stream and shutdown will experience roughly the same latency), we only need to consider client processing time. Client processing within 10ms is extremely reasonable, generous even, but not so long as to be worth fine-tuning either.
> 3. If we block the thread calling shutdown, we will be blocking the Notification thread (among others), and that blocks sending messages to other handlers.  We cannot use a simple timeout (I tested that, it locks up the messaging unless I add even more complex code).
> 4. Adding a cancel in `clientEndStreamReceived` is an option.  It is _highly unreliable_ (we generally do not receive end streams for the same reason the client wasn't receiving ours, they close the connection too fast); but we can consider that in a future PR.



Fair pushback. If the team's position is "block onNext is the correct mechanism" and that's not negotiable, then isPaused/parkNanos is the only thing that works. Didn't realise the upstream had already been pursued and rejected.
  
The blocking-the-Notification-thread point is also fair, inline parkNanos in checkMidBlockAndShutdown would cascade through whatever thread calls in there. The scheduled approach is load-bearing, not just groundwork.

And on clientEndStreamReceived being unreliable for the same reason the original bug exists. If CN closes too fast to fire the callback, await-or-timeout collapses to just-timeout most of the time, which is what's already in place.

That leaves the blocking item below as the only real ask. Everything else, you've answered. But we can improve on a follow-up.

Unblocking ✅

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
AlfredoG87: Not a blocker, is just a nit, but maybe we can consider on a follow-up to cap it at 16 threads.?

timfn-hg: I'm just passing through, but virtual threads should _never_ be pooled.

> [Virtual threads should never be pooled](https://openjdk.org/jeps/444#Do-not-pool-virtual-threads) since each is intended to run only a single task over its lifetime.

jsync-swirlds: They aren't, the virtual thread "pool" size is really just a count of threads permitted to run simultaneously in this executor service. We're using the JVM executor services behind these (in production).

PR #2834, opened 2026-05-17
feat(tool): networkCapacity Tool Refactor + Production Defaults

AlfredoG87:
Refactors the `networkCapacity` tool to produce accurate CN→BN throughput measurements and adds gRPC compression support.

## Problem

The old tool had two measurement artifacts:
1. Server used `publishBlockStream()` — fully deserializes every `BlockItem`, which production BN never does.
2. Client parsed and batched blocks *during* the gRPC stream, mixing CPU overhead into network timing.

## Changes

**`NetworkCapacityServer.java`** — replaced `publishBlockStream()` with `open()` + `Pipelines.bidiStreaming()` + `PublishStreamRequestUnparsed`, matching `StreamPublisherPlugin` exactly. Only header and proof bytes are parsed; all item bytes stay as raw `Bytes`.

**`NetworkCapacityClient.java`** — two-phase streaming: Phase 1 pre-builds all batches before opening gRPC; Phase 2 streams them with no parse work in the hot path. Waits for all ACKs before half-closing. Supports `grpcEncoding`: `"identity"`, `"gzip"`, `"zstd"`.

**`network_capacity_config.proto`** — added `grpcEncoding` field to `HelidonWebClientConfig`.

**`clientDefaultConfig.json`** — CN production defaults (Helidon HTTP/2 defaults, `priorKnowledge: false`, `grpcEncoding: "identity"`).

**`serverDefaultConfig.json`** — current BN production defaults (8 MB HTTP/2 window, 128 KB socket buffers). Updated to optimized values in companion PR.


ata-nas (approved):
Unblocking tools.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/config/proto/network_capacity_config.proto
```diff
@@ -15,6 +15,9 @@ message HelidonWebClientConfig {
   bool priorKnowledge = 10;
   uint32 serverMaxMessageSizeBytes = 12;
   uint32 delayBetweenBlocksMs = 13;
+  // gRPC message encoding for outgoing requests: "identity" (no compression) or "gzip".
```
Nana-EC, line 18: note in the PR description you mentioned zstd support but here only cover gzip

tools-and-tests/tools/src/main/java/org/hiero/block/tools/capacity/NetworkCapacityServer.java
```diff
@@ -178,8 +168,24 @@ public void onError(final Throwable throwable) {
...
+            metrics.reportFinal();
+        }
+
+        private long parseBlockNumber(BlockItemUnparsed headerItem) {
+            try {
+                BlockHeader header = BlockHeader.PROTOBUF.parse(
+                        headerItem.blockHeaderOrThrow().toReadableSequentialData());
+                return header.number();
+            } catch (ParseException e) {
+                return currentBlockNumber + 1;
```
Nana-EC, line 187: This means on header parse we'll still acknowledge. Something to note so it's clear we're intentionally doing this

Nana-EC (approved):
LG, minor notes

tools-and-tests/tools/src/main/java/org/hiero/block/tools/config/proto/network_capacity_config.proto
AlfredoG87: Yes, that is on the next PR. I will update the description, thanks.

PR #2835, opened 2026-05-17
perf(config): Tune BN Server Defaults: Socket Buffers, Write Queue, Flow Control Timeout

AlfredoG87:
Increases the BN server's TCP receive buffer to match the HTTP/2 flow control window. Halves the flow control timeout as a conservative improvement for stalled-window edge cases. Updates the `networkCapacity` tool's server default config to match.

## Changes

| File | Parameter | Before | After |
|---|---|---|---|
| `ServerConfig.java` | `socketReceiveBufferSizeBytes` | 131,072 | 8,388,608 |
| `WebServerHttp2Config.java` | `flowControlTimeout` | 1,000 ms | 500 ms |
| `serverDefaultConfig.json` | `receiveBufferSize`, timeout, other tool defaults | various | see tool section |
| `clientDefaultConfig.json` | window, frame, encoding | old tool defaults | updated tool defaults |

### Socket receive buffer — 128 KB → 8 MB

The BN server advertises an HTTP/2 `initialWindowSize` of 8 MB. This means a Consensus Node may have up to 8 MB of gRPC data in flight before waiting for a window update. When the TCP receive buffer is smaller than the HTTP/2 window (128 KB << 8 MB), the OS kernel fills the receive buffer and signals TCP backpressure to the sender before the application-layer window is exhausted. The result is the CN slowing down not because of HTTP/2 flow control, but because of kernel-level stalls.

Matching the TCP receive buffer to the HTTP/2 window removes this mismatch.

### Flow control timeout — 1,000 ms → 500 ms

Reduces the maximum duration a blocked write will wait for a window update from the CN. Helidon's default is 50 ms; 500 ms is a conservative middle ground. This only affects edge cases where window updates are delayed (e.g., a slow or stalled CN); it has no effect in normal operation.

## Test Results

Tests were run isolated, one change at a time, with interleaved baselines to control for network drift. Client config was held constant at CN realistic defaults (65,535 HTTP/2 window, 16,384 max frame, identity encoding) throughout.

| Config | What changed | Throughput | vs baseline |
|---|---|---|---|
| Baseline | — | ~24,500 KB/s | — |
| **recv buffer 8 MB only** | `receiveBufferSize` 128KB→8MB | **~30,000 KB/s** | **+24%** |
| send buffer 8 MB only | `sendBufferSize` 128KB→8MB | ~25,400 KB/s | +3% (noise) |
| write queue 32,768 only | `writeQueueLength` 8192→32768 | ~25,800 KB/s | 0% (noise) |
| flow timeout 500 ms only | `flowControlTimeout` 1000→500ms | ~25,900 KB/s | +3% (noise) |
| both buffers (8 MB) | recv+send | ~29,400 KB/s | +20%* |

*Adding the send buffer on top of the receive buffer shows no additional gain — the receive buffer alone accounts for the full improvement.

The send buffer and write queue changes showed no isolated improvement and were not included.

## Tool defaults (`serverDefaultConfig.json`, `clientDefaultConfig.json`)

These files are used only by the `networkCapacity` benchmarking tool and do not affect the BN server. They were updated to match the BN's actual production config so tool results are representative.

### DIFF
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
@@ -29,7 +29,7 @@ import org.hiero.block.node.base.Loggable;
 public record ServerConfig(
         @Loggable @ConfigProperty(defaultValue = "131_072_000")
             @Min(1_048_576) @Max(1_610_612_736) int maxMessageSizeBytes,
-        @Loggable @ConfigProperty(defaultValue = "8_388_608")
+        @Loggable @ConfigProperty(defaultValue = "131_072")
             @Min(32768) @Max(Integer.MAX_VALUE) int socketSendBufferSizeBytes,
         @Loggable @ConfigProperty(defaultValue = "8_388_608")
             @Min(32768) @Max(Integer.MAX_VALUE) int socketReceiveBufferSizeBytes,
@@ -41,5 +41,5 @@ public record ServerConfig(
         @Loggable @ConfigProperty(defaultValue = "30") int idleConnectionTimeoutMinutes,
         @Loggable @ConfigProperty(defaultValue = "true") boolean tcpNoDelay,
         @Loggable @ConfigProperty(defaultValue = "8_192") int backlogSize,
-        @Loggable @ConfigProperty(defaultValue = "32_768") int writeQueueLength) {}
+        @Loggable @ConfigProperty(defaultValue = "8_192") int writeQueueLength) {}
 // spotless:on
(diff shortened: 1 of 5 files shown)


berryware (dismissed):
Looks Good

jsync-swirlds:
I don't think the inital window size affects the _send_ buffer.  Not critical, but we might experiment with shrinking that specific buffer as we expect to support quite a few clients and smaller client buffers will enable us to handle more clients efficiently.

Smaller buffers also tend to be more efficient when individual messages are small.

AlfredoG87:
> I don't think the inital window size affects the _send_ buffer. Not critical, but we might experiment with shrinking that specific buffer as we expect to support quite a few clients and smaller client buffers will enable us to handle more clients efficiently.
> 
> Smaller buffers also tend to be more efficient when individual messages are small.

You are right, I repeated the tests now, running each change in isolation, and the biggest win came from `receiveBufferSize 128KB→8MB`

jsync-swirlds (dismissed):

Nana-EC (dismissed):

Nana-EC (approved):

jsync-swirlds (approved):

PR #2837, opened 2026-05-18
docs(mirror-node-integration): add concept doc and connecting a Mirror Node to a Block Node document

### LINKED ISSUE #2733: docs: create mirror-node-integration.md
## User Stories

BN-A6 - How Mirror Nodes connect to Block Nodes | BN-I4 - How to configure a Mirror Node for Block Streams

## Goal

Enable Mirror Node operators to connect to and subscribe from a Block Node without needing support from the Block Node team. Mirror Node teams must complete integration testing before the 0.75 testnet transition on June 11.

## Contents should cover

- Architecture overview - how Mirror Nodes subscribe to Block Nodes via the gRPC subscribe service
- Prerequisites - Block Node version, network access, TLS certificates
- Configuration - what to set on the Mirror Node side to point at a Block Node
- How to verify the connection is live and blocks are flowing
- Behaviour on reconnect and gap handling
- Where to look when something goes wrong (cross-reference troubleshooting doc)

The `SubscriberServicePlugin` and `BlockStreamSubscriberSession` are implemented in `block-node/stream-subscriber`. Use these as the source of truth for behaviour.

deshmukhpranali:
## Reviewer Notes

Adds two documents for Mirror Node operators preparing for the
records-to-block-streams cutover (HIP-1193):

- **`docs/block-node/mirror-node-integration.md`** - concept doc.
  Covers the rationale for subscribing, the request/response model (`BlockStreamSubscribeService.subscribeBlockStream`), the full set of status codes, multi-Block-Node selection behaviour (`hiero.mirror.importer.block.nodes[]` priority + scheduler), and the two non-obvious operator behaviours: slow consumers see gaps rather than closed streams, and reconnection is the client's responsibility.

- **`docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md`**
  - step-by-step operator guide. Lead-in links to existing Block Node deployment guides and the Mirror Node installation guide.
  Configuration tables on the Block Node side (`subscriber.*`, `server.*`, `server.http2.*`) and the real
  Mirror Node side (`hiero.mirror.importer.block.enabled`, `block.sourceType`, `block.nodes[]`, `block.scheduler.*`,
  `block.stream.*`, `block.cutover.*`) with a working YAML example for two-Block-Node failover. 
  Four-step procedure (reachability, serverStatus, configure-and-restart, smoke-test the subscribe
  call). Verification section using Block Node logs, metrics, and status.

## User stories addressed

- *As a mirror node operator, I want to understand details on how mirror nodes connect to one or more block nodes, so that I can better understand mirror node configuration for migration.*
  
- *As a mirror node operator, I want procedures on configuring mirror nodes to receive streams from block nodes, so that I can properly connect my mirror node.*
  
## Test plan

All shell commands in the connection guide were executed verbatim
against a locally-built Block Node \`app-0.35.0-SNAPSHOT\` on a clean working directory.

- [x] Step 1 reachability (\`nc -vz\`) succeeds.
- [x] Step 1 protobuf bundle download + \`serverStatus\` returns the
      documented empty-Block-Node JSON.
- [x] Step 3 smoke test (\`subscribeBlockStream\` against empty BN)
      returns \`{"status": "NOT_AVAILABLE"}\` as documented.
- [x] Verification metrics scrape returns both
      \`blocknode_subscriber_open_connections\` and
      \`blocknode_subscriber_errors_total\` (the latter
      \`_total\`-suffixed by Prometheus convention).
- [x] Troubleshooting table verified: \`Failed to dial: connection
      refused\`, \`INVALID_END_BLOCK_NUMBER\`, reflection error wording
      (\`malformed header: missing HTTP content-type\`).

## Related Issue(s)

Fixes #2733


docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,335 @@
...
+4. Diagnose the most common failure modes.
+
+## Prerequisites
+
+Before you begin, ensure you have:
+
+- A running Block Node deployment using one of the following methods:
+  - [**Manual Single-Node Kubernetes Deployment**](./single-node-k8s-deployment.md)
+  - [**Solo Provisioner Single-Node Kubernetes Deployment**](./solo-weaver-single-node-k8s-deployment.md)
+  - [**Block Node Dev Quickstart (Docker)**](../quickstart.md) — for local development and testing.
```
jsync-swirlds: Here we should add a reference to public block nodes

```suggestion
- A running Block Node deployment using one of the following methods:
  - [**Manual Single-Node Kubernetes Deployment**](./single-node-k8s-deployment.md)
  - [**Solo Provisioner Single-Node Kubernetes Deployment**](./solo-weaver-single-node-k8s-deployment.md)
  - [**Block Node Dev Quickstart (Docker)**](../quickstart.md) — for local development and testing.
  - [**Block Nodes Published on the Network**](https://hips.hedera.com/all#hip-1137) - For production use with production networks.
```

We might also want to include a paragraph explaining up front that _in production_ mirror node operators are not generally expected to run their own block node. The exception being that an operator with _many_ mirror nodes should run a "tier 2" block node to ease load on tier 1 and also provide local buffer and pre-validation.

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,335 @@
...
+- A running Mirror Node deployment ready to be reconfigured to consume from a Block Node:
+  - [**Mirror Node Installation Guide**](https://github.com/hiero-ledger/hiero-mirror-node/blob/main/docs/installation.md) — local or Docker Compose install paths.
+  - [**Mirror Node Configuration Reference**](https://github.com/hiero-ledger/hiero-mirror-node/blob/main/docs/configuration.md) — full property reference for the Mirror Node services.
+- Network connectivity between the Mirror Node host and each Block Node host on the gRPC port (default `40840`).
+- A gRPC client capable of HTTP/2 server-streaming calls:
+  - The Mirror Node's built-in gRPC stack (for production integration), **or**
+  - [**`grpcurl`**](https://github.com/fullstorydev/grpcurl) — for ad-hoc verification from a shell (optional but recommended). Install via `brew install grpcurl` on macOS or your distribution's package manager on Linux.
```
jsync-swirlds: Perhaps also note Windows install:

```suggestion
  - [**`grpcurl`**](https://github.com/fullstorydev/grpcurl) — for ad-hoc verification from a shell (optional but recommended). Install via `brew install grpcurl` on macOS, following the [official guide](https://grpcurl.com/how-do-i-install-grpcurl-on-my-system/) on Windows, or your distribution's package manager on Linux.
```

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,335 @@
...
+| Block Node version     | A release matching the proto and configuration referenced in this guide. [NEEDS-VERIFICATION — confirm the minimum supported release once a tagged version after `0.33.0` is available.] |
+| gRPC port reachable    | Default `40840`; configurable via `server.port`.                                                                                                                                         |
+| Metrics port reachable | Default `16007` for verification (optional but recommended).                                                                                                                             |
+| Block availability     | The Block Node must have ingested at least `start_block_number` before the Mirror Node subscribes. Query `serverStatus` to confirm.                                                      |
+| Transport              | HTTP/2 over plaintext, or TLS terminated at infrastructure (load balancer, ingress). The Block Node process itself does not terminate TLS.                                               |
```
jsync-swirlds: Given that we don't expect mirror node operators to run their own block node in most cases, perhaps it would be better to refer them to the block node documentation here?

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,335 @@
...
+| Proto definitions | `block_stream_subscribe_service.proto`, `node_service.proto`, and `shared_message_types.proto` from [`protobuf-sources/src/main/proto/block-node/api/`](https://github.com/hiero-ledger/hiero-block-node/tree/main/protobuf-sources/src/main/proto/block-node/api). For ad-hoc `grpcurl` use, the matching versioned bundle from the [Block Node releases](https://github.com/hiero-ledger/hiero-block-node/releases) page is the easiest source — see [Step 1](#step-1-confirm-each-block-node-is-reachable-and-serving-blocks). |
+| gRPC reflection   | The Block Node does **not** enable gRPC server reflection on the public port. Clients must supply protobuf descriptors explicitly.                                                                                                                                                                                                                                                                                                                                                                                                |
+
+## Configuration
+
+### Block Node settings that affect subscribers
```
jsync-swirlds: As above, it might be better to refer to the block node documentation here for brevity.

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,335 @@
...
+- **Expected output** with one Mirror Node connected:
+
+  ```
+  # TYPE blocknode_subscriber_open_connections gauge
+  blocknode_subscriber_open_connections 1
+  # TYPE blocknode_subscriber_errors counter
+  blocknode_subscriber_errors_total 0
+  ```
+
+  The counter is exposed with the Prometheus-conventional `_total` suffix even though the underlying metric is registered as `subscriber_errors`. The `_open_connections` gauge is updated lazily: a closed session's decrement is processed only when the next subscriber attempts to connect. In low-traffic windows the gauge can appear stuck on the previous value. Treat the gauge as approximate; use the Mirror Node side (last committed block, reconnect rate) and infrastructure-level connection counts (load balancer, ingress) for precise observation.
```
jsync-swirlds: This is specific to entities running a block node; which is not necessarily expected of mirror node operators.
Perhaps move this to an appendix or other more clearly optional section?

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,335 @@
...
+| All configured Block Nodes marked inactive                                                   | Network reachability problem, or all Block Nodes simultaneously unhealthy.                                                 | Verify host/port for each `nodes[]` entry; check that `requiresTls` matches the actual termination setup; inspect each Block Node's metrics endpoint.                                                                                     |
+| `blocknode_subscriber_open_connections` does not increment after Mirror Node connects        | Connection is not reaching the Block Node.                                                                                 | Re-check firewall rules; verify the Mirror Node is hitting the correct host and port.                                                                                                                                                     |
```
jsync-swirlds: Slight wording improvement:

```suggestion
| `blocknode_subscriber_open_connections` does not increment after Mirror Node connects        | Connection is not reaching the Block Node.                                                                                 | Re-check firewall rules; verify the Mirror Node is connecting to the correct host and port.                                                                                                                                                     |
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+This document explains why Mirror Nodes consume block streams from a Block Node and the high-level model operators must understand before configuring the integration.
+
+For step-by-step configuration and verification commands, see the companion guide on [Connecting a Mirror Node to a Block Node](./operations/connecting-a-mirror-node-to-a-block-node.md).
+
+## Why Mirror Nodes subscribe to Block Nodes
+
+After the Hiero record-stream-to-block-stream cutover ([HIP-1193](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-1193.md)), Mirror Nodes consume finalized block data from Block Nodes over a long-lived gRPC stream instead of polling record files from cloud storage. Subscribing to a Block Node gives a Mirror Node:
+
+- **Low-latency live data** - blocks are pushed as they are finalized, rather than waiting for batch file uploads to a cloud bucket.
+- **Cryptographic verification at the boundary** - Block Nodes verify each block's aggregated signature and block proof before re-streaming.
```
jsync-swirlds: This is false, block nodes stream blocks immediately, not after verification.

```suggestion
- **Self-Contained Cryptographic verification** - Block Streams incorporate a Block Proof for each block that cryptographically verifies that block with no additional data.
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+- **Random-access historical replay** - a single gRPC call can replay an arbitrary block range to a freshly provisioned Mirror Node.
+- **A decentralized data plane** - Mirror Nodes can subscribe to one or more Block Nodes (Tier 1 or Tier 2) for redundancy.
```
jsync-swirlds, line 14: Given the incorporation of Block Node installation in the companion guide, should we include a note here about optionally running your own block node?

```suggestion
- **Random-access historical replay** - a single gRPC call can replay an arbitrary block range to a freshly provisioned Mirror Node.
- **A decentralized data plane** - Mirror Nodes can subscribe to one or more Block Nodes (Tier 1 or Tier 2) for redundancy.
- **Local Storage and Consolidation** - Mirror Node operators may choose to operate a local Block Node to independently store block data and redistribute the Block Stream to multiple Mirror Nodes from a single connection.
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+
+The network-level placement of Block Nodes between Consensus Nodes and Mirror Nodes is illustrated in the Block Node overview diagram:
+
+![block-node-network-architecture](../assets/block-node-network-architecture.svg)
+
+## Subscription model
+
+A Mirror Node subscribes to a Block Node by opening a **server-streaming gRPC call** to the `BlockStreamSubscribeService.subscribeBlockStream` RPC. The Block Node streams `SubscribeStreamResponse` messages back until the requested block range is fully served, the client disconnects, or an error occurs.
```
jsync-swirlds: Block Node will also recycle the connection on occasion (typically once per day)

```suggestion
A Mirror Node subscribes to a Block Node by opening a **server-streaming gRPC call** to the `BlockStreamSubscribeService.subscribeBlockStream` RPC. The Block Node streams `SubscribeStreamResponse` messages back until the requested block range is fully served, the connection reaches the maximum connection lifetime, the client disconnects, or an error occurs.
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Request
+
+A `SubscribeStreamRequest` carries two `uint64` fields:
```
jsync-swirlds:
```suggestion
A `SubscribeStreamRequest` expects two `uint64` fields:
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+
+|    Variant     |                                                                   Carries                                                                   |           Frequency           |
+|----------------|---------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------|
+| `block_items`  | A `BlockItemSet` with one or more `BlockItem` messages, batched into chunks. Large blocks are split across multiple `block_items` messages. | Many per block.               |
+| `end_of_block` | A `BlockEnd` message containing the completed `block_number`.                                                                               | Exactly one per block.        |
+| `status`       | A terminal `Code` value indicating why the stream is ending.                                                                                | Exactly one, at stream close. |
+
+Over the wire the pattern is: `block_items+ end_of_block` repeated per block, followed by exactly one terminal `status` when the stream closes.
```
jsync-swirlds: Minor typo:
```suggestion
Over the wire the pattern is: `block_items + end_of_block` repeated per block, followed by exactly one terminal `status` when the stream closes.
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Status codes
+
+|             Code             | Value |                                                       Meaning                                                       |
+|------------------------------|-------|---------------------------------------------------------------------------------------------------------------------|
+| `UNKNOWN`                    | `0`   | Reserved sentinel; the server MUST NOT return this. If observed, treat as a server-side defect.                     |
+| `SUCCESS`                    | `1`   | Stream ended normally; all requested blocks were sent.                                                              |
```
jsync-swirlds:
```suggestion
| `SUCCESS`                    | `1`   | Stream ended normally; all requested blocks were sent or the connection lifetime limit was reached.                 |
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+| `ERROR`                      | `3`   | Block Node encountered an internal error and cannot continue. Client MAY retry.                                     |
+| `INVALID_START_BLOCK_NUMBER` | `4`   | `start_block_number` is negative or outside the allowed range (including beyond the future-window).                 |
+| `INVALID_END_BLOCK_NUMBER`   | `5`   | `end_block_number` is invalid (negative, or less than `start_block_number`).                                        |
+| `NOT_AVAILABLE`              | `6`   | The requested stream is not available from this Block Node. Client MAY retry later or query a different Block Node. |
+
+## Connecting to multiple Block Nodes
+
+A Mirror Node can be configured to subscribe to more than one Block Node. The Mirror Node holds the full list as a configured array under `hiero.mirror.importer.block.nodes[]` and chooses which node to use according to a scheduler strategy - by default, `PRIORITY_THEN_LATENCY`. Each entry carries a `priority` (lower is higher priority) and an optional `requiresTls` flag.
```
jsync-swirlds: Small suggestion:
```suggestion
A Mirror Node should be configured to subscribe to more than one Block Node. The Mirror Node holds the full list as a configured array under `hiero.mirror.importer.block.nodes[]` and chooses which node to use according to a scheduler strategy - by default, `PRIORITY_THEN_LATENCY`. Each entry carries a `priority` (lower is higher priority) and an optional `requiresTls` flag.
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+
+The exact property names and the `nodes[]` YAML shape are documented in [Connecting a Mirror Node to a Block Node - Block Node endpoints](./operations/connecting-a-mirror-node-to-a-block-node.md#block-node-endpoints).
+
+## Behaviour Mirror Node operators must know
+
+Two non-obvious behaviours shape how a Mirror Node should integrate with a Block Node.
+
+### Slow consumers see gaps, not closed streams
+
+If a Mirror Node consumes more slowly than the Block Node ingests new blocks, the Block Node does **not** close the session. It silently skips whole blocks from the head of the per-session live queue once free capacity falls below a configurable threshold, allowing the stream to stay caught up to a more recent block at the cost of skipped intermediate blocks.
```
jsync-swirlds: This is not accurate.
The block node silently switches to history and streams from history until the slow subscriber catches up.
Gaps _do not_ indicate a slow subscriber, they indicate blocks that failed to complete or other issues in the stream _from the CN_ that resulted in resend requests.

Gaps are not an optimization, they're a consequence of sending the _unverified_ stream as it's received, rather than waiting for blocks to finish and be verified (which would add somewhere from 5ms to 500ms latency, or more for very large blocks, network errors, or under exceptionally heavy load).


```suggestion
Because a Mirror Node consumes the unverified block stream, gaps, interruptions, out-of-order resend, and incomplete blocks are visible in the subscribed stream.
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+
+The implication for the Mirror Node is that it must detect gaps client-side - by checking whether each received `end_of_block.block_number` is exactly `(last_committed_block + 1)` - and reconnect to backfill the missing range from history. The Block Node will not signal the gap itself.
```
jsync-swirlds: Gaps, out-of-order, failed verification, incomplete blocks.
All of these are possible because the stream is _unverified_; the block _items_ are sent as they're received rather than waiting for each block to complete and verify.  This means almost every blip, hiccough, interruption, resend, or other disruption in the stream coming from CN is visible in the subscribe stream.

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+
+This document explains why Mirror Nodes consume block streams from a Block Node and the high-level model operators must understand before configuring the integration.
+
+For step-by-step configuration and verification commands, see the companion guide on [Connecting a Mirror Node to a Block Node](./operations/connecting-a-mirror-node-to-a-block-node.md).
+
+## Why Mirror Nodes subscribe to Block Nodes
+
+After the Hiero record-stream-to-block-stream cutover ([HIP-1193](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-1193.md)), Mirror Nodes consume finalized block data from Block Nodes over a long-lived gRPC stream instead of polling record files from cloud storage. Subscribing to a Block Node gives a Mirror Node:
+
+- **Low-latency live data** - blocks are pushed as they are finalized, rather than waiting for batch file uploads to a cloud bucket.
```
jsync-swirlds:
```suggestion
- **Low-latency live data** - blocks are pushed as they are produced, rather than waiting for batch file uploads to a cloud bucket.
```

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,335 @@
...
+  { "status": "NOT_AVAILABLE" }
+  ```
+
+### Step 4: Handle disconnects and gaps
+
+The Block Node closes the stream when the finite range is fully served, when an internal error occurs, or when the client disconnects. The Mirror Node handles reconnection automatically: it will retry against the highest-priority reachable Block Node, governed by `maxSubscribeAttempts` and `readmitDelay`.
+
+Two operator-visible patterns are worth knowing:
+
+- **Gap in `end_of_block.block_number`**: the Mirror Node fell behind, the Block Node trimmed whole blocks from the live-queue head, and the Mirror Node must reconnect to backfill the missing range from history. The Mirror Node detects this and re-subscribes with `start_block_number = (last_committed_block + 1)`. If the gap is not available on the current Block Node (terminal `NOT_AVAILABLE`), failover to a higher-priority Tier-1 archive node is required — configure such a node as a low-priority entry under `nodes[]` so the Mirror Node can fall over automatically.
```
jsync-swirlds: This is slightly inaccurate.

```suggestion
- **Gap in `end_of_block.block_number`**: The flow of data from consensus node to block node arrived out-of-order or required a resend due to verification or persistence failure. The Mirror Node may reconnect to backfill the missing range from history. The Mirror Node detects this and re-subscribes with `start_block_number = (last_committed_block + 1)`. If the gap is not available on the current Block Node (terminal `NOT_AVAILABLE`), failover to a higher-priority Tier-1 archive node is required — configure such a node as a low-priority entry under `nodes[]` so the Mirror Node can fall over automatically.
```

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,335 @@
...
+- **Repeated `ERROR (3)` from one Block Node**: the Block Node is failing internally. The Mirror Node will mark it inactive after `maxSubscribeAttempts` and try the next configured node.
+
+See [Slow consumers see gaps, not closed streams](../mirror-node-integration.md#slow-consumers-see-gaps-not-closed-streams) in the concept doc for the rationale.
```
jsync-swirlds: The "slow consumers" part is misleading here.  Slow consumers see data fed from history, not gaps. The subscribers see gaps because the stream is _unverified_.

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,335 @@
...
+| `blocknode_subscriber_open_connections` does not increment after Mirror Node connects        | Connection is not reaching the Block Node.                                                                                 | Re-check firewall rules; verify the Mirror Node is hitting the correct host and port.                                                                                                                                                     |
+| **Gap in `end_of_block.block_number` after going live**                                      | The Mirror Node fell behind and the Block Node trimmed whole blocks from the live queue head.                              | The Mirror Node will reconnect to backfill from history. To reduce the rate of trims, increase `subscriber.liveQueueSize` and/or `subscriber.minimumLiveQueueCapacity` on the Block Node, or improve the Mirror Node's commit throughput. |
```
jsync-swirlds: The cause here is not a slow mirror node, and changing these values will have no impact on gaps in block number.

```suggestion
| **Gap in `end_of_block.block_number` after going live**                                      | The live block stream experienced interruptions or errors and blocks were received out of order or were resent.            | The Mirror Node will reconnect to backfill from history.                                                                                                                                                                                  |
```

docs/block-node/mirror-node-integration.md
```diff
@@ -0,0 +1,95 @@
...
+
+The implication for the Mirror Node is that it must detect gaps client-side - by checking whether each received `end_of_block.block_number` is exactly `(last_committed_block + 1)` - and reconnect to backfill the missing range from history. The Block Node will not signal the gap itself.
+
+### Reconnection is the client's responsibility
+
+The Block Node closes the gRPC stream when (a) a finite requested range is fully served, (b) an internal error or runtime exception is raised, or (c) the client disconnects. In every case other than (c), the Mirror Node must implement reconnection logic, typically with exponential backoff and a fresh `serverStatus` check before each retry. The Mirror Node's importer ships with this logic built in and tunes it through the `block.stream.*` properties. See the [Connecting a Mirror Node to a Block Node](./operations/connecting-a-mirror-node-to-a-block-node.md#step-4-handle-disconnects-and-gaps) guide for the operator-facing details.
```
jsync-swirlds: Missing connection lifetime limit:

```suggestion
The Block Node closes the gRPC stream when (a) a finite requested range is fully served, (b) the connection reaches the connection lifetime limit configured for that block node, (c) an internal error or runtime exception is raised, or (d) the client disconnects. In every case other than (c), the Mirror Node must implement reconnection logic, typically with exponential backoff and a fresh `serverStatus` check before each retry. The Mirror Node's importer ships with this logic built in and tunes it through the `block.stream.*` properties. See the [Connecting a Mirror Node to a Block Node](./operations/connecting-a-mirror-node-to-a-block-node.md#step-4-handle-disconnects-and-gaps) guide for the operator-facing details.
```

jsync-swirlds (commented):
Thank you for writing this up.
There are a few small recommendations, and one significant misapprehension regarding what happens to slow mirror node subscribers and why/how gaps and other "bad" data can be present in the stream from a block node.

jsync-swirlds (dismissed):
Well done.

We need to have a future update to add instructions for the cloud backup source for block stream files (if all block nodes are unavailable), but that makes more sense as a separate PR.

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,309 @@
...
+The Mirror Node ships with built-in multi-Block-Node support: configure one or more Block Nodes under `hiero.mirror.importer.block.nodes[]` and the Mirror Node selects between them by priority and measured latency, failing over automatically if a node becomes inactive.
+
+This guide shows how to:
+
+1. Confirm each target Block Node is reachable and serving the desired block range.
+2. Configure the Mirror Node to subscribe — including the migration cutover toggle.
+3. Verify the connection through logs, metrics, and the Block Node's `serverStatus` endpoint.
+4. Diagnose the most common failure modes.
+
+> **Most Mirror Node operators consume the block stream from a public Block Node and do not run their own.** The exception is an operator running many Mirror Nodes, who may run a Tier 2 Block Node as a local buffer that pre-validates and redistributes the stream to multiple Mirror Nodes from a single connection upstream. This guide covers both cases; sections specific to running your own Block Node are marked accordingly.
```
Nana-EC, line 20: Can we re-word this so it's not misinterpreted as a prescription of ways MN operator could/should run BNs.
If anything just say this covers MNs pointing to Tier 1 or Tier 2 BNs

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,309 @@
...
+| `hiero.mirror.importer.block.stream.responseTimeout`      | `400ms`                 | `serverStatus` request timeout.                                                                                                                 |
+
+#### Migration cutover
+
+During the records-to-block-streams cutover ([HIP-1193](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-1193.md)), the Mirror Node can switch automatically from record-file ingestion to block-stream subscription as the network rolls forward:
+
+|                           Property                           | Default  |                                    Effect                                     |
+|--------------------------------------------------------------|----------|-------------------------------------------------------------------------------|
+| `hiero.mirror.importer.block.cutover.enabled`                | `false`  | Enables the auto-switch. Keep `false` until the operator is ready to migrate. |
```
xin-hedera: In release 0.155, we changed back the default to unset and this make it dependent on the network. The effective value when this property is unset is as follows

- `true` for `mainnet` and `testnet`
- `false` for all other networks

If the property is set, it overrides the network default.

```suggestion
| `hiero.mirror.importer.block.cutover.enabled`                |   | Enables the auto-switch. Keep `false` until the operator is ready to migrate. |
```

We also added a new hapiVersion config property

```
| `hiero.mirror.importer.block.cutover.hapiVersion`                               | 0.76.0                                               | The HAPI version following which the final cutover will happen.                                                                                                                                                                                                    |
```

So the default cutover detection will only kick in when the network runs 0.76 release.

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,309 @@
...
+| `hiero.mirror.importer.block.cutover.firstStage.enabled`     | `false`  | Enables first-stage cutover with latency safety checks.                       |
+| `hiero.mirror.importer.block.cutover.firstStage.hapiVersion` | `0.75.0` | Minimum HAPI version that enables first-stage cutover.                        |
+| `hiero.mirror.importer.block.cutover.firstStage.maxLatency`  | `4s`     | If exceeded during the latency check, falls back to record stream.            |
```
xin-hedera: I'd prefer not to recommend these to third-party operators.

```suggestion
```

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,309 @@
...
+| `hiero.mirror.importer.block.cutover.threshold`              | `16s`    | Time to wait before switching between block stream and record stream.         |
+
+For migration, the recommended sequence is: ensure record-file ingestion is healthy → set `block.enabled=true` and `sourceType=AUTO` to begin block-stream verification in parallel → flip `block.cutover.enabled=true` to authorize the switch once the network reaches the cutover release.
```
xin-hedera: cutover detection and handling is automatic, we shouldn't recommend any config changes to operators.

```suggestion
```

The most important config is setting block node server host and optionally port. I'm not sure if we have finalized on the approach: HIP-1137 registered node so it's automatic, or manual configuration, or a later mirrornode release with a list of default block node servers.

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,309 @@
...
+
+  When both values are `uint64_max`, the Block Node has not yet ingested any blocks. The Mirror Node will receive `NOT_AVAILABLE (6)` if it tries to subscribe now.
+
+  Without `-emit-defaults`, `grpcurl` elides `"onlyLatestState": false` from the output; both forms are semantically equivalent.
+
+### Step 2: Configure the Mirror Node and restart
+
+1. Edit the Mirror Node importer's `application.yml` (or equivalent override).
+2. Set the [required properties](#required) and add a `nodes[]` entry per Block Node, as shown in the example above.
+3. If you are mid-migration and want the Mirror Node to switch from record-file ingestion to block-stream subscription automatically, also set [`hiero.mirror.importer.block.cutover.enabled=true`](#migration-cutover).
```
xin-hedera: This isn't needed.

```suggestion
```

xin-hedera (changes requested):

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
```diff
@@ -0,0 +1,305 @@
...
+| Session fails shortly after reconnect with `NOT_AVAILABLE (6)`                               | Mirror Node reconnected before the Block Node re-indexed the requested range after a restart.                              | Add a short delay and re-query `serverStatus` before each reconnect (already handled by the Mirror Node's readmit logic via `readmitDelay`).                                                                             |
+| Cutover does not switch to block-stream source                                               | `hiero.mirror.importer.block.cutover.enabled` is `false`, or the network has not reached `cutover.firstStage.hapiVersion`. | Set `cutover.enabled=true`; confirm the network HAPI version meets `cutover.firstStage.hapiVersion`.                                                                                                                     |
```
xin-hedera:
```suggestion
| Cutover does not switch to block-stream source                                               | `hiero.mirror.importer.block.cutover.enabled` is `false`, or the network has not reached `cutover.hapiVersion`. | Set `cutover.enabled=true`; confirm the network HAPI version meets `cutover.hapiVersion`.                                                                                                                     |
```

xin-hedera (changes requested):

jsync-swirlds (dismissed):

xin-hedera (approved):
LGTM

jsync-swirlds (approved):

Nana-EC (approved):

docs/block-node/operations/connecting-a-mirror-node-to-a-block-node.md
Nana-EC: Will push out so it's not blocking

PR #2846, opened 2026-05-18
feat: the ApplicationStateFacility should performantly persist block set ranges (#2823)

### LINKED ISSUE #2823: the ApplicationStateFacility should performantly persist block set ranges
### Story Form

As a BlockNode engineer I would like the ApplicationStateFacility to store data as performantly as possible.

### Technical Notes

We need to update the block ranges every block.
The task to save that data to disk (which is controlled entirely by the Application State Facility) should only run every {N} blocks (where {N} is configurable, perhaps 100 as a default).

imshubham22apr-gif:
This pull request implements performant block set ranges persistence within the ApplicationStateFacility to address the performance overhead of tracking block ranges on every block.

Summary of Implementation:
Performance-focused serialization: Persists the set of block ranges to a compact binary format file (block-ranges-data.bin) rather than scanning filesystem directories or using heavy text-based serialization.
Configurable Save Interval: Range updates are written to disk asynchronously using virtual thread executors every {N} blocks (defined by the blockRangesSaveInterval configuration property, defaulting to 100).
Graceful Shutdown Flush: Ensures all pending range updates since the last interval-based save are safely flushed to disk during the node shutdown sequence.
JDK 25 Windows Compatibility: Normalized directory paths within the Gradle Exec task in protobuf-sources to avoid backslash escaping issues in local Windows environments using Eclipse Adoptium JDK 25.
Robust Test Suite: Added a thorough unit test testBlockRangesPersistence() to BlockNodeAppTest.java verifying range serialization, save-on-interval behavior, and clean state restoration.
Related Issue(s)
Resolves #2823

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "interactive"
+}
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/node/NodeConfig.java
@@ -4,6 +4,7 @@ package org.hiero.block.node.app.config.node;
 import com.swirlds.config.api.ConfigData;
 import com.swirlds.config.api.ConfigProperty;
 import com.swirlds.config.api.validation.annotation.Min;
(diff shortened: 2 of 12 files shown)


AlfredoG87:
Closing due to inactivity and conflicts with current sprint work.

PR #2851, opened 2026-05-19
test: Test with real S3 bucket

### LINKED ISSUE #2810: Test with real S3 bucket
### Story Form

As a block node developer
I want to test the new S3 cloud arhive plugins with real S3 bucket
So that I am sure that they will store data correctly on the cloud

### Technical Notes

* Run a block node with the two new S3 cloud archive plugins
* Configure them to write into the respective S3 bucket
* Run the simulator tool and let it generate some load
* Observe how archives are uploaded in both buckets
* Kill block node and start it again
* Make sure that the storage plugin resumes work from where it left
* Task with various number of blocks per archive and various block size

ivannov:
##  Reviewer Notes
  
This PR adds a one-shot integration test for CloudStorageArchivePlugin that exercises the full plugin lifecycle against a real S3-compatible bucket. The test is @Disabled by default and skips automatically if a credentials properties file (~/.hedera/cloud-archive-integration-test.properties) is absent.

Supporting changes:
  
- Config prefix fix: CloudStorageArchiveConfig was annotated @ConfigData("cloud.archive") but the env var convention uses CLOUD_STORAGE_ARCHIVE_*. The prefix is corrected to cloud.storage.archive, and the docs table + existing tests are updated to match.
- update-env.sh: propagates the CLOUD_STORAGE_ARCHIVE_* env vars into the Docker .env file so the archive plugin can be configured for local Docker testing.

##  Related Issue(s)

Closes #2810

### DIFF
--- a/block-node/cloud-storage-archive/src/test/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchiveIntegrationTest.java
+++ b/block-node/cloud-storage-archive/src/test/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchiveIntegrationTest.java
@@ -1,0 +1,409 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.cloud.storage.archive;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+import com.hedera.bucky.S3Client;
+import com.hedera.hapi.node.base.NodeAddressBook;
+import com.hedera.pbj.runtime.OneOf;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.swirlds.config.api.ConfigurationBuilder;
+import java.io.BufferedReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Random;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.locks.LockSupport;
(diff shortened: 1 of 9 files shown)


block-node/cloud-storage-archive/src/test/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchiveIntegrationTest.java
```diff
@@ -0,0 +1,409 @@
...
+        }
+        activeContexts.clear();
+        try {
+            clearBucket();
+        } catch (final Exception ignored) {
+        }
+    }
+
+    @Test
+    @Disabled // remove this if you want to run the test
```
ata-nas: I think we should be able to put `@Disable` globally on class level. In that way all tests are disabled by default. I know we have only one test, but adding others here will disable them by default.

ata-nas (dismissed):
LG.

block-node/cloud-storage-archive/src/test/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchiveIntegrationTest.java
ivannov: I didn't plan to have this test at all pushed.

But its goal is to be a one-shot tests, that is one method only.

jsync-swirlds: It's good to have this in the codebase, and I tend to agree with Atanas that disabling the class might be worthwhile.

My thought is that while we have a single test here, we may find it worthwhile to add a few more that use real S3 services and we might use that in a "qualification" test process wherein we occasionally run this (manually or via well secured automation) to ensure that both existing S3 services and any new S3 services work as expected.

AlfredoG87 (dismissed):
LGTM

block-node/cloud-storage-archive/src/test/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchiveIntegrationTest.java
ivannov: Disabled the whole test class

ata-nas (approved):

jsync-swirlds (approved):

PR #2857, opened 2026-05-20
build: Include new S3 archive plugins in deployments

### LINKED ISSUE #2811: Include new S3 archive plugins in deployments
### Story Form

As a block node operator
I want to have the s3 cloud archive expanded and storage plugins in the helm charts
So that they are eventually deployed on the my block node instance

### Technical Notes

* Look for all the deployment descriptors where the old S3 archive plugin is included
* Add the cloud-archive-storage and cloud-archive-expanded plugins there
* Consider removing the old plugin

ivannov:
## Reviewer Notes

* Adds the `cloud-storage-expanded` and `cloud-storage archive` plugins to `all` and `rfh` profiles
* Does not remove the old `s3-archive` plugin

## Related Issue(s)

Closes #2811


AlfredoG87 (commented):
Looks good, just a curious:

1. Why did you not removed the old S3 Plugin?
2. Is the upgrade of curl download needed or just a nice to do?

What is the expected BN behaviour with all 3 Plugins?

ivannov:
> Looks good, just a curious:
> 
> 1. Why did you not removed the old S3 Plugin?
> 2. Is the upgrade of curl download needed or just a nice to do?
> 
> What is the expected BN behaviour with all 3 Plugins?

@1 I was not sure whether I should. I can do that

@2 We had this earlier. The curl library is with special version that is there limited time. I asked whether we can generalize it (it's the simulator). But I was told to keep it with concrete tags

charts/block-node-server/values-overrides/plugin-profile-rfh.yaml
```diff
@@ -1,2 +1,2 @@
 plugins:
-  names: "facility-messaging,block-access-service,health,server-status,stream-publisher,stream-subscriber,verification,blocks-file-recent,backfill,s3-archive,cloud-storage-archive,cloud-storage-expanded"
+  names: "facility-messaging,block-access-service,health,server-status,stream-publisher,stream-subscriber,verification,blocks-file-recent,backfill,cloud-storage-archive,cloud-storage-expanded"
```
jsync-swirlds, line 2: We _should_ be able to remove blocks-files-recent from this line as well.

jsync-swirlds (approved):
one minor suggestion.

Nana-EC (approved):

charts/block-node-server/values-overrides/plugin-profile-rfh.yaml
Nana-EC: This requires the PR for pulling block ranges first right?

jsync-swirlds: Probably?  Not sure exactly how those two interact.

PR #2862, opened 2026-05-20
feat: 2797 rsa bootstrap followup

### LINKED ISSUE #2762: Investigate gson handling of MirrorNodeResponse mapping
Should we just let GSon handle mapping to the reduced fields. Only the fields in the POJO are included for deserialization. All other fields are ignored. If the fields change over time, we just update the MirrorNodeNodesResponse.

MirrorNodeNodesResponse response = gson.fromJson(json, MirrorNodeNodesResponse.class);

_Originally posted by @berryware in https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3210762329_

Also see https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3211043490

### LINKED ISSUE #2797: RsaRosterBootstrapPlugin — Code Quality Follow-up (unresolved PR #2719 comments)
## Story Form

As a Block Node developer,
I want the unresolved review comments from PR #2719 to be addressed,
So that \`RsaRosterBootstrapPlugin\` follows correct Java interrupt-handling conventions, avoids exception-as-flow-control anti-patterns, uses approved JSON parsing libraries, stays current with the Mirror Node address book, and is fully documented in the configuration reference.

## Technical Notes

All items below are unresolved inline review comments from PR [#2719](https://github.com/hiero-ledger/hiero-block-node/pull/2719). Each links to the original comment.

### \`module-info.java\` — Gson must be replaced

- [x] **Replace Gson with PBJ or an approved alternative — Gson must not be used in production code.** jsync: *"We shouldn't be using GSON at all. We can use PBJ or we can look at other alternatives if necessary, but GSON shouldn't be used in production code. This isn't immediately blocking, but it's a high priority to replace."* The Mirror Node response is JSON, not protobuf, so PBJ JSON codec should be evaluated first; if the schema doesn't map cleanly, identify an alternative. [jsync comment](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3227710802)

> Note: the earlier Gson-simplification suggestion from berryware (tracked in [#2762](https://github.com/hiero-ledger/hiero-block-node/issues/2762)) is superseded by this — the goal is to remove Gson entirely, not improve how it is used.

### \`RsaRosterBootstrapPlugin.java\` — \`InterruptedException\` handling

- [x] **Do not catch \`InterruptedException\` in the retry sleep and wrap it as \`IllegalStateException\`.** ivannov and jsync both flagged this. Interruption is not an error state — it is a normal signal. The try block around \`Thread.sleep(delayMs)\` should be removed; either let \`InterruptedException\` propagate up the call stack naturally, or replace the sleep with \`LockSupport.parkNanos(delayMs * 1_000_000L)\` which does not throw. [ivannov comment](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3226826503) · [jsync comment](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3227199274) · [jsync suggestion](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3227448370)

### \`RsaRosterBootstrapPlugin.java\` — Exception flow-control in \`fetchAndParseWithRetry\`

- [x] **Refactor the exception-as-flow-control structure in \`fetchAndParseWithRetry\`.** jsync: *"We're throwing an exception as flow-control rather than handling a status return; and we're catching exceptions (including the one we just threw) just to record them for later."* Either: (a) extract the per-attempt logic into a dedicated helper that returns a success/failure code, or (b) restructure so the outer \`try\` block encloses the entire method body rather than nesting try-within-try. [jsync comment](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3227541642)

### \`RsaRosterBootstrapPlugin.java\` — Always check Mirror Node even when a local file is present

- [x] **Query Mirror Node even when the local bootstrap file is already loaded, and use whichever source is newer.** jsync: *"For follow-up, we should query mirror node even if we found a local file. If mirror node is newer, then we should use the version that is valid for the current time (or the newest if multiple are valid at the current time)."* The current file-first logic skips the MN call entirely when a file exists, which means a stale address book is never refreshed without a manual restart. [jsync comment](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3228172053)

### \`RsaRosterBootstrapPlugin.java\` — Potential NPE on \`entry.timestamp().to()\`

- [x] **Guard against NPE when \`entry.timestamp()\` itself is null.** rockysingh noted that \`entry.timestamp().to()\` will throw if \`entry.timestamp()\` returns null (not just when \`to()\` is null). The existing null check only guards \`timestamp.to\` — add a null check on \`timestamp\` first. [rockysingh comment](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3228487816)

### \`RsaRosterBootstrapPlugin.java\` — \`order=desc\` early-stop assumption

- [x] **Document (or harden) the \`order=desc\` early-stop assumption.** rockysingh: *"The code requests order=desc and assumes the first superseded entry means all remaining are also historical. But if Mirror Node has a bug or returns entries out-of-order, active nodes could be skipped."* At minimum add a comment in the code explaining the ordering guarantee being relied on. If Mirror Node cannot guarantee strict descending order, consider collecting all active entries before breaking. [rockysingh comment](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3228491007)

### \`RsaRosterBootstrapConfig.java\` — Configuration documentation

- [x] **Document the \`roster.bootstrap.rsa.*\` namespace in \`docs/block-node/configuration.md\`.** The config record was added without a corresponding entry in the configuration reference. [ivannov comment](https://github.com/hiero-ledger/hiero-block-node/pull/2719#discussion_r3225846220)

---

## References

- Source PR: [#2719](https://github.com/hiero-ledger/hiero-block-node/pull/2719)
- Gson simplification (superseded by full Gson removal): [#2762](https://github.com/hiero-ledger/hiero-block-node/issues/2762)

berryware:
## Reviewer Notes

1. Gson → PBJ: Drops MirrorNodeNodesResponse.java (hand-rolled Gson parser). Adds mn_response.proto in protobuf-sources/src/main/proto/internal/, generating types into org.hiero.block.internal via the shared protobuf-pbj module. Null or missing strings default to empty string so need to update the tests and the plugin
2. RsaRosterBootstrapPlugin Sync → Async plugin: start() no longer blocks/retries. It schedules fetchFromMirrorNode() via scheduleAtFixedRate, retrying on a shorter interval until a valid address book is obtained. It then retries at a longer interval to check for additional address book updates.
3. ScheduledBlockingExecutor bug fix: Repeating FutureTask one-shot no-op fixed.
4. TSS cleanup: queryPeerInitialDelay removed, initial delay hardcoded to 0.- replace GSON with PBJ for JSON parsing

## Related Issue(s)
Closes #2797
Closes #2762


block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapConfig.java
```diff
@@ -19,6 +20,10 @@
 @ConfigData("roster.bootstrap.rsa")
 public record RsaRosterBootstrapConfig(
         @Loggable @ConfigProperty(defaultValue = "") String mirrorNodeBaseUrl,
+
+        @Loggable @ConfigProperty(defaultValue = "10000") @Min(100)
+        int mirrorNodeQueryInterval,
+
         @Loggable @ConfigProperty(defaultValue = "5") int mirrorNodeConnectTimeoutSeconds,
         @Loggable @ConfigProperty(defaultValue = "10") int mirrorNodeReadTimeoutSeconds,
         @Loggable @ConfigProperty(defaultValue = "100") int mirrorNodePageSize) {}
```
jsync-swirlds, line 32: We've been disabling spotless on these because it mangles config badly.
(optional, of course)

```suggestion
public record RsaRosterBootstrapConfig(
// spotless:off
        @Loggable @ConfigProperty(defaultValue = "") String mirrorNodeBaseUrl,
        @Loggable @ConfigProperty(defaultValue = "10000") @Min(100) int mirrorNodeQueryInterval,
        @Loggable @ConfigProperty(defaultValue = "5") int mirrorNodeConnectTimeoutSeconds,
        @Loggable @ConfigProperty(defaultValue = "10") int mirrorNodeReadTimeoutSeconds,
        @Loggable @ConfigProperty(defaultValue = "100") int mirrorNodePageSize) {}
// spotless:on
```

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapConfig.java
```diff
@@ -19,6 +20,10 @@
 @ConfigData("roster.bootstrap.rsa")
 public record RsaRosterBootstrapConfig(
         @Loggable @ConfigProperty(defaultValue = "") String mirrorNodeBaseUrl,
+
+        @Loggable @ConfigProperty(defaultValue = "10000") @Min(100)
```
jsync-swirlds: What is the units for this?  Is this seconds, milliseconds, minutes?

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -62,8 +67,10 @@ public class RsaRosterBootstrapPlugin implements BlockNodePlugin {
     static final MetricKey<ObservableGauge> METRIC_ROSTER_LOAD_DURATION_MS =
             MetricKey.of("roster_load_duration_ms", ObservableGauge.class).addCategory(METRICS_CATEGORY);
 
-    /// Max Mirror Node fetch retries before fail-fast.
-    private static final int MAX_RETRIES = 3;
+    /// The ScheduledExecutorService used by the RosterBootstrapPlugin to query peer Mirror Nodes for a Node
+    /// Address Book
+    private ScheduledExecutorService queryMNExecutor;
```
jsync-swirlds: Generally, shortened names are treated like regular words in variable or class names.

```suggestion
    private ScheduledExecutorService queryMnExecutor;
```

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -121,21 +128,30 @@ public void start() {
...
-            book = fetchFromMirrorNode();
-            // BlockNodeApp.updateAddressBook() persists the file and calls onContextUpdate.
-            applicationStateFacility.updateAddressBook(book);
-            addressBookSource = "Mirror Node";
+            // Set up the asynchronous node address book fetcher
+            // Create thread executors via threadPoolManager.
+            queryMNExecutor = context.threadPoolManager()
+                    .createVirtualThreadScheduledExecutor(1, "queryPeerScanner", this::uncaughtExceptionHandler);
+
+            // Schedule periodic checking of bn peers for their TssData
```
jsync-swirlds: copy pasta?

```suggestion
            // Schedule periodic checking of mirror node for address book data
```

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -121,21 +128,30 @@ public void start() {
...
+            scheduledFuture = queryMNExecutor.scheduleAtFixedRate(
+                    this::fetchFromMirrorNode, 0, config.mirrorNodeQueryInterval(), TimeUnit.MILLISECONDS);
```
jsync-swirlds: 10 seconds is probably quite a bit too short an interval most of the time.  Address books don't change multiple times per minute...
Perhaps set minimum for the config to 1000 and default to 3_600_000 (1 hour)?

We might actually need two values, one for "we do not have address book data yet" (which is quite short intervals) and another for "we have address book data, and are just checking for updates" (which would be longer intervals).

block-node/roster-bootstrap-rsa/src/main/java/module-info.java
```diff
@@ -20,13 +20,14 @@
...
 
+    requires transitive com.hedera.pbj.runtime;
     requires transitive com.swirlds.config.api;
     requires transitive org.hiero.block.node.spi;
     requires org.hiero.block.node.base;
     requires org.hiero.block.protobuf.pbj;
     requires org.hiero.metrics;
-    requires com.google.gson;
     requires java.net.http;
+    requires org.antlr.antlr4.runtime;
```
jsync-swirlds: Do we actually need this ANTLR dependency?

jsync-swirlds (commented):
A few minor items (which are not required) and a couple questions.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: Agreed. This is just for the "no address book yet" scenario. Will add the second value in a future update

block-node/roster-bootstrap-rsa/src/main/java/module-info.java
berryware: Apparently so. The JSON parsing of PBJ is done by antlr.

```
* What went wrong:
Execution failed for task ':roster-bootstrap-rsa:checkModuleDirectivesScope'.
> /Users/dberry/Projects/hiero-ledger/hiero-block-node/block-node/roster-bootstrap-rsa/src/main/java/module-info.java
  
  Please add the following requires directives:
      requires org.antlr.antlr4.runtime;
```

jsync-swirlds: Ahh, I missed that we added the proto file to this plugin rather than adding it to the `internal` set in the main protos.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -183,69 +201,94 @@ private NodeAddressBook fetchFromMirrorNode() {
...
+                        config.mirrorNodeBaseUrl());
+                return;
             }
 
-            return NodeAddressBook.newBuilder().nodeAddress(addresses).build();
+            NodeAddressBook book =
+                    NodeAddressBook.newBuilder().nodeAddress(addresses).build();
+
+            // We found an address book stop the mirror node requests
+            scheduledFuture.cancel(true);
```
Nana-EC: nit: stop the schedule after a successful updateAddressBook call, to ensure data goes out first

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -183,69 +201,94 @@ private NodeAddressBook fetchFromMirrorNode() {
...
+
+            // BlockNodeApp.updateAddressBook() persists the file and calls onContextUpdate.
+            applicationStateFacility.updateAddressBook(book);
+            String addressBookSource = "Mirror Node";
+            rosterEntriesLoaded = book.nodeAddress().size();
+            LOGGER.log(
+                    INFO,
+                    "RSA roster available: {0} entries obtained from {1}",
+                    rosterEntriesLoaded,
+                    addressBookSource,
```
Nana-EC:
```suggestion
                    "RSA roster available: {0} entries obtained from {1}",
                    book.nodeAddress().size(),
                    "Mirror Node",
```

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
```diff
@@ -197,24 +183,37 @@ private void registerStaticHandler(final String path, final int statusCode, fina
...
+                @Override
+                public void onContextUpdate(BlockNodeContext context) {
+                    super.onContextUpdate(context);
+                    latch.countDown();
+                }
+            };
+
+            start(plugin, new SimpleInMemoryHistoricalBlockFacility(), serverConfig());
+
+            latch.await();
```
Nana-EC: This (and others) should have a timeout

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -183,69 +201,94 @@ private NodeAddressBook fetchFromMirrorNode() {
...
+                "RSA address book could not be created — Mirror Node API unavailable at {0}. BN cannot verify WRB proofs. Provide rsa-bootstrap-roster.json or ensure Mirror Node is reachable.",
+                config.mirrorNodeBaseUrl(),
+                lastCause);
+        return null;
+    }
+
+    /// UncaughtExceptionHandler for logging uncaught exceptions
+    private void uncaughtExceptionHandler(Thread thread, Throwable throwable) {
+        LOGGER.log(
+                WARNING, "Uncaught exception in RsaRosterBootstrapPlugin thread {0}: {1}", thread.getName(), throwable);
```
Nana-EC: I think the throwable needs to be the  trailing argument after all substitutions right?
```suggestion
                WARNING, "Uncaught exception in RsaRosterBootstrapPlugin thread {0}", thread.getName(), throwable);
```

Nana-EC (commented):
Nice additions

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
berryware: Updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: Updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: Updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: Updated

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
```diff
@@ -197,24 +184,37 @@ private void registerStaticHandler(final String path, final int statusCode, fina
...
+                @Override
+                public void onContextUpdate(BlockNodeContext context) {
+                    super.onContextUpdate(context);
+                    latch.countDown();
+                }
+            };
+
+            start(plugin, new SimpleInMemoryHistoricalBlockFacility(), serverConfig());
+
+            assertTrue(latch.await(5, TimeUnit.SECONDS));
```
jsync-swirlds: My most recent PR (#2833) makes the test-only scheduled blocking executor actually block until instructed to run (like the test-only blocking executor).
That makes these awaits never finish; so we'll need to add a line above this line once that PR merges (found this when TSS bootstrap test deadlocked).

Depending on merge order, that may need to be added here, if it does, the line to add is as follows:
`testThreadPoolManager.scheduledExecutor().executeSerially();`

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
```diff
@@ -227,7 +227,20 @@ void paginatedResponseCollectsAllNodes() {
...
+            CountDownLatch latch = new CountDownLatch(1);
+
+            RsaRosterBootstrapPlugin plugin = new RsaRosterBootstrapPlugin() {
+
+                @Override
+                public void onContextUpdate(BlockNodeContext context) {
+                    super.onContextUpdate(context);
+                    latch.countDown();
+                }
+            };
```
jsync-swirlds: This is repeated a few times in this class.
It might be worth extracting that into a single inner class implementation.
There is an example where I made that change for TSS bootstrap plugin test in PR #2833.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -183,69 +201,93 @@ private NodeAddressBook fetchFromMirrorNode() {
...
+            Thread.currentThread().interrupt();
+            lastCause = ie;
+        } catch (IOException | RuntimeException | ParseException e) {
+            lastCause = e;
+        }
+        LOGGER.log(
+                ERROR,
+                "RSA address book could not be created — Mirror Node API unavailable at {0}. BN cannot verify WRB proofs. Provide rsa-bootstrap-roster.json or ensure Mirror Node is reachable.",
+                config.mirrorNodeBaseUrl(),
+                lastCause);
```
jsync-swirlds: if we want stack trace, we need to do this a little different (System.Logger only recognizes a throwable if it's the _only_ extra parameter).

```suggestion
        final String message = "RSA address book could not be created — Mirror Node API unavailable at %s. BN cannot verify WRB proofs. Provide rsa-bootstrap-roster.json or ensure Mirror Node is reachable.".formatted(config.mirrorNodeBaseUrl());
        LOGGER.log(ERROR, message, lastCause);
```

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -183,69 +201,93 @@ private NodeAddressBook fetchFromMirrorNode() {
...
+        return null;
+    }
+
+    /// UncaughtExceptionHandler for logging uncaught exceptions
+    private void uncaughtExceptionHandler(Thread thread, Throwable throwable) {
+        LOGGER.log(WARNING, "Uncaught exception in RsaRosterBootstrapPlugin thread {0}", thread.getName(), throwable);
```
jsync-swirlds: Similar here, we have to use formatted to include extra parameters with a throwable if we want a stack trace.

```suggestion
        final String message = "Uncaught exception in %s thread %s%n".formatted(getClass().getSimpleName(), thread.getName());
        LOGGER.log(WARNING, message, throwable);
```

jsync-swirlds (dismissed):
A couple small logging fixes needed, but this looks good so far.

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
berryware: Updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: updated

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
berryware: Not sure this works as expected. Here is the changes made to get the tests to work:
```java
            testThreadPoolManager.scheduledExecutor().executeSerially();
            try {
                assertTrue(latch.await(5, TimeUnit.SECONDS));
                // ignore
            }
```
An InterruptedException gets thrown now that I have to ignore to proceed with the rest of the test.

One test fails. 
```java
        @Test
        @DisplayName("HTTP 500 triggers retry and succeeds on the next attempt")
        void http500TriggersRetryThenSucceeds() {
```
This test makes two calls to the http server. The second call is not being made. Not sure why yet, but am pointing it out in case you have any insight. It worked prior.

berryware: The first call fails and the second call is supposed to succeed. The next scheduled run of the virtual thread does not happen.

jsync-swirlds: Correct, the scheduled blocking executor only runs when told to (just like the regular blocking executor), so the execute serially call is needed for each expected run of a scheduled event.
Good: This means you don't have to wait for scheduled tasks
Bad: This means you have to tell the executor to run the tasks each time.

The executor _does_ re-execute the same task if it's scheduled to be repeated, you just have to call executeSerially or executeAsync for each such run.

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
```diff
@@ -203,13 +190,21 @@ void singlePageLoadsAllNodes() {
...
+            CountDownLatch latch = new CountDownLatch(1);
+            RsaRosterBootstrapPlugin plugin = new TestBootstrapPlugin(latch);
 
+            start(plugin, new SimpleInMemoryHistoricalBlockFacility(), serverConfig());
+            testThreadPoolManager.scheduledExecutor().executeSerially();
+            try {
+                assertTrue(latch.await(5, TimeUnit.SECONDS));
+            } catch (InterruptedException e) {
+                // ignore
+            }
```
jsync-swirlds: With the use of `executeSerially()` you could probably eliminate the latch entirely; the method will have always completed because it's run on the same thread as this unit test.

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
jsync-swirlds: Side note, because the scheduled tasks are executing serially, you don't actually need the latch anymore; you could just remove it.

jsync-swirlds (dismissed):

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -81,7 +81,7 @@ public class RsaRosterBootstrapPlugin implements BlockNodePlugin {
     // Metric values stored after startup so ObservableGauge can read them
     private volatile long rosterEntriesLoaded = 0L;
     private volatile long rosterLoadDurationMs = 0L;
-    private long rosterLoadStartMs = 0L;
+    private volatile long rosterLoadStartMs = 0L;
```
jsync-swirlds: (very minor, no need to update)
`volatile` isn't actually needed here.
The as-if-serial guarantee ensures that the schedule thread cannot start before this value is visible to the scheduled thread because the task is scheduled after the initial value is set.
After that, this value is not updated (it could be final if it was set in the constructor instead of `start()`), so there is no concern about updates.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -121,33 +133,40 @@ public void start() {
...
+            // Create thread executors via threadPoolManager.
+            queryMnExecutor = context.threadPoolManager()
+                    .createVirtualThreadScheduledExecutor(1, "queryMnScanner", this::uncaughtExceptionHandler);
+
+            // No retry limit — the node requires a roster to verify WRB proofs; keep polling
+            // until Mirror Node is reachable. Each failed attempt is logged at ERROR level.
+            // Will be canceled once an address book is retrieved.
+            scheduledFuture = queryMnExecutor.scheduleAtFixedRate(
+                    this::fetchFromMirrorNode, 0, config.mirrorNodeQueryIntervalMillis(), TimeUnit.MILLISECONDS);
+        } else {
```
jsync-swirlds, line 146: What if the local file is out of date?  This if/else just accepts the local file and never queries the mirror node in that case.  The MN data is always preferred if available (I think), so even if we have local data that lets the BN launch quickly, it's probably a good idea to check with MN as well.

The other concern I have here is how do we continue to query mirror node for updates?  The RSA roster _can_ change at any time (typically it's membership only changes at upgrades, but we don't necessarily want to require BN upgrade at the same time as CN either).

These might be handled in a later issue, but we should at least keep them in mind.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -161,14 +180,17 @@ private NodeAddressBook fetchFromMirrorNode() {
...
-                for (final MirrorNodeNodesResponse.NodeEntry entry : response.nodes()) {
+                final MirrorNodeNodesResponse response = fetchAndParse(client, nextUrl);
+                if (response == null) {
+                    return;
+                }
+                for (final NodeEntry entry : response.nodes()) {
                     // A non-null timestamp.to means this entry has been superseded — all subsequent
                     // entries (in descending order) are also historical; stop here.
-                    if (entry.timestamp() != null && entry.timestamp().to() != null) {
+                    if (entry.timestamp() != null && !entry.timestamp().to().isBlank()) {
```
jsync-swirlds, line 203: There is a subtle NPE opportunity here.  That might be acceptable, but I thought it worth noting.

```suggestion
                    if (entry.timestamp() != null && entry.timestamp().to() != null && !entry.timestamp().to().isBlank()) {
```

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -161,14 +180,17 @@ private NodeAddressBook fetchFromMirrorNode() {
...
                         break outer;
                     }
-                    if (entry.publicKey() == null || entry.publicKey().isBlank()) {
+                    if (entry.publicKey().isBlank()) {
```
jsync-swirlds, line 206: Kind of the same thing here, this change means a null public key field will throw instead of being treated as no data...

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -161,14 +180,17 @@ private NodeAddressBook fetchFromMirrorNode() {
 
             outer:
             while (nextUrl != null) {
-                final MirrorNodeNodesResponse response = fetchAndParseWithRetry(client, nextUrl);
-                for (final MirrorNodeNodesResponse.NodeEntry entry : response.nodes()) {
+                final MirrorNodeNodesResponse response = fetchAndParse(client, nextUrl);
+                if (response == null) {
+                    return;
```
jsync-swirlds, line 198: The returns are a little confusing (it looks like we give up).

We might want to document in the method javadoc that the method is scheduled _repeatedly_ until it succeeds, so return from this method without cancelling the future just means we will try again later.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
jsync-swirlds: There is a subtle bug here.

We are cancelling the thread currently running this method.
We probably shouldn't pass `true` for the parameter that says to interrupt the running thread...

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -183,69 +205,92 @@ private NodeAddressBook fetchFromMirrorNode() {
...
+        Exception lastCause;
+        try {
+            final HttpRequest request = HttpRequest.newBuilder()
+                    .uri(URI.create(url))
+                    .timeout(Duration.ofSeconds(config.mirrorNodeReadTimeoutSeconds()))
+                    .GET()
+                    .build();
+            final HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+            if (response.statusCode() != 200) {
+                throw new IOException("HTTP " + response.statusCode() + " from " + url);
```
jsync-swirlds: Instead of throwing, can we just log the error and let the code try again later?

jsync-swirlds (commented):
A few questions, and one subtle bug.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: PBJ defaults null and missing string entries to "". It shouldn't be null. 

from lint:
Condition 'entry.timestamp().to() != null' is always 'true'

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: PBJ defaults null and missing string entries to "". It shouldn't be null. 

from lint:
Condition 'entry.publicKey() != null' is always 'true'

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: Here's the new logic. @Nana-EC please let me know if you are okay with it. 

- There are two MN Query intervals: initial and subsequent.
- initial is used when there is no local file. It is a short interval that runs until a node address book is received.
- subsequent is used, after a local file is found, or after the initial intervals have found an address book.

Nana-EC: So this covers 2 then
1. BN gets initial address book from local file then after the subsequent MN query interval will get an update if it exists and after the subsequent MN query interval will get an update as scheduled if it exists
2. BN gets the initial address book from MN if no local is found and after the subsequent MN query interval will get an update as scheduled if it exists

\Is that right?

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
jsync-swirlds: OK.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
jsync-swirlds: This looks reasonable.

I wonder if we could simplify, however.
1. If a local file is configured, read that value on startup.
1. If MN URL is configured, begin querying mirror node on startup until we get a useable response.
   - If the response is newer than the what we have, update our internal state.
1. If an address book re-query interval is configured (and MN URL is configured), query MN every interval to get the latest address book.
   - If the response is newer than the what we have, update our internal state.

That makes decisioning a little simpler (local file if configured, mirror node if configured, interval if configured) because all three options are completely independent of each other.

Not sure if that's enough better to be worth changing at this point; just a suggestion, really.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
```diff
@@ -240,12 +240,23 @@ public List<CompletableFuture<Void>> executeAsync(
 
     /// Move all current scheduled tasks to the work queue for execution.
     /// Any non-recurring tasks are then removed from the work map, but
-    /// recurring tasks remain to be executed again on the next cycle
+    /// recurring tasks remain to be executed again on the next cycle.
+    ///
+    /// For repeating schedules the raw Runnable is enqueued directly rather than
+    /// the ScheduledTask (FutureTask) wrapper, because FutureTask.run() is a
+    /// one-shot operation and silently becomes a no-op after its first execution.
```
jsync-swirlds, line 247: hmmm, this is less than ideal...
The point of using ScheduledTask was to avoid this kind of "extract a runnable" action.
This works for now, but it would be good to find a more elegant resolution in the future.

jsync-swirlds (approved):

Nana-EC (approved):
LG.
My expectation is BN loads an initial address book either from local file if present or from a MN is reachable.
After that regardless of initial source the MN can be used to update the initial address book if there's a newer addressBook

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/async/ScheduledBlockingExecutor.java
berryware: Captured in [#2900](https://github.com/hiero-ledger/hiero-block-node/issues/2900)

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: Captured in [#2901](https://github.com/hiero-ledger/hiero-block-node/issues/2901)

PR #2863, opened 2026-05-20
chore: add RSA bootstrap roster support

jan-milenkov:
## Description

Adds RSA bootstrap roster support for block node deployments.The block-node Helm chart now supports receiving an RSA bootstrap roster through chart values, writing the roster file during pod initialization, and exposing the configured file path through `APP_STATE_RSA_BOOTSTRAP_FILE_PATH`.This enables Solo to provide RSA bootstrap data directly during deployment without relying on Mirror Node APIs.

## Changes

- Added `blockNode.rsaBootstrap` chart values.- Added init container support for writing the RSA bootstrap roster file.
- Added support for configuring `APP_STATE_RSA_BOOTSTRAP_FILE_PATH`.
- Mounted the generated roster file into the block-node container.
- Added conditional handling so deployments without RSA bootstrap configuration continue to behave as before.

## Validation

Manually verified using solo that:

- Helm receives the generated roster values.
- The block-node pod includes the RSA bootstrap init container.
- The roster file is written to the pod.
- `APP_STATE_RSA_BOOTSTRAP_FILE_PATH` points to the generated roster file.

## Related Issue(s)

- Requirement for # https://github.com/hiero-ledger/solo/issues/4426

### DIFF
--- a/charts/block-node-server/templates/statefulset.yaml
+++ b/charts/block-node-server/templates/statefulset.yaml
@@ -32,6 +32,9 @@ spec:
       serviceAccountName: {{ include "hiero-block-node.serviceAccountName" . }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      {{- if and .Values.blockNode.rsaBootstrap.enabled (not .Values.blockNode.rsaBootstrap.rosterBase64) }}
+        {{- fail "blockNode.rsaBootstrap.enabled is true but rosterBase64 is empty" }}
+      {{- end }}
       {{- if .Values.hostNetwork }}
       hostNetwork: true
       dnsPolicy: {{ default "ClusterFirstWithHostNet" .Values.dnsPolicy }}
@@ -77,7 +80,6 @@ spec:
         {{- with .Values.blockNode.initContainers }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-
         {{- if .Values.blockNode.rsaBootstrap.enabled }}
         - name: write-rsa-bootstrap-roster
           image: {{ .Values.blockNode.rsaBootstrap.image | quote }}
@@ -96,7 +98,6 @@ spec:
             - name: {{ default "verification-storage" .Values.blockNode.persistence.verification.volumeName }}
               mountPath: {{ .Values.blockNode.rsaBootstrap.mountPath | quote }}
         {{- end }}
-
         {{- if .Values.plugins.names }}
         - name: copy-core-modules
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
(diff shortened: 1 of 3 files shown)


charts/block-node-server/templates/statefulset.yaml
```diff
@@ -199,80 +219,80 @@ spec:
...
+            periodSeconds: {{ .Values.blockNode.health.liveness.periodSeconds }}
+            timeoutSeconds: {{ .Values.blockNode.health.liveness.timeoutSeconds }}
+            successThreshold: {{ .Values.blockNode.health.liveness.successThreshold }}
+            failureThreshold: {{ .Values.blockNode.health.liveness.failureThreshold }}
+          readinessProbe:
+            httpGet:
+              path: {{ .Values.blockNode.health.readiness.endpoint }}
+              port: {{ .Values.service.port }}
+            initialDelaySeconds: {{ .Values.blockNode.health.readiness.initialDelaySeconds }}
+            timeoutSeconds: {{ .Values.blockNode.health.readiness.timeoutSeconds }}
```
AlfredoG87: why the change in indentation. when trying to: `kubectl apply --dry-run=server` rejects this:
```
 error: unknown field "spec.template.spec.containers[0].limits", unknown field "spec.template.spec.containers[0].requests"
```

I believe is related to changes on indentation.


  The container block was reindented from 8 spaces to 10, but {{- toYaml . | nindent 10 }} was left unchanged. It must become nindent 12 to match the new parent indentation. Reproducible by rendering the chart and running kubectl apply 
  --dry-run=server -f -. This breaks every default deployment regardless of rsaBootstrap.enabled.

However, we should not touch anything that is out of the scope (such as refformatting the entire containers block)

charts/block-node-server/templates/statefulset.yaml
```diff
@@ -72,11 +72,31 @@ spec:
...
+            - sh
+            - -c
+            - |
+              mkdir -p {{ .Values.blockNode.rsaBootstrap.mountPath }} && \
+              printf "%s" "$RSA_BOOTSTRAP_ROSTER_BASE64" | base64 -d > {{ .Values.blockNode.rsaBootstrap.filePath }} && \
+              chown 2000:2000 {{ .Values.blockNode.rsaBootstrap.filePath }} && \
+              chmod 0444 {{ .Values.blockNode.rsaBootstrap.filePath }}
+          env:
+            - name: RSA_BOOTSTRAP_ROSTER_BASE64
+              value: {{ .Values.blockNode.rsaBootstrap.rosterBase64 | quote }}
```
AlfredoG87, line 96: If enabled: true but rosterBase64: "", the init container silently writes an empty file, should we have a fail fast error message? 
```
 {{- if and .Values.blockNode.rsaBootstrap.enabled (not .Values.blockNode.rsaBootstrap.rosterBase64) }}
    {{- fail "blockNode.rsaBootstrap.enabled is true but rosterBase64 is empty" }}
  {{- end }} 

```

charts/block-node-server/values.yaml
```diff
@@ -142,6 +142,30 @@ blockNode:
...
+
+    # Base64-encoded rsa-bootstrap-roster.json content.
+    rosterBase64: ""
+
+    # Path where the init container writes the decoded roster file.
+    # This path is also exposed to the block node through APP_STATE_RSA_BOOTSTRAP_FILE_PATH.
+    filePath: "/opt/hiero/block-node/verification/rsa-bootstrap-roster.json"
+
+    # Existing mounted verification PVC path used to persist the generated file.
+    mountPath: "/opt/hiero/block-node/verification"
```
AlfredoG87: These filepaths don't match the expected default path on BN App configuration.

https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java#L26-L29

AlfredoG87 (changes requested):

charts/block-node-server/templates/statefulset.yaml
jan-milenkov: updated, it was caused by `.editorconfig`

charts/block-node-server/templates/statefulset.yaml
jan-milenkov: updated

charts/block-node-server/values.yaml
jan-milenkov: updated

AlfredoG87:
Not Needed anymore

berryware:
@AlfredoG87, Why is this not needed anymore? Does the solo team need to do any other work for this? Is this covered by the WRB CLI?

PR #2869, opened 2026-05-21
feat: Verification Plugin Rework

### LINKED ISSUE #1582: Validation Plugin Should Manage Block Order
### Story Form

As a Block Node Engineer
I want to ensure blocks are processed in order
So that I can assert critical invariants about blocks

### Technical Notes

The validation plugin is in unique position to ensure ordering of validation events for blocks.  We intend to add minimal order checking to the validation plugin to ensure that validation success events are ordered for the newest/latest blocks.
Basic concept:
```text
if the block is >= the latest validated block
    if the block is the _next_ block in order
        Send validation event
    else
        Buffer (if recent) or fail validation (if too old or too far in the future)
else
  Send validation event based on further checks
```

This should only apply for Live Flow of verification.

Other Notifications that request work (Backfill or other plugins) _may_ not need to verify in order (e.g. backfill of history covered by a jumpstart data file would not need to be verified in order).

### LINKED ISSUE #2449: Limited Buffer to Support In-Order Block Verification
### Story Form

As a Verification Plugin
I want to have a limited buffer to store incoming blocks
So that I can collect and verify blocks in order

### Technical Notes

- Verification plugin should have a limited size buffer to collect incoming blocks
- The buffer size should be configurable
- This buffer will allow the plugin to temporarily store incoming blocks in order to ensure that blocks are verified in order
- There are multiple sources from where a block could come from
  - Two main ones are from a publisher and from backfill
- Sources have priority

### LINKED ISSUE #2527: Comprehensive Error Handling for Verification
### Story Form

As a Verification Plugin
I want to have robust, sophisticated and comprehensive error handling
So that I can confidently respond and continue normal operations

### Technical Notes

- Verification Plugin's error handling should be very robust, comprehensive and sophisticated
- We must be able to catch any and all failures
- We must be able to respond properly to different kinds of failures, for instance: failed verification, failed block parsing etc. so that interested parties can take appropriate action(s)
- The plugin should never throw, but handle all kinds of failures
- With the addition of a more comprehensive and extended health facility, we will also be able to signal unhealthy state to the facility

### LINKED ISSUE #2528: Improved and Comprehensive Verification Result
### Story Form

As a Verification Plugin
I want to have a comprehensive verification result
So that I can then process it and send the correct message or take an appropriate action

### Technical Notes

- Each verification attempt should return a comprehensive result
- This result should contain all data needed to construct a meaningful message based on the result, or to be able to take an appropriate action
- The result should be able to discern successful verification attempts as well as different types of failure
- The result should contain block source
- The result should contain the block data in case of success

### LINKED ISSUE #2609: Utilize `FailureType` for Verification Notifications
### Story Form

As a Verification Plugin
I want to give a descriptive reason as to why a verification has failed
So that interested plugins can take appropriate actions

### Technical Notes

**DEPENDS ON**: #2529 

- we've recently extended the verification notification to include a failure type
- we now have to use the failure types appropriately

ata-nas:
* Complete Verification Plugin Overhaul:
  * Verification Plugin now receives data in any order and starts a new session for each new block received.
* Verification Sessions are now async:
  * A session consists of multiple stages (units of work) that are chained.
    * Block Hasher - the first stage, accepts the raw data and produces a root hash. Fails fast in case of any issues during hashing.
    * Block Verifier - the second stage, accepts the hashing result of the previous stage and proceeds to do verification on block proofs of the block.
    * Block Ordering Manager - the third stage, accepts the verification result from the previous stage and awaits its order to complete. Successful verificaitons must complete in order so that we can comply with a strict ordering guarantee. Failures are immediate and must not wait here.
    * Verification Result Handler - the last stage, accepts the verification result produced from the chain before and proceeds to send the correct notification to the internal messaging system. Finally, it increments the last verified block flag wich the ordering stage uses.
* Active Sessions Buffer:
  * A configurable buffer of active sessions.
  * When new data comes in and the buffer is full, oldest sessions will be canceled to make room for new ones

Related Issues:
Resolves #2528
Resolves #2527
Resolves #2449
Resolves #1582
Resolves #2609


block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/TSSParameters.java
jsync-swirlds, line 1: Why do we not have the WRAPS verification key here?

ata-nas: For now, this is a pure copy of what the previous plugin did. Let's discuss this.

ata-nas:
Opening for review. Even with cleanup due.

_Note_ this code does not yet switch over from the current verification plugin.

Since this is already massive, tests will be a follow up.
Let's collect feedback to address in a follow up.

Notable things to do:

- evaluate changes (reviewers)
- utilize TssData from app state facility instead of a local implementation (to be added)

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/HashingResult.java
```diff
@@ -0,0 +1,48 @@
...
+        Objects.requireNonNull(rootHash, "Block root hash cannot be null");
+        Objects.requireNonNull(blockHeader, "Block header cannot be null");
+        Objects.requireNonNull(blockFooter, "Block footer cannot be null");
+        Objects.requireNonNull(hapiProtoVersion, "HapiVersion cannot be null");
+        if (blockNumber < 0) {
+            throw new IllegalArgumentException("Block number cannot be negative");
+        }
+        if (blockProofs == null || blockProofs.isEmpty()) {
+            throw new IllegalArgumentException("Block proofs cannot be null or empty");
+        }
```
ata-nas: This check should simply be a null check. If the collection is empty, next stage will take action upon that.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/HashingResult.java
```diff
@@ -0,0 +1,48 @@
...
+import org.hiero.block.node.spi.blockmessaging.BlockSource;
+
+/// A result returned after a successful hashing of a block.
+/// @param blockNumber the number of the block, cannot be negative
+/// @param blockSource the source of the block, cannot be null
+/// @param block the whole block, cannot be null
+/// @param rootHash the root hash of the block, cannot be null
+/// @param blockHeader the parsed header of the block, cannot be null
+/// @param blockFooter the parsed footer of the block, cannot be null
+/// @param blockProofs all block proofs of the block, parsed, cannot be null or empty
```
ata-nas:
```suggestion
/// @param blockProofs all block proofs of the block, parsed, cannot be null
```

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/TSSParameters.java
ata-nas: We are now using `TssData`

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/VerificationSessionFailedException.java
```diff
@@ -0,0 +1,69 @@
...
+    private final long blockNumber;
+    private final SessionFailureType failureType;
+    private final BlockSource blockSource;
+
+    public VerificationSessionFailedException(
+            final long blockNumber, final SessionFailureType failureType, final BlockSource blockSource) {
+        this.blockNumber = blockNumber;
+        this.failureType = failureType;
+        this.blockSource = blockSource;
+        super(DEFAULT_MESSAGE.formatted(blockNumber, blockSource, failureType));
```
Nana-EC, line 19: `super(...)` should be first right?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -0,0 +1,236 @@
...
+    public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
+        this.context = context;
+        this.verificationConfig = context.configuration().getConfigData(VerificationConfig.class);
+        // todo(2528) Bootstrap TSS parameters from persisted file if available. The file contains a
+        // serialized LedgerIdPublicationTransactionBody with ledger ID, address book, and WRAPS VK.
+        this.executor = context.threadPoolManager()
+                .getVirtualThreadExecutor(
+                        "VerificationSession", VerificationServicePlugin::getUncaughtExceptionHandler);
+        // todo(2528) initialize tss parameter before passing them
+        this.sessionHandler = new BlockSessionHandler(
```
Nana-EC, line 84: `metricsHolder` is still null currently
```suggestion
        // initialize metrics
        this.metricsHolder = MetricsHolder.create(context.metricRegistry());
        this.sessionHandler = new BlockSessionHandler(
```

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/AllBlocksHasherHandler.java
```diff
@@ -292,12 +288,17 @@ private byte[] calculateBlockHashFromBlockNumber(long blockNumber) throws ParseE
...
+                            //                            final VerificationSession session =
+                            // HapiVersionSessionFactory.createSession(
+                            //                                    blockNumber, BlockSource.HISTORY,
+                            // blockHeader.hapiProtoVersion(), null, null, null);
+                            //                            final VerificationNotification result =
+                            // session.processBlockItems(blockItemsMessage);
+                            //                            if (result != null && result.success()) {
+                            //                                blockHash = result.blockHash().toByteArray();
+                            //                            }
+                            throw new UnsupportedOperationException("need to use the new sessions");
```
Nana-EC, line 301: Q: is this the final form or should we be wiring it up?
There's commented out code and the comment says todo(2528) which this PR addresses so it's unclear

block-node/block-verification/build.gradle.kts
```diff
@@ -17,6 +17,6 @@ mainModuleInfo {
 testModuleInfo {
     requires("org.junit.jupiter.api")
     requires("org.hiero.block.node.app.test.fixtures")
-    requires("org.mockito")
-    requires("org.junit.jupiter.params")
+    // requires("org.mockito")
+    // requires("org.junit.jupiter.params")
```
jsync-swirlds, line 21: Let's go ahead and remove these two if we can.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
```diff
@@ -37,9 +38,32 @@ public enum FailureType {
         MISSING_MANDATORY_ITEM,
         /// This type indicates that the block is missing a mandatory field (null)
         MISSING_MANDATORY_FIELD,
+        /// Missing RSA Roster Keys
+        MISSING_RSA_ROSTER_KEYS,
+        /// This type indicates missing TSS parameters
+        MISSING_TSS_PARAMETERS,
```
jsync-swirlds: Can we combined these into a single "MISSING_VERIFICATION_DATA"?
That would prevent needing to change this enum when we add other proof types.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
```diff
@@ -37,9 +38,32 @@ public enum FailureType {
...
+        UNSUPPORTED_HAPI_VERSION,
         /// This type indicates that the session was cancelled
         CANCELLED,
-        /// This type indicates that an unknow error occurred
+        /// This type indicates that an unknown error occurred
         UNKNOWN_ERROR
     }
+
+    /// todo(2528) add documentation
+    public record FailureInfo(FailureType failureType, boolean isInformational) {
```
jsync-swirlds, line 59: This is a little odd to me.
Why are we adding another record inside the Notification instead of just adding one field to the enclosing record?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/ResultOrderingManager.java
```diff
@@ -0,0 +1,41 @@
...
+    private final AtomicBoolean isCanceled;
+
+    public ResultOrderingManager(final AtomicBoolean isCanceled, final AtomicLong lastVerifiedBlock) {
+        this.isCanceled = Objects.requireNonNull(isCanceled);
+        this.lastVerifiedBlock = Objects.requireNonNull(lastVerifiedBlock);
+    }
+
+    @Override
+    public BlockVerificationResult apply(final BlockVerificationResult blockVerificationResult) {
+        while (!isCanceled()) {
```
Nana-EC, line 31: Are we deadlocking ourselves here?
If the `blockVerificationResult.blockNumber() > nextExpectedBlock` remains true this will loop indefinitely unless there's an external trigger to cancel!
There's missing logic to ensure this scenario eventually is invalidated, at worst the `while` should have a timeout but it's likely better to have a state condition instead of a time based approach if possible.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
```diff
@@ -0,0 +1,190 @@
...
+                true,
+                null,
+                verificationResult.blockNumber(),
+                verificationResult.blockHash(),
+                verificationResult.block(),
+                verificationResult.source());
+        sendNotification(notification);
+        // todo(2528) ensure the correct CAS and sequence of actions
+        if (verificationResult.blockNumber() > lastVerified) {
+            final boolean incremented = lastVerifiedBlock.compareAndSet(lastVerified, lastVerified + 1);
```
Nana-EC, line 194: Would it be more correct to use `verificationResult.blockNumber()`.
Is there's a chance `lastVerifiedBlock` can be modified in unexpected ways the increment may become unexpected
```suggestion
            final boolean incremented = lastVerifiedBlock.compareAndSet(lastVerified, verificationResult.blockNumber());
```

Nana-EC (changes requested):
Lots of good changes.
1st pass, left some comments, will circle back again

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/VerificationSessionFailedException.java
jsync-swirlds: [JEP 513](https://openjdk.org/jeps/513) makes that rule more flexible.
"Safe" code is now allowed before the superclass call.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/ResultOrderingManager.java
jsync-swirlds: There is external trigger to cancel.
When new blocks come in, the session handler checks the count of pending sessions and cancels the oldest one if a limit is exceeded.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
jsync-swirlds: The block number needs to always increment 1 block at a time; otherwise we could jump ahead (i.e. after block 5 is complete, block 20 passes the CAS) and that might cause a number of failures for the block numbers skipped.

This could happen if block 20 were very small and blocks 6-19 were all quite large, or if a resend caused a group of blocks to "stack up", because the threads complete in undefined order.

We might need a more careful check to ensure _this block_ is the next in line, and not some later block.

jsync-swirlds (commented):
Only about 1/4 through, but posting what I have so far.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/VerificationSessionFailedException.java
ata-nas: Semantically, this reads better. That is why I've done super at the bottom.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: Nice catch! Metrics tests, which I haven't gotten to yet, would have caught that, but glad we have caught it before :)

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/AllBlocksHasherHandler.java
ata-nas: So, this was just to compile the code. The all blocks hasher is completely disabled for now.

I've left this on purpose because we will move this out of here (verification plugin) in the future. We will have to then think about how to properly rebuild from storage, as moving this out, will not have the ability to directly invoke verification specific code, as it has been doing until now.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/ResultOrderingManager.java
ata-nas: We cannot really do a time-based timeout/cancel. And yes, we have a configurable buffer of maximum active sessions. If that gets full, we will, externally, start to cancel sessions. Until then, we cannot proceed forward as it is not our turn yet.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
ata-nas: Yes, we can certainly explore the CASing logic here, hence the todo.

But the reasoning here is we only want to increment on success, if the successful block is higher than our high water mark. Now, that block which will come to this increment logic, should theoretically be exactly the next one, for instance if last verified was 50, we expect block 51 to be the only possible block to reach this CAS and be successful! If we have two or more active sessions for block 51 that complete at the same time, they can simply race and it is irrelevant which one wins the CAS.

Mind you, there is a caveat (we have to add that). With the addition of application state facility's stored blocks ranged set, we can receive runtime updates! This means that last verified could be at 50, but then we see highest stored is actually 100, for instance. In those cases, I think we have to set the last verified to 100, as nothing can be persisted w/o first being verified. That also means that any queued up sessions between 52 and 100 will immediately complete in any order, but that is OK, because we know that we do not need to keep an order on those anymore. That being said, this CAS can (should) only, and only, fail if an update for a latest persisted block arrives, which is higher than the value we have locally stored. In those cases, we know we have moved forward. 

Finally, if we have a block that is below or equal to the last verified block, it will simply send the success and not even attempt the CAS. For such blocks, no ordering on success is enforced.

The ordering manager uses this atomic value to determine if the current block should wait or go forward.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
ata-nas: Yes, we can. I was wondering if there was a benefit to differentiating between these.

Also, I think that I will use that combined failure type for NoSuchAlgorithmException as well. Kind of fits the narrative.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
ata-nas: Right, so I had to add the informational failure type. Initially, that was going to be a prefix to the existing failure types. For instance, `BAD_BLOCK_PROOF` would have its pair - `INFO_BAD_BLOCK_PROOF`. The reason for that is so that users are intentionally forced to handle informational failures and not skip them.

Then, another possibility is to simply add another boolean to the notification. It would be set to true in case of a failure that is informational. That is bad, however, because we will always see a boolean value on success, failure and is kind of error prone for users. It could easily be skipped. 

I decided to go with a failure info record, as that contains all information about the failure we need, currently a type and if it is informational. In this way, if we are observing success, this info record will be null.

My last motivation for this, is that if we ever need to have more data added to the info record, we can easily do that without breaking the world, as a simple change in the verification notification takes significant effort to accomplish.

block-node/block-verification/build.gradle.kts
ata-nas: Yes, will address in tests PR, which is a follow up. Mockito I am 99.9% certain will be absolutely dropped. Will do my best not to use it at all.

rockysingh (commented):
Left a few suggestions

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
```diff
@@ -0,0 +1,199 @@
...
+                recentlyVerifiedBlocks,
+                executor,
+                isHandlerShuttingDown,
+                context);
+    }
+
+    /// Activate a new session.
+    /// This method will activate the session and potentially cancel the oldest one if the buffer for
+    /// maximum allowed sessions is full.
+    private void activateSession(final BlockVerificationSession session) {
```
rockysingh, line 155: Session eviction uses session ID instead of block number.

activeSessions is keyed by session ID, so `pollFirstEntry()` removes lowest ID, not oldest block.

Example failure: blocks 100,101,102 (sessions 1,2,3), then block 50 (session 4) arrives. Evicts session 1 (block 100) instead of session 4 (block 50). Key by block number or use custom comparator?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/CompletableVerificationSession.java
```diff
@@ -0,0 +1,139 @@
...
+                blockSource);
+        final CompletableFuture<BlockVerificationResult> completionChain = CompletableFuture.supplyAsync(
+                        hasher, executor)
+                .thenApply(verifier)
+                .thenApply(resultOrderManager);
+        completionChain.whenComplete(sessionResultHandler);
+        sessionCompletionChain = completionChain;
+    }
+
+    @Override
```
rockysingh, line 118: Race condition: `sessionCompletionChain` read without synchronization between start() and cancel().

If `cancel()` runs after `start()` begins but before `sessionCompletionChain` assignment, throws `IllegalStateException`.

Set `isCancelled` flag first, then cancel chain if non-null without throwing.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -0,0 +1,379 @@
...
+            final BlockSource blockSource,
+            final TssData currentTssData) {
+        this.context = Objects.requireNonNull(context);
+        if (blockNumber < 0) {
+            throw new IllegalArgumentException("Block number must be non-negative");
+        }
+        this.blockNumber = blockNumber;
+        this.hashingMetrics = Objects.requireNonNull(hashingMetrics);
+        this.blockSource = Objects.requireNonNull(blockSource);
+        this.currentTssData = currentTssData; // can be null @todo(2528) see if we actually need it here
```
rockysingh: TSS data captured by value at session creation. If TSS arrives after session starts but before verification stage, verifier fails with `MISSING_TSS_PARAMETERS`.

Scenario: Block N starts (TSS=null) -> hashing done -> TSS update arrives -> verification fails with stale null.

Maybe pass as `Supplier<TssData>` or `AtomicReference<TssData>`.

rockysingh (commented):
Code review with inline comments on critical issues.

rockysingh (changes requested):

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
jsync-swirlds: I completely missed that.
You are correct, we should be using a two-part key of [block number, session ID] that implements Comparator with compare to comparing block number first and session ID only if block number is the same.

ata-nas: That's the whole point I added ids. So that I can make the comparator first by block number then by id. I had a todo for it which I seem to have lost... Good catch.

P.S. we cannot use block number as a key. We can totally allow multiple sessions for the same block simultaneously.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Same as with the RSA keys map.
If it is unavailable when the session is created, we fail. But that should be acceptable.
We can further discuss this.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/CompletableVerificationSession.java
ata-nas: This is very tricky.

The isCancelled flag is very deliberately set to true after calling cancel.
Completable future does not interrupt a thread. That is why we need the flag.
But calling cancel will actually go through the when complete stage on the caller thread.
Then, we have to be able to stop the background thread.
The reason for throwing when start is not called before cancel is to ensure we do not do it.
Then, by having the volatile read guarantees, we can assert that the session was started or not.
We can discuss this further.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
ata-nas: Funny, my todo for this is actually there, just on the field 😆

ata-nas: ✅ Done, a `SessionKey` object is now used.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
ata-nas: I am thinking, shall we keep the different missing verification data types?
In theory, the rsa and tss bootrstap plugins could listen to these to try to speed up a data fetch?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/ResultOrderingManager.java
ata-nas: What we need to think about is how to remove completed sessions from the active sessions map.
This could happen just before we put new sessions in the map (we have just received new data regardless of the source), or the sessions, when complete (terminating stage of completion chain) remove themselves from the map.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: ✅ Done, updated.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Testing this today. I think I've hit the case you mention. Dynamic updates (ones initiated during the hashing process) will not be received on time if done by the same session. And this is not the same as RSA keys map, where we do not have this situation.
Issue is we do want to update tss data via the app state facility. If we keep local copy as well, not sure about that. But that actually might be what we need for this one. Will have to explore a little bit more.

What we might do is to pass the most recent TssData optionally as the hashing result. Then, the verifier can use those for the current session. We do also send that update to the application state facility as well, however, so that it will be updated plugin level. This, however, is still somewhat fragile. 

Another approach, will be to completely ignore application state updates (called from the facility) and always rely on the most recent local find. We still keep sending updates to the facility, however. That might not be applicable cause I think the tss plugin will also fetch data from other sources.

Interesting thing to solve.

For RSA keys, we can also have an atomic ref that verifier will check for. But that one is easier cause updates will always be from a single place. They are either present or not.

ata-nas: OK. Did a quick change here.

1. A session starts with whatever TssData is available at the moment, could be null.
2. During hashing, a new TssData update could be made. If so, we go ahead and push that to the app state facility.
3. The hashing result now contains the most recent TssData. This means either from the start of the session received externally or from the hashing process. The tss data is in any case made available to the verifier via the hashing result.
4. In the meantime, the app state facility will make the update to the plugin itself and other sessions, when started, will possibly see this new update.

This does not cover the case if block 0 for instance initializes or updates tss data, then block 1 or higher cannot see that immediate update. They could be running in the background.
Another thing to think about is what if we have 2 or more sessions for block 0? That should not be a problematic update to do a couple of times in app state facility?

ata-nas: We still might consider another option which will be better. The atomic ref proposed might be better.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
jsync-swirlds: I don't think we really want to be listening to verification failures in those plugins; and I doubt we have time to improve those plugins before we have to finalize things.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockVerificationSession.java
```diff
@@ -0,0 +1,57 @@
...
+    record SessionKey(long blockNumber, long uniqueId) implements Comparable<SessionKey> {
+        @Override
+        public int compareTo(final SessionKey o) {
+            final int blockNumberCompare = Long.compare(this.blockNumber, o.blockNumber);
+            if (blockNumberCompare != 0) {
+                return blockNumberCompare;
+            } else {
+                return Long.compare(this.uniqueId, o.uniqueId);
+            }
+        }
```
jsync-swirlds: This is an expensive way to do this, you're allocating several objects when you could just use conditional operators and simple primitive comparisons.

```suggestion
            if (this == o) {
                return 0;
            } else if (o == null) {
                return 1; // null comes before non-null.
            } else {
                // spotless double indents makes this unreadable
                // spotless:off
                return blockNumber != o.blockNumber ? blockNumber > o.blockNumber ? 1 : -1
                    : uniqueId == o.uniqueId ? 0 : uniqueId > o.uniqueId ? 1 : -1
                // spotless:on
        }
```

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/CompletableVerificationSession.java
jsync-swirlds: I wonder if this is a rare case when a lock (likely ReentrantLock member of `CompletableVerificationSession`) is appropriate?

Lock beginning of `start()`, unlock in finally at very end.
Lock beginning of `cancel()`, unlock in finally at very end.
That ensures only one of the two can run at a time, and these _should_ be rare enough (one each per block) that the cost of the lock is trivial.

Just some food for thought.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/AllBlocksHasherHandler.java
jsync-swirlds: I've been thinking on this, and if we make the hashing depend on verification success notifications, and just receive the block hash from there (assuming it's in the notification, we might want to add it) then this functionality becomes dead simple (a trivial plugin that maintains something like an RRD of hashes and can provide a tree path, linear path, or combination, on demand).

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
```diff
@@ -0,0 +1,73 @@
...
+                result = SessionFailureType.MISSING_TSS_PARAMETERS;
+            } else {
+                // TSS.verifyTSS() handles both the genesis (Schnorr aggregate) and post-genesis (WRAPS) paths.
+                // Signatures without a recognized proof suffix are rejected by the library.
+                // todo(2528) check for currentTssData.validFromBlock()? Fail in case?
+                try {
+                    if (!TSS.verifyTSS(
+                            currentTssData.ledgerId().toByteArray(),
+                            signature.toByteArray(),
+                            hashToVerify.toByteArray())) {
```
jsync-swirlds: Where are we setting the WRAPS verification key?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -0,0 +1,152 @@
...
+            } else if (!stateProof.hasSignedBlockProof()) {
+                LOGGER.log(WARNING, "Block {0} state proof is missing signed block proof", blockNumber);
+                result = SessionFailureType.BAD_BLOCK_PROOF;
+            } else {
+                final List<SiblingNode> siblings = siblingPath.siblings();
+                final int totalSiblings = siblings.size();
+                if (totalSiblings < 3 || (totalSiblings - 3) % 4 != 0) {
+                    LOGGER.log(
+                            WARNING,
+                            "Block {0} state proof sibling count {1} is invalid (need >= 3, remainder must be multiple of 4)",
```
jsync-swirlds, line 69: The remainder multiple of 4 is false here.
Remainder can have almost any count, it is a _bug_ that we assumed only one possible path option and ignored the possibility to use the _tree_ instead of the chain path.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -0,0 +1,152 @@
...
+                        final byte[] depth3Node1 = combineSibling(depth4Node1, depth4Node2Sibling);
+                        final byte[] depth2Node2 = hashInternalNodeSingleChild(depth3Node1);
+                        current = hashInternalNode(hashedTimestampSibling.hash().toByteArray(), depth2Node2);
+                        if (firstIteration) {
+                            final Bytes reconstructed = Bytes.wrap(current);
+                            if (!rootHash.equals(reconstructed)) {
+                                LOGGER.log(
+                                        WARNING,
+                                        "Block {0} state proof integrity check failed: hash reconstructed from path 1 siblings"
+                                                + " [{1}] does not match block root hash computed from block content [{2}]",
```
jsync-swirlds, line 109: This seems off.
The block being proved is one of the _inputs_ to the hash (it's the leaf being proven), so we cannot reconstruct it's hash from the sibling paths.
The signed root hash is the hash of a _different_ block; we don't have that block and don't know its root hash, we use the block-being-proven root hash, and the sibling hashes, to compute a root hash, then check the signature validly signed _that_ root hash.

There is no point where this rootHash.equals(reconstructed) could possibly be true (also, reconstructed is not accurate; it's just a "computed" hash).

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -0,0 +1,382 @@
...
+    /// [HashingResult] is returned.
+    /// @throws VerificationSessionFailedException in case a known failure occurs
+    @Override
+    public HashingResult get() {
+        try {
+            final long hashingStartTime = System.nanoTime();
+            while (!isCanceled()) {
+                final BlockItems blockItemsRecord = blockItemsRecordsDeque.poll();
+                if (blockItemsRecord == null) {
+                    LockSupport.parkNanos(TimeUnit.MICROSECONDS.toNanos(200));
```
jsync-swirlds: You might want to assign `TimeUnit.MICROSECONDS.toNanos(200)` to a constant so we aren't making this calculation every time we reach this line.

jsync-swirlds (commented):
Apologies it took so long to get to this.
A few things here we should look into.

Also, a few of your own comments you might need to resolve...

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/CompletableVerificationSession.java
ata-nas: I wonder what situation we would see where cancel and start will be called simultaneously. When we receive a block and need to start a session, we create it, then start it and only then do we activate it (put it in the active sessions map) where we have the only possible place where call cancel on it. It is possible that I might be missing something.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/AllBlocksHasherHandler.java
ata-nas: That's a cool idea. I like the simplicity of this. We have to, however, completely extract the all blocks hasher from verification so it does not hit on performance at all. For the all blocks hasher, we are only interested in the root hash and nothing else.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
ata-nas: I will go with a single consolidated failure type for simplicity reasons. We can always change this easiliy.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
ata-nas: That happens when we do `safeUpdateTssData`, see verification plugin class.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Ok, after further discussions, this has been changed again.
I opted to create a simple verification data provider to consolidate rsa public keys and tss data in a single place. This provider is updatable and sees the newest.
Please take another look at this.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
ata-nas: ✅ Done.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/HashingResult.java
ata-nas: ✅ Done.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/HashingResult.java
ata-nas: ✅ Done.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockVerificationSession.java
ata-nas: Good note on null safety and reference comparison. Added that.

About the rest, I think nested ternary operators are very unreadable and error prone.
Also, what object allocation happens? I only save an `int` value temporarily so I do not re-calculate it again. Internally, `Long.compare` does not do any object allocation.

I think the below is more readable:
```suggestion
            if (this == o) {
                return 0;
            } else if (o == null) {
                return 1; // null comes before non-null.
            } else {
                final int blockNumberCompare = Long.compare(this.blockNumber, o.blockNumber);
                if (blockNumberCompare != 0) {
                    return blockNumberCompare;
                } else {
                    return Long.compare(this.uniqueId, o.uniqueId);
                }
            }
        }
```

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
ata-nas: Setting WRAPS keys now happens inside an update done to a verification data provider class.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: ✅ Good point, done.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/CompletableVerificationSession.java
jsync-swirlds: Depends on which threads call start and cancel, if the "stalled" session handling is in any thread other than the messaging thread then cancel could be called in the middle of the start method. It's pretty low cost to have the lock in this case; so it might be worthwhile just for the belt-and-suspenders value.

ata-nas: Regardless of the thread, you will not be able to cancel an active session before it is started because you will not see it in the map before it starts.

Now, I just happen to see another caveat. That is the current active live data session. That is also saved as a field so if we drop that for some reason, we have to flag for the live data RB thread. Will have to explore that for a bit.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockVerificationSession.java
jsync-swirlds: When possible I prefer to _not_ use comparison methods for primitives (because primitive values always compare directly).
That's why I prefer the 2 line ternary over multiple calls to `Long.compare`.

That's preference, not correctness, however.
Your version will work.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/CompletableVerificationSession.java
jsync-swirlds: After a discussion, Atanas has shown that start and cancel cannot be called in a sequence that could encounter this edge case.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockVerificationSession.java
AlfredoG87: I agree with Atanas on readability 👍

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: This is related to a comment that you made in the past: https://github.com/hiero-ledger/hiero-block-node/pull/2704#discussion_r3212294883

I thought I had a todo in the code, but the todo is in an offline list.
So is the correct check here simply `totalSiblings < 3` if so fail?
I base this on your comment.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: Not sure what to do here.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/ResultOrderingManager.java
ata-nas: We now have removal of completed sessions.

rockysingh (dismissed):

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
jsync-swirlds: I wouldn't bother checking the number of siblings at all.
The root hash won't be correct if we do not have the correct number of siblings, and we don't save much by checking the number in advance.
Truthfully, a state proof can have _almost_ any number of siblings and still be a valid proof, so the safest option is just build the root hash then verify it.

(we might put an _upper_ limit someday, but that's not necessary at the moment)

Nana-EC (dismissed):
Unblocking

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
jsync-swirlds: I think this entire set of checks related to a "reconstructed" hash is just wrong, it will never pass for a real state proof.

I believe the whole set is based on a complete misunderstanding of how a state proof works (assuming I'm reading the code correctly).

recommended changes:
1. remove any concept of a "reconstructed" hash; that doesn't exist.
1. remove any comparison between the block root hash and anything computed from the state proof
1. The first item combined with the first (lowest in the tree) sibling is the block root hash to be proven.  The end result is a root hash (for a different block, one we do not have!) that is passed to the signature verification.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/HashingResult.java
```diff
@@ -0,0 +1,48 @@
...
+        Objects.requireNonNull(block, "Block cannot be null");
+        Objects.requireNonNull(rootHash, "Block root hash cannot be null");
+        Objects.requireNonNull(blockHeader, "Block header cannot be null");
+        Objects.requireNonNull(blockFooter, "Block footer cannot be null");
+        Objects.requireNonNull(hapiProtoVersion, "HapiVersion cannot be null");
+        if (blockNumber < 0) {
+            throw new IllegalArgumentException("Block number cannot be negative");
+        }
+        if (blockProofs == null) {
+            throw new IllegalArgumentException("Block proofs cannot be null or empty");
```
ata-nas, line 45: Nit: 
```suggestion
            throw new IllegalArgumentException("Block proofs cannot be null");
```

jsync-swirlds (approved):

Nana-EC (approved):

rockysingh (approved):

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/ResultOrderingManager.java
ata-nas: See follow up #3007

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
ata-nas: See follow up #3007

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockVerificationSession.java
ata-nas: See follow up #3007

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: See follow up #3007

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: See follow up #3007

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/HashingResult.java
ata-nas: See follow up #3007

PR #2877, opened 2026-05-21
feat(publisher): configurable SkipBlock window for duplicate blocks

### LINKED ISSUE #2531: Configuration to send SkipBlock for duplicate blocks within a window
### Story Form

As a Block Node I want a configuration option to be able to send SkipBlock for duplicates (instead of EndOfStream with DUPLICATE_BLOCK) within a small window (perhaps 5-10 blocks at most). This should, if implemented, correspond to the maximum amount a CN is allowed to be "behind" before it reconnects (it should be at least 2 blocks shorter so it's an early indicator).

### Technical Notes

TBD

AlfredoG87:
## Summary

Adds `producer.duplicateBlockSkipWindow` to the publisher plugin. Duplicate block headers within `N` blocks of `lastPersistedBlockNumber` now receive `SkipBlock` so a slightly-behind publisher can fast-forward instead of being disconnected. Duplicates farther behind than the window still receive `EndOfStream(DUPLICATE_BLOCK)` so a hopelessly-behind publisher reconnects.

Fixes #2531.

## Config

- `producer.duplicateBlockSkipWindow` — `int`, default `5`, `@Min(1)`, `@Max(10)`.

## Behavior change

`LiveStreamPublisherManager.getActionForHeader`: when `blockNumber <= lastPersistedBlockNumber`,
- return `SKIP` if `(lastPersistedBlockNumber - blockNumber) <= duplicateBlockSkipWindow`
- otherwise return `END_DUPLICATE` (unchanged path)

`PublisherHandler.handleSkip` already builds the `SkipBlock` response and keeps the stream open — no handler-side change.

## Tests

- New UT cases in `LiveStreamPublisherManagerTest` for in-window SKIP and just-outside END_DUPLICATE boundaries.
- New API E2E `BlockNodeAPITests#duplicateBlockSkipWindowAppliesWithinWindowAndEndsOutsideWindow` covering both branches end-to-end (publish 0..19, republish block 19 → SkipBlock, fresh stream republishing block 0 → EndOfStream(DUPLICATE_BLOCK)).
- Adjusted existing assertions that exercised the legacy "every duplicate ends the stream" path:
  - `BlockNodeAPITests.e2eBlockStreamsBaseScenarios` scenario 2 now asserts SkipBlock for the duplicate genesis.
  - `LiveStreamPublisherManagerTest.testBlockIsEnding…SamePublisherDifferentBlockDuplicate` now asserts SkipBlock.
  - `StreamPublisherPluginTest.PluginTestsPreEarliestManagedBlock` pins the window to its minimum (`1`); the two distance-zero re-publish tests assert SkipBlock.
  - `PositiveMultiplePublishersTests.publisherShouldResumeAfterError` parameterized expectations updated from `status: DUPLICATE_BLOCK` to `skip_block`.

## Test plan

- [x] `./gradlew :stream-publisher:check`
- [x] `./gradlew :suites:runAPISuites`
- [x] `./gradlew :suites:runSuites` (covers `PositiveMultiplePublishersTests`)


jsync-swirlds (dismissed):

Nana-EC (approved):
Approval based on primary reviewer assessment.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
```diff
@@ -19,13 +19,24 @@
 /// Entries within the buffer remain fillable by a publisher that still has the block;
 /// entries older than the buffer are dropped to avoid asking publishers to resend a
 /// block they cannot supply (which kills the connection with TOO_FAR_BEHIND).
+/// @param flowControlPauseDelayNanos Duration in nanoseconds that `onNext` parks the request-processing
+/// thread when a request arrives while the handler is paused. The handler is paused as part of the
+/// staged shutdown sequence in `checkMidBlockAndShutdown`, which schedules the actual `shutdown` to
+/// run after a fixed delay; parking the next incoming request prevents it from racing that pending
+/// shutdown and lets any final outbound messages flush before the connection closes.
```
jsync-swirlds, line 26: Very minor:
I wouldn't recommend including why the pause happens, as it will expand as Issue 1931 is implemented (and possibly at other times).
It's sufficient to note that this controls pause duration

```suggestion
/// thread when a request arrives while the handler is paused. The handler is paused to signal flow
/// control txoff to the Helidon layer (which can then exercise flow control at the HTTP layer by
/// not refilling the frame window).
```

docs/block-node/configuration.md
```diff
@@ -188,11 +188,13 @@ Currently, no specific options.
...
+| PRODUCER_STALE_RESEND_PRUNE_BUFFER        | Number of blocks behind `lastPersistedBlockNumber` that a `blocksToResend` entry may sit before `handlePersisted` prunes it. Entries within the buffer are still considered fillable by a publisher; entries older than the buffer are dropped (the gap is owned by backfill at that point). Set to ~CN block-history depth. Must be 0 ≤ value ≤ 200.                                                                                                          |                       100 |
+| PRODUCER_FLOW_CONTROL_PAUSE_DELAY_NANOS   | Duration in nanoseconds that `onNext` parks the request-processing thread when a request arrives while the handler is paused. The handler is paused as part of the staged shutdown sequence, which schedules the actual shutdown to run after a fixed delay; parking the next incoming request prevents it from racing that pending shutdown and lets any final outbound messages flush before the connection closes. Must be 100,000 ≤ value ≤ 5,000,000,000. |               100,000,000 |
```
jsync-swirlds, line 196: Similar here:

```suggestion
| PRODUCER_FLOW_CONTROL_PAUSE_DELAY_NANOS   | Duration in nanoseconds that `onNext` parks the request-processing thread when a request arrives while the handler is paused. The handler is paused to signal flow control txoff to the networking layer (which can then exercise flow control at the HTTP layer by not refilling the frame window). Must be 100,000 ≤ value ≤ 5,000,000,000. |               100,000,000 |
```

jsync-swirlds (approved):

ata-nas (approved):

PR #2878, opened 2026-05-22
feat: S3 Cloud Archive Plugin to update stored blocks

### LINKED ISSUE #2703: S3 Cloud Archive Plugin to update store blocks
### Story Form

As a S3 Cloud Archive plugin
I want to update the application state with stored blocks
So that other parties know what is stored

### Technical Notes

* Set the stored range when startup recovery tasks completes
* Add new range to the aplication state when a part is uploaded to S3 (the range should contain the block numbers in the part)

ivannov:
## Reviewer Notes

Both cloud storage plugins (`CloudStorageArchivePlugin` and `ExpandedCloudStoragePlugin`) now call `ApplicationStateFacility` to report which blocks have been durably stored in S3, completing the integration started in #2738.

###  CloudStorageArchivePlugin / BlockUploadTask

`BlockUploadTask` gains an `ApplicationStateFacility` parameter. After each successful S3 part flush — both the fixed-size mid-loop flushes in `flushPartIfNeeded` and the final partial-buffer upload in `uploadFinalPart` — it calls  `addStoredBlockRange(LongRange)` with the contiguous range of block numbers whose bytes are now durably stored in that part.

 `CloudStorageArchivePlugin.completeRecovery()` also calls `addStoredBlockRange` once after recovery completes, reporting the range `[0, nextBlockToQueue - 1]` for blocks already committed to S3 before the restart.

### ExpandedCloudStoragePlugin
  
After a successful single-block upload, `handleUploadResult` calls `addStoredBlockRange(new LongRange(blockNumber, blockNumber)). Failed uploads leave the facility untouched.

## Related Issue(s)

Closes: #2703

### DIFF
--- a/block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
+++ b/block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
@@ -17,6 +17,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import org.hiero.block.node.spi.ApplicationStateFacility;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
 import org.hiero.block.node.spi.ServiceBuilder;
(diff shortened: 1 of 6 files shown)


block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
jsync-swirlds, line 1: Note, we might _not_ want to update stored blocks with the expanded cloud plugin, because those blocks are not actually durable.

Not for this PR, but something to discuss for possible adjustment later.

jsync-swirlds (dismissed):
Looks good, one question for future consideration.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: I decided that this PR is the right time for it :)

jsync-swirlds (approved):

Nana-EC (approved):
LG

PR #2880, opened 2026-05-22
feat: Make API Services Port Server Configurable

### LINKED ISSUE #1937: feat(server): Make API Services Port Server Configurable
### Story Form

As a BlockNode Operator
I want to give permission to CN only to stream to BN on publish method.
So that everyone else can subscribe and consume, but cannot publish.

### Technical Notes

When we discussed this feature with @nathanklick, @Nana-EC  and @jsync-swirlds  there was consensus that for any operator is simpler to do filtering of access-list IPs by port (Layer 4) rather than Layer 7 (gRPC method). due to the overhead, complexity and the lack of general support for Bidirectional gRPC Streaming in networking tools such as firewalls and load balancers.

So the first approach was divide the services in 2.
Use current port `40840` for publishers only and move all the other services to a new port that could be `40940`.

@jsync-swirlds suggested that each plugin could define the server port that it wants to use.
This presents with a bigger challenge and complexity but gives maximum flexibility to Operators and third party plugin developers.

ivannov:
## Reviewer Notes

Implements dual-port model. The motivation is to let operators use Layer 4 (IP/port) filtering to restrict which hosts can publish blocks to the node - Consensus Nodes get access to the publisher port, while all other clients (subscribers, block access, health checks) hit a separate consumer port.

* `ServerConfig` gains a `consumerPort` property (default `40940`). Setting `consumerPort` equal to `port` (default `40840`) retains the previous single-listener behaviour with zero config change required.
* `ServiceBuilder` SPI is extended with a `Socket` enum (`PUBLISHER` / `CONSUMER`) and a changed method `registerGrpcService(ServiceInterface, Socket)`. Every plugin that registers a gRPC service now supplies the target socket explicitly
* `BlockNodeApp` creates two `WebServer` instances (publisher and consumer). When both ports are identical the same `WebServer` object is reused for both

## Related Issue(s)
  
Closes #1937

### DIFF
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
@@ -15,9 +15,9 @@ import org.hiero.block.node.base.Loggable;
  * @param maxMessageSizeBytes the PBJ max message size in bytes
  * @param socketSendBufferSizeBytes the socket send buffer size in bytes
  * @param socketReceiveBufferSizeBytes the socket receive buffer size in bytes
- * @param port the port the publisher service (CN -> BN ingest) will listen on
- * @param consumerPort the port that consumer services (subscribers, block access, health) will listen on;
- *     set it equal to {@code port} to run all services on a single port
+ * @param port the default port all services listen on; individual plugins may override this by
+ *     passing an explicit port to {@link org.hiero.block.node.spi.ServiceBuilder}, or pass
+ *     {@code null} to fall back to this value
  * @param shutdownDelayMillis the delay in milliseconds for the service
  * @param maxTcpConnections the maximum number of TCP connections
  * @param idleConnectionPeriodMinutes the period of idle connections check in minutes
@@ -37,8 +37,6 @@ public record ServerConfig(
             @Min(32768) @Max(Integer.MAX_VALUE) int socketReceiveBufferSizeBytes,
         @Loggable @ConfigProperty(defaultValue = "40840")
             @Min(1024) @Max(65_535) int port,
-        @Loggable @ConfigProperty(defaultValue = "40940")
-            @Min(1024) @Max(65_535) int consumerPort,
         @Loggable @ConfigProperty(defaultValue = "500") int shutdownDelayMillis,
         @Loggable @ConfigProperty(defaultValue = "1000") int maxTcpConnections,
         @Loggable @ConfigProperty(defaultValue = "5") int idleConnectionPeriodMinutes,
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
@@ -15,7 +15,9 @@ import org.hiero.block.node.base.Loggable;
  * @param maxMessageSizeBytes the PBJ max message size in bytes
  * @param socketSendBufferSizeBytes the socket send buffer size in bytes
  * @param socketReceiveBufferSizeBytes the socket receive buffer size in bytes
- * @param port the port the server will listen on
+ * @param port the port the publisher service (CN -> BN ingest) will listen on
+ * @param consumerPort the port that consumer services (subscribers, block access, health) will listen on;
+ *     set it equal to {@code port} to run all services on a single port
  * @param shutdownDelayMillis the delay in milliseconds for the service
  * @param maxTcpConnections the maximum number of TCP connections
  * @param idleConnectionPeriodMinutes the period of idle connections check in minutes
@@ -35,6 +37,8 @@ public record ServerConfig(
             @Min(32768) @Max(Integer.MAX_VALUE) int socketReceiveBufferSizeBytes,
         @Loggable @ConfigProperty(defaultValue = "40840")
             @Min(1024) @Max(65_535) int port,
+        @Loggable @ConfigProperty(defaultValue = "40940")
+            @Min(1024) @Max(65_535) int consumerPort,
         @Loggable @ConfigProperty(defaultValue = "500") int shutdownDelayMillis,
         @Loggable @ConfigProperty(defaultValue = "1000") int maxTcpConnections,
         @Loggable @ConfigProperty(defaultValue = "5") int idleConnectionPeriodMinutes,
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -39,7 +39,9 @@ import java.security.spec.X509EncodedKeySpec;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HexFormat;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -63,7 +65,6 @@ import org.hiero.block.node.spi.ApplicationStateFacility;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodeContext.Builder;
 import org.hiero.block.node.spi.BlockNodePlugin;
-import org.hiero.block.node.spi.ServiceBuilder;
 import org.hiero.block.node.spi.ServiceLoaderFunction;
 import org.hiero.block.node.spi.blockmessaging.BlockMessagingFacility;
 import org.hiero.block.node.spi.health.HealthFacility;
@@ -93,13 +94,8 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     private static final Logger LOGGER = System.getLogger(BlockNodeApp.class.getName());
     /** The state of the server. */
     private final AtomicReference<State> state = new AtomicReference<>(State.STARTING);
-    /** The web server for publisher-side services (CN -> BN ingest). Package-private for testing. */
-    final WebServer publisherWebServer;
-    /**
-     * The web server for consumer-side services (subscribers, health, block access).
-     * Equal to {@link #publisherWebServer} when both ports are the same. Package-private for testing.
-     */
-    final WebServer consumerWebServer;
+    /** All WebServer instances, one per distinct port. Package-private for testing. */
+    final Set<WebServer> webServers;
     /** The server configuration. */
     private final ServerConfig serverConfig;
     /** The historical block node facility */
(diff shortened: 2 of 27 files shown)


block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
```diff
@@ -35,6 +37,8 @@ public record ServerConfig(
             @Min(32768) @Max(Integer.MAX_VALUE) int socketReceiveBufferSizeBytes,
         @Loggable @ConfigProperty(defaultValue = "40840")
             @Min(1024) @Max(65_535) int port,
```
AlfredoG87, line 39: should we rename this to publisherPort?

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
```diff
@@ -15,7 +15,9 @@
  * @param maxMessageSizeBytes the PBJ max message size in bytes
  * @param socketSendBufferSizeBytes the socket send buffer size in bytes
  * @param socketReceiveBufferSizeBytes the socket receive buffer size in bytes
- * @param port the port the server will listen on
+ * @param port the port the publisher service (CN -> BN ingest) will listen on
```
AlfredoG87:
```suggestion
 * @param consumerPort the port the publisher service (CN -> BN ingest) will listen on
```

so that is explicit and easily understood?

AlfredoG87 (commented):
Looking good, however a lot of the lift for this PR/Issue is related to deployment.

We are lacking updates to documentation, Solo impacted features, Charts, Docker, etc... 

Besides we need to give a heads-up to Solo, Perf, MN and CN teams about this change is coming... (not needed to capture as part of the PR, just a heads up that the work does not end with the PR)

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
Nana-EC: If we were first building yes but at this stage I'm concerned on the breaking change impact.
If we change it we need a migration path for operators and need to coordinate with provisioner

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -89,8 +93,13 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     private static final Logger LOGGER = System.getLogger(BlockNodeApp.class.getName());
     /** The state of the server. */
     private final AtomicReference<State> state = new AtomicReference<>(State.STARTING);
-    /** The web server. */
-    private final WebServer webServer;
+    /** The web server for publisher-side services (CN -> BN ingest). Package-private for testing. */
+    final WebServer publisherWebServer;
```
Nana-EC: I'm curious about this dual WebServer approach.
Why 2 vs multiple socket ports on a single WebServer?
I imagine this impacts  considerations for threads and other resource allocations.

Additionally long term we want plugin to be able to specify their own port if different from the consumerPort default.
Does this approach support that or add more work that needs to be unwound in the future?

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
```diff
@@ -8,20 +8,38 @@
...
+ *   <li>{@link Socket#PUBLISHER} — for inbound block-stream publishing (CN → BN)</li>
+ *   <li>{@link Socket#CONSUMER} — for all other services (subscribers, block access, health)</li>
+ * </ul>
+ * When both sockets share the same port the implementation merges them onto a single listener.
+ * Every registration requires an explicit {@link Socket} — use {@link Socket#CONSUMER} for all
+ * services except the CN -> BN publisher, which uses {@link Socket#PUBLISHER}.
  */
 public interface ServiceBuilder {
+    /** Identifies which network socket a service should be bound to. */
+    enum Socket {
```
Nana-EC: Is Socket the right concept here?
Have we essentially bundled the plugins into a group? Is this really an aggregation for Publisher plugin vs all other plugins?
In the future when we want to support plugins having their own port how does this help or how would this be expanded?

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
AlfredoG87: Resolving due to comment above.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
AlfredoG87: Makes sense, less disruptive this way 👍

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
```diff
@@ -8,20 +8,38 @@
...
 
     /**
-     * Registers a gRPC service with the web server.
+     * Registers a gRPC service on the given socket.
      *
      * @param service the gRPC service to register
+     * @param socket the target socket ({@link Socket#PUBLISHER} or {@link Socket#CONSUMER})
      */
-    void registerGrpcService(@NonNull ServiceInterface service);
+    void registerGrpcService(@NonNull ServiceInterface service, @NonNull Socket socket);
```
Nana-EC: is it possible rather to have plugins register with their desired port?
Based on the current suggestion if they select Consumer then they adopt the default consumerPort but maybe if they specify CUSTOM then they also provide a custom socket port ?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -259,25 +268,56 @@ public BlockNodeApp(final ServiceLoaderFunction serviceLoader, final boolean sho
...
+                .tcpNoDelay(serverConfig.tcpNoDelay())
                 .build();
 
+        // Create the web server(s). When publisher and consumer ports match, a single server
+        // is created with all routes merged; otherwise two separate servers are used so that
+        // network-level (Layer 4) firewalling can restrict CN access to the publisher port only.
+        final int publisherPort = serverConfig.port();
+        final int consumerPort = serverConfig.consumerPort();
+        if (publisherPort == consumerPort) {
+            final WebServerConfig.Builder wsBuilder = WebServerConfig.builder()
```
Nana-EC: Do we know if pbj-helidon PBJ registers itself at the protocol level or at the routing level? I think Helidon has a socket overload of addRouting that may support the addition of multiple ports on a WebServer.

Nana-EC (changes requested):
Good changes.
Curious on the approach and how it supports long term goal while enabling short term desire to split publisher port from consumer port(s)

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
jsync-swirlds: I agree, I wonder why we could not just add a `port` parameter instead of this enum, and keep a map of ports to servers (assuming WebServer is the right thing to duplicate) so we continue to behave as expected if only one port is used.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: This is exactly my thought.
We should be able to use `.socket` and `.namedRouting` on the Builder to accomplish what we're doing here with multiple WebServer instances.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: About the two server instances. Here is what I read from @jsync-swirlds in Slack (highlight is mine):

> FYI, I was incorrect in daily.  We planned the work to enable multiple ports, but never actually did that work.
Fredy's correct that it's not trivial (**we need to allow for creating new "server" instances in the service builder register method**), but I don't know that implementing a two-port option is significantly more difficult than just allowing plugins to choose their own ports.

That's why I took this way. But maybe I misunderstood it all

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
ivannov: Replaced the enum with port parameters. And now we create a server per port

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
ivannov: Replaced the enum with port parameters. And now we create a server per port

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: This I did not understand unfortunately

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Changed to single server with multiple ports

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Changed to single server with multiple ports

Nana-EC (commented):
Breaking it out into 2 ports was a stop gap solution.
What we really wanted was to go from 1 to n.
I think this PR does that but has additional complexity. I'd prefer keeping it simple and not breaking dev and test considerations, but allowing this PR to support production deployment scenarios.

Here's my suggestion
- Drop `server.consumerPort` and turn it into a per service ports exposed in each plugins config. When a plugin registers a service it passes in the port configured in it's own Config class. Or it passes in null and we use the default server.port 
- Drop the Socket enum 

This should allow for opt-in, no current breakages and per service ports as desired

ivannov:
> Breaking it out into 2 ports was a stop gap solution. What we really wanted was to go from 1 to n. I think this PR does that but has additional complexity. I'd prefer keeping it simple and not breaking dev and test considerations, but allowing this PR to support production deployment scenarios.
> 
> Here's my suggestion
> 
> * Drop `server.consumerPort` and turn it into a per service ports exposed in each plugins config. When a plugin registers a service it passes in the port configured in it's own Config class. Or it passes in null and we use the default server.port
> * Drop the Socket enum
> 
> This should allow for opt-in, no current breakages and per service ports as desired

I removed the `consumerPort` config property. The `Socket` enum was removed in previous commits

jsync-swirlds (approved):

Nana-EC (approved):
Approving to unblock but I think a follow up PR is needed by each service except publisher to actually expose the the ability to node operators to customize this

PR #2888, opened 2026-05-26
feat: Add metadata generation and S3 support for WRB CLI

rockysingh:
Phase 4 draft PR


rockysingh:
Closing Phase 4 as the team has agreed it's not worth pursuing.

## Rationale

Phases 1-3 have already successfully validated the core goal: **proving CN hashing matches WRB CLI hashing** through three independent validation paths:
- **Phase 1**: Direct block validation - verified WRB CLI can validate blocks using BlockStreamBlockHasher
- **Phase 2**: Wrapping validation - verified `blocks wrap` produces identical hashes to CN
- **Phase 3**: Jumpstart validation - verified CN verification hashes in jumpstart.bin match WRB CLI hashes

## Phase 4 Issues

Phase 4 attempted to test the same hashing logic in a different execution mode (live-sequential streaming pipeline). However, it required extensive Solo-specific workarounds:

- `--skip-signatures` flag (Solo: 1 node vs Mainnet: 13 nodes)
- `--allow-timestamp-discovery` flag (Solo Mirror Node returns .blk files)
- S3/MinIO ETag format conversion (hex → base64 MD5)
- Node account range extensions (Solo uses 0.0.3, not 0.0.0)
- Conditional SignatureValidation exclusion
- Multiple debug logging additions

**Result**: 4 files modified, ~150 lines of Solo-specific code, ongoing debugging - all to test the same hashing logic already validated in Phases 1-3.

## Decision

The cost (continued Solo workarounds, production code complexity, maintenance burden) outweighs the benefit (testing operational workflow with already-validated hashing logic). If operational testing is needed in the future, we should test against multi-node testnet/mainnet where standard workflows function without environment-specific modifications.

PR #2903, opened 2026-05-28
feat(state-live): live Hashgraph state plugin (spike, beta)

Nana-EC:
## Context

A Block Node ingests and verifies the block stream but has no queryable view of the
**network state** those blocks produce. Consumers that need state (other plugins, external
clients) must currently reach a Consensus Node. This PR adds a beta `state-management-hashgraph`
plugin that maintains a live, queryable copy of the Hashgraph state inside the Block Node by
replaying verified `state_changes` onto a merkle-backed state, and exposes it over a new
`StateService` gRPC surface.

Scope is the current-state read path only. Historical/point-in-time state queries, remote
(S3) snapshot archival, and block-stream filtering are explicitly out of scope.

> **Status: beta / experimental.** The plugin is not listed in any chart deployment manifest;
> plugins outside the manifest are considered beta.

## What it provides

- **`StateService` gRPC API** with three RPCs — `getBinaryKV`, `getBinarySingleton`,
  `getBinaryQueue` — each taking a `BinaryStateQuery` and returning a `BinaryStateQueryResponse`
  that always carries `StateMetadata` (block number, round number, state root hash, size) so a
  caller knows which block the answer corresponds to. The service binds to a configurable
  `port` (unset ⇒ shares the default `server.port`).
- **A live state** kept current by applying each verified block's `state_changes`, persisted as
  on-disk snapshots under `${stateSnapshotRecentPath}/<blockNumber>` in the canonical
  consensus-node `data/state/` layout (hard-linked, so snapshots are cheap).
- **Only network-attested state is ever served** — readers never observe state the network has
  not confirmed (see lag-1 below).

### State ownership

State is owned by `com.swirlds.state.merkle.VirtualMapStateLifecycleManager` (a merkle-backed
`VirtualMapState`). The plugin drives it through a small pipeline of **three concurrent state
versions**:

- **(3) MUTABLE** — `getMutableState()`: receives the current block's `state_changes`.
- **(2) HASHING** — `hashingImmutable`: the copy sealed by `copyMutableState()`, awaiting hash
  and attestation by the next block.
- **(1) ATTESTED** — `attestedImmutable`: the network-confirmed, query-serving copy (a
  plugin-local cache; the only version the query API reads).

### Lag-1 commit — never expose unconfirmed state

A block N is applied into the mutable and sealed, but is **only exposed to readers once block
N+1's `BlockFooter.startOfBlockStateRootHash` attests it**. On each block the plugin:

1. Reads N+1's footer start-of-block state root hash.
2. Hashes state 2 (post-N) and compares — a match means N+1 attests post-N. A mismatch marks the
   plugin degraded and exposes nothing (the mismatching block stays queued for inspection).
3. Promotes state 2 → state 1 (published to queries) with the just-computed hash as its metadata.
4. Applies N+1 into the mutable and fast-copies to seal the new state 2.

At genesis nothing is attested yet, so queries report `NOT_READY` until the second block confirms
the first.

### Hash only on promotion

Hashing a `VirtualMap` is expensive, so the root hash is computed **once, lazily, at the moment a
sealed state is promoted** (and cached by `VirtualMap` thereafter) — never on the live mutable
(which would seal it and break the next apply) and never eagerly at seal-time. A stream tip that
no successor has yet confirmed is therefore never hashed.

### Immediate, ordered apply

Verified blocks are applied the instant they arrive. `handleVerification` stages the block in an
ordered `ConcurrentSkipListMap` (the gap buffer — apply is strictly contiguous, so N+1 waits for
N) and signals a single **apply-worker thread** that blocks on a queue and drains the buffer
immediately. The same worker first performs one-shot **catch-up** from `HistoricalBlockFacility`
on start (so a fresh or restarted node reaches the tip without waiting for notifications), then
serves live traffic. Apply is guarded by a single lock, so state mutation always happens on one
path.

### Snapshots

A periodic task snapshots the attested state to `recent/<blockNumber>`, keeps the newest
`stateSnapshotRecentRetentionCount` directories (oldest pruned), and rewrites `stateMetadata.json`
atomically. On start the plugin reloads the latest snapshot + metadata and resumes. Long-term
archival (compaction, off-box transfer) is left to a future archiving plugin.

### Storage encoding

The applier writes each change as its PBJ-encoded carrier (`SingletonUpdateChange`, `MapChangeKey`
/`MapChangeValue`, `QueuePushChange`), so clients read back the exact bytes they can parse with the
same codecs.

## Changes

- **New module** `block-node/state-management-hashgraph` — plugin, config, state-change applier,
  metadata store, and module wiring (`settings.gradle.kts`, `gradle/modules.properties`,
  `hiero-dependency-versions`).
- **Protobuf** — `StateMetadata`, `BinaryStateQuery`, `BinaryStateQueryResponse`, and the three
  `getBinary*` RPCs on `state_service.proto`.
- **Tests** — plugin unit + acceptance suites against the real `VirtualMapStateLifecycleManager`,
  gRPC query rules, historical catch-up, lifecycle/restart round-trip, and an end-to-end suite
  driving `StateService` over real gRPC.
- **Docs** — `docs/design/state/live-state.md` design doc and the module `README.md`.

## Configuration (`@ConfigData("state.management")`)

| Property | Default | Notes |
|---|---|---|
| `stateMetadataPath` | `/opt/hiero/block-node/data/state/stateMetadata.json` | Metadata JSON file. |
| `stateSnapshotRecentPath` | `/opt/hiero/block-node/data/state/snapshot/recent` | Recent snapshot dirs. |
| `snapshotIntervalMillis` | `900000` (15m) | Snapshot cadence. |
| `historicCatchUpBatchSize` | `64` | Blocks pulled per catch-up iteration. |
| `stateSnapshotRecentRetentionCount` | `3` | Recent snapshot dirs kept (oldest pruned). |
| `port` | `null` | Dedicated `StateService` port; `null` shares `server.port`. |

## Testing

`:state-management-hashgraph:build` (unit + acceptance + spotless) is green; the E2E suite drives
`StateService` over gRPC against a booted Block Node. Built and tested at `hederaVersion 0.76.0-rc.1`.

## Notes

- No `enabled` flag — Block-Node plugins are active whenever their jar is on the classpath; opt-in
  is via the deployment manifest, and this plugin is intentionally left out for beta.
- The plugin depends on the swirlds state libraries (`swirlds-state-api`/`-impl`,
  `swirlds-virtualmap`) at the repo's existing `hederaVersion`.

## Why I'm marking this **beta**, not GA

- The plugin runs end-to-end: verified blocks apply, state persists across restarts, gRPC queries serve real data.
- **But**: no hash-mismatch recovery, no metrics, no health-facility integration, no backpressure under sustained throughput, no Merkle-proof RPCs, swirlds dependency-tree is heavier than it should be, and the plugin still contributes swirlds library config records that should live elsewhere.
- Production rollout should wait until at least two things land: the hash-mismatch recovery protocol, and the metrics + health-facility integration. Until then the plugin should run only in opt-in deployments behind the manifest gate.


block-node/state-management-hashgraph/src/main/java/org/hiero/block/node/state/management/SnapshotArchiver.java
```diff
@@ -0,0 +1,150 @@
...
+ * the same 512-byte header layout, octal numeric fields, ustar magic, and
+ * 1024-byte end-of-archive marker. The difference: that iterator emits one tar
+ * entry per BlockAccessor; this archiver walks an arbitrary directory tree (the
+ * consensus-node {@code data/state/} layout produced by
+ * {@code VirtualMapStateLifecycleManager.createSnapshot}) recursively.
+ *
+ * <p>Writes go through {@code &lt;historicPath&gt;/&lt;blockNumber&gt;.tar.tmp} and
+ * atomic move so a crash mid-write leaves the previous archive intact.
+ */
+final class SnapshotArchiver {
```
jasperpotts: I am not sure why we want to tar snapshots? I think initially we would just keep N recent snapshots on disk. Later we would add a archiving plugin that archives those snapshots. When we do that we need to think about how to run compaction on them and how to make them as small as possible. Also think about if we want some new indexes in them for random read ability depending on what services we want the BN to provide. To start with though it just needs to keep enough snapshots to have a reliable one for seeding/reconnecting other BNs and later CNs.  They will only ever need the last good snapshots. The snapshots are new directories on disk each time but actually share a lot of files with current live DB via `hard-links` that reduces snapshot time and disk use. If you tar then you loose that performance gain and disk space reduction that comes from hardlink use.

jasperpotts:
I think we can do something much simpler for now. I don't think we need a state status or state update notification. We can just always work on latest state. Any api to get a saved state, just returns the most recent a block node has. No historical saved state fetching. That is fine for almost all use cases. Basically reconnect for BNs or CNs.

Then for plugins that need to read state, they always get to read from latest. The challenge he is managing the life cycle of state versions. There will always be the latest mutable state that is being written into by the state changes coming from latest block.  Then there is the most recent immutable state, This is what people should be querying.  I am not sure if we want to do state validation before we expose or expose immediately. That is a critical question. If we want to validate first then at the end of a block we create a new version for next block to write to. The old immutable version is then hashed and its root hash checked with block footer stored/trusted root hash for state. If it matches we know we have a trusted state. Then we hold a copy of that state till the next blocks state if finished, hashed and validated. That copy should never be aloud to escape the plugin as we have to manually release it. State versions can not depend on garbage collector.  So all queries will have to be like `value mapGet(stateID, key)` or `value singletonGet(stateID)` directly on plugin. The plugin then backs those with the latest validated state. 

Let me know if this makes sense or you want to chat about it.

Nana-EC:
> I think we can do something much simpler for now. I don't think we need a state status or state update notification. We can just always work on latest state. Any api to get a saved state, just returns the most recent a block node has. No historical saved state fetching. That is fine for almost all use cases. Basically reconnect for BNs or CNs.
> 
> Then for plugins that need to read state, they always get to read from latest. The challenge he is managing the life cycle of state versions. There will always be the latest mutable state that is being written into by the state changes coming from latest block. Then there is the most recent immutable state, This is what people should be querying. I am not sure if we want to do state validation before we expose or expose immediately. That is a critical question. If we want to validate first then at the end of a block we create a new version for next block to write to. The old immutable version is then hashed and its root hash checked with block footer stored/trusted root hash for state. If it matches we know we have a trusted state. Then we hold a copy of that state till the next blocks state if finished, hashed and validated. That copy should never be aloud to escape the plugin as we have to manually release it. State versions can not depend on garbage collector. So all queries will have to be like `value mapGet(stateID, key)` or `value singletonGet(stateID)` directly on plugin. The plugin then backs those with the latest validated state.
> 
> Let me know if this makes sense or you want to chat about it.

Makes sense, I thought through those all and I believe my implementation covers most of this. See the design doc where it's all noted.
The main thing is it seems like I didn't need to expose the metadata in the status API, if that's needed or desired I can drop it
Other than that I already hand both states, I transition current mutable state to immutable after the hash verification from the next blocks footer confirms I applied the state correctly. All state queries implemented `getBinarySingleton()`, `getSingleton()` and `getBinaryQueue()` are all off of immutable state.
Let me know what gaps you're seeing and I can add that to simplifying as needed

block-node/state-management-hashgraph/src/main/java/org/hiero/block/node/state/management/SnapshotArchiver.java
Nana-EC: I went for tar to support the remote archive of snapshots.
I left out reading local archive as we were focusing on live state.
One consideration is a similar zip structure that we use for blocks to allow for som compression but the ability to read as needed.
Need to look into how that would scale as state snapshots are generally much bigger than blocks.

Nana-EC: Yeah I just learned about the shared hard-links so I need to explore how this affects the current logic where I archive the older snapshots.
I'll update it to keep it more simple and just keep n number recent snapshots.
I might finish up the logic here before simplifying but feedback noted.

Nana-EC (commented):
Self review

---

Feedback on state-management-spike. 

- (P2) In block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockNotificationRingEvent.java in the set methods you utilize clearAll() then do a set. Here teh order is important and a dev could change this without understanding the impact. Add appropriate comments in line
- (P2) In ServerStatusDetailResponse the field should be stateMetadata not hashStateMetadata as it's the whole object
- (P3) In LiveStatePlugin.start() we leave the BN status as NOT_READY until we're caught up. This makes sense but the impacts are unclear, can you elaborate on considerations
- (P2) The plugin should be called state-management-hashgraph not state-hashgraph-live and the class LiveStatePlugin should be StateManagementPlugin
- (P1) The LiveStateE2ETests seems a bit light. Let's add at least 2 more tests
  - 1 that starts from genesis and streams some blocks with state changes and verifies that we can get valid state objects across the 3 items
  - 1 that restore a statesnapshot and verifies that we can get valid state objects across the 3 items

In LiveStatePlugin
- (P1) why schedule applyPending before we've caught up. Could this lead to  failed attempts to apply or worse out of order attempts to apply? If you agree move the schedule call after the catchUpFromHistoricialBlocks has completed. If you disagree share why.
- (P1) In applyPending() what if there's a failure in applying a block, since we removed it before attempting to apply we'd ever be able to retry. Why not peek and then apply and on apply then remove?
- (P3) rename ready variable to something more informative like stateIsCaughtUp or stateIsRecent
- (P2) Test hooks are a smell in production code. DO we need them in LiveStatePlugin. What other approaches could we take to make LiveSTatePlugin shorter and without bloat
- (P1) In applyOne() we call the lifecycleManager.copyMutableState() after we apply the state changes applier.applyBlock(mutable, block). Is that correct. If the state at start was representing state after applying 100, then we'd have just got the BlockFooter of 101 and validated the startOfBlockStateRootHash matched our current stateRootHash. Shouldn't we immediately transition the mutable state to immutable if validateStartHash returns true and send out the StateUpdate notification, before applying the state changes from block 101?
- (P3) Rename applyOne to applyBlockStateChanges()
- (P3) Rename pruneRecentExcept() to pruneAndArchiveRecentSnapshotsExcept
- (P2) In enforceHistoricRetention() can you confirm that we are deleting the oldest archives first when enforcing the retention policy. Also add a test case for this to validate. For example of the retentionPolicy is 3 and we have 0.tar, 1.tar, 2.tar and we just added 3.tar, then i'd expect 0.tar to be removed.
- (P3) Rename snapshotDirectoryFor() to recentSnapshotDirectoryFor()
- (P1) matchesLatestBlock() assumes block 0 is invalid, that's not the case. Block 0 is the genesis block and is a valid blockNumber to provide in a request. This seems to map to a comments in BinaryStateQuery.block_number in the state_service.proto. We should update this to match the convention in BlockRequest message in block_access_servie.proto. Here it uses a oneof block_specifier where one must specify block_number or retrieve_latest. Let's copy this and make a note in the future to consider reusing by moving the message to shared_message_type.proto. We either go this approach or allow  BinaryStateQuery.block_number to be nullable.
- (P1) readableState() doesn't seem like it should exist a caller should always just call lifecycleManager.getLatestImmutableState(), if the issue is at load this is null then at load of a snapshot we should do a copyMutableState to ensure it's not null. The result is mutable and immutable wil be the same until the first state_change is applied to mutable, that's fine I think.
- (P3) Rename deleteRecursively to deleteFileRecursively()
- (P1) let's update the logic to keep n number of recent snapshots under config.stateSnapshotRecentPath(). We can use a similar logic where there is a retention policy for recent and on snapshot creation we remove the additional older dirs. To make it clean separate the snapshot archive to historic logic from the historic deletion logic as I may make changes to the archive logic later. Note recent and historic should have different retention policies
- (P1) Why do we need inspectBlock(), it forces us to walk the block twice. Could we not add StateMetaData to the ApplyResult class, that way we build it as we walk the block and apply the state changes. If this won't work why?
- (P3) In StateChangeApplier for the case STATE_ADD, STATE_REMOVE, UNSET we should log this unexpected observance. 
- (P3) Some of the write logic in SnapshotArchiver.java seems similar to TaredBlockIterator. We should make a note to consider how to consolidate these later.
- (P3) Update all the Java comments in the feature branch to use MD style also add missing comments to methods and classes

Nana-EC:
@jasperpotts I simplified items where I was jumping ahead for future expansion.
I took out the notification and I confirmed the lagging nature of moving state to immutable only after network attestation.
I wasn't aware of the state rook release logic so after researching I added that.
It should be simpler now but I might need another pass to see if I'm happy with the test coverage and if there's anyother simplifications to make

Issue #2899, opened 2026-05-27
bug(solo-e2e): seems like all the solo-e2e-tests have been failing since last week, possibly since the WRB addition


AlfredoG87:
### Description

https://github.com/hiero-ledger/hiero-block-node/actions/runs/26497615883/job/78029646494

**UPDATE 1:**
Confirmed regression on Receiving oversized Block for first WRAPS Proof. from Schnorr to Wraps proof.
Potentially from latest round of optimizations over previous sprint.

### Steps to Reproduce

Just run the Solo-e2e-test using the Daily schedule

PR #2905, opened 2026-05-28
fix(publisher): remove closeConnection() from shutdown() to preserve shared TCP connections

AlfredoG87:
fix(publisher): remove closeConnection() from shutdown() to preserve shared TCP connections (#2899)

## Summary

Fixes a latent bug in `PublisherHandler.shutdown()` exposed by the receive-buffer increase in #2835. One production bug is fixed, the test suite is updated to match the corrected behaviour, and one regression test is added.

### Root cause

During the Schnorr→WRAPS signature transition the CN sends `EndStream(RESET)` on stream A then immediately opens stream B on the **same TCP connection** to deliver the first WRAPS-signed block.

`PublisherHandler.shutdown()` (introduced in #1859) calls both `replies.onComplete()` and `replies.closeConnection()`. In Helidon, `onComplete()` closes only the HTTP/2 stream (stream A). `closeConnection()` closes the underlying TCP socket — shared by all HTTP/2 streams, including stream B. Stream B never delivers its block; the BN gets stuck and loops.

The bug was masked until #2835 increased `socketReceiveBufferSizeBytes` from 128 KB to 8 MB. The larger buffer lets the CN push the full 90 MB WRAPS block into the kernel without TCP back-pressure, then immediately send EndStream(RESET) and open stream B — all before the BN application has processed the block. The tighter timing window made the race reliably observable.

## Changes

- **`PublisherHandler.java`** — remove `replies.closeConnection()` from `shutdown()`; `replies.onComplete()` is sufficient.
- **`BlockNodeAPITests.java`** — `e2eSocketClosureTest`: assert `EndOfStream(DUPLICATE_BLOCK)` instead of socket-closed exception (no longer thrown). Update `@Disabled` reason on `e2eDuplicateBlockPublisherObserversOnComplete` to reflect `SKIP_BLOCK` window from #2877.
- **`BlockNodeApiRegressionTest.java`** — add `streamBDeliversBlockBodyAfterStreamAShutdownDelay`: stream B sends a block header before EndStream on stream A, waits for stream A `onComplete`, then sends the block body; asserts ACK and no errors on stream B.

## Test plan

- [x] `./gradlew :suites:runAPISuites` — 14 tests pass, 1 skipped
- [x] `tss-signature-transition` solo-e2e passes on CI see [here](https://github.com/hiero-ledger/hiero-block-node/actions/runs/26551647484/job/78214819257)

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
@@ -766,7 +766,6 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
             try {
                 replies.onComplete();
                 isPaused.set(false); // unpause to prevent deadlocks.
-                replies.closeConnection();
                 // @todo() Add labeled metric when possible.
                 //    Metric: "handler-closed" Labels: "clean" or "with-exception"
                 //    with-exception should be set in all exception cases, even if not logged.
@@ -782,7 +781,7 @@ public final class PublisherHandler implements Pipeline<PublishStreamRequestUnpa
                         .formatted(handlerId, correlationIdPrefix);
                 LOGGER.log(DEBUG, message, e);
             }
-            LOGGER.log(DEBUG, "[{0}] Handler {1} issued onComplete/closeConnection", correlationIdPrefix, handlerId);
+            LOGGER.log(DEBUG, "[{0}] Handler {1} issued onComplete", correlationIdPrefix, handlerId);
         }
     }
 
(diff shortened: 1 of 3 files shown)


jsync-swirlds:
I have two concerns here:
1. Why is the CN closing a stream and opening a new one.
1. We have been very clear from the beginning that Block Node does _not_ support multiple streams on a single connection.  I do not think we want to start trying to support that now as it creates a huge array of race conditions and failure modes we've ignored because "we do not support multiple streams on a connection".

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -766,7 +766,6 @@ private void shutdown() {
             try {
                 replies.onComplete();
                 isPaused.set(false); // unpause to prevent deadlocks.
-                replies.closeConnection();
```
jsync-swirlds, line 769: This is not OK.  We still need to close the connection, for a lot of reasons.
Failing to close the connection leaves abandoned connections open and doesn't disconnect misbehaving publishers, so a buggy publisher could just send random data in a way that consumes a lot of threads and a huge amount of memory that never gets cleared or processed.

jsync-swirlds:
Given the underlying behavior turned out to not be the root cause, can we close this PR?

AlfredoG87:
Closing the connection is the expected behaviour, this is not an acceptable FIX, also, it passed the CI, but seems coincidental.

PR #2906, opened 2026-05-28
feat: Add flow control and basic overload handling

### LINKED ISSUE #1931: Add handling for rapid "block dump" delivery from a publisher or publishers
### Story Form

As a Block Node
I want to gracefully handle a large surge of blocks from publisher(s)
So that I can secure blocks that were buffered due to a short-term disruption such as a brief network interruption.

### Technical Notes

We intend to add detection and streamlined handling to the publisher plugin to receive a volume of blocks in very rapid succession without out-of-memory errors or other failures while strongly prioritizing verification and persistence of the blocks.
This involves at least the following
- Limit the number of blocks "pending" in the publisher at any one time to a configurable maximum.
   - Pause (but do not terminate) incoming streams (use HTTP stream backpressure mechanisms) when the limit is reached
   - Allow streams to continue as space is freed by delivery of block data to messaging.
- Detect that a large number of blocks is coming from a publisher and fast-track delivering those blocks to messaging, where practical
- Detect a publisher surge and publish a notification that "recommmends" backfill and other non-essential activities pause until the blocks are verified and stored.
- If a publisher is detected to surge but the blocks are not passing verification, discard further blocks from that publisher, disconnect the publisher, and add that publisher to the penalty box.

There may be other items needed to fully support this type of surge bulk delivery of pending blocks.

jsync-swirlds:
## Reviewer Notes
* Added budget-based flow control
* Added configuration for flow control
* Added basic overload control via "penalty" mechanism
* Added initial tests for new functionality
* Adjusted shutdown delay usage to only delay when necessary
* Simplfied some overly complicated methds in Handler
* Adjusted scheduled thread handling

## Related Issue(s)
 Resolve #1931

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -10,6 +10,8 @@ import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_BLOCKS_CLOSED_COMPLETE;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_BLOCK_BATCHES_MESSAGED;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_BLOCK_ITEMS_MESSAGED;
+import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_FLOW_CONTROL_AGGREGATE_PAUSES;
+import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_FLOW_CONTROL_PENALTIES_APPLIED;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_HIGHEST_BLOCK_NUMBER_INBOUND;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_LATEST_BLOCK_NUMBER_ACKNOWLEDGED;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_LOWEST_BLOCK_NUMBER_INBOUND;
@@ -21,6 +23,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.ArrayList;
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -69,6 +72,11 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             "[{0}] Stall detected: handler {1} has not completed block {2}, another handler completed block {3}";
     private static final String BLOCK_ABANDONED_LOG_MESSAGE =
             "[{0}] Replacement persistence detected: handler {1} has not completed block {2}, but a different source persisted the same block {3}";
+    /// Limit the maximum scheduled threads to 12, more than that is almost certainly excessive.
+    private static final int MAX_SCHEDULE_THREADS = 12;
+    /// Value assigned as a Handler message budget when that handler is being paused
+    /// due to overload or other ill-advised behaviors.
+    private static final long MESSAGE_BUDGET_PENALTY_FLAG_VALUE = -1L;
     private final System.Logger LOGGER = System.getLogger(LiveStreamPublisherManager.class.getName());
     private final MetricsHolder metrics;
     private final BlockNodeContext serverContext;
@@ -87,6 +95,10 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     private final int duplicateBlockSkipWindow;
     private final ScheduledExecutorService scheduledExecutor;
     private volatile ScheduledFuture<Boolean> publisherUnavailabilityTimeoutFuture;
+    /// nanoTime of the last penalty-count reset (tumbling window).
+    private long lastPenaltyResetNanos;
+    /// Scheduled handle for the flow-control refresh task.
+    private ScheduledFuture<?> flowControlRefreshFuture;
 
     /// Future tracking the queue forwarder task.
     ///
@@ -109,6 +121,8 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     private final ConcurrentSkipListSet<Long> endBlocksReceived;
     private final ConcurrentSkipListSet<Long> blocksToResend;
     private final ConcurrentSkipListSet<Long> activeResendBlocks;
+    /// Flow state per handler ID; created in addHandler, removed in removeHandler.
+    private final ConcurrentSkipListMap<Long, HandlerFlowState> handlerFlowState;
     /// Maps block number to the ID of the handler that currently holds ACCEPT for that block.
     /// An entry is added when a handler wins ACCEPT via registerQueueForBlock().
     /// An entry is removed when endOfBlock() is called for that block, when blockIsEnding()
@@ -132,12 +146,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         dataReadyLatch = dataReadyLock.newCondition();
         NodeConfig nodeConfiguration = serverContext.configuration().getConfigData(NodeConfig.class);
         earliestManagedBlock = nodeConfiguration.earliestManagedBlock();
-        int threadCount = Runtime.getRuntime().availableProcessors();
-        // Ensure at least 2 threads for scheduled tasks so we do not have
-        // excessive contention with the idle detector task.
-        threadCount = threadCount < 2 ? 2 : threadCount;
-        scheduledExecutor = threadManager.createVirtualThreadScheduledExecutor(
-                threadCount, null, this::uncaughtScheduledExecutorException);
+        scheduledExecutor = setupScheduledExecutor(threadManager);
         publisherConfig = serverContext.configuration().getConfigData(PublisherConfig.class);
         maxBlocksBeforeStalled = publisherConfig.MaxFutureBlocksBeforeStalled();
         staleResendPruneBuffer = publisherConfig.staleResendPruneBuffer();
@@ -148,7 +157,21 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         blocksToResend = new ConcurrentSkipListSet<>();
         activeResendBlocks = new ConcurrentSkipListSet<>();
         activeStreamHandlerByBlock = new ConcurrentSkipListMap<>();
+        handlerFlowState = new ConcurrentSkipListMap<>();
         initializeBlockNumbers(serverContext);
+        scheduleFlowControlRefresh();
+    }
+
+    /// todo(1420) add documentation
+    private ScheduledExecutorService setupScheduledExecutor(final ThreadPoolManager threadManager) {
+        // start with half the available processors.
+        int threadCount = Runtime.getRuntime().availableProcessors() / 2;
+        // Ensure at least 2 threads for scheduled tasks so we do not have
+        // excessive contention with the idle detector task.
+        threadCount = threadCount > 2 ? threadCount : 2;
+        threadCount = threadCount > MAX_SCHEDULE_THREADS ? MAX_SCHEDULE_THREADS : threadCount;
+        return threadManager.createVirtualThreadScheduledExecutor(
+                threadCount, null, this::uncaughtScheduledExecutorException);
     }
 
     @Override
@@ -164,6 +187,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         // The cancel of the existing future must happen immediately prior to updating the active handlers map!
         cancelExistingFuture();
         handlers.put(handlerId, newHandler);
+        handlerFlowState.put(handlerId, new HandlerFlowState());
         // Now we can safely update the metrics and send the notification
         // for the new publisher.
         metrics.currentPublisherCount().set(handlers.size());
@@ -175,6 +199,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     @Override
     public void removeHandler(final long handlerId) {
         final PublisherHandler handlerRemoved = handlers.remove(handlerId);
+        handlerFlowState.remove(handlerId);
         // If there are no more active publishers, schedule the
         // unavailability timeout task.
         if (handlerRemoved != null && handlers.isEmpty()) {
@@ -304,6 +329,10 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         }
         queueByBlockMap.clear();
         activeStreamHandlerByBlock.clear();
+        // Cancel the flow control refresh task before executor shutdown.
+        if (flowControlRefreshFuture != null) {
+            flowControlRefreshFuture.cancel(false);
+        }
         // We can safely shut down the scheduled executor abruptly. The timeout
         // tasks can be ignored completely as we shut down the manager.
         scheduledExecutor.shutdownNow();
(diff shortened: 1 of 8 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -132,10 +143,12 @@ public LiveStreamPublisherManager(
...
         earliestManagedBlock = nodeConfiguration.earliestManagedBlock();
-        int threadCount = Runtime.getRuntime().availableProcessors();
+        // start with half the available processors.
+        int threadCount = Runtime.getRuntime().availableProcessors() / 2;
         // Ensure at least 2 threads for scheduled tasks so we do not have
         // excessive contention with the idle detector task.
         threadCount = threadCount < 2 ? 2 : threadCount;
+        threadCount = threadCount > MAX_SCHEDULE_THREADS ? MAX_SCHEDULE_THREADS : threadCount;
         scheduledExecutor = threadManager.createVirtualThreadScheduledExecutor(
                 threadCount, null, this::uncaughtScheduledExecutorException);
```
ata-nas: Cleanup suggestion:
```suggestion
scheduledExecutor = setupScheduledExecutor(threadManager);
```

Then the method called looks like this:

```java
    private ScheduledExecutorService setupScheduledExecutor(final ThreadPoolManager threadPoolManager) {
        // Start with half the available processors.
        final int threadCountCandidate = Runtime.getRuntime().availableProcessors() / 2;
        // Ensure at least 2 threads for scheduled tasks so we do not have
        // excessive contention with the idle detector task.
        // Also, ensure that thread count does not exceed the maximum allowed value
        final int threadCount = threadCountCandidate <= 2 ? 2 : Math.min(threadCountCandidate, MAX_SCHEDULE_THREADS);
        // Create and return the scheduled executor.
        return threadPoolManager.createVirtualThreadScheduledExecutor(
                threadCount, null, this::uncaughtScheduledExecutorException);
    }
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -69,6 +72,8 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager
             "[{0}] Stall detected: handler {1} has not completed block {2}, another handler completed block {3}";
     private static final String BLOCK_ABANDONED_LOG_MESSAGE =
             "[{0}] Replacement persistence detected: handler {1} has not completed block {2}, but a different source persisted the same block {3}";
+    /// Limit the maximum scheduled threads to 12, more than that is almost certainly excessive.
+    private static final int MAX_SCHEDULE_THREADS = 12;
```
ata-nas, line 76: I'd even limit this to 6 or 8.

Could be configurable value also.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -175,6 +191,7 @@ public PublisherHandler addHandler(
     @Override
     public void removeHandler(final long handlerId) {
         final PublisherHandler handlerRemoved = handlers.remove(handlerId);
+        handlerFlowState.remove(handlerId);
```
ata-nas, line 202: If we add the flow state after adding the handler to the handlers map, should we also remove the flow state before removing the handler from the handlers map? Not sure about the order of operations here.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -304,6 +321,10 @@ public void shutdown() {
         }
         queueByBlockMap.clear();
         activeStreamHandlerByBlock.clear();
+        // Cancel the flow control refresh task before executor shutdown.
+        if (flowControlRefreshFuture != null) {
+            flowControlRefreshFuture.cancel(false);
```
ata-nas, line 334: Why not interrupt?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -223,6 +236,26 @@ long getId() {
...
+        return messageBudget.get();
+    }
+
+    /// Sets the message budget using CAS so that a concurrent shutdown (Long.MIN_VALUE) is
+    /// never overwritten by the refresh task.
+    ///
+    /// @param expectedValue the value the budget must currently hold for the set to succeed
+    /// @param newValue the new budget value to apply
+    void setMessageBudget(final long expectedValue, final long newValue) {
+        messageBudget.compareAndSet(expectedValue, newValue);
```
ata-nas, line 247: What happens if the CAS fails? We are missing an update? Is it possible to be in a situation where the CAS can fail not due to a concurrent shutdown?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -947,11 +968,120 @@ private Future<Long> launchQueueForwarder() {
...
+                            handlerId,
+                            state.penaltyCount,
+                            currentPenaltySeconds);
+                }
+            } else {
+                state.consecutiveFullBudgetIntervals = 0;
+            }
+
+            if (state.isPenalized(nowNanos)) {
+                handler.setMessageBudget(snapshotBudget, -1L);
```
ata-nas: The `-1` is probably better as a static flag. Conveys the intentionality of the usage.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -765,7 +800,7 @@ private void shutdown() {
         } finally {
             try {
                 replies.onComplete();
-                isPaused.set(false); // unpause to prevent deadlocks.
+                messageBudget.set(1); // release the onNext thread enough to close
```
ata-nas: `messageBudget.set(1);` is not guaranteed because `onComplete` could throw.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -947,11 +968,120 @@ private Future<Long> launchQueueForwarder() {
...
+            LOGGER.log(
+                    DEBUG,
+                    "Aggregate message budget exceeded ({0} consumed, limit {1}); withholding refresh for all handlers",
+                    totalConsumed,
+                    totalBudget);
+            return;
+        }
+
+        // 4. Per-handler streak, penalty update, and budget refresh.
+        for (final Map.Entry<Long, PublisherHandler> entry : handlers.entrySet()) {
```
ata-nas, line 1036: In a concurrent skip list map/set, the entries/values/keys views are weakly consistent. Not sure if we can improve to not use that directly in a loop as updates or partial updates may or may not happen. But maybe that is not an issue.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -947,11 +968,120 @@ private Future<Long> launchQueueForwarder() {
...
+                            handler.getCorrelationId(),
+                            handlerId,
+                            state.penaltyCount,
+                            currentPenaltySeconds);
+                }
+            } else {
+                state.consecutiveFullBudgetIntervals = 0;
+            }
+
+            if (state.isPenalized(nowNanos)) {
```
ata-nas, line 1070: Will this possibly repeat on each iteration while the handler is penalized? How does that affect each iteration?

ata-nas (commented):
General questions and some cleanup suggestions

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: To be clear, this is just "extract method", correct? No other change to the logic?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I'm trying to limit configurable values, we already have too many, and we do not want to end up where CN is with multiple thousands that never get changed.
12 is, I feel, a good balance between current and future needs; particularly when we have larger numbers (20+) of Publishers sending to a single BN.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: It's pretty loose as there isn't a tight timing requirement.  If we set the budget one last time after the handler is removed, or the reverse, it doesn't break anything.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Generally, it is best to cancel without interrupt if it is acceptable for the task to complete an in-progress action.  In this case the refresh process is very fast, with no pauses or locks, so it's pretty safe to let it finish if the thread is active.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: The CAS could only fail if the caller had incorrect knowledge of the current budget.  That's possible, and would generally result in simply not refreshing a budget on a cycle.
I'll take another look, but that should be acceptable, as it tends to slow down a busy handler, but doesn't slow a less busy handler.  This is the point of the flow control, so that seems reasonable.

Note, this shouldn't affect global pauses or penalty pauses (though I need to check the penalty case).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Good idea, will update.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: Good point, and it affects connection close as well (which is far more critical; not receiving the next message is much less concerning).
I'll fix this.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: The entry set will never be in a partially updated state; but it will change during the iteration.
In theory I could copy the set up front, but that seems excessive, as we're just updating budgets; if we update a budget early or a little late, that shouldn't be a major concern.

I'll think more on this.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: This repeats on each refresh, and it checks "is the penalty done, if so refresh the budget, otherwise skip this handler".

Not sure what the concern is?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: Yes, method is extracted.

Changes are ternary operator is `<= 2` instead of `< 2`, a minor optimization. Then, `Math.min` is used so it reads better in case thread count is strictly higher than 2. Finally, it returns the constructed executor. In constructor, it becomes a one-liner.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: 12 is not a bad choice, just felt we can go even lower. But if we do not have large amount of processors, we will have a low value anyway. So I guess we are effectively achieving the same thing.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: Just a clarifying question. No concerns in particular.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Will take a look.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -188,10 +192,19 @@ public void onSubscribe(@NonNull final Subscription subscription) {
...
+        boolean paused = false;
+        while (messageBudget.get() <= 0) {
+            paused = true;
+            LockSupport.parkNanos(configuration.flowControlRefreshIntervalNanos());
         }
+        if (paused) {
+            metrics.flowControlIndividualPauses().increment();
+        }
+        // Now we decrement, because we're actually processing the message.
+        messageBudget.decrementAndGet();
```
ata-nas, line 207: I feel like we should guard this logic behind the `isActive` flag. If we are not active, we should simply not go through this. It might also be a flag to exit the `while` loop as well. So we keep sleeping while we have not budget && we are active.

We have 3 types of shutdowns.

1. Scheduled - flag is set and the next pipeline call does the shutdown;
   a. The actual shutdown may be immediate or delayed, but in either case, it will happen on the next pipeline call.
3. Delayed - shutdown is asynchronously delayed but does not wait for the next pipeline call;
4. Immediate (end of stream received) - happens immediately on this pipeline call, we do not need to delay as we are not sending any responses back.

In either case, the flag `isActive` should be CASed. It might be a good idea to not wait for the budged if we are inactive.

On another note, do we need to make the checkMidBlockAndShutdown (or shutdowns in general) idempotent?

jsync-swirlds: The budget check is deliberately before checking for active and doesn't obey active/inactive, because we need to block this thread to signal HTTP flow control, and allowing it to return means we could accept many megabytes of additional data before we get a chance to block the thread again.
Returning doesn't really allow a single additional message, it allows the buffers to be (re)filled.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -87,6 +95,10 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager
     private final int duplicateBlockSkipWindow;
     private final ScheduledExecutorService scheduledExecutor;
     private volatile ScheduledFuture<Boolean> publisherUnavailabilityTimeoutFuture;
+    /// nanoTime of the last penalty-count reset (tumbling window).
+    private long lastPenaltyResetNanos;
+    /// Scheduled handle for the flow-control refresh task.
+    private ScheduledFuture<?> flowControlRefreshFuture;
```
ata-nas, line 101: Should this future be volatile?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -188,27 +186,30 @@ public void onSubscribe(@NonNull final Subscription subscription) {
...
         if (!isActive.get()) {
-            checkMidBlockAndShutdown(currentStreamingBlockNumber.get());
+            checkMidBlockAndShutdown(currentStreamingBlockNumber.get(), true);
```
ata-nas, line 212: Not sure about this being set to `true`.

Here, if we had to send a message, we would already have done that at another point.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Not sure why; it's only started once and not restarted, so it shouldn't have any cache-coherency issues.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: Good point, will change to false (do not delay) in this case.

We probably (in follow up) need to go through the setting of `isActive` to make sure we're not using that as a pseudo-delay anywhere, however.

jsync-swirlds: Actually, checking the path(s) that can set `isActive` to false, it looks to me like we need the `true` here for delay.
Scenario:
* Publisher is sending data
* A non-urgent End-of-stream message sent (perhaps timed reset), `isActive` is set `true`
* Publisher next message arrives microseconds later
* We shut down with no delay

I think we need the delay here to ensure that doesn't happen.  Given that the main way `isActive` is set true is when we send an end-of-stream message, we need to make sure the publisher has time to pause properly.

ata-nas: So, yes, this is tricky and I did think of what you note as a scenario.

Note that onNext will only be called in series, so we have to release the current onNext call before we are able to receive the next one. If someone wants to schedule a shutdown (different than delay, this is when manager wants to say to the publisher "shutdown on your next pipeline call") they will send the response (or at least send it to the pipeline) before actually setting the isActive boolean. This still gives some time, but we can actually consider to schedule that as a task as well.

My worry is that we want to shutdown on next because we are no longer active, but we actually delay it and release the thread. This means that the next pipeline call can now be made which can cause issues.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: I see what you are saying. I was just wondering if that would possibly affect discovery on shutdown, as that will be for another thread.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: We set a flow control block when delaying a shutdown, so yes, this thread will be released on this call, but the next `onNext` call will be blocked until after the shutdown completes.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Publisher Manager `shutdown` is called from the plugin `stop()` method; while the Future is created in Publisher Manager constructor called from the plugin `start()` method. I cannot think of a scenario with where we would have a cache (or NUMA) coherency issue between those two calls.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: Found and corrected a couple subtle issues related to the penalty box, but the CAS is fine.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I worked through this a couple times.
I do not think there is significant concern here, but perhaps in a future enhancement we will replace the repeated use of the handler entry set with a single pass that updates all the flow state values, then use that for all subsequent iterations.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -310,17 +340,24 @@ void scheduleShutdown() {
...
-            publisherManager.scheduleAfterDelay(this::shutdown, SHUTDOWN_DELAY_TIME);
+            if (scheduleWithDelay) {
+                // Set budget to Long.MIN_VALUE so any subsequent onNext calls park
+                // until shutdown completes. Long.MIN_VALUE is also a signal to the
+                // refresh task to skip this handler.
+                messageBudget.set(Long.MIN_VALUE);
+                // Do not shutdown instantly, allow for final messages to send first.
+                publisherManager.scheduleAfterDelay(this::shutdown, SHUTDOWN_DELAY_TIME);
+            } else {
+                isActive.compareAndSet(true, false); // shouldn't be needed, but set just in case.
```
ata-nas, line 358: Why did this CAS move only to the `else` clause?
Then, if it is truly not needed, shall we simply remove this?
Finally, can we simply move this CAS to the `shutdown` method?
Initially, this value was used simply to schedule a shutdown (the next pipeline call to shutdown the handler).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -310,17 +340,24 @@ void scheduleShutdown() {
         isActive.set(false);
     }
 
-    private void checkMidBlockAndShutdown(final long blockNumber) {
+    /// todo add documentation
+    private void checkMidBlockAndShutdown(final long blockNumber, final boolean scheduleWithDelay) {
```
ata-nas, line 344: We can consider a better name for `scheduleWithDelay`.
Historically, a "scheduled" delay was used to denote that the handler will shutdown itself on the next pipeline call.
This unfortunately conflicts with a "scheduled task". But this delay does not wait for a pipeline call.
We can, perhaps, consider `delayShutdown`?

ata-nas (approved):
Looks good. Unblocking.
Looking forward to seeing how this behaves.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: We _probably_ can remove this, but it hurts nothing to include it.

It moved to the else because it is _definitely_ not needed in the schedule-a-delay case, and setting _before_ the if increases the probability that we receive a message that causes this method to be called twice, on two different threads, with very close timing.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: I think we perhaps change the _other_ meaning of schedule, as this more closely matches a schedule, and that more closely matches a deferral.

berryware (approved):
Looks good.

PR #2918, opened 2026-05-29
fix(solo-e2e-test): CN properties override and improved log collection

### LINKED ISSUE #2899: bug(solo-e2e): seems like all the solo-e2e-tests have been failing since last week, possibly since the WRB addition

AlfredoG87:
Fixes #2899

## Summary

- Add `generate_cn_application_properties` function to `solo-deploy-network.sh` that writes a complete `application.properties` file for Consensus Nodes and passes it to `solo consensus network deploy` via `--application-properties`. The generated file includes all standard CN properties, TSS settings, block stream configuration (`streamMode=BOTH`, `writerMode=FILE_AND_GRPC`), and two new properties: `blockNode.connectionStallThresholdMillis=5000` and `blockStream.streamWrappedRecordBlocks=false`.

- Improve CN log collection in the CI workflow: extract all files from per-node diagnostic zips (not just a handful of BN-relevant ones) while excluding only large state/binary entries (`swirlds-hashstream/`, `saved/*`, `keys/*`, `*.bin`, etc.). Add `du -sh` output alongside `ls -laR` so artifact size regressions are immediately visible. Solo.log continues to be truncated to the last 1000 lines.

### `solo-deploy-network.sh`
- New `generate_cn_application_properties` function generates `out/cn-application.properties` with the full CN property set on each deploy
- `deploy_consensus_nodes` calls the generator and passes `--application-properties` to `solo consensus network deploy`

### `.github/workflows/solo-e2e-test.yml`
- Replace narrow per-file `unzip` extractions with a single broad `unzip -x` that excludes known large/binary entries, retaining all diagnostic logs and config files (block-node-comms, hgcaa, swirlds, grpc-access, journalctl, settingsUsed, data/config)
- Add `du -sh` size summary to the collected logs output step

## **Successful runs:**

**With CN 0.74.0 and BN 0.35.0-SNAPSHOT**
https://github.com/hiero-ledger/hiero-block-node/actions/runs/26611479480

**With CN 0.75.0-rc.3 and BN 0.35.0-SNAPSHOT**
https://github.com/hiero-ledger/hiero-block-node/actions/runs/26613267191/job/78423572967

### DIFF
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-network.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-network.sh
@@ -413,6 +413,37 @@ function deploy_block_nodes {
   OVERLAY_DIR="${overlay_dir}"
 }
 
+function generate_cn_application_properties {
+  local output_file="${1}"
+  cat > "${output_file}" << 'EOF'
+hedera.config.version=0
+ledger.id=0x01
+netty.mode=TEST
+contracts.chainId=298
+hedera.recordStream.logPeriod=1
+balances.exportPeriodSecs=400
+files.maxSizeKb=2048
+hedera.recordStream.compressFilesOnCreation=true
+balances.compressOnCreation=true
+contracts.maxNumWithHapiSigsAccess=0
+autoRenew.targetTypes=
+nodes.gossipFqdnRestricted=false
+hedera.profiles.active=TEST
+nodes.updateAccountIdAllowed=true
+blockStream.streamMode=RECORDS
+# TODO: we can remove this after we no longer need less than v0.59.x
+networkAdmin.exportCandidateRoster=true
+# for v0.59+, write the network.json file when you freeze the network
+networkAdmin.diskNetworkExport=ONLY_FREEZE_BLOCK
+hedera.realm=0
+hedera.shard=0
+nodes.webProxyEndpointsEnabled=true
+nodes.nodeRewardsEnabled=false
+blockNode.connectionStallThresholdMillis=5000
+blockStream.streamWrappedRecordBlocks=false
+EOF
+}
+
 function deploy_consensus_nodes {
   log_line ""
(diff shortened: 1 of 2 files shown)


PavelSBorisov (approved):

tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-network.sh
```diff
@@ -413,6 +413,45 @@ function deploy_block_nodes {
...
+balances.exportPeriodSecs=400
+files.maxSizeKb=2048
+hedera.recordStream.compressFilesOnCreation=true
+balances.compressOnCreation=true
+contracts.maxNumWithHapiSigsAccess=0
+autoRenew.targetTypes=
+nodes.gossipFqdnRestricted=false
+hedera.profiles.active=TEST
+nodes.updateAccountIdAllowed=true
+blockStream.streamMode=BOTH
```
Nana-EC, line 433: If we're doing TSS we should be doing it fully
```suggestion
blockStream.streamMode=BLOCKS
```

Nana-EC, line 433: From 0.75 onwards we need an additional run that is WRB.
You can name it explicitly but we need that for the milestones until we enable TSS.
If this is the default and other versions can override let me know

Nana-EC (commented):
We should set TSS fully default and allow for a WRB instance

Nana-EC (approved):
@AlfredoG87 confirmed he has a follow up planned so this is the base form to unblock

PR #2922, opened 2026-06-01
feat: add S3/HMAC credential support for previewnet storage access

rockysingh:
## Summary

Add support for accessing GCS buckets using HMAC credentials via S3-compatible API, required for previewnet and other networks without service account credentials.

When `GCS_HMAC_ACCESS_ID` and `GCS_HMAC_SECRET` environment variables are present, the system automatically uses S3-compatible access via Hedera's Bucky library. Otherwise, it falls back to MainNetBucket with standard GCS Application Default Credentials.

### New Files
- **BucketLister.java**: Interface for cloud storage abstraction
- **BuckyBucketLister.java**: S3-compatible implementation using Hedera Bucky library

### Modified Files
- **MainNetBucket.java**: Now implements BucketLister interface
- **UpdateDayListingsCommand.java**: Auto-detects HMAC credentials and chooses appropriate implementation
- **build.gradle.kts**: Added Bucky dependency and com.google.auth module requirement
- **gradle/modules.properties**: Added com.google.auth module mapping

## How It Works

The `UpdateDayListingsCommand.createBucketLister()` method checks for HMAC credentials in environment variables:
- If `GCS_HMAC_ACCESS_ID` and `GCS_HMAC_SECRET` are present → uses `BuckyBucketLister`
- Otherwise → uses `MainNetBucket` with Application Default Credentials

## Environment Variables

For previewnet (or any HMAC-based access):
```
export GCS_HMAC_ACCESS_ID="your-access-id"
export GCS_HMAC_SECRET="your-secret-key"
export GCS_BUCKET_NAME="hedera-preview-testnet-streams"
export GCS_ENDPOINT="https://storage.googleapis.com"  # optional, defaults to this
```

## Known Limitation

**Metadata Limitation**: Bucky's \`listObjectsPage()\` API only returns object keys, not metadata (size, etag). Current implementation creates \`ChainFile\` objects with \`size=0\` and \`etag=null\` as placeholders.

**Potential Solutions**:
1. Add AWS S3 SDK dependency (returns full metadata with S3-compatible APIs)
2. Make individual HEAD requests for each file (slow but complete)
3. Determine if size/etag are actually critical for the use case

## Testing

To test with previewnet:

```bash
export GCS_HMAC_ACCESS_ID=\$(jq -r .access_id ~/previewnet/gcp.json)
export GCS_HMAC_SECRET=\$(jq -r .secret ~/previewnet/gcp.json)
export GCS_BUCKET_NAME="hedera-preview-testnet-streams"

java -jar tools.jar days updateDayListings --network previewnet
```

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
@@ -1,0 +1,219 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.tools.utils.s3;
+
+import static org.hiero.block.tools.config.NetworkConfig.current;
+
+import com.hedera.bucky.S3Client;
+import java.util.ArrayList;
+import java.util.List;
+import org.hiero.block.tools.records.ChainFile;
+import org.hiero.block.tools.records.RecordFileDates;
+import org.hiero.block.tools.utils.BucketLister;
+
+/**
+ * S3-compatible bucket lister using Hedera Bucky client.
+ *
+ * <p>Provides file listing capabilities for S3-compatible storage (AWS S3, GCS S3-compatible API,
+ * MinIO, etc.) using HMAC credentials.
+ *
+ * <p>This implementation uses Hedera's Bucky library which provides a lightweight S3 client
+ * that works with any S3-compatible endpoint.
+ */
+public final class BuckyBucketLister implements BucketLister {
+
+    private final S3Client s3Client;
+    private final String bucketName;
+    private final int minNodeAccountId;
+    private final int maxNodeAccountId;
+
+    /**
+     * Creates a new Bucky-based S3 bucket lister.
+     *
+     * @param endpoint S3 endpoint URL (e.g., "https://storage.googleapis.com")
+     * @param accessKey S3/HMAC access key
+     * @param secretKey S3/HMAC secret key
+     * @param bucketName S3 bucket name
+     * @param minNodeAccountId minimum node account ID
+     * @param maxNodeAccountId maximum node account ID
+     */
+    public BuckyBucketLister(
+            final String endpoint,
+            final String accessKey,
+            final String secretKey,
+            final String bucketName,
+            final int minNodeAccountId,
+            final int maxNodeAccountId) {
+        this.bucketName = bucketName;
+        this.minNodeAccountId = minNodeAccountId;
+        this.maxNodeAccountId = maxNodeAccountId;
+
+        try {
+            // Bucky S3Client constructor: S3Client(regionName, endpointUrl, bucketName, accessKey, secretKey)
+            // Region is not critical for GCS S3-compatible API
+            this.s3Client = new S3Client("us-east-1", endpoint, bucketName, accessKey, secretKey);
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to create Bucky S3 client", e);
+        }
+    }
+
+    /**
+     * Creates a Bucky bucket lister from environment variables.
+     *
+     * <p>Reads configuration from:
+     * <ul>
+     *   <li>GCS_ENDPOINT or defaults to "https://storage.googleapis.com"</li>
+     *   <li>GCS_HMAC_ACCESS_ID - required</li>
+     *   <li>GCS_HMAC_SECRET - required</li>
+     *   <li>GCS_BUCKET_NAME - required</li>
+     * </ul>
+     *
+     * @param minNodeAccountId minimum node account ID
+     * @param maxNodeAccountId maximum node account ID
+     * @return configured BuckyBucketLister
+     * @throws IllegalArgumentException if required environment variables are missing
+     */
+    public static BuckyBucketLister fromEnvironment(final int minNodeAccountId, final int maxNodeAccountId) {
+        final String endpoint = System.getenv().getOrDefault("GCS_ENDPOINT", "https://storage.googleapis.com");
+        final String accessKey = System.getenv("GCS_HMAC_ACCESS_ID");
+        final String secretKey = System.getenv("GCS_HMAC_SECRET");
+        final String bucketName = System.getenv("GCS_BUCKET_NAME");
+
+        if (accessKey == null || accessKey.isBlank()) {
+            throw new IllegalArgumentException("Missing required environment variable: GCS_HMAC_ACCESS_ID");
+        }
+        if (secretKey == null || secretKey.isBlank()) {
+            throw new IllegalArgumentException("Missing required environment variable: GCS_HMAC_SECRET");
+        }
+        if (bucketName == null || bucketName.isBlank()) {
+            throw new IllegalArgumentException("Missing required environment variable: GCS_BUCKET_NAME");
+        }
+
+        System.out.println("[BuckyBucketLister] Using S3-compatible access with endpoint: " + endpoint);
+        System.out.println("[BuckyBucketLister] Bucket: " + bucketName);
+
+        return new BuckyBucketLister(endpoint, accessKey, secretKey, bucketName, minNodeAccountId, maxNodeAccountId);
+    }
+
+    @Override
+    public List<ChainFile> listDay(long blockStartTime) {
+        final String datePrefix = RecordFileDates.blockTimeLongToRecordFilePrefix(blockStartTime);
+        // crop to the day (e.g., "2024-01-01")
+        final String dayPrefix = datePrefix.substring(0, datePrefix.indexOf('T'));
+        final String bucketPrefix = current().bucketPathPrefix();
+
+        System.out.println("[BuckyBucketLister] Listing files for day: " + dayPrefix);
+
+        final List<ChainFile> allFiles = new ArrayList<>();
+        int failedNodes = 0;
+
+        // List files for each node
+        for (int nodeId = minNodeAccountId; nodeId <= maxNodeAccountId; nodeId++) {
+            try {
+                // List record files
+                String recordPrefix = bucketPrefix + "record0.0." + nodeId + "/" + dayPrefix;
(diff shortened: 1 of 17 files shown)


rockysingh:
This was tested on previewnet and it works now:

```
user@previewnet-wrapper:~/previewnet$ nohup java -jar tools-0.35.0-SNAPSHOT-all.jar --network previewnet days updateDayListings --listing-dir metadata/listingsByDay > update-day-listings.log 2>&1 &
[1] 93656
user@previewnet-wrapper:~/previewnet$ tail -f update-day-listings.log
nohup: ignoring input
Found 34 missing day(s) to process
First missing: 2026-04-28, Last missing: 2026-05-31
Using nodes 3 to 9


Completed updating 34 day(s) in 11m 45s
```

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/gcp/MainNetBucket.java
```diff
@@ -53,7 +57,76 @@ public class MainNetBucket {
...
+     * Otherwise, falls back to Application Default Credentials.
+     *
+     * @return a configured Storage client
+     */
+    private static Storage createStorageClient() {
+        final String hmacAccessId = System.getenv("GCS_HMAC_ACCESS_ID");
+        final String hmacSecret = System.getenv("GCS_HMAC_SECRET");
+        final String projectId = System.getenv("GCP_PROJECT_ID");
+
+        if (hmacAccessId != null && !hmacAccessId.isBlank() && hmacSecret != null && !hmacSecret.isBlank()) {
```
Nana-EC: Q: Does projectId also need to be checked early for a valid string?

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
```diff
@@ -0,0 +1,202 @@
...
+        final String secretKey = System.getenv("GCS_HMAC_SECRET");
+        final String bucketName = System.getenv("GCS_BUCKET_NAME");
+
+        if (accessKey == null || accessKey.isBlank()) {
+            throw new IllegalArgumentException("Missing required environment variable: GCS_HMAC_ACCESS_ID");
+        }
+        if (secretKey == null || secretKey.isBlank()) {
+            throw new IllegalArgumentException("Missing required environment variable: GCS_HMAC_SECRET");
+        }
+        if (bucketName == null || bucketName.isBlank()) {
```
Nana-EC, line 87: nit: at this point a StringUtils that encompasses the null and blank check might be valuable. Not necessary for this change though

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/gcp/MainNetBucket.java
```diff
@@ -53,7 +57,76 @@ public class MainNetBucket {
...
+            System.out.println("[MainNetBucket] Using HMAC credentials from environment variables");
+
+            try {
+                // Create HMAC credentials using a synthetic service account JSON
+                // The GCS SDK doesn't have a direct HMAC credentials class, so we use this workaround
+                final String serviceAccountJson = String.format(
```
Nana-EC: This might be better as a json file

Nana-EC (commented):
1st pass

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/gcp/MainNetBucket.java
rockysingh: This HMAC code has been removed in f99050a4. MainNetBucket now only uses Application Default Credentials, so no projectId validation is needed here.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
rockysingh: Good suggestion\! I agree a StringUtils helper would be cleaner. I'll note this as a future refactor opportunity, but keeping it as-is for this PR to maintain focus on the HMAC credential support.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/gcp/MainNetBucket.java
rockysingh: Agreed\! That fake service account JSON was problematic. I've removed all the HMAC code from MainNetBucket in f99050a4. The approach is now much cleaner - MainNetBucket only handles Application Default Credentials, and HMAC goes through BuckyBucketLister.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
```diff
@@ -0,0 +1,202 @@
...
+            try {
+                // List record files
+                String recordPrefix = "recordstreams/record0.0." + nodeId + "/" + dayPrefix;
+                allFiles.addAll(listObjectsWithPrefix(recordPrefix, nodeId));
+
+                // List sidecar files
+                String sidecarPrefix = "recordstreams/record0.0." + nodeId + "/sidecar/" + dayPrefix;
+                allFiles.addAll(listObjectsWithPrefix(sidecarPrefix, nodeId));
+            } catch (Exception e) {
+                System.err.println(
```
Nana-EC, line 121: A failure here doesn't alter the behavior. So it'll look like a full days listing was obtained.
You should have some failure criteria here so callers know it failed and that potentially it needs to be called again

Nana-EC: Same issue with `listDayFileNames()`

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
```diff
@@ -0,0 +1,202 @@
...
+    @Override
+    public List<String> listDayFileNames(long blockStartTime) {
+        final String datePrefix = RecordFileDates.blockTimeLongToRecordFilePrefix(blockStartTime);
+        final String dayPrefix = datePrefix.substring(0, datePrefix.indexOf('T'));
+
+        final List<String> allNames = new ArrayList<>();
+
+        for (int nodeId = minNodeAccountId; nodeId <= maxNodeAccountId; nodeId++) {
+            try {
+                String recordPrefix = "recordstreams/record0.0." + nodeId + "/" + dayPrefix;
```
Nana-EC: nit: `recordstreams/` should be pulled from `NetworkConfig.current().bucketPathPrefix()` right?
Works fine for default Hiero cases but anyone testing that changes that will be broken here

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
```diff
@@ -0,0 +1,202 @@
...
+            } catch (Exception e) {
+                System.err.println(
+                        "[BuckyBucketLister] Failed to list file names for node " + nodeId + ": " + e.getMessage());
+            }
+        }
+
+        return allNames;
+    }
+
+    private List<ChainFile> listObjectsWithPrefix(String prefix, int nodeAccountId) {
```
Nana-EC, line 168: nit: both `listObjectsWithPrefix` and `listObjectNamesWithPrefix` loop over the pages.
Could share logic here

Something like
```java
private List<String> listAllKeys(String prefix) {
    final List<String> keys = new ArrayList<>();
    String token = null;
    do {
        S3Client.ListPage page = s3Client.listObjectsPage(prefix, token, null, S3Client.LIST_OBJECTS_MAX);
        keys.addAll(page.keys());
        token = page.continuationToken();
    } while (token != null);
    return keys;
}

private List<ChainFile> listObjectsWithPrefix(String prefix, int nodeAccountId) {
    return listAllKeys(prefix).stream()
            .map(key -> ChainFile.createOrNull(nodeAccountId, key, 0, null))
            .filter(Objects::nonNull)
            .toList();
}

private List<String> listObjectNamesWithPrefix(String prefix) {
    return listAllKeys(prefix);
}
```

Nana-EC (dismissed):
Non blocking suggestions.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
rockysingh: Fixed in b5f57456. Now tracking failed nodes and throwing `RuntimeException` if all nodes fail to list.

rockysingh: Fixed in b5f57456. Applied the same error handling to `listDayFileNames()` - now throws if all nodes fail.

rockysingh: fixed.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
rockysingh: fixed.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/utils/s3/BuckyBucketLister.java
rockysingh: This has been refactored. Good find.

Nana-EC (approved):

jsync-swirlds (approved):

PR #2924, opened 2026-06-01
chore: CloudStorageArchivePlugin Cleanup and Error Handling Improvements

### LINKED ISSUE #1166: `S3ArchivePlugin` Cleanup and Error Handling Improvements

ivannov:
## Reviewer Notes

Resolves six `TODO(1166)` comments left as placeholders in the cloud-storage-archive plugin.

The most significant change is fixing the issue that `UploadResult.FAILED` or throwing an exception would deadlock on next block verification. The old code cleared `currentUploadFuture` and `currentGroupPending` and immediately started a new `BlockUploadTask` from `currentGroupStart`. If the task had already consumed blocks 0 - N from its queue and uploaded them to S3, the new task would wait for block 0 again. The task would deadlock.

Both failure paths (`ExecutionException` and `UploadResult.FAILED`) now call a new `triggerMidRunRecovery()` method. It moves any blocks still in currentGroupPending to blocksStash, initializes the state, and submits a fresh `StartupRecoveryTask`.

New tests are added to cover these paths.

## Related Issue(s)

Closes #1166


block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ata-nas: If the task was interrupted, does that mean we want to interrupt this thread? This is the messaging thread, not sure we want to interrupt that. It might just be that we have a validly canceled task?

If we cannot continue, we must notify the server health plugin (which is not here yet). I just wonder if this thread interruption here is valid, as that would affect this change here. We do not cleanup on interrupt/cancel it seems.

Then, it makes me wonder, who do we expect to throw the interrupted exception?

in the `checkCompletedUpload` method, we have the following guards:

```java
            if (currentUploadFuture.isCancelled()) {
            ...
            }
        ...
        }
```

If a task is canceled, it will throw a cancellation exception (that should be the case even if we throw an exception from the task upon cancellation, I believe).

Then, I think it is a very good idea to save a local copy of the current future so that there is no possibility of the reference to change or become null.

block-node/cloud-storage-archive/src/test/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePluginTest.java
```diff
@@ -1030,6 +1031,98 @@ private void sendVerification(TestBlock block) {
...
+
+            // Simulate blocks 5 and 6 arrived but were not yet drained to the task queue.
+            plugin.currentGroupPending.put(
+                    5L, new BlockWithSource(blocks.get(5).blockUnparsed(), BlockSource.PUBLISHER));
+            plugin.currentGroupPending.put(
+                    6L, new BlockWithSource(blocks.get(6).blockUnparsed(), BlockSource.PUBLISHER));
+
+            // Inject a done-FAILED future so checkCompletedUpload() sees a completed-but-failed task.
+            plugin.currentUploadFuture = CompletableFuture.completedFuture(BlockUploadTask.UploadResult.FAILED);
+
```
ata-nas: These tests are bit fragile.

We seem to be testing internal implementation details, directly manipulating internal state, rather than actually testing the production logic.

ata-nas (commented):
Some questions.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: We don't actually need this exception at all.
Below is a diff that makes two small changes (both of which _do not change semantics_ for the code) that remove the interrupted exception here.
Note, I also remove a catch-and-rethrow instance, which is a code smell.
```diff
diff --git a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
index 369f706d..2d4d38e6 100644
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
@@ -147,54 +147,50 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
     /// {@inheritDoc}
     @Override
     public void handleVerification(VerificationNotification notification) {
         try {
             if (notification != null && notification.success() && notification.block() != null) {
                 // If the task was cancelled (i.e. stop() was called), do not start a new task or process
                 // the current block since the plugin is shutting down.
                 if (!checkCompletedUpload()) {
                     // While recovery is running, stash every block.  currentGroupStart is still -1, so
                     // routeNotification sends it directly to blocksStash without any extra logic.
-                    if (recoveryFuture != null && recoveryFuture.isDone()) {
-                        completeRecovery();
-                    }
+                    completeRecoveryIfReady();
                     final long blockNumber = notification.blockNumber();
                     // Start a new upload task when there is none active
                     if (currentUploadFuture == null && recoveryFuture == null) {
                         startNewUploadTask(blockNumber);
                     }
                     // Route the verified block; if it belongs to the active task's range, attempt to drain
                     // consecutive blocks into the queue. Out-of-range blocks go to the cross-group stash.
                     routeVerifiedBlock(blockNumber, notification);
                 }
             } else {
                 logInvalidOrFailedNotification(notification);
             }
-        } catch (InterruptedException ignore) {
-            Thread.currentThread().interrupt();
         } catch (ExecutionException e) {
             LOGGER.log(WARNING, "Could not complete uploading blocks to cloud archive storage", e);
             metricsHolder.failedTasks().increment();
             // TODO(1166) Handle properly block upload task failure
         }
     }

     /// Checks whether the active [BlockUploadTask] has finished and cleans up its state.
     /// Returns `true` if the task was cancelled, signalling the caller to skip further processing.
-    private boolean checkCompletedUpload() throws ExecutionException, InterruptedException {
+    private boolean checkCompletedUpload() throws ExecutionException {
         boolean cancelled = false;
         if (currentUploadFuture != null && currentUploadFuture.isDone()) {
             if (currentUploadFuture.isCancelled()) {
                 LOGGER.log(TRACE, "Block upload task was cancelled");
                 cancelled = true;
             } else {
-                final UploadResult uploadResult = currentUploadFuture.get();
+                final UploadResult uploadResult = currentUploadFuture.resultNow();
                 if (uploadResult == UploadResult.FAILED) {
                     LOGGER.log(WARNING, "Block upload task failed");
                     metricsHolder.failedTasks().increment();
                     // TODO(1166) Handle properly block upload task failure
                 } else {
                     metricsHolder.successfulTasks().increment();
                 }
                 // Clear the future regardless of outcome so the next block starts a fresh task.
                 currentUploadFuture = null;
                 currentGroupPending = new ConcurrentSkipListMap<>();
@@ -207,47 +203,46 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
     /// sets [currentGroupStart], creates a fresh [BlockingQueue], and submits a [BlockUploadTask]
     /// for the recovered group before replaying any blocks that arrived during recovery.
     ///
     /// When the result is a fresh start (`currentGroupStart == -1`), [currentGroupStart] remains
     /// `-1` and the next call to [handleVerification] will fall through to [startNewUploadTask]
     /// as normal.
     ///
     /// When the result carries an upload ID, [BlockUploadTask] is created with the result so
     /// that it resumes the existing upload and starts its block loop from
     /// [RecoveryResult#nextBlockNumber()] rather than the group start.
-    private void completeRecovery() throws ExecutionException, InterruptedException {
+    private void completeRecoveryIfReady() throws ExecutionException {
         try {
-            final RecoveryResult result = recoveryFuture.get();
-            if (result.currentGroupStart() != -1) {
-                currentGroupSize = Math.powExact(10, config.groupingLevel());
-                currentGroupStart = result.currentGroupStart();
-                nextBlockToQueue = result.uploadId() != null ? result.nextBlockNumber() : currentGroupStart;
-                if (nextBlockToQueue > 0) {
-                    context.applicationStateFacility().addStoredBlockRange(new LongRange(0, nextBlockToQueue - 1));
+            if (recoveryFuture != null && recoveryFuture.isDone()) {
+                final RecoveryResult result = recoveryFuture.resultNow();
+                if (result.currentGroupStart() != -1) {
+                    currentGroupSize = Math.powExact(10, config.groupingLevel());
+                    currentGroupStart = result.currentGroupStart();
+                    nextBlockToQueue = result.uploadId() != null ? result.nextBlockNumber() : currentGroupStart;
+                    if (nextBlockToQueue > 0) {
+                        context.applicationStateFacility().addStoredBlockRange(new LongRange(0, nextBlockToQueue - 1));
+                    }
+                    currentBlockQueue = new LinkedBlockingQueue<>();
+                    currentUploadFuture = virtualThreadExecutor.submit(new BlockUploadTask(
+                            config,
+                            context.blockMessaging(),
+                            currentGroupStart,
+                            currentGroupSize,
+                            currentBlockQueue,
+                            result.uploadId() != null ? result : null,
+                            metricsHolder,
+                            context.applicationStateFacility()));
+                    tryReplayStash();
                 }
-                currentBlockQueue = new LinkedBlockingQueue<>();
-                currentUploadFuture = virtualThreadExecutor.submit(new BlockUploadTask(
-                        config,
-                        context.blockMessaging(),
-                        currentGroupStart,
-                        currentGroupSize,
-                        currentBlockQueue,
-                        result.uploadId() != null ? result : null,
-                        metricsHolder,
-                        context.applicationStateFacility()));
-                tryReplayStash();
+                // No else: currentGroupStart == -1 means a fresh start with no prior S3 state.
+                // currentGroupStart stays -1, so the calling handleVerification picks it up via startNewUploadTask.
             }
-            // No else: currentGroupStart == -1 means a fresh start with no prior S3 state.
-            // currentGroupStart stays -1, so the calling handleVerification picks it up via startNewUploadTask.
-        } catch (ExecutionException e) {
-            LOGGER.log(TRACE, "Startup recovery task failed", e);
-            throw e;
         } finally {
             recoveryFuture = null;
         }
     }

     /// Logs an INFO message explaining why the given [notification] was ignored.  Called only
     /// when the notification is `null`, carries a failed verification, or has a `null` block.
     private void logInvalidOrFailedNotification(VerificationNotification notification) {
         if (notification == null) {
             LOGGER.log(INFO, "Received null verification notification, ignoring");
```

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -338,7 +365,8 @@ public void stop() {
         if (recoveryFuture != null) {
             recoveryFuture.cancel(true);
```
berryware, line 384: Can triggerMidRunRecovery() be called more than once? If so, recoveryFuture is overwritten. This may not be closing all the recoveryFutures

ivannov: Not sure when that could happen. But thanks for catching - I added a guard in the method to avoid that

block-node/cloud-storage-archive/src/test/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePluginTest.java
ivannov: You are right. I tried to make the tests more production like

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Thanks for the insight, didn't know about the `resultNow` method. It doesn't make sense to wait for a future, that we know is done

I applied your diff and also added a couple of trace logs

jsync-swirlds (approved):

rockysingh (approved):
lgtm

AlfredoG87 (approved):
Looking good 💯

PR #2936, opened 2026-06-03
chore: Replace hard coded BlockAccessor.MAX_BLOCK_SIZE_BYTES with protobuf config max message size

### LINKED ISSUE #2707: Replace Block Accessor Constant with Server Config value
### Story Form

N/A

### Technical Notes

There is a constant in BlockAccessor interface that is used _everywhere_.  It should not exist at all, because it is not configurable and we have a configuration value for the same purpose.
This Issue requests that the constant be removed entirely, and the proper configuration be used instead.
This requires method and class constructor changes in Backfill, Stream Subscriber, and Verification.
The impact, including test code, is hundreds of touch-points and nearly every plugin.

Nana-EC:
## Reviewer Notes
Removes the hard-coded `BlockAccessor.MAX_BLOCK_SIZE_BYTES` constant (125 MiB, used across nearly
every plugin) and replaces it with an operator-configurable, node-wide protobuf parse limit sourced
from a new `ProtobufConfig` and applied through the centralized `ProtobufHandler`. Previously the
parse limit was a fixed constant, so raising the message-size config had no effect on block
parsing — operators could not ingest or serve blocks larger than 125 MiB. They can now.

This PR also separates the two concerns that were previously conflated under one value:

- **Parsing** limit → `protobuf.maxMessageSizeBytes` (`ProtobufConfig`), default `131,072,000` (125 MiB).
- **gRPC/HTTP-2 transport** ceiling → `server.maxMessageSizeBytes` (`ServerConfig`), default raised to
  `134,217,728` (128 MiB) so it sits just above the parse limit and absorbs framing/transport overhead.

### Approach

`protobuf-pbj` is depended on by every block-node module (including `spi-plugins`), so the parse limit
lives where everyone can reach it — no constructor/field threading required:

- New **`ProtobufConfig`** (`@ConfigData("protobuf")`) in `app-config` — the single, operator-settable
  source of truth for the parse limit.
- **`ProtobufHandler`** (`protobuf-pbj`, now exported) holds a node-wide `maxMessageSizeBytes()` value,
  defaulted to the canonical value and set once at start-up by `BlockNodeApp` from `ProtobufConfig`.
  It also gains a generic parse helper: `T msg = ProtobufHandler.parse(SomeType.PROTOBUF, bytes)`
  (plus an explicit-size overload), parsing directly into the target type with the node-wide limit.
- **`BlockAccessor`** drops the constant; its default `blockUnparsed()` uses
  `ProtobufHandler.maxMessageSizeBytes()`.
- Every former `MAX_BLOCK_SIZE_BYTES` parse site now calls `ProtobufHandler.maxMessageSizeBytes()`.
- `BlockNodeApp` registers `ProtobufConfig` and pushes its value into `ProtobufHandler` at boot, and
  loads start-up `TssData` via the new generic `ProtobufHandler.parse`.

> Design note: the boot-set static in `ProtobufHandler` is a deliberate trade-off — it removes all
> constructor threading at the cost of a single global value set once at start-up. An alternative
> pure dependency-injection implementation exists on
> `chore/2707-replace-block-size-constant-with-config` for comparison.

### Changes

- `app-config`: **new** `ProtobufConfig`; `ServerConfig.maxMessageSizeBytes` default → `134_217_728`
  (transport ceiling, documented as distinct from the parse limit).
- `protobuf-pbj`: `ProtobufHandler` gains the node-wide size holder + generic `parse(...)`; package exported.
- `app`: `BlockNodeApp` registers `ProtobufConfig`, sets `ProtobufHandler` at boot, loads `TssData`
  via `ProtobufHandler.parse`.
- `spi-plugins`: `BlockAccessor` constant removed; `blockUnparsed()` uses `ProtobufHandler`.
- Parse-site swaps in: `blocks-file-recent`, `blocks-file-historic`, `verification`,
  `block-verification`, `stream-subscriber`, `backfill`, `roster-bootstrap-tss`, and test fixtures.
- `docs/block-node/configuration.md`: documents `PROTOBUF_MAX_MESSAGE_SIZE_BYTES` (parse) and the
  updated `SERVER_MAX_MESSAGE_SIZE_BYTES` (transport), with a TOC entry.
- `tools` `ConvertToJson` (its own unrelated CLI constant) intentionally left untouched.

20 files changed.

### Operator-facing config

| Env var | Concern | Default |
|---|---|---|
| `PROTOBUF_MAX_MESSAGE_SIZE_BYTES` | maximum block/protobuf size the node will **parse** | `131,072,000` |
| `SERVER_MAX_MESSAGE_SIZE_BYTES` | gRPC/HTTP-2 **transport** ceiling (a little larger for framing) | `134,217,728` |

To handle unusually large blocks, raise both (keep transport ≥ parse).

### Testing

- Repo-wide `compileJava` + `compileTestJava` + `spotlessCheck` — green.
- `:app:test` (boots `BlockNodeApp`, exercising `ProtobufConfig` registration + `ProtobufHandler` wiring) — passing.
- Module test suites for all affected modules — passing.
- Confirmed no `MAX_BLOCK_SIZE_BYTES` references remain (outside the unrelated `ConvertToJson` tool).

## Related Issue(s)
Resolves #2707

### DIFF
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ProtobufConfig.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ProtobufConfig.java
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.app.config;
+
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import com.swirlds.config.api.validation.annotation.Max;
+import com.swirlds.config.api.validation.annotation.Min;
+import org.hiero.block.node.base.Loggable;
+
+/**
+ * Configuration for protobuf message handling.
+ *
+ * <p>This is the single, operator-settable source of truth for the maximum protobuf message size
+ * the node will parse. It is read once at start-up and pushed into
+ * {@code org.hiero.block.node.protobuf.ProtobufHandler}, which every module uses when parsing block
+ * data — so raising this value lets a node ingest or serve unusually large blocks (e.g. TSS Wraps
+ * transition blocks) without a code change.
+ *
+ * @param maxMessageSizeBytes the maximum protobuf message size, in bytes, the node will parse
+ */
+@ConfigData("protobuf")
+public record ProtobufConfig(
+        @Loggable @ConfigProperty(defaultValue = "131_072_000") @Min(1_048_576) @Max(1_610_612_736)
+        int maxMessageSizeBytes) {}
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -676,9 +676,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         final Path tssDataJsonPath = appStateConfig.tssBootstrapFilePath();
         if (Files.exists(tssDataJsonPath)) {
             try {
-                TssData tssData = ProtobufHandler.parse(
-                        TssData.JSON,
-                        Bytes.wrap(Files.readAllBytes(tssDataJsonPath)).toReadableSequentialData());
+                TssData tssData = ProtobufHandler.parse(TssData.JSON, Bytes.wrap(Files.readAllBytes(tssDataJsonPath)));
                 updateTssData(tssData);
                 LOGGER.log(INFO, "Loaded TssData from file: {0}", tssDataJsonPath);
             } catch (ParseException | IOException e) {
@@ -699,7 +697,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         if (Files.exists(rsaFilePath)) {
             try {
                 final byte[] raw = Files.readAllBytes(rsaFilePath);
-                final NodeAddressBook book = NodeAddressBook.JSON.parse(Bytes.wrap(raw));
+                final NodeAddressBook book = ProtobufHandler.parse(NodeAddressBook.JSON, Bytes.wrap(raw));
                 validateAddressBook(book, rsaFilePath.toString());
                 pendingAddressBook.set(book);
                 LOGGER.log(
@@ -730,7 +728,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         if (Files.exists(blockRangesPath)) {
             try {
                 final BlockRangesState rangeSet =
-                        BlockRangesState.JSON.parse(Bytes.wrap(Files.readAllBytes(blockRangesPath)));
+                    ProtobufHandler.parse(BlockRangesState.JSON, Bytes.wrap(Files.readAllBytes(blockRangesPath)));
                 rangeSet.storedBlocks().forEach(r -> storedBlocks.add(new LongRange(r.rangeStart(), r.rangeEnd())));
                 rangeSet.availableBlocks()
                         .forEach(r -> availableBlocks.add(new LongRange(r.rangeStart(), r.rangeEnd())));
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -49,6 +49,7 @@ import org.hiero.block.api.BlockRange;
 import org.hiero.block.api.TssData;
 import org.hiero.block.internal.BlockRangesState;
 import org.hiero.block.node.app.config.AutomaticEnvironmentVariableConfigSource;
+import org.hiero.block.node.app.config.ProtobufConfig;
 import org.hiero.block.node.app.config.ServerConfig;
 import org.hiero.block.node.app.config.WebServerHttp2Config;
 import org.hiero.block.node.app.config.node.NodeConfig;
@@ -56,6 +57,7 @@ import org.hiero.block.node.app.config.state.ApplicationStateConfig;
 import org.hiero.block.node.app.logging.CleanColorfulFormatter;
 import org.hiero.block.node.app.logging.ConfigLogger;
 import org.hiero.block.node.base.ranges.ConcurrentLongRangeSet;
+import org.hiero.block.node.protobuf.ProtobufHandler;
 import org.hiero.block.node.spi.ApplicationStateFacility;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodeContext.Builder;
@@ -186,6 +188,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         // Collect all the config data types from the plugins and global server level
         final List<Class<? extends Record>> allConfigDataTypes = new ArrayList<>();
         allConfigDataTypes.add(ServerConfig.class);
+        allConfigDataTypes.add(ProtobufConfig.class);
         allConfigDataTypes.add(WebServerHttp2Config.class);
         allConfigDataTypes.add(NodeConfig.class);
         allConfigDataTypes.add(ApplicationStateConfig.class);
@@ -207,6 +210,10 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         ConfigLogger.log(configuration);
         // now that configuration is loaded we can get config for server
         serverConfig = configuration.getConfigData(ServerConfig.class);
+        // Push the configured protobuf max message size into ProtobufHandler so every module parses
+        // block data with the same operator-configurable bound (see ProtobufConfig).
+        ProtobufHandler.setMaxMessageSizeBytes(
+                configuration.getConfigData(ProtobufConfig.class).maxMessageSizeBytes());
         WebServerHttp2Config webServerHttp2Config = configuration.getConfigData(WebServerHttp2Config.class);
         // ==== METRICS ================================================================================================
         // discover all metrics providers via SPI
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -728,7 +728,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         if (Files.exists(blockRangesPath)) {
             try {
                 final BlockRangesState rangeSet =
-                    ProtobufHandler.parse(BlockRangesState.JSON, Bytes.wrap(Files.readAllBytes(blockRangesPath)));
+                        ProtobufHandler.parse(BlockRangesState.JSON, Bytes.wrap(Files.readAllBytes(blockRangesPath)));
                 rangeSet.storedBlocks().forEach(r -> storedBlocks.add(new LongRange(r.rangeStart(), r.rangeEnd())));
                 rangeSet.availableBlocks()
                         .forEach(r -> availableBlocks.add(new LongRange(r.rangeStart(), r.rangeEnd())));
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -676,7 +676,9 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         final Path tssDataJsonPath = appStateConfig.tssBootstrapFilePath();
         if (Files.exists(tssDataJsonPath)) {
             try {
-                TssData tssData = TssData.JSON.parse(Bytes.wrap(Files.readAllBytes(tssDataJsonPath)));
+                TssData tssData = ProtobufHandler.parse(
+                        TssData.JSON,
+                        Bytes.wrap(Files.readAllBytes(tssDataJsonPath)).toReadableSequentialData());
                 updateTssData(tssData);
                 LOGGER.log(INFO, "Loaded TssData from file: {0}", tssDataJsonPath);
             } catch (ParseException | IOException e) {
--- a/block-node/protobuf-pbj/src/main/java/module-info.java
+++ b/block-node/protobuf-pbj/src/main/java/module-info.java
@@ -68,6 +68,7 @@ module org.hiero.block.protobuf.pbj {
     exports com.hedera.hapi.block.stream.schema;
     exports org.hiero.block.api;
     exports org.hiero.block.internal;
+    exports org.hiero.block.node.protobuf;
 
     requires transitive com.hedera.pbj.runtime;
     requires org.antlr.antlr4.runtime;
--- a/block-node/protobuf-pbj/src/main/java/org/hiero/block/node/protobuf/ProtobufHandler.java
+++ b/block-node/protobuf-pbj/src/main/java/org/hiero/block/node/protobuf/ProtobufHandler.java
@@ -4,6 +4,7 @@ package org.hiero.block.node.protobuf;
 import com.hedera.pbj.runtime.Codec;
 import com.hedera.pbj.runtime.ParseException;
 import com.hedera.pbj.runtime.io.ReadableSequentialData;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
 import edu.umd.cs.findbugs.annotations.NonNull;
 
 /**
@@ -58,7 +59,7 @@ public final class ProtobufHandler {
      * @return the parsed message
      * @throws ParseException if parsing fails
      */
-    public static <T> T parse(@NonNull final Codec<T> codec, @NonNull final ReadableSequentialData input)
+    public static <T> T parse(@NonNull final Codec<T> codec, @NonNull final Bytes input)
             throws ParseException {
         return parse(codec, input, maxMessageSizeBytes);
     }
@@ -75,8 +76,8 @@ public final class ProtobufHandler {
      * @throws ParseException if parsing fails
      */
     public static <T> T parse(
-            @NonNull final Codec<T> codec, @NonNull final ReadableSequentialData input, final int maxMessageSizeBytes)
+            @NonNull final Codec<T> codec, @NonNull final Bytes input, final int maxMessageSizeBytes)
             throws ParseException {
-        return codec.parse(input, false, false, Codec.DEFAULT_MAX_DEPTH, maxMessageSizeBytes);
+        return codec.parse(input.toReadableSequentialData(), false, true, maxMessageSizeBytes / 8, maxMessageSizeBytes);
     }
 }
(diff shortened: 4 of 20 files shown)


block-node/protobuf-pbj/src/main/java/org/hiero/block/node/protobuf/ProtobufHandler.java
jsync-swirlds, line 1: This can go somewhere else, but we deliberately do not have code in this module; it's where the PBJ outputs of the HAPI definition live.

If you want to "centralize" PBJ parsing (something we decided _not to do_ for good reasons, BTW), this would belong in the `base` module.

A better choice; consistent with the "keep plugins independent" design philosophy would be to just copy this file in each plugin that needs it.
- Also, `PbjParsingUtility` would be a better name. It's not really a "Handler", and that term is getting overused; I think Claude likes the word

block-node/protobuf-pbj/src/main/java/module-info.java
```diff
@@ -68,6 +68,7 @@
     exports com.hedera.hapi.block.stream.schema;
     exports org.hiero.block.api;
     exports org.hiero.block.internal;
+    exports org.hiero.block.node.protobuf;
```
jsync-swirlds, line 71: This doesn't seem ideal, in my opinion.
There is significant value in this module _not_ exporting any classes that aren't PBJ-generated code as it keeps the module tightly focused on generated code.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
```diff
@@ -51,7 +45,11 @@ default BlockUnparsed blockUnparsed() {
                 return null;
             }
             return BlockUnparsed.PROTOBUF.parse(
-                    rawData.toReadableSequentialData(), false, false, Codec.DEFAULT_MAX_DEPTH, MAX_BLOCK_SIZE_BYTES);
+                    rawData.toReadableSequentialData(),
+                    false,
+                    false,
+                    Codec.DEFAULT_MAX_DEPTH,
```
jsync-swirlds, line 51: default max depth probably _is not_ the correct value here; the size/8 value is almost certainly more correct.  We just used default here because the PR that added this line didn't have time to work out the "correct" value.

Also, the enclosing method should probably require passing in the maximum size, rather than trying to set it internally.

This is one of the reasons I expected a more difficult process for this PR; the change (if we do this with high quality) cascades through a lot more code than it would seem.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -459,7 +460,7 @@ private void sendHistoricalBlocks() throws ParseException {
                             false,
                             false,
                             Codec.DEFAULT_MAX_DEPTH,
-                            BlockAccessor.MAX_BLOCK_SIZE_BYTES);
+                            ProtobufHandler.maxMessageSizeBytes());
```
jsync-swirlds, line 463: We're relying on static values here.
That's almost certainly not going to work out correctly (because we now have _mutable static data_, which needs to have a write lock).
We need to flow configuration down to this point, not replace one static "cheat" with another static "cheat".

Also, we might prefer to use a Subscriber-specific configuration, or even ask the client to express a maximum size limit (in a future API extension), so clients can inform this plugin of their own limits.

block-node/protobuf-pbj/src/main/java/org/hiero/block/node/protobuf/ProtobufHandler.java
```diff
@@ -0,0 +1,80 @@
...
+public final class ProtobufHandler {
+
+    /**
+     * Default maximum protobuf message size in bytes, equal to the {@code protobuf.maxMessageSizeBytes}
+     * configuration default. Used until {@link #setMaxMessageSizeBytes(int)} is called at start-up.
+     */
+    public static final int DEFAULT_MAX_MESSAGE_SIZE_BYTES = 131_072_000;
+
+    /** The node-wide maximum protobuf message size in bytes, set once at start-up from configuration. */
+    private static volatile int maxMessageSizeBytes = DEFAULT_MAX_MESSAGE_SIZE_BYTES;
```
jsync-swirlds, line 27: both `int` and `volatile` are doing a lot of work here.
We need to document that changing either one will cause unpredictable behavior and likely result in a crash or data loss.

Or not use a static "bypass" to get around flowing configuration through to places that need it.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
```diff
@@ -146,7 +146,7 @@ public BlockNodeClient(
                 "application/grpc",
                 GrpcCompression.IDENTITY,
                 GrpcCompression.getDecompressorNames(),
-                BlockAccessor.MAX_BLOCK_SIZE_BYTES,
+                ProtobufHandler.maxMessageSizeBytes(),
```
jsync-swirlds, line 149: This is a place where local configuration actually makes more sense.  We're not receiving block data here, we're receiving TSS data, which will absolutely never be large (at most it might be a few hundred kilobytes).

Part of why using a constant was such a bad practice is that places like this _really_ should have their own config that sets a much smaller value to avoid accidentally using way too much memory.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -368,13 +368,13 @@ private LedgerIdPublicationTransactionBody findLedgerIdPublication(Bytes signedT
                 false,
                 false,
                 Codec.DEFAULT_MAX_DEPTH,
-                BlockAccessor.MAX_BLOCK_SIZE_BYTES);
+                ProtobufHandler.maxMessageSizeBytes());
```
jsync-swirlds, line 371: Verification is an area where I think we might choose to just use MAX_INT always.
The block has already gone through parsing multiple times, so it's not too big, and just telling the parser "accept anything" is probably the safest choice.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -207,6 +210,10 @@ public BlockNodeApp(final ServiceLoaderFunction serviceLoader, final boolean sho
         ConfigLogger.log(configuration);
         // now that configuration is loaded we can get config for server
         serverConfig = configuration.getConfigData(ServerConfig.class);
+        // Push the configured protobuf max message size into ProtobufHandler so every module parses
+        // block data with the same operator-configurable bound (see ProtobufConfig).
+        ProtobufHandler.setMaxMessageSizeBytes(
+                configuration.getConfigData(ProtobufConfig.class).maxMessageSizeBytes());
```
jsync-swirlds, line 216: This feels not good.
We're replacing a static constant with a slightly different, _and mutable_, static variable.
I believe we're better off if we use the configuration where it's needed, and remove the reliance on static values entirely. There are some remarkably ugly (and hard to find) bugs that show up when relying on _mutable_ static variables across modules (particularly if the deployment uses class data sharing).

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -669,7 +676,7 @@ private void loadApplicationState(final Configuration configuration) {
         final Path tssDataJsonPath = appStateConfig.tssBootstrapFilePath();
         if (Files.exists(tssDataJsonPath)) {
             try {
-                TssData tssData = TssData.JSON.parse(Bytes.wrap(Files.readAllBytes(tssDataJsonPath)));
+                TssData tssData = ProtobufHandler.parse(TssData.JSON, Bytes.wrap(Files.readAllBytes(tssDataJsonPath)));
```
jsync-swirlds, line 679: All of these "application state" loads should use much smaller limits.  The data has no connection to block data sizes and is generally in the tens of kilobytes, not hundreds of megabytes.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ProtobufConfig.java
jsync-swirlds, line 1: We try to place configuration in the same module as the "owning" code.
That's not matching here, which suggests that config, "handler", or both are not in the correct modules.

jsync-swirlds (commented):
This feels like a "fastest possible" way to replace a static _constant_ with a static _variable_.
It feels like an ugly hack rather than a well designed change.
I appreciate the effort, but I'm not convinced this is how we want to approach this issue.

It's been a truism for at least the past 40 years that global variables are not good design; and this creates a global variable.

I won't block this, but I cannot, in good conscience, approve this change either.

---
I suspect we need some rethinking around what should drive protobuf parsing limits, whether we even set a limit in some (many) of these plugins.
For instance, why limit the size when reading a block from disk? That block already been processed by this node and represents internal data. Would you normally place a hard limit on the size of data read from an internal database?

Many of these limits for PBJ parsing should probably be replaced with `Integer.MAX_INT` instead of a configurable value.

Others should probably have _local_ configuration, because the parsed data isn't block data and doesn't need to allow the same sizes.

In at least one case we might allow a _client_ to specify limits so they aren't receiving overlarge blocks. For those clients we might use a filtering process to skip over blocks that are too large, just providing the header, some filtered items, the footer, and the block proof(s).

Nana-EC:
> This feels like a "fastest possible" way to replace a static _constant_ with a static _variable_. It feels like an ugly hack rather than a well designed change. I appreciate the effort, but I'm not convinced this is how we want to approach this issue.
> 
> It's been a truism for at least the past 40 years that global variables are not good design; and this creates a global variable.
> 
> I won't block this, but I cannot, in good conscience, approve this change either.
> 
> I suspect we need some rethinking around what should drive protobuf parsing limits, whether we even set a limit in some (many) of these plugins. For instance, why limit the size when reading a block from disk? That block already been processed by this node and represents internal data. Would you normally place a hard limit on the size of data read from an internal database?
> 
> Many of these limits for PBJ parsing should probably be replaced with `Integer.MAX_INT` instead of a configurable value.
> 
> Others should probably have _local_ configuration, because the parsed data isn't block data and doesn't need to allow the same sizes.
> 
> In at least one case we might allow a _client_ to specify limits so they aren't receiving overlarge blocks. For those clients we might use a filtering process to skip over blocks that are too large, just providing the header, some filtered items, the footer, and the block proof(s).

There's missing context that wasn't apparent when I made the decision to go this way. You're spot on that I was looking for the "fastest possible" whiles also trying to separate streaming from internal parsing whiles also trying to centralize the size value.

I had an initial approach that I shelve once I realize the protobuf-pbj module was a dependency to all the other plugins, it made sense as the home for protobuf related config and logic but it seems we had a reason to avoid all java code in there.

You're also right that this highlights a few places where the default parsing limits may not be the best and we can be more relaxed and considerate.

I'll reply to a few of the comments with any more comments but expect a revision here or a separate PR with my other approach

block-node/protobuf-pbj/src/main/java/org/hiero/block/node/protobuf/ProtobufHandler.java
Nana-EC: Every other plugin takes `protobuf-pbj` as a dependency and if we're looking for a place to centralize some logic is made sense as a consideration.
`base` module is a valid consideration but I'm curious of the good reasons to not centralize PBJ parsing given the considerations of `strictMode`, `parseUnknownFields` and `maxDepth`.
Parsing happens in some many locations, for the core plugins it seems we'd want to take a similar approach on those values at least and reduce the probability of devs setting them wrong.
I like `PbjParsingUtility` better.

Nana-EC: Each plugin specifying it's own pbj config details is certainly a consideration. Maybe it becomes a recommendation also - if your plugin deals with pbj you must add a config for `maxSize`, `strictMode`, `parseUnknownFields` and `maxDepth`

block-node/protobuf-pbj/src/main/java/module-info.java
Nana-EC: This is one consideration that is missing in documentation. Indeed the module is thin but it's not clear it was intended to only ever export pbj-generated code.

I'm unclear on what the downside would be if this also held configs and utilities that dealt with default handling logic for pbj generated classes.
What's the non-ideal consideration here? Is it that modules should rather just specify their own pbj configs and handling logic?

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
Nana-EC: This is the exact reason why I wanted to centralize the parse logic. 

> This is one of the reasons I expected a more difficult process for this PR; the change (if we do this with high quality) cascades through a lot more code than it would seem.

💯

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
Nana-EC: The static data was my least favorite part of this solution, was hoping to think of something better.
I think we've both consider at least twice no that maybe the best solution is just each plugin specify it's own configs for pbj handling.

block-node/protobuf-pbj/src/main/java/org/hiero/block/node/protobuf/ProtobufHandler.java
Nana-EC: Same note as above, I'd hoped this wasn't the final form. I expect this to be removed

block-node/protobuf-pbj/src/main/java/org/hiero/block/node/protobuf/ProtobufHandler.java
jsync-swirlds: Possible.

We've tried to avoid shared code in most plugins, mostly because we want to keep plugins completely independent and easy to replace, as well as avoiding a place where a change we make to improve our own plugins unexpectedly breaks a third-party plugin.

A good example is the local block storage plugins, which replicate block accessor code rather than sharing a common implementation.

The interesting element here is that a lot of these parsing locations _should_ be different parameters. Another is that we are parsing data in some places that shouldn't need to parse the data, and we should probably remove some of those.

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/client/BlockNodeClient.java
Nana-EC: The constant indeed was hiding a lot of local consideration.
The replication of the constant was also avoid sharing.
The inefficiency was hiding us from other problems but also hinting to the need for separate plugin consideration

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ProtobufConfig.java
Nana-EC: I'd initially considered it for the protobuf-pbj module itself.
I think the better approach is adding the config value to each plugin that needs it as the default

block-node/protobuf-pbj/src/main/java/module-info.java
jsync-swirlds: This is a public API module, so it's expected that _any_ entity can depend on this module and treat it as API, meaning it changes rarely, any class in the module can be expected to always be backward compatible, won't change without significant public advance warning, and doesn't include code that isn't intended for use with the public API.

It's arguable whether this class constitutes an API utility or an internal utility.  I tend to think we should start with the class somewhere else and assume it is _not_ an API utility.  We can move it _in_ to this module if we later decide that it is a public API utility far more easily than we can move it _out_ if we make that assumption initially.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
jsync-swirlds: The concern I have with centralizing is that it's assuming every use is the same, which is one of same erroneous assumptions the _constant_ makes (implicitly), while the less centralized the parsing is (within reason), the more closely it can match each use case, and the smaller the "blast radius" if we need to make changes later.

It's always a balance, central to avoid repeated code and repeated avoidable errors, versus dispersed to avoid one-size-fits-all errors and the concurrency difficulties represented by global state.

Nana-EC:
> > This feels like a "fastest possible" way to replace a static _constant_ with a static _variable_. It feels like an ugly hack rather than a well designed change. I appreciate the effort, but I'm not convinced this is how we want to approach this issue.
> > It's been a truism for at least the past 40 years that global variables are not good design; and this creates a global variable.
> > I won't block this, but I cannot, in good conscience, approve this change either.
> > I suspect we need some rethinking around what should drive protobuf parsing limits, whether we even set a limit in some (many) of these plugins. For instance, why limit the size when reading a block from disk? That block already been processed by this node and represents internal data. Would you normally place a hard limit on the size of data read from an internal database?
> > Many of these limits for PBJ parsing should probably be replaced with `Integer.MAX_INT` instead of a configurable value.
> > Others should probably have _local_ configuration, because the parsed data isn't block data and doesn't need to allow the same sizes.
> > In at least one case we might allow a _client_ to specify limits so they aren't receiving overlarge blocks. For those clients we might use a filtering process to skip over blocks that are too large, just providing the header, some filtered items, the footer, and the block proof(s).
> 
> There's missing context that wasn't apparent when I made the decision to go this way. You're spot on that I was looking for the "fastest possible" whiles also trying to separate streaming from internal parsing whiles also trying to centralize the size value.
> 
> I had an initial approach that I shelve once I realize the protobuf-pbj module was a dependency to all the other plugins, it made sense as the home for protobuf related config and logic but it seems we had a reason to avoid all java code in there.
> 
> You're also right that this highlights a few places where the default parsing limits may not be the best and we can be more relaxed and considerate.
> 
> I'll reply to a few of the comments with any more comments but expect a revision here or a separate PR with my other approach

I think the better solution arising is 
- mimic the ServerConfig on a plugin level. Each plugin can have it's own `maxProtobufMessageSizeBytes`. If it does parsing and transport we may need two configs.
- Set local constraints for internal constants that won't need tweaking e.g. AddressBook or TssData parsing are unlikely to change and need optimizing so don't need external configuration ability
- Add consideration for plugin level configuration of the `strictMode`, `parseUnknownFields` and `maxDepth`

Nana-EC:
Closing this PR, taking feedback, my note on the new direction and adding to my initial consideration for PR 2938

PR #2938, opened 2026-06-03
chore: per-plugin protobuf parse limits replacing the shared MAX_BLOCK_SIZE_BYTES

### LINKED ISSUE #2594: Harden CN <> BN protocol for WRB transmission
### Story Form

As a block node engineer
I want to harden the communication protocol
So that I have confidence that even with larger BlockItems from WRB the connections will support continuous streaming of BlockItems

### Technical Notes

Additional consideration will need to be made to support this variant of larger block items.
It's unclear where we may see some latency slow down and how that may affect node communication.
We'll need to analyze and also see where additional test coverage is needed.
Potentially it may be worth expanding the simulator to send test WRBs during 0.75 development until a CN tag with such support is made.

### LINKED ISSUE #2707: Replace Block Accessor Constant with Server Config value

Nana-EC:
### Summary

Removes the hard-coded `BlockAccessor.MAX_BLOCK_SIZE_BYTES` constant (125 MiB, used across nearly
every plugin) and replaces it with **per-plugin** protobuf parse limits chosen by *what is actually
being parsed*. Plugins stay independent — there is **no shared utility and no global state**.

### Approach — limit chosen by data, not a shared value

| Site | Data | Limit |
|---|---|---|
| `BlockAccessor.blockUnparsed()` (SPI default) | a block this node already produced/stored | `Integer.MAX_VALUE` (no constant) |
| blocks-file-recent / blocks-file-historic (disk + archive reads) | block already produced and stored | `Integer.MAX_VALUE` |
| verification + block-verification (`ExtendedMerkleTreeSession`, `AllBlocksHasherHandler`) | block already parsed/validated multiple times | `Integer.MAX_VALUE` |
| consumers reading stored blocks (block-access-service, s3-archive) | trusted internal blocks | `Integer.MAX_VALUE` |
| stream-subscriber | block streamed to a network client | `subscriber.maxProtobufMessageSizeBytes` (config) |
| backfill | block fetched from another BN over the wire | `backfill.maxProtobufMessageSizeBytes` (config) |
| roster-bootstrap-tss client | TSS data (small, never block-sized) | local constant `512 * 1024` (500 KB) |
| `BlockNodeApp` application-state loads (TssData, RSA address book, block ranges) | internal state, ~tens of KB | local constant `1 * 1024 * 1024` (1 MB) |

Trusted/internal reads use `Integer.MAX_VALUE` because the data is something this node produced and
already verified — rejecting it for size would be wrong. Only the two external/transport paths expose
an operator knob.

### Parse flags and depth

All touched parse calls standardize on `strictMode=false`, `parseUnknownFields=true`, and
`maxDepth = size / 8`:

- **`parseUnknownFields=true`** (was `false`) preserves fields from future proto versions so a
  newer-format block round-trips safely instead of being silently truncated — better forward
  compatibility for the wire paths and harmless for internal reads.
- **`maxDepth = size / 8`** replaces `Codec.DEFAULT_MAX_DEPTH`. The `/ 8` is named via a per-class
  constant (`MAX_BLOCK_MESSAGE_DEPTH` / `MAX_MESSAGE_DEPTH` / `MAX_APP_STATE_MESSAGE_DEPTH`) with a
  comment: each nesting level needs ≥ ~8 bytes on the wire, so `size/8` bounds the deepest a
  non-degenerate message can nest.

### New configuration

| Env var | Plugin | Default |
|---|---|---|
| `SUBSCRIBER_MAX_PROTOBUF_MESSAGE_SIZE_BYTES` | stream-subscriber (`SubscriberConfig`) | `131,072,000` |
| `BACKFILL_MAX_PROTOBUF_MESSAGE_SIZE_BYTES` | backfill (`BackfillConfiguration`) | `131,072,000` |

Each is a property on the plugin's own `@ConfigData` record, bounded `@Min(1_048_576) @Max(1_610_612_736)`,
threaded to the parse site by dependency injection (the subscriber captures it once at session
construction, not per block). Each config also exposes a `DEFAULT_MAX_PROTOBUF_MESSAGE_SIZE_BYTES`
constant so tests reference the default rather than copying the literal. `server.maxMessageSizeBytes`
is unchanged and remains the gRPC/HTTP-2 transport ceiling.

### Why no shared `ProtobufHandler` / `PbjParsingUtility`

Surveyed the parse sites (per the review request): each plugin has only 1–3 calls that need an
explicit limit, they share config *within* a plugin, but the correct config *differs across* plugins
(internal reads want "accept anything", transport wants a bound, rosters/state want tiny limits). 

### Notable / for reviewers

- `spi-plugins`: `MAX_BLOCK_SIZE_BYTES` removed; `blockUnparsed()` stays no-arg and parses unbounded
  (trusted internal data).
- Nothing added to `protobuf-pbj` (generated-code only) or `app-config`.
- `tools` `ConvertToJson` keeps its own unrelated CLI constant.

### Testing

- Repo-wide `compileJava` + `compileTestJava` + `spotlessCheck`.
- Module suites for every affected plugin (spi-plugins, blocks-file-recent, blocks-file-historic,
  verification, block-verification, stream-subscriber, backfill, roster-bootstrap-tss,
  roster-bootstrap-rsa, block-access-service, s3-archive, app), and `:app:test` boots `BlockNodeApp`
  with the new configs.
- Confirmed no `MAX_BLOCK_SIZE_BYTES` references remain (outside the `ConvertToJson` tool).

## Related Issue(s)

Resolves #2707
Resolves #2594

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
@@ -46,6 +46,7 @@ public record BackfillConfiguration(
         @Loggable @ConfigProperty(defaultValue = "1000") @Min(500) int perBlockProcessingTimeout,
         @Loggable @ConfigProperty(defaultValue = "60000") @Min(10000) int grpcOverallTimeout,
         @Loggable @ConfigProperty(defaultValue = "104857600") @Min(10_485_760) @Max(314_572_800) int maxIncomingBufferSize,
+        // defaultValue must match DEFAULT_MAX_PROTOBUF_MESSAGE_SIZE_BYTES (annotation requires a String literal)
         @Loggable @ConfigProperty(defaultValue = "131_072_000") @Min(1_048_576) @Max(1_610_612_736) int maxProtobufMessageSizeBytes,
         @Loggable @ConfigProperty(defaultValue = "false") boolean enableTLS,
         @Loggable @ConfigProperty(defaultValue = "false") boolean greedy,
@@ -54,7 +55,10 @@ public record BackfillConfiguration(
         @Loggable @ConfigProperty(defaultValue = "10") @Min(1) @Max(100) int liveTailQueueCapacity,
         // Health scoring constants
         @Loggable @ConfigProperty(defaultValue = "1000.0") double healthPenaltyPerFailure,
-        @Loggable @ConfigProperty(defaultValue = "300000") @Min(30000) long maxBackoffMs) {}
+        @Loggable @ConfigProperty(defaultValue = "300000") @Min(30000) long maxBackoffMs) {
+    /** Default for {@code maxProtobufMessageSizeBytes}; must match the {@code @ConfigProperty(defaultValue = ...)} literal above. */
+    public static final int DEFAULT_MAX_PROTOBUF_MESSAGE_SIZE_BYTES = 131_072_000;
+}
 
 // restore spotless formatting
 // spotless:on
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
@@ -28,6 +28,7 @@ import org.hiero.block.node.base.Loggable;
  *                                  process indefinitely in case something unexpected happens, this would allow for self-recovery
  * @param grpcOverallTimeout single timeout configuration for gRPC Client construction, connectTimeout, readTimeout and pollWaitTime
  * @param maxIncomingBufferSize maximum incoming buffer size in bytes for the gRPC client, must be larger than max block size to account for protobuf metadata and streaming
+ * @param maxProtobufMessageSizeBytes the maximum protobuf message size, in bytes, accepted while parsing a block fetched over the wire during backfill
  * @param enableTLS if enabled will assume block-node client supports tls connection.
  * @param greedy if enabled will search and retrieve blocks beyond latestAcknowledged to ensure BN doesn't fall too far behind.
  */
@@ -45,6 +46,7 @@ public record BackfillConfiguration(
         @Loggable @ConfigProperty(defaultValue = "1000") @Min(500) int perBlockProcessingTimeout,
         @Loggable @ConfigProperty(defaultValue = "60000") @Min(10000) int grpcOverallTimeout,
         @Loggable @ConfigProperty(defaultValue = "104857600") @Min(10_485_760) @Max(314_572_800) int maxIncomingBufferSize,
+        @Loggable @ConfigProperty(defaultValue = "131_072_000") @Min(1_048_576) @Max(1_610_612_736) int maxProtobufMessageSizeBytes,
         @Loggable @ConfigProperty(defaultValue = "false") boolean enableTLS,
         @Loggable @ConfigProperty(defaultValue = "false") boolean greedy,
         // Queue capacity settings for bounded queues
--- a/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
+++ b/block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
@@ -33,24 +33,17 @@ public interface BlockAccessor extends AutoCloseable {
     /**
      * Get the block as an unparsed {@code BlockUnparsed} Java object.
      *
-     * <p>The caller supplies the maximum protobuf message size in bytes; there is intentionally no
-     * shared constant, because the appropriate limit depends on the caller. A consumer reading a
-     * block this node already produced and stored may pass {@link Integer#MAX_VALUE} (the data is
-     * trusted/internal); a consumer parsing block data received over the wire should pass its own
-     * configured limit. The parse depth is derived as {@code maxMessageSizeBytes / 8}.
-     *
-     * @param maxMessageSizeBytes the maximum protobuf message size, in bytes, to accept while parsing
      * @return the block as a {@code BlockUnparsed} Java object, or null if parsing failed.
      *     Also returns null if the data cannot be read from a source.
      */
-    default BlockUnparsed blockUnparsed(final int maxMessageSizeBytes) {
+    default BlockUnparsed blockUnparsed() {
         try {
             final Bytes rawData = blockBytes(Format.PROTOBUF);
             if (rawData == null) {
                 return null;
             }
             return BlockUnparsed.PROTOBUF.parse(
-                    rawData.toReadableSequentialData(), false, true, maxMessageSizeBytes / 8, maxMessageSizeBytes);
+                    rawData.toReadableSequentialData(), false, true, Integer.MAX_VALUE / 8, Integer.MAX_VALUE);
         } catch (final RuntimeException | ParseException e) {
             final System.Logger LOGGER = System.getLogger(getClass().getName());
             LOGGER.log(WARNING, "Failed to parse block", e);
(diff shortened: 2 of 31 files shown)


berryware:
`parseUnknownFields=true (was false) preserves fields from future proto versions so a
newer-format block round-trips safely instead of being silently truncated — better forward
compatibility for the wire paths and harmless for internal reads.`

Isn't the deserializing of unknown fields a security problem? I thought that is why we always chose false. Maybe PBJ doesn't have the same issues as generic protobuf. Here is what Claude says. 

**Memory/DoS risks**
- Unknown fields are preserved in memory by default in proto3. An attacker can send a message stuffed with unknown fields, inflating memory usage and potentially causing OOM or denial of service.
- No size validation happens on unknown fields — they're just buffered as raw bytes.

**Data smuggling / confused deputy**
- If a service deserializes a message, passes it downstream, and the downstream re-serializes it, unknown fields are forwarded transparently. An attacker can embed fields that the upstream ignores but the downstream processes — effectively bypassing upstream validation.
- This is particularly dangerous in proxy/middleware patterns where one service validates and forwards to another with a different proto schema version.

**Schema version confusion**
- A newer-schema field might be "unknown" to an older service but "known" to a downstream newer service. Attackers who know the downstream schema can exploit this gap.

**Canonicalization / signature bypass**
- If you're signing or hashing serialized protobuf messages, unknown fields can be injected after signing. The signer's code ignores them, but a consumer that processes unknown fields sees different effective content than was signed.

**Mitigations**
- Explicitly discard unknown fields at trust boundaries (proto3 parsers can be configured to reject or drop them).
- Validate message size before deserialization.
- Re-serialize → deserialize at trust boundaries to canonicalize (strip unknowns).
- Use `DiscardUnknown` options available in most protobuf libraries.

In the context of this codebase (a block node using gRPC/protobuf for peer communication), the **confused deputy** and **DoS** risks are most relevant — especially since the roster bootstrap work involves accepting protobuf messages from potentially untrusted peers.

jsync-swirlds:
> `parseUnknownFields=true (was false) preserves fields from future proto versions so a newer-format block round-trips safely instead of being silently truncated — better forward compatibility for the wire paths and harmless for internal reads.`
> 
> Isn't the deserializing of unknown fields a security problem? I thought that is why we always chose false. Maybe PBJ doesn't have the same issues as generic protobuf. Here is what Claude says.
> 
`<<SNIP>>`

We have to enable this in general because we promise "forward compatibility", which means we need to preserve fields even if they were added later than the schema we currently recognize.
Absent this setting we would have to require that every block node everywhere always match the latest Consensus Node protobuf schema for the block stream.  Worse, it runs the risk of making older blocks unreadable in a newer block node (which would be fatal to the critical requirement to have all of network history readable without running multiple matching software versions).

We'll still have a validation step that hashes the binary content of each block item, unparsed and including all unrecognized data.  If even a single bit in the entire block is different from what the consensus network agreed to, then it will be detected and the entire block discarded.

Data limits are applied to the unparsed data, there is no bypassing data limits (which, in reality, are necessarily extremely loose).
There is no parse-forward-serialize gap, the _serialized_ data is what is hashed and validated, so unknown fields cannot be abused to bypass verification. All downstream consumers of the data are expected to perform their own validation; the data delivered by block nodes is _not_ considered trustworthy. Blocks are self-contained and easily verified, and every consumer of a block should always validate the block proof(s).
There is no sneaking data through the "schema version gap" because all data is hashed and verified in its serialized form.
There is no injection after signing because the binary data is hashed, not the parsed data.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
```diff
@@ -50,8 +42,11 @@ default BlockUnparsed blockUnparsed() {
             if (rawData == null) {
                 return null;
             }
+            // Trusted internal block — parse unbounded. maxDepth = size / 8: each level of message
+            // nesting needs >= ~8 bytes on the wire, so size/8 bounds the deepest a non-degenerate
+            // message can nest.
             return BlockUnparsed.PROTOBUF.parse(
-                    rawData.toReadableSequentialData(), false, false, Codec.DEFAULT_MAX_DEPTH, MAX_BLOCK_SIZE_BYTES);
+                    rawData.toReadableSequentialData(), false, true, Integer.MAX_VALUE / 8, Integer.MAX_VALUE);
```
AlfredoG87: should have a const MAX_BLOCK_MESSAGE_DEPTH as all other classes?

AlfredoG87 (dismissed):
LG, just a nit can be done in a followup

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
Nana-EC: Added

AlfredoG87 (approved):
LGTM

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
```diff
@@ -45,14 +46,19 @@ public record BackfillConfiguration(
...
         @Loggable @ConfigProperty(defaultValue = "false") boolean greedy,
         // Queue capacity settings for bounded queues
         @Loggable @ConfigProperty(defaultValue = "20") @Min(1) @Max(1000) int historicalQueueCapacity,
         @Loggable @ConfigProperty(defaultValue = "10") @Min(1) @Max(100) int liveTailQueueCapacity,
         // Health scoring constants
         @Loggable @ConfigProperty(defaultValue = "1000.0") double healthPenaltyPerFailure,
-        @Loggable @ConfigProperty(defaultValue = "300000") @Min(30000) long maxBackoffMs) {}
+        @Loggable @ConfigProperty(defaultValue = "300000") @Min(30000) long maxBackoffMs) {
+    /** Default for {@code maxProtobufMessageSizeBytes}; must match the {@code @ConfigProperty(defaultValue = ...)} literal above. */
+    public static final int DEFAULT_MAX_PROTOBUF_MESSAGE_SIZE_BYTES = 131_072_000;
```
jsync-swirlds, line 60: Do we have an issue to remove this value?
The point of having a configuration is that the default can be overridden, but a constant cannot be overridden, so as long as this constant exists we effectively have have a split between configurable and non-configurable values for the same purpose.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -364,17 +365,13 @@ private LedgerIdPublicationTransactionBody findLedgerIdPublication(Bytes signedT
...
+                signedTxBytes.toReadableSequentialData(), false, true, MAX_BLOCK_MESSAGE_DEPTH, Integer.MAX_VALUE);
         TransactionBody body = TransactionBody.PROTOBUF.parse(
                 signedTx.bodyBytes().toReadableSequentialData(),
                 false,
-                false,
-                Codec.DEFAULT_MAX_DEPTH,
-                BlockAccessor.MAX_BLOCK_SIZE_BYTES);
+                true,
+                MAX_BLOCK_MESSAGE_DEPTH,
+                Integer.MAX_VALUE);
```
jsync-swirlds, line 374: This isn't a test file.
We probably need a follow-up PR to make this a configurable value or parameter to the method.

jsync-swirlds (approved):
This is good.
There is more to do to remove the remaining constant usage, but we can do that in a later PR.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillConfiguration.java
Nana-EC: Noted in https://github.com/hiero-ledger/hiero-block-node/issues/2995

PR #2874, opened 2026-05-21
docs(reset-and-upgrade): add Block Node reset and upgrade guide for Kubernetes deployments

deshmukhpranali:
## Reviewer Notes

Adds `docs/block-node/operations/resetting-and-upgrading-the-block-node.md`, a guide covering when and why operators reset or upgrade a Block Node, plus the step-by-step procedure for each on a Kubernetes deployment. The doc also adds a see-also pointer from the existing manual Kubernetes deployment guide.

## Related Issue(s)

Fixes #2855

PR #2940, opened 2026-06-03
docs(configuration): add Plugin Management section (#2736)

### LINKED ISSUE #2736: docs: create upgrade-and-plugin-management.md (0.73 → 0.74 scope)
## User Stories

BN-I2 - Upgrade procedures and plugin management

## Goal

Give Block Node operators a clear, tested upgrade path from 0.73 to 0.74 and a reference for enabling, disabling, and configuring plugins via Helm.

## Contents should cover

- Pre-upgrade checklist: configuration keys to verify, backfill completion status, health check
- Upgrade steps using `task helm-upgrade`
- Post-upgrade validation: how to confirm the node is healthy and receiving blocks
- How to enable and disable plugins via Helm values (referencing `docs/design/deployment-with-selected-plugins.md`)
- Rollback procedure if the upgrade fails
- Known differences between 0.73 and 0.74 that affect operator configuration

Scope this doc to the 0.73 → 0.74 path only. Expand for later releases in Q3.

deshmukhpranali:
## Reviewer Notes
Adds a `Plugin Management` section to `configuration.md`.

**User story: As a Block Node operator, I want a reference for enabling, disabling, and configuring plugins via Helm.**

## Why does this lives in `configuration.md` and not a new file:

Issue #2736 also calls for upgrade procedures. Those are already covered by PR #2874 (`resetting-and-upgrading-the-block-node.md`). The only operator-facing content #2736 asks for that wasn't anywhere is plugin enable/disable via Helm. That belongs alongside the existing per-plugin configuration tables in `configuration.md`.

A cross-link from PR #2874's reset/upgrade doc into this section is included in that PR (one-line blockquote under `## Upgrading the Block Node`).

## Other changes

- **Removed the `## Table of Contents` section** - GitBook auto-generates its own ToC sidebar; the in-doc ToC was duplicative.

## Related Issue(s)
 Fixes #2736

### DIFF
--- a/docs/block-node/configuration.md
+++ b/docs/block-node/configuration.md
@@ -1,29 +1,5 @@
 # Hiero Block Node Configuration Document
 
-## Table of Contents
-
-- [Overview](#overview)
-- [Core Configuration Options](#core-configuration-options)
-- [Server Configuration](#server-configuration)
-- [Application State Configuration](#application-state-configuration)
-- [Metrics Endpoint Configuration](#metrics-endpoint-configuration)
-- [Configurations By Plugin](#configurations-by-plugin)
-  - [Cloud Storage Archive Plugin Configuration](#cloud-storage-archive-plugin-configuration)
-  - [Backfill Plugin Configuration](#backfill-plugin-configuration)
-  - [Block Access Plugin Configuration](#block-access-plugin-configuration)
-  - [Files Historic Plugin Configuration](#files-historic-plugin-configuration)
-  - [Files Recent Plugin Configuration](#files-recent-plugin-configuration)
-  - [Health Plugin Configuration](#health-plugin-configuration)
-  - [Messaging Plugin Configuration](#messaging-plugin-configuration)
-  - [Archive Plugin Configuration (S3 Archive)](#archive-plugin-configuration-s3-archive)
-  - [Server Status Plugin Configuration](#server-status-plugin-configuration)
-  - [Publisher Plugin Configuration](#publisher-plugin-configuration)
-  - [RSA Bootstrap Plugin Configuration](#rsa-bootstrap-plugin-configuration)
-  - [Subscriber Plugin Configuration](#subscriber-plugin-configuration)
-  - [TSS Bootstrap Plugin Configuration](#tss-bootstrap-plugin-configuration)
-  - [Verification Plugin Configuration](#verification-plugin-configuration)
-  - [Cloud Storage Expanded Plugin Configuration](#cloud-storage-expanded-plugin-configuration)
-
 ## Overview
 
 This document outlines configuration options for the Hiero Block Node. Most settings are controlled via environment variables for flexible deployment.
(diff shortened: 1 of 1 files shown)


docs/block-node/configuration.md
```diff
@@ -97,6 +73,45 @@ Prometheus-format metrics. Its properties are prefixed with
 > mapping (`AutomaticEnvironmentVariableConfigSource`). The chart injects them as
 > JVM system properties through `JAVA_TOOL_OPTIONS`.
 
+## Plugin Management
+
+The Block Node is composed of plugins. The Helm chart determines which plugins are loaded into the running pod from the `plugins.*` values block. The next section ([Configurations By Plugin](#configurations-by-plugin)) covers the per-plugin environment-variable settings; this section covers how plugins themselves are selected and added. For the design rationale, see [Deployment with Selected Plugins](../design/deployment-with-selected-plugins.md).
+
+### Built-in plugins
+
+The chart ships with a default set of built-in plugins enabled via `plugins.names`. The value is a comma-separated string of plugin identifiers:
```
Nana-EC:
```suggestion
The chart ships with a default set of built-in plugins focused on a Tier 1 LFH profile enabled via `plugins.names`. The value is a comma-separated string of plugin identifiers:
```

docs/block-node/configuration.md
```diff
@@ -97,6 +73,45 @@ Prometheus-format metrics. Its properties are prefixed with
...
+|------------------------|---------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `plugins.names`        | Comma-separated list of built-in plugin identifiers to load   | `facility-messaging,block-access-service,health,server-status,stream-publisher,stream-subscriber,verification,blocks-file-historic,blocks-file-recent,backfill` |
+| `plugins.mavenImage`   | Container image used to resolve plugin artifacts at pod start | `maven:3-eclipse-temurin-25-alpine`                                                                                                                             |
+| `plugins.repositories` | Maven repositories used to resolve plugin artifacts           | `central` (`https://repo1.maven.org/maven2`), `sonatype-snapshots` (`https://central.sonatype.com/repository/maven-snapshots`)                                  |
+
```
Nana-EC, line 105:
```suggestion

> Note: Most Tier 2 Block Nodes will not require a `stream-publisher` and should consider dropping the `stream-publisher` if not needed.
```

docs/block-node/configuration.md
```diff
@@ -1,29 +1,5 @@
 # Hiero Block Node Configuration Document
 
-## Table of Contents
```
Nana-EC, line 3: Let's keep the table of contents.
The page has lots of configs and it's growing so being able to go to a desired plugin quickly is useful

Nana-EC (commented):
LG, some suggestions

docs/block-node/configuration.md
deshmukhpranali: Hey @Nana-EC, but when we host these documents on GitBook, do we want a double table of contents?

docs/block-node/configuration.md
jsync-swirlds: The term "built-in" is a bit misleading here, as these plugins are not part of the block node; they're just build from the same repo.
A base block node deployment doesn't have any plugins at all; the list in the chart is plugins that are downloaded into the block node container during setup.

We might use "Hiero" or "Default" plugins instead.

docs/block-node/configuration.md
```diff
@@ -97,6 +73,45 @@ Prometheus-format metrics. Its properties are prefixed with
...
+### Built-in plugins
+
+The chart ships with a default set of built-in plugins enabled via `plugins.names`. The value is a comma-separated string of plugin identifiers:
+
+```yaml
+plugins:
+  names: "facility-messaging,block-access-service,health,server-status,stream-publisher,stream-subscriber,verification,blocks-file-historic,blocks-file-recent,backfill"
+```
+
+To enable or disable a built-in plugin, edit the list. Removing a name skips that plugin on the next pod start; adding a name loads it. Operators do not need to rebuild the image — the chart resolves and downloads plugins on pod start using the Maven image configured in `plugins.mavenImage`.
```
jsync-swirlds: This isn't quite accurate. It only removes a plugin if the plugin folder is cleared in between.

Also, there are multiple sources for plugins; not just the image pre-built for local Solo testing.

Most production deployments should be using the plugins.repositories source; not the plugins.mavenImage source. It's recommended that operators use a local mirror, if available.

docs/block-node/configuration.md
jsync-swirlds: Would be slightly more accurate to note that _all_ Tier 2 block nodes do not need a stream-publisher; that plugin is the main difference between Tier 1 and Tier 2.

jsync-swirlds (approved):

docs/block-node/configuration.md
Nana-EC: Did you take this out due to consideration of GitBook?
This is an interesting problem then as readers from the repo or those reading locally would have one experience but readers online on GitBook would experience something else

deshmukhpranali: Yes, I took it out due to consideration of GitBook. Let me think of a better solution for the GitHub and GitBook audience. Once I have the content on GitBook, I will be able to visualise it then work on the solution, hope that is ok.

berryware (approved):
Looks good.

AlfredoG87 (approved):

PR #2942, opened 2026-06-03
feat: Expose available and stored blocks to interested plugins

### LINKED ISSUE #2852: Expose available and stored blocks to interested plugins
### Story Form

As a blog node plugin
I want to be aware of stored and available blocks
So that I track the progress of the block node

### Technical Notes

Consider the two possible approaches:

* Interested parties poll the app state facility via accessor method
* App state facility updates the context similar to other data managed by it

berryware:
## Reviewer Notes
The ServerStatusDetailResponse proto added availableRanges and storedRanges fields to expose full block range information to clients. Previously the server-status plugin could only surface a single contiguous range (first/last available block). This change threads the full range sets through the context so any plugin can access them, and wires them into the detail status response.

- Adds storedBlocks and availableBlocks as List<BlockRange> fields to BlockNodeContext, making block range state available to all plugins via the context
- Updates BlockNodeApp to populate these fields when rebuilding the context, converting from the internal ConcurrentLongRangeSet to the protobuf BlockRange list
- Updates ServerStatusServicePlugin to serve availableRanges and storedRanges in the ServerStatusDetailResponse directly from the context, replacing the previous single-range approach
- Updates test fixtures (PluginTestBase) and all affected tests to pass the two new constructor parameters

## Related Issue(s)
Closes #2852


jsync-swirlds (dismissed):

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -572,12 +581,12 @@ private boolean updateBlockNodeContext(TssData tssData, NodeAddressBook addressB
             builder.nodeAddressBook(addressBook);
         }
 
-        if (storedBlocks != null) {
-            builder.storedBlocks(toBlockRange(storedBlocks));
+        if (storedBlockRange.hashCode() != context.storedBlocks().hashCode()) {
```
jsync-swirlds: Hash code has a decent chance of false match.
We really should be checking equality either directly or after checking `hashCode`, using `Arrays.equals`.

Nana-EC (dismissed):

jsync-swirlds (commented):
One small concern.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Added equality checking

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -555,26 +558,55 @@ private TssData getPendingTssData() {
...
+    private boolean updateBlockNodeContext(
+            TssData tssData, NodeAddressBook addressBook, BlockRangeSet storedBlocks, BlockRangeSet availableBlocks) {
+        BlockNodeContext context = blockNodeContext;
+
+        List<BlockRange> storedBlockRange = toBlockRange(storedBlocks);
+        List<BlockRange> availableBlockRange = toBlockRange(availableBlocks);
+
+        if (tssData == null
+                && addressBook == null
+                && storedBlockRange.hashCode() == context.storedBlocks().hashCode()
```
jsync-swirlds, line 647: Do we need equality checking here too?
If the hashCode just happens to be 19283 for two different array lists; would we miss an important update if this returns false?

ivannov: Maybe: 

```
storedBlockRange.equals(context.storedBlocks())
```

Are we afraid that this is a costly operation?

jsync-swirlds: I do not think so; we already filtered based on hashCode, so it will be called rarely. When it is called, it's fairly similar in magnitude to the cost of `BlockRange.equals`.

ata-nas: Could we simply use ranged sets instead of array lists? In that way we will have a well managed collection. We can still push updates as we do here.

Also, I think, although extremely rarely, hashcode collisions are possible.

berryware: added equality checking. We use the pbj objects directly so that the stored value does not need to be converted to respond to an api call.

jsync-swirlds (dismissed):

protobuf-sources/src/main/proto/block-node/api/node_service.proto
```diff
@@ -53,6 +53,14 @@ message ServerStatusDetailResponse {
      * Node Id, rsa public key information
      */
     proto.NodeAddressBook node_address_book = 4;
+
+    /**
+     * Block availability information.<br/>
```
ivannov: Nit: maybe _availability_ is not the correct word here? We should distinguish this from available ranges

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -628,26 +631,57 @@ private TssData getPendingTssData() {
...
+                && availableBlockRange.hashCode() == context.availableBlocks().hashCode()
+                && storedBlockRange.equals(context.storedBlocks())
+                && availableBlockRange.equals(context.storedBlocks())) {
```
ivannov: I think you should compare available blocks range to the stored blocks? CopyPasteException here? :)

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -628,26 +631,57 @@ private TssData getPendingTssData() {
...
+    private boolean updateBlockNodeContext(
+            TssData tssData, NodeAddressBook addressBook, BlockRangeSet storedBlocks, BlockRangeSet availableBlocks) {
+        BlockNodeContext context = blockNodeContext;
+
+        List<BlockRange> storedBlockRange = toBlockRange(storedBlocks);
+        List<BlockRange> availableBlockRange = toBlockRange(availableBlocks);
+
+        if (tssData == null
+                && addressBook == null
+                && storedBlockRange.hashCode() == context.storedBlocks().hashCode()
```
ivannov, line 647: Do we need these guards once we added the equals checks below?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -628,26 +631,57 @@ private TssData getPendingTssData() {
...
+
+        if (storedBlockRange.hashCode() != context.storedBlocks().hashCode()
+                || !storedBlockRange.equals(context.storedBlocks())) {
+            builder.storedBlocks(storedBlockRange);
+        }
+
+        if (availableBlockRange.hashCode() != context.availableBlocks().hashCode()
+                || !availableBlockRange.equals(context.availableBlocks())) {
+            builder.availableBlocks(availableBlockRange);
+        }
```
ivannov, line 671: Here again, why do we check the hashcode along with equals?

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -114,7 +114,8 @@ public ServerStatusDetailResponse serverStatusDetail(@NonNull final ServerStatus
         }
         // return detailed block node status information.
         return detailsBuilder
-                .availableRanges(blockRanges)
+                .availableRanges(context.availableBlocks())
```
ivannov: So far we've used the block ranges from the historical block provider, which at the moment are the source of truth for available blocks and I think that isn't changed yet.

Maybe we should keep that. And BTW the code that creates and fills `blockRanges` seems to be dead with the change here.

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
```diff
@@ -852,4 +860,39 @@ void blockRangesPersistenceRoundTrip() throws IOException {
 
         app2.stopApplicationStateFacility();
     }
+
+    /**
+     * Test ApplicationStateFacility.
+     */
+    @Test
+    @DisplayName("Test ApplicationStateFacility")
```
ivannov: Maybe a more descriptive display name? E.g.: _Block ranges trigger context update_

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/BlockNodeContext.java
```diff
@@ -49,7 +51,9 @@ public record BlockNodeContext(
         ThreadPoolManager threadPoolManager,
         BlockNodeVersions blockNodeVersions,
         TssData tssData,
-        NodeAddressBook nodeAddressBook) {
+        NodeAddressBook nodeAddressBook,
+        List<BlockRange> storedBlocks,
+        List<BlockRange> availableBlocks) {
```
ivannov, line 58: Maybe add javadoc for the new parameters?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: Yes-ish, it looks like this should be:
```suggestion
                && availableBlockRange.equals(context.availableBlocks())) {
```

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: Yes, the hashcode check avoids the more expensive equals check most of the time (perhaps as much as 90% or more, depends on how good the hashcode function is).

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: To avoid the equals check most of the time.

protobuf-sources/src/main/proto/block-node/api/node_service.proto
berryware: Changed

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Fixed

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
berryware: Updated

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/BlockNodeContext.java
berryware: Added

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: reverted and added a todo

ivannov (dismissed):

Nana-EC (dismissed):

Nana-EC (approved):

ivannov (approved):

PR #2945, opened 2026-06-03
feat: add generate-rsa-roster-bootstrap.sh operator script (#2660)

### LINKED ISSUE #2660: Implement `generate-rsa-roster-bootstrap.sh` operator script

Nana-EC:
Adds a standalone operator script, `tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh`,
that builds a Block Node `rsa-bootstrap-roster.json` by querying a Hedera Mirror Node REST API. This is the
**deploy-time** equivalent of the runtime fallback in `RsaRosterBootstrapPlugin.fetchFromMirrorNode()`: it lets
an operator stage the bootstrap roster on disk in advance so the BN has the `node_id → RSA public key` mapping at
startup and never has to call the Mirror Node to verify WRB (`SignedRecordFileProof`) blocks.

### What it does
- Queries `GET <endpoint>/api/v1/network/nodes?limit=<N>&order=desc` and follows `links.next` pagination
  (resolving root-relative links against the origin).
- Keeps only **active** entries (`timestamp.to` blank/null) that have a **non-empty `public_key`**.
- Maps `node_id → nodeId` and `public_key → RSAPubKey` (strips a leading `0x`, lowercases the hex), de-dupes by
  `nodeId`, and sorts ascending.
- Emits `{ "nodeAddress": [ { "RSAPubKey", "nodeId" } ] }` — the same shape produced by
  `genesisNetworkBNRSARoster.sh` and parsed by the plugin via `NodeAddressBook.JSON`.

This mirrors the field mapping, filtering, and pagination described in
`docs/design/wrb-streaming/bootstrap-roster-plugin.md` section 4.3 and the operator-tooling requirements in section 9.

### Interface
```
generate-rsa-roster-bootstrap.sh (--network <mainnet|testnet|previewnet> | --mirror-node-url <url>) \
    [--output <abs-path>] [--page-size <1..100>] [--help]
```
- `--network` resolves to the public Hedera MN URLs; `--mirror-node-url` takes an explicit endpoint (a trailing
  `/api/v1/network/nodes` is accepted and normalized away).
- Writes JSON to stdout by default, or atomically (temp file + `mv`) to `--output`.
- Prints an entry-count / file-size summary to stderr.

### Testing
- `bash -n` clean.
- All argument-validation paths return exit 2 with clear errors (missing/both sources, unknown network, bad URL
  scheme, out-of-range page size, relative `--output`).
- **Deterministic mock Mirror Node** (two pages with a relative `links.next`): verified pagination, `0x`-strip +
  lowercase, exclusion of a superseded entry (`timestamp.to` set) and a no-`public_key` node, and ascending sort.
- **Live `https://testnet.mirrornode.hedera.com/`** run (page-size 5 to exercise pagination): returned the 7
  active testnet nodes; the output for nodeIds **4, 5, 6 is byte-identical** to the existing fixture
  `block-node/app/src/test/resources/data/config/rsa-bootstrap-roster.json` (canonicalized `diff` clean).

### Notes for reviewers
- Output uses standard `jq` formatting (`"key": value`); the PBJ fixture uses `"key" : value` (space before colon).
  Keys/values/order/encoding are identical and the BN parses via `NodeAddressBook.JSON.parse`, so the difference is
  cosmetic. Can switch to PBJ-exact spacing if a golden-file byte match is desired.
- The script collects across all pages and filters on `timestamp.to` blank (robust), rather than the plugin's
  early-break-on-first-historical optimization. Net result is the same active roster; `/api/v1/network/nodes`
  normally returns only current entries.

## Related Issue(s)
Resolves #2660

### DIFF
--- a/tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
+++ b/tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
@@ -22,9 +22,7 @@
 # historical entries (timestamp.to set) are excluded. Nodes without a
 # public_key are excluded — they cannot participate in RSA WRB verification.
 #
-# This output matches both:
-#   * genesisNetworkBNRSARoster.sh (the genesis-network.json -> BN converter), and
-#   * NodeAddressBook.JSON as parsed by RsaRosterBootstrapPlugin at startup.
+# This output matches NodeAddressBook.JSON as parsed by RsaRosterBootstrapPlugin at startup.
 #
 # The script makes no assumptions about its own location or the current working
 # directory. It writes to stdout by default, or to the path given via -o/--output.
--- a/tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
+++ b/tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
@@ -1,0 +1,290 @@
+#!/usr/bin/env bash
+#
+# generate-rsa-roster-bootstrap.sh
+#
+# Build a hiero-block-node rsa-bootstrap-roster.json by querying a Hedera
+# Mirror Node (MN) REST API. This is the operator-side, deploy-time equivalent
+# of the runtime fallback in RsaRosterBootstrapPlugin.fetchFromMirrorNode():
+# it lets an operator generate the bootstrap roster file in advance so the
+# Block Node has it on disk at startup and never has to call the Mirror Node.
+#
+#   Source (MN):  GET <endpoint>/api/v1/network/nodes
+#                 -> nodes[].{node_id, public_key, timestamp:{from,to}}
+#                    + links.next (pagination)
+#
+#   Output (BN):  rsa-bootstrap-roster.json
+#                 -> { nodeAddress: [ { RSAPubKey, nodeId } ] }
+#                    where RSAPubKey is the lowercase hex public key with any
+#                    leading "0x" stripped, sorted ascending by nodeId.
+#
+# Only the *active* roster is captured: entries whose timestamp.to is blank
+# (null/empty) are the currently-active address-book entries; superseded
+# historical entries (timestamp.to set) are excluded. Nodes without a
+# public_key are excluded — they cannot participate in RSA WRB verification.
+#
+# This output matches both:
+#   * genesisNetworkBNRSARoster.sh (the genesis-network.json -> BN converter), and
+#   * NodeAddressBook.JSON as parsed by RsaRosterBootstrapPlugin at startup.
+#
+# The script makes no assumptions about its own location or the current working
+# directory. It writes to stdout by default, or to the path given via -o/--output.
+#
+# ===========================================================================
+# RUNBOOK — generating the RSA bootstrap roster before a Phase 2a cutover
+# ===========================================================================
+#
+# Before Consensus Nodes begin streaming Wrapped Record Blocks (WRBs), each
+# Block Node must know the current node_id -> RSA-public-key mapping so it can
+# verify SignedRecordFileProof block proofs. Pre-staging the file avoids a
+# Mirror Node round-trip (and its failure modes) at BN startup.
+#
+#   1. Pick the network (or an explicit MN URL):
+#        generate-rsa-roster-bootstrap.sh --network mainnet \
+#          --output /opt/hiero/block-node/node/rsa-bootstrap-roster.json
+#
+#      or point at a specific Mirror Node:
+#        generate-rsa-roster-bootstrap.sh \
+#          --mirror-node-url https://mainnet-public.mirrornode.hedera.com \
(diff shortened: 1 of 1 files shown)


AlfredoG87 (approved):
LGTM

tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
```diff
@@ -0,0 +1,288 @@
...
+#
+# generate-rsa-roster-bootstrap.sh
+#
+# Build a hiero-block-node rsa-bootstrap-roster.json by querying a Hedera
+# Mirror Node (MN) REST API. This is the operator-side, deploy-time equivalent
+# of the runtime fallback in RsaRosterBootstrapPlugin.fetchFromMirrorNode():
+# it lets an operator generate the bootstrap roster file in advance so the
+# Block Node has it on disk at startup and never has to call the Mirror Node.
+#
+#   Source (MN):  GET <endpoint>/api/v1/network/nodes
```
berryware, line 11: The gRPC service can return all the nodes in one call. Perhaps grpcurl could be used to simplify the script.  [network_service.proto](https://github.com/hiero-ledger/hiero-mirror-node/blob/main/protobuf/src/main/proto/com/hedera/mirror/api/proto/network_service.proto)

Nana-EC: Good insight.
I forgot about that endpoint.
In that case instead of jq a node operator would have to have grpcurl installed.

berryware (approved):
Looks good

tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
```diff
@@ -0,0 +1,288 @@
+#!/usr/bin/env bash
```
jsync-swirlds, line 1: Missing SPDX identifier.
`# SPDX-License-Identifier: Apache-2.0`

tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
```diff
@@ -0,0 +1,288 @@
...
+    -o|--output)         [[ $# -ge 2 ]] || { echo "Error: $1 requires an argument" >&2; usage >&2; exit 2; }; OUTPUT="$2"; shift 2 ;;
+    -l|--page-size)      [[ $# -ge 2 ]] || { echo "Error: $1 requires an argument" >&2; usage >&2; exit 2; }; PAGE_SIZE="$2"; shift 2 ;;
+    -h|--help)           usage; exit 0 ;;
+    --) shift; break ;;
+    -*) echo "Error: unknown option $1" >&2; usage >&2; exit 2 ;;
+    *)  echo "Error: unexpected positional argument: $1" >&2; usage >&2; exit 2 ;;
+  esac
+done
+if [[ $# -gt 0 ]]; then
+  echo "Error: unexpected positional argument(s): $*" >&2
```
jsync-swirlds, line 130: nitpick: These are flag arguments, not positional (positional means order matters).

```suggestion
    *)  echo "Error: unexpected flag argument: $1" >&2; usage >&2; exit 2 ;;
  esac
done
if [[ $# -gt 0 ]]; then
  echo "Error: unexpected flag argument(s): $*" >&2
```

tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
Nana-EC: Will evaluate with https://github.com/hiero-ledger/hiero-block-node/issues/2994

tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
Nana-EC: Will address in https://github.com/hiero-ledger/hiero-block-node/issues/2994

tools-and-tests/scripts/node-operations/generate-rsa-roster-bootstrap.sh
Nana-EC: Will address in https://github.com/hiero-ledger/hiero-block-node/issues/2994

PR #2946, opened 2026-06-03
feat: add genesis-network-to-bn-rsa-roster.sh operator script

Nana-EC:
Adds a standalone operator script, `tools-and-tests/scripts/node-operations/genesis-network-to-bn-rsa-roster.sh`,
that converts a hiero-consensus-node `genesis-network.json` into the hiero-block-node `rsa-bootstrap-roster.json`
shape. This is the **CN / genesis-file** path for producing the BN's `node_id → RSA public key` bootstrap roster,
complementing the Mirror Node path in `generate-rsa-roster-bootstrap.sh`. Both emit the same output shape, which the
Block Node loads at startup via `NodeAddressBook.JSON` to verify WRB (`SignedRecordFileProof`) blocks.
Required in test cases where MN is not running but you have access to `genesis-network.json`

### What it does
- **Input (CN):** `nodeMetadata[].rosterEntry.{nodeId, gossipCaCertificate}`, where `gossipCaCertificate` is a
  base64-encoded X.509 DER certificate.
- For each entry: base64-decode the cert → extract the public key → re-encode as DER `SubjectPublicKeyInfo` (PKIX) →
  lowercase hex (`base64 -d | openssl x509 -pubkey | openssl pkey -outform DER | xxd -p`).
- Verifies each key is RSA by asserting the `rsaEncryption` OID (`06092a864886f70d010101`) is present in the SPKI;
  fails fast otherwise.
- **Output (BN):** `{ "nodeAddress": [ { "RSAPubKey", "nodeId" } ] }`, sorted ascending by `nodeId` — identical to
  the shape produced by `generate-rsa-roster-bootstrap.sh` and parsed by `RsaRosterBootstrapPlugin`.

### Interface
```
genesis-network-to-bn-rsa-roster.sh -i <input> [-o <output>] [-h]
  -i <input>   Absolute path to genesis-network.json (required)
  -o <output>  Absolute path to write rsa-bootstrap-roster.json (default: stdout)
  -h           Help
```

### Testing
- `bash -n` clean. (shellcheck not available in my environment — worth a CI/local shellcheck pass.)
- **Functional:** generated two self-signed RSA certs, base64-DER-encoded them into a `genesis-network.json` with
  `nodeId`s deliberately out of order (7, 3), and ran the conversion. Result: 2 entries, **sorted ascending (3, 7)**,
  all `RSAPubKey` values lowercase hex with the rsaEncryption OID present.
- **Error paths:** empty `nodeMetadata` → exit 1 with a clear message; relative `-i` path → exit 2; missing/invalid
  input and non-RSA certs are rejected with explicit errors.

### Notes for reviewers
- Output uses standard `jq` formatting (`"key": value`); the PBJ fixture
  `block-node/app/src/test/resources/data/config/rsa-bootstrap-roster.json` uses `"key" : value` (space before
  colon). Keys/values/order/encoding are identical and the BN parses via `NodeAddressBook.JSON.parse`, so the
  difference is cosmetic.

### DIFF
--- a/tools-and-tests/scripts/node-operations/genesis-network-to-bn-rsa-roster.sh
+++ b/tools-and-tests/scripts/node-operations/genesis-network-to-bn-rsa-roster.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
 #
 # genesis-network-to-bn-rsa-roster.sh
 #
--- a/tools-and-tests/scripts/node-operations/genesis-network-to-bn-rsa-roster.sh
+++ b/tools-and-tests/scripts/node-operations/genesis-network-to-bn-rsa-roster.sh
@@ -1,0 +1,196 @@
+#!/usr/bin/env bash
+#
+# genesis-network-to-bn-rsa-roster.sh
+#
+# Convert a hiero-consensus-node genesis-network.json into the
+# hiero-block-node rsa-bootstrap-roster.json shape.
+#
+#   Input  (CN):  nodeMetadata[].rosterEntry.{nodeId, gossipCaCertificate}
+#                 where gossipCaCertificate is a base64-encoded X.509 DER cert.
+#
+#   Output (BN): nodeAddress[].{RSAPubKey, nodeId}
+#                 where RSAPubKey is the lowercase-hex SubjectPublicKeyInfo
+#                 (PKIX) DER bytes of the cert's public key.
+#
+# The script makes no assumptions about its own location or the current
+# working directory. The caller passes the absolute path to the
+# genesis-network.json via -i; the output path (optional) goes via -o.
+
+set -euo pipefail
+# Locale-fixed for deterministic byte handling in tr/sort/openssl.
+export LC_ALL=C
+
+usage() {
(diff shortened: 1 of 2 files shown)


AlfredoG87 (dismissed):
LGTM

tools-and-tests/scripts/node-operations/genesis-network-to-bn-rsa-roster.sh
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env bash
```
jsync-swirlds, line 1: Missing SPDX identifier.
`# SPDX-License-Identifier: Apache-2.0`

Nana-EC: Will address in https://github.com/hiero-ledger/hiero-block-node/issues/2994

Nana-EC: Actually went ahead and addressed now

jsync-swirlds (approved):

AlfredoG87 (approved):
LG 💯

PR #2952, opened 2026-06-04
docs: update the design doc `bootstrap-roster-plugin.md

### LINKED ISSUE #2793: Bootstrap RSA Roster Plugin — Design Improvements (post-merge review feedback)
## Story Form

As a Block Node developer,
I want the Bootstrap RSA Roster Plugin design document and implementation to incorporate all post-merge feedback from jsync-swirlds,
So that the plugin is built with correct assumptions about address book lifecycle, configuration namespace, security model, and operational behaviour before implementation work in #2561 is completed.

## Technical Notes

These items originate from the post-merge review thread on PR #2631 by [@jsync-swirlds](https://github.com/jsync-swirlds) (review [4209380029](https://github.com/hiero-ledger/hiero-block-node/pull/2631#pullrequestreview-4209380029)). All items were validated against `main` on 2026-05-12 and remain unaddressed — the design doc and the plugin stub are unchanged.

### Security Section (§8)

- [x] **Replay attack section (§8.2) should be removed or simplified.** jsync: *"We actually don't care, at all, about replay. If a publisher sends the same block 100 times then we will keep verifying it until it's persisted, then respond with 'duplicate' for all the rest."* The section overstates the risk; block-number deduplication is handled upstream. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171582684)

- [x] **Bootstrap file integrity risk is overstated (§8.1).** jsync: *"True, but tampering requires filesystem access, and at that point an attacker could inject any content they like without forging proofs. If the BN is compromised to the point of arbitrary filesystem access then the system is already irrecoverable."* Simplify §8.1 to remove the threat model that assumes partial filesystem compromise. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171587284)

- [x] **Drop the SHA-256 checksum deferred item entirely.** jsync: *"This has very limited value, and is costly in terms of processing. Also, anything that could replace the block file could replace the checksum, so it is highly questionable there is any value here."* Remove the checksum mitigation from §8.1 and the deferred-items table (Open Question #2). [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171590085) [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171610873)

### Operator Tooling (§9)

- [x] **Check whether Mirror Node offers a single-call address book endpoint.** jsync: *"Is there a mirror-node API to get the full address book as a single JSON response? That would be simpler; storing the whole address book is simpler and not much different for data size."* Investigate `GET /api/v1/network/nodes` pagination vs any bulk alternative; document the decision in §9. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171595923)

- [x] **Note that the bootstrap file format is synthetic and may have a limited lifetime.** jsync: *"Is it worth noting here that this 'file' is completely synthetic, is generated by the CN right after upgrade, and might not be generated much longer?"* Add a lifecycle note to §9 clarifying that the pb format is a Phase 2a transitional artifact. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171606574)

- [x] **Operators should not need to restart to update the address book.** jsync: *"This shouldn't be necessary. The address book (if managed by ApplicationStateFacility) can be updated dynamically and we probably want this plugin to do so (likely by querying MN daily or hourly to get the last update timestamp)."* Remove the "regenerate and restart" requirement from §9 in favour of periodic MN polling. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171603754)

- [x] **Include generation timestamp in the bootstrap file (or a sidecar).** jsync: *"We should probably keep the generation timestamp, so we know when we get newer data."* The design doc currently states the file embeds no metadata; update to include a timestamp so the plugin can compare freshness against MN data. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171990537)

### Configuration (§6)

- [x] **Move config properties to `app.state` namespace.** jsync: *"This should be in the application state facility config namespace."* The current namespace is `roster.bootstrap.rsa.*`; evaluate whether these properties belong under `app.state.*` since `ApplicationStateFacility` owns address book persistence. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171598614)

### Startup Sequence & Address Book Lifecycle (§4.4)

- [x] **Roster should not be limited to load-once at startup.** jsync: *"I'm not sure why we'd limit the roster to only loading on start. We can update from mirror node while running without any difficulty; and even if we do not choose to do so initially, I don't think it's wise to design around a non-changing roster. We should assume that it might change and design to be data-safe if it does."* Revise the "No mid-instance reload" stance to at minimum allow for future-safe design. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171980723)

- [x] **Even when a local bootstrap file exists, still query Mirror Node for updates.** jsync: *"I would still query mirror node, for updates in case the data was updated after the file was written."* The current design skips the MN call if the file exists. Update §4.4 to describe a file-load + MN-freshness-check sequence. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171985382)

- [x] **Check address book hourly (not only at startup).** jsync: *"We should probably check address book roughly hourly (much as SDKs do). Waiting for an upgrade is not easy for BN to manage, and there is little reason not to check more often."* Update Open Question #6 (currently resolved as "No, not initially") to reflect this direction. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171614754)

- [x] **Prefer retry-with-backoff over fail-fast when Mirror Node is unreachable.** jsync: *"I would recommend retrying the mirror node every 5 minutes or so until it succeeds rather than failing the node entirely. We might need to defer that until we have plugin health reporting, though."* Update Open Question #1 and the fail-fast rationale in §4.4. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171982372)

- [x] **Add a note that fail-fast behaviour should evolve once plugin health reporting is available.** jsync: *"Should we include a note that this should change when plugin health reporting is available? I don't think we want hard failures if we can avoid them."* [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171984026)

- [x] **Nodes with no public key should never exist — clarify filter vs failure treatment.** jsync: *"Do we filter these, or do we count them as failed verifications? Honestly, there should never be an entry with no public key."* The design currently says "filter them silently"; clarify whether a missing key triggers a metric or a warning. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171988606)

### RSA Verification (§4.5)

- [x] **Clarify that ApplicationStateFacility (not the plugin directly) supplies NodeAddressBook.** jsync suggestion: *"The `ApplicationStateFacility` supplies the `NodeAddressBook` to plugins via `onContextUpdate`; the verification service reads it from `BlockNodeContext.nodeAddressBook()`."* Update the proof-type routing description in §4.5. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171969672)

- [x] **Document proof validation priority when multiple block proofs are present.** jsync: *"Perhaps note here that Verification will attempt to validate multiple block proofs if multiple proofs are available? We still need to work out the priority for that; I suspect it will need to change a few times as the network completes the shift from records to TSS in stages."* [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171977057)

### Phase 2b Transition (§4.6)

- [x] **Update wording: WRBs have no state data at all** (not just "will not contribute to state services"). jsync: *"Not just 'will not contribute', WRBs have no state data at all."* Suggested text: *"Notably WRBs have no state data or event data and will not support state management services."* [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171674279)

### Open Questions (§12)

- [x] **Resolve plugin lifetime question (Open Question #7).** jsync: *"We can definitely remove this plugin once WRBs all have block proofs generated with a TSSSignature (StateProof type). After that there is no generally applicable reason for a block node to try to manage the old address books."* The current answer says the plugin should remain indefinitely; update to reflect jsync's view that it can be retired post-TSS transition. [comment](https://github.com/hiero-ledger/hiero-block-node/pull/2631#discussion_r3171619038)

---

## References

- Source PR: [#2631](https://github.com/hiero-ledger/hiero-block-node/pull/2631)
- Post-merge review: [jsync-swirlds review 4209380029](https://github.com/hiero-ledger/hiero-block-node/pull/2631#pullrequestreview-4209380029)
- Design doc: `docs/design/wrb-streaming/bootstrap-roster-plugin.md`
- Plugin stub: `block-node/roster-bootstrap-rsa/src/main/java/.../RsaRosterBootstrapPlugin.java`
- Depends on: [#2561](https://github.com/hiero-ledger/hiero-block-node/issues/2561) (plugin implementation)

berryware:
## Reviewer Notes

- §4.4 Startup Sequence (renamed to "Startup Sequence & Address Book Lifecycle") — Replaced the load-once / fail-fast flow with a file-load + MN-freshness-check model: the plugin always queries MN even when a file exists, schedules 2 configurable background polls, one interval when no address book exists, and a longer interval once an address book exists.
- §4.5 RSA Signature Verification — Clarified that ApplicationStateFacility (not the plugin directly) supplies NodeAddressBook to downstream services. Added a "Multiple proofs" callout documenting that the verification layer should attempt all available proof types and that priority rules are an open question (§12 #7).
- §4.6 Phase 2b Transition — Fixed wording to "WRBs have no state data or event data and will not support state management services."
- §5 Mermaid diagram — Updated to reflect the new startup sequence (always-query-MN, freshness comparison, retry paths, periodic poll).
- §6 Configuration — Moved all properties to the app.state namespace; added app.state.rsaAddressBookPollIntervalMinutes for the background poll interval.
- §8 Security — Simplified §8.1 (removed overstated partial-compromise threat model and the SHA-256 checksum mitigation). Removed §8.2 (Replay attack) entirely.
- §9 Operator Tooling — Added investigation note for a bulk MN endpoint, added generation-timestamp sidecar (step 4), removed the "restart required" guidance, added the Phase 2a lifecycle note.
- §10 Acceptance Tests — Updated all .pb references to .json, replaced the fail-fast test with retry-with-backoff tests, added the "file updated when MN book differs" case, and clarified metric names.
- §12 Open Questions — Removed Q#2 (checksum, closed as not worth doing). Resolved Q#1 (retry-not-fail-fast), Q#5/6 (hourly polling), Q#7 (plugin retirement post-TSS). Added Q#7 for multi-proof priority ordering.

## Related Issue(s)
Closes: #2793

### DIFF
--- a/docs/design/wrb-streaming/bootstrap-roster-plugin.md
+++ b/docs/design/wrb-streaming/bootstrap-roster-plugin.md
@@ -11,13 +11,14 @@
    - 4.1 [Plugin Structure](#41-plugin-structure)
    - 4.2 [Bootstrap File Format](#42-bootstrap-file-format)
    - 4.3 [Mirror Node Fetch](#43-mirror-node-fetch)
-   - 4.4 [Startup Sequence](#44-startup-sequence)
+   - 4.4 [Startup Sequence & Address Book Lifecycle](#44-startup-sequence--address-book-lifecycle)
    - 4.5 [RSA Signature Verification](#45-rsa-signature-verification)
    - 4.6 [Phase 2b Transition](#46-phase-2b-transition)
 5. [Diagram](#5-diagram)
 6. [Configuration](#6-configuration)
 7. [Metrics](#7-metrics)
 8. [Security](#8-security)
+   - 8.1 [Bootstrap File Integrity](#81-bootstrap-file-integrity)
 9. [Operator Tooling](#9-operator-tooling)
 10. [Acceptance Tests](#10-acceptance-tests)
 11. [Follow-on Ticket Mapping](#11-follow-on-ticket-mapping)
@@ -368,10 +369,14 @@ after retries the plugin logs a clear error and fails fast (see §4.4 for MN-dow
 
 ---
 
-### 4.4 Startup Sequence
+### 4.4 Startup Sequence & Address Book Lifecycle
 
-The plugin follows a **file-first** strategy consistent with `TssBootstrapPlugin`. The file check and any network call
-happen in `start()`, after `init()` has stored the facility reference.
+The plugin follows a **file-load + MN-freshness-check** strategy. If a local bootstrap file exists it is loaded first
+so the BN has an immediately usable address book. The Mirror Node is then queried (when configured) to check for
+updates. If the file does not exist, the MirrorNode is queried at the `initialQueryIntervalMillis` interval until an
+address book is returned. This allows for Mirror Nodes being started after Block Nodes. The Mirror Node
+will be queried at the `subsequentQueryIntervalMillis` interval if a file is found or when an initial query returns an
+address book.
 
 ```
 BlockNodeApp.loadApplicationState() [before any plugin is started]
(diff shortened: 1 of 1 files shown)


AlfredoG87 (approved):
LGTM

berryware:
Closing and adding to PR #2979

PR #2962, opened 2026-06-04
refactor: consolidate BlockNodeClient into block-node/base

### LINKED ISSUE #2800: Make BlockNodeClient a shared class for all plugins
### Story Form

As a blocknode engineer I would like to be able to easily allow a plugin to be a client of the blocknode api

### Technical Notes

Currently the BackFill and TssBootstrap plugins have copies of the BlockNodeClient class and the bn_source.proto file.
[ ] Move BlockNodeClient to spi, base, or common and reuse the same code across plugins
[ ] Move bn_sources.proto to protobuf sources

berryware:
## Summary

- Extracts the three duplicate `BlockNodeClient` / `BlockStreamSubscribeUnparsedClient` implementations from `backfill`, `roster-bootstrap-tss`, and `roster-bootstrap-rsa` into a single shared class in `block-node/base` under the new `org.hiero.block.node.base.client` package.
- Moves the peer-node configuration proto (`block_node_source.proto`) from module-local `src/main/proto/` directories into `protobuf-sources/src/main/proto/internal/`, following the existing naming convention and compiling into `org.hiero.block.internal` via `protobuf-pbj`.
- Renames `BackfillSourceConfig` → `BlockNodeSourceConfig` and `BackfillSource` → `BlockNodeSource` throughout the backfill module so all three modules share the same generated types.

## Changes

| Module | Change |
|--------|--------|
| `protobuf-sources` | Added `block_node_source.proto` (`BlockNodeSource`, `BlockNodeSourceConfig`, `GrpcWebClientTuning`) to `internal/` |
| `block-node/base` | Added `base.client` package with shared `BlockNodeClient` and `BlockStreamSubscribeUnparsedClient`; removed `pbj-compiler` plugin; updated `module-info.java` |
| `block-node/backfill` | Removed local `client/` package and proto; updated all `BackfillSourceConfig` → `BlockNodeSourceConfig` references across 5 source and 5 test files; removed `pbj-compiler` plugin |
| `block-node/roster-bootstrap-tss` | Removed local `client/` package and proto; updated `TssDataFetcher` and `RosterBootstrapTssPlugin` imports; removed `pbj-compiler` plugin |

## Motivation

This is a prerequisite for [issue #2682](https://github.com/hiero-ledger/hiero-block-node/issues/2682) (peer BN query in `roster-bootstrap-rsa`). Having a shared client in `base` means `roster-bootstrap-rsa` can adopt it without introducing a fourth duplicate.

Closes #2800  

## Test plan

- [x] `./gradlew :base:test` — 108 tests pass
- [x] `./gradlew :roster-bootstrap-tss:test` — 3 tests pass
- [x] `./gradlew :backfill:test` — 121 tests pass


block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/TssDataFetcher.java
```diff
@@ -77,7 +79,7 @@ protected BlockNodeClient getNodeClient(BlockNodeSourceConfig node) {
             LOGGER.log(DEBUG, "Removed unreachable client for node [{0}], will attempt to recreate", node.address());
         }
         return nodeClientMap.computeIfAbsent(
-                node, n -> new BlockNodeClient(n, enableTls, maxIncomingBufferSize, n.grpcWebclientTuning()));
+                node, n -> new BlockNodeClient(n, 10_000, enableTls, maxIncomingBufferSize, n.grpcWebclientTuning()));
```
Nana-EC: in Backfill this is a plugin config.
Given the minor usage of the client it's probably okay to not require a confgi.
However, I'd move it to a final class field with comments so it's clear we're setting the gRPC timeout to 10s

Nana-EC (dismissed):
LG, 1 suggestion

jsync-swirlds (dismissed):
N/A

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/TssDataFetcher.java
jsync-swirlds: Given how often we've had to fiddle with timeouts, I would recommend making this a value in the Bootstrap plugin configuration.

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
```diff
@@ -19,9 +21,11 @@
 import java.util.Optional;
 import java.util.function.ToIntFunction;
 import org.hiero.block.api.BlockNodeServiceInterface;
+import org.hiero.block.internal.BlockNodeSourceConfig;
+import org.hiero.block.internal.GrpcWebClientTuning;
 import org.hiero.block.node.spi.historicalblocks.BlockAccessor;
 
-public class BlockNodeClient {
+public class BlockNodeClient implements Closeable {
```
jsync-swirlds: If you're going to implement Closeable, it might be worth implementing AutoCloseable; so this can (optionally) be used in try-with-resources.

jsync-swirlds (dismissed):
We need to note somewhere that this is a temporary client.
We should be using a proper Block Node SDK for this purpose rather than a partial client in the `base` module (not least because we're now including reference to Helidon client libraries in every plugin, even where not needed or desired).

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
berryware: Updated

block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/TssDataFetcher.java
berryware: Updated

docs/block-node/configuration.md
```diff
@@ -223,6 +223,7 @@ Currently, no specific options.
 |:----------------------------------------------|:------------------------------------------------------------------------------------------|----------:|
 | ROSTER_BOOTSTRAP_TSS_BLOCK_NODE_SOURCES_PATH  | File path to the JSON file containing a list of block node servers to query for TSS data. |        "" |
 | ROSTER_BOOTSTRAP_TSS_QUERY_PEER_INTERVAL      | The amount of time in milliseconds between queries to the Peer Block Nodes for TSS data.  |     60000 |
+| ROSTER_GRPC_OVERALL_TIMEOUT                   | Overall gRPC timeout (connect, read, poll) in ms.                                         |     10000 |
```
jsync-swirlds: The min is 10000, but the default remains 60000...

docs/block-node/configuration.md
```diff
@@ -126,7 +126,7 @@ Prometheus-format metrics. Its properties are prefixed with
 | BACKFILL_DELAY_BETWEEN_BATCHES        | Delay (ms) between block batches.                               |      1000 |
 | BACKFILL_INITIAL_DELAY                | Initial delay (ms) before starting backfill.                    |     15000 |
 | BACKFILL_PER_BLOCK_PROCESSING_TIMEOUT | Timeout (ms) to wait for a block batch.                         |      1000 |
-| BACKFILL_GRPC_OVERALL_TIMEOUT         | Overall gRPC timeout (connect, read, poll) in ms.               |     60000 |
+| BACKFILL_GRPC_OVERALL_TIMEOUT         | Overall gRPC timeout (connect, read, poll) in ms.               |     10000 |
```
jsync-swirlds: This did not actually change; the min is 10000, but the default is still 60000 (at least as of this PR).

jsync-swirlds (dismissed):
Minor error in the docs (or the config, not sure which is actually intended).

jsync-swirlds (approved):

ivannov (approved):
LGTM

PR #2966, opened 2026-06-05
test(solo-e2e): Wrb solo e2e integration test

AlfredoG87:
## What

Adds a "special" solo-e2e network topology that exercises the Phase-2a path: Consensus Nodes stream **Wrapped Record Blocks (WRB)** carrying gossiped RSA signatures, and Block Nodes verify them against an **RSA roster** loaded by the `roster-bootstrap-rsa` plugin (fetched from the Mirror Node) — instead of TSS-signed blocks.

Two topologies are added: `single-wrb-rsa` (1 CN, 1 BN, 1 MN) and `3cn-2bn-wrb-rsa` (3 CN, 2 BN, 1 MN).

## How it works

Topologies opt in via a top-level `verification_mode: rsa-wrb` marker (default `tss`), registered in the topology JSON schema. When the marker is present, `solo-deploy-network.sh`:

- forces TSS off and emits CN properties `blockStream.streamWrappedRecordBlocks=true` + `tss.*=false`;
- generates a Block Node overlay that enables the `roster-bootstrap-rsa` plugin, pointing it at the in-cluster Mirror Node REST-Java service (`http://mirror-1-restjava.<ns>.svc.cluster.local`, which serves `/api/v1/network/nodes`), and drops `--block-node-tss-overlay`;
- keeps the standard **BN -> CN -> MN** order (Solo only wires Block Nodes as Consensus-Node stream sources if the BN exists at CN-deploy time). The roster plugin polls the Mirror Node indefinitely, so the BN tolerates the MN starting later.

The overlay generator gains `--verification-mode`; in `rsa-wrb` it emits the WRB-cutover Mirror Node importer config (`block.cutover` + `firstStage`, `DISABLE_IMPORTER_SPRING_PROFILES=true`).

## Changes

- **Topologies:** `topologies/single-wrb-rsa.yaml`, `topologies/3cn-2bn-wrb-rsa.yaml`; `verification_mode` enum added to `network-topology.schema.yaml`.
- **Deploy:** `solo-deploy-network.sh` marker handling, conditional CN props, BN WRB overlay, `generate_overlays` refactor (order-independent).
- **Overlay generator:** `--verification-mode` + MN scenario-1 cutover branch.
- **Versions:** `resolve-versions.sh` compatibility floor -> CN 0.75.0-rc.4, MN 0.154.0, BN 0.35.0.
- **Test:** `tests/rsa-roster-verification.yaml` + `rsa-roster-verification` assertion in `solo-test-runner.sh` (roster loaded, RSA successes accumulating, zero roster mismatches; tolerates the startup-window failures before the roster loads).
- **CI/docs:** both topologies added to the scheduler matrix and workflow inputs; README topology/test/assertion/matrix tables updated.
- **Cleanup:** moved `monitor-block-proofs.sh` into `solo-e2e-test/scripts/` (it's used only by the framework) and simplified its two callers.

## Validation

Deployed `single-wrb-rsa` end-to-end and confirmed: roster loaded from the Mirror Node (`blocknode_roster_entries_loaded=1`), WRB blocks verified via RSA (`blocknode_verification_proof_total{proof_type="rsa",result="success"}` climbing, `rsa_roster_mismatch_total=0`), the Mirror Node syncing blocks from the Block Node, and the `rsa-roster-verification` test passing.

## Version note

MN is floored at `0.154.0` (latest GA) because the importer images for `0.155.0-rc2` .. `0.156.0-rc1` (and `main`) have a GraalVM native-image crash (`RecordFileConsensusTimestampsRecalculateMigration$RecordFileSlice` record components missing from the importer's reflection config). `0.154.0` is the newest version that both avoids the crash and supports WRB cutover (since `0.153.0`). Worth filing upstream with the Mirror Node team.

Test Runs:

https://github.com/hiero-ledger/hiero-block-node/actions/runs/27146357023

Latest run with latest MN version and workaround fixes:

https://github.com/hiero-ledger/hiero-block-node/actions/runs/27178245208


jsync-swirlds (dismissed):

berryware (approved):
Looks good.

jsync-swirlds (approved):

PR #2969, opened 2026-06-05
fix(Messaging Facility): non-backpressure item handler blocks disruptor Block Node

### LINKED ISSUE #2932: fix(Messaging Facility): non-backpressure item handler blocks disruptor
There is a bug in the Messaging Facility implementation where we are handling messaging for a non-blocking message handler. The messaging implementation allows the non-blocking handler to block the message thread such that it's gating sequence is never released.

Non-Blocking handlers are allowed to block indefinitely _without_ blocking the disruptor ring buffer. Currently, that does not work correctly, because we are still using gating sequences and unregistering the handlers when we detect _after a message is handled_ that the handler is too far behind.

What we need to do is use a non-gating sequence or similar mechanism so that a non-blocking handler that does not release its handling thread cannot block the ring buffer.

### Technical Notes
In the Messaging Facility, there is a concept of a non-blocking item handler, which is supposed to deliver messages to handlers _without blocking_ the disruptor ring buffer.
Currently, there is a bug in this code [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java#L388-L426)

```java
        final InformedEventHandler<BlockItemBatchRingEvent> informedEventHandler =
// This lambda contains the point where the disruptor ring buffer can be blocked:
                (event, sequence, endOfBatch, percentageBehindRingHead) -> {
                    // send on the event block items
                    handler.handleBlockItemsReceived(event.get()); // The handler blocks the thread here.
                    if (percentageBehindRingHead > 80) { // This is never reached while the handler blocks the thread.
                        // If the event processor is more than 80% behind, we need to stop it.
                        // This is a sign that the event processor is not able to keep up with the
                        // rate of events being published.
                        unregisterBlockItemHandler(handler);
                        // the handler is too far behind, call its too far behind method
                        handler.onTooFarBehindError();
                    }
                };
```
When the handler blocks the thread on line 393, line 394 is dead code. The eviction never fires. The handler is never removed from the gating sequences.

That matters because of this code [in registerHandler](https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java#L653-L683) on [line 672](https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java#L672)
```java
        // Dynamically add sequences to the ring buffer
        ringBuffer.addGatingSequences(batchEventProcessor.getSequence());
```
The handler's sequence is a hard gate on the ring buffer. The producer cannot wrap around until every gating sequence advances. With the handler frozen on put(), its sequence is frozen, and once the ring fills, the publisher thread blocks too.

We need to find a different mechanism (non-gating sequence, perhaps) that allows the non-blocking subscriptions to still be processed, but prevents those subscriptions from blocking the ring buffer.

berryware:
## Reviewer Notes

###What was done

- Added a GatingHandler interface that all Handlers extend. The isGating() method returns true by default. NoBackPressureBlockItemHandler overrides isGating() to return false.

- Added import java.util.concurrent.atomic.AtomicBoolean — needed for the one-shot guard.

- Fixed the registerNoBackpressureBlockItemHandler lambda — two problems corrected:

1. Lag check moved before dispatch. The original code called handleBlockItemsReceived first, then checked lag — dead code when the handler blocks. The check now runs first: if the handler is already >80% behind, it's evicted immediately without ever calling handleBlockItemsReceived.
2. One-shot eviction guard (AtomicBoolean evicted). After halt() is called, the BatchEventProcessor may deliver a few more buffered events from its current batch before stopping. Without the guard, onTooFarBehindError() and unregisterBlockItemHandler() would be called repeatedly. compareAndSet(false, true) ensures exactly one eviction fires.

## Related Issue(s)
Closes: #2932

### DIFF
--- a/block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
+++ b/block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.ServiceBuilder;
 import org.hiero.block.node.spi.blockmessaging.BackfilledBlockNotification;
@@ -28,6 +29,7 @@ import org.hiero.block.node.spi.blockmessaging.BlockItemHandler;
 import org.hiero.block.node.spi.blockmessaging.BlockItems;
 import org.hiero.block.node.spi.blockmessaging.BlockMessagingFacility;
 import org.hiero.block.node.spi.blockmessaging.BlockNotificationHandler;
+import org.hiero.block.node.spi.blockmessaging.GatingHandler;
 import org.hiero.block.node.spi.blockmessaging.NewestBlockKnownToNetworkNotification;
 import org.hiero.block.node.spi.blockmessaging.NoBackPressureBlockItemHandler;
 import org.hiero.block.node.spi.blockmessaging.PersistedNotification;
@@ -387,18 +389,23 @@ public class BlockMessagingFacilityImpl implements BlockMessagingFacility {
     @Override
     public synchronized void registerNoBackpressureBlockItemHandler(
             final NoBackPressureBlockItemHandler handler, final boolean cpuIntensiveHandler, final String handlerName) {
+        // One-shot eviction guard: the event processor may deliver a few more buffered events after halt()
+        // is called, so we use an AtomicBoolean to ensure unregister and onTooFarBehindError fire exactly once.
+        final AtomicBoolean evicted = new AtomicBoolean(false);
         final InformedEventHandler<BlockItemBatchRingEvent> informedEventHandler =
                 (event, sequence, endOfBatch, percentageBehindRingHead) -> {
-                    // send on the event block items
-                    handler.handleBlockItemsReceived(event.get());
-                    if (percentageBehindRingHead > 80) {
-                        // If the event processor is more than 80% behind, we need to stop it.
-                        // This is a sign that the event processor is not able to keep up with the
-                        // rate of events being published.
-                        unregisterBlockItemHandler(handler);
-                        // the handler it got too far behind
-                        handler.onTooFarBehindError();
+                    if (evicted.get()) {
+                        // Already evicted; skip remaining buffered events.
+                        return;
                     }
+                    if (percentageBehindRingHead > 80 && evicted.compareAndSet(false, true)) {
+                        // Check lag BEFORE dispatching: if already too far behind, evict now rather
+                        // than calling handleBlockItemsReceived (which might block indefinitely).
+                        unregisterBlockItemHandler(handler);
+                        handler.onTooFarBehindError();
+                        return;
+                    }
+                    handler.handleBlockItemsReceived(event.get());
                 };
         if (blockItemDisruptor.hasStarted()) {
             registerHandler(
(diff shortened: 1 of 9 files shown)


block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
```diff
@@ -387,18 +391,34 @@ public synchronized void registerBlockItemHandler(
...
+                    if (percentageBehindRingHead > 80 && evicted.compareAndSet(false, true)) {
+                        // Check lag BEFORE dispatching: if already too far behind, evict now rather
+                        // than calling handleBlockItemsReceived (which might block indefinitely).
                         unregisterBlockItemHandler(handler);
-                        // the handler it got too far behind
                         handler.onTooFarBehindError();
+                        return;
                     }
+                    final BlockItems blockItems = event.get();
+                    LOGGER.log(DEBUG, "Received event {0}", blockItems);
```
jsync-swirlds: We should not log on every event in the messaging factility.
Logging has a cost, even if it doesn't write anything, and that cost should not be imposed on a per-message basis.

In the _very rare_ event that we _must_ log on every message, it should be `TRACE`.

I would suggest that, in this case, we do not need or want any logging at all in this Lambda expression.

block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
```diff
@@ -387,18 +391,34 @@ public synchronized void registerBlockItemHandler(
...
+
+                    long totalBlockItems =
+                            blockItemsReceived.addAndGet(blockItems.blockItems().size());
+                    LOGGER.log(
+                            DEBUG,
+                            "Total Block {0}, block items = {1} for handler {2}:",
+                            blockItems.blockNumber(),
+                            totalBlockItems,
+                            handlerName);
```
jsync-swirlds: Same here, this is per-message, and should not be present.

berryware: removed

block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
```diff
@@ -669,7 +675,15 @@ private static <H, E> void registerHandler(
...
         // Dynamically add sequences to the ring buffer
-        ringBuffer.addGatingSequences(batchEventProcessor.getSequence());
+        if (handler.isGating()) {
+            ringBuffer.addGatingSequences(batchEventProcessor.getSequence());
+        } else {
+            // Non-gating (no-backpressure) handlers must start from the current cursor so they
+            // only receive future events. The default starting sequence is -1, which causes the
+            // processor to replay every past event still in the ring buffer — for a backfill
+            // subscriber this fills liveBlockQueue and bypasses the historical read path.
+            batchEventProcessor.getSequence().set(ringBuffer.getCursor());
```
jsync-swirlds, line 685: This is correct.
The Stream Subscriber plugin expects that a non-backpressure handler is expected to deliver from "here forward" when connected, it should not offer any "history". This is also semantically correct to the intent of the non-backpressure handler (it should act like a best-effort delivery).

jsync-swirlds (dismissed):

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/GatingHandler.java
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.spi.blockmessaging;
+
+public interface GatingHandler {
```
ivannov, line 23: Maybe add Javadoc explaining the "gating" terminology and when it should be overriden with `false`?

ivannov (approved):

jsync-swirlds (approved):

PR #2972, opened 2026-06-05
docs(testing-simulator): fix inaccurate and incomplete content (#2965)

### LINKED ISSUE #2965: docs: Fix inaccurate and incomplete content in Block Node simulator testing guide
## Summary
An independent node operator tested the Block Node simulator guide after a Tier 1 install and found multiple inaccurate instructions, missing version guidance, wrong endpoint information, and absent log examples that are critical for users to verify success.

### 1. `localhost` endpoint doesn't work on cloud VMs
The guide states: *"For Local deployment: localhost:40840"*

On a GCP VM, `localhost` did not work. The operator had to use the internal IP (e.g., `10.128.0.4`).

**Fix:** Clarify that on cloud VMs, the internal IP should be used instead of `localhost`. Add a note explaining how to retrieve the internal IP (e.g., `hostname -I` or from GCP console).

### 2. "Create a new file named docker-compose-publisher.yaml" instruction is confusing
The guide suggests using `touch` and then populating the file. The operator noted this is unnecessarily indirect.

**Fix:** Replace with direct file creation instruction (e.g., using `cat > file << 'EOF'` or a text editor).

### 3. No guidance on how to determine the installed Block Node version
The guide says: *"Replace `<BLOCK_NODE_VERSION_TAG>` with the Block Node version you are testing."*

The operator had no way to determine which version was installed. They had to guess based on protobuf version (0.33.0) and releases page.

**Fix:** Add a step showing how to determine the running Block Node version, for example:
```bash
kubectl -n block-node exec block-node-block-node-server-0 -- solo-provisioner version
# or
kubectl get deployment -n block-node -o yaml | grep image
```
Also clarify explicitly: *"The simulator image tag must match your Block Node version."*

### 4. "Package registry" link is unhelpful without explanation
The link to the simulator package registry is provided without context. The operator didn't know whether the simulator version should equal the Block Node version.

**Fix:** Add explicit guidance: *"The simulator image tag must match your Block Node version exactly. For example, if your Block Node is `0.33.0`, use `simulator-image:0.33.0`."*

### 5. Simulator image `0.33.0` is not published / image pull fails
The operator ran into a pull failure:
`Error response from daemon: failed to resolve reference
"ghcr.io/hiero-ledger/hiero-block-node/simulator-image:0.33.0": not found`

They had to guess `0.33.0-rc2` to get a working image.

**Fix:** 
- Verify that simulator images are published for all GA Block Node releases, and fix the publishing gap if not.
- Add a note in the docs pointing operators to check available tags at the registry and warning that RC tags may be needed in some cases.
- Engineering follow-up: confirm whether `simulator-image` tags are published on every GA release.

### 6. External IP is blocked by firewall — Kubernetes service IP guidance is wrong
The guide states: *"Use the Kubernetes service IP from the EXTERNAL-IP column."*

On the operator's VM, the firewall blocked external IPs even from inside the VM. The internal IP was needed.

**Fix:** Add a note that firewall rules may block the external IP even from inside the VM, and provide guidance on using the internal/cluster IP as a fallback.

### 7. "Environment Variable Explanation" section is confusing
The section appears after variables have already been set, making it look like a second set of instructions rather than an explanation.

**Fix:** Move the explanation *before* the commands that use the variables, or reframe it clearly as a reference: *"The variables you set above mean the following..."*

### 8. No example of healthy simulator log output
The guide says: *"You will see logs indicating the blocks are being streamed"* — but doesn't show what healthy output looks like, leaving operators unsure whether their setup is working.

**Fix:** Add the following (or similar) as an example of healthy simulator output:
`simulator-publisher  | 2026-06-04 10:48:18.759+0000 INFO    PublisherClientModeHandler#millisPerBlockStreaming Streaming block took 0ms. Delaying for 999ms to maintain 1,000 milliseconds per block.
simulator-publisher  | 2026-06-04 10:48:18.776+0000 INFO    PublishStreamObserver#onNext Received Response: acknowledgement { block_number: 44 }`

### 9. Step 3 "Verify Block Streaming" points to simulator logs, not Block Node logs
The operator expected to verify success on the Block Node side, not just the simulator side.

**Fix:** Add a Block Node-side verification step using:
```bash
kubectl logs -n block-node block-node-block-node-server-0 | grep ExtendedMerkleTreeSession
```
With expected output example showing `Created ExtendedMerkleTreeSession for block N`.

### 10. Spurious `kubectl` "Defaulted container" message is unexplained
Several `kubectl exec` and `kubectl logs` commands produce:
`Defaulted container "block-node-server" out of: block-node-server, init-storage-dirs (init), copy-core-modules (init), resolve-plugins (init)`

This confused the operator who wasn't sure if something was wrong.

**Fix:** Add a note that this message is expected when a pod has multiple containers and is informational only — the command still executes correctly.

## Reporter
Giuseppe (independent Mirror Node operator) — via Nana Essilfie-Conduah

## References
- Feedback file: `testing-a-deplyed-block-node-using-the-simulator-feedback.md`
- Block Node version tested: 0.33.0 (or 0.33.0-rc2)
- Solo Provisioner v0.18.1

Slack: https://swirldslabs.slack.com/archives/C08HECALMNF/p1780591381656499

deshmukhpranali:
## Reviewer Notes
Address all 10 findings from issue #2965 (operator-tested feedback) 

## Related Issue(s)
Fixes #2965

### DIFF
--- a/docs/block-node/operations/testing-a-deployed-block-node-using-the-simulator.md
+++ b/docs/block-node/operations/testing-a-deployed-block-node-using-the-simulator.md
@@ -13,171 +13,162 @@ This container streams blocks to your deployed Block Node, verifying connectivit
 
 Before you begin, ensure you have:
 
-- A running Block Node deployment on your system using one of the following methods:
-  - [**Manual Single-Node Kubernetes Deployment**](./single-node-k8s-deployment.md)
-  - [**Solo Provisioner Single-Node Kubernetes Deployment**](./solo-weaver-single-node-k8s-deployment.md)
-- **[Docker](https://docs.docker.com/get-started/get-docker/) and Docker Compose** installed and available on your machine where you will run the simulator.
-- The **gRPC** service address and port for your Block Node:
-  - For Local deployment: `localhost:40840`
-  - For cloud deployment server address:
-
-    ```bash
-    kubectl get svc -n block-node
-    ```
-- A gRPC client tool for verification (optional but recommended):
-  - **Postman** (with gRPC support), or
+- A running Block Node deployment, set up via one of:
+  - [Manual Single-Node Kubernetes Deployment](./single-node-k8s-deployment.md)
+  - [Solo Provisioner Single-Node Kubernetes Deployment](./solo-weaver-single-node-k8s-deployment.md)
+- **[Docker](https://docs.docker.com/get-started/get-docker/) and Docker Compose** installed on the machine where the simulator will run.
+- The **gRPC address and port** of your Block Node: the default port is `40840` (see `server.port` in [configuration.md](../configuration.md)). The address depends on where the simulator runs; [Choose the Block Node address](#choose-the-block-node-address) in Step 1 explains how to pick it.
+- The **installed Block Node version**: this becomes the simulator image tag. [Pick a simulator image tag](#pick-a-simulator-image-tag) in Step 1 explains how to retrieve it.
+- A **gRPC client tool** for verification (optional but recommended):
+  - **Postman** with gRPC support, or
   - **grpcurl** command-line tool
 
 ## Step 1: Create the Simulator Docker Compose File
 
-1. Create a new file named **`docker-compose-publisher.yaml`** in the root of your project (or any location where you plan to run the test):
+Create `docker-compose-publisher.yaml` in the directory where you plan to run the test, with the following content:
 
-   ```bash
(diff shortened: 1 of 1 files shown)


deshmukhpranali:
@kfbr, can you please review this?

jsync-swirlds (approved):

AlfredoG87 (approved):
LGTM

PR #2979, opened 2026-06-08
feat: add peer BN query step to RsaRosterBootstrapPlugin.

### LINKED ISSUE #2682: Query peer BN for RSA roster bootstrap data

### LINKED ISSUE #2793: Bootstrap RSA Roster Plugin — Design Improvements (post-merge review feedback)

berryware:
## Reviewer Notes

Implements the three-step bootstrap sequence described in issue #2682:
1. Bootstrap file (pre-loaded by BlockNodeApp) — unchanged
2. Peer BN gRPC query (NEW) — queries configured peer block nodes via serverStatusDetail and extracts the NodeAddressBook
3. Mirror Node REST fallback — existing behaviour, unchanged

Changes:
- AddressBookFetcher: new class modeled on TssDataFetcher; queries peers in order, validates ≥1 non-blank RSA key, pools BlockNodeClient instances, falls through on exception or empty result
- RsaRosterBootstrapConfig: three new fields — blockNodeSourcesPath, peerQueryIntervalMillis (default 5s), peerQueryMaxRetries (default 3), plus enableTLS and maxIncomingBufferSize for gRPC tuning
- RsaRosterBootstrapPlugin: rewrites start() to run peer query before Mirror Node; adds rsa_roster_peer_requests / rsa_roster_peer_errors metrics; closes AddressBookFetcher on stop()
- module-info.java / build.gradle.kts: add gRPC runtime deps
- AddressBookFetcherTest: 16 new unit tests (success, validation, failover, metrics, error eviction)
- RsaRosterBootstrapPluginTest: 3 new peer-query integration tests

## Related Issue(s)
Closes #2682
Closes: #2793

### DIFF
--- a/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
+++ b/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
@@ -1,0 +1,136 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.roster.bootstrap.rsa;
+
+import static java.lang.System.Logger.Level.DEBUG;
+import static java.lang.System.Logger.Level.INFO;
+import static java.lang.System.Logger.Level.WARNING;
+
+import com.hedera.hapi.node.base.NodeAddressBook;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.concurrent.ConcurrentHashMap;
+import org.hiero.block.api.ServerStatusDetailResponse;
+import org.hiero.block.api.ServerStatusRequest;
+import org.hiero.block.internal.BlockNodeSource;
+import org.hiero.block.internal.BlockNodeSourceConfig;
+import org.hiero.block.node.base.client.BlockNodeClient;
+import org.hiero.block.node.roster.bootstrap.rsa.RsaRosterBootstrapPlugin.MetricsHolder;
+
+/// Fetches a `NodeAddressBook` from one or more configured peer block nodes via gRPC.
+///
+/// Each peer is queried in order via `serverStatusDetail`. The first peer that returns a
+/// valid address book (at least one entry with a non-blank RSA public key) is accepted and
+/// its address book is returned. If all peers fail or return an unusable book, `null` is
+/// returned and the caller falls through to the Mirror Node fetch.
+///
+/// Connection pooling: a `BlockNodeClient` is created per peer on first use and cached for
+/// subsequent calls. If a client becomes unreachable it is removed from the pool so it can
+/// be recreated on the next attempt.
+public class AddressBookFetcher implements Closeable {
+
+    private static final System.Logger LOGGER = System.getLogger(AddressBookFetcher.class.getName());
+
+    private final BlockNodeSource blockNodeSource;
+    private final boolean enableTls;
+    private final int maxIncomingBufferSize;
+    private final MetricsHolder metrics;
+
+    /// Package-private for testing — allows tests to inject mock clients.
+    final ConcurrentHashMap<BlockNodeSourceConfig, BlockNodeClient> nodeClientMap = new ConcurrentHashMap<>();
+
+    /// Constructs a fetcher for the given peer list and config.
+    ///
+    /// @param blockNodeSource the list of peer block nodes to query
+    /// @param config the RSA bootstrap configuration (TLS and buffer-size settings)
+    /// @param metrics holder for peer-query metrics counters
+    public AddressBookFetcher(
+            @NonNull BlockNodeSource blockNodeSource,
+            @NonNull RsaRosterBootstrapConfig config,
+            @NonNull MetricsHolder metrics) {
+        this.blockNodeSource = blockNodeSource;
+        this.enableTls = config.enableTLS();
+        this.maxIncomingBufferSize = config.maxIncomingBufferSize();
+        this.metrics = metrics;
+        for (BlockNodeSourceConfig node : blockNodeSource.nodes()) {
+            LOGGER.log(INFO, "Loaded peer block node: {0}", node);
+        }
+    }
+
+    /// Queries each configured peer in order and returns the first valid `NodeAddressBook`.
+    ///
+    /// A `NodeAddressBook` is considered valid if it contains at least one `NodeAddress` with
+    /// a non-blank RSA public key. If no peer returns a usable book, `null` is returned.
+    ///
+    /// @return a valid `NodeAddressBook`, or `null` if all peers fail or return empty books
+    public NodeAddressBook getNodeAddressBook() {
+        for (BlockNodeSourceConfig node : blockNodeSource.nodes()) {
+            BlockNodeClient client = getNodeClient(node);
+            if (client == null || !client.isNodeReachable()) {
+                LOGGER.log(DEBUG, "Peer [{0}] is not reachable, skipping", node.address());
+                continue;
+            }
+
+            try {
+                final ServerStatusDetailResponse response =
+                        client.getBlockNodeServiceClient().serverStatusDetail(new ServerStatusRequest());
+                final NodeAddressBook book = response.nodeAddressBook();
+                metrics.peerRequests().increment();
+
+                if (isValid(book)) {
+                    LOGGER.log(
+                            INFO,
+                            "Received valid NodeAddressBook with {0} entries from peer [{1}]",
+                            book.nodeAddress().size(),
+                            node.address());
+                    return book;
+                }
+                LOGGER.log(
+                        DEBUG, "Peer [{0}] returned an empty or key-less NodeAddressBook, trying next", node.address());
+            } catch (RuntimeException e) {
+                LOGGER.log(
+                        WARNING,
+                        "Failed to retrieve NodeAddressBook from peer [{0}]: {1}",
+                        node.address(),
+                        e.getMessage());
+                metrics.peerErrors().increment();
+                // Remove so a fresh client is created on the next attempt
+                nodeClientMap.remove(node);
+            }
+        }
+        return null;
+    }
+
+    /// Returns a cached or newly-created `BlockNodeClient` for the given peer.
+    /// If the existing client is unreachable it is evicted and a fresh one is created.
+    ///
+    /// @param node the peer node configuration
+    /// @return a `BlockNodeClient` for the node
+    protected BlockNodeClient getNodeClient(BlockNodeSourceConfig node) {
+        BlockNodeClient existing = nodeClientMap.get(node);
+        if (existing != null && !existing.isNodeReachable()) {
+            try {
+                nodeClientMap.remove(node).close();
+            } catch (IOException e) {
+                LOGGER.log(WARNING, "Unable to close BlockNodeClient [{0}]: {1}", node.name(), e);
+            }
+            LOGGER.log(DEBUG, "Removed unreachable client for peer [{0}], will recreate", node.address());
+        }
+        return nodeClientMap.computeIfAbsent(
+                node, n -> new BlockNodeClient(n, 10_000, enableTls, maxIncomingBufferSize, n.grpcWebclientTuning()));
+    }
+
+    /// Returns `true` if the book has at least one entry with a non-blank RSA public key.
+    static boolean isValid(NodeAddressBook book) {
+        if (book == null || book.nodeAddress().isEmpty()) return false;
+        return book.nodeAddress().stream()
+                .anyMatch(addr -> addr.rsaPubKey() != null && !addr.rsaPubKey().isBlank());
+    }
+
+    @Override
+    public void close() throws IOException {
+        for (BlockNodeClient client : nodeClientMap.values()) {
+            client.close();
+        }
+    }
+}
--- a/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
+++ b/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
@@ -112,17 +112,21 @@ public class AddressBookFetcher implements AutoCloseable {
     ///
     /// @param node the peer node configuration
     /// @return a `BlockNodeClient` for the node
-    protected BlockNodeClient getNodeClient(BlockNodeSourceConfig node) {
-        BlockNodeClient existing = nodeClientMap.get(node);
-        if (existing != null && !existing.isNodeReachable()) {
-            try {
-                nodeClientMap.remove(node).close();
-            } catch (IOException e) {
-                LOGGER.log(WARNING, "Unable to close BlockNodeClient [{0}]: {1}", node.name(), e);
+    BlockNodeClient getNodeClient(BlockNodeSourceConfig node) {
+        return nodeClientMap.compute(node, (key, current) -> {
+            if (current != null) {
+                if (current.isNodeReachable()) {
+                    return current;
+                }
+                try {
+                    current.close();
+                } catch (IOException e) {
+                    LOGGER.log(WARNING, "Unable to close BlockNodeClient [{0}]: {1}", key.name(), e);
+                }
             }
-            LOGGER.log(DEBUG, "Removed unreachable client for peer [{0}], will recreate", node.address());
-        }
-        return nodeClientMap.computeIfAbsent(node, this::fromBlockNodeSourceConfig);
+            LOGGER.log(DEBUG, "Client unreachable for peer [{0}], will recreate", key.address());
+            return fromBlockNodeSourceConfig(key);
+        });
     }
 
     /// Returns `true` if the book has at least one entry with a non-blank RSA public key.
--- a/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
+++ b/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
@@ -5,6 +5,7 @@ import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.WARNING;
 
+import com.hedera.hapi.node.base.NodeAddress;
 import com.hedera.hapi.node.base.NodeAddressBook;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
@@ -91,7 +92,7 @@ public class AddressBookFetcher implements AutoCloseable {
                         DEBUG, "Peer [{0}] returned an empty or key-less NodeAddressBook, trying next", node.address());
             } catch (RuntimeException e) {
                 LOGGER.log(
-                        WARNING,
+                        INFO,
                         "Failed to retrieve NodeAddressBook from peer [{0}]: {1}",
                         node.address(),
                         e.getMessage());
@@ -118,23 +119,36 @@ public class AddressBookFetcher implements AutoCloseable {
             }
             LOGGER.log(DEBUG, "Removed unreachable client for peer [{0}], will recreate", node.address());
         }
-        return nodeClientMap.computeIfAbsent(
-                node,
-                n -> new BlockNodeClient(
-                        n, grpcOverallTimeout, enableTls, maxIncomingBufferSize, n.grpcWebclientTuning()));
+        return nodeClientMap.computeIfAbsent(node, this::fromBlockNodeSourceConfig);
     }
 
     /// Returns `true` if the book has at least one entry with a non-blank RSA public key.
     static boolean isValid(NodeAddressBook book) {
         if (book == null || book.nodeAddress().isEmpty()) return false;
-        return book.nodeAddress().stream()
-                .anyMatch(addr -> addr.rsaPubKey() != null && !addr.rsaPubKey().isBlank());
+        for (NodeAddress nodeAddress : book.nodeAddress()) {
+            if (nodeAddress.rsaPubKey() != null && !nodeAddress.rsaPubKey().isBlank()) {
+                return true;
+            }
+        }
+        return false;
     }
 
     @Override
-    public void close() throws IOException {
+    public void close() {
         for (BlockNodeClient client : nodeClientMap.values()) {
-            client.close();
+            try {
+                client.close();
+            } catch (IOException e) {
+                LOGGER.log(
+                        WARNING,
+                        "Unable to close BlockNodeClient [{0}]: {1}",
+                        client.getBlockNodeServiceClient().fullName(),
+                        e);
+            }
         }
+    }
+
+    private BlockNodeClient fromBlockNodeSourceConfig(BlockNodeSourceConfig n) {
+        return new BlockNodeClient(n, grpcOverallTimeout, enableTls, maxIncomingBufferSize, n.grpcWebclientTuning());
     }
 }
--- a/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
+++ b/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
@@ -33,10 +33,12 @@ public class AddressBookFetcher implements AutoCloseable {
 
     private final BlockNodeSource blockNodeSource;
     /// Global timeout in milliseconds for gRPC calls to block nodes (used as fallback).
-    private final int grpcOverallTimeout;
+    private final int globalGrpcTimeoutMs;
     private final boolean enableTls;
     private final int maxIncomingBufferSize;
     private final MetricsHolder metrics;
+    /// RSA data is small (key material, not full blocks) — 512 KB is a safe ceiling.
+    private static final int RSA_DATA_MAX_PROTOBUF_MESSAGE_SIZE_BYTES = 512 * 1024;
 
     /// Package-private for testing — allows tests to inject mock clients.
     final ConcurrentHashMap<BlockNodeSourceConfig, BlockNodeClient> nodeClientMap = new ConcurrentHashMap<>();
@@ -51,7 +53,7 @@ public class AddressBookFetcher implements AutoCloseable {
             @NonNull RsaRosterBootstrapConfig config,
             @NonNull MetricsHolder metrics) {
         this.blockNodeSource = blockNodeSource;
-        this.grpcOverallTimeout = config.grpcOverallTimeout();
+        this.globalGrpcTimeoutMs = config.grpcOverallTimeout();
         this.enableTls = config.enableTLS();
         this.maxIncomingBufferSize = config.maxIncomingBufferSize();
         this.metrics = metrics;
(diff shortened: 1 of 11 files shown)


block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+                            "Received valid NodeAddressBook with {0} entries from peer [{1}]",
+                            book.nodeAddress().size(),
+                            node.address());
+                    return book;
+                }
+                LOGGER.log(
+                        DEBUG, "Peer [{0}] returned an empty or key-less NodeAddressBook, trying next", node.address());
+            } catch (RuntimeException e) {
+                LOGGER.log(
+                        WARNING,
```
ata-nas: Not sure if this should be `WARNING` or `INFO`

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+                nodeClientMap.remove(node).close();
+            } catch (IOException e) {
+                LOGGER.log(WARNING, "Unable to close BlockNodeClient [{0}]: {1}", node.name(), e);
+            }
+            LOGGER.log(DEBUG, "Removed unreachable client for peer [{0}], will recreate", node.address());
+        }
+        return nodeClientMap.computeIfAbsent(
+                node,
+                n -> new BlockNodeClient(
+                        n, grpcOverallTimeout, enableTls, maxIncomingBufferSize, n.grpcWebclientTuning()));
```
ata-nas: Nit, we can replace with static method ref.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /// Returns `true` if the book has at least one entry with a non-blank RSA public key.
+    static boolean isValid(NodeAddressBook book) {
+        if (book == null || book.nodeAddress().isEmpty()) return false;
+        return book.nodeAddress().stream()
+                .anyMatch(addr -> addr.rsaPubKey() != null && !addr.rsaPubKey().isBlank());
```
ata-nas: Nit, we can replace with a simple loop which is more readable and performant.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public void close() throws IOException {
+        for (BlockNodeClient client : nodeClientMap.values()) {
+            client.close();
+        }
```
ata-nas, line 155: Let's catch each and every exception on close. If we see an exception, we should still continue to close all clients. Logging on exception.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapConfig.java
```diff
@@ -10,24 +10,41 @@
...
+/// @param grpcOverallTimeout the timeout for grpc connections.
+/// @param maxIncomingBufferSize maximum incoming gRPC message buffer size in bytes
 @ConfigData("roster.bootstrap.rsa")
 public record RsaRosterBootstrapConfig(
         // spotless:off
         @Loggable @ConfigProperty(defaultValue = "") String mirrorNodeBaseUrl,
-        @Loggable @ConfigProperty(defaultValue = "5000") @Min(100) int initialQueryIntervalMillis,
-        @Loggable @ConfigProperty(defaultValue = "60000") @Min(10000) int subsequentQueryIntervalMillis,
+        @Loggable @ConfigProperty(defaultValue = "5000") @Min(100) int mnInitialQueryIntervalMillis,
+        @Loggable @ConfigProperty(defaultValue = "60000") @Min(10000) int mnSubsequentQueryIntervalMillis,
```
ata-nas: Nit for readability: we can have the "_" separator for numbers;
For instance `5000` -> `5_000`;

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -95,12 +120,37 @@ public void init(final BlockNodeContext context, final ServiceBuilder serviceBui
...
+                    LOGGER.log(
+                            WARNING,
+                            "Failed to parse block node sources from [{0}], peer query disabled: {1}",
+                            config.blockNodeSourcesPath(),
+                            e.getMessage());
+                }
+            } else {
+                LOGGER.log(
+                        WARNING,
+                        "blockNodeSourcesPath [{0}] is not a regular file, peer query disabled",
```
ata-nas: `isRegularFile` also fails (returns false) when the file does not exist. We can add that to the log. Either not existing or not a regular file.

ata-nas (commented):
Some optional nits.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
berryware: Updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
berryware: Done

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
berryware: Done

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
berryware: replaced with non-static private function

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
berryware: Changed to INFO.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapConfig.java
berryware: Done

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: Updated log message

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
```diff
@@ -0,0 +1,154 @@
...
+
+                if (isValid(book)) {
+                    LOGGER.log(
+                            INFO,
+                            "Received valid NodeAddressBook with {0} entries from peer [{1}]",
+                            book.nodeAddress().size(),
+                            node.address());
+                    return book;
+                }
+                LOGGER.log(
```
ivannov, line 94: I wonder whether we might need a metric for invalid address books?

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -109,85 +159,110 @@ public void onContextUpdate(final BlockNodeContext updatedContext) {
...
 
-    /// {@inheritDoc}
-    @Override
-    public void stop() {
-        shutdownExecutor();
+    private void fetchFromPeer() {
+        final NodeAddressBook book = addressBookFetcher.getNodeAddressBook();
+        if (book != null) {
+            applicationStateFacility.updateAddressBook(book);
+            recordSuccessMetrics(book, System.currentTimeMillis(), "Peer BN");
```
ivannov: Maybe the `System.currentTimeMillis` should be in the beginning of the method, right before we call the address book fetcher.

That's what we do in fetching from mirror node

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
```diff
@@ -361,4 +530,93 @@ private static NodeAddressBook buildAddressBook(final int count) {
...
+                throw new IllegalStateException("backfillSourcePath is required");
+            }
+
+            return new HashMap<>(Map.of(
+                    "roster.bootstrap.rsa.blockNodeSourcesPath", blockNodeSourcesPath,
+                    "roster.bootstrap.rsa.bnInitialQueryIntervalMillis", String.valueOf(bnInitialQueryIntervalMillis),
+                    "roster.bootstrap.rsa.bnSubsequentQueryIntervalMillis",
+                            String.valueOf(bnSubsequentQueryIntervalMillis),
+                    "roster.bootstrap.tsa.maxIncomingBufferSize", String.valueOf(maxIncomingBufferSize),
+                    "roster.bootstrap.tsa.grpcOverallTimeout", String.valueOf(grpcOverallTimeout),
```
ivannov: Should these rather be `.rsa.` and not `.tsa.`?

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
```diff
@@ -339,6 +373,141 @@ void http500TriggersRetryThenSucceeds() throws InterruptedException {
...
+    class PeerQueryPath {
+
+        @TempDir
+        Path tempDir;
+
+        private Map<String, String> peerConfig(String sourcesPath, int maxRetries) {
+            return Map.of(
+                    "roster.bootstrap.rsa.blockNodeSourcesPath",
+                    sourcesPath,
+                    "roster.bootstrap.rsa.peerQueryMaxRetries",
```
ivannov: I didn't see this property in the config. I don't see it used productively either (just in the JavaDoc). Was it missed?

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
```diff
@@ -339,6 +373,141 @@ void http500TriggersRetryThenSucceeds() throws InterruptedException {
...
+            Path file = tempDir.resolve("bn-sources.json");
+            java.nio.file.Files.writeString(file, json);
+            return file.toString();
+        }
+
+        @Test
+        @DisplayName("Peer success: Mirror Node is never called")
+        void peerSuccessMirrorNodeNeverCalled() throws IOException {
+            // Set up a test server that returns a valid address book via serverStatusDetail
+            org.hiero.block.node.app.fixtures.server.TestBlockNodeServer server =
```
ivannov: Nit: can you import the package instead of using the FQN?

docs/block-node/configuration.md
```diff
@@ -225,14 +225,20 @@ Currently, no specific options.
...
+| ROSTER_BOOTSTRAP_RSA_MIRROR_NODE_BASE_URL                | The URL of the mirror node from which to request the node address book.                      |        "" |
+| ROSTER_BOOTSTRAP_RSA_MN_INITIAL_QUERY_INTERVAL_MILLIS    | The initial period between queries to the mirror node until a node address book is found.    |      5000 |
+| ROSTER_BOOTSTRAP_RSA_MN_SUBSEQUENT_QUERY_INTERVAL_MILLIS | The subsequent period between queries to the mirror node after a node address book is found. |     60000 |
+| ROSTER_BOOTSTRAP_RSA_MIRROR_NODE_CONNECT_TIMEOUT_SECONDS | TCP connect timeout when calling the Mirror Node.                                            |         5 |
+| ROSTER_BOOTSTRAP_RSA_MIRROR_NODE_READ_TIMEOUT_SECONDS    | Per-request read timeout when calling the Mirror Node.                                       |        10 |
+| ROSTER_BOOTSTRAP_RSA_MIRROR_NODE_PAGE_SIZE               | Number of nodes requested per paginated Mirror Node call.                                    |       100 |
+| ROSTER_BOOTSTRAP_RSA_BLOCK_NODE_SOURCES_PATH             | File path to the JSON file containing a list of block node servers to query for TSS data.    |        "" |
```
ivannov: I see a CopyPasteException appearing here ;)

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: Done

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
berryware: Updated

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
berryware: Removed

block-node/roster-bootstrap-rsa/src/test/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPluginTest.java
berryware: Updated

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapConfig.java
```diff
@@ -10,24 +10,41 @@
...
+        @Loggable @ConfigProperty(defaultValue = "5_000") @Min(100) int mnInitialQueryIntervalMillis,
+        @Loggable @ConfigProperty(defaultValue = "60_000") @Min(10000) int mnSubsequentQueryIntervalMillis,
         @Loggable @ConfigProperty(defaultValue = "5") int mirrorNodeConnectTimeoutSeconds,
         @Loggable @ConfigProperty(defaultValue = "10") int mirrorNodeReadTimeoutSeconds,
-        @Loggable @ConfigProperty(defaultValue = "100") int mirrorNodePageSize) {}
+        @Loggable @ConfigProperty(defaultValue = "100") int mirrorNodePageSize,
+        @Loggable @ConfigProperty(defaultValue = "") String blockNodeSourcesPath,
+        @Loggable @ConfigProperty(defaultValue = "5_000") @Min(100) int bnInitialQueryIntervalMillis,
+        @Loggable @ConfigProperty(defaultValue = "60_000") @Min(10000) int bnSubsequentQueryIntervalMillis,
+        @Loggable @ConfigProperty(defaultValue = "false") boolean enableTLS,
```
jsync-swirlds, line 48: TLS should probably be a feature of individual block node sources; rather than global feature of the client.

jsync-swirlds (dismissed):

docs/block-node/configuration.md
berryware: Fixed

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/AddressBookFetcher.java
berryware: Added

ivannov (approved):

Nana-EC (approved):

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapConfig.java
berryware: captured https://github.com/hiero-ledger/hiero-block-node/issues/3009

Issue #2973, opened 2026-06-05
Bug: `cloud-storage-archive` plugin rejects `ARCHIVE` storage class


AlfredoG87:
### Description

**Symptom**

Node fails to start with:

```
Caused by: java.lang.IllegalArgumentException: No enum constant
org.hiero.block.node.cloud.storage.archive.CloudStorageArchiveConfig.S3StorageClass.ARCHIVE
```

**Root cause**

`CloudStorageArchiveConfig.S3StorageClass` is missing `ARCHIVE` as a valid enum constant. The old `s3-archive` plugin used a plain `String` for `storageClass`, so it accepted any value including GCS-native classes like `ARCHIVE`. The new plugin uses a typed enum that only covers AWS S3 storage classes.

**File to fix**

`block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchiveConfig.java:27`

Add `ARCHIVE` to the `S3StorageClass` enum:

```java
enum S3StorageClass {
    STANDARD,
    STANDARD_IA,
    INTELLIGENT_TIERING,
    ONEZONE_IA,
    EXPRESS_ONEZONE,
    GLACIER,
    GLACIER_IR,
    DEEP_ARCHIVE,
    REDUCED_REDUNDANCY,
    ARCHIVE   // GCS-native storage class (S3-compatible API)
}
```

### Steps to Reproduce

Try to deploy with:

CLOUD_STORAGE_ARCHIVE_STORAGE_CLASS: "ARCHIVE"

### Additional context

**Workaround**

Set `CLOUD_STORAGE_ARCHIVE_STORAGE_CLASS: STANDARD` and configure a GCS bucket lifecycle policy to transition objects to the ARCHIVE storage class automatically.

jsync-swirlds:
We should add GLACIER for AWS as well.

ivannov:
> We should add GLACIER for AWS as well.

It's there

PR #2981, opened 2026-06-08
feat: Add `objectKeyPrefix` Config Property to Cloud Storage Archive Plugin

### LINKED ISSUE #2980: Add `objectKeyPrefix` Config Property to Cloud Storage Archive Plugin
### Story Form

I want to support a configurable object key prefix
So that operators can organize archived blocks under a custom path within the S3 bucket

### Technical Notes

- Add `objectKeyPrefix` property to `CloudStorageArchiveConfig` with an empty string default
- In `BlockUploadTask`, the key is currently built as `ArchiveKey.format(firstBlock, config.groupingLevel())`; when a prefix is set it should become `{prefix}/{ArchiveKey.format(...)}`
- `StartupRecoveryTask` starts its listing at the bucket root via `findLastCommonPrefix`/`findLastObject`; when a prefix is configured, `{prefix}/` must be used as the initial listing path
- `ArchiveKey.parse()` concatenates all path segments to reconstruct the block number -- a prefixed key such as `myprefix/0000/0000/0001/23.tar` would corrupt parsing; the prefix must be stripped from the key before calling `parse()`

ivannov:
## Reviewer Notes

* Adds an `objectKeyPrefix` property (default: "") to `CloudStorageArchiveConfig`, matching the equivalent property already present in `ExpandedCloudStorageConfig`. When set, the prefix is prepended to every S3 object key: `{prefix}/AAAA/BBBB/CCCC/DDDD/EEE.tar`.
* `format` and `parse` now each take an explicit `objectKeyPrefix` parameter. format prepends the prefix (plus /) when non-empty; parse strips it before decoding the block number
*  All three `ArchiveKey.parse()` call sites pass `config.objectKeyPrefix()`. The right-hand S3 directory traversal in `findLastKey` already works correctly with a prefix (it starts from "" and the prefix is just part of the path structure), so only the parse calls needed updating.
* S3StorageClass was missing ARCHIVE, causing node startup to fail when the property was set to that value (used by GCS S3-compatible deployments). The enum constant is added.

## Related Issue(s)
Closes: #2980, #2973


block-node/cloud-storage-archive/src/test/java/org/hiero/block/node/cloud/storage/archive/ArchiveKeyTest.java
```diff
@@ -14,61 +16,72 @@ class ArchiveKeyTest {
...
+                Arguments.of(12340000L, 4, "", "0000/0000/0001/234.tar"),
+                Arguments.of(154000000L, 4, "", "0000/0000/0015/400.tar"),
+                Arguments.of(100000L, 5, "", "0000/0000/0000/1.tar"),
+                Arguments.of(3400000L, 5, "", "0000/0000/0000/34.tar"),
+                Arguments.of(1234500000L, 5, "", "0000/0000/0123/45.tar"),
+                Arguments.of(1000000L, 6, "", "0000/0000/0000/1.tar"),
+                Arguments.of(56000000L, 6, "", "0000/0000/0005/6.tar"),
+                Arguments.of(1234000000L, 6, "", "0000/0000/0123/4.tar"),
+                Arguments.of(0L, 1, "myblocks", "myblocks/0000/0000/0000/0000/0.tar"),
+                Arguments.of(1230L, 1, "hiero/mainnet", "hiero/mainnet/0000/0000/0000/0001/23.tar"));
```
jsync-swirlds, line 39: Perhaps add an example that looks like a resettable network:

```
                Arguments.of(1230L, 1, "hiero/testnet/2024-02", "hiero/testnet/2024-02/0000/0000/0000/0001/23.tar")
```

jsync-swirlds (approved):

Nana-EC (approved):

block-node/cloud-storage-archive/src/test/java/org/hiero/block/node/cloud/storage/archive/ArchiveKeyTest.java
ivannov: Will add this case in the next PR, thanks!

PR #2982, opened 2026-06-09
feat: Improve and extend handling of persisted notifications

### LINKED ISSUE #2198: Improve and Extend Handling of Persisted Notifications
### Story Form

As a Publisher Manager
I want to have robust and improved handling of the Persisted Notifications that I handle
So that I will enable improved usage messages sent to publishers

### Technical Notes

- We currently handle the Persisted Notifications
- We plan to add improvements to the way we handle a failed persistence

jsync-swirlds:
## Reviewer Notes
* Improved handling of failed persistence notifications.
* Cleared most maps on failed verification
* Added a flag and wait method to prevent certain tasks while a reset is pending.
   * Could have used a Lock or Condition, but those are vastly more expensive
     when not resetting, which is expected to outnumber reset by a ratio
     over one million to one.
* Add flag to ensure `checkMidBlockAndShutdown` only executes `blockIsEnding`
  once, regardless of the number of calls.
* Ensured `getActionForBlock` returns error response if called by a handler
  that is no longer in the `handlers` map.
* Removed an unnecessary condition in `endAllHandlersAndClearTracking`.
* Fixed some tests that were not structured correctly
  (e.g. using a disconnected handler to send more data)

## Related Issue(s)
 Resolves #2198

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -5,6 +5,7 @@ import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.TRACE;
 import static java.lang.System.Logger.Level.WARNING;
+import static java.util.concurrent.locks.LockSupport.parkNanos;
 import static org.hiero.block.api.PublishStreamResponse.EndOfStream.Code.TIMEOUT;
 import static org.hiero.block.node.spi.BlockNodePlugin.UNKNOWN_BLOCK_NUMBER;
 import static org.hiero.block.node.stream.publisher.StreamPublisherPlugin.METRIC_PUBLISHER_BLOCKS_CLOSED_COMPLETE;
@@ -43,6 +44,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Condition;
@@ -78,6 +80,10 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     /// Value assigned as a Handler message budget when that handler is being paused
     /// due to overload or other ill-advised behaviors.
     private static final long MESSAGE_BUDGET_PENALTY_FLAG_VALUE = -1L;
+    /// Wait 10 microseconds per pause if a reset is in progress. A reset
+    /// should generally take less than one microsecond, but waiting
+    /// less than 10 microseconds is not completely reliable.
+    private static final long RESET_WAIT_TIME_NANOS = 10_000L;
     private final System.Logger LOGGER = System.getLogger(LiveStreamPublisherManager.class.getName());
     private final MetricsHolder metrics;
     private final BlockNodeContext serverContext;
@@ -90,6 +96,9 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     private final AtomicLong blocksClosedComplete;
     private final Condition dataReadyLatch;
     private final ReentrantLock dataReadyLock;
+    /// Flag indicating that the plugin is in the middle of a reset and must
+    /// not accept new connections, create new Handlers, or forward data.
+    private final AtomicBoolean resetIsActive;
     private final long earliestManagedBlock;
     private final int maxBlocksBeforeStalled;
     private final long staleResendPruneBuffer;
@@ -133,6 +142,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     /// todo(1420) add documentation
     public LiveStreamPublisherManager(
             @NonNull final BlockNodeContext context, @NonNull final MetricsHolder metricsHolder) {
+        resetIsActive = new AtomicBoolean(true);
         serverContext = Objects.requireNonNull(context);
         metrics = Objects.requireNonNull(metricsHolder);
         threadManager = serverContext.threadPoolManager();
@@ -161,6 +171,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         handlerFlowState = new ConcurrentSkipListMap<>();
         initializeBlockNumbers(serverContext);
         scheduleFlowControlRefresh();
+        resetIsActive.compareAndSet(true, false);
     }
 
     /// todo(1420) add documentation
@@ -180,6 +191,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             @NonNull final Pipeline<? super PublishStreamResponse> replies,
             @NonNull final PublisherHandler.MetricsHolder handlerMetrics,
             final String correlationId) {
+        awaitReset();
         final long handlerId = nextHandlerId.getAndIncrement();
         final PublisherHandler newHandler =
                 new PublisherHandler(handlerId, replies, handlerMetrics, this, correlationId);
@@ -195,6 +207,20 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         sendPublisherStatusUpdate(UpdateType.PUBLISHER_CONNECTED, handlers);
         LOGGER.log(DEBUG, "Added new handler {0}", handlerId);
         return newHandler;
+    }
+
+    /// Method to wait for a boolean value to be false.
+    /// We do _not_ want a lock here, a lock ensures a single entry
+    /// at a time, while here we just do not want to proceed when this value is
+    /// true. It's not quite a condition wait, and not quite a lock situation,
+    /// it's more like a gate condition.
+    private void awaitReset() {
+        // Busy wait loop, but will almost always return immediately
+        // when it does pause, it will almost always return after a single pause
+        // because reset takes a few hundred nanoseconds.
+        while (resetIsActive.get()) {
+            parkNanos(RESET_WAIT_TIME_NANOS);
+        }
     }
 
     @Override
@@ -215,16 +241,21 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     @Override
     public BlockAction getActionForBlock(
             final long blockNumber, final BlockAction previousAction, final long handlerId) {
-        return switch (previousAction) {
-            case null -> getActionForHeader(blockNumber);
-            case ACCEPT -> getActionForCurrentlyStreaming(blockNumber);
-            case END_ERROR, END_DUPLICATE ->
-                // This should not happen because the Handler should have shut down.
-                BlockAction.END_ERROR;
-            case SKIP, RESEND, SEND_BEHIND ->
-                // This should not happen because the Handler should have reset the previous action.
-                BlockAction.END_ERROR;
-        };
+        awaitReset();
+        if (handlers.containsKey(handlerId)) {
+            return switch (previousAction) {
+                case null -> getActionForHeader(blockNumber);
+                case ACCEPT -> getActionForCurrentlyStreaming(blockNumber);
+                case END_ERROR, END_DUPLICATE ->
+                    // This should not happen because the Handler should have shut down.
+                    BlockAction.END_ERROR;
+                case SKIP, RESEND, SEND_BEHIND ->
+                    // This should not happen because the Handler should have reset the previous action.
+                    BlockAction.END_ERROR;
+            };
+        } else {
+            return BlockAction.END_ERROR;
+        }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -455,19 +456,28 @@ public void handlePersisted(@NonNull final PersistedNotification notification) {
...
                 }
             } else {
-                if (blockNumber <= lastPersistedBlockNumber.get()) {
-                    // @todo(2198): We may have an extremely rare race condition here
-                    nextUnstreamedBlockNumber.set(blockNumber);
+                if (notification.blockSource() == BlockSource.PUBLISHER) {
+                    if (blockNumber <= lastPersistedBlockNumber.get()) {
+                        nextUnstreamedBlockNumber.set(blockNumber);
+                    }
+                    Collection<PublisherHandler> handlersToClose = handlers.values();
```
ata-nas: Saving the `handlersToClose` should probably happen before setting the `nextUnstreamedBlockNumber` to a new value. In that way, we ensure that we are going to close everyone who is active prior to changing what we expect to receive next.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -455,19 +456,28 @@ public void handlePersisted(@NonNull final PersistedNotification notification) {
                     ensureNextGreaterThanPersisted(blockNumber);
                 }
             } else {
-                if (blockNumber <= lastPersistedBlockNumber.get()) {
-                    // @todo(2198): We may have an extremely rare race condition here
-                    nextUnstreamedBlockNumber.set(blockNumber);
+                if (notification.blockSource() == BlockSource.PUBLISHER) {
+                    if (blockNumber <= lastPersistedBlockNumber.get()) {
```
ata-nas: If `blockNumber > lastPersistedBlockNumber` this means that this is a block that we have probably just received and it failed to persist. We still need that block, however, to be supplied again. In the current scenario, we will not make any changes neither to the next unstreamed, nor scheduling that block for resend.

We can add tests to cover this case so we do not miss it.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -455,19 +456,28 @@ public void handlePersisted(@NonNull final PersistedNotification notification) {
                     ensureNextGreaterThanPersisted(blockNumber);
                 }
             } else {
-                if (blockNumber <= lastPersistedBlockNumber.get()) {
-                    // @todo(2198): We may have an extremely rare race condition here
-                    nextUnstreamedBlockNumber.set(blockNumber);
+                if (notification.blockSource() == BlockSource.PUBLISHER) {
```
ata-nas, line 490: I think the `notification.blockSource() == BlockSource.PUBLISHER` is correct. If backfill is providing a very old block and that fails, backfill should take an action.

We can add tests to cover cases where we receive notifications from other sources.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -455,19 +456,28 @@ public void handlePersisted(@NonNull final PersistedNotification notification) {
...
+                    if (blockNumber <= lastPersistedBlockNumber.get()) {
+                        nextUnstreamedBlockNumber.set(blockNumber);
+                    }
+                    Collection<PublisherHandler> handlersToClose = handlers.values();
+                    // Notify the handlers that persistence failed.
+                    handlersToClose.parallelStream()
+                            .unordered()
+                            .forEach((handler) -> handler.endStreamWithCode(Code.PERSISTENCE_FAILED, false));
+                    queueByBlockMap.clear();
```
ata-nas: Not sure about calling `clear()` on the queue by block map. As per the note below, it is ok for new connections to happen while we are executing this logic. We will then possibly drop their block as not needed and they will get a SKIP if the block is in multiple batches.

Moreover, the `endStreamWithCode` calls to all handlers will make them shutdown on next pipeline call. In those cases, blocks that are still active will be scheduled for resend if they are still in the map when the handler shuts down mid block. So we could be left in a state where we have rolled back to say block 5, but block 6 and 7 are scheduled for resend.

Another thing to consider is to stop the message forwarder task. This will stop data flowing downstream and it will be restarted when the next block arrives in full (I believe) as it seems to be started on the `closeBlock` method. This should improve partial data to be sent downstream which might be made unavailable at map clear.

Finally, we are safe for handling other notifications as we only receive and handle one event regardless of type. This is quite important as other notifications will not interfere with this. There is a caveat, however. If say we receive block 5 and it goes to verification and just passed verification. Then we send block 6 and 7 to verification. Block 5 fails persistence, so we roll back to block 5, but immediately after block 6 persists and we receive the notification for it (this can happen before verification seeing the failed persistence - new plugin - but in any case verification will now expect the next block to verify to be block 6 as we have just verified successfully block 5). This will set the last persisted block to 6, and if block 5 which is the actual next expected block starts being received, we will have a skip for it in the `getActionForHeader` and the `ensureNextGreaterThanPersisted` will roll forward. With new verification changes, we have already verified block 5 and 6 so we expect block 7 next. Verification for any block lower than 7 will be an immediate success, so verification will be able to continue. Effectively we will be unaware that block 5 is needed. I believe we should be able to reproduce this case in a test.

Maybe I am missing something.

ata-nas (commented):
A few questions.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -455,19 +456,28 @@ public void handlePersisted(@NonNull final PersistedNotification notification) {
...
-                    nextUnstreamedBlockNumber.set(blockNumber);
+                if (notification.blockSource() == BlockSource.PUBLISHER) {
+                    if (blockNumber <= lastPersistedBlockNumber.get()) {
+                        nextUnstreamedBlockNumber.set(blockNumber);
+                    }
+                    Collection<PublisherHandler> handlersToClose = handlers.values();
+                    // Notify the handlers that persistence failed.
+                    handlersToClose.parallelStream()
+                            .unordered()
+                            .forEach((handler) -> handler.endStreamWithCode(Code.PERSISTENCE_FAILED, false));
```
AlfredoG87: Do we need a test for this in order to prevent future regressions?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: We need to set the correct value first so that the "end of stream" messages sent back to the publishers have the correct last known good block.

Note, we aren't concerned with the "next block" from any of the current handlers; we're asking all of them to close and that starts with locking their incoming message threads (because that's what PBJ plugin requires for flow control).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: No, we do not schedule for resend, because we are disconnecting every publisher.
There is nothing being received, and no residual blocks.

What we'll do is set the next block we expect to be the block that failed (if not already lower) and when a publisher establishes a new connection, they'll either get a behind result (if they try to send a later block) or send the block that failed to persist.

Basically, we're resetting the entire plugin to start from right after the last block that succeeded, or the one that failed to persist (if lower).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: That's what this line does.

I'm not sure what the comment here is saying.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: This line is the same code that was here before (just moved a bit).
Existing tests already verify that we close all handlers on failed persistence.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: 1. We need to remove all existing data, because we do not want to attempt any pending blocks (see earlier "reset" comment above).
   - Note, it's OK to drop data that's partial as that handler will also be disconnected. The only edge is a publisher that connects after the current are disconnected (which is partly why we set the last persisted before disconnecting handlers).
   - The queue clear does need to happen _after_ the handlers are signaled to disconnect, however.
      - Worst case, there is a tiny window (almost impossible, but not quite) where a publisher might start streaming the failed block, then have it's queue cleared, and receive an error (because it's mid-block and the queue is gone unexpectedly, I believe that turns into an error, but will check).
1. Blocks scheduled for resend is a good point, I might need to clear that map as well.
1. Stopping the forwarder task shouldn't be a concern because there is no data to forward.
1. There will almost always be some amount of partial data in flight and already forwarded.  That is not unexpected and is tolerable (verification will fail that block, and subscribers will have to detect and discard).
1. Regarding notifications, we will _not_ acknowledge block 6 or set the last persisted, because we defer acknowledgement and last persisted update when it comes out of order; we would wait for block 5 to be persisted, then the next block after that (presumably 7 or 8) would acknowledge all the way forward.
   - The guards against out-of-order acknowledgement ensure, in the case of a failure followed by success, that we have behavior _similar_ to resend in that we refuse to move forward until the missing block is received.

AlfredoG87 (dismissed):

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I added more maps to the items cleared out in this case.
I also added a flag and flag-wait method (busy-wait alternative to a Condition because the double lock on bypass is way more expensive than a volatile boolean check in this case) to prevent the few concerning edge cases where things might happen in the middle of clearing the maps and closing all of the handlers.

Please take a look and let me know if I missed any other edges.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: Just pointing out. No concern here.
The comment is basically to add a test that covers the case that only publisher source notifications affect this.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: My point is, if last persisted is 4, block 5 was received, passed verification and is now being persisted. Block 6 is being streamed at the moment and next unstreamed is 7. Then, block 5 fails persistence. We do the check `blockNumber <= lastPersistedBlockNumber.get()` and that will not do anything about changing next unstreamed, so it remains 7. Then, we execute all the rest of the logic, drop block 6 in flight, clear containers etc. etc. and we are left with a publisher manager which expects block 7 next. What happens with blocks 5 and 6? Why do we want to roll back only if block that failed is <= last persisted?

I believe the check here should be either against the current value of next unstreamed or no check at all. This is publisher source data, we know it is recent. But we must not roll forward, only backwards.

P.S. just to note that what I also just stumbled upon is that `ensureNextGreaterThanPersisted` will be called on the next publisher! This means that whatever value we set here will 100% be ignored on next pipeline call. I think we will never reach the case where we see a failed persistence from publisher and that failure is for a block <= last persisted. If we do see that, it most likely is for a block being resent, or a possible greedy backfill happening just before.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: I see what you mean.

The logic that sends the "end of stream" message never ever uses the next unstreamed value. In fact, the next unstreamed value is not used to send a message back ever. What we use for that is the last persisted value.

Now, on another hand, yes, we still need that value, because a publisher that is ending, in case it is also ending mid-block, the next unstreamed value is important to end correctly, mainly to determine if to schedule for resend. Which, if happens is fine, because we will clear those values lower down the method call here.

So, the atomic long set here is, I think, most likely in the correct place, as you stated.

My main concern here is that an active publisher can be in the get action for header, where it will possibly CAS the value. We should not miss this update here or allow it to be overwritten while we are resetting.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: This is much improved.

First, to give back an answer:
1. Yes, I am aware that we need to drop all data and reset. And yes, it is fine to forward some partial data. Finally, no error should be seen, simply we keep sending some data and all of a sudden we stop cause we have no data any longer.
1. Yes, we do have to clear that as well. I still think there is a little more to it, but will expand below.
1. It is ok not to stop the message forwarder. I was mainly exploring that scenario.
1. Yes, partial in flight data being there is totally tolerable and expected.
1. This is a recent addition, but I did not see that the rolling forward of the last persisted is also behind that guard. This makes the point irrelevant, in that case. Thanks for pointing out.

Now:
- There is another caveat to what we've done. I like the addition of the resetting barrier. It lacks coverage of a publisher already past that point, but we should be able to account for possible changes the publisher will make while we are resetting. And that will be only once possibly per publisher. The issue that I find here, is that if a publisher has been made to wait for the barrier, this definitely means that it must be shutting down. Only exception is if it waits behind the barrier on publisher add (new publisher just registering). Once the barrier lifts, we allow that publisher to still go through the logic, possibly polluting the internal state. And since we have just cleared critical flags (the containers like queueByBlockMap which is critical when a handler is shutting down), we are possibly going to not ask for a resend after the fact.
  - We cannot really shutdown immediately also, as that causes other issues.

Not saying that the above is unacceptable, just want to explore the cases.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I see what you're saying.
I'll remove the if test.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Will add in follow-up.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I see where this is going.  I think there are a couple items:
1. We need a check in the getActionForBlock method for if the handler is still in the map after the awaitReset call.
1. We need to add a barrier in the checkMidBlockAndShutdown method of the handler to ensure it doesn't do anything if it has already been called for that handler.

ata-nas: Note, when sending end stream and closing from the handle persisted method, we must make the shutdown immediate. This is so that we will properly call the block is ending method.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Removed.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Done.

AlfredoG87 (approved):
LGTM

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -214,16 +241,21 @@ public void removeHandler(final long handlerId) {
...
-            case ACCEPT -> getActionForCurrentlyStreaming(blockNumber);
-            case END_ERROR, END_DUPLICATE ->
-                // This should not happen because the Handler should have shut down.
-                BlockAction.END_ERROR;
-            case SKIP, RESEND, SEND_BEHIND ->
-                // This should not happen because the Handler should have reset the previous action.
-                BlockAction.END_ERROR;
-        };
+        awaitReset();
+        if (handlers.containsKey(handlerId)) {
```
ata-nas, line 245: It may be worthwhile to leave a comment to explain why we do the check if handler is still in the map.
Also, it may be worthwhile to leave a comment as to what happens if we start resetting and a handler is already beyond that barrier.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -1112,11 +1171,8 @@ public Long call() {
             // End this thread after some number of batches, so we don't have
             // a thread that becomes overly "old" and experiences "senescence".
             while (!forwardingLimitReached) {
-                // @todo(2347) we can improve the loop because we have new way of doing resends, we can also
-                //   improve on the way we determine the current block number below, but also the way we decide
-                //   which items and how to hold them for a bit, to await for the end of block message so we can
-                //   send them to messaging as the end of the block.
                 final long currentBlockNumber = determineCurrentBlockNumber();
+                publisherManager.awaitReset(); // Hold here if the manager is resetting
```
ata-nas, line 1175: Nit: We can hold on L:1174, just before determining the current block number. No reason to hold after determining the current block number. If we hold before that, we give a chance for the next correct block number to be determined. 

This is completely optional, however. After a reset, if we find nothing for the current block number, we will simply do nothing.

ata-nas (approved):
Looks good!

PR #2983, opened 2026-06-09
feat: Expose per-service ports for the API plugins

Nana-EC:
## Reviewer Notes

[PR #2880](https://github.com/hiero-ledger/hiero-block-node/pull/2880) made the API service port
server-configurable by adding a nullable per-service `port` to `ServiceBuilder` (and the multi-socket
wiring in `BlockNodeApp`), but left every plugin passing `null` — so all services still bind the single
default `server.port`. This PR finishes that work: each API plugin now reads an optional `port` from its
own config and registers its service on that dedicated port, letting node operators put individual
services on their own ports when needed.

Five plugins gain a port:

| Plugin | Config key | Service |
|--------|-----------|---------|
| Block Access | `block.access.port` | gRPC |
| Health | `health.port` | HTTP (`/healthz/*`) |
| Server Status | `server.status.port` | gRPC |
| Publisher | `producer.port` | gRPC |
| Subscriber | `subscriber.port` | gRPC |

Each port is a **nullable `Integer`**. When unset (the default) the plugin passes `null` and the service
shares `server.port` — i.e. **default behavior is unchanged**. Set a port to move just that service onto
its own listener.

## What changed

- **Configs**
  - New `HealthConfig`, `ServerStatusConfig`, `BlockAccessConfig` records (those plugins had no config
    class), each registered via the plugin's `configDataTypes()`.
  - Added a `port` field to the existing `PublisherConfig` (`producer.*`) and `SubscriberConfig`
    (`subscriber.*`).
  - The port is declared `@ConfigProperty(defaultValue = ConfigProperty.NULL_DEFAULT_VALUE) Integer port`
    so it resolves to `null` when unset.
- **Plugin wiring** — each plugin reads `context.configuration().getConfigData(<X>Config.class).port()`
  in `init()` and passes it to `registerGrpcService(...)` / `registerHttpService(...)`. A `null` port
  resolves to `server.port` in `ServiceBuilderImpl`.
- **`module-info`** — the three new-config modules export their config package to the config framework
  (`com.swirlds.config.impl`, `com.swirlds.config.extensions`) and `org.hiero.block.node.app`, and
  declare `requires transitive com.swirlds.config.api` (the config records expose the config annotations
  as API). Matches the existing `stream-publisher` module setup.
- **`ConfigLogger` null-tolerance (required)** — `ConfigLogger.collectConfig` called
  `accessor.invoke(record).toString()` unconditionally, so any config property that resolves to `null`
  threw an NPE during `BlockNodeApp` construction. These nullable ports are the first null-valued config
  in the codebase, so this would break **every** app boot. Guarded the null case (logs `"null"`).
- **Docs** — `docs/block-node/configuration.md` documents the new `*_PORT` keys.

## Design notes

- **Nullable, not a sentinel.** `null` cleanly means "share `server.port`". No magic default value.
- **No `@Min`/`@Max` on the port.** swirlds' `@Min`/`@Max` validators call `requireNonNull` on the value,
  so a nullable port with those annotations would NPE at startup when unset. The valid range
  (`1024`–`65535`) is documented in each config's Javadoc and enforced by the web server when it binds.
- **Backward compatible / non-intrusive.** With no port set, the registration path is identical to before
  (`null` → `server.port`); nothing about the default single-port deployment changes.

## Testing

- New E2E test `BlockNodeAPITests.servicesBindToConfiguredUniquePortsOnly`: boots `BlockNodeApp` with a
  unique port per plugin (`40891`–`40895`) and asserts, for **all five** services, that each is reachable
  on its own port (publish→ACK, server-status, get-block, subscribe-streams-block-0, health `200`) **and**
  is **not** reachable on the default `server.port` (gRPC unary calls throw / streams are rejected; health
  returns `404`).
- Updated `HealthServiceTest` to stub `context.configuration()` now that the plugin reads its port in
  `init()`.
- `./gradlew :suites:runAPISuites` is green (14 passing); unit tests for all changed modules pass;
  `qualityCheck` (incl. `checkModuleDirectivesScope` and spotless) passes on every changed module.

> Note: run `:suites:runAPISuites` on its own — running the real-port E2E suite in parallel with other
> port-binding module tests can cause port contention.

## Related Issue(s)
Resolves #2880

### DIFF
--- a/block-node/app/src/main/java/org/hiero/block/node/app/logging/ConfigLogger.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/logging/ConfigLogger.java
@@ -70,8 +70,11 @@ public final class ConfigLogger {
                         final String fieldName = component.getName();
                         final Record configRecord = configuration.getConfigData(configType);
                         try {
-                            final String value =
-                                    component.getAccessor().invoke(configRecord).toString();
+                            final Object rawValue = component.getAccessor().invoke(configRecord);
+                            // A null value means an optional property (e.g. a nullable Integer declared with
+                            // ConfigProperty.NULL_DEFAULT_VALUE, such as a per-service port) was not set. Log it
+                            // as "null" rather than throwing an NPE on toString().
+                            final String value = rawValue == null ? "null" : rawValue.toString();
                             // If the field is not annotated as '@Loggable' then it's a sensitive value and needs to
                             // be handled differently.
                             if (component.getAnnotation(Loggable.class) == null) {
--- a/block-node/block-access-service/src/main/java/org/hiero/block/node/access/service/BlockAccessConfig.java
+++ b/block-node/block-access-service/src/main/java/org/hiero/block/node/access/service/BlockAccessConfig.java
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.access.service;
+
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import org.hiero.block.node.base.Loggable;
+
+/**
+ * Configuration for the block-access service plugin.
+ *
+ * @param port the dedicated port the block-access gRPC service binds to. When {@code null} (the
+ *     default) the plugin shares the default {@code server.port}. When set it must be a valid
+ *     port in the range {@code 1024}–{@code 65535}. No {@code @Min}/{@code @Max} is declared
+ *     because those validators reject a {@code null} value (they would fail when the property is
+ *     unset); the range is enforced by the web server when binding.
+ */
+@ConfigData("block.access")
+public record BlockAccessConfig(
+        @Loggable @ConfigProperty(defaultValue = ConfigProperty.NULL_DEFAULT_VALUE)
+        Integer port) {}
--- a/block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
+++ b/block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.health;
+
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import org.hiero.block.node.base.Loggable;
+
+/**
+ * Configuration for the health service plugin.
+ *
+ * @param port the dedicated port the health HTTP endpoints bind to. When {@code null} (the
+ *     default) the plugin shares the default {@code server.port}. When set it must be a valid
+ *     port in the range {@code 1024}–{@code 65535}. No {@code @Min}/{@code @Max} is declared
+ *     because those validators reject a {@code null} value (they would fail when the property is
+ *     unset); the range is enforced by the web server when binding.
+ */
+@ConfigData("health")
+public record HealthConfig(
+        @Loggable @ConfigProperty(defaultValue = ConfigProperty.NULL_DEFAULT_VALUE)
+        Integer port) {}
--- a/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusConfig.java
+++ b/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusConfig.java
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.node.server.status;
+
+import com.swirlds.config.api.ConfigData;
+import com.swirlds.config.api.ConfigProperty;
+import org.hiero.block.node.base.Loggable;
+
+/**
+ * Configuration for the server-status service plugin.
+ *
+ * @param port the dedicated port the server-status gRPC service binds to. When {@code null} (the
+ *     default) the plugin shares the default {@code server.port}. When set it must be a valid
+ *     port in the range {@code 1024}–{@code 65535}. No {@code @Min}/{@code @Max} is declared
+ *     because those validators reject a {@code null} value (they would fail when the property is
+ *     unset); the range is enforced by the web server when binding.
+ */
+@ConfigData("server.status")
+public record ServerStatusConfig(
+        @Loggable @ConfigProperty(defaultValue = ConfigProperty.NULL_DEFAULT_VALUE)
+        Integer port) {}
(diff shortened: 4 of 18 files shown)


block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @param port the dedicated port the health HTTP endpoints bind to. When {@code null} (the
+ *     default) the plugin shares the default {@code server.port}. When set it must be a valid
+ *     port in the range {@code 1024}–{@code 65535}. No {@code @Min}/{@code @Max} is declared
+ *     because those validators reject a {@code null} value (they would fail when the property is
+ *     unset); the range is enforced by the web server when binding.
+ */
+@ConfigData("health")
+public record HealthConfig(
+        @Loggable @ConfigProperty(defaultValue = ConfigProperty.NULL_DEFAULT_VALUE)
```
jsync-swirlds, line 19: Why struggle with nullable and null values?
`0` is not a valid port, so we could just use `0` as default and interpret that as "do not set a separate port".

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
```diff
@@ -52,6 +56,7 @@ public record PublisherConfig(
         @Loggable @ConfigProperty(defaultValue = "15") @Min(10) int consecutiveFullBudgetIntervalsForPenalty,
         @Loggable @ConfigProperty(defaultValue = "30") @Min(2) @Max(600) int penaltyDurationSeconds,
         @Loggable @ConfigProperty(defaultValue = "3600") @Min(600) @Max(86400) long penaltyResetIntervalSeconds,
-        @Loggable @ConfigProperty(defaultValue = "5") @Min(1) @Max(10) int duplicateBlockSkipWindow) {
+        @Loggable @ConfigProperty(defaultValue = "5") @Min(1) @Max(10) int duplicateBlockSkipWindow,
+        @Loggable @ConfigProperty(defaultValue = ConfigProperty.NULL_DEFAULT_VALUE) Integer port) {
```
ivannov, line 60: Quick question (feel free to resolve). I thought that we want to provide the override port config to all non-publisher plugins. Because the publisher will use the app config's port?

ivannov (approved):
Looks good. Just one question

berryware (approved):
Looks good

block-node/app/src/main/java/org/hiero/block/node/app/logging/ConfigLogger.java
```diff
@@ -70,8 +70,11 @@ static Map<String, String> collectConfig(@NonNull final Configuration configurat
                         final String fieldName = component.getName();
                         final Record configRecord = configuration.getConfigData(configType);
                         try {
-                            final String value =
-                                    component.getAccessor().invoke(configRecord).toString();
+                            final Object rawValue = component.getAccessor().invoke(configRecord);
+                            // A null value means an optional property (e.g. a nullable Integer declared with
+                            // ConfigProperty.NULL_DEFAULT_VALUE, such as a per-service port) was not set. Log it
+                            // as "null" rather than throwing an NPE on toString().
+                            final String value = rawValue == null ? "null" : rawValue.toString();
```
AlfredoG87, line 77: should we replicate this behaviour in the sensitive case too?

block-node/block-access-service/src/main/java/org/hiero/block/node/access/service/BlockAccessConfig.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Configuration for the block-access service plugin.
+ *
+ * @param port the dedicated port the block-access gRPC service binds to. When {@code null} (the
+ *     default) the plugin shares the default {@code server.port}. When set it must be a valid
+ *     port in the range {@code 1024}–{@code 65535}. No {@code @Min}/{@code @Max} is declared
+ *     because those validators reject a {@code null} value (they would fail when the property is
+ *     unset); the range is enforced by the web server when binding.
+ */
```
AlfredoG87, line 16: I believe we were using `///` markdown for javadoc on new stuff

block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Configuration for the health service plugin.
+ *
+ * @param port the dedicated port the health HTTP endpoints bind to. When {@code null} (the
+ *     default) the plugin shares the default {@code server.port}. When set it must be a valid
+ *     port in the range {@code 1024}–{@code 65535}. No {@code @Min}/{@code @Max} is declared
+ *     because those validators reject a {@code null} value (they would fail when the property is
+ *     unset); the range is enforced by the web server when binding.
+ */
```
AlfredoG87, line 16: same, javadoc could be markdown

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusConfig.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Configuration for the server-status service plugin.
+ *
+ * @param port the dedicated port the server-status gRPC service binds to. When {@code null} (the
+ *     default) the plugin shares the default {@code server.port}. When set it must be a valid
+ *     port in the range {@code 1024}–{@code 65535}. No {@code @Min}/{@code @Max} is declared
+ *     because those validators reject a {@code null} value (they would fail when the property is
+ *     unset); the range is enforced by the web server when binding.
+ */
```
AlfredoG87, line 16: same, md javadoc

AlfredoG87 (approved):
LGTM, just a couple nits that can be handled in the future

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherConfig.java
jsync-swirlds: Publisher should set its own; I think the application default port is most appropriate for the Health plugin, rather than the Publisher.

Really, we should allow any plugin to use a custom port, and then the app default is either unused or not even set (if that's possible).

We might even need (as a future enhancement) to add a further option to set not just a port, but also request a new endpoint, because (as examples) the Health plugin may need to run on localhost only, or the publisher plugin might need to be a different interface than the subscriber plugin.

rockysingh (approved):
lgtm

block-node/app/src/main/java/org/hiero/block/node/app/logging/ConfigLogger.java
Nana-EC: I'll walk these changes back in a follow up by using port 0

block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
Nana-EC: Yes, good catch. Will add to follow up

block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
Nana-EC: Good idea. I'll walk these changes back in a follow up by using port 0

PR #2988, opened 2026-06-09
fix(ci): upgrade Node.js to 22.x for k6 tests to satisfy Solo 0.72.0

Nana-EC:
## Summary

Solo 0.72.0 declares engines.node >= 22.0.0 but BN k6 logic is using version 20x
Running it under Node 20 emits `EBADENGINE` warnings and can produce silent runtime failures, which may have been contributing to mirror-node pods not becoming ready in the k6 CI job.

Updating setup-node to "22.x" resolves the engine mismatch.


berryware (approved):
Looks good.

AlfredoG87 (approved):
LGTM

PR #2996, opened 2026-06-11
test(e2e): WRB differential test - validate Mirror Node database comparison

### LINKED ISSUE #2767: Epic: E2E differential test - validate MN database from WRB files matches live MN
## Overview

Validate that a Mirror Node (MN) database built by ingesting Wrapped Record Block (WRB) files from a Block Node produces identical results to a Mirror Node ingesting live blocks.

## Architecture Discovery

Research shows:
- Mirror Node **only** ingests from Block Nodes (not Consensus Nodes directly)
- Block Nodes can serve both live blocks and wrapped blocks transparently
- Solo topologies support multiple Block Nodes → multiple Mirror Nodes
- MN config: `sourceType: BLOCK_NODE` (no changes needed for wrapped blocks)

## Proposed Test Architecture in Solo

```
Consensus Node → Block Node 1 (live blocks) → Mirror Node 1 (baseline DB)
                  Block Node 2 (wrapped blocks) → Mirror Node 2 (test DB)
```

Compare MN1 DB vs MN2 DB to validate WRB integrity.

## Success Criteria

- [ ] Both Mirror Nodes ingest same block range (0 to N)
- [ ] Database schemas match
- [ ] All critical tables have identical data:
  - `transaction`
  - `account_balance`
  - `token`
  - `nft`
  - `contract_state`
- [ ] Any differences are documented and explained
- [ ] If discrepancies found, root cause identified in WRB wrapping logic

## Benefits of Solo-Based Approach

- **Controlled environment** - no cloud infrastructure needed
- **Repeatable** - can run test multiple times
- **Fast iteration** - local testing with small block ranges
- **Existing framework** - leverage Solo E2E test infrastructure

### LINKED ISSUE #2769: Configure Block Node to serve wrapped blocks in Solo
## Overview

Configure the second Block Node in the WRB differential test topology to serve wrapped blocks instead of live blocks from Consensus Node.

## Background

Block Nodes support historical backfill mode where they can serve pre-wrapped blocks from storage instead of receiving live blocks from CN. This is used for jumpstart and archival scenarios.

## Acceptance Criteria

- [ ] Understand Block Node historical backfill configuration
- [ ] Configure BN2 to:
  - Not connect to Consensus Node
  - Serve blocks from local/mounted storage
  - Accept wrapped block files in expected format
- [ ] Document configuration in test README
- [ ] Verify BN2 starts successfully and exposes streaming API

## Investigation Needed

1. How does Block Node serve historical blocks?
2. What configuration flags/env vars control this mode?
3. Where should wrapped blocks be mounted in the BN pod?
4. Does BN need special startup flags?

**Files to check**:
- `charts/block-node-server/values.yaml`
- `charts/block-node-server/values-overrides/solo-dev.yaml`
- Block Node app configuration
- `docs/block-node/` documentation

## Expected Configuration

```yaml
# bn-block-node-wrapped-values.yaml
blockNode:
  mode: historical  # or similar flag
  storage:
    wrappedBlocks:
      path: /data/wrapped-blocks
      mountFrom: <volume with wrapped blocks>
  consensus:
    enabled: false  # Don't connect to CN
```

## Testing

- [ ] BN2 starts without errors
- [ ] BN2 exposes gRPC streaming API on port 40840
- [ ] BN2 can serve block 0 (test with grpcurl or client)
- [ ] MN2 can successfully connect to BN2

### LINKED ISSUE #2771: Implement database comparison framework for MN differential test
## Overview

Create tooling and scripts to compare PostgreSQL databases from two Mirror Node instances to identify any discrepancies between live block ingestion and wrapped block ingestion.

## Acceptance Criteria

- [ ] Script to export both MN databases
- [ ] Comparison logic for critical tables
- [ ] Report generation showing:
  - Tables compared
  - Row counts (expected to match)
  - Data differences (if any)
  - Categorization of differences (expected vs unexpected)
- [ ] Documentation of expected differences
- [ ] Pass/fail determination logic

## Critical Tables to Compare

Based on Mirror Node schema:
- `transaction` - all transactions
- `account_balance` - account balances
- `token` - token definitions
- `nft` - NFT ownership
- `contract` - smart contracts
- `contract_state` - contract state storage
- `topic_message` - HCS messages
- `file_data` - file service data

## Expected Differences (Document These)

Some differences may be acceptable:
- **Timestamps**: Ingestion timestamps may differ
- **Internal IDs**: Auto-increment IDs may differ
- **Ordering**: As long as consensus_timestamp ordering is preserved

### Option 1: Full Database Dump Diff
```bash
kubectl port-forward svc/postgres-mn1 5432:5432 &
pg_dump -h localhost -U mirror_node mirror_node > mn1.sql

kubectl port-forward svc/postgres-mn2 5433:5432 &
pg_dump -h localhost -U mirror_node mirror_node > mn2.sql

diff mn1.sql mn2.sql
```

### Option 2: SQL-Based Comparison
```sql
-- Compare row counts
SELECT 'transaction', COUNT(*) FROM transaction
UNION ALL
SELECT 'account_balance', COUNT(*) FROM account_balance;

-- Compare specific records
SELECT * FROM transaction ORDER BY consensus_timestamp;
```

### Option 3: Automated Tool
Build a Python/Java tool that:
1. Connects to both databases
2. Iterates through tables
3. Compares row-by-row (ignoring expected differences)
4. Generates HTML/JSON report

## Script Location

```
tools-and-tests/scripts/solo-e2e-test/scripts/compare-mn-databases.sh
```

## Output Example

```
=== Mirror Node Database Comparison ===
Baseline: MN1 (live blocks)
Test: MN2 (wrapped blocks)
Block Range: 0 - 1500

Table: transaction
  Rows (MN1): 4,523
  Rows (MN2): 4,523
  Status: MATCH

Table: account_balance
  Rows (MN1): 1,250
  Rows (MN2): 1,250
  Data Differences: 3 records
    - Row 100: balance field differs (12345 vs 12346)
    - Row 250: timestamp differs (expected - ingestion time)
  Status: REVIEW NEEDED

=== Summary ===
Total Tables: 15
Exact Matches: 12
Expected Differences: 2
Unexpected Differences: 1
Status: INVESTIGATE DISCREPANCIES
```

### LINKED ISSUE #2772: Execute WRB differential test and document results
## Overview

Execute the complete WRB E2E differential test in Solo and document findings.

## Prerequisites

- [ ] #2768 - Topology created
- [ ] #2769 - BN2 configured for wrapped blocks
- [ ] #2770 - Blocks extracted and wrapped
- [ ] #2771 - Comparison framework implemented

### 1. Start Solo with WRB Differential Topology

```bash
cd tools-and-tests/scripts/solo-e2e-test
task up TOPOLOGY=wrb-differential-test
```

### 2. Monitor Block Production

```bash
# Wait for blocks to be produced
task test:run TEST_FILE=tests/network-status.yaml

# Verify both MNs are ingesting
kubectl logs -f mirror-baseline-0 -n solo-network
kubectl logs -f mirror-wrb-test-0 -n solo-network
```

### 3. Wait for Ingestion Completion

- Let both MNs ingest to same block number
- Confirm both are caught up (no backlog)
- Target: at least 1000 blocks for meaningful test

### 4. Run Database Comparison

```bash
./scripts/compare-mn-databases.sh \
  --baseline mirror-baseline \
  --test mirror-wrb-test \
  --namespace solo-network \
  --output comparison-report.html
```

### 5. Analyze Results

**If Exact Match**:
- Document success
- WRB files validated for MN ingestion
- Update documentation with findings

**If Discrepancies Found**:
- Document each difference
- Categorize as expected vs unexpected
- For unexpected differences:
  - Identify affected block/transaction
  - Compare original record file vs wrapped block
  - Trace through WRB CLI wrapping logic
  - File bugs for any issues found
  - Fix and re-test

## Deliverables

- [ ] Test execution logs
- [ ] Database comparison report
- [ ] Summary document:
  - Test configuration
  - Block range tested
  - Results (match/discrepancy)
  - Root cause analysis (if discrepancies)
  - Recommendations
- [ ] Update to WRB CLI documentation if needed

## Success Criteria

- [ ] Test completed without infrastructure errors
- [ ] Both MNs ingested same block range
- [ ] Comparison completed for all critical tables
- [ ] All discrepancies are either:
  - Documented as expected, OR
  - Root caused and fixed
- [ ] Final re-test shows exact match (or documented expected differences)

## Documentation Location

```
docs/block-node/testing/wrb-differential-test-results.md
```

Should include:
- Test date and version
- Configuration used
- Block range tested
- Results summary
- Any issues found and resolutions

rockysingh:
## Summary

Closes #2767
Closes #2769
Closes #2771
Closes #2772

### Problem

The `wrb-sequential-comparison.sh` script was deploying Mirror Nodes with `--enable-ingress`, causing HAProxy ingress controller conflicts:
```
ServiceAccount "mirror-ingress-controller-solo-network" exists and cannot be imported 
into release "haproxy-ingress-2": current value is "haproxy-ingress-1"
```

### Solution

Removed `--enable-ingress` flag from `deploy_mn_to_bn()` function since:
- Sequential script uses port-forwarding for API access (`localhost:5551`)
- Ingress is not required for the test workflow
- Avoids resource conflicts when redeploying Mirror Nodes

### Test Results

**wrb-differential-test.yaml PASSES**
- Events: 11/11 passed
- Assertions: 4/4 passed

 **Sequential Comparison VALIDATED**
- Phase 1: MN1 ingested 11 live blocks from BN1
- Phase 2: MN2 ingested 11 wrapped blocks from BN2
- Phase 3: All API responses identical
  - `network_nodes.json` ✓ Match
  - `blocks.json` ✓ Match
  - `transactions.json` ✓ Match
  - `balances.json` ✓ Match

**Report:** `/tmp/mn-comparison-report.json`

### Changes

- `wrb-sequential-comparison.sh`: Remove `--enable-ingress` from Mirror Node deployment
- Add trailing newlines to multiple files (code style)

### Testing

Run the test:
```bash
cd tools-and-tests/scripts/solo-e2e-test
task test:run TEST_FILE=tests/wrb-differential-test.yaml TOPOLOGY=wrb-differential-test
```

This was tested on solo run here:

https://github.com/hiero-ledger/hiero-block-node/actions/runs/27524417667/job/81348700240

### DIFF
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
@@ -34,6 +34,10 @@ MN_WRB_CAPTURE_DIR="/tmp/mn-wrb-responses"
 BLOCK_RANGE="0:100"
 MIN_TRANSACTIONS=10
 
+# Mirror Node version used for both MN1 (Solo deployment) and the standalone MN2.
+# Override with MIRROR_NODE_VERSION env var. Use a GA version, not 'main' or 'latest'.
+MIRROR_NODE_VERSION="${MIRROR_NODE_VERSION:-v0.156.0}"
+
 # Work directories
 WORK_DIR="/tmp/wrb-test-$$"
 RECORD_FILES_DIR="${WORK_DIR}/record-files"
@@ -164,10 +168,10 @@ importer:
             responseTimeout: 10s
 EOF
 
-        # Deploy Mirror Node with v0.156.0+
+        # Deploy Mirror Node (version from MIRROR_NODE_VERSION env var)
         solo mirror node add \
             --deployment "${DEPLOYMENT}" \
-            --mirror-node-version "v0.156.0" \
+            --mirror-node-version "${MIRROR_NODE_VERSION}" \
             --pinger \
             --cluster-ref "${CONTEXT}" \
             -f "${overlay_file}" || {
@@ -721,8 +725,8 @@ function compare_phase_responses {
     log "Comparing Phase 1 vs Phase 2 responses..."
 
     if [ ! -d "${MN_CAPTURE_DIR}" ] || [ ! -d "${MN_WRB_CAPTURE_DIR}" ]; then
-        log "WARNING: Missing response directories, skipping comparison"
-        return 0
+        log "ERROR: Missing response directories, cannot run differential comparison"
+        return 1
     fi
 
     python3 "${PYTHON_DIR}/compare_captured_responses.py" \
@@ -730,9 +734,9 @@ function compare_phase_responses {
         "${MN_WRB_CAPTURE_DIR}" \
         --output /tmp/comparison-report.json \
         --verbose || {
-        log "⚠️  Responses differ between Phase 1 and Phase 2"
-        log "This is expected if block production continued between phases"
-        return 0
+        log "✗ Phase 1 vs Phase 2 differential comparison FAILED"
+        log "  Report: /tmp/comparison-report.json"
+        return 1
     }
 
     log "✓ Phase 1 and Phase 2 responses match"
@@ -1014,14 +1018,18 @@ function deploy_mn2_to_bn2 {
     local mn_importer_image=$(kubectl --context "${CONTEXT}" --namespace "${NAMESPACE}" get deployment mirror-1-importer -o jsonpath='{.spec.template.spec.containers[0].image}' 2>/dev/null || echo "")
     local mn_rest_image=$(kubectl --context "${CONTEXT}" --namespace "${NAMESPACE}" get deployment mirror-1-rest -o jsonpath='{.spec.template.spec.containers[0].image}' 2>/dev/null || echo "")
 
+    # Fallback to the configured GA version with the leading 'v' stripped to match
+    # the Docker image tag convention (e.g. v0.156.0 -> 0.156.0).
+    local mn_image_tag="${MIRROR_NODE_VERSION#v}"
+
     if [ -z "${mn_importer_image}" ]; then
-        log "WARNING: Could not get MN1 importer image, using default"
-        mn_importer_image="gcr.io/mirrornode/hedera-mirror-importer:0.156.0"
+        log "WARNING: Could not get MN1 importer image, using default for version ${MIRROR_NODE_VERSION}"
+        mn_importer_image="gcr.io/mirrornode/hedera-mirror-importer:${mn_image_tag}"
     fi
 
     if [ -z "${mn_rest_image}" ]; then
-        log "WARNING: Could not get MN1 REST image, using default"
-        mn_rest_image="gcr.io/mirrornode/hedera-mirror-rest-java:0.156.0"
+        log "WARNING: Could not get MN1 REST image, using default for version ${MIRROR_NODE_VERSION}"
+        mn_rest_image="gcr.io/mirrornode/hedera-mirror-rest-java:${mn_image_tag}"
     fi
 
     log "Using Mirror Node importer image: ${mn_importer_image}"
@@ -1172,66 +1180,6 @@ spec:
       - name: init-scripts
         configMap:
           name: mn2-postgres-init
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: mn2-rest-config
-  namespace: ${WRB_NAMESPACE}
-data:
-  application.yml: |
-    hiero:
-      mirror:
-        common:
-          shard: 0
-          realm: 0
-        rest:
-          cache:
-            entityId:
-              maxAge: 600
-              maxSize: 100000
-            response:
-              enabled: false
-            token:
-              maxSize: 50000
-          db:
-            host: mirror-1-postgres-postgresql.${NAMESPACE}.svc.cluster.local
-            port: 5432
-            name: mirror_node_wrb
-            username: mirror_node
-            password: mirror_node_pass
-            pool:
-              connectionTimeout: 30000
-              maxConnections: 50
-              statementTimeout: 30000
-            tls:
-              enabled: false
-          log:
-            level: INFO
-          openapi:
-            enabled: false
-            specVersion: openapi
-            validation:
-              enabled: false
-          port: 5551
-          query:
-            maxRecordFileCloseInterval: 2m
-            maxScheduledTransactionConsensusTimestampRange: 30d
-            maxTimestampRange: 7d
-            maxTransactionConsensusTimestampRange: 30d
-            maxTransactionsTimestampRange: 1h
-            maxValidStartTimestampDrift: 24h
-            transactions:
-              precedingTransactionTypes: []
-          redis:
-            enabled: true
-            sentinel:
-              enabled: false
-            uri: redis://mirror-1-redis-node.${NAMESPACE}.svc.cluster.local:6379
-          response:
-            limit:
-              default: 25
-              max: 100
 ---
 apiVersion: apps/v1
 kind: Deployment
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
@@ -1,0 +1,1604 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#
+# Block Ingestion Validation Test
+#
+# Tests that Mirror Node correctly ingests blocks from Block Node.
+# Solo's single-node setup doesn't have TSS metadata, so this validates
+# basic block ingestion without signature verification.
+#
+# Flow:
+#   1. Wait for network to generate blocks
+#   2. Deploy MN → BN1
+#   3. Wait for ingestion
+#   4. Capture MN API responses
+#   5. Validate basic ingestion metrics
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PYTHON_DIR="${SCRIPT_DIR}/python"
+
+NAMESPACE="${NAMESPACE:-solo-network}"
+WRB_NAMESPACE="solo-wrb-test"
+CONTEXT="${CONTEXT:-kind-solo-cluster}"
+DEPLOYMENT="${DEPLOYMENT:-deployment-solo}"
+
+MN_PORT="${MN_PORT:-5551}"
+MN_URL="http://localhost:${MN_PORT}/api/v1"
+MN_CAPTURE_DIR="/tmp/mn-responses"
+MN_WRB_PORT="5552"
+MN_WRB_URL="http://localhost:${MN_WRB_PORT}/api/v1"
+MN_WRB_CAPTURE_DIR="/tmp/mn-wrb-responses"
+
+BLOCK_RANGE="0:100"
+MIN_TRANSACTIONS=10
+
+# Work directories
+WORK_DIR="/tmp/wrb-test-$$"
+RECORD_FILES_DIR="${WORK_DIR}/record-files"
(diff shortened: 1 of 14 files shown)


tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
```diff
@@ -0,0 +1,1133 @@
...
+    fi
+
+    python3 "${PYTHON_DIR}/compare_captured_responses.py" \
+        "${MN_CAPTURE_DIR}" \
+        "${MN_WRB_CAPTURE_DIR}" \
+        --output /tmp/comparison-report.json \
+        --verbose || {
+        log "⚠️  Responses differ between Phase 1 and Phase 2"
+        log "This is expected if block production continued between phases"
+        return 0
```
Nana-EC: Shouldn't this return non 0 response? If not we're missing the actual differential logic no?
There should actually be a test that confirms this.
If this is intentional then the calls for Phase 1 vs 2 should be different so the validation is explicit

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
```diff
@@ -0,0 +1,1133 @@
...
+    matchLabels:
+      app: block-node-2
+  template:
+    metadata:
+      labels:
+        app: block-node-2
+    spec:
+      containers:
+      - name: block-node-server
+        image: ghcr.io/hiero-ledger/hiero-block-node:0.31.0-rc4
```
Nana-EC: version should be an input or at least a GA version initially

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
```diff
@@ -0,0 +1,1137 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: mn2-importer-config
+  namespace: ${WRB_NAMESPACE}
+data:
+  application.yml: |
+    spring:
+      datasource:
+        url: jdbc:postgresql://mirror-1-postgres-postgresql.${NAMESPACE}.svc.cluster.local:5432/mirror_node_wrb?sslmode=disable
```
Nana-EC: Q: Should this be mirror-2-postgres?
Comments say MN2 but I see `...-1-...` in a few places

Nana-EC (changes requested):
Some clarifications on matchup, but also you seem to be deleting unrelated files.
Might need to rebase to clarify changes are all applicable

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
rockysingh: we use the same postgres: with a different db:  `mirror_node_wrb`

.github/workflows/solo-e2e-test.yml
```diff
@@ -227,6 +227,8 @@ jobs:
 
       - name: Install Helm
         uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+        with:
+          version: v4.2.0
```
Nana-EC: Check https://github.com/hiero-ledger/hiero-block-node/pull/3021, this change conflicts and apparently will break solo

rockysingh:
## Known Limitation: Single-Block Wrap on Solo Networks

During investigation of extending block coverage past block 0, we discovered that **WRB CLI only wraps 1 block from a 25-record-file input** when running against a Solo network, even though the input metadata correctly declares `firstBlockNumber: 0` and `lastBlockNumber: 24`.

### Evidence

Diagnostic run output:
```
Processing day files from 2026-06-19 to 2026-06-19
Starting processing day: ...tar.zstd at block: 0
[SIGNATURE DEBUG] Block ...: 1/1 signatures passed validation
Address book updated at block 0: Address Book Changed, via node lifecycle (DAB):
   Node 3 key changed from 2d6d11bf to 300e0603
[==                                                ] 4.00%  | Block 0 ETA: 4s
...
Conversion complete. Blocks written: 1
Total Days:        1
Total Blocks:      1
Total Signatures:  1
```

Only **one** `[SIGNATURE DEBUG]` entry appears for the 25 input record files. WRB exits cleanly (no crash) after block 0 with `Blocks written: 1`.

### Root Cause

Solo's genesis block writes file `0.0.102` (the network address book) as a regular transaction. WRB's `updateAddressBookAndReverify` (in `ToWrappedBlocksCommand.java`) treats this as a Dynamic Address Book (DAB) update and rewrites the in-memory address book at block 0:

1. Block 0 verifies with our extracted `s-public-node1.pem` key (`2d6d11bf...`)
2. Block 0's `0.0.102` transaction triggers WRB's DAB update
3. WRB's `AddressBookRegistry` now uses the key encoding from `0.0.102` (`300e0603...`)
4. Block 1 is signed by the same physical node, but its signature no longer matches the post-DAB book
5. The Stage 1 filter silently drops block 1+ and the loop terminates

This is a key-encoding mismatch between the PEM file we extract from the CN pod and whatever is stored in `0.0.102` at Solo genesis. The actual signing key is the same, but the in-memory comparison `!currentBook.equals(preVerified.addressBook())` fires, and re-verification with the new book fails.

### What This Means for This PR

This PR validates the full MN2 ingestion pipeline (signature verification, address book bootstrap, schema setup, importer behavior). All of that works correctly for the one block WRB produces.

Extending the test to block 1+ requires resolving Solo's genesis bootstrap interaction with WRB's DAB logic, which is a separate engineering problem unrelated to MN2 ingestion. Tracking this would be appropriate as a follow-up issue if multi-block differential coverage is required.

### Workarounds Considered

1. **Disable DAB auto-update in WRB**: would change WRB behavior globally, not just for Solo
2. **Match the `0.0.102` key encoding when extracting from CN pod**: would need to reverse-engineer Solo's genesis state file
3. **Skip block 0's address book update**: hacky, easy to introduce silent correctness bugs

None are appropriate to land as part of this PR.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
rockysingh: Good catch. Fixed in f5f56c2b: the function now returns 1 on a real differential failure. The comparison code already tolerates WRB-expected differences (hash, previous_hash, name, size on block 0; pagination link on the blocks list) so this won't trip on the intentional chain-origin diff, but will fail loudly if a real data mismatch ever appears.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
rockysingh: Fixed in f5f56c2b: added MIRROR_NODE_VERSION env var (default v0.156.0, a GA release). The Solo deployment of MN1 and the standalone MN2 image fallbacks now both honor it.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
rockysingh: The mirror-1-postgres reference was in a stale mn2-rest-config ConfigMap that was defined but never mounted into the mirror-2-rest pod. The actual REST deployment uses inline env vars pointing at mn2-postgres in the solo-wrb-test namespace, so functionally we already had separate postgres for MN2. Removed the dead ConfigMap in f5f56c2b to eliminate the confusing reference.

Nana-EC (approved):

.github/workflows/solo-e2e-test.yml
Nana-EC: Left over comment from early review but I think this was already addressed with David's PR

xin-hedera:
@Nana-EC the different `block0.previousHash` is expected, because

- one stream is native blockstream (not WRB)
- the other stream is WRB (essentially record stream files wrapped in blockstream format)

The genesis blockstream block should have a precalculated non-zero previous block hash.
The genesis recordstream block should always have a all-zero previous hash.

AlfredoG87 (approved):
LGTM

PR #3002, opened 2026-06-11
feat: Helm chart per-plugin port configuration (#3001)

### LINKED ISSUE #3001: Update chart to support custom plugin service ports
### Story Form

As a block node operator
I want to expose the service ports in the helm chart
So that I or helm install/upgrade tools can easily customize the block node during deployment

### Technical Notes

With https://github.com/hiero-ledger/hiero-block-node/pull/2983 it's possible to set config values that specify each plugins service port.
However, the block node in production is installed with helm so the helm chart must be updated to ensure a smooth experience for node operators.
It is possible to set each plugins port config under the blockNode.config.<plugin>.port. 
However, the chart also has support for `blockNode.hostPorts`, `service.port` and `loadBalancer` settings that need to be considered as they have previously assumed the all services were on a single port.

Nana-EC:
## Reviewer Notes

PR #2983 added optional per-plugin port overrides to the block node application (`PRODUCER_PORT`, `SUBSCRIBER_PORT`, `BLOCK_ACCESS_PORT`, `HEALTH_PORT`, `SERVER_STATUS_PORT`). This PR updates the Helm chart so that node operators can configure those ports through a structured, documented `blockNode.ports` section and have all downstream Kubernetes resources (container ports, Service, ConfigMap env vars, probes) updated automatically.

**Before this PR**, an operator who set e.g. `HEALTH_PORT: "8080"` in `blockNode.config` would get:
- the env var injected ✓
- container port `8080` **not declared** ✗
- Service port `8080` **not exposed** ✗
- liveness/readiness probes still targeting the old `service.port` → **pod never becomes Ready** ✗

**After this PR**, setting `blockNode.ports.health: 8080` handles all of the above automatically, and the probes correctly redirect to port `8080`.

### `values.yaml` — new `blockNode.ports` section

```yaml
blockNode:
  ports:
    publisher: null     # PRODUCER_PORT    — gRPC, block producers
    subscriber: null    # SUBSCRIBER_PORT  — gRPC, block subscribers
    blockAccess: null   # BLOCK_ACCESS_PORT — gRPC/HTTP, block query API
    health: null        # HEALTH_PORT      — HTTP, /healthz endpoints
    serverStatus: null  # SERVER_STATUS_PORT — HTTP, server-status endpoint
```

All fields default to `null` — **zero behaviour change for existing deployments**.

### `templates/_helpers.tpl` — three new YAML-output helpers

| Helper | Used by |
|---|---|
| `hiero-block-node.pluginPortEnvVars` | `configmap.yaml` |
| `hiero-block-node.pluginServicePorts` | `service.yaml` |
| `hiero-block-node.pluginContainerPorts` | `statefulset.yaml` |

Helpers emit raw YAML and are composed with `include … | nindent`. No JSON round-trip — works with Helm 3.11+.

### `templates/statefulset.yaml` — two fixes

1. **Dynamic container ports**: each non-null `blockNode.ports` entry is declared as a named `containerPort` (with optional `hostPort` if set in `blockNode.hostPorts`).
2. **Health probe port fix**: liveness and readiness probes now use `blockNode.ports.health` when set, falling back to `service.port` when null.

### `templates/service.yaml`

Each non-null `blockNode.ports` entry is added as a named port to the ClusterIP Service so in-cluster traffic can reach any plugin that has moved off the shared port.

### `templates/configmap.yaml`

Plugin port env vars are injected from `blockNode.ports` before the `blockNode.config` range, so explicit values in `blockNode.config` still take precedence if both are set.

### `NOTES.txt` (new)

Rendered at install/upgrade time. Includes:
- Active per-plugin port summary and port-forward commands
- Health endpoint URLs (resolved against the actual health port)
- **Migration warning** if `HEALTH_PORT` is found in `blockNode.config` but `blockNode.ports.health` is not set — this is the misconfiguration that causes silent probe failure

### `README.md`

New _Per-Plugin Port Configuration_ section covering:
- How to configure `blockNode.ports` with a worked example
- The conflict warning (don't set both `blockNode.ports.X` and the raw env var in `blockNode.config`)
- The health probe safety note
- LoadBalancer guidance: single-IP/multi-port use case and the follow-on planned for multi-port LB support

### `ci/multi-port-values.yaml` (new)

CI test values used by both `ct install` and the new workflow job:

| Plugin | Port |
|---|---|
| health | 40940 |
| publisher | 40941 |
| subscriber | 40942 |
| blockAccess | 40943 |
| serverStatus | 40944 |
| server.port (shared fallback) | 40840 — **no plugin uses this** |

### `.github/workflows/helm-charts.yaml` — new `multi-port-install-test` job

Independent job that:
1. Builds the block-node Docker image and loads it into a fresh kind cluster
2. Installs the chart with `ci/multi-port-values.yaml`
3. Waits for pod ready (readiness probe on port 40940 — the health port)
4. Port-forwards all five plugin ports and runs per-port connectivity checks:
   - **40940** (health): `curl /healthz/livez` — full HTTP response check
   - **40941–40943** (gRPC): TCP connect — verifies port is open and accepting connections
   - **40944** (server-status): TCP connect
   - **40840** (server.port): asserts the health plugin is **not** responding here, confirming per-plugin port routing is active end-to-end

Query logic is intentionally minimal and will be expanded in a follow-on.

## Backward Compatibility

| Scenario | Impact |
|---|---|
| No `blockNode.ports` set (all null) | Zero change — helpers emit nothing; probes use `service.port` as before |
| `blockNode.config` with raw env vars (old pattern) | Continues to work; probes remain on `service.port` unless `blockNode.ports.health` is also set. NOTES.txt emits a migration warning. |
| Existing `loadBalancer.port/portName` | Unchanged — the LB template was not modified |
| `blockNode.hostPorts` with `http`/`metrics` | Unchanged — still supported; new plugin port names can also be added |

## Testing

- `helm lint charts/block-node-server` — clean (default values)
- `helm lint charts/block-node-server -f charts/block-node-server/ci/multi-port-values.yaml` — clean
- `helm template` verified: service ports, container ports, configmap env vars, and probe ports all render correctly for both default and multi-port values

## Related Issue(s)
Resolves #3001

### DIFF
--- a/.github/workflows/helm-charts.yaml
+++ b/.github/workflows/helm-charts.yaml
@@ -96,3 +96,145 @@ jobs:
       - name: Run chart-testing (install)
         if: steps.install-needed.outputs.needed == 'true'
         run: ${CGEXEC} ct install --config ct.yaml --all
+
+  multi-port-install-test:
+    timeout-minutes: 20
+    name: Multi-Port Plugin Install Test
+    runs-on: hl-bn-lin-md
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          egress-policy: audit
+
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          fetch-depth: 0
+
+      - name: Check if install test is needed
+        id: install-needed
+        run: |
+          changed_files=$(git diff --name-only origin/${{ github.event.repository.default_branch }}...HEAD)
+          if echo "$changed_files" | grep -qE "^(block-node/app/(docker/|build\.gradle\.kts)|charts/|tools-and-tests/simulator/(docker/|build\.gradle\.kts))"; then
+            echo "needed=true" >> "$GITHUB_OUTPUT"
+            echo "Image/chart/build files changed - will run multi-port test"
+          else
+            echo "needed=false" >> "$GITHUB_OUTPUT"
+            echo "No relevant changes - skipping multi-port install test"
+          fi
+
+      - name: Set up Helm
+        if: steps.install-needed.outputs.needed == 'true'
+        uses: Azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+        with:
+          version: v3.14.4
+
+      - name: Set up Java
+        if: steps.install-needed.outputs.needed == 'true'
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
+        with:
+          distribution: temurin
+          java-version: 25
+
+      - name: Create kind cluster
+        if: steps.install-needed.outputs.needed == 'true'
+        uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc # v1.14.0
+
+      - name: Build Docker images
+        if: steps.install-needed.outputs.needed == 'true'
+        run: ${CGEXEC} ./gradlew :app:createDockerImage
+
+      - name: Load image into kind
+        if: steps.install-needed.outputs.needed == 'true'
+        run: |
+          APP_VERSION=$(cat version.txt)
+          docker tag "block-node-server:${APP_VERSION}" "ghcr.io/hiero-ledger/hiero-block-node:${APP_VERSION}"
+          kind load docker-image "ghcr.io/hiero-ledger/hiero-block-node:${APP_VERSION}" --name chart-testing
+
+      - name: Update chart dependencies
+        if: steps.install-needed.outputs.needed == 'true'
+        run: |
+          helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+          helm repo add grafana https://grafana.github.io/helm-charts
+          helm dependency update charts/block-node-server
+
+      - name: Install chart with multi-port values
+        if: steps.install-needed.outputs.needed == 'true'
+        run: |
+          APP_VERSION=$(cat version.txt)
+          helm install bn-multi-port-test charts/block-node-server \
+            --values charts/block-node-server/ci/multi-port-values.yaml \
+            --set image.tag="${APP_VERSION}" \
+            --wait \
+            --timeout 300s
+
+      - name: Verify per-plugin ports are reachable
+        if: steps.install-needed.outputs.needed == 'true'
+        run: |
+          POD=$(kubectl get pod -l app.kubernetes.io/instance=bn-multi-port-test -o jsonpath='{.items[0].metadata.name}')
+          echo "Testing pod: ${POD}"
+
+          # Forward all five plugin ports to the local runner
+          kubectl port-forward "pod/${POD}" 40940:40940 40941:40941 40942:40942 40943:40943 40944:40944 &
+          PF_PID=$!
+          # Give port-forward time to establish
+          sleep 5
+
+          failed=0
+
+          # Health port (40940): HTTP liveness endpoint
+          echo "--- port 40940 (health) ---"
+          if curl -sf --max-time 5 http://localhost:40940/healthz/livez; then
+            echo "OK"
+          else
+            echo "FAIL: health port 40940 did not respond"
+            failed=1
+          fi
+
+          # Publisher port (40941): TCP connect (gRPC)
+          echo "--- port 40941 (publisher/gRPC) ---"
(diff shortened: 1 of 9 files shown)


charts/block-node-server/NOTES.txt
```diff
@@ -0,0 +1,73 @@
...
+Then connect gRPC clients to localhost:{{ .Values.service.port }}.
+
+{{- if or .Values.blockNode.ports.publisher .Values.blockNode.ports.subscriber .Values.blockNode.ports.blockAccess .Values.blockNode.ports.health .Values.blockNode.ports.serverStatus }}
+
+==> Per-Plugin Ports Active
+
+The following plugins have been assigned dedicated ports:
+{{- with .Values.blockNode.ports }}
+{{- if .publisher }}
+  publisher   (PRODUCER_PORT)    → {{ .publisher }}
```
AlfredoG87: should we replace arrow with `->`

same on lines 24,26,28 and 30

.github/workflows/helm-charts.yaml
```diff
@@ -96,3 +96,145 @@ jobs:
...
+            failed=1
+          fi
+
+          # Confirm that the shared server.port (40840) is NOT serving plugin traffic
+          echo "--- port 40840 (server.port — must NOT have plugin traffic) ---"
+          if curl -sf --max-time 5 http://localhost:40840/healthz/livez 2>/dev/null; then
+            echo "WARN: health plugin responded on server.port 40840 — expected it to be on 40940 only"
+          else
+            echo "OK (health plugin correctly not on 40840)"
+          fi
```
AlfredoG87: This will always fail regardless since we are not doing port-forward for port `40840` in order to really verify, we need to actually port-forward.

charts/block-node-server/NOTES.txt
```diff
@@ -0,0 +1,73 @@
...
+  kubectl port-forward svc/{{ include "hiero-block-node.fullname" . }} \
+  {{- with .Values.blockNode.ports }}
+  {{- if .publisher }} {{ .publisher }}:{{ .publisher }} \{{- end }}
+  {{- if .subscriber }} {{ .subscriber }}:{{ .subscriber }} \{{- end }}
+  {{- if .blockAccess }} {{ .blockAccess }}:{{ .blockAccess }} \{{- end }}
+  {{- if .health }} {{ .health }}:{{ .health }} \{{- end }}
+  {{- if .serverStatus }} {{ .serverStatus }}:{{ .serverStatus }}{{- end }}
+  {{- end }}
+
+{{- end }}
```
AlfredoG87: port-forward command template produces invalid shell (inline `\` instead of line continuations)
```
kubectl port-forward svc/NAME \ 40941:40941 \ 40942:40942 \ 40943:40943 \...
```

The `\` characters are standalone shell arguments, not line continuations (those require `\` immediately before a newline). kubectl port-forward will reject the `\` args. For subsets where serverStatus is not the last active port, there's also a trailing `\`. Fix: remove the backslash-continuation pattern and
  just emit one port per line without continuations, or use a range over a pre-built list so indentation is controlled in one place.

charts/block-node-server/templates/_helpers.tpl
```diff
@@ -92,6 +92,121 @@ Build JAVA_TOOL_OPTIONS combining logging config and metrics system properties.
 {{- end }}
 {{- end }}
 
+{{/*
+Emit ConfigMap data lines (KEY: "value") for each non-null plugin port in blockNode.ports.
+These are emitted before the blockNode.config range so that explicit config values win on conflict.
+Usage: include "hiero-block-node.pluginPortEnvVars" . | nindent 0
```
AlfredoG87: we should remove the `| nindent 0`

configmap.yaml correctly omits | nindent 0. Adding nindent 0 would prepend an extra newline to content that already starts with \n, producing a blank line in the ConfigMap.

AlfredoG87 (approved):
LG! 💯

.github/workflows/helm-charts.yaml
```diff
@@ -96,3 +96,148 @@ jobs:
...
+          # Block-access port (40943): TCP connect (gRPC)
+          echo "--- port 40943 (block-access/gRPC) ---"
+          if timeout 5 bash -c 'echo > /dev/tcp/localhost/40943' 2>/dev/null; then
+            echo "OK"
+          else
+            echo "FAIL: block-access port 40943 not reachable"
+            failed=1
+          fi
+
+          # Server-status port (40944): TCP connect (HTTP)
```
jsync-swirlds, line 223: Server Status is gRPC, not HTTP.

jsync-swirlds (approved):

.github/workflows/helm-charts.yaml
Nana-EC: Will cover in follow up

PR #3021, opened 2026-06-16
test(solo-e2e): WRAPS v1.0.0 transition support and WRB RSA verification improvements

AlfredoG87:
## Summary

Solo-e2e improvements in two areas:

1. **WRAPS** — make `task down` + `task up` reliably produce a Schnorr -> WRAPS signature transition using WRAPS v1.0.0.
2. **WRB RSA** — add a verification task and speed up RSA roster loading.

## WRAPS v1.0.0 transition

The consensus node builds its genesis history construction within seconds of going ACTIVE. If the WRAPS proving-key library is not ready by then, CN v0.75.0-rc.6 permanently drops the WRAPS proof (`WrapsHistoryProver - Skipping publication of POST_AGGREGATION output: WRAPS library is not ready`), construction #1 stays `WRAPS-extensible? false`, and the network never transitions off Schnorr. Solo's default WRAPS is v0.2.0, which is incompatible with current consensus-node releases.

CN-side runtime download (`tss.wrapsProvingKeyDownloadEnabled=true`) downloads + verifies the ~2 GB key fine, but it runs async and finishes minutes after genesis — always too late. The readiness-retry that would let a late download succeed (`Deferring ... will retry each consensus round until ready`) only exists on consensus-node `main`, not in any released/rc version. So the proving key must be on disk before the JVM starts.

Changes in `scripts/solo-deploy-network.sh`:

- New `ensure_wraps_keys_cached`: downloads and extracts the WRAPS v1.0.0 archive once into `~/.solo/cache/wraps-v1.0.0-keys` (the four `.bin` files), skipping the download on later deploys.
- `deploy_consensus_nodes` passes `--wraps true --wraps-key-path "${WRAPS_KEYS_DIR}"` to `solo consensus network deploy` (previously bare `--wraps true`), so Solo stages the local v1.0.0 keys before the JVM starts and the library is ready at genesis. `--tss` is unchanged so Solo still patches `block-nodes.json` with the TSS message-size limits the large WRAPS genesis block needs.
- No changes to the consensus node `application.properties` (still just `tss.wrapsEnabled=true`; CN-side download stays disabled).

Validate:

```
task down
task up
task check:signature   # SCHNORR 2920 bytes -> WRAPS 3432 bytes
```

The genesis WRAPS proof takes ~15 min of single-node compute, so the flip lands a few hundred blocks in (observed first WRAPS block 437); the block where it flips is oversized (~31 MB) and the Block Node returns NOT_AVAILABLE for it (expected).

Note: Solo stages the keys under its cache directory named `wraps-v0.2.0` (its default `directoryName`, not overridable here), so that directory holds v1.0.0 contents on this setup.

## WRB RSA verification improvements

`Taskfile.yml`:

- New `verify:wrb` task (`NODE=n`, default 1): checks RSA-roster WRB verification on a Block Node via metrics — `blocknode_roster_entries_loaded`, `blocknode_verification_proof_total{proof_type="rsa"}` success/failure, and `blocknode_rsa_roster_mismatch_total`. Passes when the roster is loaded, there are RSA-verified blocks, and no mismatches; startup-window RSA failures (blocks received before the roster loaded) are reported as expected.

`scripts/solo-deploy-network.sh`:

- Lower `ROSTER_BOOTSTRAP_RSA_INITIAL_QUERY_INTERVAL_MILLIS` from `15000` to `3000` so the RSA roster loads sooner after the Mirror Node is up, narrowing the startup window in which an early block can arrive before the roster.

## Notes

- The underlying Solo limitations behind the WRAPS workaround (v0.2.0 pin, absolute `TSS_LIB_WRAPS_ARTIFACTS_PATH` rejected in CN download mode, ineffective version overrides) are written up for the Solo team in `.claude/tmp/solo-wraps-issue.md`, and the workaround is documented in `.claude/knowledge/project-patterns.md`.

### DIFF
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -227,6 +227,10 @@ jobs:
 
       - name: Install Helm
         uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+        with:
+          # Pin to Helm 3.x: Solo cannot parse Helm 4.x install output
+          # ("Failed to deserialize the output into the specified class: Release").
+          version: v3.14.4
 
       - name: Setup Kind
         uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc # v1.14.0
--- a/tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
+++ b/tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
@@ -247,6 +247,52 @@ tasks:
     cmds:
       - '"{{.SCRIPTS_DIR}}/monitor-block-proofs.sh" "{{.PROTO_PATH}}" "localhost:40840"'
 
+  verify:wrb:
+    desc: "Verify WRB blocks are accepted via the RSA roster on a Block Node (NODE=n, default: 1). Needs port-forward."
+    vars:
+      NODE: '{{.NODE | default "1"}}'
+    cmds:
+      - |
+        PORT=$((16006 + {{.NODE}}))
+        echo "Verifying RSA-roster WRB verification on Block Node {{.NODE}} (metrics: localhost:${PORT})..."
+        METRICS=$(curl -s --max-time 10 "http://localhost:${PORT}/metrics" 2>/dev/null)
+        if [[ -z "${METRICS}" ]]; then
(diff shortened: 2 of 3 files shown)


Nana-EC (approved):
LG

berryware (approved):
Looks good

PR #3023, opened 2026-06-16
feat: Cloud Storage Archive Plugin to Support EMB-aware Startup and Temporary Archives

### LINKED ISSUE #2987: Cloud Storage Archive Plugin to Support EMB-aware Startup and Temporary Archives
### Story Form

I want to start archiving from the next clean group boundary after `earliestManagedBlock` (EMB), write temporary archives for blocks that arrive before that boundary, and consolidate them once the full group range is available
So that no verified or backfilled block is silently dropped or indefinitely held in memory

### Technical Notes

- Read `NodeConfig#earliestManagedBlock()` during `start()`. The first regular archive starts at the next clean group boundary after EMB: `((EMB / groupSize) + 1) * groupSize`. When EMB is 0, the boundary is 0
- Introduce _temporary archives_: S3 objects covering an incomplete block range, started when blocks arrive that cannot yet fill a complete aligned group and ended when a gap appears or a group boundary is reached
- A temporary archive must use a key format clearly distinguishable from a final archive (e.g. a `.tmp` suffix or a separate prefix segment) so that `StartupRecoveryTask` and `findLastKey` can differentiate them during startup listing
- Only one regular archive upload runs at a time. Multiple temporary archive uploads may run simultaneously but are bounded (configurable with a plugin config property)
- Gap management is backfill's responsibility. The archive plugin constructs complete archives; it does not fill gaps or wait for predecessors
- The plugin must be aggressive: blocks must not be held in memory waiting for a specific predecessor. The in-memory cross-group stash (`blockStash`) is replaced by immediate upload to temporary S3 archives.
- Introduce a _durable temporary archive tracker_ separate from `storedBlocks`. `storedBlocks` is published application state shared with backfill and other consumers
- The _durable tracker_ is internal plugin state. It needs to know, per temp archive: the exact S3 key (so it can download parts for merging), the block range (so it can check whether a complete group is covered), and whether the multipart upload is still open
- `storedBlocks` must be updated to include blocks in temporary archives (not only final archives), so that backfill does not re-fetch blocks already durably stored in cloud
- On startup, list the bucket (scoped by `objectKeyPrefix` when configured) to reconcile actual cloud state with the durable tracker and rebuild both `storedBlocks` and the local final-archive set
- Implement a _consolidation step_: once backfill or live stream delivers all missing blocks for a temporary group, merge the relevant temporary archives into a single final archive
 - On startup, scan both the regular and temporary key spaces, reconcile against the durable tracker, and resume any open uploads using the existing hanging-upload recovery logic

ivannov:
### EMB-aware startup boundary
  
During `init()` the plugin reads `NodeConfig#earliestManagedBlock()` and computes `firstRegularGroupStart`: the first aligned  group boundary at or above the EMB. Regular `BlockUploadTask` instances only start for blocks at orabove that boundary; everything below it goes directly to temporary archives.

### Temporary archives

Blocks that cannot enter the regular pipeline are uploaded immediately to S3 as streaming multipart uploads via the new `TempArchiveUploadTask`. Each segment writes to a temporary "directory". When the segment closes (gap detected, last block of group) the task completes the upload and atomically writes a _meta_ companion whose sole content is the decimal `lastBlock`. A completed `.meta` file is the only durability signal; StartupRecovery trusts this as the ground truth without downloading the tar content.

The in-memory cross-group stash (`blockStash`) is held only while the `StartupRecoveryTask` is in flight.

Concurrent temp-upload slots are bounded by the new `maxConcurrentTempArchives` config property. Overflow blocks are held in `tempOverflowStash` (a `TreeMap`), drained in block-number order each time a slot frees up.

### Consolidation

`checkGroupCoverage()` walks completed `TempArchiveEntry` records for a group in ascending order and checks whether they cover the full `[groupStart, groupStart+groupSize)` range without gaps.  When the range is fully covered, the group is handed to a `ConsolidationTask`, which downloads each `.tmp` object in order and pipes its raw bytes into a new multipart upload for the final `ArchiveKey`-formatted key (no re-encoding -- temp and final archives share the same tar-entry format). After completion the `.tmp` and `.meta` companions are deleted.

### Startup recovery extension

`StartupRecoveryTask` now additionally scans the `tmp/` virtual directory. 

## Related Issue(s)

Closes #2987


block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ConsolidationTask.java
```diff
@@ -0,0 +1,155 @@
...
+
+/// Merges a complete set of temporary archives covering one aligned group into a single final
+/// `.tar` archive and removes the temporary objects from S3.
+///
+/// The caller guarantees that the [entries] list, sorted by [TempArchiveEntry#firstBlock()],
+/// together cover exactly the block range `[groupStart, groupStart + groupSize)`.  The task:
+///
+/// 1. Creates a new multipart upload at [ArchiveKey#format(long, int, String)].
+/// 2. Downloads each `.tmp` object and pipes its raw bytes directly into the upload (no tar
+///    re-encoding — temp archives and final archives share the same tar-entry format).
```
jsync-swirlds, line 26: This is OK as a first pass, but raises a massive risk of a corrupted archive (whether accidental or deliberate).
I think we need a follow-up task to change this to actually read the tar content, process each block in that content, and write the final tar from those.
This would also open the _possibility_ for partial consolidation, which might be of value in some cases.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -203,31 +269,160 @@ private boolean checkCompletedUpload() throws ExecutionException {
...
+                LOGGER.log(TRACE, "Temp archive completed: blocks [{0}, {1}]", result.firstBlock(), result.lastBlock());
+            } catch (IllegalStateException e) {
+                LOGGER.log(INFO, "Temp archive upload failed for firstBlock {0}", entry.getKey(), e);
+                metricsHolder.failedTasks().increment();
+            }
+        }
+    }
+
+    /// Submits pending [ConsolidationTask]s and drains completed ones.
+    private void checkAndDrainConsolidations() {
```
Nana-EC, line 312: Logic is good but I think we need a test that checks the outer plugin-level retry i.e. it checks that after a consolidation failure the group is re-added to `pendingConsolidations` and a second `ConsolidationTask` eventually succeeds.

Suggestion you can do in follow up is
Add a test in `CloudStorageArchivePluginTest` (probably in the existing temp-archive nested class) that uses a `BlockingExecutor` to:
1. Drive blocks through a full temp-archive group.
2. Make the first `ConsolidationTask` fail (via a test-subclass override).
3. Assert that the group reappears in `pendingConsolidations`.
4. Run a second successful consolidation and assert the final tar is present and the temp objects are deleted.

Nana-EC (commented):
Still reviewing, but might be valuable to expand the tests with one more

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
```diff
@@ -15,9 +14,9 @@
 import java.util.List;
 import java.util.Map;
 import java.util.SortedMap;
+import java.util.TreeMap;
```
jsync-swirlds: This is not thread safe, I recommend using the ConcurrentSkipListMap that was already in use.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
```diff
@@ -162,7 +155,7 @@ public UploadResult call()
                     ? resumeState.trailingBytes()
                     : new byte[0];
             // Maps each buffered block number to its source, in insertion order.
-            final SortedMap<Long, BlockSource> blocksInBuffer = new ConcurrentSkipListMap<>();
+            final SortedMap<Long, BlockSource> blocksInBuffer = new TreeMap<>();
```
jsync-swirlds: Why choose a TreeMap instead of a ConcurrentSkipListMap?
The latter is at least as efficient and also thread safe.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -85,27 +97,70 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
...
+    final NavigableMap<Long, BlockWithSource> tempOverflowStash = new TreeMap<>();
+
+    /// Groups whose temp archives fully cover [groupStart, groupStart+groupSize), queued for
+    /// consolidation.  Keyed by groupStart.
+    /// Accessed only from the notification handler thread.
+    final NavigableMap<Long, List<TempArchiveEntry>> pendingConsolidations = new TreeMap<>();
+
+    /// In-flight consolidation task futures keyed by groupStart.
+    /// Accessed only from the notification handler thread.
+    final Map<Long, Future<UploadResult>> consolidationFutures = new HashMap<>();
```
jsync-swirlds: 1. These values are used quite frequently, we should avoid the virtual call overhead and just use concrete types for the variables.
1. Unless the Key type is not `Comparable` (and for all of these it is) we should probably prefer `ConcurrentSkipListMap` over `HashMap` or `TreeMap`.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -131,6 +186,12 @@ public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
             context.blockMessaging().registerBlockNotificationHandler(this, false, "Cloud Storage Archive");
             configValid = true;
         }
+        groupSize = Math.powExact(10, config.groupingLevel());
+        final long emb = context.configuration().getConfigData(NodeConfig.class).earliestManagedBlock();
+        // Round EMB up to the nearest group boundary: groups starting at or above this point use
+        // the regular pipeline; groups starting below it (including those straddling the EMB) go
+        // to temp archives.
+        firstRegularGroupStart = ((emb + groupSize - 1) / groupSize) * groupSize;
```
jsync-swirlds, line 193: There is a bit more than EMB to consider.
At _any time_ we encounter a gap we should make used of temporary archives.
If we do that, use a temporary archive any time a gap is encountered, then we don't need to pay any attention to EMB in this plugin. We will just store what we can (whether in primary archives or temporary archives) and reassemble into complete archives whenever we can.
In many cases, the temporary archives may even self-complete.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -203,31 +269,160 @@ private boolean checkCompletedUpload() throws ExecutionException {
                 } else {
                     metricsHolder.successfulTasks().increment();
                     currentUploadFuture = null;
-                    currentGroupPending = new ConcurrentSkipListMap<>();
+                    currentGroupPending = new TreeMap<>();
```
jsync-swirlds: Why replace this with `TreeMap`?

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -203,31 +269,160 @@ private boolean checkCompletedUpload() throws ExecutionException {
...
+                tempArchiveTracker.put(result.firstBlock(), result);
+                metricsHolder.successfulTasks().increment();
+                final long groupStart = (result.firstBlock() / groupSize) * groupSize;
+                checkGroupCoverage(groupStart);
+                LOGGER.log(TRACE, "Temp archive completed: blocks [{0}, {1}]", result.firstBlock(), result.lastBlock());
+            } catch (IllegalStateException e) {
+                LOGGER.log(INFO, "Temp archive upload failed for firstBlock {0}", entry.getKey(), e);
+                metricsHolder.failedTasks().increment();
+            }
+        }
```
jsync-swirlds: Instead of looping twice to filter then process; why not just loop once.

```suggestion
        for (final Map.Entry<Long, Future<TempArchiveEntry>> entry : tempUploadFutures.entrySet()) {
            if (entry.getValue.isDone()) {
                tempUploadFutures.remove(entry.getKey());
                try {
                    final TempArchiveEntry result = entry.getValue().resultNow();
                    tempArchiveTracker.put(result.firstBlock(), result);
                    metricsHolder.successfulTasks().increment();
                    final long groupStart = (result.firstBlock() / groupSize) * groupSize;
                    checkGroupCoverage(groupStart);
                    LOGGER.log(TRACE, "Temp archive completed: blocks [{0}, {1}]", result.firstBlock(), result.lastBlock());
                } catch (IllegalStateException e) {
                    LOGGER.log(INFO, "Temp archive upload failed for firstBlock {0}", entry.getKey(), e);
                    metricsHolder.failedTasks().increment();
                }
            }
        }
```

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -203,31 +269,160 @@ private boolean checkCompletedUpload() throws ExecutionException {
...
+    }
+
+    /// Submits pending [ConsolidationTask]s and drains completed ones.
+    private void checkAndDrainConsolidations() {
+        // Process completed consolidations.
+        final List<Long> doneGroups = consolidationFutures.entrySet().stream()
+                .filter(e -> e.getValue().isDone())
+                .map(Map.Entry::getKey)
+                .toList();
+        for (final long groupStart : doneGroups) {
```
jsync-swirlds: Similar comment, instead of filter-then-loop (which runs two loops), we may wish to use a single loop instead.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: At comment 1:

Here: https://github.com/hiero-ledger/hiero-block-node/pull/2473#discussion_r3021758863 the feedback was for me to use the interface type (SortedMap) rather than the concrete type. I followed that guidance here as well. If the team now prefers concrete types for these fields, I'm happy to update them. I'd appreciate a consistent guideline

At comment 2:

Answered in another comment

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
ivannov: To start with: I changed all over the place the types of the maps from concurrent to non-concurrent, because they are only used in a single thread. And there is price we pay for concurrent maps.

I tried to take this decision driven by data, so I created this JMH benchmark to compare `TreeMap` to `ConcurrentSkipListMap`: https://github.com/ivannov/treemap-vs-concurrentmap-benchmark

Here's the result of a run on my laptop:

https://gist.github.com/ivannov/18776c804fa06cd2e53d8c986ec486ad

This comes to prove that a tree map outperforms the concurrent counterpart in a single-thread environment (apart from the iteration case and maybe massive removal case as well).

That is why my proposal is to stick to the non-concurrent types. But if you wish, I can revert back.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Answered in another comment

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Changed with a single loop, although used iterator to avoid concurrent modification exception

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Thanks for catching, added the missing test. Also found another possible bug (midway failing temp archive task) and added support and test for that too

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Implemented here as well with single loop

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ConsolidationTask.java
ivannov: Created this task:

https://github.com/hiero-ledger/hiero-block-node/issues/3041

Can you please check whether it captures what you intended?

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Well, EMB was the central part of the requirement as I put it in the issue.

If you think that we should really extend it, I may create another follow-up issue.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
jsync-swirlds: It is difficult to cleanly compare the two map types as performance is highly dependent on the exact data and ordering.  Red-Black trees will outperform sometimes, and Skip Lists will outperform other times.
This is further confounded by the probabilistic nature of Skip Lists.

Also, Skip Lists almost always outperform when _iterating_ over a map; and (at least from what I see) it looks like iteration is the more common interaction for most of these.

In general, however, these maps are pretty small, and the differences are small as well.

We've used concurrent maps for almost all Maps in the system as a pattern so we don't need to think as hard about whether any particular map might be used from multiple threads, even if it's currently used on only one.

I do not intend to _require_ any particular option.  My intent is to ensure we're thinking through all of the considerations and making consistent choices.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: Good point regarding the interfaces, we've been inconsistent in expecting interfaces or concrete types.

There are two items in tension here:
1. Interfaces provide more flexibility and can reduce code churn
1. Interfaces introduce vcall overhead, which gets bad when a variable is accessed with high frequency.

I don't know that there is a "guideline" level recommendation here that always applies. We need to think through each case individually.

If a variable is used in a tight loop where performance might be important, we should use a concrete type. If the variable is never used in loops or is only referenced at fairly low frequency, then the balance tips more toward using an interface.

This set of items appear to be used in loops a lot, hence my comment, but if those loops are rare or not performance critical, then we can leave the interface types in place here.

I answered the concurrent types in the other comment.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: Please do create that follow up.  While the issue was written in terms of EMB, I believe that was based on flawed assumptions and over-specifying.  The _underlying requirement_ is to handle gaps, not just support EMB.  Of particular note is that EMB will be `0` in the most critical environments, so this plugin must handle gaps and out-of-order backfill when EMB is `0`.

People creating issues are working from limited time and information (or, increasingly, asking a LLM/GPT to create the text), which means there is always some question whether a given issue is complete or entirely correct.

We should not, in my opinion, take issues as "do exactly this" but rather as a starting point for possible conversations to ensure we're doing the work that is needed, rather than uncritically accepting what is written.

ivannov: @jsync-swirlds does this capture what you think we should cover: https://github.com/hiero-ledger/hiero-block-node/issues/3045

jsync-swirlds: That's close.
For the following:
> Replace the EMB-based routing condition with a continuity check: compare the incoming block number against the last block handed off to any upload task.

Perhaps include that we should still allow for a small (perhaps 5 blocks?) buffer held in memory (and not notified as persisted yet) to avoid splitting off to a temporary archive too easily.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ConsolidationTask.java
jsync-swirlds: That's a good issue to resolve, but I don't see how it addresses the possibility that the temp file tar content might be corrupted or otherwise need to be verified as we read it.

Currently the code (per this comment) just does a binary concatenation and assumes that the tar content is always clean and complete blocks, that are zstd compressed correctly, and comprise the exact range indicated in the metadata file.

I believe we should read and process the tar data as it's received from the S3 API, to ensure we are retrieving the individual block files before feeding them back to a consolidation task that writes the data to a final consolidated tar file.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Updated the task

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ConsolidationTask.java
ivannov: Not sure whether I understand you correctly. But we do not do such checks for the "regular" archives. We just trust our code that it doesn't create garbage.

We use the same code to generate the temporary archives as well. So why would we not trust it there?

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
ivannov: Reverted back to concurrent structures

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Used concrete variable types for variables participating in loops

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ConsolidationTask.java
jsync-swirlds: The difference is that regular files are written in a single upload that isn't generally visible to other users, whereas these temp archives are closed (and visible to other users) for potentially extended time before they're read back.

It's a resiliency issue that may be deferred for a while, but I would like it captured in detail so we don't lose it.

ivannov: OK, got it. I added one more note to #3041

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -263,30 +470,59 @@ private void logInvalidOrFailedNotification(VerificationNotification notificatio
...
+                    TRACE,
+                    "Block {0} is below firstRegularGroupStart {1}; routed to temp archive instead of regular pipeline",
+                    blockNumber,
+                    firstRegularGroupStart);
+        }
+    }
+
+    /// Returns `true` when the group starting at [groupStart] has any in-flight or completed
+    /// temporary archive data (active streaming segments, in-flight futures, or tracker entries).
+    private boolean hasAnyTempDataForGroup(long groupStart) {
```
Nana-EC, line 509: This is an interesting area.

My understanding is `drainOverflowStash()` is called from `checkAndDrainTempUploadResults()` and restores blocks in ascending key order via `pollFirstEntry()`. No test exercises this path.
It seems the default `maxConcurrentTempArchives = 4` is always large enough to keep all groups active in every test. The gap matters because the interaction between stashing, `tempGroupNextExpected`, and gap detection in `routeToTempArchive` is non-trivial: a stashed block whose group later receives a normally-arriving block will trigger a segment close on the gap path, potentially fragmenting the group into many one-block segments before coverage is achieved.

I'd suggest in follow up adding one test in `CloudStorageArchivePluginTest` that sets `maxConcurrentTempArchives = 1` and sends blocks for two different pre-EMB groups interleaved, verifying that all blocks end up in `tempArchiveTracker` with full coverage and no blocks are stranded in `tempOverflowStash`.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -383,10 +705,14 @@ public void stop() {
...
-        // cancel(true) interrupts the virtual thread at blockQueue.take(), so the task completes
-        // quickly without needing an explicit get() here.
+        tempUploadFutures.values().forEach(f -> f.cancel(true));
+        consolidationFutures.values().forEach(f -> f.cancel(true));
         currentGroupPending.clear();
         currentBlockQueue.clear();
+        tempGroupActiveQueues.clear();
+        tempUploadFutures.clear();
+        pendingConsolidations.clear();
+        consolidationFutures.clear();
```
Nana-EC, line 728: I think we're missing the clearing of tempArchiveTracker, tempGroupNextExpected and tempOverflowStash

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -211,23 +276,165 @@ private boolean checkCompletedUpload() throws ExecutionException {
...
+
+    /// Checks whether the temp archives in [tempArchiveTracker] together cover the entire aligned
+    /// group `[groupStart, groupStart + groupSize)`.  If so, queues the group for consolidation.
+    private void checkGroupCoverage(long groupStart) {
+        final long groupEnd = groupStart + groupSize - 1;
+
+        // Only consider completed (non-in-flight) archives within this group's range.
+        final List<TempArchiveEntry> entries =
+                tempArchiveTracker.subMap(groupStart, true, groupEnd, true).values().stream()
+                        .filter(e -> e.uploadId() == null)
```
Nana-EC: is this required or defensive in nature?
I don't see a code path that puts a non-null `uploadId` entry into the tracker.
My understanding is in-flight entries live in `tempUploadFutures` right.
Maybe an additional comment would help or remove the filter if it's a no-op.

Nana-EC (dismissed):
LG. Some non-blocking suggestions that you can address in the follow up PR

jsync-swirlds (dismissed):
Remaining concerns can be addressed in follow-up issues.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Great catch! There was a bug in `drainOverflowStash`. It would keep blocks in the stash indefinitely once we crossed the `maxConcurrentTempArchives` limit.

As I had to fix a failing build due to a spotless plugin change, I fixed this (and other comments) too.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Good catch, added the missing `clear()` calls

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: You are right, it was overly defensive. I removed the filter

Nana-EC (approved):

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -383,10 +715,17 @@ public void stop() {
         if (recoveryFuture != null) {
             recoveryFuture.cancel(true);
         }
-        // cancel(true) interrupts the virtual thread at blockQueue.take(), so the task completes
-        // quickly without needing an explicit get() here.
+        tempUploadFutures.values().forEach(f -> f.cancel(true));
+        consolidationFutures.values().forEach(f -> f.cancel(true));
```
jsync-swirlds, line 719: Have we verified that cancel actually ends the thread in this case?  @ata-nas found that not all Futures implement cancel in a way that interrupts the running task.

jsync-swirlds (approved):

PR #3024, opened 2026-06-16
chore: follow-up items from PR #2738 (block range persistence in ASF)

### LINKED ISSUE #2875: chore: follow-up items from PR #2738 (block range persistence in ASF)
## Context

PR #2738 adds `addBlockRange` to `ApplicationStateFacility` and persists stored/available block ranges in `BlockNodeApp`. Several non-blocking review comments were deferred to a follow-up. This ticket captures those items.

---

## Technical Notes

- [ ] **Wire block ranges into the server status endpoint** — `addStoredBlockRange` and `addAvailableBlockRange` are now tracked in `BlockNodeApp` but the `serverDetailStatus` response still reads only from `historicalBlockFacility`. Expose the stored/available ranges via a getter on `ApplicationStateFacility` and wire them into the status response.
  - Raised by @berryware: https://github.com/hiero-ledger/hiero-block-node/pull/2738#discussion_r3202322256

- [ ] **Settle `availableBlocks` semantics — per-provider vs consolidated** — @ata-nas and @jsync-swirlds raised that availability is inherently per-provider and a consolidated `ConcurrentLongRangeSet` cannot accurately represent cases where one provider removes a block another still holds. Decide between: (A) keep consolidated with documented union semantics, (B) remove `availableBlocks` and delegate to `HistoricalBlockFacility`, or (C) per-provider tracking.
  - Raised by @ata-nas: https://github.com/hiero-ledger/hiero-block-node/pull/2738#discussion_r3207820223
  - jsync follow-up: https://github.com/hiero-ledger/hiero-block-node/pull/2738#discussion_r3314656676

- [ ] **Drop or standardise `spotless:off/on` in `ApplicationStateConfig`** — @Nana-EC asked to stop adding spotless suppression flips; @ivannov restored them. Settle the convention and apply it consistently. If the flips are needed, add a comment explaining why.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2738#discussion_r3212274142

- [ ] **Resolve `updateInitialDelay` value — 0 or 100 ms** — @Nana-EC left a code suggestion to use literal `0` for the initial delay. The current code uses `appStateConfig.updateInitialDelay()` which defaults to `100` ms. @berryware stated the initial delay "has been removed" but the config property still exists. Decide on the correct default and either apply the suggestion or close the thread with an explanation.
  - Nana-EC suggestion: https://github.com/hiero-ledger/hiero-block-node/pull/2738#discussion_r3270988694
  - berryware note: https://github.com/hiero-ledger/hiero-block-node/pull/2738#discussion_r3282173238

- [x] **Persist-on-change for all application state** — already captured in #2872.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2738#discussion_r3212294266

ivannov:
## Reviewer Notes

This PR resolves the four deferred review items from PR #2738 (block range persistence in `ApplicationStateFacility`).

### Available-blocks semantics settled

`addAvailableBlockRange` has been removed from `ApplicationStateFacility` entirely. Available blocks are now derived from `HistoricalBlockFacility.availableBlocks()` at query time rather than tracked as a parallel `ConcurrentLongRangeSet` inside `BlockNodeApp`. The `availableBlocks` field and its persistence to a separate `.bin` file are gone. `BlockRangesState` proto now only carries `storedBlocks`.
  
### updateInitialDelay resolved to 0

The config property default is now `0` (the `@Min(100)` constraint is removed). The initial delay was a workaround that was no longer needed; removing it eliminates the 100 ms startup stall.

### Spotless convention across config files

The convention was settled in favor of keeping the flips, and they are now consistent across all config files with a standardized comment.

## Related Issue(s)

Closes #2875

### DIFF
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -751,7 +751,12 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
                 .streamRanges()
                 .map(r -> new BlockRange(r.start(), r.end()))
                 .toList();
-        return new BlockRangesState(stored);
+        final List<BlockRange> available = historicalBlockFacility
+                .availableBlocks()
+                .streamRanges()
+                .map(r -> new BlockRange(r.start(), r.end()))
+                .toList();
+        return new BlockRangesState(stored, available);
     }
 
     /**
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -5,6 +5,7 @@ import static java.lang.System.Logger;
 import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.ERROR;
 import static java.lang.System.Logger.Level.INFO;
+import static java.lang.System.Logger.Level.TRACE;
 import static java.lang.System.Logger.Level.WARNING;
 import static org.hiero.block.common.constants.StringsConstants.APPLICATION_PROPERTIES;
 import static org.hiero.block.common.constants.StringsConstants.APPLICATION_TEST_PROPERTIES;
@@ -39,6 +40,7 @@ import java.security.spec.X509EncodedKeySpec;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HexFormat;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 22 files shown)


jsync-swirlds:
Did we discuss removing the available blocks element from Application State?
That seems, to me, to be very backwards. Relying on a specific plugin for blocks available seems completely the reverse of where we should be moving the system.

I would, instead, expect that we want to make the Application State the Authoritative source for available blocks and the _other_ plugin would be modified to update available blocks as needed via calls to Application State.

Edit: After discussion, we'll make this change now and design a more consolidated approach later.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -497,12 +494,6 @@ public void addStoredBlockRange(LongRange blockRange) {
         storedBlocks.add(blockRange);
     }
 
-    @Override
-    public void addAvailableBlockRange(LongRange blockRange) {
-        availableBlocks.add(blockRange);
-        addStoredBlockRange(blockRange);
-    }
-
```
jsync-swirlds, line 505: This is one of the key items, it's important to ensure that the stored block range _also_ includes available blocks; otherwise we'll have incomplete information at critical points and every plugin will have to manage both stored and available.

We don't need to keep this method, but we need to ensure that when we update context, we include a value for "stored" blocks that includes available blocks as well.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds, line 646: This needs to read from the historical blocks facility.

ivannov: Restored the available blocks to the block range state proto

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Just to confirm that as per our agreement this method should not be restored, right?

jsync-swirlds: I don't think we need this method; but we _do_ need to ensure that when we report stored blocks we merge in the _available_ blocks; so the range(s) reported include both stored and available.
Basically, it's critical that stored (as _reported_) be a superset of available, not orthogonal.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -848,8 +840,6 @@ private void loadApplicationState(final Configuration configuration) {
                         MAX_APP_STATE_MESSAGE_DEPTH,
                         MAX_APP_STATE_MESSAGE_SIZE_BYTES);
                 rangeSet.storedBlocks().forEach(r -> storedBlocks.add(new LongRange(r.rangeStart(), r.rangeEnd())));
-                rangeSet.availableBlocks()
-                        .forEach(r -> availableBlocks.add(new LongRange(r.rangeStart(), r.rangeEnd())));
```
jsync-swirlds, line 852: Do we need this filled in?

Nana-EC (dismissed):

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: To my understanding, the current source of truth for available blocks is the historical block facility. We don't keep track of available blocks in the application state facility.

As discussed during the meeting, it will be addressed in a task part of epic #3037

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: We don't keep track of available blocks here. We rely on the historical blocks facility

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: That's fine, but you keep missing the other part.

Stored blocks _must be_ a superset of Available blocks.
This isn't optional, and it _does not matter_ where the data is "owned".

When reporting Stored blocks, we _must_ include all available blocks, this is a hard requirement in order for backfill, publisher, status, and verification to work correctly.

I'm not saying to store that data here, I'm saying we must merge that data in dynamically.

I've commented below with the exact point where we need to update stored blocks.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: We still need to include that data when updating block node context, however.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds, line 640: This is the point where we need an update, if available blocks changed, then we need to create a merged stored value, and set that as the value for stored blocks on the builder.

jsync-swirlds (approved):

AlfredoG87 (approved):
LGTM

PR #3027, opened 2026-06-16
chore: Add schema for statusz endpoint networking data

jsync-swirlds:
## Reviewer Notes
* Added network-data proto file
    * This is for use in a local-only API used by Provisioner
    * The purpose is to deliver information needed for firewall configuration and traffic shaping.

This does not close any issue, it will be used in future PRs that do close issues.

### DIFF
--- a/protobuf-sources/src/main/proto/block-node/api/network-data.proto
+++ b/protobuf-sources/src/main/proto/block-node/api/network-data.proto
@@ -1,0 +1,94 @@
+// SPDX-License-Identifier: Apache-2.0
+syntax = "proto3";
+
+package org.hiero.block.api;
+
+option java_package = "org.hiero.block.api.protoc";
+// <<<pbj.java_package = "org.hiero.block.api">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/// A set of network endpoints representing communication channels required
+/// for a particular system to interact with other systems.
+///
+/// The `active_endpoints` field MUST NOT be empty.
+message NetworkData {
+    /// A list of network endpoints.
+    /// <p>
+    /// This field is REQUIRED.<br/>
+    /// This field MUST NOT be empty.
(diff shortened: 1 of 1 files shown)


protobuf-sources/src/main/proto/block-node/api/network-data.proto
```diff
@@ -0,0 +1,90 @@
...
+    ConnectionReference local = 1;
+
+    /// A remote endpoint that the producing system connects to or that
+    /// connects to the `local` endpoint.
+    ConnectionReference remote = 2;
+
+    /// A category for this connection.
+    ///
+    /// Categories are defined by the producing system.
+    string category = 3;
```
Nana-EC, line 48: This is a source of dynamic specification on a set of details that may not be highly variable.
Might it be worth considering explicit types?
Previously we explored `PUBLISHER`, `PARTNER`, `PUBLIC`, `RESTRICTED`, should this be the starting set that can be added to as needed?
Leaving it dynamic makes it easy for BN do change intent and potentially break expectation with consumers of the API

Nana-EC (commented):
Nice.
One consideration on the category string

protobuf-sources/src/main/proto/block-node/api/network-data.proto
jsync-swirlds: The use of a string is intentional.
Rather than constrain the value and need translations, the intent is to match the TC or nftables category string so the consumer of the API can just do a direct string match to their QoS categories.

Nana-EC: Gotcha. We should note that in the proto comments so it's clear there is some structure to it

jsync-swirlds: Updated.

rockysingh (approved):
lgtm

Nana-EC (approved):

PR #3038, opened 2026-06-17
feat: add blocks bulk-load command for historic plugin.

### LINKED ISSUE #2955: CLI: backfill historical WRBs by pushing to a Block Node

**Prereq:** none. (Full historical-WRB serving also needs T4+T5, but T1 is independently buildable/testable.)

## Story Form

As a node operator running the CLI conversion server,
I want a CLI command that reads the existing wrapped-block zips and pushes each block to a Block Node's ingestion endpoint,
So that the special-purpose BN is seeded with the full historical WRB set without restructuring storage.

## Technical Notes

- New/extended `blocks` subcommand (e.g. `blocks push`) reading the wrapped-block zip storage (10k blocks/zip) and publishing **one block at a time**, in order, over gRPC.
- Reuse the publish client used by `networkCapacity` client mode / simulator `PublishStreamGrpcClient`; server side is the standard `stream-publisher` ingestion.
- Resumable: start from the BN's latest stored block (query it) so a restart continues, not restarts.
- Options: source dir, target endpoint, start/resume block, rate limit. Emit progress + last-pushed block metrics/logs. Handle BN-unavailable with backoff; never drop a block.

## Design

rockysingh:
## Description

Implements CLI `blocks bulk-load` command to copy wrapped block zips from CLI output into Block Node historic storage for the BlockFileHistoricPlugin.

Closes #2955

## Features

- **Bulk copy**: Copies wrapped block zip files from source directory to Block Node historic storage directory
- **Resume support**: Tracks progress in `historic-plugin-bulk-load-state.json` and can resume from interruption
- **Filtering**: `--start-block` option to only copy blocks >= specified number
- **Dry-run mode**: Preview what would be copied without making changes
- **Progress tracking**: Shows copy progress every 10 files and saves state periodically

## Implementation

- `BulkLoadBlocksCommand.java`: Main command implementation with resume functionality
- `BulkLoadBlocksCommandTest.java`: Comprehensive test suite with 16 tests covering:
  - Command line interface (3 tests)
  - File copying (4 tests)
  - State file management (3 tests)
  - Start block filtering (2 tests)
  - Dry run mode (1 test)
  - Edge cases (3 tests)

## Usage

```bash
# Basic bulk load
blocks bulk-load --source-dir /path/to/cli/output --dest-dir /opt/hiero/block-node/data/historic

# Resume from previous run
blocks bulk-load --source-dir /path/to/cli/output --dest-dir /opt/hiero/block-node/data/historic --resume

# Copy only blocks >= 1000
blocks bulk-load --source-dir /path/to/cli/output --dest-dir /opt/hiero/block-node/data/historic --start-block 1000

# Dry run to preview
blocks bulk-load --source-dir /path/to/cli/output --dest-dir /opt/hiero/block-node/data/historic --dry-run
```

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
@@ -1,0 +1,418 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.tools.blocks;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import java.io.IOException;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicLong;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Help.Ansi;
+import picocli.CommandLine.Option;
+
+/**
+ * Bulk-load wrapped blocks from CLI output into a Block Node's historic storage.
+ *
+ * <p>This command copies wrapped block zip files from the CLI's output directory (as produced by
+ * the {@code wrap} command) directly into a Block Node's historic storage directory. The Block Node
+ * MUST be stopped before running this command.
+ *
+ * <p>On Block Node startup, the {@code BlockFileHistoricPlugin} will detect all copied blocks and
+ * make them available for serving.
+ *
+ * <h2>Storage Format Compatibility</h2>
+ * <p>Both the CLI's {@code wrap} command and the Block Node's {@code BlockFileHistoricPlugin} use
+ * the same storage format (as defined by {@code BlockWriter}), so zip files can be copied directly
+ * without any conversion or restructuring.
+ *
+ * <h2>Resumable</h2>
+ * <p>This command is resumable: if run multiple times, it will only copy files that don't already
+ * exist in the destination. Files are compared by path and size.
+ *
+ * <h2>Safety</h2>
+ * <p>The Block Node MUST be stopped before running this command to avoid concurrent writes to the
+ * same storage location.
+ *
+ * <h2>Example Usage</h2>
+ * <pre>
+ * # Stop the Block Node first!
+ * # systemctl stop hiero-block-node
+ *
+ * # Bulk-load wrapped blocks
+ * blocks bulk-load --source wrappedBlocks --dest /opt/hiero/block-node/data/historic
+ *
+ * # Start the Block Node
+ * # systemctl start hiero-block-node
+ * </pre>
+ *
+ * <h2>Directory Structure</h2>
+ * <p>The command copies the entire directory tree from the source to destination, preserving the
+ * nested directory structure required by the Block Node:
+ * <pre>
+ * source/
+ *   000/123/456/789/012/345/
+ *     60000s.zip
+ *     70000s.zip
+ *
+ * dest/
+ *   000/123/456/789/012/345/
+ *     60000s.zip  (copied)
+ *     70000s.zip  (copied)
+ * </pre>
+ */
+@Command(
+        name = "bulk-load",
+        description = "Bulk-load wrapped blocks from CLI output into Block Node historic storage (BN must be stopped)",
+        mixinStandardHelpOptions = true)
+public class BulkLoadBlocksCommand implements Callable<Integer> {
+
+    @Option(
+            names = {"-s", "--source"},
+            description = "Source directory containing wrapped block zip files (from CLI wrap command)",
+            required = true)
+    private Path sourceDir;
+
+    @Option(
+            names = {"-d", "--dest"},
+            description =
+                    "Destination directory (Block Node historic root path, e.g., /opt/hiero/block-node/data/historic)",
+            required = true)
+    private Path destDir;
+
+    @Option(
+            names = {"--dry-run"},
+            description = "Show what would be copied without actually copying")
+    private boolean dryRun = false;
+
+    @Option(
+            names = {"--start-block"},
+            description = "Start copying from this block number onwards (for resuming interrupted loads)")
+    private long startBlock = -1;
+
+    private static final String STATE_FILE_NAME = "historic-plugin-bulk-load-state.json";
+    private static final ObjectMapper JSON_MAPPER = new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);
+
+    /**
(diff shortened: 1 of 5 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
```diff
@@ -0,0 +1,633 @@
...
+            return;
+        }
+
+        try {
+            // Extract the first block number from the zip file name
+            String numStr = fileName.substring(0, fileName.length() - 5); // Remove "s.zip"
+            long zipFirstBlock = Long.parseLong(numStr);
+
+            // Each zip contains 10,000 blocks
+            long zipLastBlock = zipFirstBlock + 10_000 - 1;
```
Nana-EC: Use config to allow for customization
```suggestion
            long zipLastBlock = zipFirstBlock + BlockZipsUtilities.DEFAULT_BLOCKS_PER_ZIP - 1;
```

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
```diff
@@ -0,0 +1,633 @@
...
+        try {
+            // Extract the first block number from the zip file name
+            // e.g., "10000s.zip" -> 10000
+            String numStr = fileName.substring(0, fileName.length() - 5); // Remove "s.zip"
+            long zipFirstBlock = Long.parseLong(numStr);
+
+            // Determine the last block in this zip
+            // Assume powersOfTen=4 (10,000 blocks per zip) as that's the default
+            long blocksPerZip = 10_000;
+            long zipLastBlock = zipFirstBlock + blocksPerZip - 1;
```
Nana-EC: use config
```suggestion
            long zipLastBlock = zipFirstBlock + BlockZipsUtilities.DEFAULT_BLOCKS_PER_ZIP - 1;
```

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
```diff
@@ -0,0 +1,633 @@
...
+                    copiedBytes.set(state.totalBytesCopied);
+                    progress.advance(attrs.size());
+
+                    return FileVisitResult.CONTINUE;
+                }
+
+                @Override
+                public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
+                    // Skip staging, links, and zipwork directories if they exist in source
+                    String dirName = dir.getFileName().toString();
```
Nana-EC: Missing a NPE check like you for for scanWorkToDo
```suggestion
                    String dirName = dir.getFileName() == null ? "" : dir.getFileName().toString();
```

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
```diff
@@ -0,0 +1,633 @@
...
+                return FileVisitResult.CONTINUE;
+            }
+
+            @Override
+            public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
+                if (!file.toString().endsWith(".zip")) {
+                    return FileVisitResult.CONTINUE;
+                }
+                Path rel = source.relativize(file);
+                if (state.copiedFiles.contains(rel.toString())) {
```
Nana-EC, line 519: On a resume run, every file visit scans the entire list. This will be large for mainnet no?
Should we explore an improvement to speed things up?
Might be worth exploring if copiedFiles should be HashSet.

Nana-EC (dismissed):
LG, some suggested improvements and consideration that you can address in follow up if needed.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
rockysingh: Done in a6f0f226 — replaced the hardcoded `10_000` in `shouldSkipBasedOnBlockNumber` with `BlockZipsUtilities.DEFAULT_BLOCKS_PER_ZIP`, matching the fix applied to `updateBlockNumber` earlier. Also added the missing import (the earlier Apply Suggestion didn't add it, which was breaking the build).

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
rockysingh: Same fix as the line 420 thread — done in a6f0f226.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
rockysingh: Good catch on the O(n²) resume scan. Changed `copiedFiles` from `ArrayList<String>` to `HashSet<String>` in a6f0f226 so `.contains()` is O(1) instead of O(n). Jackson serializes `Set` as a JSON array, so on-disk state files remain backward-compatible (existing state files deserialize fine into a HashSet).

Nana-EC (dismissed):

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
```diff
@@ -0,0 +1,633 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.tools.blocks;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
```
jsync-swirlds: Instead of Jackson; can we use the PBJ JSON support for this?

jsync-swirlds (commented):
Might be good to stick with PBJ if we can for JSON processing.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/BulkLoadBlocksCommand.java
rockysingh: Yep will get that in.

rockysingh: Done in c92d6bab — added `internal/bulk_load_state.proto` defining `BulkLoadState` and migrated `BulkLoadBlocksCommand` to `BulkLoadState.JSON.parse / .write` (via PBJ's `ReadableStreamingData / WritableStreamingData`). Kept the mutable `LoadState` POJO in-memory so `copiedFiles` stays a `HashSet` for O(1) lookups during the walk; convert to/from the PBJ message at the I/O boundary. Also dropped Jackson from `mainModuleInfo` and migrated the tests to the same PBJ JSON path. PBJ emits camelCase keys so on-disk state files written by previous versions remain readable.

jsync-swirlds (approved):

Nana-EC (approved):

PR #3046, opened 2026-06-18
feat: Add statusz inbound and outbound network data to Health plugin.

### LINKED ISSUE #2992: Expose backfill target information on Health Plugin API
### Story Form

As a node operator
I want to be able to expose the backfill target information on a ReST health plugin API
So that I can have solo provisioner observe changes and update the targets I support

### Technical Notes

Add the ability for the health plugin to get backfill targets from application state and expose it in a new API.
A config already exists that allows a node operator to override this, do not break it as it's still  needed.
Use the existing data schema for the JSON data

jsync-swirlds:
## Reviewer Notes
* Added configuration for baseline network data for publisher, inbound partners, and outbound partners
   * The publisher data will be replaced with roster data in future work.
* Adjusted the Backfill plugin to fill in Backfill connections when loading backfill configuration.
   * This may also be modified in future work to use HIP-1137 RegisteredNode data.
* Modified the backfill plugin configuration data schema to include missing values needed to fill in NetworkData connections.
* Adjusted Application State Facility to manage network data
* Added endpoints to the Health plugin for inbound and outbound network data
* Added initial tests for all new functionality.

## Related Issue(s)
 Resolves #2992

### DIFF
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
@@ -22,6 +22,12 @@ import org.hiero.block.node.base.Loggable;
  *     The file is written automatically every {@code BLOCK_RANGE_PERSIST_INTERVAL} blocks and on
  *     shutdown, then loaded on startup. Block availability is derived from
  *     {@code HistoricalBlockFacility} at query time and is not persisted here.
+ * @param knownPublishersFilePath path to the JSON file (a serialized {@code NetworkData}) describing the
+ *     known inbound publishers, loaded on startup and exposed for the {@code /statusz/inbound} endpoint.
+ * @param inboundPartnersFilePath path to the JSON file (a serialized {@code NetworkData}) describing the
+ *     designated inbound partners, loaded on startup and exposed for the {@code /statusz/inbound} endpoint.
+ * @param outboundPartnersFilePath path to the JSON file (a serialized {@code NetworkData}) describing the
+ *     designated outbound partners, loaded on startup and exposed for the {@code /statusz/outbound} endpoint.
  * @param updateScanInterval The amount of milliseconds that the {@code ApplicationStateFacility} waits between
  *     checking to see if there are any {@code TssData} updates to process.
  * @param updateInitialDelay The amount of milliseconds before the first scheduled scan. Defaults to {@code 0}
@@ -33,6 +39,9 @@ public record ApplicationStateConfig(
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/tss-bootstrap-roster.json") Path tssBootstrapFilePath,
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/rsa-bootstrap-roster.json") Path rsaBootstrapFilePath,
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/block-ranges.json") Path blockRangesFilePath,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/known-publishers.json") Path knownPublishersFilePath,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/inbound-partners.json") Path inboundPartnersFilePath,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/outbound-partners.json") Path outboundPartnersFilePath,
         @Loggable @ConfigProperty(defaultValue = "500") @Min(100) long updateScanInterval,
         @Loggable @ConfigProperty(defaultValue = "0") @Min(0) int updateInitialDelay) {
         // spotless:on
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -55,6 +55,7 @@ import java.util.stream.Collectors;
 import org.hiero.block.api.BlockNodeVersions;
 import org.hiero.block.api.BlockNodeVersions.PluginVersion;
 import org.hiero.block.api.BlockRange;
+import org.hiero.block.api.NetworkData;
 import org.hiero.block.api.TssData;
 import org.hiero.block.internal.BlockRangesState;
 import org.hiero.block.node.app.config.AutomaticEnvironmentVariableConfigSource;
@@ -129,6 +130,18 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     /// Blocks reported as stored by plugins that do not serve them for retrieval
     final ConcurrentLongRangeSet storedBlocks = new ConcurrentLongRangeSet();
 
+    /// Known inbound publishers loaded from configuration on startup; exposed for /statusz/inbound.
+    private final AtomicReference<NetworkData> knownPublishers = new AtomicReference<>(NetworkData.DEFAULT);
+
+    /// Designated inbound partners loaded from configuration on startup; exposed for /statusz/inbound.
+    private final AtomicReference<NetworkData> inboundPartners = new AtomicReference<>(NetworkData.DEFAULT);
+
+    /// Designated outbound partners loaded from configuration on startup; exposed for /statusz/outbound.
+    private final AtomicReference<NetworkData> outboundPartners = new AtomicReference<>(NetworkData.DEFAULT);
+
+    /// Backfill source connections reported by the backfill plugin; exposed for both /statusz endpoints.
+    private final AtomicReference<NetworkData> backfillSources = new AtomicReference<>(NetworkData.DEFAULT);
+
     /// Block count at the time of the last scheduled persist; only read/written by the scanner thread
     private long lastPersistedBlockCount = 0;
 
@@ -150,7 +163,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         // ==== LOAD LOGGING CONFIG ====================================================================================
         final boolean externalLogging = System.getProperty("java.util.logging.config.file") != null;
         if (externalLogging) {
-            LOGGER.log(INFO, "External logging configuration found");
+            LOGGER.log(DEBUG, "External logging configuration found");
         } else {
             // load the logging configuration from the classpath and make it colorful
             try (var loggingConfigIn = BlockNodeApp.class.getClassLoader().getResourceAsStream("logging.properties")) {
@@ -163,7 +176,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
                 LOGGER.log(INFO, "Failed to load logging configuration", e);
             }
             CleanColorfulFormatter.makeLoggingColorful();
-            LOGGER.log(INFO, "Using default logging configuration");
+            LOGGER.log(DEBUG, "Using default logging configuration");
         }
         // tell helidon to use the same logging configuration
         System.setProperty("io.helidon.logging.config.disabled", "true");
@@ -177,7 +190,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
             final String[] moduleClassPathArray = moduleClassPath.split(":");
             for (String module : moduleClassPathArray) {
                 if (module.contains("hiero")) {
-                    LOGGER.log(INFO, "    " + module);
+                    LOGGER.log(INFO, "    {0}", module);
                 }
             }
         }
@@ -248,9 +261,8 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         final ServiceBuilderImpl serviceBuilder = new ServiceBuilderImpl(serverConfig.port());
         // ==== INITIALIZE PLUGINS =====================================================================================
         // Initialize all the facilities & plugins, adding routing for each plugin
-        LOGGER.log(INFO, "Initializing plugins:");
         for (BlockNodePlugin plugin : loadedPlugins) {
-            LOGGER.log(INFO, "    " + plugin.name());
+            LOGGER.log(INFO, "    {0}", plugin.name());
             plugin.init(blockNodeContext, serviceBuilder);
         }
         // ==== LOAD & CONFIGURE WEB SERVER ============================================================================
@@ -383,7 +395,6 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     /// Starts the block node server. This method initializes all the plugins, starts the web server,
     /// and starts the metrics.
     public void start() {
-        LOGGER.log(INFO, "Starting BlockNode Server...");
         webServer.start();
         LOGGER.log(
                 INFO,
@@ -431,7 +442,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         webServer.stop();
         // Stop all the facilities &  plugins
         for (BlockNodePlugin plugin : loadedPlugins) {
-            LOGGER.log(INFO, "Stopping plugin: {0}", plugin.name());
+            LOGGER.log(INFO, "    {0}", plugin.name());
             plugin.stop();
         }
         // Stop metrics
@@ -442,7 +453,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
             LOGGER.log(DEBUG, "Could not properly close metric registry.", e);
         }
         // finally exit
-        LOGGER.log(INFO, "Bye bye");
+        LOGGER.log(INFO, "System Exiting");
         if (shouldExitJvmOnShutdown) System.exit(0);
     }
 
@@ -458,10 +469,8 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     /// Start the loadedPlugins. Use a separate method to make starting plugins testable
     protected void startPlugins(List<BlockNodePlugin> plugins) {
         // Start all the facilities & plugins asynchronously
-        LOGGER.log(INFO, "Asynchronously Starting plugins:");
         // Asynchronously start the plugins
         plugins.parallelStream().forEach(plugin -> {
-            LOGGER.log(INFO, "    " + plugin.name());
             plugin.start();
         });
     }
@@ -483,6 +492,31 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         storedBlocks.add(blockRange);
     }
 
+    @Override
+    public NetworkData knownPublishers() {
+        return knownPublishers.get();
+    }
+
+    @Override
+    public NetworkData inboundPartners() {
+        return inboundPartners.get();
+    }
+
+    @Override
+    public NetworkData outboundPartners() {
+        return outboundPartners.get();
+    }
+
+    @Override
+    public NetworkData backfillSources() {
+        return backfillSources.get();
+    }
+
+    @Override
+    public void updateBackfillSources(NetworkData sources) {
+        backfillSources.set(sources != null ? sources : NetworkData.DEFAULT);
+    }
+
     /// Allow plugins to update the NodeAddressBook for this BlockNodeApp.
     /// The address book is staged in a last-write-wins reference; if
     /// `updateAddressBook` is called more than once before the next
@@ -498,7 +532,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
         try {
             validateAddressBook(nodeAddressBook, "runtime update");
         } catch (IllegalStateException e) {
-            LOGGER.log(WARNING, "Rejecting invalid address book update: {0}", e.getMessage());
+            LOGGER.log(INFO, "Rejecting invalid address book update: {0}", e);
             return false;
         }
         pendingAddressBook.set(nodeAddressBook);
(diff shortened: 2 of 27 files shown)


block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -30,6 +36,9 @@ public record ApplicationStateConfig(
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/tss-bootstrap-roster.json") Path tssBootstrapFilePath,
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/rsa-bootstrap-roster.json") Path rsaBootstrapFilePath,
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/block-ranges.json") Path blockRangesFilePath,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/known-publishers.json") Path knownPublishersFilePath,
```
Nana-EC, line 42: Initially I had to remember the diff between known vs designated and why they are separately needed for the single `statusz/inbound`.

On a related note, this PR should have a design doc to capture the goal and approach here for the new APIs and the underlying logic.
There's a lot of intent that's important to capture for easy reference by node operators.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillNetworkData.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param tls whether TLS is required (from backfill configuration)
+     * @return the equivalent NetworkConnection
+     */
+    static NetworkConnection toNetworkConnection(@NonNull final BlockNodeSourceConfig node, final boolean tls) {
+        final String scheme = node.scheme().isBlank() ? (tls ? "https" : "http") : node.scheme();
+        final IpProtocol protocol = node.protocol() == IpProtocol.UNIDENTIFIED ? IpProtocol.TCP : node.protocol();
+        return NetworkConnection.newBuilder()
+                .remote(new ConnectionReference(node.address(), Integer.toString(node.port())))
+                .local(new ConnectionReference("$", "*"))
+                .category("partner")
```
Nana-EC: I still think we should try our best to codify the options for category.
An example is cloud bucket storage types.
We don't define them but they encompass S3 compatible API options i.e. GPC, AWS, etc definitions are applicable and easy to reference.
"partner" here is unclear to me as the meaning.

block-node/health/src/main/java/org/hiero/block/node/health/InboundStatusHandler.java
```diff
@@ -0,0 +1,57 @@
...
+            @NonNull final ServerResponse response,
+            @NonNull final ApplicationStateFacility appState) {
+        this.request = request;
+        this.response = response;
+        this.appState = appState;
+    }
+
+    /** Builds the inbound {@link NetworkData} and sends it as an {@code application/json} response. */
+    void createAndSendResponse() {
+        final List<NetworkConnection> endpoints = new ArrayList<>();
```
Nana-EC, line 41: Q: Are we intending to support mid operation updates to publishers, partners and backfill sources?
If not why not cache this once after init so that responses to this call  are predictable and light weight

protobuf-sources/src/main/proto/internal/block_node_source.proto
```diff
@@ -39,6 +41,18 @@ message BlockNodeSourceConfig {
    * All tuning parameters are in this sub-message for cleaner organization
    */
   GrpcWebClientTuning grpc_webclient_tuning = 6;
+
+  /**
+   * RFC 3986 connection scheme for this source (e.g. "http" or "https").
+   * When empty, the scheme is derived from the backfill TLS setting.
+   */
+  string scheme = 7;
```
Nana-EC, line 50: We should make this an enum to be more restrictive of support. `scheme` could also be `ftp`, `mailto` etc

protobuf-sources/src/main/proto/internal/block_node_source.proto
```diff
@@ -39,6 +41,18 @@ message BlockNodeSourceConfig {
    * All tuning parameters are in this sub-message for cleaner organization
    */
   GrpcWebClientTuning grpc_webclient_tuning = 6;
+
+  /**
+   * RFC 3986 connection scheme for this source (e.g. "http" or "https").
+   * When empty, the scheme is derived from the backfill TLS setting.
```
Nana-EC: Q: why based on backfill?

Nana-EC (changes requested):
Solid start.
I'd like to see a design capturing the context here and I had a few questions to confirm some choices.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
jsync-swirlds: I'll include a design document in the next PR.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillNetworkData.java
jsync-swirlds: Would it be OK to detail "typical" categories in the design document?
I don't know that these values are permanent (and to be clear, I still need to make several of these values constants); we need to get the provisioner work a little further along before finalizing them.

block-node/health/src/main/java/org/hiero/block/node/health/InboundStatusHandler.java
jsync-swirlds: Publishers and backfill are expected to be dynamic and read from on-chain data when we start using HIP-1137 data.
At least some partners (particularly mirror nodes registered on chain) are also expected to be dynamic eventually.

protobuf-sources/src/main/proto/internal/block_node_source.proto
jsync-swirlds: We could create an enum, but I don't know that there is a lot of value in supporting all of the several hundred possible schemes; and any enum, because of protobuf requirements, requires _translation_; using a string here does not.

As a matter of API design, for things that are not typically enumerated (schemes are not) and that would require translation code if enumerated (which creates enum mismatch problems for all API users), I generally prefer to leave it a string and accept that if someone enters a bad value then that entry will not work correctly.

We aren't supporting anything, this is information for a firewall or traffic prioritization; which supports whatever it supports (often a very large set of schemes).

protobuf-sources/src/main/proto/internal/block_node_source.proto
jsync-swirlds: This internal file describes backfill sources.  Everything in this file refers to backfill.

It is a little short of detail, but if TLS is _not_ enabled we'll use `grpc` and if TLS is enabled we'll use `grpcs`.

block-node/health/src/main/java/org/hiero/block/node/health/InboundStatusHandler.java
Nana-EC: Ah good reminder. Forgot about 1137, I was thinking the static file updates.
Okay dynamic approach makes sense again.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -30,6 +36,9 @@ public record ApplicationStateConfig(
...
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/inbound-partners.json") Path inboundPartnersFilePath,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/outbound-partners.json") Path outboundPartnersFilePath,
```
ivannov, line 44: Should we document these in configuration.md?

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppNetworkDataLoadTest.java
```diff
@@ -0,0 +1,58 @@
...
+        assertTrue(BlockNodeApp.loadNetworkData(dir.resolve("does-not-exist.json"))
+                .activeEndpoints()
+                .isEmpty());
+    }
+
+    @Test
+    @DisplayName("A null path yields an empty NetworkData")
+    void nullPathReturnsEmpty() {
+        assertTrue(BlockNodeApp.loadNetworkData(null).activeEndpoints().isEmpty());
+    }
```
ivannov, line 57: Is it worth adding a test for when `loadNetworkData` throws parser or IO exception?

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
```diff
@@ -90,4 +101,29 @@ public final void handleReadyz(@NonNull final ServerRequest req, @NonNull final
...
+            if (path.endsWith(INBOUND_PATH)) {
+                new InboundStatusHandler(req, res, applicationStateFacility).createAndSendResponse();
+            } else if (path.endsWith(OUTBOUND_PATH)) {
+                new OutboundStatusHandler(req, res, applicationStateFacility).createAndSendResponse();
+            } else {
+                res.status(404).send("Unknown statusz subpath");
+                LOGGER.log(INFO, "Responded code 404 (Unknown statusz subpath) for {0}", path);
+            }
+        } catch (final RuntimeException e) {
+            LOGGER.log(WARNING, "Failed to respond to statusz check due to %s".formatted(e), e);
```
ivannov, line 117: Shouldn't we set status to the response object here (e.g. 500)? What will happen with the client in case of runtime exception?

block-node/health/src/main/java/org/hiero/block/node/health/InboundStatusHandler.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Builds and sends the {@code /statusz/inbound} response: the merge of the known publishers, the
+ * inbound designated partners, and the backfill sources, serialized as {@link NetworkData} JSON.
+ *
+ * <p>This is an independent, {@code final} class with no shared base or interface (to avoid
+ * virtual-call overhead). Its single {@link #createAndSendResponse()} entry point is invoked
+ * synchronously on the (virtual) request thread by {@link HealthServicePlugin}.
+ */
+final class InboundStatusHandler {
+    private final Logger LOGGER = System.getLogger(getClass().getName());
```
ivannov, line 24: Maybe this should be static, to avoid creating a logger object on every request. Same applies to the outbound handler

protobuf-sources/src/main/proto/internal/block_node_source.proto
```diff
@@ -39,6 +41,18 @@ message BlockNodeSourceConfig {
    * All tuning parameters are in this sub-message for cleaner organization
    */
   GrpcWebClientTuning grpc_webclient_tuning = 6;
+
+  /**
+   * RFC 3986 connection scheme for this source (e.g. "grpc" or "grpcs").
+   * When empty, the scheme is derived from the backfill TLS setting, using
+   * `grpc` if TLS is not enabled and `grpcs` if TLS is enabled.
```
ivannov, line 48: This documentation implies that the default scheme is grpc(s). However the actual implementation defaults to to http(s).

block-node/health/src/test/java/org/hiero/block/node/health/HealthServiceTest.java
```diff
@@ -152,4 +160,82 @@ public void testRouting() {
...
+        healthServicePlugin.init(context, serviceBuilder);
+
+        TestServerResponse response = new TestServerResponse();
+        healthServicePlugin.handleStatusz(new TestServerRequest("/statusz/inbound"), response);
+
+        assertEquals(200, response.sentStatus());
+        assertEquals("application/json", response.contentType());
+        NetworkData sent =
+                NetworkData.JSON.parse(Bytes.wrap(((String) response.sentEntity()).getBytes(StandardCharsets.UTF_8)));
+        assertTrue(!sent.activeEndpoints().isEmpty());
```
ivannov: NIT: isn't `assertFalse(sent.activeEndpoints().isEmpty())` better?

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
```diff
@@ -90,4 +101,29 @@ public final void handleReadyz(@NonNull final ServerRequest req, @NonNull final
...
+        try {
+            final String path = req.path().path();
+            if (path.endsWith(INBOUND_PATH)) {
+                new InboundStatusHandler(req, res, applicationStateFacility).createAndSendResponse();
+            } else if (path.endsWith(OUTBOUND_PATH)) {
+                new OutboundStatusHandler(req, res, applicationStateFacility).createAndSendResponse();
+            } else {
+                res.status(404).send("Unknown statusz subpath");
+                LOGGER.log(INFO, "Responded code 404 (Unknown statusz subpath) for {0}", path);
+            }
```
ivannov, line 115: Is this code reachable? Won't any path different than inbound and outbound be filtered out already by Helidon?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -862,6 +900,44 @@ private void loadApplicationState(final Configuration configuration) {
...
+                    MAX_APP_STATE_MESSAGE_DEPTH,
+                    MAX_APP_STATE_MESSAGE_SIZE_BYTES);
+            LOGGER.log(
+                    INFO,
+                    "Loaded network data from file: {0} ({1} endpoints)",
+                    path,
+                    data.activeEndpoints().size());
+            return data;
+        } catch (ParseException | IOException e) {
+            LOGGER.log(WARNING, "Failed to read network data file: " + path, e);
```
ivannov: NIT: in other occasions you used the string interpolation instead of concatenation

ivannov (commented):
Some small questions / nits.

So far we've avoided to add accessors to the application state facility, we only had mutators. Now this is not the case anymore

protobuf-sources/src/main/proto/internal/block_node_source.proto
Nana-EC: Applying the restrictive nature is exactly the reason I wanted an enum. If there are hundred of options I wanted us to list only the 1-3 that we did support. Anything else listed would immediately fail validation plus the docs for node operators would clearly cover what is supported.
Also i don't see the translation code as a high barrier to be a blocker.

What i do agree with that you've highlighted is that the support is really by provisioner or whichever firewall implementing tool is in use. With that in mind this is more a pass through so we can say the validation is done by the tool so we should make sure provisioner has clear documentation on what it supports and doens't.
For us let's be clear in our docs that this is pass through and node operators should confirm provisioner or their tool of choice supports the setting they provide as Block Node software does not confirm it.

protobuf-sources/src/main/proto/internal/block_node_source.proto
Nana-EC: You're right, the name and file aren't explicit on it but indeed this is used by Backfill primarily.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
jsync-swirlds: Yes, I overlooked that.  Will add.

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppNetworkDataLoadTest.java
jsync-swirlds: I don't _think_ it's worth doing so; but I'm not opposed to it.

block-node/health/src/main/java/org/hiero/block/node/health/InboundStatusHandler.java
jsync-swirlds: Loggers are already singletons.  This doesn't create new loggers except for the first time it is called.

protobuf-sources/src/main/proto/internal/block_node_source.proto
jsync-swirlds: Good catch, will fix the default.

block-node/health/src/test/java/org/hiero/block/node/health/HealthServiceTest.java
jsync-swirlds: Very probably.
Will fix.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
jsync-swirlds: Helidon isn't filtering the last part of the path (at least, not unless I made a mistake).

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: This is an exception, so we cannot use _log_ interpolation (due to unfortunate choices in the Logger API).  I will change the concatenation to use `String.formatted`, however.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
jsync-swirlds: Good catch.

protobuf-sources/src/main/proto/internal/block_node_source.proto
jsync-swirlds: fixed.

block-node/health/src/test/java/org/hiero/block/node/health/HealthServiceTest.java
jsync-swirlds: Fixed.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
jsync-swirlds: (I did have an issue in the routing setup) Fixed.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: Done.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
jsync-swirlds: Fixed.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
jsync-swirlds: Added.

protobuf-sources/src/main/proto/internal/block_node_source.proto
jsync-swirlds: I believe the docs are clear on this point now.

jsync-swirlds:
Note, Spotless failure is due to a bug in Spotless setup for Hiero plugin.
It is not resolvable and must be ignored for now.

Nana-EC (dismissed):
Unblocking

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppNetworkDataLoadTest.java
ivannov: I don't mind to leave it like that

ivannov (dismissed):

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
```diff
@@ -11,51 +11,58 @@
...
         healthFacility = context.serverHealth();
+        applicationStateFacility = context.applicationStateFacility();
         // A null port (the default) shares server.port
         final Integer port =
                 context.configuration().getConfigData(HealthConfig.class).port();
         serviceBuilder.registerHttpService(HEALTHZ_PATH, port, httpRules -> httpRules
                 .get(LIVEZ_PATH, this::handleLivez)
                 .get(READYZ_PATH, this::handleReadyz));
+        serviceBuilder.registerHttpService(
+                STATUSZ_PATH, port, httpRules -> httpRules.get(STATUSZ_PATH, this::handleStatusz));
```
berryware: Won't this look for a pattern of /statusz/statusz? Should the inbound and outbound handlers be added here?

jsync-swirlds: This should just be for `/statusz`  We don't want to have additional routes for subpaths because it gets confused in Helidon and doesn't feed information the way we expect.
It's been difficult to directly test this, but I'll verify and adjust as needed.

Nana-EC: @berryware is right. registerHttpService takes the base path and the httpRules.get provides the subpath, which should be inbound and outbound
```suggestion
                STATUSZ_PATH, port, httpRules -> httpRules
                    .get(INBOUND_PATH, this::handleStatusz)
                    .get(OUTBOUND_PATH, this::handleStatusz));
```

jsync-swirlds: I'll fix, we _do not_ want the subpaths specified, as it makes the methods not work; so I'll adjust appropriately.

I removed those subpaths because it broke the ability to manage handlers properly.

berryware (dismissed):

AlfredoG87 (dismissed):
LGTM

Nana-EC (approved):
Approving assuming E2E rerun is successful

berryware (approved):
Looks Good

PR #3055, opened 2026-06-18
docs: add subscribeBlockStream protocol reference

### LINKED ISSUE #2735: docs: create subscribe-block-stream.md - subscribe protocol specification
## User Story

BN-A6 - As a Mirror Node operator or developer, I want to understand how the Block Node subscribe protocol works so I can implement a correct and reliable subscription client.

## Goal

Provide a protocol-level reference for the subscribe service, equivalent to the existing publish protocol spec, so Mirror Node developers can implement subscription correctly before the June 11 testnet transition.

## Contents should cover

- gRPC service definition and proto message types for subscribe requests and responses
- How to initiate a subscription (start block number, request format)
- Streaming behaviour - how block items are batched and delivered
- The `SUBSCRIBER_MAXIMUM_FUTURE_REQUEST` constraint and what happens when a subscriber requests too far ahead
- Error codes and what each means
- Reconnection behaviour - what the Block Node does when a subscriber disconnects
- How a subscriber handles gaps in received blocks

The existing publish protocol spec at `docs/design/communication-protocol/publish-block-stream.md` is the format reference.

deshmukhpranali:
## Reviewer Notes

- Adds subscribe-block-stream.md, a protocol-level reference for the subscribeBlockStream gRPC API equivalent to the existing publish-block-stream.md. 
- Also updates docs/design/communication-protocol/README.md to include the new document in the contents table.

## Related Issue(s)
Fixes  #2735

### DIFF
--- a/docs/design/communication-protocol/subscribe-block-stream.md
+++ b/docs/design/communication-protocol/subscribe-block-stream.md
@@ -1,0 +1,226 @@
+# subscribeBlockStream Protocol
+
+The `subscribeBlockStream` RPC is a server-streaming call exposed by `BlockStreamSubscribeService`. A client sends one `SubscribeStreamRequest` and the Block Node responds with a sequence of `SubscribeStreamResponse` messages until the requested range is delivered or an error occurs. This document describes the wire protocol a subscriber — typically a Mirror Node — must implement to consume that stream correctly.
+
+For how to configure a Mirror Node to connect to a Block Node in production, see [Connecting a Mirror Node to a Block Node](../../block-node/operations/connecting-a-mirror-node-to-a-block-node.md).
+
+![Block Node network architecture showing the Block Node between Consensus Nodes and downstream consumers such as Mirror Nodes](../../assets/block-node-network-architecture.svg)
+
+## gRPC service definition
+
+The subscribe service is defined in `block_stream_subscribe_service.proto` and exposes a single RPC:
+
+```
+rpc subscribeBlockStream(SubscribeStreamRequest)
+    returns (stream SubscribeStreamResponse);
+```
+
+| Message                   | Fields                                                                          |
+|:--------------------------|:--------------------------------------------------------------------------------|
+| `SubscribeStreamRequest`  | `start_block_number` (uint64, field 1), `end_block_number` (uint64, field 2)    |
+| `SubscribeStreamResponse` | oneof: `status` (Code), `block_items` (BlockItemSet), `end_of_block` (BlockEnd) |
+| `BlockItemSet`            | repeated `BlockItem` — defined in `shared_message_types.proto`                  |
+| `BlockEnd`                | `block_number` (uint64) — defined in `shared_message_types.proto`               |
+
+## Pre-flight: query available blocks
+
+Before constructing a request, call the `serverStatus` RPC defined in `node_service.proto`. The `ServerStatusResponse` returns two fields:
+
+- `first_available_block` — the lowest block number currently in persistent storage.
+- `last_available_block` — the highest block number currently available.
+
+A request with a `start_block_number` below `first_available_block` is rejected with `INVALID_START_BLOCK_NUMBER`. A request whose `start_block_number` exceeds the future-block limit is rejected with `NOT_AVAILABLE` (see [Future-block constraint](#future-block-constraint)).
+
+## Request format
+
+Both request fields are `uint64`. The value `18446744073709551615` (`0xFFFFFFFFFFFFFFFF`, `uint64` maximum) is a sentinel value with two distinct meanings depending on which field it appears in:
+
+- In `start_block_number`: begin from `first_available_block` as reported by `serverStatus`.
+- In `end_block_number`: stream indefinitely; no terminal `SUCCESS` is sent until the stream is closed by an error or a Block Node shutdown.
+
+The four supported streaming modes are:
+
+| Mode                           | `start_block_number` | `end_block_number` | Behavior                                               |
+|:-------------------------------|:--------------------:|:------------------:|:-------------------------------------------------------|
+| Live-only                      |     `uint64_max`     |    `uint64_max`    | Stream from the next complete live block, indefinitely |
+| First-available to fixed end   |     `uint64_max`     |         M          | Stream from `first_available_block` through block M    |
+| From fixed start, indefinitely |          N           |    `uint64_max`    | Stream from block N forward, indefinitely              |
+| Bounded range                  |          N           |      M (≥ N)       | Stream blocks N through M inclusive                    |
+
+The request is rejected with `INVALID_END_BLOCK_NUMBER` if a specific (non-sentinel) `end_block_number` is less than `start_block_number`.
+
+## Response stream structure
+
+Each `SubscribeStreamResponse` carries exactly one of three variants:
+
+**`block_items`** — a `BlockItemSet` containing one or more `BlockItem`s from the same block. A large block is split across multiple consecutive `block_items` messages. The following ordering invariants apply across those messages:
+
+- A `BlockItemSet` SHALL NOT span more than one block.
+- The `BlockHeader` appears exactly once per block, in the first `block_items` message for that block, and MUST be the first item in that set.
+- The `BlockProof` appears exactly once per block, in the last `block_items` message for that block, and MUST be the last item in that set.
+- All `block_items` messages between the header and proof carry intermediate block items with no header or proof.
+
+When reassembling a block from the stream, a new block begins on each `BlockHeader` and is complete when the corresponding `BlockProof` is received and confirmed by the subsequent `end_of_block` message.
+
+**`end_of_block`** — a `BlockEnd` message carrying the `block_number` of the completed block. The Block Node sends exactly one `end_of_block` per block, emitted immediately after all `block_items` for that block. The message immediately following a `BlockEnd` is either a new `block_items` starting with a `BlockHeader`, or the terminal `status`.
+
+**`status`** — a `Code` value that terminates the stream. This is always the last message. Subscribers MUST NOT expect any further messages after receiving `status`.
+
+For a bounded request (blocks N through M), the message sequence is:
+
+```
+block_items...  end_of_block(N)
+block_items...  end_of_block(N+1)
+...
+block_items...  end_of_block(M)
+status(SUCCESS)
+```
+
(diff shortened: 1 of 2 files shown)


docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Response stream structure
+
+Each `SubscribeStreamResponse` carries exactly one of three variants:
+
+**`block_items`** — a `BlockItemSet` containing one or more `BlockItem`s from the same block. A large block is split across multiple consecutive `block_items` messages. The following ordering invariants apply across those messages:
+
+- A `BlockItemSet` SHALL NOT span more than one block.
+- The `BlockHeader` appears exactly once per block, in the first `block_items` message for that block, and MUST be the first item in that set.
+- The `BlockProof` appears exactly once per block, in the last `block_items` message for that block, and MUST be the last item in that set.
```
jsync-swirlds: This is no longer true.
Any given block may contain multiple block proof items, and those items will always be _after_ the BlockFooter.
There is always exactly one BlockFooter, and it is the last non-proof block item.

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+- All `block_items` messages between the header and proof carry intermediate block items with no header or proof.
```
jsync-swirlds: Better to say "between header and footer", as there may be multiple proof items.

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+
+When reassembling a block from the stream, a new block begins on each `BlockHeader` and is complete when the corresponding `BlockProof` is received and confirmed by the subsequent `end_of_block` message.
```
jsync-swirlds:
```suggestion
When reassembling a block from the stream, a new block begins on each `BlockHeader`, is complete when the corresponding `BlockFooter` is received, is followed by one or more `BlockProof` items, and terminated by a single `end_of_block` message.
```

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+
+```
+block_items...  end_of_block(N)
+block_items...  end_of_block(N+1)
+...
+block_items...  end_of_block(M)
+status(SUCCESS)
+```
+
+For an indefinite request (`end_block_number` = `uint64_max`) the `status` message is sent only when the stream is closed by an error or a Block Node shutdown.
```
jsync-swirlds: We include the possibility for the Block Node to limit connection duration in the subscribe protocol as well. This is not _currently_ implemented but is supported and permitted. A request that has reached maximum stream limits will end with `status(SUCCESS)`.
```suggestion
For an indefinite request (`end_block_number` = `uint64_max`) the `status` message is sent only when the stream is closed by an error, the Block Node shuts down, or when the stream reaches maximum connection duration (typically around 600,000 blocks, if enabled).
```

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+    BlockNode-->>Subscriber: status(SUCCESS)
+```
+
+## Streaming modes
+
+### Historical streaming
+
+In bounded-range mode (`start_block_number` = N, `end_block_number` = M where M ≥ N) the Block Node reads the requested blocks from persistent storage. The stream ends with `status(SUCCESS)` once block M has been fully delivered.
+
+If a block in the requested range is not found in storage, the Block Node closes the stream with `NOT_AVAILABLE`.
```
jsync-swirlds: This is not correct.  If the range extends beyond the latest available block, the block node will continue to deliver blocks from the _live stream_ until the end block number is reached (see [comment here](https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java#L416)).

If the _first_ requested block is in the future, then the request is still accepted as long as the value is not too far in the future, limited by the `maximumFutureRequest` configuration which defaults to 4,000 blocks.

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+
+In live-only mode (both fields = `uint64_max`) the Block Node attaches the subscriber to the live block queue and begins delivery from the next complete block it receives. No historical blocks are delivered. If the Block Node is mid-block at connection time, delivery begins after that block completes.
+
+### Hybrid streaming
+
+In from-start-indefinite mode (`start_block_number` = N, `end_block_number` = `uint64_max`) the Block Node first delivers historical blocks from N up to the latest persisted block, then transitions seamlessly to the live queue. The subscriber receives a single continuous stream with no gap at the transition.
+
+This is the most common pattern for Mirror Nodes, which require a contiguous block history from a known checkpoint through the current tip and onward.
+
+> **Note:** The Block Node maintains a live-block queue per session. If a subscriber processes historical blocks more slowly than the Block Node receives new live blocks, older live-queue entries may be dropped to prevent the queue from filling. Dropped entries are not lost: the Block Node re-reads those blocks from persistent storage when the historical phase catches up to them.
```
jsync-swirlds, line 126: We should add another note that blocks streamed from the live stream _are not verified, may contain errors, may be repeated, and may be incomplete_. The receiver is responsible for handling these conditions and is encouraged to verify each block using all attached `BlockProof`s.

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+
+| Config key                        |        Default |     Minimum |
+|:----------------------------------|---------------:|------------:|
+| `subscriber.maximumFutureRequest` | `4,000` blocks | `10` blocks |
+
+A request for a start block within the permitted future window but not yet produced will be held — the Block Node waits for that block to arrive on the live stream before beginning delivery.
+
+> **Note:** `NOT_AVAILABLE` (not `INVALID_START_BLOCK_NUMBER`) is returned when `start_block_number` is too far ahead of the current tip. `INVALID_START_BLOCK_NUMBER` is returned only when `start_block_number` is below `first_available_block` or is otherwise structurally invalid.
+
+## Error codes
```
jsync-swirlds: Not all are errors:

```suggestion
## Result Codes
```

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+| `INVALID_END_BLOCK_NUMBER`   |   5   | `end_block_number` is less than `start_block_number`                                                                             | Correct `end_block_number`; retry                                                     |
+| `NOT_AVAILABLE`              |   6   | `start_block_number` exceeds the future-block limit, a required block is absent from storage, or the Block Node is shutting down | Call `serverStatus`; adjust the block range; retry with exponential back-off          |
+
+When the Block Node shuts down gracefully it closes each open session with `status(SUCCESS)`. New connection attempts received during shutdown are immediately closed with `NOT_AVAILABLE`.
+
+Because `SUCCESS` does not imply all requested blocks were delivered (orderly shutdown is also reported as `SUCCESS`), subscribers MUST NOT assume delivery is complete on receiving `SUCCESS` for an indefinite or partially-delivered stream.
```
jsync-swirlds:
```suggestion
Because `SUCCESS` does not imply all requested blocks were delivered (orderly shutdown and connection limit are also reported as `SUCCESS`), subscribers MUST NOT assume delivery is complete on receiving `SUCCESS` for an indefinite or partially-delivered stream.
```

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+        Note over Subscriber: Reconnect from block K+1; call serverStatus on another Block Node if needed
+    end
+```
+
+## Reconnection and gap handling
+
+If the stream closes for any reason, the subscriber SHOULD:
+
+1. Record the `block_number` from the last `end_of_block` message received. Use `block_number + 1` as `start_block_number` for the next request. If no `end_of_block` was received before the stream closed, reconnect using the original `start_block_number`.
+2. Call `serverStatus` to confirm the new start block is within the available range on the target Block Node.
```
jsync-swirlds:
```suggestion
2. Call `serverStatus` to confirm the new start block is within the available range on the target Block Node or not overly far in the future.
```

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+3. Issue a new `subscribeBlockStream` request from that start block, applying exponential back-off if consecutive attempts fail.
+4. Consider falling back to a different Block Node if errors on the primary persist.
```
jsync-swirlds: I would recommend swapping these.

```suggestion
4. Switch to a different Block Node and issue a new `subscribeBlockStream` request from that start block.
3. Apply exponential back-off if consecutive attempts fail across multiple Block Nodes.
```

docs/design/communication-protocol/subscribe-block-stream.md
```diff
@@ -0,0 +1,226 @@
...
+
+The Block Node streams blocks in strict ascending order with no gaps within a session. If a block in the requested range cannot be provided mid-stream (for example because it was pruned from storage after the request was accepted), the Block Node closes the stream with `NOT_AVAILABLE`. On receiving `NOT_AVAILABLE` mid-stream, call `serverStatus` and reconnect from the earliest block that covers the needed range, potentially from a different Block Node.
```
jsync-swirlds: Slightly incorrect.

```suggestion
The Block Node makes a reasonable effort to stream blocks in strict ascending order with no gaps within a session, but some exceptions may occur, particularly for "live" blocks. If a block in the requested range cannot be provided mid-stream (for example because it was pruned from storage after the request was accepted), the Block Node closes the stream with `NOT_AVAILABLE`. On receiving `NOT_AVAILABLE` mid-stream, call `serverStatus` and reconnect from the earliest block that covers the needed range, potentially from a different Block Node.
```

jsync-swirlds (commented):
A few improvements and corrections.

jsync-swirlds (approved):

Nana-EC (approved):

Issue #3007, opened 2026-06-12
Verification Rework Follow Up


Nana-EC:
## Story Form

As a maintainer
I want all non-blocking review feedback from #2869 to be tracked
So that it is not lost after the PR merges

## Technical Notes

Follow-up items from unresolved review comments on #2869 (feat: Verification Plugin Rework).

### Correctness / Bugs

- [x] **`StateProofVerifier.java` line 109 — reconstructed hash logic is wrong** (`jsync-swirlds`): The block being proved is a leaf input to the hash — its hash cannot be reconstructed from sibling paths. The signed root hash belongs to a _different_ block; the code uses sibling hashes to compute a root hash and then checks the signature on _that_ hash. The `rootHash.equals(reconstructed)` check can never pass for a real state proof. jsync: "I think this entire set of checks related to a 'reconstructed' hash is just wrong, it will never pass for a real state proof." ([link](https://github.com/hiero-ledger/hiero-block-node/pull/2869#discussion_r3384897318))

- [x] **`StateProofVerifier.java` line 69 — siblings count assumes remainder is multiple of 4** (`jsync-swirlds`): The check assumes only one possible sibling path shape and ignores the tree path. jsync: "Remainder can have almost any count; it is a _bug_ that we assumed only one possible path option and ignored the possibility to use the _tree_ instead of the chain path." Suggested fix: remove the count check entirely and let the signature mismatch surface the error. ([link](https://github.com/hiero-ledger/hiero-block-node/pull/2869#discussion_r3384886062))

- [x] **`SessionResultHandler.java` line 194 — `lastVerifiedBlock` increment may skip blocks** (`Nana-EC`, `jsync-swirlds`): Question raised whether `compareAndSet(lastVerified, verificationResult.blockNumber())` is more correct than always incrementing by 1. Author acknowledged the TODO and noted the intent is to increment strictly one block at a time, but left a `todo` comment indicating this needs revisiting. ([link](https://github.com/hiero-ledger/hiero-block-node/pull/2869#discussion_r3337858748))

### Design Concerns

[...]

PR #3058, opened 2026-06-19
refactor(verification): dump block + proof to disk on BAD_BLOCK_PROOF

### LINKED ISSUE #2745: Improvement(verification): dump block + proof to disk on BAD_BLOCK_PROOF
## What we want

When a block fails verification, the BN should:

1. Log the failure category at WARNING (`MISSING_FOOTER`, `MALFORMED_PROOF_STRUCTURE`, `STATE_PROOF_INVALID`, `SIGNATURE_MISMATCH`, …).
2. Dump the offending `BlockUnparsed` bytes to a configurable directory, with a sibling `.meta.json` (correlation id, handler id, failure category, BN identity, HAPI version).
3. Auto-purge dumps older than configurable retention (default ~7 days).
4. Rate-limit per `(blockNumber, failureCategory)` so a retried bad block doesn't flood disk.
5. Off by default in production, enabled in non-prod via config.

## Why

In the 2026-05-07 incident the BN logged only `INFO "Verification failed for block=N"`. We had no failure category and no captured bytes — couldn't confirm the state-proof root cause (issue 2) until WARNING-tier logs were separately retrieved. Dumped bytes would also let us byte-diff failing vs. successful copies of the same block across BNs.

## Affected code

- `VerificationServicePlugin.java:298` — failure log site
- `ExtendedMerkleTreeSession` failure points: 162 (null footer), 197 (multi-proof), 233 (sig mismatch), 320+ (state proof)

ivannov:
### Failure type granularity

The catch-all `BAD_BLOCK_PROOF` failure type is replaced with five specific types in both `VerificationNotification.FailureType` and `SessionFailureType`:  `MISSING_FOOTER`, `MALFORMED_PROOF_STRUCTURE`, `STATE_PROOF_INVALID`, `SIGNATURE_MISMATCH`, and `INVALID_BLOCK_HEADER`. Each failure site in `ExtendedMerkleTreeSession`, `RSAProofVerifier`, `StateProofVerifier`, `TSSVerifier`, and `BlockHasher` now emits the appropriate type. The log call at the verification failure site is promoted from `INFO` to `WARNING` and now includes the failure type.

### Bad-block dumper

`BadBlockDumper` writes two files per unique `(blockNumber, failureType)` pair: a protobuf-encoded `.blk` file containing the raw `BlockUnparsed` bytes, and a `.meta.json` sidecar with correlation ID, failure category, BN identity (hostname), and HAPI version. A `ConcurrentHashMap` key set acts as a per-process rate limiter so a retried bad block cannot flood disk. A daily `ScheduledExecutorService` task auto-purges files older than `verification.dumpRetentionDays` (default 7). The feature is disabled by default (`verification.dumpEnabled=false`) and must be explicitly turned on in non-prod environments.

## Related Issue(s)

Closes #2745


block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/TestVerificationPlugin.java
```diff
@@ -29,7 +29,7 @@ public class TestVerificationPlugin implements BlockNodePlugin, BlockNotificatio
     private BlockNodeContext context;
     private VerificationSession currentSession;
     private volatile BlockSource blockSource = BlockSource.PUBLISHER;
-    private volatile FailureType failureType = FailureType.BAD_BLOCK_PROOF;
+    private volatile FailureType failureType = FailureType.UNKNOWN_ERROR;
```
ata-nas: We should not change this. By default, this test fixture fails with bad proof.

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiterTest.java
```diff
@@ -247,7 +247,7 @@ void shouldReleaseLatchOnVerificationFailure() {
             // when - verification failed
             subject.handleVerification(new VerificationNotification(
                     false,
-                    FailureInfo.standard(FailureType.BAD_BLOCK_PROOF),
+                    FailureInfo.standard(FailureType.UNKNOWN_ERROR),
```
ata-nas: Why this change? I do not think we should change tests to unknown error. If they fail, maybe we are breaking something? Please consider this for all tests that have this change.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -117,6 +117,10 @@ public HashingResult get() {
                             switch (kind) {
                                 case BLOCK_HEADER -> {
                                     this.blockHeader = BlockHeader.PROTOBUF.parse(item.blockHeader());
+                                    if (this.blockHeader.number() != this.blockNumber) {
+                                        throw new VerificationSessionFailedException(
+                                                blockNumber, SessionFailureType.INVALID_BLOCK_HEADER, blockSource);
+                                    }
```
ata-nas: This is probably a bad check to add, not 100% sure. Proof verification will fail if the block is in any way invalid. Not sure about the failure type, it seems too specific to me.

EDIT: in fact, a [test below](https://github.com/hiero-ledger/hiero-block-node/pull/3058/changes#r3452682322) kind of confirmed my suspicion. I think we might want to reconsider this.

@jsync-swirlds any opinions about this check?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionFailureType.java
```diff
@@ -39,6 +47,11 @@ public FailureType asFailureType() {
             case UNSUPPORTED_HAPI_VERSION -> FailureType.UNSUPPORTED_HAPI_VERSION;
             case CANCELLED -> FailureType.CANCELLED;
             case UNKNOWN_ERROR -> FailureType.UNKNOWN_ERROR;
+            case MISSING_FOOTER -> FailureType.MISSING_FOOTER;
+            case MALFORMED_PROOF_STRUCTURE -> FailureType.MALFORMED_PROOF_STRUCTURE;
+            case STATE_PROOF_INVALID -> FailureType.STATE_PROOF_INVALID;
+            case SIGNATURE_MISMATCH -> FailureType.SIGNATURE_MISMATCH;
+            case INVALID_BLOCK_HEADER -> FailureType.INVALID_BLOCK_HEADER;
```
ata-nas: Not fond of this change. Most of these are already covered. For instance missing footer is covered by missing mandatory item. Signature mismatch, state proof invalid and malformed proof structure are all bad block proof. Not sure why we are removing bad block proof.

Being overly specific here, I believe leads to fragility and much worse handling by receivers of failed block notifications. Instead of having to handle a single case for bad proof, now we have to handle multiple. Then, testing is much harder, but also any change has the possibility to break so many things.

If we dump a bad block, we know it is bad and we can investigate. I see 0 value in this part here.
@jsync-swirlds any opinions?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
```diff
@@ -158,13 +162,15 @@ private VerificationNotification processCompletionException(final CompletionExce
...
-            return new VerificationNotification(
+            final VerificationNotification notification = new VerificationNotification(
                     false,
                     getFailureInfo(vfe.getBlockNumber(), vfe.getFailureType()),
                     vfe.getBlockNumber(),
                     null,
-                    null,
+                    vfe.getBlock(),
                     vfe.getBlockSource());
+            badBlockDumper.attemptDump(notification, vfe.getHapiVersion());
```
ata-nas: I think we must dump the block in the `handleThrowable` method. I think we should wrap the logic there with `try-finally` and put the dump in the finally clause. In any case, if we are handling a throwable, we have some kind of failure. This, being here, misses a few cases. Only when a block is canceled, maybe then we have the case where we can say we do not want a dump, but I would argue we better have it as well.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/VerificationSessionFailedException.java
```diff
@@ -10,12 +12,37 @@ public class VerificationSessionFailedException extends RuntimeException {
     private final long blockNumber;
     private final SessionFailureType failureType;
     private final BlockSource blockSource;
+    /// The raw block bytes captured at the failure site, for diagnostic dump. May be {@code null}
+    /// when the block had not yet been fully assembled (e.g. failures during hashing).
+    private final BlockUnparsed block;
```
ata-nas: `BlockUnparsed` is actually a parsed proto, we just name it so because its contents are unparsed. In order to provide the actual raw bytes, we actually want to use a byte array or `Bytes`. I would actually recommend that. The reason for that is because it is possible that we have the full block received (we know that because `BlockItems` will tell us), but the block fails to parse. If the block fails to parse, but we have the whole thing, we will not be able to transform it to `BlockUnparsed` anyway, but we will still have the complete block available for dumping.

In the case of partial block, I think we can still dump it. We can still parse the items and do some diagnostics. I feel like we should never have a `null` case for raw bytes. In any case we throw this exception only when a block fails the verification process.
@jsync-swirlds opinions?

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/BadBlockDumperTest.java
```diff
@@ -0,0 +1,172 @@
...
+import org.hiero.block.node.spi.blockmessaging.VerificationNotification;
+import org.hiero.block.node.spi.blockmessaging.VerificationNotification.FailureInfo;
+import org.hiero.block.node.spi.blockmessaging.VerificationNotification.FailureType;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class BadBlockDumperTest {
+
+    @TempDir
+    Path tempDir;
```
ata-nas: Non-blocking nit: we can use `jimfs` instead.

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/VerificationServicePluginTest.java
```diff
@@ -1,27 +1,77 @@
...
+                new BlockItems(block0Info.blockUnparsed().blockItems(), block0Info.blockNumber(), true, true));
+        testThreadPoolManager.executor().executeSerially();
+
+        assertEquals(1, blockMessaging.getSentVerificationNotifications().size());
+        final VerificationNotification notification =
+                blockMessaging.getSentVerificationNotifications().getFirst();
+        assertNotNull(notification);
+        assertTrue(notification.success(), "Block 0 TSS-wraps block must verify successfully");
+        assertEquals(block0Info.blockNumber(), notification.blockNumber());
+    }
```
ata-nas: You can safely remove this positive test. I have this covered in another branch already and will conflict with you. I have even expanded the resource test sources to work with a `TestBlock` abstraction in my branch instead of the SampleBlockInfo.

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/VerificationServicePluginTest.java
```diff
@@ -1,27 +1,77 @@
...
+        final BlockUtils.SampleBlockInfo block0Info = BlockUtils.getSampleBlockInfo(BlockUtils.SAMPLE_BLOCKS.BLOCK_0);
+        blockMessaging.sendBlockItems(new BlockItems(block0Info.blockUnparsed().blockItems(), 1L, true, true));
+        testThreadPoolManager.executor().executeSerially();
+
+        assertEquals(1, blockMessaging.getSentVerificationNotifications().size());
+        final VerificationNotification notification =
+                blockMessaging.getSentVerificationNotifications().getFirst();
+        assertNotNull(notification);
+        assertFalse(notification.success(), "Block with mismatched header number must fail verification");
+        assertEquals(
```
ata-nas: This is the fragility I am talking about. Now, we  say we have invalid header, that might not be the case, as it is even with this test. We do not have an invalid header, but we witness a mismatch between a valid block's valid header and what the internal messaging is reporting. The block will fail proof verification if the block is actually invalid. Here, we are actually misleading investigation with the failure type. And imagine, any change to the failure type will now have to come with the burden to support tests as well.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -777,7 +777,7 @@ void testHandleVerificationEqualToLatest() {
                 // Source must be publisher.
                 final VerificationNotification notification = new VerificationNotification(
                         false,
-                        FailureInfo.standard(FailureType.BAD_BLOCK_PROOF),
+                        FailureInfo.standard(FailureType.UNKNOWN_ERROR),
```
ata-nas: Again, this should not be changed I believe. The publisher manager will very shortly start using these failure types. Tests will break. Currently, the whole system assumes bad block proof because we have never had failure types until recently, hence everywhere you see failure type used, it is bad proof by default. This change in tests invites trouble when plugins actually start to use the failure types. Worst case is false positives will be present.

ata-nas (commented):
I have a few questions.
Reviewed the reworked verification part + shared notifications and failure types. Most of what I question is also valid for the current verification plugin.

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/BadBlockDumperTest.java
jsync-swirlds: Use of `jimfs` is strongly encouraged where possible. It avoids thrash on the CI container disk mounts and improves unit test performance. It also improves isolation, which helps with running tests in parallel.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
jsync-swirlds: Also, unknown error is a terrible default in general. Except for _specific_ tests that need this enum, we should not use it in tests.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionFailureType.java
jsync-swirlds: I tend to agree.  We tried to achieve a good balance between _enough_ detail without becoming _over specified_ in the initial definitions.

At least some of these are well into the over specified region.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/VerificationSessionFailedException.java
jsync-swirlds: If we have the raw bytes, we should dump that data.
Otherwise we should serialize the object before writing.

ata-nas: We receive `BlockItemsUnparsed`. I think we should be always able to wrap them into a `BlockUnparsed`, even if we do not have the complete block. The only exception is when we fail to parse the block items (a `ParsedException` is thrown). In that exceptional case, we are still able to persist the data to disk, but I am not sure if we can wrap them into a `BlockUnparsed`. I think we will be able to, just that when we try to parse an individual item that fails, we will see a parsed exception again. We have to verify this, however. But that case is also important to be covered so we can investigate. If what I am saying is correct, the verification failure exception must never accept a null `BlockUnparsed` field. Semantic Hapi Version could be null because it can actually fail to parse.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/BadBlockDumper.java
```diff
@@ -0,0 +1,212 @@
...
+                        LOGGER.log(System.Logger.Level.DEBUG, "Failed to check/delete dump file {0}", file, e);
+                    }
+                });
+            } catch (final IOException e) {
+                LOGGER.log(System.Logger.Level.DEBUG, "Failed to list dump directory {0} for purge", dir, e);
+            }
+        }
+    }
+
+    private void removeDumpedKey(@NonNull final Path file) {
```
Nana-EC, line 206: The need for the string parsing in this method feels like any changes to one of a few assumptions (e.g. failureType doesn't include a `-`, no additional hyphen delimited part is added) could silently break us.
Instead of key being <blockNumber>:<failureType> should it just be the whole file name e.g. `block-{blockNumber}-{failureType}-{uuid}`?
This would make reference easy, avoid string parsing and avoid breakages

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/BadBlockDumper.java
```diff
@@ -0,0 +1,212 @@
...
+            @NonNull final Path path,
+            @NonNull final String correlationId,
+            final long blockNumber,
+            @NonNull final FailureType failureType,
+            final SemanticVersion hapiVersion)
+            throws IOException {
+        final String hapiVersionStr = hapiVersion == null
+                ? "unknown"
+                : "%d.%d.%d".formatted(hapiVersion.major(), hapiVersion.minor(), hapiVersion.patch());
+        final String json = """
```
Nana-EC, line 169: Works for now but might be better to use an actual JSON builder to avoid any potential of malformed JSON based on the values of items that will replace `%s`

Nana-EC (dismissed):
LG, non blocking suggestions that can be address in follow up

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/VerificationSessionFailedException.java
jsync-swirlds: We've already parsed the data if it's a `BlockItemsUnparsed`.
We should not be parsing at all, we should just be serializing so we can write the bytes to disk.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -515,22 +550,42 @@ private VerificationNotification verifyRsaProof(
...
+                        null,
+                        new BlockUnparsed(blockItems),
+                        blockSource);
+            } catch (final SignatureException e) {
+                LOGGER.log(
+                        WARNING,
+                        "RSA signature verification failed for node {0} in block {1}: {2} — rejecting block",
+                        nodeId,
+                        blockNumber,
+                        e.getMessage());
```
jsync-swirlds: `e.getMessage` is somewhat low reliability; `e` is almost always a better value to pass to a log statement.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/BadBlockDumper.java
jsync-swirlds: This data isn't intended to be read by software; as long as it's readable by a person it should be useable.
Also, perhaps worth noting that this option is a lot faster than running a full JSON codec process.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionFailureType.java
ivannov: In the requirements in #2745 there was this:

> Log the failure category at WARNING (MISSING_FOOTER, MALFORMED_PROOF_STRUCTURE, STATE_PROOF_INVALID, SIGNATURE_MISMATCH, …)

I assumed that part of the requirements (aside the warning level and dumping the result) was to have more detailed failure types.

As requested here and in other comments, I restored the `BAD_BLOCK_PROOF` and removed the detailed types

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ivannov: Removed the check altogether (and the test for it)

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/BadBlockDumper.java
ivannov: I'd call this log message with structure. Not a service to service communication protocol :)

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/BadBlockDumper.java
ivannov: Changed as per your request. Also simplified the file name - removed the `{uuid}` suffix as I don't think we need it

Did it for both `BadBlockDumper` classes

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/VerificationServicePluginTest.java
ivannov: Removed it

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/VerificationServicePluginTest.java
ivannov: Removed the test altogether

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
ivannov: It was like that before the change. I can change that, but:

* Because the exception overrive of the log message doesn't take vararg, I have to change from vararg to string interpolation
* To be consistent, I have to do it everywhere
* This is the old verification plugin, should we invest in the log message overhaul there?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
ivannov: Moved it to a finally block in the `handleThrowable` method

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/VerificationSessionFailedException.java
ivannov: Changed the field to `List<BlockItemUnparsed>`

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/BadBlockDumperTest.java
ivannov: Easy fix, changed to JimFS. Will have that in mind next time

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Thanks, I'll consider it in #3007.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds: This is not what I suggested.

If you just pass `e` this isn't to get the exception override, it's to get a useable string.
e.getMessage is frequently null or empty.
e.toString prints the exception type then the message, so the string is more useful.

Anytime you see `e.getMessage()` in a log statement, _without any other change_ you can replace with `e` and the net result is a clearer and more useful message.

That is not to say that you cannot make your proposed change; only that it is not what I was suggesting.

Here is what I was suggesting:
```suggestion
                        e);
```

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
jsync-swirlds, line 1: Slight concern about high level design here.
Not blocking, but something we should probably discuss more widely.

Why are we building this bad block dump into the core mainline of the verification plugin?

If this is truly a debugging feature (and it absolutely is), then why not make the function to write out bad blocks a separate plugin that listens to the verification notifications?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -226,4 +234,12 @@ public void handlePersisted(final PersistedNotification notification) {
...
         }
     }
+
+    private static String resolveHostname() {
+        try {
+            return InetAddress.getLocalHost().getHostName();
+        } catch (final UnknownHostException e) {
+            return "unknown";
+        }
+    }
```
jsync-swirlds, line 244: Hostname is low value information, in general, particularly in containerized workloads.
I would recommend pushing this into the `BadBlockDumper` class and not make it something passed in.

I get the "easier to test" aspect, but it's also something that doesn't actually need to be tested, as it's not more complex than a simple accessor method.

jsync-swirlds (approved):
Some concerns about design direction, but those are better addressed in future work.

ata-nas (approved):
LG. @jsync-swirlds has a valid concern which I agree with.
If this had been a simple plugin, things would have been so much easier and cleaner.
Now, we still have to require a PVC for verification, something which we removed recently.

@AlfredoG87 do some changes need to be done to charts to support the new path?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
ata-nas: I agree, decoupling this functionality from verification all together sounds like the path forward.

PR #3070, opened 2026-06-22
feat: BN: historical address-book roster model + bootstrap

### LINKED ISSUE #2644: Add Candidate Roster updates to ApplicationStateFacility
### Story Form

As a BlockNodeEngineer I would like BlockNode to accept new Rosters and update the plugins with the new Roster

### Technical Notes

Rosters can be updated during upgrades, but new Rosters may be created in between upgrades

### LINKED ISSUE #2958: BN: historical address-book roster model + bootstrap

**Prereq:** none. Foundational — T3, T5, T6 build on it.

## Story Form

As a Block Node,
I want to hold multiple historical address books each scoped to a block-number range and load them at startup,
So that I can later verify a WRB against the address book that was in effect for that block.

## Technical Notes

- Define the roster model: ordered, range-keyed `{addressBook, startBlock, endBlock}` with efficient lookup-by-block. Define + document the on-disk roster file format (input for T3).
- Extend `RsaRosterBootstrapPlugin` / `RsaRosterBootstrapConfig` from a single current book to the roster; load (bootstrap) the roster file at startup. Keep backward compatibility with a single-book deployment (one entry covering all blocks).
- No verification behavior change in this ticket (T5); this is the model + load + expose-for-lookup.

## Design

berryware:
## Summary

- Adds a block-number-keyed address-book history model (`RangedAddressBookHistory` proto) so the
  Block Node can select the `NodeAddressBook` that was in effect for any given block number,
  enabling correct RSA verification of historical Wrapped Record Blocks (WRBs).
- Implements O(log n) range lookup (`AddressBookHistoryLookup`) and wires the history through the
  full application stack: config → `BlockNodeContext` → `BlockNodeApp` load/persist →
  `RsaRosterBootstrapPlugin` startup path.
- Backward-compatible: existing single-book deployments require no operator action — if only the
  old `rsa-bootstrap-roster.json` is present the BN wraps it into a single open-ended era at startup.

### Proto
- `protobuf-sources/…/ranged_address_book_history.proto` — new `RangedAddressBookHistory` /
  `RangedNodeAddressBook` messages; `endBlock = 0` is the open-ended sentinel for the latest era.

### Lookup utility
- `AddressBookHistoryLookup` — builds a `NavigableMap<startBlock, RangedNodeAddressBook>` and
  resolves the correct `NodeAddressBook` via `floorEntry(b)` + endBlock check (O(log n)).

### Configuration
- `ApplicationStateConfig.rsaAddressBookHistoryFilePath` — new config property (default:
  `/opt/hiero/block-node/application-state/rsa-address-book-history.json`).
  Env var: `APP_STATE_RSA_ADDRESS_BOOK_HISTORY_FILE_PATH`.

### SPI
- `BlockNodeContext` — adds `nodeAddressBookHistory` field (12th); 15 call sites updated.
- `ApplicationStateFacility` — adds `updateAddressBookHistory()` as a `default` no-op to avoid
  breaking existing implementations.

### Application
- `BlockNodeApp` — loads the history file at startup (validates non-empty); falls back to the
  single-book backward-compat bridge; persists updates via the atomic `.tmp`-then-rename pattern;
  drains pending history through the existing scan-tick cycle.

### Plugin
- `RsaRosterBootstrapPlugin.start()` — checks `nodeAddressBookHistory` first; records metrics via
  new `recordHistoryMetrics()` and returns early; falls back to single-book / Mirror-Node paths
  when no history is present.
- New gauge: `blocknode:roster_eras_loaded`; existing `roster_entries_loaded` now reflects the
  total across all eras in history mode.

### Docs
- `docs/design/wrb-streaming/sp-wrb-bn-design.md` — new §2a documents the on-disk format,
  invariants table, minimal JSON example, lookup algorithm, codec, precedence rules, and config key.
- `docs/block-node/configuration.md` — new `APP_STATE_RSA_ADDRESS_BOOK_HISTORY_FILE_PATH` row and
  "history vs. single-book file" explanation.

## Test plan

- [ ] `AddressBookHistoryLookupTest` — 14 unit tests: empty history, multi-entry index, exact
  start/end boundaries, one-past-end, mid-range, first block of second era, open-ended sentinel
  (large block, start of open-ended entry, block before open-ended entry), gaps between eras
  (mid-gap, gap-start), single-entry backward-compat (all blocks covered).
- [ ] `RsaRosterBootstrapPluginTest` — 4 new tests in `PreloadedAddressBookHistory`: history path
  skips Mirror Node, records era count metric, records entry count metric, single-book path still
  works when no history present.
- [ ] `BlockNodeAppTest` — 4 new tests: two-era history file loads correctly, single-book file is
  wrapped into backward-compat era, corrupt history file throws on startup, empty history list
  throws on startup.
- [ ] All 53 `roster-bootstrap-rsa` tests pass; all 72 `app` tests pass; full project compiles
  clean.

## Related Issue(s)
Closes: #2958


block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -32,6 +38,7 @@ public record ApplicationStateConfig(
         // spotless:off
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/tss-bootstrap-roster.json") Path tssBootstrapFilePath,
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/rsa-bootstrap-roster.json") Path rsaBootstrapFilePath,
+        @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/rsa-address-book-history.json") Path rsaAddressBookHistoryFilePath,
```
Nana-EC: Q: Did we consider supporting both formats and catching it in file parse instead of needing a new path?
We'll have to maintain the previous path going forward

.github/workflows/k6-tests.yml
```diff
@@ -20,7 +20,7 @@ permissions:
 
 env:
   GRADLE_EXEC: ionice -c 2 -n 2 nice -n 19 ./gradlew --no-daemon
-  SOLO_VERSION: "0.72.0"
+  SOLO_VERSION: "0.73.0"
```
jsync-swirlds: Should we be using Solo `0.79` (latest release)?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -127,6 +131,12 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     /// Pending address book loaded at startup; consumed by the first checkForApplicationStateUpdates run.
     private final AtomicReference<NodeAddressBook> pendingAddressBook = new AtomicReference<>();
 
+    /// Pending address book history loaded at startup; consumed by the first checkForApplicationStateUpdates run.
+    private final AtomicReference<RangedAddressBookHistory> pendingAddressBookHistory = new AtomicReference<>();
+
+    /** Cached O(log n) index built from the current address book history; rebuilt whenever history changes. */
+    private volatile NavigableMap<Long, RangedNodeAddressBook> addressBookIndex = new TreeMap<>();
```
jsync-swirlds, line 137: Is there any chance of this map being accessed by one thread while also being written by another?
If so, we should probably use a `ConcurrentSkipListMap` instead.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -831,6 +964,60 @@ private void loadApplicationState(final Configuration configuration) {
...
+        // Load RSA address book history (JSON format) — takes precedence over the single-book file.
+        // If absent but a single-book file was loaded above, wrap it into a single open-ended era
+        // so the rest of the application sees a consistent RangedAddressBookHistory regardless of
+        // which file is present (backward-compatibility bridge).
+        final Path historyFilePath = appStateConfig.rsaAddressBookHistoryFilePath();
+        if (Files.exists(historyFilePath)) {
+            try {
+                final RangedAddressBookHistory history = standardParse(
+                        RangedAddressBookHistory.JSON,
+                        Bytes.wrap(Files.readAllBytes(historyFilePath)).toReadableSequentialData());
```
jsync-swirlds: The toReadableSequentialData is unnecessary:

```suggestion
                        Bytes.wrap(Files.readAllBytes(historyFilePath)));
```

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -748,6 +853,34 @@ private void persistNodeAddressBook(NodeAddressBook nodeAddressBook) {
...
+                .rsaAddressBookHistoryFilePath();
+        try {
+            final Path tmp = filePath.resolveSibling(filePath.getFileName() + ".tmp");
+            final Bytes encoded = RangedAddressBookHistory.JSON.toBytes(history);
+            Files.write(tmp, encoded.toByteArray());
+            try {
+                Files.move(tmp, filePath, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE);
+            } catch (AtomicMoveNotSupportedException e) {
+                LOGGER.log(
+                        WARNING,
```
jsync-swirlds: This isn't indicating anything broken, so perhaps it should be `INFO` level.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -748,6 +853,34 @@ private void persistNodeAddressBook(NodeAddressBook nodeAddressBook) {
...
+                        "Atomic move not supported on this filesystem for {0}; falling back to non-atomic replace",
+                        filePath);
+                Files.move(tmp, filePath, StandardCopyOption.REPLACE_EXISTING);
+            }
+            LOGGER.log(
+                    INFO,
```
jsync-swirlds: This is happy-path detail, which means the highest permitted level is `DEBUG` and it is recommended to trace `TRACE` or remove the log entirely.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -831,6 +964,60 @@ private void loadApplicationState(final Configuration configuration) {
...
+                final RangedAddressBookHistory history = standardParse(
+                        RangedAddressBookHistory.JSON,
+                        Bytes.wrap(Files.readAllBytes(historyFilePath)).toReadableSequentialData());
+                if (history.addressBooks().isEmpty()) {
+                    throw new IllegalStateException(
+                            "RSA address book history file contains no entries: " + historyFilePath);
+                }
+                pendingAddressBookHistory.set(history);
+                LOGGER.log(
+                        INFO,
```
jsync-swirlds: Here also, recommend we use `TRACE` or just remove the log.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -371,6 +526,22 @@ private void recordSuccessMetrics(NodeAddressBook book, long startTimeMillis, St
...
+    private void recordHistoryMetrics(RangedAddressBookHistory history, long startTimeMillis) {
+        rosterErasLoaded = history.addressBooks().size();
+        rosterEntriesLoaded = history.addressBooks().stream()
+                .map(RangedNodeAddressBook::addressBook)
+                .filter(book -> book != null)
+                .mapToLong(book -> book.nodeAddress().size())
+                .sum();
+        rosterLoadDurationMs = System.currentTimeMillis() - startTimeMillis;
+        LOGGER.log(
+                INFO,
```
jsync-swirlds, line 538: Generally, most of the log statements in this file should probably be logged at least 1 level lower.

Principles:
1. Generally do not log success; if necessary, log at `TRACE`.
1. Events that indicate minor and recoverable errors should be logged either `DEBUG` or `INFO`.
1. Events that indicate major errors, but are not fatal, should be logged at `WARNING`.
1. An `ERROR` log indicates the system is about to fail or shut down due to an error.

jsync-swirlds (dismissed):
Non-Blocking; but we should probably audit and quiet down the logging in this change.

AlfredoG87 (dismissed):
LGTM

docs/design/wrb-streaming/sp-wrb-bn-design.md
```diff
@@ -107,6 +107,84 @@ to a `[startBlock, endBlock]` range**:
...
+    },
+    {
+      "addressBook": {
+        "nodeAddress": [
+          { "nodeId": "1", "RSA_PubKey": "3082..." },
+          { "nodeId": "3", "RSA_PubKey": "3082..." }
+        ]
+      },
+      "startBlock": "5000001",
+      "endBlock":   "0"
```
Nana-EC, line 174: Q: to confirm this is incorrect and endBlock on the latest AddressBook shouldn't be 0.
It should be something that indicates an open range.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
```diff
@@ -256,6 +259,16 @@ private TssData buildTssData(
                     .build();
         }
 
+        private static RangedAddressBookHistory buildRangedHistory(final int count) {
```
Nana-EC, line 262: Okay to start but i'm guessing if this is the only method then the tests are limited to a History with only 1 address book that covers only block 0.
An expansion would be to handle the number of desired addressBooks and their start and end blocks

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/BlockNodeContext.java
```diff
@@ -70,6 +76,7 @@ public static class Builder {
         BlockNodeVersions blockNodeVersions;
         TssData tssData;
         NodeAddressBook nodeAddressBook;
+        public RangedAddressBookHistory rangedAddressBookHistory;
```
Nana-EC, line 85:
```suggestion
       RangedAddressBookHistory rangedAddressBookHistory;
```

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -328,6 +365,124 @@ private void fetchFromMirrorNode() {
...
+    /// Converts a (from, to) timestamp pair to a [startBlock, endBlock] pair by querying the
+    /// Mirror Node blocks API. Returns {@code null} if the block range cannot be determined.
+    ///
+    /// Blank timestamps are treated as sentinels: blank {@code from} means genesis (startBlock=0)
+    /// with no API call; blank {@code to} means open-ended (endBlock=0) with no API call.
+    private long[] fetchBlockRange(final HttpClient client, final String from, final String to) {
+        final long startBlock, endBlock;
+        if (from == null || from.isBlank()) {
+            startBlock = 0L; // no from-timestamp → treat as genesis
+            endBlock = 0L;
```
Nana-EC: Q: are we sure about endBlock = 0.
To me that assumes a new addressBook starts with 1

Nana-EC (commented):
Good changes.
I'm still on the fence about
- choice of endBlock as 0 by default
- new config field

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: changed to -1

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/BlockNodeContext.java
berryware: Changed

docs/design/wrb-streaming/sp-wrb-bn-design.md
berryware: Changing to -1 for endBlock sentinel value.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
berryware: Removed. ParserException will be used to determine which bootstrap file it is.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
berryware: Understood.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: removed

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: removed

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Done

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: Went back to the Codec.parse in order to raise a ParseException when the first codec does not match.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: It is never written to directly. The entire index is rebuilt on change and the volatile reference is overwritten by the BlockNodeApp thread. Other threads may read from it.

.github/workflows/k6-tests.yml
berryware: Will do this in a separate PR for clarity

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: `standardParse` should still throw the ParseException if `Codec.parse` would have.
Exception handling shouldn't require switching back to direct Codec use.

berryware: standardParse does not let me set parseUnknownFields. At 'true' is swallows the old format and does not throw ParseException.

berryware: I think it's a problem that parseUnknownFields=true swallows a completely different object and returns the default of the Object being requested. I don't think that should be allowed.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
Nana-EC: Okay, about to review but i'd expect
1. Parse new format
2. Catch exception try old format
3. Catch and throw as it's try bad.

PBJ does support a tryParse unfortunately

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: I see what you're intending now.

For forward compatibility we must set `parseUnknownFields`, which is why that's true in the standard parse. For anything coming from CN we _must_ be non-strict and allow unknown fields.

For resilience, we generally want the same for local config, but it's possible we might want something different.

For this specific case we have three choices I see:
1. Make the new structure backwards compatible with the new; properly reserve old fields and put new data in new fields (with new field numbers).
   - Detect the different types based on which fields are filled in; or because the old structure uses a reserved field number, which causes parse exception, even with unknown fields allowed.
1. Use a completely different structure, with a different message (which Nana seems to prefer not doing)
    - Detection is parse exception due to wrong message type
1. Change the type in a manner that is not backwards compatible
   - Detection relies on setting strict parsing, no parsing of unknown fields, and parse exception.

I tend to prefer options (1) or (2).

Nana-EC (approved):
LG.
My initial feedback was addressed.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: @Nana-EC ?

jsync-swirlds (approved):
Let's merge this, but we should revisit the structure/form of the proto for storing the address data, and see if we can standardize parsing.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/server/TestBlockNodeServer.java
berryware: captured in #3088

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
berryware: captured in #3088

AlfredoG87 (approved):
LGTM

Issue #2957, opened 2026-06-04
Tool/script: convert CLI address-book history to BN roster format


Nana-EC:
Part of epic #2954 — _Distribute historical & live WRBs and TSS to Tier-1 Block Nodes_.

**Prereq:** T4 (the roster file format it must emit). Can be built in parallel once that format is fixed.

## Story Form

As a node operator,
I want a script/tool that converts the CLI's historical address-book JSON files into the Block Node's address-book-roster format,
So that the special-purpose BN can be bootstrapped to verify historical WRBs across all past address books.

## Technical Notes

- Input: the CLI's ~20 historical address-book JSON files (from `mirror generateAddressBook*` / `docs/address-book-updating.md`), each effectively scoped to a consensus-time/block range.
- Output: the BN roster format defined by T4 (set of `{addressBook, startBlock, endBlock}`).
- Extends `genesis-network-to-bn-rsa-roster.sh` / `generate-rsa-roster-bootstrap.sh`. Deliver as a standalone operator script (preferred) or a CLI subcommand — decide with T4's format.
- Needs consensus-time → block-number mapping (see open question); document the source used.

## Design

`docs/design/wrb-streaming/sp-wrb-bn-design.md`

rockysingh:
This is a duplicate.

Issue #2954, opened 2026-06-04
Distribute historical & live WRBs and TSS to Tier-1 Block Nodes


Nana-EC:
## Goal

Stand up a special-purpose ("Tier 0") Block Node co-located with the CLI that ingests historical and live Wrapped Record Blocks (WRBs) via push and serves them (and TSS) to Council Tier-1 Block Nodes, with verification of any historical WRB, before the WRB release (0.75, July).

## Intention

We will push WRBs from the CLI into the special-purpose BN through the existing ingestion API — a one-time backfill of stored wrapped-block zips plus a live feed of each new block — instead of restructuring storage. We will make the BN verify historical WRBs by holding a roster of address books scoped to block-number ranges, bootstrapped from the CLI's existing address-book history and kept current from Mirror Node. Tier-1 nodes then backfill blocks and query TSS from the special-purpose BN using existing paths. The work reuses the CLI's wrap/address-book tooling, the gRPC publish client, and the BN's RSA-roster and verification plugins.

## Considerations

- Push via the ingestion API (chosen) vs. restructuring on-disk storage (rejected: needs BN restarts, bypasses validation).
- One-block-at-a-time live push (chosen) vs. 6-hour zip batches (rejected: operational floods, lag).
- Historical address-book roster in the BN (chosen, durable) vs. bypassing verification or current-book-only (rejected: can't validate historical WRBs; throwaway).
- A simplified backfill-source API plugin to reduce failure points — discussed, undecided (open).
- Timeline: hardware by mid-June, backfill before the July release; CLI + special BN run continuously for the ~2-month transition.

## Technical Details

- CLI = `tools-and-tests/tools` (`BlockStreamTool`). Reuse `networkCapacity`/simulator `PublishStreamGrpcClient` for push; `mirror generateAddressBook*` for address-book history.
- BN verification = `block-verification`/`verification` `ExtendedMerkleTreeSession`; current roster = `RsaRosterBootstrapPlugin` / `RsaRosterBootstrapConfig`.
- TSS retrieval already implemented (`RosterBootstrapTssPlugin.queryPeerTssData()`); only an operator file move is required (covered by T7).

## Tasks and Features

- [ ] #2955 — T1 — CLI: backfill-push historical WRBs to a BN
- [ ] #2956 — T2 — CLI: live-push each new WRB to a BN (prereq: T1)
[...]

PR #3071, opened 2026-06-22
feat:  add blocks convert-address-book-history subcommand in wrb cli (#2957)

rockysingh:
## Summary

Adds the `blocks convert-address-book-history` CLI subcommand. It converts the CLI's historical address-book JSON (`AddressBookHistory` proto, produced by `mirror generateAddressBook*`) into a block-range-keyed roster history file for the special-purpose Block Node to use when verifying historical Wrapped Record Blocks.

This is **T3** of the WRB-distribution epic ([#2954](https://github.com/hiero-ledger/hiero-block-node/issues/2954)). The intended consumer is the BN-side historical RSA-roster bootstrap (**T4 / [#2958](https://github.com/hiero-ledger/hiero-block-node/issues/2958)**, not in this PR).

### Output format (T4-facing, **Option B** per team alignment)

```json
{
  "entries": [
    {
      "startBlock": 0,
      "endBlock":   4999,
      "blockTimestamp": { "seconds": 1234567890, "nanos": 0 },
      "addressBook": "<base64-encoded NodeAddressBook proto>"
    },
    {
      "startBlock": 5000,
      "endBlock":   null,
      "blockTimestamp": { "seconds": 1234600000, "nanos": 0 },
      "addressBook": "<base64-encoded NodeAddressBook proto>"
    }
  ]
}
```

- The full `NodeAddressBook` proto is preserved per era (Base64) so T4 can extract whatever fields it needs at load time.
- The last era's `endBlock` is `null` (open-ended).
- Consensus-time → block-number resolution uses the existing `BlockTimeReader` (`block_times.bin`); no Mirror Node round-trip needed.

### Usage

```sh
blocks convert-address-book-history \
  --input addressBookHistory.json \
  --output rsa-bootstrap-roster-history.json \
  [--block-times-file metadata/block_times.bin]
```

## Refs

- Epic: #2954
- T3: #2957 (this)
- T4 (downstream consumer): #2958
- Design: `docs/design/wrb-streaming/sp-wrb-bn-design.md`


tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
```diff
@@ -0,0 +1,190 @@
...
+        JSON_MAPPER.writeValue(outputFile.toFile(), root);
+        System.out.println(Ansi.AUTO.string(
+                "@|green Wrote " + sorted.size() + " roster entries to " + outputFile.toAbsolutePath() + "|@"));
+        return 0;
+    }
+
+    private static AddressBookHistory loadAddressBookHistory(Path file) throws IOException {
+        try (ReadableStreamingData in = new ReadableStreamingData(Files.newInputStream(file))) {
+            return AddressBookHistory.JSON.parse(in);
+        } catch (Exception e) {
```
Nana-EC: nit: A bit general. Should we break this into an IOException and RuntimeException. This would help distinguish malformed proto vs other types of errors

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
```diff
@@ -0,0 +1,190 @@
...
+            tsNode.put("seconds", ts.seconds());
+            tsNode.put("nanos", ts.nanos());
+            entry.put("addressBook", encodeAddressBook(book));
+        }
+
+        Path parent = outputFile.toAbsolutePath().getParent();
+        if (parent != null) {
+            Files.createDirectories(parent);
+        }
+        JSON_MAPPER.writeValue(outputFile.toFile(), root);
```
Nana-EC: Could do an atomic move to ensure a complete write
```suggestion
final Path tmp = outputFile.resolveSibling(outputFile.getFileName() + ".tmp");
JSON_MAPPER.writeValue(tmp.toFile(), root);
Files.move(tmp, outputFile, java.nio.file.StandardCopyOption.ATOMIC_MOVE,
           java.nio.file.StandardCopyOption.REPLACE_EXISTING);
```

Nana-EC (dismissed):
LG
2 suggestions

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
rockysingh: Done in 6edc2164 — `loadAddressBookHistory` now lets `IOException` propagate from the underlying input stream (so missing/unreadable files surface as IO errors) and catches PBJ's `ParseException` separately, rewrapping it as a `RuntimeException` with a "Malformed AddressBookHistory JSON in <path>" message. Added a `malformedInput` test to lock that path in.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
rockysingh: Done in 6edc2164 — the output is now serialized to a sibling `<output>.tmp` and renamed via `Files.move(..., ATOMIC_MOVE, REPLACE_EXISTING)`, with the `.tmp` deleted if the serialize step throws. Added an `atomicWriteLeavesNoTmpFile` test that asserts the destination exists and no stale `.tmp` is left behind on success.

Nana-EC (dismissed):

tools-and-tests/tools/build.gradle.kts
```diff
@@ -12,36 +12,38 @@ description = "Hiero Block Stream Tools"
 application { mainClass = "org.hiero.block.tools.BlockStreamTool" }
 
 mainModuleInfo {
-    requires("org.hiero.block.protobuf.pbj")
-    requires("org.hiero.block.node.base")
+    requires("com.hedera.bucky")
+    requires("com.hedera.pbj.grpc.client.helidon")
+    requires("com.hedera.pbj.grpc.helidon")
+    requires("com.hedera.pbj.runtime")
```
berryware, line 18: Duplicated

tools-and-tests/tools/src/test/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommandTest.java
```diff
@@ -0,0 +1,419 @@
...
+                    .toList();
+        }
+    }
+
+    private static DatedNodeAddressBook dated(Instant instant, NodeAddressBook book) {
+        return new DatedNodeAddressBook(new Timestamp(instant.getEpochSecond(), instant.getNano()), book);
+    }
+
+    private static NodeAddressBook addressBook(long firstNodeId, int count, String rsaPubKeyHexBase) {
+        java.util.List<NodeAddress> nodes = new java.util.ArrayList<>();
```
berryware: nit: import List and ArrayList and use short names

berryware (commented):
Some nits

tools-and-tests/tools/build.gradle.kts
rockysingh: Done in 24e27553 dropped the duplicate `requires("com.hedera.pbj.runtime")` line.

tools-and-tests/tools/src/test/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommandTest.java
rockysingh: Done in 24e27553  added imports for `ArrayList` and `Arrays` and switched the three call sites to short names (`List<NodeAddress>`, `new ArrayList<>()`, `Arrays.stream(...)`).

Nana-EC (approved):

berryware (approved):
Looks good

PR #3075, opened 2026-06-23
chore: Allow Out of Order Verification for Sources Other Than Publisher

ata-nas:
* Added a new config that will allow enabling/disabling strict ordering for blocks with sources other than Publisher
* Order Manager uses that config


AlfredoG87 (commented):
Looking good, we are missing Configuration Properties on Docs. 

https://github.com/hiero-ledger/hiero-block-node/blob/main/docs/block-node/configuration.md

I think we should add a new section for `Block Verification` and eventually deprecate and delete the current section for `Verification`

ata-nas:
> Looking good, we are missing Configuration Properties on Docs.
> 
> https://github.com/hiero-ledger/hiero-block-node/blob/main/docs/block-node/configuration.md
> 
> I think we should add a new section for `Block Verification` and eventually deprecate and delete the current section for `Verification`

I will remove most of the configs and will update the doc then, making a new section for the reworked plugin. That will be in another PR which will tackle the follow ups.

AlfredoG87 (approved):

jsync-swirlds (approved):

PR #3076, opened 2026-06-23
feat(solo-e2e): per-plugin port support and deployment lock fix

AlfredoG87:
## Summary

- **Helm chart dedup**: `pluginServicePorts` / `pluginContainerPorts` helpers now skip duplicate port entries (Kubernetes rejects Services/containers with the same port number declared twice).
- **Topology-driven plugin ports**: `generate_bn_overlay` now emits `blockNode.ports` from the topology `plugin_ports` map, and a new `generate_cn_block_nodes_json` function produces per-CN `block-nodes.json` overrides with separate `streamingPort` (publisher) and `servicePort` (serverStatus) — applied to CN pods after deployment to replace Solo's hardcoded port 40840 entries.
- **Solo deployment lock fix**: a background `solo block node add` process from a prior run can survive cluster teardown and continue renewing the K8s `solo-network` Lease in a newly created cluster, blocking the next deployment. Fixed by killing matching processes and deleting the Lease at both `cluster:destroy` time and at the start of `deploy_block_nodes`.

## Changes

**`charts/block-node-server/templates/_helpers.tpl`**
- `pluginServicePorts` and `pluginContainerPorts` iterate a single ordered list and track seen port numbers with a `$seen` dict, emitting only the first plugin entry per unique port number.

**`tools-and-tests/scripts/network-topology-tool/generate-chart-values-config-overlays.sh`**
- `generate_bn_overlay`: extended to emit `blockNode.ports.*` from `plugin_ports` topology keys in addition to backfill config; both sections are written under a single `blockNode:` root using a `{ } > file` block to prevent duplicate YAML top-level keys.
- New `generate_cn_block_nodes_json`: for any CN whose BNs declare non-default `plugin_ports.publisher` or `plugin_ports.serverStatus`, generates `cn-<name>-block-nodes.json` with explicit `streamingPort` and `servicePort` fields. Falls back to `topology port // 40840` when a plugin port is not overridden.

**`tools-and-tests/scripts/network-topology-tool/network-topology.schema.yaml`**
- Added `plugin_ports` (map of camelCase plugin name → port number) to the block node schema.

**`tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-network.sh`**
- `deploy_block_nodes`: kill orphaned `solo block node` processes + delete K8s Lease before iterating BNs.
- New `apply_cn_block_nodes_configs`: finds `cn-*-block-nodes.json` overlay files and copies them into each CN pod's `data/config/block-nodes.json`, replacing Solo's auto-generated version.
- `apply_cn_block_nodes_configs` called in `main` after `deploy_consensus_nodes`.
- Overlay cleanup now includes `cn-*-block-nodes.json`.

**`tools-and-tests/scripts/solo-e2e-test/Taskfile.yml`**
- `cluster:destroy`: kill orphaned processes + delete Lease before running Solo CLI teardown.

## Known Limitation

Solo's own "Check block node readiness" step in `solo block node add` hardcodes port 40840 for both the HTTP health probe and the gRPC `serverStatus` probe. Per-plugin port overrides for `health` or `serverStatus` will cause Solo's deployment to hang at this step, even though the K8s pod becomes `Ready` correctly. A separate issue has been filed with the Solo team to fix this in their readiness check flow.

## Test Plan

- [x] `task up` with default `single` topology (no `plugin_ports`) completes without lock errors
- [x] `task down` followed immediately by `task up` does not hit "lock already acquired" error
- [x] `task up` with a topology using `plugin_ports.publisher` / `plugin_ports.serverStatus` — verify `cn-node1-block-nodes.json` is generated with the expected `streamingPort` / `servicePort` and is applied to the CN pod
- [x] Helm template dry-run with two plugins sharing a port produces no duplicate Service or container port entries

### DIFF
--- a/.github/workflows/solo-e2e-scheduler.yml
+++ b/.github/workflows/solo-e2e-scheduler.yml
@@ -189,12 +189,16 @@ jobs:
           MATRIX_COUNT=$(echo "${MATRIX_JSON}" | jq '. | length')
           echo "matrix_count=${MATRIX_COUNT}" >> "$GITHUB_OUTPUT"
 
+          # MN pinned: v0.157.0+ (HIP-1137) broke Solo CLI's block node env var injection.
+          # Track https://github.com/hiero-ledger/solo/issues/4863 to unpin MN_PINNED.
+          MN_PINNED="v0.156.0"
+
           # Configure versions based on run type and trigger
           if [[ "${{ github.event_name }}" == "push" && "${{ github.ref_type }}" == "tag" ]]; then
             # TAG release: use the tag for BN, latest GA for CN/MN
             BN_VERSION="${{ github.ref_name }}"
             CN_VERSION="latest"
-            MN_VERSION="latest"
+            MN_VERSION="${MN_PINNED}"
           elif [[ "${RUN_TYPE}" == "rc" ]]; then
             # RC run: use SNAPSHOT BN against RC versions of CN/MN
             BN_VERSION="${{ inputs.block-node-version }}"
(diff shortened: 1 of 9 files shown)


berryware (dismissed):
Looks good.

jsync-swirlds (approved):

berryware (approved):
Looks good

PR #3080, opened 2026-06-24
build(deps): upgrade PBJ to 0.15.10 and Helidon to 4.5.0

### LINKED ISSUE #3056: Ensure PBJ parse max depth set appropriately with next version of PBJ
### Story Form

N/A

### Technical Notes

PBJ is reducing the default maximum depth for parsing as part of ongoing memory usage optimizations.
When we adopt the next version of the PBJ library, we should check everywhere we parse using a PBJ `Codec` and ensure we are deliberately setting a reasonable maximum depth (perhaps 256, certainly not more than 384) so that our parsing code is not dependent on future changes to the library default.
This default value should come from configuration (perhaps ServerConfig?).
Some plugins may also choose to set smaller limits when appropriate for memory efficiency; these smaller limits should be set either to an appropriate constant when depth is known to always be constant, or to a value derived from the configured default.

AlfredoG87:
Fixes https://github.com/hiero-ledger/hiero-block-node/pull/3080

## Summary

Upgrades PBJ to 0.15.10 and Helidon to 4.5.0, and replaces all manual PBJ `.parse()` call sites with a new `ParseHelper.standardParse()` helper. Resolves #3056.

**Why two dependency upgrades are bundled:** PBJ 0.15.10 fixes the binary incompatibility with Helidon 4.5.0 (`Http2ClientStream` constructor gained a parameter), so both upgrades must land together.

**Parse depth regression:** PBJ 0.15.9 silently lowered `Codec.DEFAULT_MAX_DEPTH` from 512 to 128. Block stream data can exceed 128 levels of nesting. All `.parse()` call sites now go through `ParseHelper.standardParse()` which hard-codes `depth=256`, `strict=false`, and `unknownFields=true` (preserves forward compatibility).

**Max size strategy:** `standardParse()` defaults to `Codec.DEFAULT_MAX_SIZE` (2 MB). Call sites that originally used a large explicit max — full-block reads, stream-publisher (configured `maxMessageSize`), stream-subscriber (configured `maxProtobufMessageSizeBytes`) — pass the relevant size via the overload. This preserves each site's original intent while centralizing the depth/strict/unknownFields settings.

### Changes

| What | Where |
|---|---|
| PBJ plugin bumped to `0.15.10` | `settings.gradle.kts` |
| Helidon restored to `4.5.0` | `hiero-dependency-versions/build.gradle.kts` |
| `zstd-jni` bumped to `1.5.7-11` | `hiero-dependency-versions/build.gradle.kts` |
| `ParseHelper` with `standardParse()` overloads (new) | `base/ParseHelper.java` |
| `ParseConstants` removed | `base/ParseConstants.java` deleted |
| `PbjParseMaxDepthTest` documents the regression (new) | `base/PbjParseMaxDepthTest.java` |
| All PBJ `.parse()` call sites updated | ~107 files across block-node, tools CLI, simulator |
| `BlockAccessor` mirrors `MAX_PARSE_DEPTH` as interface constant | `spi-plugins` (no `base` dep — circular) |
| Stream-publisher passes configured `maxMessageSize` | `StreamPublisherPlugin.java` |
| Stream-subscriber passes configured `maxProtobufMessageSizeBytes` | `BlockStreamSubscriberSession.java` |
| Full-block reads pass `Integer.MAX_VALUE` | `BlockFileBlockAccessor`, `ZipBlockAccessor`, `BlockStagingFileAccessor`, `BlockHasher`, `ExtendedMerkleTreeSession`, `BlockItemUtils` |
| Simulator keeps local `MAX_PARSE_DEPTH` | standalone module, no `base` dep |
| `app/testFixtures/module-info.java` — `requires org.hiero.block.node.base` added | testFixtures |

## Notes on `BlockAccessor.MAX_PARSE_DEPTH`

`spi-plugins` (module `org.hiero.block.node.spi`) **cannot** depend on `base` — `base` itself requires `org.hiero.block.node.spi`, which would be a circular module dependency. The `MAX_PARSE_DEPTH = 256` constant is mirrored directly in the interface with a Javadoc reference to `ParseHelper`, so the two stay in sync by convention.

## Test plan

- [x] `./gradlew test` — full suite passes (`facility-messaging`, `s3-archive`, `stream-publisher` are pre-existing timing-sensitive flakes that pass on re-run, unrelated to this change)
- [x] `./gradlew qualityCheck` — passes
- [x] `PbjParseMaxDepthTest` — asserts `DEFAULT_MAX_DEPTH == 128`, depth-130 fails at default, depth-130 succeeds at explicit 256
- [x] TSS Wraps Transition Block (466, ~31 MB) parses successfully with `Integer.MAX_VALUE` size limit


block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -147,9 +146,12 @@ public HashingResult get() {
...
-                                case BLOCK_FOOTER -> this.blockFooter = BlockFooter.PROTOBUF.parse(item.blockFooter());
+                                case BLOCK_FOOTER ->
+                                    this.blockFooter =
+                                            BlockFooter.PROTOBUF.parse(item.blockFooter(), false, MAX_PARSE_DEPTH);
                                 case BLOCK_PROOF -> {
-                                    final BlockProof blockProof = BlockProof.PROTOBUF.parse(item.blockProof());
+                                    final BlockProof blockProof =
+                                            BlockProof.PROTOBUF.parse(item.blockProof(), false, MAX_PARSE_DEPTH);
                                     blockProofs.add(blockProof);
                                 }
```
ata-nas, line 153: Huh, I thought I changed these parse calls to be like the ones below, for example:

```java
 TransactionBody.PROTOBUF.parse(
                    signedTx.bodyBytes().toReadableSequentialData(),
                    false,
                    true,
                    MAX_BLOCK_MESSAGE_DEPTH,
                    Integer.MAX_VALUE);
```
Should these cases also adopt the readable sequential data and then `true` and `false` for the next arguments? This is based on a change @Nana-EC recently did.

ata-nas (commented):
Question for block hasher.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: It's actually more interesting, in that all of the convenience methods just call other methods on the same interface (often more than one deep).
The most efficient call would be as follows:
```java
BlockProof.PROTOBUF.parse(item.blockProof().toReadableSequentialData(), false, true, MAX_PARSE_DEPTH, Codec.DEFAULT_MAX_SIZE);
```

This would also ensure we do not throw unknown fields out (which is important for forward compatibility).

If we wanted a "standardParse" method (as noted elsewhere), this would be even simpler:
```java
                                            StandardParse(BlockProof.PROTOBUF, item.blockProof());
```

block-node/block-access-service/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
```diff
@@ -74,7 +75,8 @@ public Pipeline<? super Bytes> open(
         return switch (blockAccessServiceMethod) {
             case getBlock ->
                 Pipelines.<BlockRequest, BlockResponseUnparsed>unary()
-                        .mapRequest(bytes -> BlockRequest.PROTOBUF.parse(bytes.toReadableSequentialData()))
+                        .mapRequest(bytes ->
+                                BlockRequest.PROTOBUF.parse(bytes.toReadableSequentialData(), false, MAX_PARSE_DEPTH))
```
jsync-swirlds: I see the same call _many_ times in this PR, and I wonder if we should make a convenience method in the `ParseConstants` file (which might rename to `ParseHelper`).
This would then become (assuming static import of `ParseHelper.standardParse`)
```suggestion
                                standardParse(BlockRequest.PROTOBUF, bytes)
```

block-node/base/src/main/java/org/hiero/block/node/base/ParseConstants.java
```diff
@@ -0,0 +1,17 @@
...
+ * All parse call sites must pass this constant explicitly so they are not silently affected
+ * by future changes to the library default (see issue #3056).
+ */
+public final class ParseConstants {
+
+    /** Maximum nesting depth for all PBJ codec parse calls. */
+    public static final int MAX_PARSE_DEPTH = 256;
+
+    private ParseConstants() {}
+}
```
jsync-swirlds: I wonder if a "standard parsing" class would be a better choice here.

```suggestion
public final class ParseHelper {
    /** Maximum nesting depth for all PBJ codec parse calls. */
    private static final int MAX_PARSE_DEPTH = 256;

    private ParseHelper() {}

    private static <T> T standardParse(final Codec<T> parser, final Bytes input) {
        return parser.parse(input.toReadableSequentialData(), false, true, MAX_PARSE_DEPTH, Codec.DEFAULT_MAX_SIZE)
    }

    private static <T> T standardParse(final Codec<T> parser, final Bytes input, final long maxParseBytes) {
        return parser.parse(input.toReadableSequentialData(), false, true, MAX_PARSE_DEPTH, maxParseBytes)
    }

    private static <T> T standardParse(final Codec<T> parser, final ReadableSequentialData input) {
        return parser.parse(input, false, true, MAX_PARSE_DEPTH, Codec.DEFAULT_MAX_SIZ)
    }

    private static <T> T standardParse(final Codec<T> parser, final ReadableSequentialData input, final long maxParseBytes) {
        return parser.parse(input, false, true, MAX_PARSE_DEPTH, maxParseBytes)
    }
}
```

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
AlfredoG87: I think is worth the refactor to do the StandardParse Helper method. so we can centralize all calls in one place, and only change it if needed in other places

Will push that change shortly

block-node/base/src/main/java/org/hiero/block/node/base/ParseConstants.java
AlfredoG87: Done the changes and rebase and resolve conflicts ✅

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
AlfredoG87: Done ✅

block-node/block-access-service/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java
AlfredoG87: ✅  Done!

block-node/base/src/main/java/org/hiero/block/node/base/ParseHelper.java
```diff
@@ -0,0 +1,37 @@
...
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+
+/**
+ * Convenience helpers for PBJ codec parsing.
+ *
+ * <p>PBJ 0.15.9 reduced {@code Codec.DEFAULT_MAX_DEPTH} from 512 to 128, which is insufficient
+ * for block stream data. All parse call sites must use these helpers so the depth and unknown-field
+ * settings are applied consistently and cannot silently regress (see issue #3056).
+ */
+public final class ParseHelper {
```
jsync-swirlds, line 16: I do not think we should allow Integer.MAX_VALUE parse limits in general.
The default from PBJ is mostly correct; and in a few places we already have configured limits (which we should continue to enforce).

We should have overloads here that allow specifying the maximum size (as offered in a prior comment) but, importantly, do not create an extra method call depth.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/historicalblocks/BlockAccessor.java
```diff
@@ -50,7 +46,7 @@ default BlockUnparsed blockUnparsed() {
                 return null;
             }
             return BlockUnparsed.PROTOBUF.parse(
-                    rawData.toReadableSequentialData(), false, true, MAX_BLOCK_MESSAGE_DEPTH, Integer.MAX_VALUE);
+                    rawData.toReadableSequentialData(), false, true, MAX_PARSE_DEPTH, Integer.MAX_VALUE);
```
jsync-swirlds, line 49: We should probably use the standardParse method here too.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -144,14 +145,7 @@ public Pipeline<? super Bytes> open(
...
                 Pipelines.<PublishStreamRequestUnparsed, PublishStreamResponse>bidiStreaming()
-                        .mapRequest(
-                                bytes -> PublishStreamRequestUnparsed.PROTOBUF.parse(
-                                        bytes.toReadableSequentialData(), // input data
-                                        false, // strictMode
-                                        true, // parseUnknownFields
-                                        maxMessageSize / 8,
-                                        maxMessageSize) // maxDepth
-                                )
+                        .mapRequest(bytes -> standardParse(PublishStreamRequestUnparsed.PROTOBUF, bytes))
```
jsync-swirlds: This is one of the places we _must_ use the configured `maxMessageSize`.
A standardParse variant that accepts max size as a parameter would work for this case.

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -459,14 +460,7 @@ private void sendHistoricalBlocks() throws ParseException {
...
                     final int blockByteSize = (int) blockBytes.length();
-                    // maxProtobufMessageSizeBytes / 8: each level of message nesting needs >= ~8 bytes on the wire,
-                    // so size/8 bounds the deepest a non-degenerate message can nest.
-                    final BlockUnparsed block = BlockUnparsed.PROTOBUF.parse(
-                            blockBytes.toReadableSequentialData(),
-                            false,
-                            true,
-                            maxProtobufMessageSizeBytes / 8,
-                            maxProtobufMessageSizeBytes);
+                    final BlockUnparsed block = standardParse(BlockUnparsed.PROTOBUF, blockBytes);
```
jsync-swirlds: Here again, we must adhere to the configured maximum size.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/generator/itemhandler/AbstractBlockItemHandler.java
```diff
@@ -33,7 +35,7 @@ public BlockItem getItem() {
     @Override
     public BlockItemUnparsed unparseBlockItem() throws BlockSimulatorParsingException {
         try {
-            return BlockItemUnparsed.PROTOBUF.parse(Bytes.wrap(getItem().toByteArray()));
+            return BlockItemUnparsed.PROTOBUF.parse(Bytes.wrap(getItem().toByteArray()), false, MAX_PARSE_DEPTH);
         } catch (ParseException e) {
```
jsync-swirlds, line 39: Here too, using the `standardParse` method seems like a better option.

jsync-swirlds (commented):
There is one change in Publisher that needs to be adjusted.

jsync-swirlds (dismissed):

jsync-swirlds (approved):

rockysingh (approved):

PR #3081, opened 2026-06-24
feat: Replace EMB-based routing with gap-triggered temporary archives

### LINKED ISSUE #3045: Replace EMB-based routing with gap-triggered temporary archives
### Story Form

I want to switch to a temporary archive whenever a gap appears in the incoming block stream, and stay on a regular archive for contiguous blocks
So that the plugin has no dependency on `earliestManagedBlock` and handles any gap correctly regardless of where in the stream it occurs

### Background

The implementation introduced by #2987 routes blocks based on `earliestManagedBlock` (EMB): it computes `firstRegularGroupStart` as the first aligned group boundary at or above the EMB and routes all blocks below that boundary to temporary archives. This over-specifies the solution: EMB is not the right signal for choosing between a regular and a temporary archive. The right signal is a gap in the block stream.

If a temporary archive is opened whenever a gap is detected, the plugin can discard EMB awareness entirely. It will store what it can - in regular archives when the stream is contiguous, in temporary archives when it is not, and consolidate into complete archives whenever coverage allows. In many cases a temporary archive will self-complete if the run it captures happens to cover a full aligned group.

### Changes Required

- Remove `firstRegularGroupStart` and all reads of `NodeConfig#earliestManagedBlock()` from `CloudStorageArchivePlugin`.
- Replace the EMB-based routing condition with a continuity check: compare the incoming block number against the last block handed off to any upload task.
    - Contiguous block (no gap): route to the current regular `BlockUploadTask`.
    - Gap detected: hold arriving blocks in a small in-memory buffer (configurable, default ~5 blocks) without notifying them as persisted. If the gap closes before the buffer is exhausted, drain the buffer into the current regular task and continue. If the buffer fills up without the gap closing, flush the buffered blocks to a new `TempArchiveUploadTask`.
- A temporary-archive run closes when the next gap is detected or the run reaches an aligned group boundary. After closing, call `checkGroupCoverage()` as before.
- `StartupRecoveryTask` must initialize the "last archived block" correctly so the first post-recovery block is routed correctly.
- Introduce a new CloudStorageArchiveConfig property for the gap buffer size (e.g. gapBufferSize, default 5).

ivannov:
### Gap-triggered routing replaces EMB-based routing

`firstRegularGroupStart` and all reads of `NodeConfig#earliestManagedBlock()` are removed. Routing is now driven by a single field: `lastHandedOffBlock` (initialized to -1), which tracks the last block number handed off to any upload task (regular or temp). A contiguous arrival (`blockNumber == lastHandedOffBlock + 1`, or `lastHandedOffBlock == -1` on a fresh start) feeds the regular pipeline. Any gap triggers the buffer described below.

### Gap buffer

`gapBuffer` is a `ConcurrentSkipListMap` that accumulates up to `gapBufferSize` (new config property, default 5, range 1-10) blocks when a gap is detected. If the expected block arrives before the buffer fills, `drainGapBufferToRegular()` drains the contiguous prefix into the regular pipeline - the buffer may still hold a residual gap. If the buffer fills first, `flushGapBufferToTemp()` routes the contents: blocks still within the active regular task's remaining range are salvaged to `currentGroupPending` so the task does not stall waiting for them; all other blocks go to a `TempArchiveUploadTask`. On first gap detection, if the target group already has temp data or is two or more groups ahead, the buffer is bypassed and the block goes directly to temp.

## Related Issue(s)

Closes #3045

### DIFF
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
@@ -150,6 +150,19 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
     /// Accessed only from the notification handler thread.
     final Map<Long, Long> tempGroupNextExpected = new ConcurrentHashMap<>();
 
+    /// The first block of the currently active (streaming) temp segment per group.
+    /// Set by [startNewTempSegment]; cleared by [closeActiveTempSegment].
+    /// Accessed only from the notification handler thread.
+    final Map<Long, Long> tempGroupActiveSegmentFirstBlock = new ConcurrentHashMap<>();
+
+    /// Last block number successfully queued to each in-flight temp segment, keyed by the
+    /// segment's firstBlock.  Updated in [routeToTempArchive] each time a block is enqueued.
+    /// Removed when the segment's future completes and moves to [tempArchiveTracker].
+    /// Used by [isBlockCoveredByAnyTempSegment] to distinguish true duplicates from late-arriving
+    /// blocks that fell into an abandoned gap between two temp segments.
+    /// Accessed only from the notification handler thread.
+    final NavigableMap<Long, Long> tempSegmentLastBlock = new ConcurrentSkipListMap<>();
+
     /// In-flight temp archive upload futures keyed by the firstBlock of the uploaded segment.
     /// Accessed only from the notification handler thread.
     final ConcurrentSkipListMap<Long, Future<TempArchiveEntry>> tempUploadFutures = new ConcurrentSkipListMap<>();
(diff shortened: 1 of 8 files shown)


AlfredoG87 (dismissed):
LGTM

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -562,24 +556,145 @@ Callable<TempArchiveEntry> newTempArchiveUploadTask(
...
+                drainPendingToQueue();
+            } else if (currentUploadFuture == null
+                    && blockNumber >= nextBlockToQueue
+                    && !hasAnyTempDataForGroup((blockNumber / groupSize) * groupSize)) {
+                // The regular task completed but temp-archive routing advanced lastHandedOffBlock,
+                // making this block appear retrograde. Start a new regular task for its group.
+                LOGGER.log(TRACE, "Block {0} starting new regular task", blockNumber);
+                routeToCloudArchive(blockNumber, blockWithSource);
+            }
+            // if we reach here, it is truly retrograde/duplicate -- discard silently.
```
jsync-swirlds: Are you discarding the block (something we must not do) or just not adding it to the gap buffer?

Any block that doesn't go to the gap buffer or directly to a current upload task or temp archive task must go to a new temp archive task.

Principle: no blocks are ever discarded, every block is uploaded _somewhere_.

Nana-EC (dismissed):

Nana-EC (approved):

AlfredoG87 (approved):
LGTM

PR #3086, opened 2026-06-25
fix(solo-e2e): give Block Node a realistic heap to prevent OOM during streaming of Oversized Wraps first Block

AlfredoG87:
## Summary

The solo-e2e Block Node was crashing with `java.lang.OutOfMemoryError: Java heap space` after roughly 428 blocks of live streaming. The packaged `block-node-server` chart that `solo block node add` pulls ships placeholder JVM limits — `JAVA_OPTS: -Xmx24m` and `resources.limits.memory: 2Gi` — so the node ran with a 24 MB heap. This adds a memory-override values overlay **(16 GB heap, 20Gi container limit)** and applies it on every Block Node deploy.

## Root cause

Captured live from the Block Node's own logs at the moment of failure:

```
Wrote verified block 427 ... 0000000000000000427.blk.zstd
Started new block verification session for block number 428
java.lang.OutOfMemoryError: Java heap space
Dumping heap to /tmp/dump.hprof ...
```

`helm get values block-node-1` confirmed the deployed release ran with `JAVA_OPTS: -Xmx24m -XX:MaxDirectMemorySize=8m -XX:MaxMetaspaceSize=32m` and `limits.memory: 2Gi`, while the consensus node stayed healthy and kept producing blocks. The solo profile (`custom-spec.yaml`) has no block-node section, so nothing raised these chart placeholder values.

## Changes

**`tools-and-tests/scripts/solo-e2e-test/overrides/bn-memory.yaml`** (new)
- Sets `resources` to requests 8Gi / 2 CPU, limits 20Gi / 4 CPU.
- Sets `blockNode.config.JAVA_OPTS` to `-Xms8G -Xmx16G -XX:MaxDirectMemorySize=2G -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath="/tmp/dump.hprof"`.

**`tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-network.sh`**
- `deploy_block_nodes`: appends the overlay as the final `-f` values-file in the `solo block node add` invocation for every Block Node, so it overrides the chart defaults.

## Test Plan

- [x] `task up` (default `single` topology) deploys with the override applied — pod env shows `JAVA_OPTS=-Xms8G -Xmx16G` and `limits.memory: 20Gi`.
- [x] Block Node streams past the previous failure point with no OOM — confirmed `lastAvailableBlock` climbing cleanly through 428/431 to 701.
- [x] Multi-BN topology (e.g. `fan-out-3cn-2bn`) applies the override to every Block Node.

### DIFF
--- a/tools-and-tests/scripts/solo-e2e-test/overrides/bn-memory.yaml
+++ b/tools-and-tests/scripts/solo-e2e-test/overrides/bn-memory.yaml
@@ -1,0 +1,22 @@
+# Block Node memory override for solo-e2e local testing.
+#
+# The packaged block-node-server chart fetched by `solo block node add`
+# (e.g. 0.37.0-SNAPSHOT) ships tiny placeholder JVM limits
+# (JAVA_OPTS: -Xmx24m, resources.limits.memory: 2Gi). Under sustained live
+# streaming the heap exhausts and the Block Node throws OutOfMemoryError
+# around block ~428, killing the gRPC listener. This overlay restores a
+# realistic heap so the node can run indefinitely.
+#
+# Applied as the final `-f` values-file in solo-deploy-network.sh, so it
+# overrides the chart defaults.
+resources:
+  requests:
+    cpu: "2"
+    memory: "8Gi"
+  limits:
+    cpu: "4"
+    memory: "20Gi"
+
+blockNode:
+  config:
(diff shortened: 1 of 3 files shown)


Nana-EC (approved):

rockysingh (approved):
lgtm

PR #913, opened 2025-03-27
docs: Create MAINTAINERS.md

hendrikebbers:
Maintainers file added as described at https://lf-decentralized-trust.github.io/governance/governing-documents/MAINTAINERS-file.html

## Related Issue(s)
 Use keywords `Fix, Fixes, Fixed, Close, Closes, Closed, Resolve, Resolves, Resolved`
to connect an issue to be closed by this pull request.

Issue #912, opened 2025-03-27
Block Streaming to Consumer Test Plan


Nana-EC:
### Story Form

As a block node developer
I want to have a reference test plan that explores the block streaming of block items to a consumer
So that I have confidence of coverage and understanding for the flow of block items out of a BN

### Technical Notes

Test Plan should explicitly cover
- BN -> MN or any gRPC client
- BN -> BN in client mode

PR #3087, opened 2026-06-26
chore: Add E2E node operator deployment and operations check

Nana-EC:
## Reviewer Notes

An end-to-end test that exercises a standalone Block Node through **solo-provisioner** (the
`solo-weaver` CLI) on a **kind** cluster — the full operator lifecycle, with real blocks replayed
from fixtures and real TSS verification. No consensus node is involved; it focuses purely on the
operator data path.

```
install (base) → push 0–2 → verify → upgrade (snapshot) → verify + push 3–4
   → reset → re-push 0–4 → uninstall
```

It proves what an operator cares about: blocks and TSS state **survive an upgrade**, `reset` returns
the node to a clean state, and `uninstall` tears everything down.

## Changes (CI / test tooling only)

| File | Purpose |
|------|---------|
| `.github/workflows/e2e-node-operator.yaml` | The lifecycle workflow. Triggers: `pull_request` (relevant paths), nightly `schedule`, `workflow_dispatch` (base/snapshot/profile/provisioner-version inputs). |
| `tools-and-tests/scripts/e2e-node-operator/solo-ci-config.yaml` | solo-provisioner `--config` (profile `local`, `/tmp/bn/*` storage, chart ref). |
| `…/solo-ci-values.yaml` | Helm `--values` for kind (image, resources, init containers, PVC sizes). |
| `…/assert-server-status.sh` | grpcurl assertions: `serverStatus`, `getBlock`, `subscribe`. |
| `…/bn-kind-port-forward.sh` | Waits for pod Ready, (re)establishes the port-forward after each restart. |
| `tools-and-tests/suites/.../e2e/BlockNodeE2EClient.java` + `suites/build.gradle.kts` | Reusable gRPC client (push `.blk` fixtures + subscribe/assert) packaged as a `blockPusherJar` fat-jar. |
| `block-node/app/build.gradle.kts` | `createDockerImage` now `dependsOn(prepareDockerPlugins)` so the solo-dev image bakes plugins (previously done ad-hoc in CI). |

**No product code changes** beyond that one build-graph line.

## Key choices & workarounds (for reviewers)

These are deliberate and commented in place; each links an upstream ask where relevant.

| Choice / workaround | Why |
|---------------------|-----|
| **solo-provisioner pinned to `v0.21.0`** | v0.22.0+ `install` unconditionally installs a tc-egress **systemd** service that fails on the systemd-less CI runner. Upstream: `hashgraph/solo-weaver#915`. |
| **Manual block-store clear after `reset`** | On v0.21.0, `reset` doesn't purge storage on the `local` profile; the step wipes block data + restarts the pod to reach the empty state the assert expects. Removable once #915 lets us move to a version whose `reset` clears. |
| **solo-dev image + `plugins.names: ""` + `seed-plugins` init container** (copy + clear + profile-prune) | The default image resolves plugins via a Maven init container (~4 min cold) that blows the provisioner's 5-min install timeout (`#912`); and the provisioner force-mounts a managed plugins PVC read-only over the image's baked plugins (`#913`), so we seed that PVC from the baked image and prune to the selected profile. |
| **cloud-provider-kind + `--load-balancer-enabled=false`** | solo-provisioner makes the BN Service a LoadBalancer; kind has no LB controller, so `helm --wait` hangs on a `<pending>` external IP. cloud-provider-kind assigns IPs; the flag keeps the service ClusterIP-friendly. |
| **`crane` to load the base image; local OCI registry for the snapshot chart** | The published base image is multi-arch and `kind load docker-image` chokes on the manifest list; crane pulls a single-platform archive. solo-provisioner won't take a local chart path, so the built snapshot chart is pushed to a localhost registry. |
| **Readiness gated by `kubectl wait --for=condition=Ready`** | The BN serves health on a dedicated port (`40983`); gating on the k8s readiness probe is port-agnostic (rather than curling a hardcoded port). |
| **Base = latest published GA/RC (auto); snapshot = `version.txt`** | The upgrade must go base → snapshot; this holds as long as `version.txt` stays above the latest release. Both are assumed 0.37+ (stable `application-state` volume), so the upgrade is a plain `helm upgrade` on chart defaults. |

## Testing

Green end-to-end on the Linux + kind runner (latest: run `29956826753`) — every phase from install
through uninstall passes, including TSS verification surviving the upgrade. The job runs on PRs that
touch the relevant paths and nightly.

## Upstream follow-ups filed

`hashgraph/solo-weaver#912` (configurable helm timeout), `#913` (opt out of the managed plugins PVC),
`#915` (skip tc-egress on systemd-less hosts) — landing any of these lets us drop the corresponding
workaround above and move to a newer provisioner.

## CI failure note (runs 30284247718 — failed twice, not a test bug)

Two recent CI runs failed at `BlockNodeE2EClient: Timed out waiting for 3 acknowledgement(s)`.
Root cause: at the time those runs executed, `0.39.0-rc3` was the latest published release and was
resolved as the base image. rc3 had two compounding problems both absent in `0.39.0` GA:

1. **Plugin name mismatch.** PR #3253 renamed the verification plugin from `verification` to
   `block-verification` on `main`; that rename was cherry-picked into `release/0.39` only _after_
   rc3 was tagged. The seed-plugins prune loop whitelists names from the current branch's
   `plugin-profile-lfh.yaml` (which already has `block-verification`), so it pruned the
   `verification` jar from rc3, leaving the server without a verification plugin.

2. **Non-thread-safe native TSS library.** The same PR (#3253) introduced `NativeVerificationLibrary`
   — a `ReentrantLock` wrapper around the hinTS/WRAPS native singleton. Its javadoc states:
   _"empirically even two concurrent `verifyTSS` calls corrupt the library's shared native state,
   producing sporadic, load-dependent `BAD_BLOCK_PROOF` failures."_ rc3 lacked the lock.

Both problems are structural to the 0.39.x RC cycle: `release/0.39` was branched before #3253
merged to `main`. Future cycles branch from `main` _after_ #3253, so this will not recur.

With `0.39.0` GA now published, the version resolver picks it as base and the test runs cleanly as
`0.39.0 → 0.40.0-SNAPSHOT`.


tools-and-tests/suites/build.gradle.kts
jsync-swirlds, line 1: This (build change) is, in my opinion, going to be, ultimately, self-defeating.
We're trying to make E2E tests easier by making sure they never run in an environment that looks like production.
This means we never see plugin related bugs, but it also means we never _catch_ plugin related bugs.

berryware (dismissed):
Looks good

tools-and-tests/suites/build.gradle.kts
```diff
@@ -105,6 +103,58 @@ val prepareTestPlugins by
...
+
+    // The compiled classes of this module plus the full runtime classpath, flattened.
+    from(sourceSets["main"].output)
+    inputs.files(blockPusherRuntimeJars)
+    from({ blockPusherRuntimeJars.filter { it.name.endsWith(".jar") }.map { zipTree(it) } })
+
+    // Drop signature files from signed dependency jars so the flattened jar remains valid.
+    exclude("META-INF/*.SF", "META-INF/*.DSA", "META-INF/*.RSA", "META-INF/*.EC")
+}
+
```
jsync-swirlds, line 153: There is support in the hiero plugin for this now; we should use that instead.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeE2EClient.java
jsync-swirlds, line 1: Would it not be better to incorporate this into the existing picocli-based CLI tool?

jsync-swirlds (commented):
A couple of high-level concerns here.
Also, the fat-jar can be done with hiero-plugin support now, so if we really believe a fat-jar is appropriate, we should use that.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeE2EClient.java
Nana-EC: Yes, but the intent with this PR was to make things self contained and share after proven resiliency. I wanted to see a few runs after merge before sharing. Intent was to update E2E tests to also use this

tools-and-tests/suites/build.gradle.kts
Nana-EC: @jsync-swirlds the comment is a bit vague, I can't tell which part of the changes you're referring to. This file contains the fat-jar creation and forced spotless changes.
Can you clarify the issue you're referring to? Why would we not see plugin related bugs?
The BN runs real plugins

jsync-swirlds: The fat jar is the core concern, we, from what I can tell, are trying to avoid issues with loading plugins by packaging them all in a single jar, but that makes the testing run in an environment that doesn't function the way production does _because_ it doesn't load plugins.

Rather than changing packaging to hide those issues, I would think we want to see those plugin loading issues as early as possible.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeE2EClient.java
jsync-swirlds: Apparently I'm a little confused the entire purpose of this PR.

It _seems like_ we're finding testing hard so we're making the testing environment unique (and very unlike production) in order to make that easier; which also makes the testing less valuable precisely because the environment no longer matches production...

tools-and-tests/suites/build.gradle.kts
Nana-EC: @jsync-swirlds please check one more time. The fat jar has nothing to do with avoiding loading plugins or loading all plugins to a single jar. The fat jar logic bundles BlockNodeE2EClient and needed dependencies to make it easy to push a set of blocks in a test.
Are you mixing this up with the solo-dev image that's used in the operator run or something else.
No existing build logic is modified. This only adds logic for the test logic specifically.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeE2EClient.java
Nana-EC: The PR simple simulates the full cycle of operations a node operator will face.
Not sure the confusion here. There were some workarounds I had to do in the 0.36-0.37 upgrade due to the breaking change that we had, I've since reverted those but if there's anything left over that relating to this let me know. There was also an issue with provisioner not supporting plugin selection which i had to work around to get the rsa plugin in there - hence the use of the solo-dev machine to pull from.
Beyond that it should be pretty straight forward and I don't agree on the less value in testing, this actually unearthed what 3rd party operators who want to customize their deployments beyond our initial recommendations would face.

tools-and-tests/suites/build.gradle.kts
Nana-EC: Will take a look

tools-and-tests/suites/build.gradle.kts
Nana-EC: https://github.com/hiero-ledger/hiero-block-node/pull/3087/changes#r3599710291
I see this comment. Are you potentially mixing up the plugin selection from the initContainers in tools-and-tests/scripts/e2e-node-operator/solo-ci-values.yaml?
If so i've shared before the non-optimal nature of relying on plugins published to maven and I believe pulling from the image with all plugins was a path noted to work around that.
In my case as others may find I experienced timeouts pulling the snapshot plugins images from maven, this would result in flaky tests where runs would fail.

A few of these cases are workarounds and have some follow ups from BN team and potentially provisioner to improve the experiences but those are less timely than ensuring we have a good base of migration like testing that covers the node operators experiences and ensuring coverage early on in the development and release cycle.

I'll take a look to see what I can simplify or pull out now but let me know what you consider blocking feedback and what's improvements so i can tackle them appropriately.
I'd like to get runs on this going in 0.39 to ensure we're catching regressions early

jsync-swirlds: I'm mostly concerned we're creating an environment for testing with this change that doesn't reflect actual runtime.  If you're confident that this isn't the case, then this concern is non-blocking and we can revisit later.

We can examine simplifying options for this process later.

tools-and-tests/suites/build.gradle.kts
jsync-swirlds: There is a slack message I tagged you in 13 days ago.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeE2EClient.java
jsync-swirlds: I tend to think we should test as close to production as practical.
I also think we should be working with solo (specifically solo-provisioner, since that's what's intended for production deployments) to avoid the need for workarounds and customizations.

Those can be later.

I find these LLM-generated extra detailed PR descriptions hard to follow accurately; and that's probably a major source of confusion in this case.

AlfredoG87 (dismissed):
Great Work! 💯

berryware (dismissed):
Looks good

tools-and-tests/suites/build.gradle.kts
```diff
@@ -26,6 +26,7 @@ mainModuleInfo {
     runtimeOnly("org.hiero.block.node.stream.publisher")
     runtimeOnly("org.hiero.block.node.stream.subscriber")
     runtimeOnly("org.hiero.block.node.tss.boostrap")
+    runtimeOnly("org.hiero.block.node.verification")
```
jsync-swirlds: This re-adds an obsolete module, probably shouldn't add that.

Nana-EC: Left over from previous sprints, removed along with references to it

jsync-swirlds (dismissed):

tools-and-tests/scripts/e2e-node-operator/solo-ci-values.yaml
```diff
@@ -33,6 +33,9 @@ blockNode:
     # block 0 with the embedded TSS bootstrap. JAVA_OPTS sized for the kind envelope; SERVER_PORT
     # inherits the chart default.
     JAVA_OPTS: "-Xms1G -Xmx2G -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/dump.hprof"
+  logs:
+    loggingProperties:
+      org.hiero.block.level: "FINEST"
```
jsync-swirlds: We should probably be careful with this, it will crater performance during tests and write many gigabytes of trace data.
`FINE` is the typical "debug" setting, with `FINEST` reserved for specific cases.

Nana-EC: You caught me mid troubleshooting, it's a temp change to understand some failures I'm observing.
Recent changes in the latest BN versions have the CI failing in places it wasn't before.
Let me move the PR to draft whiles I investigate.

Nana-EC: Reverted, this helped me find the verification race condition noted

tools-and-tests/tools/src/test/java/org/hiero/block/tools/blocks/BlocksCommandTest.java
```diff
@@ -35,6 +35,7 @@ void help_printsUsageHelp() {
         StringWriter out = new StringWriter();
         CommandLine cmd = new CommandLine(new BlocksCommand());
         cmd.setOut(new PrintWriter(out));
+        cmd.setErr(new PrintWriter(new StringWriter()));
```
jsync-swirlds, line 38: This seems like it's hiding errors.
Shouldn't we capture the output (assign the string writer) and verify it's not containing some significant failure report?

AlfredoG87 (approved):

tools-and-tests/tools/src/test/java/org/hiero/block/tools/blocks/BlocksCommandTest.java
Nana-EC: Seems consistent with the other test forms.
`noSubcommand_printsUsageHelp()` above breaks it out into another variable but is also passing it through.
@rockysingh is the desired approach, if not we'd need to change it across multiple tests

jsync-swirlds: Not worth blocking this PR, just something we might want to note and fix more thoroughly later.

jsync-swirlds (approved):
There is a lot of dense code here.
I gave a surface scan, but do not have time for a deep inspection.
Approving to unblock.

PR #3102, opened 2026-06-29
chore: Verification Rework Follow Up

### LINKED ISSUE #3007: Verification Rework Follow Up

### LINKED ISSUE #3157: Fix State Proof Verification
### Technical Notes
There are two items in the reworked State Proof implementation that need modest changes.
1. The Timestamp leaf at top-of-tree is a _separate MerklePath_ so we need to find and follow that MerklePath in order to produce the signed Block Root (it is the left-sibling of the endpoint of the other path(s) we follow from block root hash up).
1. Any given entry in the MerklePath list _might_ be a single-child internal node.  This will be represented as a MerklePath entry with no content and no siblings.
   - A single-child internal node is computed by taking the hash computed from the previous MerklePath entry and hashing that alone with a prefix of `01` (for 1 child).

We need to add these two cases to the current updated State Proof type Block Proof verification code.

ata-nas:
* Added more tests
* Fixed state proof verification logic
* Resolve outstanding todos
* Removing all blocks hasher to simplify the codebase
* Added additional branches to hasher for item types and a default branch to be tackled by #3195 
* Remove "em-dashes"

Resolves #3007
Resolves #3157


block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
```diff
@@ -178,38 +178,36 @@ private VerificationNotification processCompletionException(final CompletionExce
...
+        final long verifiedBlockNumber = verificationResult.blockNumber();
         final VerificationNotification notification = new VerificationNotification(
                 true,
                 null,
-                verificationResult.blockNumber(),
+                verifiedBlockNumber,
                 verificationResult.rootHash(),
                 verificationResult.block(),
                 verificationResult.source());
         safeSendNotification(notification);
```
ata-nas, line 215: Maybe unsuccessful sending of notification should be a sign we are unhealthy. We can throw here.

jsync-swirlds: Not throw, but perhaps return false from `safeSendNotification` if it failed, and set the plugin "unhealthy".
Won't do much at the moment, but when we add proper plugin and node health monitoring it will inform the Health Plugin and Status Plugin (if present) to report the system health properly.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/AllBlocksHasherHandler.java
jsync-swirlds, line 1: I wonder if we could just remove this class in the new plugin...

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds, line 188: Just a reminder to add the hashing of the remaining known block item types and `@todo` for field-number based hashing of `UnknowndFields` here.

jsync-swirlds (commented):
Getting close.
Thanks for hanging in with all the changes.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: If we have an `UNSET` value, I believe we should fail as unable to parse. I think we should add that as well.

jsync-swirlds: Agreed.  That should fail to parse anyway, but no harm adding it to the switch.
One caveat, an unknown (forward compatible) field may _also_ be `UNSET` in the enum (because we don't know the correct schema), so for safety in `UNSET` instead of always failing, we should check for anything in unknown fields.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -184,8 +193,6 @@ private boolean isCanceled() {
     private HashingResult finalHashingResult() throws NoSuchAlgorithmException {
         final HashingResult hashingResult;
         if (blockHeader == null || blockFooter == null || blockProofs.isEmpty()) {
-            // todo(2528) validate that when we see an WRB we have:
-            //    1x Block Header, 1x RECORD_ITEM, 1x Block Footer, N number (at least 1) Block Proof
```
jsync-swirlds, line 188: Do we have this check?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -7,130 +7,144 @@
...
-                                LOGGER.log(
-                                        WARNING,
-                                        "Block {0} state proof integrity check failed: hash reconstructed from path 1 siblings"
-                                                + " [{1}] does not match block root hash computed from block content [{2}]",
-                                        blockNumber,
-                                        reconstructed,
-                                        rootHash);
-                                proofVerificationMetrics.stateProofFailure().increment();
+                    final MerklePath path = paths.get(leafStartingIndex);
+                    if (isLeaf(path)) {
```
jsync-swirlds: This should check both is leaf _and_ not visited.
It shouldn't happen (the graph should be acyclic), but it's a cheap check here.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -7,130 +7,144 @@
...
+                                    case STATE_ITEM_LEAF ->
+                                        hashLeaf(path.stateItemLeaf().toByteArray());
+                                    case BLOCK_ITEM_LEAF ->
+                                        hashLeaf(path.blockItemLeaf().toByteArray());
+                                    case TIMESTAMP_LEAF ->
+                                        hashLeaf(path.timestampLeaf().toByteArray());
+                                };
+                        pathResult = mergeSiblings(path, pathResult);
+                        // Now we must follow the next index, which must always be a join point, and
+                        // we keep following that until we can
```
jsync-swirlds:
```suggestion
                        // we keep following that while we can
```

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -7,130 +7,144 @@
...
+                        boolean canFollowNextIndex = true;
+                        int currentJoinPointIndex = path.nextPathIndex();
+                        while (canFollowNextIndex) {
```
jsync-swirlds: This might read better as do-while...

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -7,130 +7,144 @@
...
+                                        }
+                                        yield hash;
+                                    }
+                                    case STATE_ITEM_LEAF ->
+                                        hashLeaf(path.stateItemLeaf().toByteArray());
+                                    case BLOCK_ITEM_LEAF ->
+                                        hashLeaf(path.blockItemLeaf().toByteArray());
+                                    case TIMESTAMP_LEAF ->
+                                        hashLeaf(path.timestampLeaf().toByteArray());
+                                };
```
jsync-swirlds: This can be a method --> `computePathResult(path)`

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -7,130 +7,144 @@
...
+                                    } else {
+                                        checkpoints.put(
+                                                currentJoinPointIndex,
+                                                new MergeCheckpoint(leafStartingIndex, pathResult));
+                                        canFollowNextIndex = false;
+                                    }
+                                    currentJoinPointIndex = nextPath.nextPathIndex();
+                                } else {
+                                    return SessionFailureType.BAD_BLOCK_PROOF;
+                                }
```
jsync-swirlds: This (with the while loop) can be a method --> followPathUntilJoin(path, checkpoints)

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -7,130 +7,144 @@
...
+                if (stateProof.hasSignedBlockProof()) {
                     final TSSVerifier tssVerifier = new TSSVerifier(
                             proofVerificationMetrics,
                             Bytes.wrap(signedBlockRoot),
                             stateProof.signedBlockProof().blockSignature(),
                             verificationDataProvider);
                     result = tssVerifier.verify();
+                } else {
+                    result = SessionFailureType.BAD_BLOCK_PROOF;
                 }
```
jsync-swirlds: Another method --> completeVerification(signedBlockRoot, foundComputedRootHashMatch, stateProof)

jsync-swirlds (commented):
Needs another pass or two to know if I really followed it, but a few suggestions for now.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: I am not sure I understood what you are asking here.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: No, if you remember, we decided to drop that as we check for a header, footer and a proof present.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: ✅ Done.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: Not a fan of do-whiles. I would say it does fit here.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: ✅ Done.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: It can be, the reason it is here is because I want to have the ability to return a failure type in the middle. I will lose that if I now have to return another result instead.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: ✅ Done. Good call, extracted.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/AllBlocksHasherHandler.java
ata-nas: ✅ Done, removed.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: So, we only visit the next leaf always, because of the for loop, but if we have to follow a next index, we only do so if that next index is not a leaf, but a join point. If we have a cyclic graph in the join points, this will result in a merge, that will be wrong and eventually fail.

Let's examine this together later.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: I think I was overthinking the UNSET case, actually (having had a bit more time to think about it).

ata-nas: Okay, so simply fail verification on UNSET should be what we need to do? Just to confirm.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: For WRB _only_ we do need to ensure it is exactly one RecordFileItem too.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
jsync-swirlds: I agree, just suggesting the check for visited as well; it's not expensive and makes the intent clearer in the code.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Ok, we can add that check.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
jsync-swirlds: I also find do-while a rarely valuable option, but (as noted) it seems to be a better fit in this specific case.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
jsync-swirlds: The new method can return a failure type (or null if no failure).
We get a 4 line conditional instead of a 30 line loop.
Much easier to read through, in my opinion.

non-blocking as it can easily be done later if preferred.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: Yeah, does that check also mean that we have to keep track of visited ones? Cause using the iteration's `i` value seems a bit redundant.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: Let me consider it.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: True, but I have to somehow have the updated value present. Let me explore more, maybe I am missing something. In order to do this, I have to save some values as fields.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Added a check for WRB, if one is detected, that the second item is indeed record file. See if that makes sense.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: Okay, have made changes. The signed block root variable is now a field, however, as both the `for i` loop and the `do-while` loop need this value and they are now not within the same scope. We do achieve, however, a more readable `verify` method. Let me know what you think.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: I've switched to `do-while`, let me know what you think.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: @jsync-swirlds what about this one?

jsync-swirlds: I expected we'd both iterate by index and also track visited because we follow each chain when we reach a leaf, so we can encounter previously visited paths while iterating.

Checking visited lets us skip the leaf check, which might be faster.
It's not critical; if we want to add later we can do so easily.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: I was not clear before, and for that I apologize.

1. We should not just check that the second item is a record item (that's risky and might not be true in all cases).
2. What we need to do is ensure we do not get _2 or more_ record file items (that is, if it's present, there is only one).

That's the check we must make here, that record file items are singular ("lone wolf" check) within a block.

ata-nas: Okay, what I am thinking then is that if we hit the switch case for record file more than once, we fail. That seems to me the most performant check we can possibly make.

BTW, we can also add this check for header and footer as well. We only expect to have a single header and a single footer ever. Is that correct? Or is it more nuanced, considering possible forward compatibility default branch?

jsync-swirlds: I think it's reasonable to check for single header and single footer as well.

The approach you recommended is a good one.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
ata-nas: Left a todo since we do not have health plugin updates yet.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Checks added along with a parameterized test. Check it out.

docs/block-node/configuration.md
```diff
@@ -297,6 +297,21 @@ for the JSON schema.
...
+> _NOTE_ even if the `VERIFICATION_ALL_SOURCES_REQUIRE_ORDERING` value is set to false, strict ordering will always
+> be imposed for the `PUBLISHER` source. In that scenario, it is possible that gaps can happen, because sources,
+> other than `PUBLISHER`, can supply a valid block, much higher than last verified. This concern is understood and
+> accepted.
+
+### Verification Plugin Configuration (DEPRECATED)
+
 | ENV Variable                                        | Description                                                                    |                                                                 Default |
 |:----------------------------------------------------|:-------------------------------------------------------------------------------|------------------------------------------------------------------------:|
 | VERIFICATION_ALL_BLOCKS_HASHER_ENABLED              | Enable the all-blocks hasher to compute and verify a rolling root hash.        |                                                                   false |
```
berryware, line 320: These are removed from VerificationConfig. That is more than deprecated. Setting these are no-ops. Perhaps these should be removed or more clear.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -173,14 +196,22 @@ public void onContextUpdate(final BlockNodeContext updatedContext) {
...
+                            blockItems.blockNumber(), source, SessionFailureType.MISSING_VERIFICATION_DATA);
+                }
             } else {
-                LOGGER.log(WARNING, "Received null block items on live items ring buffer");
+                LOGGER.log(INFO, "Received null block items on live items ring buffer");
             }
-        } catch (final Exception e) {
-            LOGGER.log(WARNING, "Failed to handle live block items in verification ", e);
+        } catch (final RuntimeException e) {
+            LOGGER.log(INFO, "Failed to handle live block items in verification ", e);
```
berryware, line 212: Should a RuntimeException be an INFO and not a instead of WARNING? Is it expected to happen often?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
```diff
@@ -26,29 +28,31 @@
...
 
     /// Constructor.
     public SessionResultHandler(
             final BlockNodeContext context,
+            final VerificationConfig verificationConfig,
             final SessionResultMetrics sessionResultMetrics,
             final BadBlockDumper badBlockDumper,
             final AtomicLong lastVerifiedBlock,
-            final ConcurrentSkipListSet<Long> recentlyVerifiedBlocks,
+            final ConcurrentLinkedDeque<Long> recentlyVerifiedBlocks,
```
berryware, line 49: Will this introduce duplicates into the processing? The set would have filtered them out?

ata-nas: That is an interesting observation. And yes, duplicates will be present.
The thing about recency is that we only care about which were our actual latest 100 verified blocks.
With a set, we will not have duplicates, but also, we will be unable to identify what to delete when this is full. Duplicates should be fine, because they actually reflect the reality.
I think queue is the better data structure here.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: So, following our logging guidelines, ERROR is a catastrophic event, WARNING is this should not happen, an operator must take an action now, INFO is an operator is notified of an event. I think in this case INFO is appropriate, but maybe WARNING could also be appropriate.

docs/block-node/configuration.md
ata-nas: You are looking at the deprecated config, I keep both.
But I have made more changes to the new ones, adding block dumper.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: Is that code not pushed?

ata-nas: It is, look at the switch branches where we determine item type. 
What I have forgotten is to remove the valid wrb method and check in the finalize method. Removing now.

ata-nas: Ok, that method is removed and the if statement below the todo which is now also gone is restored.
Checks for more than one of these mandatory items happen on the switch cases. Checks for having the mandatory items remain the same.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
```diff
@@ -97,67 +99,72 @@ public SessionFailureType verify() {
...
+                            // cryptographic verification means the block or proof has been tampered with.
+                            LOGGER.log(
+                                    DEBUG,
+                                    "RSA signature from node {0} failed verification in block {1} - rejecting block",
+                                    nodeId,
+                                    blockNumber);
+                            proofVerificationMetrics.rsaFailure().increment();
+                            return SessionFailureType.BAD_BLOCK_PROOF;
+                        }
+                    } catch (final InvalidKeyException | SignatureException e) {
```
rockysingh, line 148: Real bug: `InvalidKeyException` is thrown by `Signature.initVerify(PublicKey)` when the *local* key material is unsuitable for the algorithm — it's a local roster/config problem, not evidence of tampering. Coalescing it with `SignatureException` here means a single malformed RSA public key in the roster will silently reject every otherwise-valid block from that node as `BAD_BLOCK_PROOF`. Please split the catches: `InvalidKeyException` should surface as `MISSING_VERIFICATION_DATA` (or propagate), and only `SignatureException` should imply block-level rejection.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -7,146 +7,254 @@
...
+    }
+
+    /// Compute the content of a leaf.
+    private byte[] computeLeafContent(final MerklePath leaf) {
+        return switch (leaf.content().kind()) {
+            case UNSET -> throw new IllegalArgumentException("Unexpected MerklePath content kind");
+            case HASH -> {
+                final byte[] hash = leaf.hash().toByteArray();
+                if (!foundComputedRootHashMatch) {
+                    foundComputedRootHashMatch = Arrays.equals(rootHash.toByteArray(), hash);
```
rockysingh, line 126: Correctness question: `foundComputedRootHashMatch` fires on *any* HASH-typed leaf anywhere in `paths` whose bytes equal `rootHash`, and `completeVerification` at L187 accepts on `signedBlockRoot != null && foundComputedRootHashMatch`. Nothing here ties the matching HASH leaf to the subtree traversal that produced `signedBlockRoot`. If the on-wire proof format guarantees that a HASH leaf equal to `rootHash` can *only* appear as the anchor of the reconstructed subtree, this is fine — please confirm and add a comment noting the invariant. Otherwise a malformed proof with an unrelated HASH leaf plus a `signedBlockRoot` reconstructed from a different subtree would still pass.

rockysingh (commented):
Big refactor and cleanup good work! The general direction is good (removing `AllBlocksHasherHandler`, generalising `StateProofVerifier` to walk arbitrary DAG-shaped merkle paths, cleaner cancellation plumbing across the verifier stack, and materially better test coverage). Six findings inline: one bug (RSA InvalidKeyException), two correctness concerns worth sanity-checking with someone who owns the proof format, one behavioural regression (Set → Deque), and two nits. Skipped a couple of speculative findings after checking the code more carefully.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -142,15 +150,42 @@ public HashingResult get() {
...
+                                }
+                                case UNSET -> {
+                                    throw new VerificationSessionFailedException(
+                                            blockNumber, SessionFailureType.UNABLE_TO_PARSE, blockSource);
+                                }
+                                case null -> {
+                                    throw new VerificationSessionFailedException(
+                                            blockNumber, SessionFailureType.UNABLE_TO_PARSE, blockSource);
+                                }
+                                default -> {
```
rockysingh, line 183: Silent `default -> {}` in a hasher is a footgun. If a future CN adds a new `ItemOneOfType` (protobuf forward-compat is the norm), this code will happily produce a root hash that doesn't cover the new items — proof verification will then fail with a confusing signal, or worse silently accept a hash missing data. Until #3195 lands, please either fail loudly (matching `REDACTED_ITEM/FILTERED_SINGLE_ITEM` → `UNABLE_TO_PARSE`) or at minimum log a WARN and increment a metric so operators notice. Leaving an empty default in a security-critical hasher during a rework is exactly the kind of thing that gets forgotten.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -61,7 +72,7 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
     /// Constructor.
     public VerificationServicePlugin() {
         this.lastVerifiedBlock = new AtomicLong(-1);
-        this.recentlyVerifiedBlocks = new ConcurrentSkipListSet<>();
+        this.recentlyVerifiedBlocks = new ConcurrentLinkedDeque<>();
```
rockysingh, line 75: Behavioural regression: `recentlyVerifiedBlocks` changed from `ConcurrentSkipListSet<Long>` (main) to `ConcurrentLinkedDeque<Long>` here. Consequences: (1) `markRecentlyVerified` at `SessionResultHandler.java:235` uses `.offer()` unconditionally, so re-verifying the same block adds a duplicate entry; (2) `recentlyVerifiedBlocks.remove(notification.blockNumber())` at L272 is `Deque.remove(Object)` which removes only the *first* occurrence, so duplicates linger until buffer overflow drains them via `pollFirst()`; (3) `contains()` degrades from O(log n) to O(n). If duplicates are impossible by construction, revert to the Set. If they aren't, either dedupe on insert or drain all matching entries on remove.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -7,146 +7,254 @@
...
-                    }
-                    final byte[] depth5Node1 = combineSibling(current, siblings.get(index));
-                    final byte[] depth4Node1 = combineSibling(depth5Node1, siblings.get(index + 1));
-                    final byte[] depth3Node1 = combineSibling(depth4Node1, siblings.get(index + 2));
-                    final byte[] depth2Node2 = hashInternalNodeSingleChild(depth3Node1);
-                    final byte[] hashedTimestampLeaf =
-                            hashLeaf(timestampPath.timestampLeaf().toByteArray());
-                    final byte[] signedBlockRoot = hashInternalNode(hashedTimestampLeaf, depth2Node2);
-                    // todo(2528) check if block signature has value else throw missing field
+                    currentJoinPointIndex = nextPath.nextPathIndex();
```
rockysingh: Nit: `currentJoinPointIndex = nextPath.nextPathIndex()` runs even in the `checkpoint == null` branch above, which just set `canFollowNextIndex = false` and will exit the loop next iteration. The assignment is dead in that branch. Moving it inside `if (checkpoint != null)` makes the intent ("only advance when we merged") explicit and matches how the read of `currentJoinPointIndex` is guarded.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationConfig.java
```diff
@@ -10,29 +10,24 @@
...
+///     been recently verified will result in an informational failure rather than a standard one.
 /// @param activeSessionsBufferSize size of maximum allowed active sessions. When full and a new session needs to
-///     start, room will be made for it by canceling the longest running one. todo(2528) could be lowest block nubmer
+///     start, room will be made for it by canceling the one with the lowest block being verified, except when
+///     the session we just started is the one with the lowest block being verified.
 /// @param allSourcesRequireOrdering if true, strict ordering of the next expected block to verify will be enforced
 ///     for blocks received from any source. If false, only
+///     [org.hiero.block.node.spi.blockmessaging.BlockSource#PUBLISHER] will be strictly ordered.<br/>
+///     _NOTE_ that this value should remain `true` in all "Tier 1" Block Nodes.
 ///     [org.hiero.block.node.spi.blockmessaging.BlockSource#PUBLISHER] will be strictly ordered.
```
rockysingh: Nit: duplicate leftover — `[...BlockSource#PUBLISHER] will be strictly ordered.` already appears on L22 as the "else" clause of the sentence. Here on L24 it's dangling after the `_NOTE_` line with no context. Delete.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: I see it now, I was expecting a boolean, so it wasn't quickly clear.

Thanks for the clarification.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: Good point, I would advocate that we fail with unable to parse here until the follow-on issue is resolved.

Another concern, which can be a separate PR:
    We must go through the "unknown fields" map as well, and fail if it contains anything (again, until 3195 is resolved).  It's important that this hasher either fail or hash every field in the data, leaving nothing to silently pass through.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
jsync-swirlds: We process every leaf in the list, and track join points, so we're pretty close to a completely closed process.
1. As long as _at least one_ leaf in the proof matches the root hash (value-to-prove), that is sufficient for the value-to-prove check.
1. We can (should?) check that there are no "dangling" merkle paths (i.e. paths that were not included in the final result). That might be a good follow-up item, however.

berryware (dismissed):
Looks good

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
ata-nas: We've discussed this with @jsync-swirlds and decided that we have bad block proof failure in both cases.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: Already answered in another thread: https://github.com/hiero-ledger/hiero-block-node/pull/3102#discussion_r3588394925

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: Moved inside the if statement, just after merging siblings.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationConfig.java
ata-nas: Good catch. Removed.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Good catch. I agree as well. For now, throwing an exception in default branch.

@jsync-swirlds about your other point, just throwing in the default branch should be sufficient? We will only hit it in case none of the rest of the switches hit a match. If we are in the default branch, we expect unknown fields to not be empty. But by throwing now, we cover for now until we implement #3195. Am I missing something?

berryware (dismissed):
Reapproving

rockysingh (dismissed):

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: To close the loop:
We added a separate check for the unknown fields list; initially this returns a bad block proof; we'll add proper hashing of those in follow-up.

jsync-swirlds (approved):

ivannov (approved):

PR #3105, opened 2026-06-29
fix(solo-e2e): fix wrb-cli-wrapping-validation regressions (unsigned + oversized record files)

AlfredoG87:
✅  Dry-Run test run: https://github.com/hiero-ledger/hiero-block-node/actions/runs/28400298918/job/84159984106

## Summary

The `wrb-cli-wrapping-validation` test in the `single` solo-e2e topology was failing repeatedly (3 attempts in a row on the scheduler, and again on `main`). After the OOM fix landed and the Block Node could stream past the previous crash point, the test reached the WRB CLI wrap step and exposed two separate, pre-existing problems. This PR fixes both.

Both failures surface in the `Run Test Definitions` step; the other tests (`smoke-test`, `tss-signature-transition`, `wrb-cli-hash-validation`) pass.

## Root causes

**1. Unsigned in-progress record file (intermittent)**

`wrb-cli-wrap-and-compare.sh` extracts the Consensus Node's record stream with `kubectl cp` while the CN is actively writing. That captures the newest record file mid-write: a truncated `.rcd.gz` whose `.rcd_sig` signature has not been flushed yet. The `blocks wrap` command hard-fails on any record file set missing its signature:

```
gzip: .../2026-..._..Z.rcd.gz: unexpected end of file
java.lang.RuntimeException: No signature files found for baseKey='2026-..._..Z' ...
Conversion complete. Blocks written: 0
```

**2. Oversized record file exceeds PBJ parse cap (deterministic)**

Once past the signature issue, wrap fails parsing the oversized "wraps transition" block (~30 MB) because the tools CLI parsed whole record files/blocks/transactions with PBJ's 2 MB default size:

```
com.hedera.pbj.runtime.ParseException: record_stream_items size 31570579 is greater than max 2097152
```

This was a latent regression from the PBJ 0.15.10 upgrade (#3080): call sites that previously used the 1-arg `Codec.parse(...)` were moved to `ParseHelper.standardParse(...)`, which defaults to `Codec.DEFAULT_MAX_SIZE` (2 MB). The block-node modules preserved their large parses with `Integer.MAX_VALUE`; the tools record-file paths were left at 2 MB.

## Changes

**Signature prune (`wrb-cli-wrap-and-compare.sh`)**
- Add `prune_incomplete_record_files`, called at the start of `generate_metadata_from_files` (the shared chokepoint for both the CN and MinIO extraction paths). It drops any `.rcd`/`.rcd.gz` lacking a matching `.rcd_sig`/`.rcd_sig.gz` before metadata generation and packaging. One rule covers both observed symptoms, since the truncated gz and the missing signature belong to the same in-progress file.

**Parse size cap consolidation (`tools` module)**
- Replace `MAX_RECORD_FILE_SIZE` (128 MB) and `MAX_PARSE_SIZE` (120 MB) in `ProtobufParsingConstants` with a single `MAX_MESSAGE_SIZE = Integer.MAX_VALUE`. Both were arbitrary bounded caps for the same purpose (parsing a whole trusted record file/block/transaction), and oversized blocks could exceed them. The cap only guards against runaway allocation on corrupt input, which is not a concern for data the tool itself produced or fetched.
- Route every whole-record/whole-block/transaction parse site through `MAX_MESSAGE_SIZE`: `ParsedRecordFile`, `RecordFileHasher`, `UnparsedRecordBlockV6`, `BlockExtractionUtils`, and the existing validation/model sites (`NodeStakeUpdateValidation`, `AddressBookUpdateValidation`, `TssEnablementValidation`, `HbarSupplyValidation`, `BlockReader`, `BlockZipsUtilities`, `BlockStreamBlockHasher`, `NetworkCapacityClient`, `LiveSequential`).
- Drop the duplicate local `MAX_RECORD_FILE_SIZE` constant in `TssEnablementValidation` in favor of the shared one.

## Test plan

- [x] `./gradlew :tools:spotlessApply :tools:compileJava :tools:test` — passes
- [x] Local solo-e2e `single` topology with the oversized ~30 MB record present: `wrb-cli-wrapping-validation` passes end-to-end — 16345 blocks wrapped, `VALIDATION PASSED`, 8/8 events, 3/3 assertions, 0 parse-size errors, 0 stake-data warnings
- [x] Confirmed the signature prune fires live (`Pruned 1 incomplete record file(s) lacking signatures`) and wrap progresses from 0 to all blocks
- [x] Solo E2E scheduler `single` job green in CI

## Reviewer Notes

This branch was cut before #3086 (the OOM/heap fix) merged, so it currently also carries that commit. Rebasing onto latest `main` drops it (already merged), leaving only the two fixes above.

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/ProtobufParsingConstants.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/ProtobufParsingConstants.java
@@ -11,11 +11,10 @@ public final class ProtobufParsingConstants {
 
     /// Maximum size for parsing a whole record file, block, or other large message.
     ///
-    /// Set to [Integer#MAX_VALUE] (effectively unbounded): these parses consume trusted data the
-    /// tool itself produced or fetched, and message sizes are determined by network activity (e.g.
-    /// oversized blocks), not by a bound we choose. The cap only guards against runaway allocation
-    /// on corrupt input, which is not a concern here.
-    public static final int MAX_MESSAGE_SIZE = Integer.MAX_VALUE;
+    /// Set to 300 MB: large enough to comfortably accommodate the biggest record files and blocks
+    /// the tool handles, while still bounding allocation so a limit is encountered here rather than
+    /// later in the Block Node. This guards against runaway allocation on corrupt input.
+    public static final int MAX_MESSAGE_SIZE = 300 * 1024 * 1024;
 
     private ProtobufParsingConstants() {}
 }
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/ProtobufParsingConstants.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/ProtobufParsingConstants.java
@@ -9,11 +9,13 @@ public final class ProtobufParsingConstants {
     /** Maximum protobuf nesting depth for parsing. */
     public static final int MAX_DEPTH = 512;
 
-    /** Maximum record file size for protobuf parsing (128 MB). */
-    public static final int MAX_RECORD_FILE_SIZE = 128 * 1024 * 1024;
-
-    /** Maximum parse size for protobuf messages (120 MB) to handle large blocks and StateChanges items. */
-    public static final int MAX_PARSE_SIZE = 120 * 1024 * 1024;
+    /// Maximum size for parsing a whole record file, block, or other large message.
+    ///
+    /// Set to [Integer#MAX_VALUE] (effectively unbounded): these parses consume trusted data the
+    /// tool itself produced or fetched, and message sizes are determined by network activity (e.g.
+    /// oversized blocks), not by a bound we choose. The cap only guards against runaway allocation
+    /// on corrupt input, which is not a concern here.
+    public static final int MAX_MESSAGE_SIZE = Integer.MAX_VALUE;
 
     private ProtobufParsingConstants() {}
 }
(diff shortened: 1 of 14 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/ProtobufParsingConstants.java
```diff
@@ -9,11 +9,13 @@ public final class ProtobufParsingConstants {
...
-
-    /** Maximum parse size for protobuf messages (120 MB) to handle large blocks and StateChanges items. */
-    public static final int MAX_PARSE_SIZE = 120 * 1024 * 1024;
+    /// Maximum size for parsing a whole record file, block, or other large message.
+    ///
+    /// Set to [Integer#MAX_VALUE] (effectively unbounded): these parses consume trusted data the
+    /// tool itself produced or fetched, and message sizes are determined by network activity (e.g.
+    /// oversized blocks), not by a bound we choose. The cap only guards against runaway allocation
+    /// on corrupt input, which is not a concern here.
+    public static final int MAX_MESSAGE_SIZE = Integer.MAX_VALUE;
```
jsync-swirlds: This is excessively large without need.
300M is larger than we should ever need; and actually keeps a reasonable limit.
We want a limit to be encountered in _this_ code rather than being encountered in the BN later.

jsync-swirlds (commented):
Recommend reducing the max message size to 300M rather than Integer.MAX_VALUE.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/ProtobufParsingConstants.java
AlfredoG87: MAX_VALUE comes down to 2GB, we were already using that value in some parsing places, so I choose the higher end and consolidated it on this constant, but I agree that 300 is more reasonable for a single block, specially since in the future the frequency of blocks production tends to go up, not down (more blocks per second = smaller blocks)

AlfredoG87: Done ✅

Nana-EC (approved):

berryware (approved):
Looks good.

jsync-swirlds (approved):

PR #3108, opened 2026-06-30
fix: Enable Subscriber to detect when messaging subscription is ended and re-register if needed.

### LINKED ISSUE #1715: `Stream Subscriber Plugin` Add Support for Request Max Duration and Fix unregistered Listener
### Story Form

As a Block Node Engineer/User
I want to have a limit to the maximum duration a subscribe stream request can live for
So that I can manage better my plugin

### Technical Notes

- we currently have no support for max duration of a single subscribe stream request
- we should add that:
  - the unit of measurement will be count of blocks sent
  - the max count of blocks sent per request has to be a configurable value
  - if this condition is reached, the session has to close with SUCCESS even if we have not reached the request's end criteria

Additional item:
- Subscriber currently does not detect when its registered listener for the messaging queue is unregistered because it falls behind.  This is incorrect.  The subscriber must:
   1. Detect that the listener is unregistered
   2. set a flag to reregister the listener when caught up
   3. When ready to transition back to live, reregister the listener
- Note, the `onTooFarBehind` method is called when the listener is unregistered. This is where we can set the flag to reregister.

jsync-swirlds:
## Reviewer Notes
* Added functional code for onTooFarBehindError to record de-registration.
* Added code to re-register when needed, gated by a CAS check.
* Fixed a disabled test and re-enabled it.

## Related Issue(s)
 Fixes #1715

### DIFF
--- a/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
+++ b/block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
@@ -540,6 +540,13 @@ public class BlockStreamSubscriberSession implements Callable<BlockStreamSubscri
         // Need to park here waiting for at least one entry on the queue, otherwise
         // we'll spin in the caller thread in an extreme form of spin wait.
         if (liveBlockQueue.isEmpty()) {
+            // Re-register the messaging handler if necessary.
+            // Use a CAS check to determine if we reregister
+            if (liveBlockHandler.canReregister()) {
+                blockNodeContext
+                        .blockMessaging()
+                        .registerNoBackpressureBlockItemHandler(liveBlockHandler, false, sessionContext.handlerName);
+            }
             // Wait briefly for a live block to be available.
             awaitNewLiveEntries();
         }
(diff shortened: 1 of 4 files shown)


AlfredoG87 (approved):
LGTM

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
```diff
@@ -875,24 +883,24 @@ private LiveBlockHandler(
...
         }
 
         @Override
         public void onTooFarBehindError() {
-            // Insert a signal to the session that it is "too far behind" live.
-            // Should not ever happen with this design.
-            LOGGER.log(Level.INFO, "Handler for {0} has fallen behind.", clientId);
+            // Mark this handler to be re-registered when the session is ready
+            // to swtich back to "live".
+            if (isRegisteredFlag.compareAndSet(true, false)) {
```
Nana-EC, line 897: I think an API/E2E test to confirm transition is pertinent.
Let's plan for a follow up

Nana-EC (approved):
LG, change highlight the value of an E2E test to capture this

block-node/stream-subscriber/src/main/java/org/hiero/block/node/stream/subscriber/BlockStreamSubscriberSession.java
jsync-swirlds: There are such tests, and always have been.
They aren't comprehensive (as evidence this issue was possible), so expanding the set is wise.

Most likely, we need E2E, because unit tests won't encounter the necessary interactions from messaging.

PR #3109, opened 2026-06-30
fix(roster-bootstrap-rsa): resolve genesis era without the blocks API

### LINKED ISSUE #3107: bug: single-wrb-rsa solo-e2e fails - RSA roster never loads (regression from #3070)
## Symptom

`single-wrb-rsa` job (`rsa-roster-verification`) has failed every Solo E2E Scheduler run since 2026-06-27:

```
FAIL rsa-verification-active: RSA roster not loaded (roster_entries_loaded=0)
| block-node-1 | Block Node | Running | Blocks: 18446744073709551615 - 18446744073709551615 |  (none stored)
```

The `roster-bootstrap-rsa` plugin loads but the address book stays empty, so every block fails RSA verification.

### Steps to Reproduce

1. Run the daily solo-e2e test schedule.

## Root cause

Regression from **#3070** (`f5b14f93`, merged 2026-06-26 — between the last pass and first fail), which reworked the bootstrap to a historical-era model. It resolves each era's block range via `/api/v1/blocks`. The Mirror Node reports the genesis era's `timestamp.from` as a near-epoch sentinel (`0.000000001`), not blank, so it isn't treated as genesis and the plugin calls `/api/v1/blocks`.

Catch-22: on a fresh network `/api/v1/blocks` is empty until the Block Node stores blocks, but the Block Node needs the roster to verify and store them. The era is skipped → roster empty → verification fails. (`/api/v1/network/nodes`, the roster source, is available the whole time.)

AlfredoG87:
## Summary

The `single-wrb-rsa` solo-e2e job (`rsa-roster-verification`) has failed since #3070 with `roster_entries_loaded=0` — the Block Node verifies no blocks because the RSA roster never loads.

#3070's historical-era bootstrap resolves each era's block range via `/api/v1/blocks`. For the genesis era the Mirror Node reports `timestamp.from` as a near-epoch sentinel (`0.000000001`) rather than blank, so it isn't recognized as genesis and the plugin calls `/api/v1/blocks`. On a fresh network that endpoint is empty until the Block Node stores blocks — but the Block Node can't store blocks without the roster. The era is skipped, the roster stays empty, and every block fails RSA verification.

Fixes #3107 

## Test plan

- [x] Unit test: genesis sentinel builds a one-era roster at block 0 and never calls the blocks API (fails without the fix)
- [x] Local solo-e2e `single-wrb-rsa`, fixed plugin injected manually (the published SNAPSHOT lacks the fix, so it can't ride the normal deploy): roster loads — `roster_entries_loaded` 0 → 1, "RSA key map updated: 1/1 nodes loaded", zero `/blocks` calls
- [ ] Full green `rsa-roster-verification` on CI after merge republishes the `main` SNAPSHOT with the fix

### DIFF
--- a/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
+++ b/block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
@@ -437,12 +437,12 @@ public class RsaRosterBootstrapPlugin implements BlockNodePlugin {
     /// Converts a (from, to) timestamp pair to a [startBlock, endBlock] pair by querying the
     /// Mirror Node blocks API. Returns {@code null} if the block range cannot be determined.
     ///
-    /// Blank timestamps are treated as sentinels: blank {@code from} means genesis (startBlock=0)
-    /// with no API call; blank {@code to} means open-ended (endBlock=-1) with no API call.
+    /// Sentinels are resolved without an API call: a genesis {@code from} (see [#isGenesisFrom])
+    /// means startBlock=0; a blank {@code to} means open-ended (endBlock=-1).
     private long[] fetchBlockRange(final HttpClient client, final String from, final String to) {
         final long startBlock, endBlock;
-        if (from == null || from.isBlank()) {
-            startBlock = 0L; // no from-timestamp → treat as genesis
+        if (isGenesisFrom(from)) {
+            startBlock = 0L; // genesis address book → block 0, no blocks API call needed
             endBlock = -1L;
         } else {
             String query = to == null || to.isBlank()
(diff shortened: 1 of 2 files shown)


block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -462,6 +462,24 @@ private long[] fetchBlockRange(final HttpClient client, final String from, final
...
+    ///
+    /// This lets the genesis era resolve to block 0 without the blocks API, which is required on
+    /// networks (e.g. solo) where the Mirror Node has no blocks until the Block Node — which needs
+    /// this roster to verify them — starts storing them.
+    private static boolean isGenesisFrom(final String from) {
+        if (from == null || from.isBlank()) {
+            return true;
+        }
+        try {
+            return Double.parseDouble(from) < 1.0;
```
Nana-EC, line 477: For completeness should this be < epoch?

Nana-EC (approved):
One suggestion.

berryware (approved):
Looks good.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
jsync-swirlds: `epoch == 0`.  This PR exists because the value is slightly greater than the epoch value in real environments.

We should have a constant for this `1.0` magic number; perhaps "EPOCH_CLOSE_SENTINEL" or similar?

jsync-swirlds (approved):

PR #3113, opened 2026-06-30
feat: add block-verification to published projects in build.gradle.kts

### LINKED ISSUE #3110: Publish `block-verification` plugin to Maven Central
Publish `block-verification` plugin to Maven Central

The `block-verification` module (#2869) is missing from `gradle/aggregation/build.gradle.kts`, so it is never included in the `publishAggregationToCentralPortal` run and is absent from Maven Central. Add it alongside `verification` so the artifact is available for E2E validation before the switchover.

## Change

`gradle/aggregation/build.gradle.kts` — add `published(project(":block-verification"))`

AlfredoG87:
## Reviewer Notes

This is needed so the plugin can be published to Maven and used by the deployment flow.

Fixes #3110

### DIFF
--- a/gradle/aggregation/build.gradle.kts
+++ b/gradle/aggregation/build.gradle.kts
@@ -20,6 +20,7 @@ dependencies {
     published(project(":roster-bootstrap-rsa"))
     published(project(":roster-bootstrap-tss"))
     published(project(":verification"))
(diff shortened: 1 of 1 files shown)


jsync-swirlds (approved):

berryware (approved):
Looks good

PR #3114, opened 2026-06-30
feat: Defer ASF stored-block registration in TempArchiveUploadTask until multipart upload completes

### LINKED ISSUE #3041: Defer ASF stored-block registration in TempArchiveUploadTask until multipart upload completes
### Story Form

I want to register a temp archive's block range with ApplicationStateFacility only after doCompleteMultipartUpload and the companion .meta object are both written successfully
So that a failed or aborted multipart upload never leaves ghost stored-block ranges in ASF that prevent backfill from re-delivering the missing blocks and closing the temp-archive coverage gap

### Background

`TempArchiveUploadTask` sends `PersistedNotification(success=true)` and calls `addStoredBlockRange` after each successfully flushed S3 part, before the multipart upload is complete. If a later part or `completeMultipartUpload` fails, `abortQuietly` removes all uploaded data from S3, but the optimistic ASF registrations are not retracted.

The gap this leaves in `tempArchiveTracker` cannot be closed in-process: `StartupRecoveryTask` only rebuilds the tracker from .meta files, and .meta is written only after a successful `doCompleteMultipartUpload`, so an aborted upload leaves nothing for recovery to find. The only path to eventual consolidation is for backfill to re-deliver the missing blocks into a fresh `TempArchiveUploadTask`. But backfill skips blocks that ASF already reports as stored, so the ghost ranges block re-delivery indefinitely and the group is never consolidated until the node restarts and ASF state is rebuilt from scratch.
  
### Notes

- Remove the per-part `addStoredBlockRange` calls inside the while loop and the final-buffer block in `TempArchiveUploadTask`.
- Add a single `addStoredBlockRange(new LongRange(firstBlock, lastBlock))` call after `doCompleteMultipartUpload` and `doUploadTextFile (.meta)` both succeed — matching the point at which the block range becomes durable and discoverable by `StartupRecoveryTask`.
- The `PersistedNotification(success=true)` sends can similarly be deferred to a single notification at the same point; or kept per-part as an optimistic hint to downstream consumers (same trade-off as `BlockUploadTask`).
- On the failure path, only the `PersistedNotification(success=false)` for the last-attempted range needs to remain; no ASF cleanup is required because `addStoredBlockRange` was never called.
- Update `TempArchiveUploadTaskTest` to assert that no `addStoredBlockRange` call is made before `doCompleteMultipartUpload` returns successfully, and that a single call covering [firstBlock, lastBlock] is made on success.
- In `StartupRecoveryTask.recoverFromCompletedObjects()`, instead of only finding the last key (`findLastKey`), enumerate all tar objects under the configured prefix and collect every group's block range.
- In `completeRecoveryIfReady()`, replace the single `addStoredBlockRange(new LongRange(0, nextBlockToQueue - 1))` call with per-archive calls, one `addStoredBlockRange` per entry in `completedRanges`.
- (Possible enhancement) Add metada JSON file for the final archives as well, not just for temporary ones. Use that file to determine the stored blocks ranges and not tar file names
- In the consolidation task, decompress and parse each temp .tar.zst fetched from S3 entry-by-entry rather than binary-concatenating the raw bytes, verifying that the extracted block numbers match the range declared in the .meta file before writing to the consolidated output.

ivannov:
## Reviewer Notes

### Deferred ASF registration and persisted notification in `TempArchiveUploadTask`                                                                                                   
                                                                                                                                                                                      
Previously `TempArchiveUploadTask` called `addStoredBlockRange` and sent a `PersistedNotification(success=true)` after each flushed S3 part, before the multipart upload was  complete. If `completeMultipartUpload` or the subsequent `.meta` write failed, `abortQuietly` removed the data from S3 but the ASF registrations were not reverted.                                                                        
                                                                                                                                                                                      
The fix removes all per-part ASF and notification calls from the while loop and the final-buffer block. A single `addStoredBlockRange(new LongRange(firstBlock, lastBlock))` and `PersistedNotification(success=true)` are now emitted only after both `doCompleteMultipartUpload` and the `.meta` write succeed. Each failure site is wrapped in its own try-catch. Metric increments are consolidated to the same point. The `lastNotifiedBlock` tracking variable is replaced by `totalBytes`, which accumulates part sizes for the deferred metrics call.
                                                        
### Per-archive range registration during startup recovery                                                                                                                          
                                                                                                                                                                                      
`StartupRecoveryTask.recoverFromCompletedObjects` previously used a right-hand-path traversal (`findLastKey`) that only located the alphabetically last completed tar and returned a single broad `[0, nextBlockToQueue-1]` range to `CloudStorageArchivePlugin`. This over-reported stored blocks when the node had gaps in its archive history.                        
                                                                                                                                                                                      
`findLastKey` is replaced by `findAllTarKeys`, which pages through all objects under the configured prefix (excluding `tmp/`) and collects every final tar key (filtering out `.tmp` and `.meta` by suffix). The new `RecoveryResult.completedRanges` field carries one `LongRange` per completed archive back to `CloudStorageArchivePlugin`. During startup, if  `completedRanges` is non-null (Case 1), the plugin registers each range individually via `addStoredBlockRange` instead of the broad fallback. Case 2 (resume path, where completed archives are not enumerated) continues to use the broad `[0, nextBlockToQueue-1]` range and sets `completedRanges` to `null` to signal this. 

## Related Issue(s)

Closes #3041

### DIFF
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
@@ -13,6 +13,14 @@ import java.util.List;
 /// round-trip `parse(format(groupStart, level), level) == groupStart` always holds.
 final class ArchiveKey {
 
+    /// Width, in characters, of each `/`-delimited path segment produced by [format] and consumed
+    /// by [parse] (and, for the segment count alone, by [StartupRecoveryTask#directoryDepth]).
+    static final int PATH_SEGMENT_WIDTH = 4;
+
+    /// Number of decimal digits `groupStart` is zero-padded to in [format]; `Long.MAX_VALUE` has 19
+    /// digits, so this fits any `long` block number and keeps S3's key ordering numeric.
+    static final int MAX_LONG_DIGITS = 19;
+
     private ArchiveKey() {}
 
     /// Formats the S3 object key for the tar group whose first block number is `groupStart`,
@@ -27,10 +35,11 @@ final class ArchiveKey {
     ///
     /// When `objectKeyPrefix` is non-empty the returned key is `{prefix}/{block-path}.tar`.
     static @NonNull String format(long groupStart, int groupingLevel, @NonNull String objectKeyPrefix) {
-        final String truncated = String.format("%019d", groupStart).substring(0, 19 - groupingLevel);
+        final String truncated =
+                String.format("%0" + MAX_LONG_DIGITS + "d", groupStart).substring(0, MAX_LONG_DIGITS - groupingLevel);
         final List<String> parts = new ArrayList<>();
-        for (int i = 0; i < truncated.length(); i += 4) {
-            parts.add(truncated.substring(i, Math.min(i + 4, truncated.length())));
+        for (int i = 0; i < truncated.length(); i += PATH_SEGMENT_WIDTH) {
+            parts.add(truncated.substring(i, Math.min(i + PATH_SEGMENT_WIDTH, truncated.length())));
         }
         parts.set(parts.size() - 1, String.valueOf(Long.parseLong(parts.getLast())));
         final String blockPath = String.join("/", parts) + ".tar";
@@ -75,7 +84,7 @@ final class ArchiveKey {
 
     /// Returns the width (number of characters) of the last path segment for a given grouping level.
     private static int lastSegmentWidth(int groupingLevel) {
-        final int width = (19 - groupingLevel) % 4;
-        return width == 0 ? 4 : width;
+        final int width = (MAX_LONG_DIGITS - groupingLevel) % PATH_SEGMENT_WIDTH;
+        return width == 0 ? PATH_SEGMENT_WIDTH : width;
     }
 }
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
@@ -17,9 +17,9 @@ final class ArchiveKey {
     /// by [parse] (and, for the segment count alone, by [StartupRecoveryTask#directoryDepth]).
     static final int PATH_SEGMENT_WIDTH = 4;
 
-    /// Number of decimal digits `groupStart` is zero-padded to in [format]; `Long.MAX_VALUE` has 19
-    /// digits, so this fits any `long` block number and keeps S3's key ordering numeric.
-    static final int MAX_LONG_DIGITS = 19;
+    /// Number of decimal digits `groupStart` is zero-padded to in [format]; this fits any `long`
+    /// block number and keeps S3's key ordering numeric.
+    static final int MAX_LONG_DIGITS = Long.toString(Long.MAX_VALUE).length();
 
     private ArchiveKey() {}
 
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/RecoveryResult.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/RecoveryResult.java
@@ -2,6 +2,7 @@
 package org.hiero.block.node.cloud.storage.archive;
 
 import java.util.List;
+import org.hiero.block.node.spi.historicalblocks.LongRange;
 
 /// The outcome of a [StartupRecoveryTask] run.
 ///
@@ -32,6 +33,11 @@ import java.util.List;
 ///                          recovery paths that skip the temporary-archive scan
 /// @param lastHandedOffBlock the last block number handed off to any upload task before the
 ///                           restart, or `-1` when nothing was archived yet (fresh start)
+/// @param completedRanges   one [LongRange] per completed tar archive found during startup, so
+///                          that [CloudStorageArchivePlugin] can register them individually
+///                          instead of a single broad `[0, nextBlockToQueue-1]` range;
+///                          `null` when the recovery path does not enumerate completed archives
+///                          (Case 2 resume path)
 record RecoveryResult(
         long currentGroupStart,
         String uploadId,
@@ -39,4 +45,5 @@ record RecoveryResult(
         long nextBlockNumber,
         byte[] trailingBytes,
         List<TempArchiveEntry> tempArchives,
-        long lastHandedOffBlock) {}
+        long lastHandedOffBlock,
+        List<LongRange> completedRanges) {}
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/RecoveryResult.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/RecoveryResult.java
@@ -2,7 +2,7 @@
 package org.hiero.block.node.cloud.storage.archive;
 
 import java.util.List;
-import org.hiero.block.node.spi.historicalblocks.LongRange;
+import org.hiero.block.node.base.ranges.ConcurrentLongRangeSet;
 
 /// The outcome of a [StartupRecoveryTask] run.
 ///
@@ -33,11 +33,11 @@ import org.hiero.block.node.spi.historicalblocks.LongRange;
 ///                          recovery paths that skip the temporary-archive scan
 /// @param lastHandedOffBlock the last block number handed off to any upload task before the
 ///                           restart, or `-1` when nothing was archived yet (fresh start)
-/// @param completedRanges   one [LongRange] per completed tar archive found during startup, so
-///                          that [CloudStorageArchivePlugin] can register them individually
-///                          instead of a single broad `[0, nextBlockToQueue-1]` range;
-///                          `null` when the recovery path does not enumerate completed archives
-///                          (Case 2 resume path)
+/// @param completedRanges   the ranges of completed tar archives found during startup, so that
+///                          [CloudStorageArchivePlugin] can register the actual coverage
+///                          (including any gaps) instead of a single broad
+///                          `[0, nextBlockToQueue-1]` range; `null` when the recovery path does
+///                          not enumerate completed archives (Case 2 resume path)
 record RecoveryResult(
         long currentGroupStart,
         String uploadId,
@@ -46,4 +46,4 @@ record RecoveryResult(
         byte[] trailingBytes,
         List<TempArchiveEntry> tempArchives,
         long lastHandedOffBlock,
-        List<LongRange> completedRanges) {}
+        ConcurrentLongRangeSet completedRanges) {}
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
@@ -154,31 +154,59 @@ class StartupRecoveryTask implements Callable<RecoveryResult> {
         return recoveryResult;
     }
 
-    /// Pages through ALL objects under the configured prefix (excluding `tmp/`) and returns
-    /// the sorted list of final tar keys -- those that do not end with `.tmp` or `.meta`.
+    /// Walks the `/`-delimited directory hierarchy under the configured prefix (excluding `tmp/`)
+    /// and returns the list of all final tar keys.
     ///
-    /// O(N/1000) list calls for N completed archives. The full list is required to populate
-    /// [RecoveryResult#completedRanges()]; a faster O(depth) last-key traversal using
-    /// delimiter listing would suffice for `currentGroupStart` alone but not for per-range
-    /// accuracy.
+    /// [ArchiveKey#format] lays out each group as `directoryDepth` path segments followed by a
+    /// leaf segment folded into the object name (e.g. `0000/0000/0000/0001/23.tar`). This walks
+    /// [directoryDepth] delimiter listings (returning `CommonPrefixes`, one per branch) before
+    /// issuing a single flat listing per leaf directory to read the actual tar keys, bounding the
+    /// list-call count by O(levels x branches) rather than the total number of completed archives,
+    /// with no loss of per-range accuracy since every tar key is still read from a leaf listing.
     private List<String> findAllTarKeys(S3Client s3) throws S3ResponseException, IOException {
         final String objectKeyPrefix = config.objectKeyPrefix();
         final String excludedPrefix = TempArchiveKey.tmpPrefix(objectKeyPrefix);
         final String rootPrefix = objectKeyPrefix.isEmpty() ? "" : objectKeyPrefix + "/";
         final List<String> tarKeys = new ArrayList<>();
+        collectTarKeys(s3, rootPrefix, directoryDepth(), excludedPrefix, tarKeys);
+        return tarKeys;
+    }
+
+    /// Number of `/`-delimited directory levels [ArchiveKey#format] emits before the leaf segment
+    /// that is folded into the object name, for the configured [CloudStorageArchiveConfig#groupingLevel()].
+    private int directoryDepth() {
+        final int digitCount = 19 - config.groupingLevel();
+        final int segmentCount = (digitCount + 3) / 4;
+        return segmentCount - 1;
+    }
+
+    /// Recursively descends `remainingDepth` levels of `/`-delimited directories starting at
+    /// [prefix], collecting completed tar keys into [out]. At `remainingDepth == 0`, [prefix] is a
+    /// leaf directory: a flat (non-delimited) listing reads its objects directly, giving the exact
+    /// tar keys without ever walking sibling leaf directories.
+    private void collectTarKeys(S3Client s3, String prefix, int remainingDepth, String excludedPrefix, List<String> out)
+            throws S3ResponseException, IOException {
+        final String delimiter = remainingDepth == 0 ? null : "/";
         String token = null;
         boolean hasMore = true;
         while (hasMore) {
-            final S3Client.ListPage page = s3.listObjectsPage(rootPrefix, token, null, MAX_LIST_RESULTS);
-            for (final String key : page.keys()) {
-                if (!key.startsWith(excludedPrefix) && !key.endsWith(".tmp") && !key.endsWith(".meta")) {
-                    tarKeys.add(key);
+            final S3Client.ListPage page = s3.listObjectsPage(prefix, token, delimiter, MAX_LIST_RESULTS);
+            if (remainingDepth == 0) {
+                for (final String key : page.keys()) {
+                    if (key.endsWith(".tar")) {
+                        out.add(key);
+                    }
+                }
+            } else {
+                for (final String childPrefix : page.keys()) {
+                    if (!childPrefix.equals(excludedPrefix)) {
+                        collectTarKeys(s3, childPrefix, remainingDepth - 1, excludedPrefix, out);
+                    }
                 }
             }
             token = page.continuationToken();
             hasMore = token != null;
         }
-        return tarKeys;
     }
 
     /// Case 2: one hanging upload — complete it to materialize a readable S3 object, start a fresh
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.stream.Collectors;
+import org.hiero.block.node.spi.historicalblocks.LongRange;
 
 /// Determines where [CloudStorageArchivePlugin] should resume uploading after a restart.
 ///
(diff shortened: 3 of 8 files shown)


block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
```diff
@@ -122,84 +124,61 @@ private Map<String, List<String>> regularUploadsUnderPrefix(Map<String, List<Str
...
-    /// the alphabetically last common prefix at each level, skipping the `tmp/` virtual directory,
-    /// then pages through the leaf objects to return the key of the last one, or `null` if the
-    /// bucket is empty.
+    /// Pages through ALL objects under the configured prefix (excluding `tmp/`) and returns
+    /// the sorted list of final tar keys -- those that do not end with `.tmp` or `.meta`.
     ///
-    /// When [CloudStorageArchiveConfig#objectKeyPrefix()] is configured the traversal is rooted
-    /// at `{prefix}/` so that sibling prefixes in the same bucket are never visited.
-    private String findLastKey(S3Client s3) throws S3ResponseException, IOException {
+    /// O(N/1000) list calls for N completed archives. The full list is required to populate
```
Nana-EC: The gain here seems accurate but it worsens as we scale.
I wonder if we have potential improvement here to consider in follow up.
Should we be capturing the completedRanges in bucket to reduce the list occurrences, this requires additional validation otherwise it forces trust. 🤔 
@rockysingh thoughts here as you've had to explore similar optimizations

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
```diff
@@ -132,27 +122,26 @@ public TempArchiveEntry call()
...
             }
 
-            // If this throws after the success notification above, the result is a false-persisted block; accepted by
-            // design, same as BlockUploadTask.
-            doCompleteMultipartUpload(s3, s3Key, uploadId, etags);
+            try {
+                doCompleteMultipartUpload(s3, s3Key, uploadId, etags);
+            } catch (S3ResponseException | IOException e) {
+                S3UploadUtils.abortQuietly(s3, s3Key, uploadId);
+                blockMessaging.sendBlockPersisted(new PersistedNotification(firstBlock, false, 1_000, lastSource));
```
Nana-EC, line 141: Q: Do we have a test case for this path explicitly?
Looking at the test cases i'm not sure we differentiate between the failed `PersistedNotification` cases thrown after `doUploadPart()`, `doCompleteMultipartUpload()` and `doUploadTextFile()`

Nana-EC (dismissed):
LG.
Made mentions of potential follow up and an additional test case

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
ivannov: I am not sure whether I understand your question, Nana :)

Let's start with this. In `TempArchiveUploadTask.call()` method we have four catch blocks. Here are the tests that cover them:

1. First `doUploadPart()`, inside while block: `partUploadFailureDuringLoopThrowsAndAbortsUpload()`
2. Second `doUploadPart()`, outside while block, if there are some more bytes in the buffer: `partUploadFailureForFinalPartSendsFailedNotificationAndThrows()`
3. `doCompleteMultipartUpload`: `completeFailureThrowsAndLeavesNoKeys()`
4. `doUploadTextFile()`: `metaWriteFailureThrowsAndLeavesTmpButNoMeta()`

Please reply to this message if you had in mind something else

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
rockysingh: Real trade-off. Current cost is O(N/1000) LIST calls per startup (`MAX_LIST_RESULTS = 1000` in `findAllTarKeys`), so on a long-lived BN with, say, 10M completed archives that's ~10K LIST calls at startup — bounded, sub-minute today, but grows.

If we cache `completedRanges` in the bucket, the failure mode we need to defend against is **staleness after out-of-band changes**: retention deletes, manual cleanup, another BN sharing the prefix, a partial abort from a previous run. Blindly trusting the manifest loses the very property this task exists to protect.

A middle-ground worth exploring in the follow-up:

1. **Persist** a `state/completed-ranges.json` sibling under the same prefix on every successful group commit (append-only, one small `PUT`, doesn't need coordination if we keep it monotonic).
2. **On startup**: read the manifest with a single `GET`, then **LIST-verify only the tail** — the last N groups, or everything since the manifest's high-water mark. That gives us `currentGroupStart` accuracy for recovery (which is what the switch statement actually depends on) without walking the whole prefix.
3. **On tail mismatch**: invalidate the manifest and fall back to the current full-paged rescan; log loudly so we can spot out-of-band mutations.

Bounds: `GET(manifest)` + `LIST(tail)` is O(1) list calls in the happy path vs. O(N/1000) today. Correctness stays local — we only ever *trust* the range that the tail-verify confirmed.

Concurrent-BN case is the sharpest edge — two writers racing the manifest need last-writer-wins with a versioned key, or a leader-per-prefix. That's the piece I'd want to nail down in the follow-up ticket before shipping.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
rockysingh: Good catch — checked the test file. The two mid-loop-part branches (`doUploadPart` failing mid-loop → **no** notification; `doUploadPart` failing on the final buffer flush → **one failed** notification) are both explicitly asserted (`assertThat(messaging.getSentPersistedNotifications()).isEmpty()` / `.hasSize(1)`).

The two remaining paths — the ones this PR is actually changing timing for — only assert **S3 state**, not the notification effect:

- `completeFailureThrowsAndLeavesNoKeys` (`doCompleteMultipartUpload` throws) asserts `.tmp` and `.meta` are absent, but doesn't check that exactly one failed `PersistedNotification(firstBlock, false, …)` was sent per line 141 of `TempArchiveUploadTask`.
- `metaWriteFailureThrowsAndLeavesTmpButNoMeta` (`doUploadTextFile` throws) asserts `.tmp` is present and `.meta` is absent, but doesn't check the success `PersistedNotification(lastBlock, true, …)` from line 157 or the `applicationStateFacility.addStoredBlockRange` side-effect from line 158.

Since the whole point of this PR is *when* those notifications fire relative to the S3 boundaries, the two existing tests should each grow one `messaging.getSentPersistedNotifications()` assertion (and the meta-write case one `applicationStateFacility` interaction check). Small addition — happy to do a follow-up if you'd rather ship this and layer it, but it's cheap enough to add here.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
rockysingh: Digging further — there's a bigger win available before we even need to reach for caching. Bucky's `S3Client.listObjectsPage` already accepts a `delimiter` parameter (it just returns `CommonPrefixes` instead of `Contents`) — the current code always passes `null`, so we're doing a flat walk of every leaf object.

For a grouping-level-4 layout (`000/000/000/000/…/tar`), a hierarchical delimiter walk needs roughly `O(levels × branches)` — call it ~5 levels × ~10 branches ≈ **50 LIST calls** — vs. the current `O(N/1000)` flat walk which is ~10K calls at 10M archives. Two orders of magnitude, and no correctness compromise.

Concretely: LIST the root with `delimiter="/"` to get the top-level prefixes, recurse down to the parent-of-tar level. At that level a group is either fully committed (the atomic multipart-complete step is what creates the `.tar` object) or fully absent, so the presence of the group prefix is itself the signal — we can convert it straight to a `LongRange` via `ArchiveKey.parse` without ever reading the leaf name.

If we don't want to trust "prefix present == group complete" (defensive against a raced abort leaving an empty prefix), we can still LIST-verify at the group-parent level with a single non-delimiter call each, which still beats the flat walk by roughly 10×.

Caching then becomes the second-line optimization for when even the delimiter walk grows painful — but the delimiter change is much smaller and doesn't introduce any manifest-staleness or concurrent-writer edges. Would suggest doing that one first.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
ivannov: @rockysingh 

Maybe I don't get your comment, but

> completeFailureThrowsAndLeavesNoKeys (doCompleteMultipartUpload throws) asserts .tmp and .meta are absent, but doesn't check that exactly one failed PersistedNotification(firstBlock, false, …) was sent per line 141 of TempArchiveUploadTask.

If you look at lines 229-233:

```java
        final List<PersistedNotification> notifications = messaging.getSentPersistedNotifications();
        assertThat(notifications).hasSize(1);
        assertThat(notifications.getFirst().succeeded()).isFalse();
        assertThat(notifications.getFirst().blockNumber()).isZero();
        assertThat(notifications.getFirst().blockSource()).isEqualTo(BlockSource.PUBLISHER);
```

> metaWriteFailureThrowsAndLeavesTmpButNoMeta (doUploadTextFile throws) asserts .tmp is present and .meta is absent, but doesn't check the success PersistedNotification(lastBlock, true, …) from line 157 or the applicationStateFacility.addStoredBlockRange side-effect from line 158.

Here the lines are 260-266:

```java
        assertThat(asf.addedRanges).hasSize(1);
        assertThat(asf.addedRanges.getFirst()).isEqualTo(new LongRange(0, 0));
        final List<PersistedNotification> notifications = messaging.getSentPersistedNotifications();
        assertThat(notifications).hasSize(1);
        assertThat(notifications.getFirst().succeeded()).isTrue();
        assertThat(notifications.getFirst().blockNumber()).isZero();
        assertThat(notifications.getFirst().blockSource()).isEqualTo(BlockSource.PUBLISHER);
```

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
ivannov: Great findings, @rockysingh !

I've implemented the proposed algorithm

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
```diff
@@ -122,84 +124,89 @@ private Map<String, List<String>> regularUploadsUnderPrefix(Map<String, List<Str
...
-                }
-            }
-            token = page.continuationToken();
-            hasMore = token != null;
-        }
-        return lastPrefix;
+    /// Number of `/`-delimited directory levels [ArchiveKey#format] emits before the leaf segment
+    /// that is folded into the object name, for the configured [CloudStorageArchiveConfig#groupingLevel()].
+    private int directoryDepth() {
+        final int digitCount = 19 - config.groupingLevel();
```
AlfredoG87: should we have constant for this magic number? `19`

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
```diff
@@ -122,84 +124,89 @@ private Map<String, List<String>> regularUploadsUnderPrefix(Map<String, List<Str
...
+        final int segmentCount = (digitCount + 3) / 4;
```
AlfredoG87: why the  +3)/4 ?  as hardcoded, maybe make constants?

AlfredoG87 (dismissed):
LGTM, just 2 nit suggestions.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
jsync-swirlds: I think the `3` is `config.groupingLevel() - 1` and the `4` is `config.groupingLevel()`.
If that's true, then we should calculate those two numbers accordingly.
```suggestion
        final int segmentOffset = config.groupingLevel() - 1;
        final int segmentSize = config.groupingLevel();
        final int segmentCount = (digitCount + segmentOffset) / segmentSize;
```

ivannov: Well, it doesn't have to do with the grouping level. 4 is the number of symbols in a "directory" in S3. That is, the width of a segment. And 3 is (4 - 1).

But as these magic numbers brought confusion, I am changing them to a constant.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
ivannov: Changed that to a constant on the `ArchiveKey` class

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
```diff
@@ -13,6 +13,14 @@
 /// round-trip `parse(format(groupStart, level), level) == groupStart` always holds.
 final class ArchiveKey {
 
+    /// Width, in characters, of each `/`-delimited path segment produced by [format] and consumed
+    /// by [parse] (and, for the segment count alone, by [StartupRecoveryTask#directoryDepth]).
+    static final int PATH_SEGMENT_WIDTH = 4;
```
jsync-swirlds, line 18: This is dependent on configuration, it shouldn't be a constant, but a calculation.
`segmentSize = config.groupingLevel();`
The config is in the cloud archive plugin, so it needs to be a method parameter in `format` below, and `lastSegmentWidth` should use groupingLevel instead of a fixed path segment width constant.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
```diff
@@ -13,6 +13,14 @@
...
+
+    /// Number of decimal digits `groupStart` is zero-padded to in [format]; `Long.MAX_VALUE` has 19
+    /// digits, so this fits any `long` block number and keeps S3's key ordering numeric.
+    static final int MAX_LONG_DIGITS = 19;
```
jsync-swirlds: This is _OK_, but it would be better as a calculation:

```suggestion
    static final int MAX_LONG_DIGITS = Long.MAX_VALUE.toString().length();
```

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
```diff
@@ -122,84 +124,89 @@ private Map<String, List<String>> regularUploadsUnderPrefix(Map<String, List<Str
...
-            final long groupStart = ArchiveKey.parse(lastKey, config.groupingLevel(), config.objectKeyPrefix());
+            final List<LongRange> completedRanges = new ArrayList<>(allTarKeys.size());
+            long maxGroupStart = -1;
+            for (final String key : allTarKeys) {
+                final long groupStart = ArchiveKey.parse(key, config.groupingLevel(), config.objectKeyPrefix());
+                completedRanges.add(new LongRange(groupStart, groupStart + groupSize - 1));
+                if (groupStart > maxGroupStart) {
+                    maxGroupStart = groupStart;
+                }
+            }
```
jsync-swirlds: Instead of managing a list, why not use a ConcurrentLongRangeSet and generate the list at the end, that way we merge together contiguous ranges.

```suggestion
            final ConcurrentLongRangeSet foundRanges = new ConcurrentLongRangeSet();
            long maxGroupStart = -1;
            for (final String key : allTarKeys) {
                final long groupStart = ArchiveKey.parse(key, config.groupingLevel(), config.objectKeyPrefix());
                foundRanges.add(new LongRange(groupStart, groupStart + groupSize - 1));
                if (groupStart > maxGroupStart) {
                    maxGroupStart = groupStart;
                }
            }
            final List<LongRange> completedRanges = foundRanges.streamRanges().toList();
```

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/RecoveryResult.java
```diff
@@ -32,11 +33,17 @@
 ///                          recovery paths that skip the temporary-archive scan
 /// @param lastHandedOffBlock the last block number handed off to any upload task before the
 ///                           restart, or `-1` when nothing was archived yet (fresh start)
+/// @param completedRanges   one [LongRange] per completed tar archive found during startup, so
+///                          that [CloudStorageArchivePlugin] can register them individually
+///                          instead of a single broad `[0, nextBlockToQueue-1]` range;
+///                          `null` when the recovery path does not enumerate completed archives
+///                          (Case 2 resume path)
```
jsync-swirlds: I wonder if a `ConcurrentLongRangeSet` might be a better option than a `List<LongRange>` because the former automatically merges contiguous ranges...

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
ivannov: This is actually a constant, independent of the grouping level and there are parameterized tests that prove it. It just says how many digits goes into the S3 "folder". For example:

level=1: 0000/0000/0000/0001/23.tar   <- all non-last segments are 4 chars
level=3: 0000/0000/0000/123.tar          <- still 4 chars
level=6: 0000/0000/0000/1.tar              <- still 4 chars

Only the last segment's width varies with `groupingLevel`.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/RecoveryResult.java
ivannov: Changed to `ConcurrentLongRangeSet`

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
jsync-swirlds: Ahh, the folders are fixed-ish (because 19 - grouping might not be divisible by 4).
I forgot we chose fixed folder sizes.

Thanks.

jsync-swirlds (approved):

ata-nas (approved):
LG.

I have some concerns as to how this change will affect publisher and backfill.

As of now, we will not have publisher in RFH, but that does not mean we will not have it in the future. We currently only have backfill with this plugin.

My thought is that the low frequency of updates of the stored blocks and success/failure of blocks received here might affect how backfill and publisher work. In the case of publisher, low frequency means no timely acknowledgements. In the case of backfill, (note that I am not sure how backfill operates and handles this case) this might mean that backfill will backfill x blocks, but since no updates to stored block and persistence notifications will come soon, backfill might attempt to backfill the same x blocks again.

cc: @jsync-swirlds @Nana-EC @AlfredoG87

ivannov:
> LG.
> 
> I have some concerns as to how this change will affect publisher and backfill.
> 
> As of now, we will not have publisher in RFH, but that does not mean we will not have it in the future. We currently only have backfill with this plugin.
> 
> My thought is that the low frequency of updates of the stored blocks and success/failure of blocks received here might affect how backfill and publisher work. In the case of publisher, low frequency means no timely acknowledgements. In the case of backfill, (note that I am not sure how backfill operates and handles this case) this might mean that backfill will backfill x blocks, but since no updates to stored block and persistence notifications will come soon, backfill might attempt to backfill the same x blocks again.
> 
> cc: @jsync-swirlds @Nana-EC @AlfredoG87

Thank you @ata-nas for catching this potential issue. Here are my thoughts about it.

To start with, this deference affects only out-of-order blocks, not the regular in-order path, which reports per S3 part uploaded.

But still the problem is there and there is the risk of the backfill plugin "hammering" the range which is handled by the temp upload task. What #3129 is introducing is an exponential backoff for ranges that do not shrink, thus avoiding unbounded repeats of backfills for such ranges. 

So the problem is reduced to "backfill may redundantly re-fetch a range once or twice while a slow temp segment is open". That's a real improvement but doesn't fully close the gap you're describing.

Even though I have enough approvals for merging the PR, I'll wait for other colleagues' opinion on the described situation.

jsync-swirlds:
Publisher won't be compatible with cloud archive plugin any time soon.
There are many reasons for this, but we can discuss them separately.

Long term we're more likely to require a local short-term storage structure for any case where Publisher is deployed along with any `cloud` plugin.

Further, I am not concerned regarding backfill, we can make adjustments to backfill if it turns out to be necessary.

The last item to consider is that we _must not_ acknowledge blocks as stored until we are sure those blocks are durable. If we update faster than that, I think we end up with more bugs than we fix, and they're more risky.

AlfredoG87 (approved):
LGTM

PR #3122, opened 2026-06-30
feat: Verify historical WRB by block-range roster

### LINKED ISSUE #2808: WRB Dynamic Roster Selection and Store
### Story Form

As a BlockNode Operator
I want to Make sure we have the historical Roster for all the block ranges
So that We can verify historical blocks using the correct roster that was valid at the time of the block

### Technical Notes

Currently we only have one Roster for Verifying RSA Signatures of WRBs, however the Address Books (ABs) changes over time with the addition on removal of some CNs, in that any case, we need to maintain an historical Roster Collection that allows the BN to select the correct Roster based on the block number to verify.

### LINKED ISSUE #2959: BN: verify historical WRB by block-range roster

**Prereq:** T4.

## Story Form

As a Block Node verifying a historical WRB,
I want to select the address book whose block range contains the block being verified,
So that any historical WRB validates against the node keys that were valid when its record file was signed.

## Technical Notes

- `ExtendedMerkleTreeSession` (both `verification` and `block-verification`) currently uses a single `rsaKeyByNodeId` map; change it to resolve the address book for the block number from the T4 roster.
- A block outside every roster range fails fast with a clear, documented cause (metric + log), not a silent pass.
- Add tests covering multiple eras and the no-matching-entry case.

## Design

berryware:
## Reviewer Notes

- VerificationNotification.java: Added NO_MATCHING_ADDRESS_BOOK failure type
- SessionFailureType.java: Added mapping for the new type
- VerificationServicePlugin (verification): Replaced single keyByNodeId cache + onContextUpdate rebuild with per-block getAddressBookForBlock() lookup
- VerificationDataProvider (block-verification): Replaced AtomicReference<Map> + safeUpdateNodeAddressBook() with rsaPublicKeysForBlock(blockNumber) that delegates to getAddressBookForBlock()
- VerificationServicePlugin (block-verification): Removed onContextUpdate address book update call
- BlockVerifier: Uses rsaPublicKeysForBlock() with null-check → NO_MATCHING_ADDRESS_BOOK
- ExtendedMerkleTreeSession + RSAProofVerifier: Empty key map → NO_MATCHING_ADDRESS_BOOK; unknown node ID → fail-fast
- PluginTestBase + TestApplicationStateFacility: Added getAddressBookForBlock() with inline era-range scan (no JPMS-violating AddressBookHistoryLookup reference)
- Tests updated throughout
- (preserving NO_MATCHING_ADDRESS_BOOK) instead of replacing it with a generic BAD_BLOCK_PROOF.

## Related Issue(s)
Closes #2959 
Closes: #2808


block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionFailureType.java
```diff
@@ -17,6 +17,9 @@ public enum SessionFailureType {
     /// This type indicates we have missing verification data. This includes we do not have TSS Data, RSA public
     /// keys, or we have witnessed a [java.security.NoSuchAlgorithmException].
     MISSING_VERIFICATION_DATA,
+    /// This type indicates that no address book era in the loaded RangedAddressBookHistory covers
+    /// the block number being verified. The block cannot be RSA-verified without the correct era keys.
+    NO_MATCHING_ADDRESS_BOOK,
```
ata-nas: We already have a failure type that covers that - `MISSING_VERIFICATION_DATA`.
I think we should probably use that as we don't gain too much value from being overly specific. In fact, that introduces harder handling and more fragility due to increased complexity.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/BlockVerifier.java
```diff
@@ -97,11 +97,19 @@ private List<ProofVerifier> createVerifiers(final HashingResult hashingResult) t
                         proof.signedBlockProof().blockSignature(),
                         verificationDataProvider));
             } else if (proof.hasSignedRecordFileProof()) {
+                final java.util.Map<Long, java.security.PublicKey> rsaKeys =
```
ata-nas: Nit: we can import Map and PublicKey.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/BlockVerifier.java
```diff
@@ -97,11 +97,19 @@ private List<ProofVerifier> createVerifiers(final HashingResult hashingResult) t
...
+                        verificationDataProvider.rsaPublicKeysForBlock(hashingResult.blockNumber());
+                if (rsaKeys == null) {
+                    throw new VerificationSessionFailedException(
+                            hashingResult.blockNumber(),
+                            SessionFailureType.NO_MATCHING_ADDRESS_BOOK,
+                            hashingResult.blockSource());
+                }
```
ata-nas: Instead of this being here, let's still make the verifier. It will fail if map is empty.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
```diff
@@ -66,14 +66,14 @@ public RSAProofVerifier(
...
         if (rsaKeyByNodeId.isEmpty()) {
             LOGGER.log(
                     ERROR,
-                    "Address book is empty — cannot verify RSA WRB proof for block {0}."
-                            + " Ensure RsaRosterBootstrapPlugin started successfully.",
+                    "No address book era covers block {0} — cannot verify RSA WRB proof."
+                            + " Ensure rsa-address-book-history.json is loaded and covers this block number.",
                     blockNumber);
-            result = SessionFailureType.MISSING_VERIFICATION_DATA;
+            result = SessionFailureType.NO_MATCHING_ADDRESS_BOOK;
```
ata-nas: We can use missing verification data failure type.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
```diff
@@ -94,22 +94,18 @@ public SessionFailureType verify() {
...
-                // @todo(2808): long-term, use a historical roster keyed by block number so signatures
-                //   from nodes that were valid at the time the block was produced are verified correctly
-                //   across address-book transitions, rather than skipping unknown node IDs.
                 final PublicKey publicKey = rsaKeyByNodeId.get(nodeId);
                 if (publicKey == null) {
-                    mismatchCount++;
                     LOGGER.log(
-                            DEBUG,
-                            "Signature from node {0} not in address book (block {1}) — skipped",
+                            WARNING,
```
ata-nas: Not sure about log level being WARNING.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
```diff
@@ -94,22 +94,18 @@ public SessionFailureType verify() {
             // a duplicate entry in the proof from inflating validCount.
             final int rosterSize = rsaKeyByNodeId.size();
             int validCount = 0;
-            int mismatchCount = 0;
             final Set<Long> validatedNodes = new HashSet<>();
             for (final RecordFileSignature sig : proof.recordFileSignatures()) {
                 final long nodeId = sig.nodeId();
-                // @todo(2808): long-term, use a historical roster keyed by block number so signatures
-                //   from nodes that were valid at the time the block was produced are verified correctly
-                //   across address-book transitions, rather than skipping unknown node IDs.
```
ata-nas, line 103: Just to confirm, do we cover the todo of #2808 here? This changes things.

ata-nas (commented):
A few suggestions.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
berryware: Yes, we now lookup the AddressBook in the RosterAddressBookHistory based on blocknumber and retrieve the key from the matching addressbook.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
berryware: Reverted to DEBUG

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
berryware: Changed to use MISSING_VERIFICATION_DATA

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/BlockVerifier.java
berryware: Removed. Good call.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/BlockVerifier.java
berryware: Done

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionFailureType.java
berryware: Removed.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
ata-nas: Is that ticket closeable or is there more to #2808?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
```diff
@@ -66,12 +66,12 @@ public RSAProofVerifier(
     @Override
     public SessionFailureType verify() {
         final SessionFailureType result;
-        // Guard: address book must be loaded before any WRB arrives
+        // Guard: no era in the address book history covers this block number
         if (rsaKeyByNodeId.isEmpty()) {
             LOGGER.log(
                     ERROR,
```
jsync-swirlds: This is non-fatal; so a WARNING is perhaps more appropriate.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationDataProvider.java
```diff
@@ -50,9 +48,21 @@ public boolean hasTssData() {
...
+        final NodeAddressBook book = context.applicationStateFacility().getAddressBookForBlock(blockNumber);
+        if (book == null) {
+            return null;
+        }
+        try {
+            return buildKeyMap(book);
+        } catch (final NoSuchAlgorithmException e) {
+            LOGGER.log(WARNING, "RSA KeyFactory not available for block {0} — returning empty key map", blockNumber);
+            return null;
+        }
```
jsync-swirlds, line 79: Can we cache the Map produced and the address book it corresponds to?
It would be good to not need to recreate the map every block when address books typically change only around once every 40,000-1,000,000 blocks.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
berryware: It is closeable and added as so to this ticket.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
```diff
@@ -41,6 +41,9 @@ public enum FailureType {
         /// This type indicates we have missing verification data. This includes we do not have TSS Data, RSA public
         /// keys, or we have witnessed a [java.security.NoSuchAlgorithmException].
         MISSING_VERIFICATION_DATA,
+        /// This type indicates that no address book era in the loaded RangedAddressBookHistory covers
+        /// the block number being verified. The block cannot be RSA-verified without the correct era keys.
+        NO_MATCHING_ADDRESS_BOOK,
```
jsync-swirlds: This duplicates `MISSING_VERIFICATION_DATA`.  Do we need the extra return value?

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -497,23 +497,20 @@ private VerificationNotification verifyRsaProof(
         // a duplicate entry in the proof from inflating validCount.
         final int rosterSize = rsaKeyByNodeId.size();
         int validCount = 0;
-        int mismatchCount = 0;
```
jsync-swirlds, line 500: We still need to handle mismatches and allow for them.
No matter how aggressive we might be about obtaining address book updates (and we cannot be very aggressive at all) there will always be gaps and times when we do not have the latest address book.

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
```diff
@@ -565,13 +562,8 @@ private VerificationNotification verifyRsaProof(
             }
         }
 
-        if (mismatchCount > 0) {
-            metrics.incrementRsaRosterMismatch(mismatchCount);
-        }
-
         // Acceptance threshold: every signature present in the proof passes validation,
-        // and at least one such signature exists. Signatures from nodes not in the local
-        // roster are skipped (see TODO(2808)). Because we fail fast on any failed
```
jsync-swirlds, line 574: We do still skip signatures not in the local roster; particularly the _latest_ roster, because that one can be a little out of date and still be used.

block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
```diff
@@ -536,14 +495,18 @@ public void handleBackfilled(BackfilledBlockNotification notification) {
                 LOGGER.log(WARNING, message, notification.blockNumber(), blockHeader.number());
                 sendFailureNotification(notification.blockNumber(), BlockSource.BACKFILL, FailureType.BAD_BLOCK_PROOF);
             } else {
+                final NodeAddressBook backfillAddressBook =
+                        context.applicationStateFacility().getAddressBookForBlock(notification.blockNumber());
+                final Map<Long, PublicKey> backfillRsaKeys =
+                        backfillAddressBook != null ? RsaKeyDecoder.buildKeyMap(backfillAddressBook) : Map.of();
```
jsync-swirlds: Some level of basic caching for the key map would probably be helpful, even with backfill.

jsync-swirlds (commented):
A few questions and suggestions.

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/AllBlocksHasherHandlerTest.java
ata-nas, line 1: Let's not add anything related to all blocks hasher. This class will become a standalone plugin and is completely non-operational. I am thinking about completely deleting it at this point...

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
ata-nas: I though we reverted this.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/BlockVerifier.java
```diff
@@ -101,11 +103,13 @@ private List<ProofVerifier> createVerifiers(final HashingResult hashingResult) t
...
                         verificationDataProvider));
             } else if (proof.hasSignedRecordFileProof()) {
+                final Map<Long, PublicKey> rsaKeys =
+                        verificationDataProvider.rsaPublicKeysForBlock(hashingResult.blockNumber());
                 final Signature sha384WithRSASig = Signature.getInstance("SHA384withRSA");
                 verifiers.add(new RSAProofVerifier(
                         proofVerificationMetrics,
                         hashingResult.blockNumber(),
-                        verificationDataProvider.currentRSAPublicKeys(),
+                        rsaKeys,
```
ivannov, line 112: If this is null (and as far as I can see null is a valid value here), this constrictor has requiresNonNull guard. Maybe that guard should be removed now null is a valid value for the `rsaKeys`?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
```diff
@@ -66,12 +66,12 @@ public RSAProofVerifier(
     @Override
     public SessionFailureType verify() {
         final SessionFailureType result;
-        // Guard: address book must be loaded before any WRB arrives
+        // Guard: no era in the address book history covers this block number
```
ivannov, line 69: What if `rsaKeyByNodeId` is null?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
```diff
@@ -94,22 +94,18 @@ public SessionFailureType verify() {
...
-                    mismatchCount++;
                     LOGGER.log(
                             DEBUG,
-                            "Signature from node {0} not in address book (block {1}) — skipped",
+                            "Signature from node {0} not in era address book for block {1} — rejecting block",
                             nodeId,
                             blockNumber);
-                    continue;
+                    proofVerificationMetrics.rsaFailure().increment();
+                    return SessionFailureType.BAD_BLOCK_PROOF;
```
ivannov: Nit: there is an inconsistency when e directly return and when we assign a `SessionFailureType` to the `result` variable. Is there a special reason why we directly return here or not just continue or break the loop?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
```diff
@@ -66,12 +66,12 @@ public RSAProofVerifier(
...
+        // Guard: no era in the address book history covers this block number
         if (rsaKeyByNodeId.isEmpty()) {
             LOGGER.log(
                     ERROR,
-                    "Address book is empty — cannot verify RSA WRB proof for block {0}."
-                            + " Ensure RsaRosterBootstrapPlugin started successfully.",
+                    "No address book era covers block {0} — cannot verify RSA WRB proof."
+                            + " Ensure rsa-address-book-history.json is loaded and covers this block number.",
                     blockNumber);
             result = SessionFailureType.MISSING_VERIFICATION_DATA;
```
ivannov, line 76: The check in the old verification plugin would return the `NO_MATCHING_ADDRESS_BOOK` failure type in this case. Was it a miss here?

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/AllBlocksHasherHandlerTest.java
berryware: I needed to increase the test coverage. I am okay with deleting this later.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
berryware: NO_MATCHING_ADDRESS_BOOK is removed based on above comments of being a duplicate to MISSING_VERIFICATION_DATA

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
berryware: Reverted to continue;

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
berryware: Added a null check

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/BlockVerifier.java
berryware: Removed the guard

block-node/verification/src/main/java/org/hiero/block/node/verification/VerificationServicePlugin.java
berryware: Added caching in

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
berryware: Modified the comment

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
berryware: Added back in

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/VerificationNotification.java
berryware: It has been reverted again. Merge conflicts had put it back

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationDataProvider.java
berryware: Added a cache for the most recent NodeAddressBook, and only recreate when it changes.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
berryware: Changed to WARNING

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/AllBlocksHasherHandlerTest.java
ata-nas: I am not sure we need to increase coverage. We do not have a blocker on no coverage.

jsync-swirlds (dismissed):
LGTM.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/BlockVerifier.java
ata-nas: RSA keys should not be null in general. If we have missing ones, let's return an empty map.

I am a bit hesitant to removing a null check. It might seem safer, but it also can indicate a bug if we supply a null map.

At least that is my opinion, but this is not a blocker.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationDataProvider.java
```diff
@@ -50,9 +54,31 @@ public boolean hasTssData() {
...
+        if (cached != null && cached.book() == book) {
+            return cached.keys();
+        }
+        try {
+            final Map<Long, PublicKey> keys = buildKeyMap(book);
+            cachedKeyMap.set(new CachedKeyMap(book, keys));
+            return keys;
+        } catch (final NoSuchAlgorithmException e) {
+            LOGGER.log(WARNING, "RSA KeyFactory not available for block {0} — returning empty key map", blockNumber);
+            return null;
```
ata-nas: As mentioned above, we can simply return an empty map instead of null as that would be much safer in general.

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/AllBlocksHasherHandlerTest.java
berryware:
```[codecov/patch](https://github.com/hiero-ledger/hiero-block-node/pull/3122/checks?check_run_id=84896358911)
codecov/patch — 76.47% of diff hit (target 80.00%)```

jsync-swirlds (dismissed):

jsync-swirlds (dismissed):

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/AllBlocksHasherHandlerTest.java
berryware: removed the file

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/BlockVerifier.java
berryware: Returns an empty Map now

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RSAProofVerifier.java
berryware: removed null check. Returns an empty map now

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationDataProvider.java
berryware: returns an empty map now

jsync-swirlds (dismissed):

jsync-swirlds (approved):

ivannov (approved):

PR #3124, opened 2026-06-30
feat: live-push each new WRB to a Block Node (#2956)

### LINKED ISSUE #2956: CLI: live-push each new WRB to a Block Node

**Prereq:** T1 (reuses its publish client).

## Story Form

As a node operator,
I want the live wrap pipeline to push each WRB to the Block Node immediately as it is produced,
So that the special-purpose BN stays continuously current during the transition without 6-hour batch floods.

## Technical Notes

- Extend the live pipeline (`days live-sequential` / `download-live2`) with an option to publish each wrapped block to the BN as it is produced, bypassing the in-memory 6-hour batching for the distribution path only.
- Reuse T1's publish client. Idempotent against already-stored blocks; retry/backoff on failure.
- Must coexist with the CLI's normal zip output (no shared-write contention).

## Design

rockysingh:
## Summary

- Adds an opt-in path to the `days live-sequential` command that pushes each freshly wrapped & validated `Block` to a Block Node via the standard `BlockStreamPublishService` gRPC stream, one block at a time, as it is produced.
- New `org.hiero.block.tools.push.LiveBlockPushClient` encapsulates the persistent publish stream, decoupling the producer (wrap thread) from the network via a bounded queue so the local zip output path is never delayed by transient BN slowness.
- Gated by an explicit `--push-enabled` flag so the host can be configured ahead of time but the feature stays off until the operator flips it on after the historical backfill (#2955 / #3038) has populated the target BN.

Closes #2956.

## Why

Part of epic #2954 — _Distribute historical & live WRBs and TSS to Tier-1 Block Nodes_. The special-purpose ("Tier 0") BN is seeded via bulk-load for history; this PR provides the steady-state live feed so the BN stays continuously current without the 6-hour zip-batch flood pattern.

### Push client (`LiveBlockPushClient`)

- Builds the gRPC publish stream on `PbjGrpcClient` + Helidon WebClient, matching the existing `NetworkCapacityClient` plumbing.
- Background worker thread drains an `ArrayBlockingQueue<QueuedBlock>` into the stream. Each block is batched into `PublishStreamRequest`s respecting `serverMaxMessageSizeBytes`.
- **"Never drop" guarantee:** blocks remain in an in-flight deque until ACKed. On stream error the worker reconnects with exponential backoff (250 ms → 30 s cap) and re-sends in-flight blocks before resuming the queue. If the queue fills (sustained BN outage) the producer's `pushBlock` call blocks rather than dropping.
- One-shot `queryLastAvailableBlock()` via `BlockNodeService.serverStatus` so the producer can skip blocks the BN already has on resume.
- `loadDefaultWebConfig()` reads the bundled `clientDefaultConfig.json` so callers don't have to hand-build a `HelidonWebClientConfig`.

### Wiring in `LiveSequential`

- New CLI options:
  - `--push-enabled` (bool, default `false`) — explicit gate
  - `--push-bn-host` (string) — BN host; required when `--push-enabled` is set
  - `--push-bn-port` (int, default `40840`)
  - `--push-queue-capacity` (int, default `32`)
- Validation:
  - `--push-enabled` without `--push-bn-host` → fail fast with `IllegalArgumentException`
  - `--push-bn-host` set but `--push-enabled` not set → log "configured but disabled" and proceed in zip-only mode
- Hook lives at the end of `wrapAndValidateBlock`, after the block is fully validated and committed to the zip. Blocks `<= lastAvailableBlock` are skipped from the push path; subsequent blocks go onto the push queue.
- Push client is started before the wrap thread and drained both on graceful completion and via the shutdown hook.

## Operational flow

1. Deploy with `--push-bn-host=<bn>` baked in, `--push-enabled` off → wrap keeps producing the local zip; BN gets backfilled via `blocks bulk-load` (#3038).
2. When backfill catches up, restart with `--push-enabled` → live feed starts from `lastAvailableBlock + 1`.

## BN side

No BN changes required. We hit:
- `stream-publisher` plugin → `BlockStreamPublishService.publishBlockStream` (already in default `values.yaml`)
- `server-status` plugin → `BlockNodeService.serverStatus` (already in default `values.yaml`)

For WRB-mode BNs, `roster-bootstrap-rsa` must be appended to the plugin list (this is already the operational pattern in `solo-deploy-network.sh` for rsa-wrb mode).

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
@@ -94,6 +94,7 @@ import org.hiero.block.tools.mirrornode.FetchBlockQuery;
 import org.hiero.block.tools.mirrornode.FixBlockTime;
 import org.hiero.block.tools.mirrornode.MirrorNodeBlockQueryOrder;
 import org.hiero.block.tools.mirrornode.UpdateBlockData;
+import org.hiero.block.tools.push.LiveBlockPushClient;
 import org.hiero.block.tools.records.RecordFileUtils;
 import org.hiero.block.tools.records.model.parsed.ParsedRecordBlock;
 import org.hiero.block.tools.records.model.parsed.RecordBlockConverter;
@@ -180,6 +181,30 @@ public class LiveSequential implements Runnable {
             names = {"--start-date"},
             description = "Start date in YYYY-MM-DD format (default: auto-detect from mirror node)")
     private String startDate;
+
+    @Option(
+            names = {"--push-bn-host"},
+            description = "Optional. If set, each wrapped block is pushed to this Block Node "
+                    + "host:port via the Publish gRPC stream as it is produced.")
+    private String pushBnHost;
+
+    @Option(
+            names = {"--push-bn-port"},
+            description =
+                    "Block Node Publish port to push to (default: 40840). Ignored if " + "--push-bn-host is not set.")
+    private int pushBnPort = 40840;
+
+    @Option(
+            names = {"--push-queue-capacity"},
+            description = "Bounded backpressure queue between the wrap thread and the push worker "
+                    + "(default: 32). When full, the wrap thread blocks; nothing is dropped.")
+    private int pushQueueCapacity = 32;
+
+    /** Live-push client; null when --push-bn-host is not set. */
+    private LiveBlockPushClient pushClient;
+
+    /** Blocks at or below this number are already on the BN; skip pushing them. */
+    private long pushSkipUpToInclusive = -1L;
 
     /** State persisted to JSON for resumability. Compatible with DownloadLive2 format. */
     private static class State {
@@ -316,6 +341,21 @@ public class LiveSequential implements Runnable {
                     + "]"
                     + " day=" + initialState.dayDate);
 
+            // Optional: live-push to a Block Node. If --push-bn-host was provided, set up the
+            // push client so the wrap thread can hand each validated block off to it as it is
+            // produced. The push subsystem owns its own thread + bounded queue (see
+            // LiveBlockPushClient javadoc).
+            if (pushBnHost != null && !pushBnHost.isBlank()) {
+                pushClient = new LiveBlockPushClient(
+                        pushBnHost, pushBnPort, pushQueueCapacity, LiveBlockPushClient.loadDefaultWebConfig());
+                pushSkipUpToInclusive = pushClient.queryLastAvailableBlock();
+                System.out.println("[live-sequential] Live push enabled: target=" + pushBnHost + ":"
+                        + pushBnPort + " queueCapacity=" + pushQueueCapacity
+                        + " BN lastAvailableBlock=" + pushSkipUpToInclusive
+                        + " (blocks <= this are skipped from push)");
+                pushClient.start();
+            }
+
             // Create producer-consumer queue
             final BlockingQueue<ValidatedBlock> queue = new LinkedBlockingQueue<>(32);
             final AtomicReference<Throwable> wrapError = new AtomicReference<>(null);
@@ -341,6 +381,12 @@ public class LiveSequential implements Runnable {
                             () -> {
                                 System.out.println("[live-sequential] Shutdown requested...");
                                 downloadManager.close();
+                                if (pushClient != null) {
+                                    System.out.println("[live-sequential] Draining live-push client (queue="
+                                            + pushClient.queueDepth() + " acked=" + pushClient.acked()
+                                            + " submitted=" + pushClient.submitted() + ")...");
+                                    pushClient.shutdown();
+                                }
                             },
                             "live-sequential-shutdown"));
 
@@ -356,6 +402,11 @@ public class LiveSequential implements Runnable {
             Throwable wrapErr = wrapError.get();
             if (wrapErr != null) {
                 throw new RuntimeException("Wrap+validate thread failed", wrapErr);
+            }
+
+            if (pushClient != null) {
+                System.out.println("[live-sequential] Draining live-push client...");
+                pushClient.shutdown();
             }
 
             System.out.println("[live-sequential] Complete.");
@@ -1106,6 +1157,19 @@ public class LiveSequential implements Runnable {
 
         // Run all validation checks
         runBlockValidations(wrappedBytes, blockNum, vc, ls, checkpointDir);
+
+        // Live-push to a Block Node if configured. Done after the block is fully validated and
+        // committed to the local zip; the push subsystem is fully decoupled (its own thread +
+        // bounded queue) so a slow/unreachable BN only stalls the push, not the local pipeline,
+        // unless the queue fills (in which case the wrap thread blocks here rather than dropping).
+        if (pushClient != null && blockNum > pushSkipUpToInclusive) {
+            try {
+                pushClient.pushBlock(blockNum, wrapped);
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+                throw ie;
+            }
+        }
 
         // Collect signature stats
         SignatureBlockStats blockStats = vc.signatureValidation().popBlockStats(blockNum);
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
@@ -362,10 +362,9 @@ public class LiveSequential implements Runnable {
                 pushClient = new LiveBlockPushClient(
                         pushBnHost, pushBnPort, pushQueueCapacity, LiveBlockPushClient.loadDefaultWebConfig());
                 pushSkipUpToInclusive = pushClient.queryLastAvailableBlock();
-                System.out.println("[live-sequential] Live push enabled: target=" + pushBnHost + ":"
-                        + pushBnPort + " queueCapacity=" + pushQueueCapacity
-                        + " BN lastAvailableBlock=" + pushSkipUpToInclusive
-                        + " (blocks <= this are skipped from push)");
+                System.out.printf(
+                        "[live-sequential] Live push enabled: target=%s:%d queueCapacity=%d BN lastAvailableBlock=%d (blocks <= this are skipped from push)%n",
+                        pushBnHost, pushBnPort, pushQueueCapacity, pushSkipUpToInclusive);
                 pushClient.start();
             }
 
@@ -1176,12 +1175,8 @@ public class LiveSequential implements Runnable {
         // bounded queue) so a slow/unreachable BN only stalls the push, not the local pipeline,
         // unless the queue fills (in which case the wrap thread blocks here rather than dropping).
         if (pushClient != null && blockNum > pushSkipUpToInclusive) {
-            try {
-                pushClient.pushBlock(blockNum, wrapped);
-            } catch (InterruptedException ie) {
-                Thread.currentThread().interrupt();
-                throw ie;
-            }
+            // Method declares `throws Exception` so InterruptedException propagates naturally.
+            pushClient.pushBlock(blockNum, wrapped);
         }
 
         // Collect signature stats
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
@@ -192,9 +192,10 @@ public class LiveSequential implements Runnable {
 
     @Option(
             names = {"--push-bn-host"},
-            description = "Block Node host to push to. Required when --push-enabled is set. Can be "
-                    + "configured ahead of time and gated on later via --push-enabled.")
-    private String pushBnHost;
+            description = "Block Node host to push to (default: localhost). The expected deployment "
+                    + "is a co-located BN, so localhost is the common case; override only when the "
+                    + "target BN runs elsewhere.")
+    private String pushBnHost = "localhost";
 
     @Option(
             names = {"--push-bn-port"},
@@ -358,9 +359,6 @@ public class LiveSequential implements Runnable {
             // side coordinating with the other. The push subsystem owns its own thread + bounded
             // queue (see LiveBlockPushClient javadoc).
             if (pushEnabled) {
-                if (pushBnHost == null || pushBnHost.isBlank()) {
-                    throw new IllegalArgumentException("--push-enabled requires --push-bn-host to be set");
-                }
                 pushClient = new LiveBlockPushClient(
                         pushBnHost, pushBnPort, pushQueueCapacity, LiveBlockPushClient.loadDefaultWebConfig());
                 pushSkipUpToInclusive = pushClient.queryLastAvailableBlock();
@@ -369,10 +367,6 @@ public class LiveSequential implements Runnable {
                         + " BN lastAvailableBlock=" + pushSkipUpToInclusive
                         + " (blocks <= this are skipped from push)");
                 pushClient.start();
-            } else if (pushBnHost != null && !pushBnHost.isBlank()) {
-                System.out.println("[live-sequential] Live push configured (host=" + pushBnHost + ":"
-                        + pushBnPort + ") but disabled (--push-enabled not set). "
-                        + "Run with --push-enabled to turn on after backfill completes.");
             }
 
             // Create producer-consumer queue
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
@@ -351,9 +351,12 @@ public class LiveSequential implements Runnable {
                     + " day=" + initialState.dayDate);
 
             // Optional: live-push to a Block Node. Gated by --push-enabled so the host can be
-            // configured ahead of time but the feature stays off until the operator flips it on
-            // (typically after the historical backfill has populated the target BN). The push
-            // subsystem owns its own thread + bounded queue (see LiveBlockPushClient javadoc).
+            // configured ahead of time but the feature stays off until the operator flips it on.
+            // Live push can run concurrently with a separate historical-backfill process pointing
+            // at the same BN — queryLastAvailableBlock() below skips blocks the BN already has, so
+            // the live side naturally sits above the historical high-water mark without either
+            // side coordinating with the other. The push subsystem owns its own thread + bounded
+            // queue (see LiveBlockPushClient javadoc).
             if (pushEnabled) {
                 if (pushBnHost == null || pushBnHost.isBlank()) {
                     throw new IllegalArgumentException("--push-enabled requires --push-bn-host to be set");
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
@@ -183,24 +183,33 @@ public class LiveSequential implements Runnable {
     private String startDate;
 
     @Option(
+            names = {"--push-enabled"},
+            description = "Enable live-push of each wrapped block to a Block Node as it is produced. "
+                    + "Off by default. Operationally, leave this off until the historical backfill "
+                    + "(see 'blocks bulk-load') has populated the target BN; then turn this on for "
+                    + "the steady-state live feed. Requires --push-bn-host.")
+    private boolean pushEnabled = false;
+
+    @Option(
             names = {"--push-bn-host"},
-            description = "Optional. If set, each wrapped block is pushed to this Block Node "
-                    + "host:port via the Publish gRPC stream as it is produced.")
+            description = "Block Node host to push to. Required when --push-enabled is set. Can be "
+                    + "configured ahead of time and gated on later via --push-enabled.")
     private String pushBnHost;
 
     @Option(
             names = {"--push-bn-port"},
             description =
-                    "Block Node Publish port to push to (default: 40840). Ignored if " + "--push-bn-host is not set.")
+                    "Block Node Publish port to push to (default: 40840). Ignored unless " + "--push-enabled is set.")
     private int pushBnPort = 40840;
 
     @Option(
             names = {"--push-queue-capacity"},
             description = "Bounded backpressure queue between the wrap thread and the push worker "
-                    + "(default: 32). When full, the wrap thread blocks; nothing is dropped.")
+                    + "(default: 32). When full, the wrap thread blocks; nothing is dropped. "
+                    + "Ignored unless --push-enabled is set.")
     private int pushQueueCapacity = 32;
 
-    /** Live-push client; null when --push-bn-host is not set. */
+    /** Live-push client; null unless push is enabled and configured. */
     private LiveBlockPushClient pushClient;
 
     /** Blocks at or below this number are already on the BN; skip pushing them. */
@@ -341,11 +350,14 @@ public class LiveSequential implements Runnable {
                     + "]"
                     + " day=" + initialState.dayDate);
 
-            // Optional: live-push to a Block Node. If --push-bn-host was provided, set up the
-            // push client so the wrap thread can hand each validated block off to it as it is
-            // produced. The push subsystem owns its own thread + bounded queue (see
-            // LiveBlockPushClient javadoc).
-            if (pushBnHost != null && !pushBnHost.isBlank()) {
+            // Optional: live-push to a Block Node. Gated by --push-enabled so the host can be
+            // configured ahead of time but the feature stays off until the operator flips it on
+            // (typically after the historical backfill has populated the target BN). The push
+            // subsystem owns its own thread + bounded queue (see LiveBlockPushClient javadoc).
+            if (pushEnabled) {
+                if (pushBnHost == null || pushBnHost.isBlank()) {
+                    throw new IllegalArgumentException("--push-enabled requires --push-bn-host to be set");
+                }
                 pushClient = new LiveBlockPushClient(
                         pushBnHost, pushBnPort, pushQueueCapacity, LiveBlockPushClient.loadDefaultWebConfig());
                 pushSkipUpToInclusive = pushClient.queryLastAvailableBlock();
@@ -354,6 +366,10 @@ public class LiveSequential implements Runnable {
                         + " BN lastAvailableBlock=" + pushSkipUpToInclusive
                         + " (blocks <= this are skipped from push)");
                 pushClient.start();
+            } else if (pushBnHost != null && !pushBnHost.isBlank()) {
+                System.out.println("[live-sequential] Live push configured (host=" + pushBnHost + ":"
+                        + pushBnPort + ") but disabled (--push-enabled not set). "
+                        + "Run with --push-enabled to turn on after backfill completes.");
             }
 
             // Create producer-consumer queue
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
@@ -1,0 +1,429 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.tools.push;
+
+import com.hedera.hapi.block.stream.Block;
+import com.hedera.hapi.block.stream.BlockItem;
+import com.hedera.pbj.grpc.client.helidon.PbjGrpcClient;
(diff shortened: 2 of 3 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
```diff
@@ -316,6 +350,28 @@ public void run() {
                     + "]"
                     + " day=" + initialState.dayDate);
 
+            // Optional: live-push to a Block Node. Gated by --push-enabled so the host can be
+            // configured ahead of time but the feature stays off until the operator flips it on
+            // (typically after the historical backfill has populated the target BN). The push
```
Nana-EC: To confirm, is it possible to have both live and historical push running in parallel?
This should be the goal.
The flag is good for operators who may want to completely backfill first as part of their process.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    private void sendBlock(final StreamSession session, final QueuedBlock qb) {
+        final List<BlockItem> items = qb.block.items();
+        final List<List<BlockItem>> batches = new ArrayList<>();
+        List<BlockItem> current = new ArrayList<>();
+        long currentSize = 0;
+        for (final BlockItem item : items) {
+            final int itemSize = BlockItem.PROTOBUF.measureRecord(item);
+            if (!current.isEmpty() && currentSize + itemSize > maxBatchBytes) {
```
Nana-EC, line 307: Unlikely to happen but we should also check if the item itself is oversized.
If so we should probably stop as the BN won't accept it, or we should at least log so we're aware the scenario we're in.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,429 @@
...
+    private final long retryMaxMs;
+
+    private final AtomicBoolean running = new AtomicBoolean(false);
+    private final AtomicLong submitted = new AtomicLong();
+    private final AtomicLong acked = new AtomicLong();
+    private final AtomicLong lastAckedBlock = new AtomicLong(-1);
+    private final AtomicLong streamReconnects = new AtomicLong();
+
+    private Thread worker;
+    private WebClient cachedWebClient; // recreated on each reconnect
```
Nana-EC: I think the comment is wrong, it's lazily created once and reused.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,429 @@
...
+        if (!current.isEmpty()) {
+            batches.add(current);
+        }
+        for (final List<BlockItem> batch : batches) {
+            final BlockItemSet set = BlockItemSet.newBuilder().blockItems(batch).build();
+            final PublishStreamRequest req =
+                    PublishStreamRequest.newBuilder().blockItems(set).build();
+            session.requestPipeline.onNext(req);
+        }
+        submitted.incrementAndGet();
```
Nana-EC, line 334: Should submitted be unique across blocks?
For example if there's a reattempt of the same block with a successful ack you'd increment this twice and ack once. The result is the logs would make it seem like there was a block that missed acknowledgment.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,429 @@
...
+                        break;
+                    }
+                } while (!lastAckedBlock.compareAndSet(prev, blockNumber));
+            }
+        }
+
+        @Override
+        public void onError(final Throwable throwable) {
+            completion.completeExceptionally(throwable);
+            throw new StreamFailure(throwable);
```
Nana-EC: I think this is thrown on Helion's I/O thread right?
Do we know that it exposes this and doesn't silently absorbs the exception?

If we're sure then leave as is.
If not maybe consider a `volatile boolean streamFailed` that we check before polling the `queue`. You can consider other alternatives to make sure failure detection is reliable

Nana-EC (commented):
Some comments

AlfredoG87 (dismissed):
Nice Work!

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
rockysingh: Confirmed yes. Live push and a separate historical backfill process can run against the same BN in parallel — nothing here prevents it. `queryLastAvailableBlock()` at start-up gives the live side a floor equal to the BN's high-water mark, so it naturally sits above whatever the historical process is filling in. Updated the comment block above to say so explicitly (69499c7f).

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Good catch. Added an explicit log line when `itemSize > maxBatchBytes` so if it ever happens the operator sees why the BN rejected the frame instead of chasing a mysterious error (69499c7f). No stop-and-drain since the BN's rejection will trip the stream anyway and the reconnect path handles it.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: You're right, the comment was wrong. Fixed to "lazily created once and reused for every session" (69499c7f).

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Fixed. Added `highestSubmittedBlock` (AtomicLong, initialised to -1) and `submitted` now only increments when a block's number is strictly greater than the previous high water mark. Resends after a stream reconnect walk in-flight in ascending block-number order so nothing above the mark ever gets recounted, and the operator log no longer shows a phantom `submitted > acked` gap after a reconnect (69499c7f).

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Good instinct. Went with the `volatile` flag approach so we don't have to depend on Helidon's error-propagation semantics. `AckPipeline` now flips a `streamAlive` AtomicBoolean to false (both in onError and onComplete) and stashes the cause; the worker loop replaces `queue.take()` with `queue.poll(200ms)` and re-throws `StreamFailure` when the flag drops. So even if Helidon absorbs the throw, the worker sees the transition within 200ms and reconnects (69499c7f).

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
```diff
@@ -181,6 +182,39 @@ public class LiveSequential implements Runnable {
...
+                    + "Off by default. Operationally, leave this off until the historical backfill "
+                    + "(see 'blocks bulk-load') has populated the target BN; then turn this on for "
+                    + "the steady-state live feed. Requires --push-bn-host.")
+    private boolean pushEnabled = false;
+
+    @Option(
+            names = {"--push-bn-host"},
+            description = "Block Node host to push to. Required when --push-enabled is set. Can be "
+                    + "configured ahead of time and gated on later via --push-enabled.")
+    private String pushBnHost;
```
jsync-swirlds: Rather than making this required, and failing if not set, could we set the default to `localhost` or `::1`?  In the expected use case we'll be feeding localhost every time.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
```diff
@@ -316,6 +350,31 @@ public void run() {
...
+                if (pushBnHost == null || pushBnHost.isBlank()) {
+                    throw new IllegalArgumentException("--push-enabled requires --push-bn-host to be set");
+                }
+                pushClient = new LiveBlockPushClient(
+                        pushBnHost, pushBnPort, pushQueueCapacity, LiveBlockPushClient.loadDefaultWebConfig());
+                pushSkipUpToInclusive = pushClient.queryLastAvailableBlock();
+                System.out.println("[live-sequential] Live push enabled: target=" + pushBnHost + ":"
+                        + pushBnPort + " queueCapacity=" + pushQueueCapacity
+                        + " BN lastAvailableBlock=" + pushSkipUpToInclusive
+                        + " (blocks <= this are skipped from push)");
```
jsync-swirlds: Might be easier to read if we use `printf`. Same for the other println uses below.

```suggestion
                final String message = "[live-sequential] Live push enabled: target=%s:%d queueCapacity=%d BN lastAvailableBlock=%d (blocks <= this are skipped from push)%n";
                System.out.printf(message, pushBnHost, pushBnPort, pushQueueCapacity, pushSkipUpToInclusive);
```

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
```diff
@@ -1107,6 +1177,19 @@ private void wrapAndValidateBlock(
...
+        // Live-push to a Block Node if configured. Done after the block is fully validated and
+        // committed to the local zip; the push subsystem is fully decoupled (its own thread +
+        // bounded queue) so a slow/unreachable BN only stalls the push, not the local pipeline,
+        // unless the queue fills (in which case the wrap thread blocks here rather than dropping).
+        if (pushClient != null && blockNum > pushSkipUpToInclusive) {
+            try {
+                pushClient.pushBlock(blockNum, wrapped);
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+                throw ie;
```
jsync-swirlds: If you're going to throw, just do not catch.
In almost all cases, it is best to only catch an exception if we can do something meaningful with it.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,481 @@
...
+     * don't have to hand-construct a {@link HelidonWebClientConfig}.
+     */
+    public static HelidonWebClientConfig loadDefaultWebConfig() {
+        try (final var stream =
+                LiveBlockPushClient.class.getClassLoader().getResourceAsStream("clientDefaultConfig.json")) {
+            if (stream == null) {
+                throw new IllegalStateException("clientDefaultConfig.json not found on classpath");
+            }
+            return HelidonWebClientConfig.JSON.parse(Bytes.wrap(stream.readAllBytes()));
+        } catch (final Exception e) {
```
jsync-swirlds: Catch `RuntimeException` in catch-all; that ensures we don't accidentally miss a new checked exception that may have a different treatment.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,481 @@
...
+                        prunePendingFromAck(inFlight);
+                    }
+                    // Wait for all ACKs (deadline) before closing.
+                    waitForPending(inFlight);
+                    session.requestPipeline.onComplete();
+                    try {
+                        session.completion.get(2, TimeUnit.MINUTES);
+                    } catch (final Exception ignored) {
+                        // best effort
+                    }
```
jsync-swirlds, line 251: Narrow try-catch is a code smell.
Couldn't we just let the outer try-catch handle any exceptions thrown from this line?

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,481 @@
...
+            } catch (final Exception e) {
+                streamReconnects.incrementAndGet();
+                System.err.println("[push] unexpected worker error: " + e.getMessage() + "; reconnecting after "
+                        + backoffMs + "ms");
+                try {
+                    Thread.sleep(backoffMs);
+                } catch (final InterruptedException ie) {
+                    Thread.currentThread().interrupt();
+                    return;
+                }
```
jsync-swirlds: This is an ugly pattern; perhaps use the following cleaner alternative:

```suggestion
                LockSupport.parkNanos(backoffMs * 1_000_000L);
```

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
rockysingh: Good call. Defaulted `--push-bn-host` to `localhost` in 61ed00fb and dropped the fail-without-it check. Remote-BN callers still pass the flag explicitly; the co-located case just doesn't need it any more.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
jsync-swirlds: Generally, we should _not_ throw from this method.
We should log the exception, instruct the stream to complete, and return clean.
The stream will end cleanly for this client if we handle this properly.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,481 @@
...
+        }
+    }
+
+    /**
+     * Tracks ACKs and signals failure on stream errors. Runs on Helidon's response-processing thread,
+     * which may or may not surface exceptions thrown from {@link #onError}. To make failure detection
+     * robust from the worker thread, {@link #streamAlive} is flipped to {@code false} before the
+     * exception is thrown; the worker polls it in the drain loop and reconnects even if Helidon
+     * swallowed the throw.
+     */
```
jsync-swirlds, line 473: Note, the publish protocol responds with a lot more than acknowledgements.
The BN may respond with Skip, resend, behind, etc..., and will send a "EndOfStream(SUCCESS)" response after the connection is open for a few days (by default, less if configured for shorter max stream duration).

jsync-swirlds (commented):
A few places where things are unclear, or improvement is possible.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
rockysingh: Applied in 4db8646a. Replaced the concatenated println with a printf format string `[live-sequential] Live push enabled: target=%s:%d queueCapacity=%d BN lastAvailableBlock=%d (blocks <= this are skipped from push)%n`.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
rockysingh: Agreed. Dropped the try/catch in 4db8646a. `wrapAndValidateBlock` already declares `throws Exception`, so InterruptedException propagates naturally without the flag-set-then-rethrow dance.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Fixed in 4db8646a. Replaced `catch (Exception)` with `catch (IOException | ParseException | RuntimeException)`. Named the checked exceptions explicitly so a new one introduced upstream will break the build instead of getting silently absorbed.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Kept a narrow catch but justified it in 4db8646a. We're in the graceful shutdown branch (`return;` right after), so letting these exceptions flow to the outer reconnect path would be wrong. Narrowed `catch (Exception ignored)` to `catch (ExecutionException | TimeoutException)` and now log the specific failure instead of silently swallowing.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Applied in 4db8646a. Both `Thread.sleep` sites collapsed into a `backoffAndAdvance` helper that uses `LockSupport.parkNanos`. Park returns silently on interrupt with the flag still set, so the caller just checks `isInterrupted()` and returns. Removed the nested try/catch entirely.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Done in 4db8646a. onError no longer throws. It logs, flips `streamAlive` to false, and completes the future exceptionally. The worker's poll loop observes `streamAlive == false` and re-throws StreamFailure on its own thread (the correct place for reconnect logic), so we don't lose the failure signal.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Good point. Added explicit handling in 4db8646a. onNext now recognises `SkipBlock`, `ResendBlock`, `NodeBehindPublisher`, and `EndOfStream`. Each variant logs so the operator can see what the BN is telling us; `EndOfStream` cleanly completes the future and flips `streamAlive` so the worker reconnects. `EndOfStream(SUCCESS)` after the max stream duration is treated as a normal event, not an error.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,528 @@
...
+            // which the outer worker treats as a reconnect trigger.
+            if (response.hasSkipBlock()) {
+                System.err.println("[push] BN sent SkipBlock: block "
+                        + response.skipBlock().blockNumber());
+            } else if (response.hasResendBlock()) {
+                System.err.println("[push] BN sent ResendBlock: block "
+                        + response.resendBlock().blockNumber());
+            } else if (response.hasNodeBehindPublisher()) {
+                System.err.println("[push] BN reports node behind publisher at block "
+                        + response.nodeBehindPublisher().blockNumber());
```
jsync-swirlds: Important, these are _not_ informational.
- If you get skip block and keep sending, you're likely to get disconnected because the block node is expecting a new block, not the middle of the current block, so you look like a bad publisher.
- If you get resend, you need to roll back to that block and resend it, or you'll get a node behind publisher because you keep going forward and the block node told you it is missing an earlier block.
- If you get node behind publisher, you need to roll back to the block offered, because the publisher cannot accept your blocks; they're too far in the future. IF you keep sending you'll get disconnected and may even be blocked for a while for misbehaving.

Now, most of these _should not_ arise, but having the client basically ignore them is a high risk option.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,528 @@
...
+            } else if (response.hasEndStream()) {
+                final PublishStreamResponse.EndOfStream end = response.endStream();
+                System.err.println("[push] BN sent EndOfStream (code=" + end.status() + ", block=" + end.blockNumber()
+                        + "); worker will reconnect");
+                streamAlive.set(false);
+                completion.complete(null);
```
jsync-swirlds, line 565: The next block to send should probably be set to the `end.blockNumber()` here.  That's not the current block, it is the last known-valid-persisted block on the block node.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -0,0 +1,528 @@
...
+                        + response.skipBlock().blockNumber());
+            } else if (response.hasResendBlock()) {
+                System.err.println("[push] BN sent ResendBlock: block "
+                        + response.resendBlock().blockNumber());
+            } else if (response.hasNodeBehindPublisher()) {
+                System.err.println("[push] BN reports node behind publisher at block "
+                        + response.nodeBehindPublisher().blockNumber());
+            } else if (response.hasEndStream()) {
+                final PublishStreamResponse.EndOfStream end = response.endStream();
+                System.err.println("[push] BN sent EndOfStream (code=" + end.status() + ", block=" + end.blockNumber()
```
jsync-swirlds: Slight improvement in clarity:

```suggestion
                System.err.println("[push] BN sent EndOfStream (code=" + end.status() + ", last persisted block=" + end.blockNumber()
```

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Addressed in 9c3d1eca2.

Each response variant now drives a concrete watermark action rather than being logged and forgotten:

- SkipBlock: advance the ACK watermark to the skipped block so we do not resend it (BN already persisted it)
- ResendBlock: rewind the ACK watermark to target-1 and tear down the stream, so the worker reconnect replays inFlight from the target block
- NodeBehindPublisher: same rewind + reconnect pattern as ResendBlock, per your note about being disconnected or blocked if we keep pushing forward

Split into advanceAckWatermarkTo / rewindAckWatermarkTo so the intent is clear at each call site.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Done in 9c3d1eca2. On EndOfStream we now call advanceAckWatermarkTo(end.blockNumber()) before completing the stream future, so the worker reconnect skips anything <= end.blockNumber() during inFlight replay and resumes from end.blockNumber() + 1.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
jsync-swirlds: Both resend and node behind are _permitted_ to continue sending without reconnect, they just have to rewind and send from the requested block.

It's not hugely important which approach is used, continuing to use the existing connection is a bit more efficient, but that's about all.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -482,25 +508,43 @@ public void onNext(final PublishStreamResponse response) {
...
+            // The publish protocol responses beyond Acknowledgement drive the client's next-block
+            // watermark: continuing to push forward after a Skip/Resend/NodeBehindPublisher (or
+            // ignoring EndOfStream's persisted-block hint) risks a disconnect or a misbehaving-
+            // publisher block from the BN. Each variant is mapped to an ACK-watermark change and,
+            // where applicable, a stream tear-down; the worker's reconnect + inFlight replay then
+            // naturally resumes at the correct next block.
             if (response.hasSkipBlock()) {
-                System.err.println("[push] BN sent SkipBlock: block "
-                        + response.skipBlock().blockNumber());
+                // BN has already persisted this block; treat it as ACKed so we do not resend it.
```
jsync-swirlds: This isn't quite true. SKIP means someone else is sending the same block, not that it's persisted.
If the block was already persisted then you'd receive an EndOfStream(DUPLICATE).

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -482,25 +508,43 @@ public void onNext(final PublishStreamResponse response) {
...
+                // inFlight from that block; continuing forward here draws a NodeBehindPublisher.
+                final long target = response.resendBlock().blockNumber();
+                System.err.println("[push] BN sent ResendBlock: rewinding to block " + target + " and reconnecting");
+                rewindAckWatermarkTo(target - 1);
+                streamAlive.set(false);
+                completion.complete(null);
             } else if (response.hasNodeBehindPublisher()) {
-                System.err.println("[push] BN reports node behind publisher at block "
-                        + response.nodeBehindPublisher().blockNumber());
+                // Publisher can't accept blocks this far ahead. Rewind and reconnect, or the BN
```
jsync-swirlds: nit: We're the Publisher; this is the Block Node that cannot accept blocks this far ahead.

Rewind is required, reconnect is optional.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
```diff
@@ -482,25 +508,43 @@ public void onNext(final PublishStreamResponse response) {
...
             } else if (response.hasEndStream()) {
                 final PublishStreamResponse.EndOfStream end = response.endStream();
-                System.err.println("[push] BN sent EndOfStream (code=" + end.status() + ", last persisted block=" + end.blockNumber()
-                        + "); worker will reconnect");
+                System.err.println("[push] BN sent EndOfStream (code=" + end.status()
+                        + ", last persisted block=" + end.blockNumber() + "); resuming from block "
+                        + (end.blockNumber() + 1));
+                // Advance to the BN's declared last-persisted watermark so the reconnect resumes
+                // at end.blockNumber() + 1 rather than replaying anything already persisted.
+                advanceAckWatermarkTo(end.blockNumber());
```
jsync-swirlds: Note, end stream can go forward or backward, depending on why the stream was ended.
You probably need to check greater or less than and either rewind or advance based on that.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Thanks for clarifying, that matches how I understand it now. Kept the rewind-and-reconnect approach here since the worker's inFlight-replay path already handles the rewind mechanically; noted in the code that continuing on the same stream is equivalent and would be a bit more efficient.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Fixed in 8d006011. Comment and log line now say another publisher is currently sending the block; noted that EndOfStream(DUPLICATE) is the persisted-already signal. Watermark behavior unchanged (still advance so we do not resend).

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Fixed in 8d006011. Comment now names the Block Node as the party that cannot accept blocks this far ahead and calls out that rewind is required and reconnect is optional. Kept the tear-down-and-reconnect approach because the worker already has an inFlight replay path that handles the rewind.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Good catch, fixed in 8d006011. EndOfStream now compares end.blockNumber() against lastAckedBlock.get() and calls advanceAckWatermarkTo or rewindAckWatermarkTo accordingly, so the reconnect resumes at exactly end.blockNumber() + 1 whether the BN's hint is ahead of or behind our current watermark. Previous code only called advance, which was a silent no-op on backward hints.

jsync-swirlds (approved):

tools-and-tests/tools/src/main/java/org/hiero/block/tools/push/LiveBlockPushClient.java
rockysingh: Applied via GitHub's commit-suggestion flow in cb8bd671 (way back on 2026-07-08). The 'last persisted block=' phrasing is in the log line as suggested; thanks for the nudge.

Nana-EC (approved):

PR #3127, opened 2026-07-01
feat: Route partial first group to temporary archive when plugin starts mid-group with no prior state

### LINKED ISSUE #3123: Route partial first group to temporary archive when plugin starts mid-group with no prior state
### Story Form

I want to route incoming blocks to a temporary archive when a completely fresh start finds the first block at a non-aligned group boundary
So that those blocks are durably stored in S3 immediately rather than accumulating in memory

### Background

When #3045 replaced EMB-based routing with gap-triggered temp archives, it relied on `lastHandedOffBlock` to detect gaps. On a completely fresh start (no prior S3 state), `lastHandedOffBlock` is initialized to -1. The `lastHandedOffBlock == -1` branch in `routeVerifiedBlock` treats the first arriving block as contiguous and routes it to `routeToCloudArchive`.

If that first block is not at an aligned group boundary (e.g. block 1,289,123 with `groupSize` 100,000), `startNewUploadTask` creates a `BlockUploadTask` for the full aligned group [1,200,000, 1,300,000) and sets `nextBlockToQueue = 1,200,000`. Blocks before the first arriving block will never appear, so `drainPendingToQueue` is permanently stalled waiting for block 1,200,000. All subsequent blocks in the group accumulate in `currentGroupPending` without ever reaching the upload task.

This scenario occurs whenever the cloud storage archive plugin is deployed against a live network for the first time and the chain is already in progress.

### Changes Required

- first block beyond group 0 whose true predecessors never arrive flushes to a temp archive once `gapBuffer` fills, instead of leaving the regular pipeline stalled indefinitely.

*Note*: The range `[groupStart, firstArrivingBlock)` will have no temp archive coverage because those blocks were never received. `checkGroupCoverage` will not trigger consolidation for that group until a backfill delivers the missing blocks. This is expected behavior for a node that joins the network mid-chain.

ivannov:
### Fresh-start routing narrowed to only the unambiguous cases

Previously any first-ever arrival went straight to the regular pipeline regardless of block number. If that first block wasn't on an aligned group boundary, the resulting task waited on predecessors that would never arrive, stalling permanently.

`routeVerifiedBlock` now only takes the direct-to-regular path on a fresh start when the block is on an aligned group boundary or falls within group 0 (block 0 is the one universally-known anchor, so any other block there is ordinary out-of-order arrival, not evidence of a live mid-chain deployment). Anything else beyond group 0 falls through into the existing gap-buffer logic from #3045: it buffers in case predecessors are just reordered, and if the buffer fills without them showing up, it routes to a temp archive instead of stalling.

### Fix for stragglers from an already-passed group being silently discarded

`blockNumber < nextBlockToQueue` was treated uniformly as "already consumed or skipped," but that's only true within the current group. A straggler from an earlier group the regular pipeline never handled (e.g. temp-archived) matched this too and was silently dropped. The check now also requires `blockNumber >= currentGroupStart`, so out-of-group stragglers fall through to `routeToTempArchive` instead.

## Related Issue(s)

Closes #3123


block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -621,9 +627,9 @@ private void routeVerifiedBlock(long blockNumber, VerificationNotification notif
                 // making this block appear retrograde. Start a new regular task for its group.
                 LOGGER.log(TRACE, "Block {0} starting new regular task", blockNumber);
                 routeToCloudArchive(blockNumber, blockWithSource);
-            } else if (blockNumber < nextBlockToQueue) {
-                // Block falls before the current (or most recent) regular task's queue pointer:
-                // already consumed or deliberately skipped (e.g. resume point after recovery).
+            } else if (blockNumber < nextBlockToQueue && blockNumber >= currentGroupStart) {
+                // Already consumed by the current task's own group; a block below currentGroupStart
+                // belongs to an earlier group the regular pipeline never handled, so it falls through.
```
jsync-swirlds, line 635: I continue to believe that we need a place to store these blocks rather than discarding them; mostly because I never assume that everything goes as expected.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -589,11 +589,17 @@ private void routeToCloudArchive(long blockNumber, BlockWithSource blockWithSour
...
-    /// the gap closes before the buffer fills, buffered blocks drain into the regular pipeline.
-    /// If the buffer fills first, all buffered blocks are flushed to a [TempArchiveUploadTask].
-    /// Retrograde or duplicate blocks are silently discarded.
+    /// After recovery, contiguous blocks go to the regular pipeline via [currentGroupPending].
+    /// A gap triggers [gapBuffer] accumulation; if the gap closes before the buffer fills,
+    /// buffered blocks drain into the regular pipeline.  If the buffer fills first, all buffered
+    /// blocks are flushed to a [TempArchiveUploadTask].  Retrograde or duplicate blocks are
+    /// silently discarded.
+    ///
+    /// On a fresh start, a first arrival in group 0 or on any group boundary is always safe to
```
jsync-swirlds: I do not understand why we think group 0 is "safe".
Consider:
group is 100,000 blocks
First block to arrive is block 24,567
Backfill never finds any block lower than 25,000; we end up with 24,567 and 25,000+, but nothing lower.
Secondary concern: blocks 25,000-100,000 arrive in order from publishers, but the check on line 630 drops every block below 24,567; so the group never completes.


How is this different from any other block?

jsync-swirlds (commented):
A couple of questions about block retention.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: OK, maybe _special_ would be a better wording here for my intentions. I was going to introduce special handling for group 0 in another task.

But now as I look from a perspective, you are right. There shouldn't be anything special about group 0 (especially for the cases with bigger tars). So I changed the code (and some tests) to treat is as a normal group.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: You're right that this needed a closer look - digging in, I found the actual place where "everything not going as expected" happens: `BlockUploadTask`'s `InterruptedException` handler was calling _abort multipart upload_, deleting parts already confirmed durable, and that interrupt fires whenever `triggerMidRunRecovery()` cancels 
the task. I've removed that abort so the interrupt path now leaves the hanging upload for StartupRecoveryTask to resume, exactly like the existing IOException path already does.

With that closed, everything reaching this line is provably a true duplicate (nextBlockToQueue only advances when a block is handed to the queue, never skips a gap), so I kept the discard. Rerouting to routeToTempArchive instead would actually be worse, since its dedup only tracks temp segments and has no idea the regular pipeline already owns this block. 

To make the invariant observable rather than assumed, I added a `duplicate_blocks_discarded` counter incremented here and at the equivalent `routeToTempArchive` discard, and bumped both log lines to DEBUG. 

One related gap I'm not fixing here: `TempArchiveUploadTask` has the identical abort pattern but no resume path on restart: captured in follow up task https://github.com/hiero-ledger/hiero-block-node/issues/3138

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
```diff
@@ -176,8 +176,10 @@ public UploadResult call()
                         blocksInBuffer.put(blockNum, blockData.source());
                     }
                 } catch (InterruptedException e) {
+                    // Do not abort here: it would delete parts already confirmed durable via
+                    // PersistedNotification. Leave the upload hanging for StartupRecoveryTask to
+                    // resume, mirroring the IOException branch below.
```
jsync-swirlds, line 181: The other thing you could do is _complete_ the upload, if the blocks are closed, as temp files are not expected to have a particular end block.

jsync-swirlds (approved):
Will look for the follow-up item.

AlfredoG87 (approved):
LGTM

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
ivannov: Maybe you meant it for `TempArchiveUploadTask` not for `BlockUploadTask`?

It's a very good idea, as there we do abort upon interruption. I created this little task to handle that: https://github.com/hiero-ledger/hiero-block-node/issues/3147

PR #3128, opened 2026-07-01
ci(sync): push to branch and open PR instead of direct push to master

deshmukhpranali:
## Reviewer Notes
hiero-docs enforces a DCO-required status check on master. A direct push from the bot was rejected with GH013 because no Signed-off-by trailer was present and the branch protection cannot be bypassed.

This change:
- Commits with -s so the DCO check passes on the resulting PR.
- Pushes to a rolling sync/block-node-docs branch (not master directly).
- Opens a PR against master if one does not already exist, or lets the existing PR auto-update on subsequent syncs.
- Adds workflow_dispatch so the sync can be triggered manually.

### DIFF
--- a/.github/workflows/sync-docs-to-gitbook.yaml
+++ b/.github/workflows/sync-docs-to-gitbook.yaml
@@ -58,10 +58,11 @@ jobs:
           fi
           BRANCH="sync/block-node-docs"
           git checkout -b "$BRANCH"
-          # -s adds the DCO Signed-off-by trailer that hiero-docs requires (matches the bot author).
-          git commit -s -m "chore: sync block-node docs from hiero-block-node@${GITHUB_SHA::7}"
+          # --signoff adds the DCO Signed-off-by trailer that hiero-docs requires (matches the bot author).
+          git commit --signoff -m "chore: sync block-node docs from hiero-block-node@${GITHUB_SHA::7}"
           # Push to a dedicated branch (master is protected; a direct push is rejected).
-          git push --force origin "$BRANCH"
+          # --set-upstream binds local and remote, reducing opportunity for mistakes.
+          git push --force --set-upstream origin "$BRANCH"
           # Open the PR if one isn't already open for this branch; otherwise the existing PR auto-updates.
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-docs-to-gitbook.yaml
```diff
@@ -43,11 +44,30 @@ jobs:
...
+          if git diff --staged --quiet; then
+            echo "No block-node doc changes to sync."
+            exit 0
+          fi
+          BRANCH="sync/block-node-docs"
+          git checkout -b "$BRANCH"
+          # -s adds the DCO Signed-off-by trailer that hiero-docs requires (matches the bot author).
+          git commit -s -m "chore: sync block-node docs from hiero-block-node@${GITHUB_SHA::7}"
+          # Push to a dedicated branch (master is protected; a direct push is rejected).
+          git push --force origin "$BRANCH"
```
jsync-swirlds: Adding track binds local and remote, which reduces opportunity for mistakes.
```suggestion
          git push --force --track origin "$BRANCH"
```

.github/workflows/sync-docs-to-gitbook.yaml
```diff
@@ -43,11 +44,30 @@ jobs:
...
-          git diff --staged --quiet || git commit -m "chore: sync block-node docs from hiero-block-node"
-          git push
+          if git diff --staged --quiet; then
+            echo "No block-node doc changes to sync."
+            exit 0
+          fi
+          BRANCH="sync/block-node-docs"
+          git checkout -b "$BRANCH"
+          # -s adds the DCO Signed-off-by trailer that hiero-docs requires (matches the bot author).
+          git commit -s -m "chore: sync block-node docs from hiero-block-node@${GITHUB_SHA::7}"
```
jsync-swirlds: `-s` can be replaced with `--signoff` if we want greater clarity.
If we have a GPG key for the bot, we could also add `--gpg-sign`.

jsync-swirlds (dismissed):
Two minor possible improvements; but these can be done as a follow-up.

Nana-EC (dismissed):

jsync-swirlds (approved):

Nana-EC (approved):

Issue #2819, opened 2026-05-14
Make ApplicationStateFacility the source of truth for all block range searches


berryware:
### Story Form
As a BlockNode engineer I would like the ApplicationStateFacility to be the source of truth for all block range searches.

### Technical Notes
- [ ] - Backfill currently searches gaps from block providers and should switch to pulling from new logic in applicationStateFacility
- [ ] - serverStatus & serverStatusDetails should be extended with the concept of stored vs available. May want to consider another word than stored.
- [ ] - Publisher must, on startup, also use the ranged set of stored blocks coming from the application state facility in order to correctly determine which block to expect next.
- [ ] - Block updates should persist every {N} blocks for better performance
- [ ] - ApplicationStateFacility needs to handle all block range updates including pruned blocks and partial storage.

Nana-EC:
@ivannov  to add a ticket to make ASF source of truth for availableBlocks

AlfredoG87:
Should we do the ApplicationStateFacility a plugin first?

berryware:
> Should we do the ApplicationStateFacility a plugin first?

I think we should. BlockNodeApp will get to bulky if we add all the code there, and there will be more to move to the ASF plugin.

berryware:
All tasks completed

PR #3129, opened 2026-07-02
fix(backfill): stop re-fetching blocks evicted by the recent-tier retention policy

### LINKED ISSUE #2698: Bug: Backfilled Blocks Immediately Evicted by Recent Storage Retention Policy
# Bug: Backfilled Blocks Immediately Evicted by Recent Storage Retention Policy

## Summary

When running the **remote plugins profile** (recent-file storage + S3 archive, no historical storage), backfill detects historical gaps and fetches old blocks from peers. Those blocks land in recent storage, which immediately deletes them because they are the oldest entries in the retention window. S3 never archives them. The next gap scan re-detects the same gap, creating an infinite retry loop.

## Impact

Historical gaps that fall outside the recent retention window can **never be permanently stored** in this configuration. S3 is left with a permanent gap matching the missed blocks. Backfill wastes peer bandwidth retrying indefinitely.

### Steps to Reproduce

1. Start a block node with the remote plugins profile and default retention (96,000 blocks).
2. Let recent storage fill to capacity (receive ≥ 96,001 live blocks).
3. Create a historical gap: stop the node before the window was full, miss a range of blocks, then restart so the node continues from a later point, leaving a gap below `latest - 96,000`. _Alternatively_: configure `startBlock` in backfill to point at a gap that is older than the current retention window.
4. Enable backfill pointing at a peer that has the missing blocks.
5. **Observe:**
   - Backfill logs `Backfilling block [N]` for the gap block.
   - Retention immediately deletes block N (check recent-storage metrics or log at DEBUG).
   - S3 never receives an upload for block N.
   - The gap re-appears in the next backfill scan cycle (infinite loop).

## Root Cause

Recent storage (`BlockFileRecentPlugin.handleVerification`) runs the retention policy in the **same event handler call** that writes a newly received block:

```
writeBlockToLivePath(block)          // block is now the OLDEST entry
→ excess = availableBlocks.size() - 96_000
→ delete from availableBlocks.min()  // min() == the block we just wrote
```

Because a historical gap is, by definition, older than all currently stored blocks, the backfilled block becomes `availableBlocks.min()` the instant it is written. With a full 96,000-block window, `excess == 1` and the block is deleted immediately.

S3 archive checks `available.contains(startBlock, endBlock)` before queuing an upload (`S3ArchivePlugin.java:224`). The window between "block written" and "block deleted" is a single event-handler pass — effectively zero from S3's perspective.

## Affected Configuration

- **Profile:** remote plugins (recent-file storage + S3 archive)
- **Setting:** `blockRetentionThreshold > 0` (default 96,000)
- **Condition:** any gap whose block number is below `latestBlock - 96,000`

### LINKED ISSUE #2820: Update Backfill to use ApplicationStateFacility
### Story Form

As a BlockNode engineer I would like the ApplicationStateFacility to be the source of truth for all block range searches.

### Technical Notes

Backfill currently searches gaps from block providers and should switch to pulling from new logic in applicationStateFacility

AlfredoG87:
fix(backfill): stop re-fetching blocks evicted by the recent-tier retention policy

## Problem

On a recent-tier deployment with retention, backfill fetched a historical block, retention evicted it, and the next scan re-detected the now-missing block and re-fetched it — forever, wasting peer bandwidth and local churn (#2698). The root cause: backfill detected gaps from *available* blocks (what's retrievable right now) and consumed only a peer's fuzzy `first/last` range.

## Change

Backfill is now a pure consumer of Application State: it reads block-range state and never writes it. The stored set is owned by the storage/archive plugins (cloud-storage-archive) and reflects what was actually persisted remotely.

- **Gap detection uses stored ∪ available.** A block a storage plugin recorded as stored is not re-backfilled after the recent tier evicts it from `available`. Retention only removes from the recent plugin's own `available` set; the stored set is monotonic and controlled elsewhere. The plugin caches the stored+available snapshot delivered via `onContextUpdate()` (the same merge `BlockNodeApp` already computes when the underlying data changes) and unions it with a live `availableBlocks()` read on every scan, so it never re-queries the Application State facility's raw stored set per scan while still reflecting blocks persisted or evicted since the last context update.

- **Exact peer ranges via `serverStatusDetail`.** Replaces the coarse `serverStatus` first/last, so peers with internal gaps are handled precisely and a chunk is only requested from a peer that actually holds it. `getNewAvailableRanges` now returns the union of peers' real ranges instead of fabricating a single contiguous span.

- **Exponential backoff for reappearing gaps.** When a gap keeps reappearing without shrinking (e.g. recent-only deployments where nothing records the block as stored), it is retried with an exponentially growing delay (base = scan interval, capped at `maxBackoffMs`) instead of every scan. Backoff resets the moment a gap makes real progress.

Net effect: with a cloud archive the loop simply cannot happen (archived blocks stay in `stored`, so they're never re-detected). Without one, backfill still won't hammer — it backs off to the cap.

## Notable

- `BackfillPlugin` now implements `onContextUpdate()` to cache the merged stored+available block ranges reported by the Application State facility, instead of adding a new read accessor to `ApplicationStateFacility` — the initial approach was flagged in review as inconsistent with the already-merged `BlockNodeContext.storedBlocks()` and as redundant per-scan work.
- Fixed a bug in the `TestBlockNodeServer` fixture where `serverStatusDetail` reported only the last available range.
- The old "lying node marked unavailable" test now reflects the new behavior: with exact ranges a peer that has blocks on both sides of a gap serves both.
- `PluginTestBase` gained `updateStoredBlocks()`, which records a stored range and delivers it to the plugin under test via `onContextUpdate()`, matching how `BlockNodeApp` merges stored and available blocks in production.
- Fixed a pre-existing flaky test (`BlockNodeAppTest.testBlockRangesTriggerOnContextUpdate`) that assumed a fixed number of `onContextUpdate` calls; two rapid state changes can coalesce into a single scanner tick, so the test now waits on the delivered content rather than a call count.

## Testing

Tests in `BackfillPluginTest` cover: stored blocks are not re-backfilled, a reappearing gap is throttled by backoff (not re-fetched every scan), and exact peer ranges are consumed while the peer's internal gap is skipped. `BackfillFetcherTest` mocks moved to `serverStatusDetail`. All affected module suites (`spi-plugins`, `app`, `backfill`, `cloud-storage-archive`, `health`, `server-status`, `block-verification`) and `qualityCheck` pass.

## Related issues

Closes #2698, closes #2820, advances #2819.

### DIFF
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -500,6 +500,11 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     }
 
     @Override
+    public BlockRangeSet storedBlocks() {
+        return storedBlocks;
+    }
+
+    @Override
     public NetworkData knownPublishers() {
         return knownPublishers.get();
     }
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -667,7 +667,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
             addressBookHistory = null;
         }
 
-        List<BlockRange> storedBlockRange = toBlockRange(storedBlocks);
+        List<BlockRange> storedBlockRange = mergeRanges(storedBlocks, availableBlocks);
         List<BlockRange> availableBlockRange = toBlockRange(availableBlocks);
 
         if (tssData == null
@@ -688,11 +688,9 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
             builder.rangedAddressBookHistory(addressBookHistory);
         }
 
-        // The next two items must remain in order (available first, stored second).
         if (availableBlockRange.hashCode() != context.availableBlocks().hashCode()
                 || !availableBlockRange.equals(context.availableBlocks())) {
             builder.availableBlocks(availableBlockRange);
-            storedBlockRange = mergeRanges(storedBlocks, availableBlocks);
         }
         if (storedBlockRange.hashCode() != context.storedBlocks().hashCode()
                 || !storedBlockRange.equals(context.storedBlocks())) {
@@ -729,14 +727,17 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     /// Merge two sets of block ranges into a single list of block ranges.
     /// This is a very expensive method, O(n<sup>2</sup>), so it should be used
     /// carefully. Perhaps a future update to BlockRangeSet will add a more
-    /// efficient merge process.
+    /// efficient merge process. Called unconditionally on every scan (not only when
+    /// `availableBlocks` changes) — `storedBlocks` can change independently (e.g. a plugin calling
+    /// `addStoredBlockRange`), and skipping the merge in that case previously let `storedBlocks()`
+    /// regress to the unmerged, durable-only set once `availableBlocks` stabilized.
     ///
     /// @param storedBlocks a set of stored blocks to merge into the result.
     /// @param availableBlocks a set of available blocks to merge into the result.
     private List<BlockRange> mergeRanges(final BlockRangeSet storedBlocks, final BlockRangeSet availableBlocks) {
         ConcurrentLongRangeSet combined = new ConcurrentLongRangeSet();
-        combined.addAll(storedBlocks.streamRanges().toList());
-        combined.addAll(availableBlocks.streamRanges().toList());
+        combined.addAll(storedBlocks);
+        combined.addAll(availableBlocks);
         return combined.streamRanges()
                 .map(longRange -> new BlockRange(longRange.start(), longRange.end()))
                 .toList();
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -500,11 +500,6 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     }
 
     @Override
-    public BlockRangeSet storedBlocks() {
-        return storedBlocks;
-    }
-
-    @Override
     public NetworkData knownPublishers() {
         return knownPublishers.get();
     }
--- a/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
+++ b/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
@@ -1044,16 +1044,25 @@ class BlockNodeAppTest {
         blockNodeApp.startApplicationStateFacility();
         blockNodeApp.loadedPlugins.add(testPlugin);
         blockNodeApp.addStoredBlockRange(new LongRange(0, 999));
-        testPlugin.expectContextUpdates(1);
         blockNodeApp.addStoredBlockRange(new LongRange(1000, 1049));
-        testPlugin.expectContextUpdates(1);
 
-        // wait for the ApplicationStateFacility scanner to pick up the update
-        testPlugin.awaitContextUpdates(15);
+        // The scanner delivers context updates asynchronously; poll the delivered context until it reports the
+        // merged range. Polling avoids racing an in-flight scanner tick whose callback could carry a stale range.
+        final long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(15);
+        BlockRange storedBlocks = null;
+        while (System.nanoTime() < deadline) {
+            final BlockNodeContext context = testPlugin.getContext();
+            if (context != null && !context.storedBlocks().isEmpty()) {
+                storedBlocks = context.storedBlocks().getFirst();
+                if (storedBlocks.rangeStart() == 0L && storedBlocks.rangeEnd() == 1049L) {
+                    break;
+                }
+            }
+            //noinspection BusyWait
+            Thread.sleep(50);
+        }
 
-        BlockNodeContext context = testPlugin.getContext();
-        assertNotNull(context);
-        BlockRange storedBlocks = context.storedBlocks().getFirst();
+        assertNotNull(storedBlocks, "onContextUpdate did not deliver the merged stored range within the timeout");
         assertEquals(0L, storedBlocks.rangeStart());
         assertEquals(1049L, storedBlocks.rangeEnd());
 
--- a/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
+++ b/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
@@ -351,10 +351,11 @@ class BlockNodeAppTest {
         BlockNodeContext awaitContext(final long timeoutSeconds, final Predicate<BlockNodeContext> condition)
                 throws InterruptedException {
             final long deadlineNanos = System.nanoTime() + TimeUnit.SECONDS.toNanos(timeoutSeconds);
-            long remainingNanos;
             BlockNodeContext delivered;
-            while ((remainingNanos = deadlineNanos - System.nanoTime()) > 0
-                    && (delivered = contextUpdates.poll(remainingNanos, TimeUnit.NANOSECONDS)) != null) {
+            // poll() treats a zero/negative timeout as "don't wait" and returns null immediately, so no
+            // separate deadline check is needed before it — this also sidesteps any ambiguity between
+            // ">" and ">=" for a near-zero remaining duration.
+            while ((delivered = contextUpdates.poll(deadlineNanos - System.nanoTime(), TimeUnit.NANOSECONDS)) != null) {
                 if (condition.test(delivered)) {
                     return delivered;
                 }
--- a/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
+++ b/block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
@@ -28,9 +28,12 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HexFormat;
 import java.util.List;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
 import java.util.stream.Stream;
 import org.hiero.block.api.BlockNodeVersions;
 import org.hiero.block.api.BlockNodeVersions.PluginVersion;
@@ -318,8 +321,9 @@ class BlockNodeAppTest {
     private static class TestPlugin implements BlockNodePlugin {
         private final AtomicInteger contextUpdated = new AtomicInteger(0);
         private volatile CountDownLatch latch = new CountDownLatch(0);
+        private final BlockingQueue<BlockNodeContext> contextUpdates = new LinkedBlockingQueue<>();
 
-        private BlockNodeContext context = null;
+        private volatile BlockNodeContext context = null;
 
         /** Call before the action under test to set how many `onContextUpdate` calls are expected. */
         void expectContextUpdates(final int count) {
@@ -336,6 +340,28 @@ class BlockNodeAppTest {
                     "onContextUpdate was not called within " + timeoutSeconds + "s");
         }
 
+        /**
+         * Blocks until a delivered context satisfies the given condition, or the timeout elapses.
+         * Drains delivered updates from a queue rather than assuming a specific number of updates
+         * will occur — updates can coalesce when several state changes land before the scanner's
+         * next tick.
+         *
+         * @return the first satisfying context, or {@code null} if the timeout elapses first
+         */
+        BlockNodeContext awaitContext(final long timeoutSeconds, final Predicate<BlockNodeContext> condition)
+                throws InterruptedException {
+            final long deadlineNanos = System.nanoTime() + TimeUnit.SECONDS.toNanos(timeoutSeconds);
+            long remainingNanos;
+            BlockNodeContext delivered;
+            while ((remainingNanos = deadlineNanos - System.nanoTime()) > 0
+                    && (delivered = contextUpdates.poll(remainingNanos, TimeUnit.NANOSECONDS)) != null) {
+                if (condition.test(delivered)) {
+                    return delivered;
+                }
+            }
+            return null;
+        }
+
         int getContextUpdated() {
             return contextUpdated.get();
         }
@@ -348,6 +374,7 @@ class BlockNodeAppTest {
         public void onContextUpdate(final BlockNodeContext context) {
             this.context = context;
             contextUpdated.incrementAndGet();
+            contextUpdates.add(context);
             latch.countDown();
         }
     }
@@ -1046,23 +1073,16 @@ class BlockNodeAppTest {
         blockNodeApp.addStoredBlockRange(new LongRange(0, 999));
         blockNodeApp.addStoredBlockRange(new LongRange(1000, 1049));
 
-        // The scanner delivers context updates asynchronously; poll the delivered context until it reports the
-        // merged range. Polling avoids racing an in-flight scanner tick whose callback could carry a stale range.
-        final long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(15);
-        BlockRange storedBlocks = null;
-        while (System.nanoTime() < deadline) {
-            final BlockNodeContext context = testPlugin.getContext();
-            if (context != null && !context.storedBlocks().isEmpty()) {
-                storedBlocks = context.storedBlocks().getFirst();
-                if (storedBlocks.rangeStart() == 0L && storedBlocks.rangeEnd() == 1049L) {
-                    break;
-                }
-            }
-            //noinspection BusyWait
-            Thread.sleep(50);
-        }
+        // The scanner delivers context updates asynchronously and may coalesce both writes into a
+        // single update, so wait on the merged content rather than a specific update count.
+        final BlockNodeContext context = testPlugin.awaitContext(
+                15,
+                ctx -> !ctx.storedBlocks().isEmpty()
+                        && ctx.storedBlocks().getFirst().rangeStart() == 0L
+                        && ctx.storedBlocks().getFirst().rangeEnd() == 1049L);
 
-        assertNotNull(storedBlocks, "onContextUpdate did not deliver the merged stored range within the timeout");
+        assertNotNull(context, "onContextUpdate did not deliver the merged stored range within the timeout");
+        final BlockRange storedBlocks = context.storedBlocks().getFirst();
         assertEquals(0L, storedBlocks.rangeStart());
         assertEquals(1049L, storedBlocks.rangeEnd());
 
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
@@ -362,10 +362,22 @@ public abstract class PluginTestBase<
      */
     public void updateStoredBlocks(LongRange blockRange) {
         appStoredBlocks.add(blockRange);
+        refreshContext();
+    }
+
+    /**
+     * Recomputes the merged stored+available context from the current {@link #appStoredBlocks} and the
+     * active {@link HistoricalBlockFacility}'s available blocks, and delivers it to the plugin via
+     * {@link BlockNodePlugin#onContextUpdate}. Mirrors how {@code BlockNodeApp}'s Application State
+     * facility periodically re-merges the two sets in production. Call this after directly mutating the
+     * historical block facility's available blocks (e.g. via block items sent through the test
+     * messaging facility) so the plugin observes the change without waiting for a real scanner tick.
+     */
+    protected void refreshContext() {
         final ConcurrentLongRangeSet merged = new ConcurrentLongRangeSet();
         merged.addAll(appStoredBlocks);
         merged.addAll(activeHistoricalBlockFacility.availableBlocks());
-        final List<BlockRange> mergedRanges = merged.toList().stream()
+        final List<BlockRange> mergedRanges = merged.streamRanges()
                 .map(r -> new BlockRange(r.start(), r.end()))
                 .toList();
         blockNodeContext = new BlockNodeContext.Builder(blockNodeContext)
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
@@ -26,6 +26,7 @@ import org.hiero.block.api.RangedAddressBookHistory;
 import org.hiero.block.api.TssData;
 import org.hiero.block.node.app.fixtures.TestMetricsExporter;
 import org.hiero.block.node.app.fixtures.async.TestThreadPoolManager;
+import org.hiero.block.node.base.ranges.ConcurrentLongRangeSet;
 import org.hiero.block.node.spi.ApplicationStateFacility;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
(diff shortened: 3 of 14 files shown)


berryware (dismissed):
Looks good.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -499,6 +499,11 @@ public void addStoredBlockRange(LongRange blockRange) {
         storedBlocks.add(blockRange);
     }
 
+    @Override
+    public BlockRangeSet storedBlocks() {
+        return storedBlocks;
+    }
+
```
jsync-swirlds: The problem with a trivial accessor here is that it will not be consistent with the latest Context; because the stored blocks are merged with available blocks on context update, while this is just the blocks explicitly labeled as `stored`.

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
```diff
@@ -1044,16 +1044,25 @@ void testBlockRangesTriggerOnContextUpdate() throws Exception {
...
+            final BlockNodeContext context = testPlugin.getContext();
+            if (context != null && !context.storedBlocks().isEmpty()) {
+                storedBlocks = context.storedBlocks().getFirst();
+                if (storedBlocks.rangeStart() == 0L && storedBlocks.rangeEnd() == 1049L) {
+                    break;
+                }
+            }
+            //noinspection BusyWait
+            Thread.sleep(50);
+        }
```
jsync-swirlds: The await used a condition wait, which is more reliable and _even that_ still fails sometimes.
We need to be improving test reliability, not adding more spin waits that sometimes work and sometimes fail.

Is there a mechanism we could be using that either uses a controllable execution service so there is no need to guess, or has an unambiguous condition wait?

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
```diff
@@ -130,14 +131,12 @@ public LongRange getNewAvailableRange(long latestStoredBlockNumber) {
...
-
-                // update the earliestPeerBlock to the max lastAvailableBlock
-                latestPeerBlock = Math.max(latestPeerBlock, lastAvailableBlock);
-                earliestPeerBlock = Math.min(earliestPeerBlock, firstAvailableBlock);
+                for (final LongRange range : resolveAvailableRanges(currentNodeClient)) {
+                    if (range.end() > latestStoredBlockNumber) {
+                        final long start = Math.max(range.start(), latestStoredBlockNumber + 1);
+                        peerRanges.add(new LongRange(start, range.end()));
+                    }
+                }
```
jsync-swirlds, line 144: This seems off.
Why are we modifying the start of ranges based on the highest stored block; are we deciding that backfill will never fill in gaps below that point?

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -296,11 +314,18 @@ private void shutdownExecutor(ExecutorService executor, String name) {
...
+        //    evicted from a volatile tier (e.g. by the recent-tier retention policy) but were already stored.
+        ConcurrentLongRangeSet knownBlocks = new ConcurrentLongRangeSet();
+        knownBlocks.addAll(
+                context.applicationStateFacility().storedBlocks().streamRanges().toList());
+        knownBlocks.addAll(context.historicalBlockProvider()
                 .availableBlocks()
                 .streamRanges()
-                .toList();
+                .toList());
+        List<LongRange> blockRanges = knownBlocks.streamRanges().toList();
```
jsync-swirlds: The ASF does this work when updating the context.
It just does so only when the data is meaningfully updated, rather than every time the data is queried.

Not sure which approach has better performance, but we know that performance is already a mild issue.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ApplicationStateFacility.java
```diff
@@ -47,6 +48,20 @@ default boolean updateAddressBookHistory(RangedAddressBookHistory history) {
...
+    /// are not re-fetched.
+    ///
+    /// The default implementation returns an empty set. Implementations that track stored blocks
+    /// (i.e. `BlockNodeApp`) override this method.
+    ///
+    /// @return the set of stored blocks; never `null` (empty when none are stored)
+    default BlockRangeSet storedBlocks() {
+        return BlockRangeSet.EMPTY;
+    }
+
```
jsync-swirlds: I'm not a fan of default implementations.
It hides places we forgot to update and really should have.

jsync-swirlds (commented):
A few concerns worth noting; all are advisory.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
ata-nas: I think we should be listening for updates instead of calling a method `storedBlock`, like in other places, and keep a local copy of the latest update. Then, we keep a local copy of those which can be served.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
AlfredoG87: Removed the trivial storedBlocks() accessor since it wasn't consistent with the already-merged BlockNodeContext.storedBlocks() ✅

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ApplicationStateFacility.java
AlfredoG87: Removed the new default storedBlocks() method entirely instead of keeping a default implementation

✅

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Replaced the per-scan ASF query + re-merge with a cached stored+available snapshot updated via onContextUpdate(), union with a live availableBlocks() read each scan ✅

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
AlfredoG87: Clarified via javadoc that the clamped range only sizes the greedy scan window and doesn't gate which gaps below baseline get backfilled (that's handled separately by getAvailabilityForRange)

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
AlfredoG87: Replaced the busy-wait polling loop with a BlockingQueue-drained condition wait (awaitContext) that doesn't assume a fixed update count ✅

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -95,6 +95,13 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
     // Deduplication: highest block scheduled for live-tail (prevents overlapping submissions)
     private final AtomicLong liveTailHighWaterMark = new AtomicLong(-1);
 
+    // Cached stored+available snapshot delivered by the last onContextUpdate() call. The Application
+    // State facility computes this merge itself whenever the underlying data actually changes, so
+    // caching it here avoids re-querying its raw stored set on every autonomous scan. Each scan still
+    // unions this with a live availableBlocks() read (see detectAndScheduleGaps) to reflect blocks
+    // persisted or evicted since the last context update.
+    private volatile List<LongRange> knownBlockRanges = List.of();
```
jsync-swirlds: Instead of a List, perhaps a `ConcurrentLongRangeSet` would be better?
That would make merging the ranges (to combine overlapping and contiguous ranges) easier.

This works even better if we add a `List` accessor to `ConcurrentLongRangeSet` (possibly returning an unmodifiable view using `Collections.unmodifiableList`) so we don't need to constantly remap via a stream just to get a List.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -187,6 +194,16 @@ public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
...
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onContextUpdate(BlockNodeContext context) {
+        knownBlockRanges = context.storedBlocks().stream()
+                .map(range -> new LongRange(range.rangeStart(), range.rangeEnd()))
+                .toList();
```
jsync-swirlds: Why are we remapping what is already a List?
`knownBlockRanges = context.storedBlocks()` would be equivalent, and avoids a streamed remap that doesn't need to happen...

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
jsync-swirlds: Grrr...
We should have an `addAll` override on `ConcurrentLongRangeSet` that takes another `ConcurrentLongRangeSet` so we're not having to perform this conversion all the time.

jsync-swirlds (dismissed):
A few more potential improvements.
These can be done in a follow-up if desired.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: knownBlockRanges is now a ConcurrentLongRangeSet built directly from context.storedBlocks(), no more mapping into List<LongRange>  ✅

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Widened addAll(ConcurrentLongRangeSet) to addAll(BlockRangeSet) (a strict superset, so all existing callers still compile) — the merge sites in detectAndScheduleGaps/handleNewestBlockKnownToNetwork now call addAll/max() straight on the cache and on availableBlocks(), no conversions left ✅

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Made knownBlockRanges a ConcurrentLongRangeSet and added a toList() accessor on ConcurrentLongRangeSet that returns the internal immutable list directly (no stream remap needed) ✅

ivannov (dismissed):

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -296,11 +332,16 @@ private void shutdownExecutor(ExecutorService executor, String name) {
...
-                .toList();
+        // 1. Get the blocks this node already has: the cached stored+available snapshot from the last
+        //    onContextUpdate() (avoids re-querying the Application State facility's raw stored set on
+        //    every scan) unioned with the live available set, so blocks persisted or evicted since the
+        //    last context update are reflected immediately. Folding in the cached snapshot — not just
+        //    live available — prevents re-backfilling blocks that have been evicted from a volatile tier
+        //    (e.g. by the recent-tier retention policy) but were already stored.
+        ConcurrentLongRangeSet knownBlocks = new ConcurrentLongRangeSet();
+        knownBlocks.addAll(knownBlockRanges);
+        knownBlocks.addAll(context.historicalBlockProvider().availableBlocks());
```
ata-nas, line 349: Why are we adding the available blocks from historical provider? Stored blocks should already be a superset of that. This seems off.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
```diff
@@ -363,10 +363,9 @@ public void addStoredBlockRange(LongRange blockRange) {
     public void updateStoredBlocks(LongRange blockRange) {
         appStoredBlocks.add(blockRange);
         final ConcurrentLongRangeSet merged = new ConcurrentLongRangeSet();
-        merged.addAll(appStoredBlocks.streamRanges().toList());
-        merged.addAll(
-                activeHistoricalBlockFacility.availableBlocks().streamRanges().toList());
-        final List<BlockRange> mergedRanges = merged.streamRanges()
+        merged.addAll(appStoredBlocks);
+        merged.addAll(activeHistoricalBlockFacility.availableBlocks());
+        final List<BlockRange> mergedRanges = merged.toList().stream()
```
jsync-swirlds: This is running two unnecessary conversions.
Use `streamRanges` instead of converting to a list, then streaming, then remapping back to a range.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -169,6 +195,16 @@ public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
         context.blockMessaging().registerBlockNotificationHandler(this, false, "BackfillPlugin");
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onContextUpdate(BlockNodeContext context) {
+        final ConcurrentLongRangeSet updated = new ConcurrentLongRangeSet();
+        context.storedBlocks().forEach(range -> updated.add(range.rangeStart(), range.rangeEnd()));
```
jsync-swirlds: This is inefficient (stream actions are always more costly than the equivalents) and somewhat confusing (it looks like you're updating context, rather than adding context into "updated".

Why not use addAll?

```suggestion
        updated.addAll(context.storedBlocks());
```

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
jsync-swirlds: This is merging in updates to historical that happened after the latest update to stored (which is much less frequent, in most cases).

block-node/base/src/main/java/org/hiero/block/node/base/ranges/ConcurrentLongRangeSet.java
```diff
@@ -219,13 +219,10 @@ public final void addAll(Collection<LongRange>... ranges) {
...
-     * @param ranges the ranges to add
+     * @param ranges the set of ranges to add
      */
-    public void addAll(ConcurrentLongRangeSet ranges) {
-        List<LongRange> otherRanges = ranges.ranges.get();
-        for (LongRange range : otherRanges) {
-            add(range);
-        }
+    public void addAll(BlockRangeSet ranges) {
+        ranges.streamRanges().forEach(this::add);
```
jsync-swirlds: Please avoid streams and lambdas in hot loops like this method can be.
The for loop avoids at least two vcall overheads and two object allocations compared to this construction.

block-node/base/src/main/java/org/hiero/block/node/base/ranges/ConcurrentLongRangeSet.java
```diff
@@ -554,6 +551,15 @@ public Stream<LongRange> streamRanges() {
         return ranges.get().stream();
     }
 
+    /**
+     * Returns the ranges in the set as an unmodifiable list, without the overhead of streaming.
+     *
+     * @return an unmodifiable, ascending, non-overlapping list of the ranges in the set
+     */
+    public List<LongRange> toList() {
+        return ranges.get();
```
jsync-swirlds: We need to protect the list we return to avoid unintended modifications.


```suggestion
        return Collections.unmodifiableList(ranges.get());
```

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: You're right — removed the live availableBlocks() merge entirely. not really needed, great catch! ✅

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
AlfredoG87: Fixed ✅

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Replaced the forEach lambda with a plain for loop ✅

block-node/base/src/main/java/org/hiero/block/node/base/ranges/ConcurrentLongRangeSet.java
AlfredoG87: Restored addAll(ConcurrentLongRangeSet) as a plain for loop over the internal list (zero-allocation fast path), and kept addAll(BlockRangeSet) as a separate overload for the generic case, also implemented as a for loop instead of streamRanges().forEach(this::add). Callers passing a ConcurrentLongRangeSet-typed reference get the fast overload automatically 

✅

block-node/base/src/main/java/org/hiero/block/node/base/ranges/ConcurrentLongRangeSet.java
AlfredoG87: Applied as suggested ✅

jsync-swirlds: Might be this did not get pushed?

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
jsync-swirlds: Seems there was an issue when removing this?

AlfredoG87: correct, I had to roll it back since it was failing e2e tests like the cached value is not reliable in practice, needs more investigation and out of scope for this PR, but will create a follow-up issue.

ata-nas: Yeah, I feel like when you get stored blocks from ASF, we have already merged all available at the point of the call, inside ASF. I agree with @AlfredoG87 We have to investigate as to why we see failures.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -321,13 +366,53 @@ private void detectAndScheduleGaps() {
...
+                final long start = gap.range().start();
+                detectedHistoricalStarts.add(start);
+                if (isHistoricalGapInBackoff(start, gap.range().size(), now)) {
+                    final String backingOffMsg = "Historical gap [{0}] is backing off, skipping this scan";
+                    LOGGER.log(DEBUG, backingOffMsg, gap.range());
+                    continue;
+                }
+            }
+            if (scheduleGap(gap)) {
+                metricsHolder.backfillGapsDetected.increment();
```
Nana-EC: Q: How will this impact metrics inferences?
Before this was incremented on every detection. Now, it's only incremented if that detection is also submitted for backfill.
Either the metric name and description need to be refined or we restore it so the intent is unchanged.
Let's validate to ensure the inferences from metrics don't change, especially if we're alerting on it.

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
```diff
@@ -336,6 +340,28 @@ void awaitContextUpdates(final long timeoutSeconds) throws InterruptedException
...
+         * next tick.
+         *
+         * @return the first satisfying context, or {@code null} if the timeout elapses first
+         */
+        BlockNodeContext awaitContext(final long timeoutSeconds, final Predicate<BlockNodeContext> condition)
+                throws InterruptedException {
+            final long deadlineNanos = System.nanoTime() + TimeUnit.SECONDS.toNanos(timeoutSeconds);
+            long remainingNanos;
+            BlockNodeContext delivered;
+            while ((remainingNanos = deadlineNanos - System.nanoTime()) > 0
```
Nana-EC: Should this be >= or should we drop the outer poll?
Considering the edge case of a really small remainingNanos e.g. ~1ns
I'm not sure on the impact of testing but watching out for flakiness here

Nana-EC (dismissed):
No blocking comments.
Let's get this in to unblock RFH and take updates in sprint starting today

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Good Catch!! Restored backfill_gaps_detected to its original unconditional semantics (matches existing docs/dashboard, no alerting change) and added a new backfill_gaps_submitted counter for the backoff-throttled case, which the backoff test now asserts on instead. Updated both metrics docs. ✅

block-node/app/src/test/java/org/hiero/block/node/app/BlockNodeAppTest.java
AlfredoG87: Yes you are correct! Dropped the outer check, poll() already treats timeout ≤ 0 as non-blocking and returns immediately, so we don't need to reason about > vs >= at all. ✅

block-node/base/src/main/java/org/hiero/block/node/base/ranges/ConcurrentLongRangeSet.java
AlfredoG87: Yes, but it should be there now 👍

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -667,7 +667,7 @@ private boolean updateBlockNodeContext(
             addressBookHistory = null;
         }
 
-        List<BlockRange> storedBlockRange = toBlockRange(storedBlocks);
+        List<BlockRange> storedBlockRange = mergeRanges(storedBlocks, availableBlocks);
```
jsync-swirlds, line 670: We shouldn't merge ranges unless the available ranges changed.
Otherwise we spend a very expensive merge when we do not need to, and also have a higher chance of updating stored when we should not need to.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -91,6 +98,27 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
     // Deduplication: highest block scheduled for live-tail (prevents overlapping submissions)
     private final AtomicLong liveTailHighWaterMark = new AtomicLong(-1);
 
+    // Cached stored+available snapshot delivered by the last onContextUpdate() call. The Application
+    // State facility already folds availableBlocks() into storedBlocks whenever either changes, so this
+    // alone is the source of truth for "blocks this node already has" — no separate live availableBlocks()
+    // read is needed here. A fresh set replaces this reference on each update rather than being mutated
+    // in place, so concurrent readers always see a consistent snapshot.
+    private volatile ConcurrentLongRangeSet knownBlockRanges = new ConcurrentLongRangeSet();
```
jsync-swirlds, line 106: Are we expecting to set this reference to a new value often?
Volatile has no effect on the content of the object here, it only applies to the reference itself.

Basically it doesn't protect the actual content in any way (which is fine if that's intended).

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -296,11 +336,18 @@ private void shutdownExecutor(ExecutorService executor, String name) {
...
+        //    onContextUpdate() unioned with a live availableBlocks() read. The Application State
+        //    facility's own periodic merge normally makes the cache alone sufficient, but its scan is
+        //    on a separate schedule and can (rarely) observe a transient/incomplete snapshot of the
+        //    underlying facility right as it's queried; re-checking the live set here is a cheap
+        //    guard against acting on such a stale snapshot. Folding in stored — not just available —
+        //    prevents re-backfilling blocks that have been evicted from a volatile tier (e.g. by the
+        //    recent-tier retention policy) but were already stored.
+        ConcurrentLongRangeSet knownBlocks = new ConcurrentLongRangeSet();
+        knownBlocks.addAll(knownBlockRanges);
+        knownBlocks.addAll(context.historicalBlockProvider().availableBlocks());
```
jsync-swirlds, line 349: I suspect this is needed because historical blocks facility updates itself much more often than context is updated.

jsync-swirlds (approved):
Looks good, one question we can address later, and a couple items to note.

You're welcome to resolve all three once read.

Nana-EC (approved):
LG

Issue #2089, opened 2026-01-29
Generic Ring Buffer Event Setter


ata-nas:
### Story Form

As a Block Node Engineer
I want to be able to easily create new notification types that will be handled by messaging
So that I can easily, at runtime, be able to introduce any type of notification, without having to make changes to the base logic and break things

### Technical Notes

- We currently have pre-defined notification types that are not generic.
- For every type of notification, we need to change the base code to support it (add setters)
- We should not have to do that, but allow any arbitrary plugin to be able to push any type of notification to messaging without having to make any changes to the base code.
- Moreover, we need a way to be able to handle any arbitrary notification from any plugin.

Issue #1744, opened 2025-10-09
Explore a POC for Async Prioritized Notifications Messaging


Nana-EC:
### Story Form

As a block node engineer
I want to have an optimize notification messaging queue that isn't overly packed and full
So that notification can travel quickly through the system and not get held up

### Technical Notes

Thea team discussed multiple options, one such solution was to implement prioritized notifications in the queue based on plugins.

Note
- Notifications must be fired in order thy were submitted.
- No 2 threads can fire notifications simultaneously.
- Consider memory implications and ensure we don't starve threads.
- Detect system load - is it per plugin or central to messaging itself

Issue #1431, opened 2025-07-25
Enable third party plugins to dynamically create and process notifications


jsync-swirlds:
### Goal

Third party plugins should be able to respond to existing notifications, create notifications, define new notifications, and coordinate across plugins with new notifications.

### Intention

We intend to modify the notification mechanism to allow for generic notifications that are not pre-defined in the block node core. We intend to enable plugins to register to receive individual notifications by notification identifier.  We intend to enable plugins to register new notification types at startup, send and receive notifications, and list available notification types.

### Considerations

Dynamic notifications is a significant change. Currently notifications are statically encoded in a central interface and any change to notifications requires a code change in the Block Node application module.
* Make sure that this change does not impact general performance for the tier 1 configurations (listed below).
* Ensure that this change is easily extensible without being overly broad.
* Ensure that a plugin developer can create several plugins that interact via notifications with no change to the block node software.

### Technical Details

Tier 1 configurations are, currently as follows:
* Full History Node
    * (facility) Messaging
    * (facility) Block Provider
    * Stream Publisher
    * Stream Subscriber
    * Verification
    * Files.Recent
    * Files.Historic
    * Backfill
    * Server Status
    * (future) State change processing
    * (future) Live State Management
    * (future) State Snapshots
    * (future) State Snapshot API
* Archive Node
    * (facility) Messaging
    * (facility) Block Provider
    * Stream Publisher
    * Verification
    * S3 Archive
    * (potentially) Files.Recent
    * Server Status
    * (restricted access) Stream Subscriber

### Tasks and Features

- Generic Notification Design Document
    - Design generic notification record(s)
    - Design generic notification interface(s)
    - Design notification registration, discovery, emission, and consumption
- Rework existing notifications
- Create a testing plugin, loaded only during testing, that registers
  notifications, sends and receives notifications, and validates the process.
  This plugin must _not_ require changes or support from the main application.
- Build end to end testing and solo-based testing to validate the design.

Nana-EC:
Having this allows new plugins to be more easily defined and testing e.g. S3 plugin and State plugins would be much easier with this. Not needed but if we could get it in before it would be great.
Depends on net number of notifications to be added in Beta-1.

PR #3131, opened 2026-07-02
feat(messaging): generic block notifications via BlockNotification marker interface

AlfredoG87:
## Summary

Replaces the five type-specific notification methods on `BlockMessagingFacility` and `BlockNotificationHandler` with a single generic pair (`sendBlockNotification`/`handleNotification`), so any plugin can define, emit, and consume its own notification type without touching `spi-plugins` or `facility-messaging`.

## Changes

- Design doc: `docs/design/architecture/generic-block-notifications.md`
- New `BlockNotification` marker interface; the five existing notification records now implement it
- `BlockNotificationRingEvent` collapsed from five typed fields/getters/setters to one
- `BlockMessagingFacility.sendBlockNotification(BlockNotification)` replaces the five `sendX` methods
- `BlockNotificationHandler.handleNotification(BlockNotification)` replaces the five `handleX` methods; consumers dispatch with a pattern-matching `switch`
- `BlockMessagingFacilityImpl` forwards notifications directly (no more type-check dispatch chain); the five per-type metric counters collapsed into one labeled counter (`notification_type` dynamic label)
- Migrated every plugin implementing `BlockNotificationHandler`: backfill, block-verification, blocks-file-historic, blocks-file-recent, cloud-storage-archive, cloud-storage-expanded, s3-archive, stream-publisher, verification, plus app test fixtures
- Added `CustomNotificationTest`, which defines a notification type inline (standing in for a third-party plugin's own type) and proves it flows through the real `BlockMessagingFacilityImpl` with the same delivery guarantees (in-order, exactly-once, dedicated handler thread) as the built-in types

## Test plan

- [x] `./gradlew compileJava compileTestJava compileTestFixturesJava` — full repo compiles
- [x] `./gradlew test` — full repo test suite passes
- [x] `CustomNotificationTest` re-run multiple times to rule out flakiness

## Related issues

Addresses #1431. Related, not addressed here: #1744, #2089.

### DIFF
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
@@ -186,8 +186,8 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
         }
     }
 
-    /// {@inheritDoc}
-    /// ---
+    /// Handles a backfilled block notification, dispatched from [#handleNotification].
+    ///
     /// This is where we receive blocks from backfill.
     /// We will always receive a complete block, one per notification.
     /// We can safely wrap the block as [BlockItems] which is both the start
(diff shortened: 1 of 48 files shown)


jsync-swirlds (changes requested):
This is large enough, and has a sufficiently widespread impact, that we should delay merging until next sprint; to give ample time for thorough review.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -228,13 +228,22 @@ public void handleBackfilled(final BackfilledBlockNotification notification) {
...
     }
 
+    /// {@inheritDoc}
+    @Override
+    public void handleNotification(final BlockNotification notification) {
+        switch (notification) {
+            case BackfilledBlockNotification b -> handleBackfilled(b);
+            case PersistedNotification p -> handlePersisted(p);
+            default -> {}
+        }
```
ata-nas, line 244: Note that in all places, this will produce an NPE if notification is null.
I would suggest we add a `null` case as well. This acts as a null check before we call the right method and we are sure we have no null value downstream.

Issue #3125, opened 2026-06-30
E2E test for the WRB-distribution flow (CLI + 3 BNs + 2 MNs + CN cutover)


rockysingh:
Part of epic #2954 — _Distribute historical & live WRBs and TSS to Tier-1 Block Nodes_. The epic currently has no ticket for an end-to-end test that exercises the full CLI + BN-distribution + MN-ingestion flow.

### Story Form

As an Engineer,
I want a repeatable CI job that truly exercises the WRB CLI and the full network distribution flow,
So that we can validate the historical-backfill push, the live-push, the BN-to-BN backfill chain, the MN switchover to BN-as-source, and the TSS cutover, end-to-end, against a real (Solo) deployment.

### Technical Notes

Repeatable CI job using Solo. Acceptance is the ordered execution and validation of each step below; intermediate state may be checked via the BN `serverStatus` / `serverStatusDetail` RPCs, the MN REST API, and the Solo provisioner outputs.

**Setup**

1. Use Solo to set up a CN v0.75 network with 3 v0.75 CNs and 1 v0.157.1 MN (call it `MN1`). Note: this setup has no BN, to avoid confusion or Solo issues at this stage.
2. Install `wrb-cli` and process records from the bucket/store. Ensure the created WRBs file structure has enough blocks to expose a nested directory structure. Ensure the live wrap continues in the background.

**BN deployments**

3. Install a Tier-0 BN (call it `BN1`) with the solo-provisioner and ensure `EMB=0`.
4. Install a 2nd Tier-1 BN (call it `BN2`) with the solo-provisioner and ensure `EMB=0` to highlight that the BN wants to start from genesis.
5. Install a 3rd Tier-1 BN (call it `BN3`) with the solo-provisioner and ensure `EMB` is set to a high enough number to support future live streaming.

**MN reconfiguration / second MN**

6. Upgrade `MN1` to point to `BN2` and `BN3` as sources of blocks (bucket remains as records backup), so it begins to pull live blocks.
7. Install a 2nd MN v0.157.1 (call it `MN2`) pointing to `BN2` and `BN3` as sources of blocks (no bucket reference), so it pulls blocks from genesis.

**CN → BN3 push and CLI → BN1 push**

8. Upgrade/reconfigure the CN in Solo to push WRBs to `BN3`.
9. Run the `wrb-cli` logic to move blocks from the CLI to `BN1`, and validate that both historical and live WRBs are moved and continue to move over.

[...]

PR #3143, opened 2026-07-02
test: WRB distribution E2E — slices 1-3 (steps 1-5 of #3125)

rockysingh:
## Summary

Implements **slices 1–3** of the WRB distribution E2E test specification from #3125.

Both new test definitions are passing in CI:

| Slice | #3125 Steps | Topology | Latest CI |
|-------|-------------|----------|-----------|
| 1–2 | CN + MN, wrb-cli, live wrap loop | `wrb-distribution-step12` | [28554266028](https://github.com/hiero-ledger/hiero-block-node/actions/runs/28554266028) |
| 3 | Add BN1/BN2/BN3 with EMB overlays | `wrb-distribution-step345` | [28612951515](https://github.com/hiero-ledger/hiero-block-node/actions/runs/28612951515) |

---

## Slice 1–2: Live WRB generation

Deploys 3 CNs (v0.75) and a Mirror Node (v0.157.1) **without Block Nodes**, allowing Solo to provision its MinIO tenant for the record stream bucket.

The test then:

1. Waits for record files to be generated.
2. Installs `wrb-cli`, downloads records from MinIO, executes `blocks wrap --network other`, and verifies the wrapped block directory layout.
3. Starts a background worker that polls MinIO every 30 seconds and reruns `blocks wrap` idempotently.
4. Observes the system for three minutes and verifies that additional wrapped blocks continue to be produced.

---

## Slice 3: Incremental Block Node deployment

Starting from the Slice 1–2 topology, Block Nodes are added one at a time while the live wrap loop continues running.

Each Block Node is configured with its own `BLOCK_NODE_EARLIEST_MANAGED_BLOCK` value:

| Block Node | EMB |
|------------|-----|
| BN1 | 0 |
| BN2 | 0 |
| BN3 | 100,000,000 |

The test:

1. Executes the Slice 1–2 baseline.
2. Installs each Block Node using `solo block node add`.
3. Applies the corresponding topology overlay.
4. Waits for the pod to become Ready.
5. Starts background `kubectl port-forward` processes for metrics assertions.
6. Verifies each Block Node's `BLOCK_NODE_EARLIEST_MANAGED_BLOCK` matches its overlay.
7. Confirms the live wrap loop continues producing wrapped blocks throughout all Block Node installations.

---

## Framework changes

A few framework changes were required to support deploying a topology with **zero Block Nodes**:

- `network-topology-tool/generate-chart-values-config-overlays.sh`
  - Treat an empty `block_nodes: {}` as informational instead of an error.
- `scripts/solo-deploy-network.sh`
  - Same relaxation for zero Block Nodes.
  - Added support for per-topology static Block Node overlays from:
    `overrides/${TOPOLOGY}/bn-block-node-<i>-values.yaml`
- `scripts/wrb-sequential-comparison.sh`
  - Wrapped `main "$@"` with a `${BASH_SOURCE[0]} == "$0"` guard so the script can be safely sourced by the WRB distribution workflow.

---

### Topologies

- `wrb-distribution-step12.yaml`
- `wrb-distribution-step345.yaml`

### Test definitions

- `tests/wrb-distribution-step12.yaml`
- `tests/wrb-distribution-step345.yaml`

### Block Node overlays

- `overrides/wrb-distribution-step345/bn-block-node-{1,2,3}-values.yaml`

### Scripts

- `install-and-run-wrb-cli.sh`
- `start-live-wrap.sh`
- `stop-live-wrap.sh`
- `assert-live-wrap-produced-new-blocks.sh`
- `add-bn.sh`
- `assert-bn-emb.sh`

### Workflow

Added `wrb-distribution-step12` and `wrb-distribution-step345` to the topology and test-definition selection lists in `.github/workflows/solo-e2e-test.yml`.

---

## Remaining work

This PR implements **slices 1–3** of #3125.

Follow-up PRs will implement:

- Step 6 — Mirror Node reconfiguration
- Step 7 — Mirror Node 2 deployment
- Step 8 — CN → BN3 streaming
- Step 9 — `wrb-cli` → BN1 streaming
- Step 10 — Block Node backfill
- Step 11 — TSS cutover
- Step 12 — CN cutover validation

---

## Test plan

-  `wrb-distribution-step12` passes in CI ([28554266028](https://github.com/hiero-ledger/hiero-block-node/actions/runs/28554266028))
-  `wrb-distribution-step345` passes in CI ([28612951515](https://github.com/hiero-ledger/hiero-block-node/actions/runs/28612951515))
-  All 12 events and the `no-errors` assertion pass for `wrb-distribution-step345`


.github/workflows/solo-e2e-test.yml
```diff
@@ -103,6 +103,8 @@ on:
           - 3cn-2bn-wrb-rsa # 3 CN, 2 BN, 1 MN; WRB verified via RSA roster
           - wrb-differential-test # 1 CN, 2 BNs (live + historic), 2 MNs
           - wrb-differential-test-minimal # 1 CN, 2 BNs (live + historic), no MN
+          - wrb-distribution-step12 # WRB Distribution E2E (#3125) — slice 1: 3 CN v0.75 + MN1 v0.157.1, no BN
```
Nana-EC, line 106: nitL not a fan of the naming as it's not intuitive, also steps might change and I'm not sure long term we would want separate options.
As a starting solution I can accept separate, in the future would they be merged to something like `wrb-distribution` or are they truly separate?
Maybe go with `wrb-dsitribution-phase1` and `wrb-dsitribution-phase2` as a phase can include multiple steps

tools-and-tests/scripts/network-topology-tool/generate-chart-values-config-overlays.sh
```diff
@@ -596,15 +602,24 @@ function main {
 
   log_line ""
   if [[ "${generated_count}" -eq 0 ]]; then
-    log_line "WARNING: No overlay files were generated."
-    log_line "  Topology: %s" "${TOPOLOGY_FILE}"
-    log_line "  Block nodes found: %s" "$(yq '.block_nodes | keys | length' "${TOPOLOGY_FILE}" 2>/dev/null || echo "0")"
-    log_line "  Mirror nodes found: %s" "$(yq '.mirror_nodes | keys | length' "${TOPOLOGY_FILE}" 2>/dev/null || echo "0")"
-    log_line "  Consensus nodes found: %s" "$(yq '.consensus_nodes | keys | length' "${TOPOLOGY_FILE}" 2>/dev/null || echo "0")"
-    fail "ERROR: Expected at least 1 overlay file but generated 0. Check topology file structure." 1
+    # A zero-overlay outcome is legitimate whenever the topology has no
```
Nana-EC, line 605: Weren't the logs removed informative?
Sounds like in some cases the topology may have no details of value here, in that case shouldn't it be handled differently?

tools-and-tests/scripts/solo-e2e-test/overrides/wrb-distribution-step345/bn-block-node-1-values.yaml
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: Apache-2.0
+# BN1 for the wrb-distribution-step345 topology (#3125 step 3).
+# Tier-0 Block Node. EMB=0 → this BN manages the full history from genesis;
+# it will later receive historical + live WRBs from the wrb-cli push path.
+blockNode:
+  config:
+    BLOCK_NODE_EARLIEST_MANAGED_BLOCK: "0"
```
Nana-EC: Instead of 3 separate files i might have a `tools-and-tests/scripts/solo-e2e-test/overrides/wrb-distribution-step345/bn-emb-values.yaml` with a placeholder value that the script can replace as needed.

tools-and-tests/scripts/solo-e2e-test/tests/wrb-distribution-step345.yaml
```diff
@@ -0,0 +1,139 @@
...
+#   * assert-bn-emb.sh exec'ing into each block-node-server container
+# Adding explicit node-healthy assertions here would fail the runner's
+# topology-target validation (block-node-1/2/3 aren't in the deploy-time
+# topology).  The `no-errors` assertion runs against whatever BNs are up
+# by the end of the test.
+assertions:
+  - id: no-errors
+    type: no-errors
+    description: "No errors in Block Node logs"
+    target: all
```
AlfredoG87, line 148: This assertion is not running over all BNs, only against first block-node, maybe that is by design, but since the `target: all` might be confuzing.

AlfredoG87 (approved):
LGTM, just one nit to consider to the follow-up

AlfredoG87 (dismissed):

Nana-EC (commented):
1st pass, will come back.

.github/workflows/solo-e2e-test.yml
rockysingh: Fair point. The current names (step12 / step345 / step67) map 1:1 to the numbered steps in the issue text, so if a reviewer opens the issue alongside the PR they can jump straight to the section that describes what the test covers. Long term I don't think we merge them into one monolithic test (it'd be a 60min+ CI run with hard-to-isolate failures), so the plan is one test-def per phase of the epic.

Happy to rename step12 → phase1, step345 → phase2, step67 → phase3 in a follow-up PR if the phaseN naming reads better; I'd rather not churn this PR's branch and lose the approvals. WDYT?

tools-and-tests/scripts/network-topology-tool/generate-chart-values-config-overlays.sh
rockysingh: Good catch. I tightened the check in a8b15c1c so it now keeps the original ERROR + generator-output context for any topology that declares block_nodes; the informational log_line only fires when the topology explicitly declares `block_nodes: {}` (the wrb-distribution slices, which install BNs post-deploy via solo block node add). So we still hard-fail on the case that would actually indicate a generator bug.

tools-and-tests/scripts/solo-e2e-test/overrides/wrb-distribution-step345/bn-block-node-1-values.yaml
rockysingh: Done in 0337bd8c. Replaced the three bn-block-node-{1,2,3}-values.yaml files with a single `overrides/wrb-distribution-step345/bn-emb-values.yaml` that carries a small map:

```yaml
embs:
  "1": "0"
  "2": "0"
  "3": "100000000"
```

`add-bn.sh` reads `embs.<index>` via yq and renders a minimal chart-values overlay to `/tmp` before passing it to `solo block node add`. One source of truth, no per-BN files.

tools-and-tests/scripts/solo-e2e-test/tests/wrb-distribution-step345.yaml
rockysingh: You're right, that was misleading. Root cause: our topology declares `block_nodes: {}` at deploy so Solo installs MinIO, and the runner's `get_all_block_nodes` fell back to just `block-node-1` when the topology had no BNs, so `target: all` was really `target: block-node-1`.

Fixed in a8b15c1c by adding a top-level `dynamic_targets:` list on the test-def:

```yaml
dynamic_targets:
  - block-node-1
  - block-node-2
  - block-node-3
```

The runner's target validator and `get_all_block_nodes` now both merge this list with the topology's own BNs, so `target: all` scrapes metrics from every BN that will actually be up by end of test.

Nana-EC (approved):

AlfredoG87 (approved):
LGTM

PR #3159, opened 2026-07-06
docs(block-node): add Block Node Integration FAQ for CN/BN/MN developers

### LINKED ISSUE #3104: docs: create Block Node integration FAQ for CN/BN/MN developers
## User story
As a CN/BN/MN developer, I want answers to common integration and protocol
questions in one place, so I can implement correct client behaviour without
reading proto files and source code.

## Scope
Create docs/block-node/integration-faq.md covering:

### Publish stream protocol (CN → BN)
- How does the CN initiate a publish stream to the BN?
- What does the BN send back during streaming (BlockAcknowledgement, SkipBlock,
  ResendBlock, BehindPublisher)?
- What does DUPLICATE_BLOCK mean and what should the CN do?
- What does TOO_FAR_BEHIND mean and what should the CN do?
- What does PERSISTENCE_FAILED mean and is the block lost?
- What does BAD_BLOCK_PROOF mean?
- When does the BN send TIMEOUT to the CN?
- Does the BN reconnect to the CN if the stream drops?

### Subscribe stream protocol (MN → BN)
- How does a Mirror Node subscribe to the block stream?
- What is the difference between INVALID_START_BLOCK_NUMBER and NOT_AVAILABLE?
- What should a subscriber do on NOT_AVAILABLE?
- How does the BN handle slow subscribers?

### Block access API
- What is the difference between NOT_FOUND and NOT_AVAILABLE in getBlock?
- Is stateSnapshot implemented?

### CN → BN wiring
- What is block-nodes.json and where does it live on the CN?
- What values go in block-nodes.json (address, ports, priority)?
- What is writerMode and when do I need FILE_AND_GRPC vs GRPC?
- Does block-nodes.json require a CN restart to take effect?
- How do I use solo block node add-external instead of editing block-nodes.json?

### WRB cutover and migration
- What changes for Mirror Nodes when the network transitions from record streams
  to block streams?
- What preparation steps does a BN need before WRB cutover?

### Versioning and compatibility
- Which BN version supports which CN version?
- Where is the compatibility matrix?

## Notes
- Seed answers from protocol reference docs, proto files, and code audit findings
- Gather additional questions from block node team via Slack before writing
- stateSnapshot should be noted as not yet implemented

## Acceptance criteria
- [ ] docs/block-node/integration-faq.md created
- [ ] Protocol-level answers cite proto file definitions

deshmukhpranali:
## Reviewer Notes

Creates `docs/block-node/integration-faq.md` - a comprehensive FAQ for Consensus Node, Block Node, and Mirror Node developers covering integration and protocol questions. All answers verified against proto files, design docs, and source code across hiero-block-node, hiero-mirror-node, and hiero-consensus-node.

## Related Issue(s)
Closes #3104

### DIFF
--- a/docs/block-node/faq/integration-faq.md
+++ b/docs/block-node/faq/integration-faq.md
@@ -26,6 +26,7 @@ responds:
 - If the block number is the next expected block → streaming continues normally.
 - If the block number ≤ last verified block → BN responds with `DUPLICATE_BLOCK`.
 - If the block number > next expected → BN responds with `BehindPublisher`.
+- If the block number matches a block currently being streamed by another publisher → BN responds with `SKIP_BLOCK`.
 
 The CN is always the client (initiator). The BN never dials a CN.
 
@@ -37,28 +38,18 @@ The BN can send five response types during an active stream:
 
 |        Response        |                    When sent                    |                What the CN should do                |
 |------------------------|-------------------------------------------------|-----------------------------------------------------|
-| `BlockAcknowledgement` | After block proof received and verified         | Continue streaming the next block                   |
-| `SkipBlock`            | Another publisher already delivered this block  | Skip to the block indicated and continue            |
+| `BlockAcknowledgement` | After block proof received and verified            | Mark block secured and continue streaming the next block |
+| `SkipBlock`            | Another publisher is currently sending this block  | Skip the current block and continue with the next        |
 | `ResendBlock`          | BN needs the CN to resend from a specific block | Resend from the block number specified              |
 | `BehindPublisher`      | BN is behind — CN is too far ahead              | Start a new stream from the block number specified  |
 | `EndOfStream`          | Terminal — stream is closed                     | See the status code and take the appropriate action |
 
 ### What does `DUPLICATE_BLOCK` mean and what should the CN do?
 
-`DUPLICATE_BLOCK` (code 5 in `EndOfStream`) means the block header the CN sent
+`DUPLICATE_BLOCK` means the block header the CN sent
 corresponds to a block that is already stored and verified by the BN. The CN closes the
 stream and resumes streaming — to this BN or a different available Block Node — beginning
 with the block after the last persisted and verified block.
-
-### What does `TOO_FAR_BEHIND` mean and what should the CN do?
-
-`TOO_FAR_BEHIND` (code 4 in `EndOfStream`) means the BN is too far behind the CN to
-catch up by streaming alone. The CN should:
-
-1. Switch streaming to a different Block Node.
-2. Include `earliest_block` and `latest_block` in the `EndStream` message so the
-   falling-behind BN can attempt to backfill from a peer.
-3. Resume streaming to this BN later once it has caught up via backfill.
 
 ### What does `PERSISTENCE_FAILED` mean and is the block lost?
 
@@ -80,7 +71,7 @@ available Block Node — from before the failed block.
 `TIMEOUT` (code 4 in `EndOfStream`) is sent when the delay between stream items
 exceeds the BN's configured timeout — the CN did not deliver the next item within the
 allowed window. The CN closes the stream and resumes streaming — to this BN or a
-different available Block Node — from before the timed-out block.
+different available Block Node — from the timed-out block.
 
 ### Does the Block Node reconnect to the Consensus Node if the stream drops?
 
@@ -114,18 +105,23 @@ closes.
 
 |               Code               |                                                           Meaning                                                           |                              What to do                               |
 |----------------------------------|-----------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------|
-| `INVALID_START_BLOCK_NUMBER` (4) | `start_block_number` is structurally invalid — negative, below `first_available_block`, or beyond the future-request window | Fix the request parameters; do not retry with the same value          |
+| `INVALID_START_BLOCK_NUMBER` (4) | `start_block_number` is structurally invalid — below `first_available_block`, or beyond the future-request window | Fix the request parameters; do not retry with the same value          |
 | `NOT_AVAILABLE` (6)              | The block is not available on this BN at this time (e.g. pruned, or node is still backfilling)                              | Retry later with exponential backoff, or query a different Block Node |
 
 ### What should a subscriber do on `NOT_AVAILABLE`?
 
 The subscriber **may retry with exponential backoff** against the same BN, or
 **fall over to a lower-priority BN** that covers the requested block range. Check
-`serverStatus` on available BNs to find one whose range includes the needed block.
+`serverStatusDetail` on available BNs to find one whose `available_ranges` includes the needed block.
 
-Detect gaps client-side by comparing `end_of_block.block_number` against
-`(last_committed_block + 1)`. Do not rely on the BN to signal gaps — the BN feeds
-slow subscribers from history, not from the live stream, and does not produce gaps.
+Clients may avoid this result by checking for gaps client-side. Gaps are reported
+clearly when calling `serverStatusDetail`, which returns `available_ranges` covering
+all blocks available on that node. If the node has all blocks it will return a single
+range from `0` through the latest persisted block.
+
+The Subscribe API stream at the live edge, because it is live and unverified, _may_
+deliver blocks out of order, but _should_ deliver the out-of-order block within a
+reasonable time.
 
 > See [Mirror Node Integration](./mirror-node-integration.md) for more details.
 
@@ -153,8 +149,8 @@ its `BlockProof`.
 | `NOT_FOUND` (4)     | The block **should** be available on this BN (within its managed range) but was not found — may be a transient storage issue |
 | `NOT_AVAILABLE` (5) | The block is **outside** this BN's managed range — the BN has never held it or it has been pruned                            |
 
-For `NOT_AVAILABLE`, call `serverStatus` to determine the BN's `firstAvailableBlock`
-and `lastAvailableBlock`, then query a BN that covers the needed range.
+For `NOT_AVAILABLE`, call `serverStatusDetail` to determine the BN's `available_ranges`,
+then query a BN that covers the needed range.
 
 ### Is `stateSnapshot` implemented?
 
@@ -217,12 +213,12 @@ block data:
 
 |      Mode       |                                     Behaviour                                      |
 |-----------------|------------------------------------------------------------------------------------|
-| `FILE`          | Write to local disk only; Block Nodes receive nothing                              |
-| `FILE_AND_GRPC` | Write to local disk **and** stream to configured Block Nodes (recommended default) |
-| `GRPC`          | Stream to Block Nodes only; no local files written                                 |
+| `FILE`          | Write to local disk only; Block Nodes receive nothing                                        |
+| `FILE_AND_GRPC` | Write to local disk **and** stream to configured Block Nodes (recommended before "cutover") |
+| `GRPC`          | Stream to Block Nodes only; no local files written (required after "cutover")               |
 
-Use `GRPC` when the CN should not retain local block files and all history is owned by
-Block Nodes. Use `FILE_AND_GRPC` to keep a local backup while streaming.
+Use `GRPC` when the CN must stream to Block Nodes only and all history is owned by Block Nodes.
+Use `FILE_AND_GRPC` before "cutover" to send blocks while still directly uploading data to legacy cloud buckets.
 
 > See [Configure Consensus Node Streaming](./operations/consensus-node-to-block-node-configuration.md) for more details.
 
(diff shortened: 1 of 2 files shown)


deshmukhpranali:
@kfbr can you please review this PR?

docs/block-node/integration-faq.md
```diff
@@ -0,0 +1,334 @@
...
+
+```bash
+# Example: find the CN version the current BN targets
+grep "hederaVersion" hiero-dependency-versions/build.gradle.kts
+# → val hederaVersion = "0.76.0-rc.1"
+```
+
+The E2E test suite (`solo-e2e-test.yml`) always runs against the latest available
+version of each component dynamically rather than pinned pairs. For production
+compatibility guidance, consult your Hashgraph PoC.
```
kfbr: Will there be a formal compatibility matrix? @Nana-EC this is something we've discussed, or at least some policy of N+/- some version.

jsync-swirlds: For each CN version, starting with `0.75`, we _might_ offer a "minimum recommended block node version".  BN has a core goal to be compatible with future CN versions for as long as practical, however, so I suspect the version will be something like BN `1.0.0` at CN `0.77` and stay at that version for a very long time.

That doesn't mean we do not recommend using the latest version of the BN in general, only that the "compatible" version **should** be very slow to change.

docs/block-node/integration-faq.md
```diff
@@ -0,0 +1,334 @@
...
+the block. **The block is not necessarily lost** — the CN still has it. The CN must:
+
+1. Start a new stream and resend the block to this BN or a different reliable BN.
+2. **Not discard the block** until it has been acknowledged as persisted and verified by
+   at least one BN.
+
+### What does `BAD_BLOCK_PROOF` mean?
+
+`BAD_BLOCK_PROOF` (code 6 in `EndOfStream`) means a `BlockProof` item could not be
+validated. The CN must start a new stream beginning before the failed block.
```
timfn-hg: Not sure if it is worth calling out (or maybe it already is somewhere), but there is a bit of nuance with how the CN responds to various EndOfStream responses, and other connection-related failures. In particular, the CN may not restart the stream to the same BN if there are other BNs available to stream to. And then then there is a concept of cool down - which you can think of as a timeout/penalty box for misbehaving BNs. Depending on the type of failure (or EndOfStream response) a BN may or may not be placed into cool down. You can see more information about this [here](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/streaming/CloseReason.java) and [here](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/streaming/CoolDownType.java). This may all be too far "in the weeds" for the documentation, but if you think it is worthwhile and want to talk through it, let me know!

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+|------------------------|-------------------------------------------------|-----------------------------------------------------|
+| `BlockAcknowledgement` | After block proof received and verified         | Continue streaming the next block                   |
+| `SkipBlock`            | Another publisher already delivered this block  | Skip to the block indicated and continue            |
+| `ResendBlock`          | BN needs the CN to resend from a specific block | Resend from the block number specified              |
+| `BehindPublisher`      | BN is behind — CN is too far ahead              | Start a new stream from the block number specified  |
+| `EndOfStream`          | Terminal — stream is closed                     | See the status code and take the appropriate action |
+
+### What does `DUPLICATE_BLOCK` mean and what should the CN do?
+
+`DUPLICATE_BLOCK` (code 5 in `EndOfStream`) means the block header the CN sent
```
jsync-swirlds: I'm not sure we want to call out the ordinal on these, as the enumeration is defined in the schema and any SDK implementations will have matching enumerations or constants defined.

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+### How does a Consensus Node initiate a publish stream to a Block Node?
+
+The CN opens a **bidirectional gRPC stream** to
+`BlockStreamPublishService.publishBlockStream`. The **first message in the stream must
+be a `BlockHeader`** for the next expected block. The BN inspects the block number and
+responds:
+
+- If the block number is the next expected block → streaming continues normally.
+- If the block number ≤ last verified block → BN responds with `DUPLICATE_BLOCK`.
+- If the block number > next expected → BN responds with `BehindPublisher`.
```
jsync-swirlds, line 28: One missing case:
`- If the block number is one currently being streamed by another CN → BN responds with `SKIP_BLOCK` and streaming continues with the next block.`

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+> See [Publish Block Stream](../design/communication-protocol/publish-block-stream.md) for more details.
+
+### What does the Block Node send back during streaming?
+
+The BN can send five response types during an active stream:
+
+|        Response        |                    When sent                    |                What the CN should do                |
+|------------------------|-------------------------------------------------|-----------------------------------------------------|
+| `BlockAcknowledgement` | After block proof received and verified         | Continue streaming the next block                   |
+| `SkipBlock`            | Another publisher already delivered this block  | Skip to the block indicated and continue            |
```
jsync-swirlds: Minor adjustments:
1. Acknowledgement means the block is secured and should be noted as such (possibly removing the block from buffers).
1. Skip does not specify a block to _skip to_, it returns the _current_ block so the skip can be matched to the corresponding header already sent (in case more than one block is sent before the skip is received), and means "skip the current block".

```suggestion
| `BlockAcknowledgement` | After block proof received and verified         | Mark block secured and continue streaming the next block |
| `SkipBlock`            | Another publisher is currently sending this block  | Skip the current block and continue with the next |
```

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+
+### What does `TOO_FAR_BEHIND` mean and what should the CN do?
+
+`TOO_FAR_BEHIND` (code 4 in `EndOfStream`) means the BN is too far behind the CN to
+catch up by streaming alone. The CN should:
+
+1. Switch streaming to a different Block Node.
+2. Include `earliest_block` and `latest_block` in the `EndStream` message so the
+   falling-behind BN can attempt to backfill from a peer.
+3. Resume streaming to this BN later once it has caught up via backfill.
```
jsync-swirlds: This enum no longer exists (see enum [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/protobuf-sources/src/main/proto/block-node/api/block_stream_publish_service.proto#L360)).

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+`BAD_BLOCK_PROOF` (code 6 in `EndOfStream`) means a `BlockProof` item could not be
+validated. The CN closes the stream and resumes streaming — to this BN or a different
+available Block Node — from before the failed block.
+
+### When does the Block Node send `TIMEOUT` to the Consensus Node?
+
+`TIMEOUT` (code 4 in `EndOfStream`) is sent when the delay between stream items
+exceeds the BN's configured timeout — the CN did not deliver the next item within the
+allowed window. The CN closes the stream and resumes streaming — to this BN or a
+different available Block Node — from before the timed-out block.
```
jsync-swirlds:
```suggestion
different available Block Node — from the timed-out block.
```

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+`block_items`, `end_of_block`, and finally a terminal `status` code when the stream
+closes.
+
+> See [Mirror Node Integration](./mirror-node-integration.md) for more details.
+
+### What is the difference between `INVALID_START_BLOCK_NUMBER` and `NOT_AVAILABLE`?
+
+|               Code               |                                                           Meaning                                                           |                              What to do                               |
+|----------------------------------|-----------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------|
+| `INVALID_START_BLOCK_NUMBER` (4) | `start_block_number` is structurally invalid — negative, below `first_available_block`, or beyond the future-request window | Fix the request parameters; do not retry with the same value          |
```
jsync-swirlds: Technically negative numbers aren't possible (the field is `uint64`), so the value is either below the first available block, or too far in the future.

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+| `NOT_AVAILABLE` (6)              | The block is not available on this BN at this time (e.g. pruned, or node is still backfilling)                              | Retry later with exponential backoff, or query a different Block Node |
+
+### What should a subscriber do on `NOT_AVAILABLE`?
+
+The subscriber **may retry with exponential backoff** against the same BN, or
+**fall over to a lower-priority BN** that covers the requested block range. Check
+`serverStatus` on available BNs to find one whose range includes the needed block.
```
jsync-swirlds:
```suggestion
`serverStatusDetail` on available BNs to find one whose available ranges include the needed block.
```

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+
+Detect gaps client-side by comparing `end_of_block.block_number` against
+`(last_committed_block + 1)`. Do not rely on the BN to signal gaps — the BN feeds
+slow subscribers from history, not from the live stream, and does not produce gaps.
```
jsync-swirlds: This is a bit confusing.
I'm not sure what gap detection is about here in relation to what to do on "not available".
BN feeds subscribers from a combination of live and history as needed, and _may_ return not available at any point if it encounters a gap.

_Possible_ rewrite (not sure, just a first pass)

```suggestion
Clients may avoid this result by checking for gaps client-side.
Gaps are reported clearly when calling `serverStatusDetail` which returns
`available_ranges` covering all blocks available on that node. If the node has
all blocks then it will return a single range from `0` through the latest
persisted block.
The Subscribe API stream delivering blocks at the "live" edge, because it is
live and unverified, _may_ deliver blocks out of order, but _should_ deliver
the out-of-order block within a reasonable time.
```

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+
+### What is the difference between `NOT_FOUND` and `NOT_AVAILABLE` in `getBlock`?
+
+|        Code         |                                                           Meaning                                                            |
+|---------------------|------------------------------------------------------------------------------------------------------------------------------|
+| `NOT_FOUND` (4)     | The block **should** be available on this BN (within its managed range) but was not found — may be a transient storage issue |
+| `NOT_AVAILABLE` (5) | The block is **outside** this BN's managed range — the BN has never held it or it has been pruned                            |
+
+For `NOT_AVAILABLE`, call `serverStatus` to determine the BN's `firstAvailableBlock`
+and `lastAvailableBlock`, then query a BN that covers the needed range.
```
jsync-swirlds:
```suggestion
For `NOT_AVAILABLE`, call `serverStatusDetail` to determine the BN's `available_ranges`,
then query a BN that covers the needed range.
```

docs/block-node/faq/integration-faq.md
```diff
@@ -0,0 +1,343 @@
...
+block data:
+
+|      Mode       |                                     Behaviour                                      |
+|-----------------|------------------------------------------------------------------------------------|
+| `FILE`          | Write to local disk only; Block Nodes receive nothing                              |
+| `FILE_AND_GRPC` | Write to local disk **and** stream to configured Block Nodes (recommended default) |
+| `GRPC`          | Stream to Block Nodes only; no local files written                                 |
+
+Use `GRPC` when the CN should not retain local block files and all history is owned by
+Block Nodes. Use `FILE_AND_GRPC` to keep a local backup while streaming.
```
jsync-swirlds: This is subtly inaccurate:

```suggestion
|      Mode       |                                     Behaviour                                      |
|-----------------|------------------------------------------------------------------------------------|
| `FILE`          | Write to local disk only; Block Nodes receive nothing                              |
| `FILE_AND_GRPC` | Write to local disk **and** stream to configured Block Nodes (recommended before "cutover") |
| `GRPC`          | Stream to Block Nodes only; no local files written (Required after "cutover")               |

Use `GRPC` when the CN must stream to block nodes only, and all history is owned by Block Nodes.
Use `FILE_AND_GRPC` before "cutover" to send blocks while still directly uploading data to legacy cloud buckets.
```

jsync-swirlds (commented):
Some suggested improvements or adjustments.
One is highly speculative and might be incorrect (noted as "_Possible_ rewrite").

jsync-swirlds:
Spotless needs to be run for this PR; that's what is causing the PR checks to fail.
`./gradlew spotlessApply` should accomplish the required fixes.

kfbr (approved):

jsync-swirlds (approved):

Nana-EC (approved):

PR #3165, opened 2026-07-07
feat: Make ServerStatus aware of the earliest managed block setting

### LINKED ISSUE #2971: Block Node Status API should consider earliestManagedBlock configuration
### Story Form

As a Publisher
I want to Stream blocks based on status API information that considers earliestManagedBlock information
So that I know when I can stream data, even if the block node is in the midst of a historical backfill.

### Technical Notes

Currently, the status API reports a "wanted" block, which the publishers use to determine if they should attempt to send blocks to that node. This "wanted" block is only based on blocks stored and/or available, and does not consider the `earliestManagedBlock` configuration.
The result is that a new backfilling block node may never get any blocks from publishers because it holds older blocks, and even though it could accept much later blocks, it claims it "wants" something the publishers cannot supply.

What we intend to change is that if `earliestManagedBlock` is set, the status API will not return a "wanted" block _lower_ than that value, but will set a _later_ value if the block node holds later blocks.

berryware:
## Reviewer Notes
This implements the fix for [#2971](https://github.com/hiero-ledger/hiero-block-node/issues/2971): the Status API's `serverStatus` RPC now floors the reported `lastAvailableBlock` at the configured `earliestManagedBlock` when the node actually holds blocks (leaving `firstAvailableBlock` untouched, per the issue's clarified invariant `first <= earliestManagedBlock <= last`), and leaves both as `UNKNOWN_BLOCK_NUMBER` when the node holds no blocks at all.

Changes:
- [ServerStatusServicePlugin.java](block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java) — reads `NodeConfig.earliestManagedBlock()` in `init()`, applies the floor to `lastAvailableBlock` in `serverStatus()`.
- [module-info.java](block-node/server-status/src/main/java/module-info.java) — added `requires org.hiero.block.node.app.config`.
- [ServerStatusServicePluginTest.java](block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java) — added 3 tests covering the raise case, the no-op case (already at/above EMB), and the empty-node case.

## Related Issue(s)
Closes: #2971

### DIFF
--- a/block-node/server-status/src/main/java/module-info.java
+++ b/block-node/server-status/src/main/java/module-info.java
@@ -1,6 +1,4 @@
 // SPDX-License-Identifier: Apache-2.0
-/* SPDX-License-Identifier: Apache-2.0 */
-
 import org.hiero.block.node.server.status.ServerStatusServicePlugin;
 
 module org.hiero.block.node.server.status {
@@ -18,8 +16,6 @@ module org.hiero.block.node.server.status {
     requires org.hiero.block.node.base;
     requires com.github.spotbugs.annotations;
 
-    uses com.swirlds.config.api.spi.ConfigurationBuilderFactory;
-    /* export configuration classes to the config module and app */
     provides org.hiero.block.node.spi.BlockNodePlugin with
             ServerStatusServicePlugin;
 }
--- a/block-node/server-status/src/main/java/module-info.java
+++ b/block-node/server-status/src/main/java/module-info.java
@@ -1,10 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
+/* SPDX-License-Identifier: Apache-2.0 */
+
 import org.hiero.block.node.server.status.ServerStatusServicePlugin;
 
 module org.hiero.block.node.server.status {
-    uses com.swirlds.config.api.spi.ConfigurationBuilderFactory;
-
-    // export configuration classes to the config module and app
     exports org.hiero.block.node.server.status to
             com.swirlds.config.impl,
             com.swirlds.config.extensions,
@@ -15,9 +14,12 @@ module org.hiero.block.node.server.status {
     requires transitive org.hiero.block.protobuf.pbj;
     requires transitive org.hiero.metrics;
     requires com.hedera.pbj.runtime;
+    requires org.hiero.block.node.app.config;
     requires org.hiero.block.node.base;
     requires com.github.spotbugs.annotations;
 
+    uses com.swirlds.config.api.spi.ConfigurationBuilderFactory;
+    /* export configuration classes to the config module and app */
     provides org.hiero.block.node.spi.BlockNodePlugin with
             ServerStatusServicePlugin;
 }
--- a/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
+++ b/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
@@ -59,12 +59,13 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
 
         final ServerStatusResponse.Builder serverStatusResponseBuilder = ServerStatusResponse.newBuilder();
         final long firstAvailableBlock = blockProvider.availableBlocks().min();
-        long lastAvailableBlock = blockProvider.availableBlocks().max();
+        long highestAvailableBlock = blockProvider.availableBlocks().max();
         // If this node is configured to manage blocks starting later than what it currently holds, report that
         // later block as the last available block so publishers know they can stream from earliestManagedBlock
-        // instead of assuming the node only wants its currently stored (older) blocks.
-        if (lastAvailableBlock != UNKNOWN_BLOCK_NUMBER && lastAvailableBlock < earliestManagedBlock) {
-            lastAvailableBlock = earliestManagedBlock;
+        // instead of assuming the node only wants older blocks between the oldest stored block and earliest managed
+        // block.
+        if (highestAvailableBlock != UNKNOWN_BLOCK_NUMBER && highestAvailableBlock < earliestManagedBlock) {
+            highestAvailableBlock = earliestManagedBlock;
         }
 
         // TODO(#579) Should get from state config or status, which would be provided by the context from responsible
@@ -78,7 +79,7 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
         // Build the response
         ServerStatusResponse response = serverStatusResponseBuilder
                 .firstAvailableBlock(firstAvailableBlock)
-                .lastAvailableBlock(lastAvailableBlock)
+                .lastAvailableBlock(highestAvailableBlock)
                 .onlyLatestState(onlyLatestState)
                 .build();
 
@@ -174,7 +175,7 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
     @Override
     @NonNull
     public List<Class<? extends Record>> configDataTypes() {
-        return List.of(ServerStatusConfig.class, NodeConfig.class);
+        return List.of(ServerStatusConfig.class);
     }
 
     /**
--- a/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
+++ b/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
@@ -12,6 +12,7 @@ import org.hiero.block.api.BlockRange;
 import org.hiero.block.api.ServerStatusDetailResponse;
 import org.hiero.block.api.ServerStatusRequest;
 import org.hiero.block.api.ServerStatusResponse;
+import org.hiero.block.node.app.config.node.NodeConfig;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
 import org.hiero.block.node.spi.ServiceBuilder;
@@ -38,6 +39,8 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
     private HistoricalBlockFacility blockProvider;
     /** The block node context, used to provide access to facilities */
     private volatile BlockNodeContext blockNodeContext;
+    /** The earliest block number this node is configured to manage */
+    private long earliestManagedBlock;
     /** Counter for the number of status requests */
     private LongCounter.Measurement requestStatusCounter;
     /** Counter for the number of detail requests */
@@ -56,7 +59,13 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
 
         final ServerStatusResponse.Builder serverStatusResponseBuilder = ServerStatusResponse.newBuilder();
         final long firstAvailableBlock = blockProvider.availableBlocks().min();
-        final long lastAvailableBlock = blockProvider.availableBlocks().max();
+        long lastAvailableBlock = blockProvider.availableBlocks().max();
+        // If this node is configured to manage blocks starting later than what it currently holds, report that
+        // later block as the last available block so publishers know they can stream from earliestManagedBlock
+        // instead of assuming the node only wants its currently stored (older) blocks.
+        if (lastAvailableBlock != UNKNOWN_BLOCK_NUMBER && lastAvailableBlock < earliestManagedBlock) {
+            lastAvailableBlock = earliestManagedBlock;
+        }
 
         // TODO(#579) Should get from state config or status, which would be provided by the context from responsible
         // facility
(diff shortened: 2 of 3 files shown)


block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -163,7 +174,7 @@ public void init(@NonNull final BlockNodeContext context, @NonNull final Service
     @Override
     @NonNull
     public List<Class<? extends Record>> configDataTypes() {
-        return List.of(ServerStatusConfig.class);
+        return List.of(ServerStatusConfig.class, NodeConfig.class);
```
jsync-swirlds: Don't add node config here, it's added by the app.
This method is only for config types that are _new_ with this plugin.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -56,7 +59,13 @@ public ServerStatusResponse serverStatus(@NonNull final ServerStatusRequest requ
 
         final ServerStatusResponse.Builder serverStatusResponseBuilder = ServerStatusResponse.newBuilder();
         final long firstAvailableBlock = blockProvider.availableBlocks().min();
-        final long lastAvailableBlock = blockProvider.availableBlocks().max();
+        long lastAvailableBlock = blockProvider.availableBlocks().max();
+        // If this node is configured to manage blocks starting later than what it currently holds, report that
+        // later block as the last available block so publishers know they can stream from earliestManagedBlock
+        // instead of assuming the node only wants its currently stored (older) blocks.
```
jsync-swirlds: This seems to be saying the block node would want blocks it already has; which is not the case; it's the block node expecting blocks below EMB that it does _not_ yet have.

```suggestion
        // instead of assuming the node only wants older blocks between the oldest stored block and earliest managed block.
```

block-node/server-status/src/main/java/module-info.java
```diff
@@ -1,10 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
+/* SPDX-License-Identifier: Apache-2.0 */
```
jsync-swirlds: Is this a spotless change?  If so it's kind of odd...

block-node/server-status/src/main/java/module-info.java
```diff
@@ -15,9 +14,12 @@
     requires transitive org.hiero.block.protobuf.pbj;
     requires transitive org.hiero.metrics;
     requires com.hedera.pbj.runtime;
+    requires org.hiero.block.node.app.config;
     requires org.hiero.block.node.base;
     requires com.github.spotbugs.annotations;
 
+    uses com.swirlds.config.api.spi.ConfigurationBuilderFactory;
+    /* export configuration classes to the config module and app */
```
jsync-swirlds: This doesn't make sense where it is, we should either move back, or remove.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -56,7 +59,13 @@ public ServerStatusResponse serverStatus(@NonNull final ServerStatusRequest requ
 
         final ServerStatusResponse.Builder serverStatusResponseBuilder = ServerStatusResponse.newBuilder();
         final long firstAvailableBlock = blockProvider.availableBlocks().min();
-        final long lastAvailableBlock = blockProvider.availableBlocks().max();
+        long lastAvailableBlock = blockProvider.availableBlocks().max();
```
ata-nas: Nit: the name `lastAvailableBlock` is kind of confusing, it could be misunderstood conceptually to mean lowest block. I think `highestAvailableBlock` or something similar could be better.

block-node/server-status/src/main/java/module-info.java
ata-nas: There have been changes with our gradle setup as of recent and some issues alongside with it.
Maybe @jjohannes knows more as to why this happens.

block-node/server-status/src/main/java/module-info.java
```diff
@@ -1,10 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
+/* SPDX-License-Identifier: Apache-2.0 */
+
 import org.hiero.block.node.server.status.ServerStatusServicePlugin;
 
 module org.hiero.block.node.server.status {
-    uses com.swirlds.config.api.spi.ConfigurationBuilderFactory;
-
-    // export configuration classes to the config module and app
```
ata-nas, line 7: This removal, I suppose, is possibly again the result of recent changes to gradle setup.
That, of course, if this is not intentional.

block-node/server-status/src/main/java/module-info.java
ata-nas: I think this is due to changes recently in gradle setup. @jjohannes I believe knows more. The word `export` seems to be treated as a keyword for the formatter.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -56,7 +59,13 @@ public ServerStatusResponse serverStatus(@NonNull final ServerStatusRequest requ
...
+        // If this node is configured to manage blocks starting later than what it currently holds, report that
+        // later block as the last available block so publishers know they can stream from earliestManagedBlock
+        // instead of assuming the node only wants its currently stored (older) blocks.
+        if (lastAvailableBlock != UNKNOWN_BLOCK_NUMBER && lastAvailableBlock < earliestManagedBlock) {
+            lastAvailableBlock = earliestManagedBlock;
+        }
```
ata-nas: I am missing something here. If we say that our highest available block == EMB, but we actually do not have certain blocks, is that not misleading?

ata-nas (commented):
Some questions

block-node/server-status/src/main/java/module-info.java
berryware: There was some goofiness with spotlessApply. Removing

block-node/server-status/src/main/java/module-info.java
ivannov: It's the comment actually, not the word export. Here is what I have from @jjohannes on a recent communication between us:

> Handling comments that are on a separate line is difficult in automatic sorting. But we should somehow deal with this better and not do a wrong formatting. (Follow up for me so that this behaves better in the future: https://github.com/hiero-ledger/hiero-gradle-conventions/issues/519). You can either remove the comment, if you are fine with that. Or add it at the end of the definition line itself

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -56,7 +59,13 @@ public ServerStatusResponse serverStatus(@NonNull final ServerStatusRequest requ
 
         final ServerStatusResponse.Builder serverStatusResponseBuilder = ServerStatusResponse.newBuilder();
         final long firstAvailableBlock = blockProvider.availableBlocks().min();
-        final long lastAvailableBlock = blockProvider.availableBlocks().max();
+        long lastAvailableBlock = blockProvider.availableBlocks().max();
+        // If this node is configured to manage blocks starting later than what it currently holds, report that
+        // later block as the last available block so publishers know they can stream from earliestManagedBlock
+        // instead of assuming the node only wants its currently stored (older) blocks.
+        if (lastAvailableBlock != UNKNOWN_BLOCK_NUMBER && lastAvailableBlock < earliestManagedBlock) {
+            lastAvailableBlock = earliestManagedBlock;
```
ivannov: Maybe I misunderstand the requirement. But let's say that the last available block for this node is 5. While the earliest managed block is 100.

Does this mean that a backfill process on another node will try to obtain block 100 from this node? And will be mislead along this path?

block-node/server-status/src/main/java/module-info.java
jjohannes: There was a bug in this area (sorry!). Would it be possible to update the Gradle plugins in `settings.gradle.kts` to `0.7.11` in this PR. And then run the spotlessApply again?

Regarding comments:
After the upgrade comments are preserved. They are "attached" to the line below the comment. So if a comment ends up in the wrong places after the "spotlessApply" you should be able to move it to where you want it and then it should stay there.

block-node/server-status/src/main/java/module-info.java
berryware: removed the comment as it was causing spotlessCheck failures, even after running spotlessApply

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: reverted

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: @jsync-swirlds explains this in the [ticket](https://github.com/hiero-ledger/hiero-block-node/issues/2971) comments.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: updated.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: renamed

block-node/server-status/src/main/java/module-info.java
berryware: Removed comment and removed the uses  com.swirlds.config.api.spi.ConfigurationBuilderFactory;

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: @jsync-swirlds explains this in the https://github.com/hiero-ledger/hiero-block-node/issues/2971 comments.

block-node/server-status/src/main/java/module-info.java
jsync-swirlds: Are we sure this plugin does not use the factory?

block-node/server-status/src/main/java/module-info.java
jsync-swirlds: Removing the comment is fine.
Are we certain removing the `uses` clause is correct?

jsync-swirlds (approved):

block-node/server-status/src/main/java/module-info.java
ata-nas: I think we need it, not sure, however.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
ata-nas: Right, I hope that does not conflict with backfill. That one should use the server status detail, however. Just making sure we do not cause issue for the backfill. I see this is intentional.

jsync-swirlds: Backfill uses detail status because it needs to detect gaps.

AlfredoG87: correct, in the latest PR, we changed it to use detailStatus indeed. 👍

AlfredoG87 (approved):
LGTM 💯

block-node/server-status/src/main/java/module-info.java
Nana-EC: Moving consideration to #3176

block-node/server-status/src/main/java/module-info.java
Nana-EC: Moving consideration to #3176

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
Nana-EC: Moving consideration to #3176

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
Nana-EC: Moving consideration to #3176

PR #3166, opened 2026-07-07
fix: align convert-address-book-history output with RangedAddressBookHistory (#3164)

### LINKED ISSUE #3164: convert-address-book-history: align output with RangedAddressBookHistory JSON
## Context

`blocks convert-address-book-history` (added in #3071 as T3 of epic #2954) currently emits a hand-rolled JSON shape (top-level `entries[]`, Base64-encoded `NodeAddressBook` payload, `null` sentinel for open-ended era, embedded `blockTimestamp`).

The BN-side loader that lands with T4 (#2958, PR #3070) reads `RangedAddressBookHistory` PBJ JSON at `ApplicationStateConfig.rsaBootstrapFilePath`. The two shapes don't match, so the file the tool produces today can't be loaded by the BN as a historical roster.

## Required changes

To align T3's output with what T4 actually reads:

1. **Top-level key**: `"entries"` → `"addressBooks"`.
2. **`addressBook` field**: currently a Base64-encoded string of raw `NodeAddressBook` protobuf bytes. It needs to become a **nested JSON object** in `NodeAddressBook.JSON` shape (e.g. `{ "nodeAddress": [ { "nodeId": ..., "RSAPubKey": "...", ... }, ... ] }`).
3. **`endBlock` sentinel**: currently `null` for open-ended (last era). `RangedAddressBookHistory.JSON` uses the integer `-1` (per `BlockNodeAppTest.java:583` — `.endBlock(-1L)`).
4. **`blockTimestamp` field**: not part of `RangedNodeAddressBook`; drop entirely.

## Additional related work

- **Default `--output` path**: currently `/opt/hiero/block-node/application-state/rsa-bootstrap-roster-history.json`. Should be `/opt/hiero/block-node/application-state/rsa-bootstrap-roster.json` — the same file the BN reads via `app.state.rsaBootstrapFilePath` (T4's loader accepts either single-book `NodeAddressBook` or `RangedAddressBookHistory` at that path).
- **`--network` inheritance**: `ConvertAddressBookHistoryCommand` uses `new BlockTimeReader(blockTimesFile)` which hardcodes mainnet genesis. Should use `BlockTimeReader.forCurrentNetwork(blockTimesFile)` (like `ToWrappedBlocksCommand.java:331`, `DownloadDaysV2.java:62`) so the top-level `--network previewnet`/`testnet` flag propagates.

## Suggested implementation

Rather than continue to hand-roll JSON, emit `RangedAddressBookHistory` via PBJ's own codec (`RangedAddressBookHistory.JSON.write(...)`). That gets format compatibility for free, drops the Jackson dependency in `tools`, and keeps this file trivially consumable by the BN loader.

## Refs

- Epic: #2954
- T3 (introduced current shape): #2957 / PR #3071
- T4 (defines the target shape): #2958 / PR #3070
- Design: `docs/design/wrb-streaming/sp-wrb-bn-design.md`

rockysingh:
## Summary

Rewrite the `blocks convert-address-book-history` output so the file the tool produces is directly loadable by the BN's T4 historical-roster bootstrap (which reads `RangedAddressBookHistory` PBJ JSON at `ApplicationStateConfig.rsaBootstrapFilePath`).

Fixes #3164.

### Format changes (from #3071)

| | Before | After |
|---|---|---|
| Top-level key | `"entries"` | `"addressBooks"` |
| Per-era `addressBook` | base64 string of `NodeAddressBook` proto | **nested** `NodeAddressBook` JSON object (`{ "nodeAddress": [ { "nodeId": …, "RSAPubKey": …, … } ] }`) |
| Per-era `endBlock` open-ended sentinel | `null` | `-1` (matches `RangedAddressBookHistory.JSON`) |
| `blockTimestamp` field | present | **removed** (not part of `RangedNodeAddressBook`) |

Implementation now emits via `RangedAddressBookHistory.JSON.write(...)` so wire compatibility is exact.

### Additional related fixes

- **Default `--output`** → `/opt/hiero/block-node/application-state/rsa-bootstrap-roster.json` (same file the BN reads via `app.state.rsaBootstrapFilePath`; T4's loader accepts either single-book or ranged-history JSON at that path).
- **`BlockTimeReader` construction** — the raw `new BlockTimeReader(blockTimesFile)` ctor is retained deliberately, even when the caller passes `--network testnet` or `--network previewnet`. The extraction writer chain hardcodes mainnet as the block-time anchor via `RecordFileDates.instantToBlockTimeLong`, so the reader must use the same mainnet anchor to round-trip the same bytes correctly. Switching to `BlockTimeReader.forCurrentNetwork(blockTimesFile)` would apply a network-specific anchor and produce a multi-year offset on non-mainnet networks. Design decision is documented inline (see the block comment above the ctor call).

### Note on dependencies

Jackson (`com.fasterxml.jackson.databind`) is still a `tools` module dependency and is used inside the reworked read-back / test path to parse the emitted JSON and assert the exact wire shape. An earlier draft of this PR removed it; that was reverted during review because Jackson is the cleanest way to walk the nested `NodeAddressBook` JSON tree in tests without pulling in PBJ codec setup.

## Test plan

- [x] `./gradlew :tools:qualityCheck :tools:test` — spotless + module-directives clean.
- [x] Reworked tests parse output with `RangedAddressBookHistory.JSON.parse` and assert the exact wire shape (`addressBooks`, nested `nodeAddress[]`, `RSAPubKey`, no `blockTimestamp`, no `entries`).
- [ ] CI green
- [ ] Manual smoke on previewnet: convert `previewnet/wrappedBlocks/addressBookHistory.json` and load into a BN test harness

## Refs

- Follow-up to #3071 (T3)
- Consumed by #3070 (T4 loader)
- Epic: #2954

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
@@ -253,10 +253,19 @@ public class ConvertAddressBookHistoryCommand implements Callable<Integer> {
     }
 
     /**
-     * Re-emit the roster JSON at {@code file} so every entry has an explicit
-     * {@code startBlock} and {@code endBlock} field, defaulting to {@code 0} and
-     * {@code -1} respectively when the PBJ codec elided them. Preserves entry order
-     * and the nested {@code addressBook} structure verbatim.
+     * Re-emit the roster JSON at {@code file} so every entry has explicit fields that PBJ's
+     * JSON codec would otherwise elide as proto3 defaults:
+     *
+     * <ul>
+     *   <li>Per-era: {@code startBlock} defaults to {@code 0}, {@code endBlock} defaults to
+     *       {@code -1} (the {@link #OPEN_ENDED_END_BLOCK} sentinel used only for the last era).</li>
+     *   <li>Per {@code nodeAddress}: {@code nodeId} defaults to {@code 0}. Without this,
+     *       every genesis-node-0 entry loses its identifier and the BN's roster verifier
+     *       can't route by nodeId.</li>
+     * </ul>
+     *
+     * Preserves entry order and the nested {@code addressBook} structure verbatim. PBJ's
+     * parse side reconstructs the same values from either shape, so the BN loader is unaffected.
      */
     private static void ensureExplicitStartAndEndBlock(Path file) throws IOException {
         final ObjectMapper mapper = new ObjectMapper();
@@ -276,6 +285,18 @@ public class ConvertAddressBookHistoryCommand implements Callable<Integer> {
             }
             if (!entryObj.has("endBlock")) {
                 entryObj.put("endBlock", Long.toString(OPEN_ENDED_END_BLOCK));
+            }
+            // Restore any nodeAddress[].nodeId that PBJ elided as the uint64 default (0).
+            final JsonNode addressBookNode = entryObj.get("addressBook");
+            if (addressBookNode instanceof ObjectNode addressBookObj) {
+                final JsonNode nodeAddressNode = addressBookObj.get("nodeAddress");
+                if (nodeAddressNode != null && nodeAddressNode.isArray()) {
+                    for (final JsonNode addr : nodeAddressNode) {
+                        if (addr instanceof ObjectNode addrObj && !addrObj.has("nodeId")) {
+                            addrObj.put("nodeId", Long.toString(0L));
+                        }
+                    }
+                }
             }
         }
         mapper.writerWithDefaultPrettyPrinter().writeValue(file.toFile(), root);
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
@@ -119,15 +119,19 @@ public class ConvertAddressBookHistoryCommand implements Callable<Integer> {
         }
         System.out.println("  Eras:            " + sorted.size());
 
-        // Use forCurrentNetwork() so that the genesis instant respects the top-level --network
-        // option (mainnet / testnet / previewnet / other). The raw BlockTimeReader(Path) ctor
-        // hardcodes the mainnet genesis, which would break resolution on any other network.
+        // Use the raw ctor (mainnet-anchored) even for --network testnet/previewnet. The
+        // block_times.bin file format stores each block's time as nanos-since-mainnet-first-
+        // block regardless of network: every writer in the extraction chain hardcodes that
+        // anchor via RecordFileDates.instantToBlockTimeLong. Using forCurrentNetwork() here
+        // reads the same bytes with a network-specific anchor and produces a multi-year
+        // offset on non-mainnet networks (see PR #3166 review). Reader and writer must
+        // agree — leave both mainnet-anchored until the write path is network-aware.
         final long[] startBlocks;
         final List<String> resolutionProblems;
         final Instant coverageStart;
         final Instant coverageEnd;
         final long coverageMaxBlock;
-        try (BlockTimeReader reader = BlockTimeReader.forCurrentNetwork(blockTimesFile)) {
+        try (BlockTimeReader reader = new BlockTimeReader(blockTimesFile)) {
             coverageMaxBlock = reader.getMaxBlockNumber();
             coverageStart = reader.getBlockInstant(0);
             coverageEnd = reader.getBlockInstant(coverageMaxBlock);
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.tools.blocks;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.hedera.hapi.node.base.NodeAddressBook;
 import com.hedera.hapi.node.base.Timestamp;
 import com.hedera.pbj.runtime.ParseException;
@@ -208,6 +211,14 @@ public class ConvertAddressBookHistoryCommand implements Callable<Integer> {
             try (WritableStreamingData out = new WritableStreamingData(Files.newOutputStream(tmp))) {
                 RangedAddressBookHistory.JSON.write(rangedHistory, out);
             }
+            // PBJ's JSON codec elides proto3 default values (uint64 == 0), which makes a
+            // legitimate genesis-era startBlock=0 or open-ended endBlock look identical to a
+            // missing field. This bootstrap file gets inspected by operators, so re-emit it
+            // with both fields always present: startBlock defaults to 0 (proto3 default),
+            // endBlock defaults to -1 (OPEN_ENDED_END_BLOCK sentinel, used only for the last
+            // era). PBJ's parse side reconstructs the same values from either shape, so the
+            // BN loader is unaffected.
+            ensureExplicitStartAndEndBlock(tmp);
             Files.move(tmp, outputFile, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
         } catch (IOException e) {
             Files.deleteIfExists(tmp);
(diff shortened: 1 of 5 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
```diff
@@ -30,31 +32,29 @@
...
  *     },
  *     ...
  *   ]
  * }
  * }</pre>
  *
- * <p>{@code endBlock} is {@code (next.startBlock - 1)}; the most recent era's {@code endBlock}
- * is {@code null} to mark it as open-ended. The full {@code NodeAddressBook} proto is preserved
- * per era (Base64) so the BN can extract the fields it needs at load time.
+ * <p>{@code endBlock} for the last era is {@code -1}, the open-ended sentinel used by
```
Nana-EC, line 56: FYI the `node_service.proto` has comments for `RangedAddressBookHistory` that contradict.

>  * Entries MUST be ordered by start_block ascending and MUST NOT overlap.
>  * The last entry typically has end_block = 0, meaning it covers all block
>  * numbers >= start_block with no upper bound (open-ended).

Your PR doesn't introduce the issue but it does improve it so here or in follow up might be good to update that file so it's consistent with your update.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
```diff
@@ -166,6 +234,74 @@ private int convertAndWrite(List<DatedNodeAddressBook> sorted, long[] startBlock
...
+     *       {@code -1} (the {@link #OPEN_ENDED_END_BLOCK} sentinel used only for the last era).</li>
+     *   <li>Per {@code nodeAddress}: {@code nodeId} defaults to {@code 0}. Without this,
+     *       every genesis-node-0 entry loses its identifier and the BN's roster verifier
+     *       can't route by nodeId.</li>
+     * </ul>
+     *
+     * Preserves entry order and the nested {@code addressBook} structure verbatim. PBJ's
+     * parse side reconstructs the same values from either shape, so the BN loader is unaffected.
+     */
+    private static void ensureExplicitStartAndEndBlock(Path file) throws IOException {
```
Nana-EC: nit: You're doing more than just Start and End block

```suggestion
    private static void ensureExplicitPbjDefaults(Path file) throws IOException {
```

Nana-EC (dismissed):
Non blocking suggestions

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
rockysingh: Good catch. Fixed in c01dc7f1. Updated node_service.proto's docs to say the sentinel is -1 (max uint64 in wire form, rendered as "-1" by PBJ's JSON codec) in all three places it was mentioned: the RangedAddressBookHistory message doc, the RangedNodeAddressBook message doc, and the end_block field doc. Went with a docs update rather than flipping the tooling to 0 because -1 is what every generated bootstrap file and the BN loader already handle.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
rockysingh: Applied in c01dc7f1. The method definition was already renamed to ensureExplicitPbjDefaults but the L226 caller still had the old name -- brought the caller in line.

berryware:
Is the PR summary correct anymore? 

- "--network propagation — use BlockTimeReader.forCurrentNetwork(blockTimesFile) instead of new BlockTimeReader(blockTimesFile). The raw ctor hardcodes mainnet genesis, which broke the consensus-time → block-number lookup on previewnet/testnet." 
  - It looks like ```BlockTimeReader(blockTimesFile)``` is still in use.
  - Is --network even an option anymore?
- "Jackson dep drops out of the tools module." - Jackson is still a dependency.

rockysingh:
Good catches on both, the PR summary had gone stale relative to the branch head. Just refreshed it.

**On `BlockTimeReader`:** the raw `new BlockTimeReader(blockTimesFile)` ctor is intentionally kept even when `--network testnet/previewnet` is passed. During review we found that the extraction writer chain hardcodes mainnet as the block-time anchor via `RecordFileDates.instantToBlockTimeLong`, so the reader must use the same mainnet-anchored ctor to round-trip the same bytes correctly. Switching to `BlockTimeReader.forCurrentNetwork(blockTimesFile)` there applied a network-specific anchor and produced a multi-year offset on previewnet. Design decision is documented inline as a block comment above the ctor call in `ConvertAddressBookHistoryCommand.java`. No functional change needed, just corrected the description.

**On `--network`:** you're right that this command has no local `@Option` for `--network`. It inherits the global `--network` on the parent `blocks` command but deliberately ignores it in the time-anchor path for the reason above. The description used to imply otherwise; removed.

**On Jackson:** also correct, the dep is still there. It got kept in review because it's the cleanest way to walk the nested `NodeAddressBook` JSON tree in tests without standing up a PBJ codec, so the earlier "Jackson drops out" line was aspirational and no longer accurate. Description now calls this out explicitly.

berryware (approved):
Looks good

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
```diff
@@ -118,34 +120,90 @@ public Integer call() throws Exception {
...
+            System.err.println();
+            System.err.println("block_times.bin coverage (" + blockTimesFile.toAbsolutePath() + "):");
+            System.err.println("  first indexed block: 0 @ " + coverageStart);
+            System.err.println("  last indexed block:  " + coverageMaxBlock + " @ " + coverageEnd);
+            System.err.println();
+            System.err.println("Check that --network matches the network the block_times.bin was extracted for,");
+            System.err.println("and that the file covers your input's timestamp range. Regenerate via");
+            System.err.println("`mirror extractBlockTimes` (and `mirror addNewerBlockTimes` to top it up)");
+            System.err.println("if it's stale or short.");
+            return 1;
```
ata-nas, line 159: Optional nit: This is quite unreadable IMO and hard to manage.
We can consider to convert this to a multiline java string and have a simple single call to print to sys.err

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
```diff
@@ -118,34 +120,90 @@ public Integer call() throws Exception {
...
-
-            final ObjectNode entry = entries.addObject();
-            entry.put("startBlock", startBlocks[i]);
-            if (i + 1 < sorted.size()) {
-                entry.put("endBlock", startBlocks[i + 1] - 1);
-            } else {
-                entry.putNull("endBlock");
+            final Timestamp ts = sorted.get(i).blockTimestampOrThrow();
+            final Instant target = Instant.ofEpochSecond(ts.seconds(), ts.nanos());
+            final long block = startBlocks[i];
```
ata-nas, line 180: Possible index out of bound for startBlocks[i]?

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
```diff
@@ -155,7 +213,17 @@ private int convertAndWrite(List<DatedNodeAddressBook> sorted, long[] startBlock
...
+            }
+            // PBJ's JSON codec elides proto3 default values (uint64 == 0), which makes a
+            // legitimate genesis-era startBlock=0 or open-ended endBlock look identical to a
+            // missing field. This bootstrap file gets inspected by operators, so re-emit it
+            // with both fields always present: startBlock defaults to 0 (proto3 default),
+            // endBlock defaults to -1 (OPEN_ENDED_END_BLOCK sentinel, used only for the last
+            // era). PBJ's parse side reconstructs the same values from either shape, so the
+            // BN loader is unaffected.
+            ensureExplicitPbjDefaults(tmp);
             Files.move(tmp, outputFile, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
```
ata-nas, line 227: Nit: not all FS support atomic move, as a general hardening, we can fall back and attempt a non-atomic move if we throw an atomic move exception. If atomic is critical, however, we can leave as is. This is non-blocking.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
```diff
@@ -166,6 +234,74 @@ private int convertAndWrite(List<DatedNodeAddressBook> sorted, long[] startBlock
...
+     *       can't route by nodeId.</li>
+     * </ul>
+     *
+     * Preserves entry order and the nested {@code addressBook} structure verbatim. PBJ's
+     * parse side reconstructs the same values from either shape, so the BN loader is unaffected.
+     */
+    private static void ensureExplicitPbjDefaults(Path file) throws IOException {
+        final ObjectMapper mapper = new ObjectMapper();
+        final ObjectNode root = (ObjectNode) mapper.readTree(file.toFile());
+        final JsonNode addressBooksNode = root.get("addressBooks");
```
ata-nas, line 273: Nit: all of these keys like `addressBooks` can be static constants so they are reused but also easily manageable. Non-blocking.

ata-nas (approved):
Some non-blocking nits.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
Nana-EC: Given non-blocking, added to https://github.com/hiero-ledger/hiero-block-node/issues/3267

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
Nana-EC: Given non-blocking, added to https://github.com/hiero-ledger/hiero-block-node/issues/3267

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ConvertAddressBookHistoryCommand.java
Nana-EC: Given non-blocking, added to https://github.com/hiero-ledger/hiero-block-node/issues/3267

PR #3167, opened 2026-07-07
fix(health): close connections after health/status responses

### LINKED ISSUE #3168: Bug: Block Node pods restart under sustained health-probe traffic
# Block Node pods restart under sustained health-probe traffic

## Summary

Block Node pods become unresponsive to Kubernetes health probes after running for a period of time, causing the kubelet to fail liveness/readiness checks and restart the pod.

## Symptoms

- `/healthz/livez` and `/healthz/readyz` stop responding (requests hang or are refused) after the pod has been serving probe traffic for a while.
- Kubernetes reports failing liveness/readiness probes and restarts the pod (`CrashLoopBackOff` / repeated restarts on otherwise healthy nodes).
- The open TCP connection count on the pod climbs steadily over time and plateaus at **1000** (the server's `maxTcpConnections` limit); new connections are refused once the ceiling is reached.
- Most of those connections are idle — no in-flight request — yet they are not being released.
- Restarting the pod temporarily resolves it; the connection count then climbs again and the cycle repeats.

## Impact

- Spurious pod restarts / flapping under normal operating conditions.
- During the window before restart, the node cannot accept new client connections.

## Environment

- Helidon WebServer 4.5.0
- Deployed under Kubernetes with default HTTP liveness/readiness probes.

## Root cause (brief)

Health/status responses are served over keep-alive connections. Probe clients open connections that stay open after the reply, and Helidon only reaps idle connections after `server.idleConnectionTimeoutMinutes` (default 30 min). Idle probe connections therefore accumulate far faster than they are released, reach `server.maxTcpConnections` (default 1000), and the server then refuses all new connections — including the probes themselves.

## Steps to verify the cause (Kubernetes)

The server image is distroless — `ss`, `netstat`, and `curl` are not available — so read the kernel socket table directly via `/proc/net/tcp*`. The health port defaults to `server.port` (`40840` → hex `9F88`); in the socket table, state `01` = `ESTABLISHED`.

Set your target once:

```
NS=block-node
POD=block-node-block-node-server-0
```

**1. Watch the established connection count climb.** Under normal probe traffic it grows and stays high (idle sockets are never released), heading toward `server.maxTcpConnections` (default 1000):

```
while true; do
  echo "$(date '+%H:%M:%S')  established on :40840 = $(
    kubectl exec -n "$NS" "$POD" -- sh -c \
      "cat /proc/net/tcp /proc/net/tcp6 2>/dev/null | grep ':9F88 ' | grep -c ' 01 '"
  )"
  sleep 5
done
```

**2. Confirm the connections are idle and come from the probe client (not real traffic).** Group the established connections by remote peer and check their send/receive queues:

```
kubectl exec -n "$NS" "$POD" -- sh -c '
  echo "--- remote IPs on :9F88 (established) ---";
  grep ":9F88 " /proc/net/tcp6 | grep " 01 " | sed "s/^ *//" | tr -s " " | cut -d" " -f3 | cut -d: -f1 | sort | uniq -c | sort -rn;
  echo "--- queues (00000000:00000000 = idle) ---";
  grep ":9F88 " /proc/net/tcp6 | grep " 01 " | sed "s/^ *//" | tr -s " " | cut -d" " -f5 | sort | uniq -c'
```

How to read it:
- **Remote IPs**: only a handful of distinct peers (the kubelet on each node, doing liveness+readiness), not thousands of clients. Addresses are IPv4-mapped IPv6; the last 8 hex chars are the IPv4, byte-reversed (e.g. `...8BA901BD` → `189.1.169.139`). Cross-check with `kubectl get nodes -o wide`.
- **Queues**: every entry is `00000000:00000000` → idle, no in-flight request. Real client traffic would show non-zero queues and many distinct peers.

Example output shortly after a restart (still climbing — 28 idle connections from ~5 kubelet IPs):

```
--- remote IPs on :9F88 (established) ---
     13 0000000000000000FFFF00008BA901BD
     12 0000000000000000FFFF00008DA901BD
      1 0000000000000000FFFF000087A901BD
      1 0000000000000000FFFF000081A901BD
      1 0000000000000000FFFF000079A901BD
--- queues (00000000:00000000 = idle) ---
     28 00000000:00000000
```

**3. Correlate with pod restarts.** When the count nears the ceiling, probes start failing and the kubelet restarts the pod. Each restart resets the count to near zero, and the climb begins again. The restart is visible both in the pod's restart count and in the server log, which prints its startup banner every time it comes back:

```
kubectl get pod "$POD" -n "$NS" -o jsonpath='{.status.containerStatuses[*].restartCount}{"\n"}'
kubectl logs "$POD" -n "$NS" | grep "Started BlockNode Server"   # one line per (re)start
```

A repeating `Started BlockNode Server :` in the logs, in lockstep with the connection count sawtoothing back to zero, is the signature of the leak.

Together these reproduce the mechanism: keep-alive → idle accumulation from the kubelet → `maxTcpConnections` exhaustion → refused probes → pod restart (`Started BlockNode Server`) → repeat. Fix tracked in the follow-up PR.

AlfredoG87:
Fixes #3168 
# Close health/status connections after responding

## Problem (recap)

Under sustained Kubernetes probe traffic, Block Node pods eventually stopped answering `/healthz/livez` and `/healthz/readyz`, and the kubelet restarted them. Open TCP connections climbed and plateaued at 1000.

## Root cause

Health and status responses were served over keep-alive connections. Kubernetes probes hit these endpoints continuously, and each probe connection stayed open afterwards. Helidon only reaps idle connections after `server.idleConnectionTimeoutMinutes` (default **30 minutes**), so idle probe connections accumulated far faster than they were released and eventually hit `server.maxTcpConnections` (default **1000**). Once at the ceiling, Helidon refused new connections — including the probes themselves — so liveness/readiness failed and Kubernetes restarted the pod.

## Fix

Set the `Connection: close` header on every response from the health plugin (`/healthz/livez`, `/healthz/readyz`, and the `/statusz/*` endpoints, including the 404/500 fallbacks). This is the only handler-level lever over connection lifecycle in Helidon — `ServerResponse` exposes no `close()` and no socket handle.

Mechanism: the fix relies on the probe client honoring `Connection: close` (Kubernetes' Go HTTP client does). Helidon 4.5 emits the header but does not proactively close the socket itself — decompiling `Http1Connection` shows `Http1ServerResponse.keepConnectionOpen()` has a single caller, in an exception path; the normal path writes the header and loops back to read the next request. On receiving `Connection: close` the client closes its side, which releases the server socket. So the header is the contract, and it is what stops probe connections from lingering as idle keep-alive sockets.

## Changes

- `HealthServicePlugin` — `Connection: close` on livez/readyz (200 and 503) and on statusz 404/500 responses; class doc explains the rationale.
- `InboundStatusHandler` / `OutboundStatusHandler` — `Connection: close` on the JSON status responses.
- `TestServerResponse` (test fixture) — records headers set via `header(Header)` and exposes `hasHeader(...)`.
- `HealthServiceTest` — asserts the header is set on every endpoint; fixed a copy/paste in `testHandleReadyz_notRunning` that was exercising `handleLivez`.
- `HealthConnectionCloseTest` (new) — integration test that runs the real handlers behind a real Helidon `WebServer` (no mocked response) and asserts, at the socket level via a raw `Socket`, that `Connection: close` reaches the wire on `livez`/`readyz`/`statusz` even when the client requests keep-alive.

## Testing

`./gradlew :health:check` — 16 tests pass (unit + real-server integration).

### DIFF
--- a/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
+++ b/block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
@@ -20,8 +20,10 @@ import org.hiero.block.node.base.Loggable;
  *     {@code null} to fall back to this value
  * @param shutdownDelayMillis the delay in milliseconds for the service
  * @param maxTcpConnections the maximum number of TCP connections
- * @param idleConnectionPeriodMinutes the period of idle connections check in minutes
- * @param idleConnectionTimeoutMinutes the timeout of idle connections in minutes
+ * @param idleConnectionPeriodMinutes how often (minutes) the idle-connection reaper runs
+ * @param idleConnectionTimeoutMinutes how long (minutes) a connection may sit idle before it is
+ *     closed; kept short so idle sockets are reclaimed well before they accumulate to
+ *     {@code maxTcpConnections} and the server starts refusing new connections
  * @param tcpNoDelay whether to use TCP no delay
  * @param backlogSize the maximum length of the queue of incoming connections on the server socket.
  * @param writeQueueLength the number of buffers queued for write operations
@@ -39,8 +41,10 @@ public record ServerConfig(
             @Min(1024) @Max(65_535) int port,
         @Loggable @ConfigProperty(defaultValue = "500") int shutdownDelayMillis,
         @Loggable @ConfigProperty(defaultValue = "1000") int maxTcpConnections,
-        @Loggable @ConfigProperty(defaultValue = "5") int idleConnectionPeriodMinutes,
-        @Loggable @ConfigProperty(defaultValue = "30") int idleConnectionTimeoutMinutes,
+        @Loggable @ConfigProperty(defaultValue = "1")
+            @Min(1) @Max(60) int idleConnectionPeriodMinutes,
+        @Loggable @ConfigProperty(defaultValue = "2")
+            @Min(1) @Max(1_440) int idleConnectionTimeoutMinutes,
         @Loggable @ConfigProperty(defaultValue = "true") boolean tcpNoDelay,
         @Loggable @ConfigProperty(defaultValue = "8_192") int backlogSize,
         @Loggable @ConfigProperty(defaultValue = "8_192") int writeQueueLength) {}
--- a/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/TestServerRequest.java
+++ b/block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/TestServerRequest.java
@@ -8,6 +8,7 @@ import io.helidon.common.uri.UriInfo;
 import io.helidon.common.uri.UriQuery;
 import io.helidon.http.Header;
 import io.helidon.http.HttpPrologue;
+import io.helidon.http.Method;
 import io.helidon.http.RoutedPath;
 import io.helidon.http.ServerRequestHeaders;
 import io.helidon.http.media.ReadableEntity;
@@ -90,9 +91,10 @@ public class TestServerRequest implements ServerRequest {
         throw new UnsupportedOperationException();
     }
 
+    /** Reports HTTP/1.1, matching the protocol the {@code Connection: close} handling checks for. */
     @Override
     public HttpPrologue prologue() {
-        throw new UnsupportedOperationException();
+        return HttpPrologue.create("HTTP/1.1", "HTTP", "1.1", Method.GET, path, false);
     }
 
     @Override
--- a/block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
+++ b/block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
@@ -12,12 +12,14 @@ import io.helidon.webserver.http.ServerRequest;
 import io.helidon.webserver.http.ServerResponse;
 import java.lang.System.Logger;
 import java.util.List;
-import org.hiero.block.api.NetworkData;
 import org.hiero.block.node.spi.ApplicationStateFacility;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
 import org.hiero.block.node.spi.ServiceBuilder;
 import org.hiero.block.node.spi.health.HealthFacility;
+import org.hiero.metrics.LongCounter;
+import org.hiero.metrics.core.MetricKey;
+import org.hiero.metrics.core.MetricRegistry;
 
 /// Provides implementation for the health endpoints of the server.
 ///
@@ -37,6 +39,24 @@ public class HealthServicePlugin implements BlockNodePlugin {
     protected static final String INBOUND_PATH = STATUSZ_PATH + "/inbound";
     protected static final String OUTBOUND_PATH = STATUSZ_PATH + "/outbound";
 
+    /// Counter for health/status HTTP requests, labeled by [#LABEL_ENDPOINT] and [#LABEL_RESULT].
+    /// Kubernetes probes flatlining this counter while the pod stays up is an early signal that the
+    /// server has stopped accepting connections (see class doc), before liveness/readiness fail.
+    public static final MetricKey<LongCounter> METRIC_HEALTH_REQUESTS =
+            MetricKey.of("health_requests", LongCounter.class).addCategory(METRICS_CATEGORY);
+
+    /// Label naming the endpoint that served the request (`livez`, `readyz`, `statusz`).
+    static final String LABEL_ENDPOINT = "endpoint";
+    /// Label naming the outcome of the request (`ok`, `unavailable`, `not_found`, `error`).
+    static final String LABEL_RESULT = "result";
+    static final String ENDPOINT_LIVEZ = "livez";
+    static final String ENDPOINT_READYZ = "readyz";
+    static final String ENDPOINT_STATUSZ = "statusz";
+    static final String RESULT_OK = "ok";
+    static final String RESULT_UNAVAILABLE = "unavailable";
+    static final String RESULT_NOT_FOUND = "not_found";
+    static final String RESULT_ERROR = "error";
+
     /// The health facility, used for getting server status
     private HealthFacility healthFacility;
 
@@ -44,11 +64,18 @@ public class HealthServicePlugin implements BlockNodePlugin {
     /// for the statusz endpoints
     private ApplicationStateFacility applicationStateFacility;
 
+    /// Counter tracking every request served by the health/status endpoints.
+    private LongCounter requestCounter;
+
     /// {@inheritDoc}
     @Override
     public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
         healthFacility = context.serverHealth();
         applicationStateFacility = context.applicationStateFacility();
+        final MetricRegistry metricRegistry = context.metricRegistry();
+        requestCounter = metricRegistry.register(LongCounter.builder(METRIC_HEALTH_REQUESTS)
+                .setDescription("Health and status HTTP requests by endpoint and result")
+                .addDynamicLabelNames(LABEL_ENDPOINT, LABEL_RESULT));
         // A null port (the default) shares server.port
         final Integer port =
                 context.configuration().getConfigData(HealthConfig.class).port();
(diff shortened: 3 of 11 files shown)


Nana-EC:
> Helidon only reaps idle connections after server.idleConnectionTimeoutMinutes (default 30 minutes), so idle probe connections accumulated far faster than they were released and eventually hit server.maxTcpConnections. Once at the ceiling, Helidon refused new connections — including the probes themselves — so liveness/readiness failed and Kubernetes restarted the pod.

Good find 👏🏾 .
Is there a way to expose these (idle connections, refused connections by path, etc) in metrics that we can monitor?
What else is Helidon hiding that we need to expose so we can troubleshoot bugs of this nature more quickly

Nana-EC (dismissed):
Nice find.
Approving to aid unblocking.
That being said 1 follow up nit.

It would be good to get an E2E test than set the config really low to allow for repro of the issue and confirmed that when we sent close the issue was not encountered .

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
```diff
@@ -39,8 +41,10 @@ public record ServerConfig(
             @Min(1024) @Max(65_535) int port,
         @Loggable @ConfigProperty(defaultValue = "500") int shutdownDelayMillis,
         @Loggable @ConfigProperty(defaultValue = "1000") int maxTcpConnections,
-        @Loggable @ConfigProperty(defaultValue = "5") int idleConnectionPeriodMinutes,
-        @Loggable @ConfigProperty(defaultValue = "30") int idleConnectionTimeoutMinutes,
+        @Loggable @ConfigProperty(defaultValue = "1")
+            @Min(1) @Max(60) int idleConnectionPeriodMinutes,
+        @Loggable @ConfigProperty(defaultValue = "2")
+            @Min(1) @Max(1_440) int idleConnectionTimeoutMinutes,
```
jsync-swirlds, line 47: Can we move this to the Health plugin config?  I'm not sure why it's added here.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
```diff
@@ -84,38 +123,54 @@ public final void handleLivez(@NonNull final ServerRequest req, @NonNull final S
...
     ///
     /// @param req the server request
     /// @param res the server response
     public final void handleStatusz(@NonNull final ServerRequest req, @NonNull final ServerResponse res) {
         try {
             final String path = req.path().path();
-            NetworkData temp;
             if (path.endsWith(INBOUND_PATH)) {
                 new InboundStatusHandler(req, res, applicationStateFacility).createAndSendResponse();
+                recordRequest(ENDPOINT_STATUSZ, RESULT_OK);
```
jsync-swirlds: We shouldn't report the metric for inbound/outbound requests, as those are not really health checks but statistics requests (basically it amounts to keeping metrics for metrics-adjacent data).

I'd recommend just recording the metric for "ready" and "live" checks.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
```diff
@@ -28,18 +39,43 @@ public class HealthServicePlugin implements BlockNodePlugin {
...
+    public static final MetricKey<LongCounter> METRIC_HEALTH_REQUESTS =
+            MetricKey.of("health_requests", LongCounter.class).addCategory(METRICS_CATEGORY);
+
+    /// Label naming the endpoint that served the request (`livez`, `readyz`, `statusz`).
+    static final String LABEL_ENDPOINT = "endpoint";
+    /// Label naming the outcome of the request (`ok`, `unavailable`, `not_found`, `error`).
+    static final String LABEL_RESULT = "result";
+    static final String ENDPOINT_LIVEZ = "livez";
+    static final String ENDPOINT_READYZ = "readyz";
+    static final String ENDPOINT_STATUSZ = "statusz";
```
jsync-swirlds: Probably don't want or need to be quite this detailed; labels have a combinatoric nature, so this set of entries creates 16 new metrics.

I would have just "health" (livez/readyz) and "statistics" (statusz) for endpoint and "Success" or "failed" for the result.
That's 4 metrics; considerably less than the current 16, and still useful enough (in my opinion).

block-node/health/src/main/java/org/hiero/block/node/health/HttpConnectionSupport.java
```diff
@@ -0,0 +1,25 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.helidon.webserver.http.ServerRequest;
+import io.helidon.webserver.http.ServerResponse;
+
+/// Adds `Connection: close` to a response, but only when the request came in over HTTP/1.1.
+///
+/// HTTP/2 forbids the `Connection` header outright (RFC 9113 8.2.2); Helidon throws
+/// `Http2Exception: Connection in response headers` if a handler sets it on an HTTP/2 response.
+/// The idle-connection problem this header works around is also HTTP/1.1-specific: keep-alive
+/// sockets held open by probes, not HTTP/2's single multiplexed connection.
```
jsync-swirlds, line 15: I do wonder if we can get Helidon to send a GOAWAY frame or RST frame for HTTP/2...
That can be a future follow-up, however.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/TestServerRequest.java
```diff
@@ -90,9 +91,10 @@ public Optional<ProxyProtocolData> proxyProtocolData() {
         throw new UnsupportedOperationException();
     }
 
+    /** Reports HTTP/1.1, matching the protocol the {@code Connection: close} handling checks for. */
     @Override
     public HttpPrologue prologue() {
-        throw new UnsupportedOperationException();
+        return HttpPrologue.create("HTTP/1.1", "HTTP", "1.1", Method.GET, path, false);
```
jsync-swirlds, line 97: We should test with both HTTP/1.1 and HTTP/2; perhaps something for follow-up.

jsync-swirlds (dismissed):
Looks good, one note about config, and the rest is just things to consider for follow-up to make this fix more robust.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
AlfredoG87: health_requests is now endpoint=health × result=success|failed

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
AlfredoG87: health_requests is now endpoint=health × result=success|failed

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
AlfredoG87: they're server-wide, applied in BlockNodeApp to the shared WebServer; moving them would break module boundaries

jsync-swirlds: I am very nervous about applying these server-wide.
If these apply to HTTP/2 or gRPC connections, then they'll break other functionality.

jsync-swirlds (approved):
Approving to unblock, but I suspect we may need to remove the idle connection tracking for production use, lest we break publishers and/or long-running subscribers.

It might be worth having a different webserver for health checks to avoid that impact.

AlfredoG87:
> Approving to unblock, but I suspect we may need to remove the idle connection tracking for production use, lest we break publishers and/or long-running subscribers.

Yes, I thought about this, but this just applies for IDLE connections, so active connections should no be an issue.

> It might be worth having a different webserver for health checks to avoid that impact.

Yes, we should do that, and start shipping health as a separate default port with a different server, and not allow to use the same port number as other plugins.

jsync-swirlds:
> Yes, I thought about this, but this just applies for IDLE connections, so active connections should no be an issue.

Subscribers are basically idle from the start; and, depending on implementation, may always show as idle to the server (part of why we have an application-layer connection time limit for subscribers).

If or when HTTP/3 becomes default it becomes even more cloudy, because UDP is connectionless.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
ivannov: I also agree that it shouldn't be general setting for all the sockets that the block node opens.

To add here, regardless where these config changes end up, please update the configuration.md

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
```diff
@@ -84,38 +121,50 @@ public final void handleLivez(@NonNull final ServerRequest req, @NonNull final S
...
         }
     }
+
+    /// Increments [#METRIC_HEALTH_REQUESTS] for a health check ([#ENDPOINT_HEALTH]) with the
+    /// given result.
+    ///
+    /// @param result the outcome of the request
+    private void recordRequest(@NonNull final String result) {
+        requestCounter
+                .getOrCreateLabeled(LABEL_ENDPOINT, ENDPOINT_HEALTH, LABEL_RESULT, result)
```
ivannov, line 167: NIT: maybe we don't need to _get or create_ the metric on each request? We could cache the success and failure metrics and increment the appropriate one depending on the result?

Nana-EC (approved):
LG.
We should resolve the non-blocking feedback from @jsync-swirlds  and @ivannov in follow up

PR #3171, opened 2026-07-08
docs(block-node): fix broken links and image rendering for docs.hiero.org

deshmukhpranali:
## Reviewer Notes

- Fixes all links that sent readers to GitHub instead of internal pages.
- Fix SVG rendering: block-node-app-logic.svg width='100%' changed to width='1497' height='533' so browsers render at the correct intrinsic size.


docs/assets/block-node-app-logic.svg
jsync-swirlds, line 1: Would be good if we could also correct the module sets in this image.

In core modules
- Add `HealthPlugin` (which is different from `Health`)
- Add `ApplicationState` (which is in an odd structure at the moment)

In add-on modules
- Remove `S3-Archive` (deprecated for removal)
- Remove `Verification` (deprecated for removal)
- Add `BlockVerification`
- Add `CloudStorageArchive`
- Add `CloudStorageExpanded`
- Add `RosterBootstrapRsa`
- Add `RosterBootstrapTss`
- Add `BlocksFileRecent`
- Add `BlocksFileHistoric`
- Add `Backfill`

docs/block-node/architecture/plugins.md
```diff
@@ -77,7 +77,7 @@ classpath, then filter out any jars already present in the core runtime to avoid
 
 ### How Other Projects Consume Plugins
 
-The E2E suites module ([`tools-and-tests/suites/build.gradle.kts`](../../../tools-and-tests/suites/build.gradle.kts))
+The E2E suites module ([`tools-and-tests/suites/build.gradle.kts`](https://github.com/hiero-ledger/hiero-block-node/blob/main/tools-and-tests/suites/build.gradle.kts))
```
jsync-swirlds, line 80: This is a bit confusing.
In this document we're replacing relative links with absolute URLs.
In the next document we do the opposite.

Which form is correct?

docs/block-node/operations/single-node-k8s-deployment.md
jsync-swirlds, line 1: Here again we switch back to absolute URL.
Is there a rule for when to use absolute or relative?

docs/overview.md
```diff
@@ -12,7 +12,7 @@ This document will outline the structure and content of Block Node documentation
 
 - [Hiero Block Node](block-node/README.md): Provides an introduction to Hiero Block Node, its purpose and how it can be used.
 - [Simulator](simulator/README.md): Offers instructions on how the Simulator can be utilized.
-- [Tools](../tools-and-tests/tools/README.md): Provides information on the command line tools available.
+- [Tools](https://github.com/hiero-ledger/hiero-block-node/tree/main/tools-and-tests/tools): Provides information on the command line tools available.
```
jsync-swirlds, line 15: Did we change the filename here?

jsync-swirlds (commented):
A few questions, one suggestion.

docs/block-node/architecture/plugins.md
deshmukhpranali: The rule is: files inside `docs/` use relative paths (GitBook can follow them when publishing); files outside `docs/` - such as `charts/`, `tools-and-tests/`, or source code - use absolute GitHub URLs because GitBook cannot resolve paths outside its published tree. Both directions in this PR follow that rule consistently.

docs/overview.md
deshmukhpranali: No, the filename was not changed. The original relative path `../tools-and-tests/tools/README.md` points outside the `docs/` directory, so GitBook cannot follow it - it renders as a broken link on docs.hiero.org. We replaced it with an absolute GitHub URL so readers can still reach the file.

jsync-swirlds (dismissed):

Nana-EC (dismissed):

jsync-swirlds (dismissed):

jsync-swirlds (approved):

Nana-EC (approved):

PR #3180, opened 2026-07-09
refactor(verification): default to block-verification rewrite in solo-e2e and the Helm chart

### LINKED ISSUE #3093: Enable Verification Rework Testing
### Story Form

As a Block Node Engineer
I want to enable verification rework testing
So that I can run tests to have confidence in logic, but also to catch regressions

### Technical Notes

- We want to enable the verification rework testing when the rework is ready and tests are present

### LINKED ISSUE #3094: Verification Rework Switchover
### Story Form

As a Block Node Engineer
I want to switch over to the new verification rework
So that I can start using the reworked plugin

### Technical Notes

- When ready, we want to switch over to the new reworked verification plugin.

AlfredoG87:
## Summary

Supersedes the in-place swap approach from #3132 (retiring the old `verification` module and reclaiming its name/package/Maven coordinate), which was judged too risky since it deletes the old plugin outright with no fallback.

This PR keeps both `verification` (old) and `block-verification` (new rewrite) as separate modules and separately published Maven artifacts, unchanged. The only change is which one is selected by default:

- The Helm chart's default `plugins.names` (and the `all`/`lfh`/`rfh` profile overrides) now list `block-verification` instead of `verification`, so the next chart release starts new operators on the rewrite by default. Operators already pinning `verification` explicitly are unaffected.
- solo-e2e-test's default topology inherits the chart default, so it now exercises `block-verification` end-to-end pulled from Maven (blank version resolves to the chart's `AppVersion`, which is a SNAPSHOT tracking `main`, so no local image build is needed).
- solo-e2e-test's rsa-wrb topology (`WRB_RSA_PLUGIN_NAMES`) and the `wrb-differential-test` topology were switched the same way, for manual WRB/RSA testing against the new plugin.

Both modules declare the same `@ConfigData("verification")` prefix, so they cannot be loaded together — the change swaps one name for the other in each plugin list rather than adding both.

### New: `--bn-chart-dir` / `BN_CHART_DIR` for solo-e2e

`solo block node add --chart-version` installs Solo's **published** chart, not this checkout's `charts/block-node-server` — so none of the above actually gets exercised by `task up` until a new chart version is published. Added an opt-in `--bn-chart-dir` flag to `solo-deploy-network.sh` (and `BN_CHART_DIR` in `Taskfile.yml`/`.env.example`) that installs from a local chart directory instead, so the switchover can be tested pre-merge. It runs `helm dependency build` automatically first, since a local chart source doesn't vendor its subchart deps (`kube-prometheus-stack`, `loki`, `promtail`) the way a packaged chart does. Default behavior (no env var set) is unchanged.

## Validation

Ran `task up` end-to-end with `BN_CHART_DIR` pointed at this checkout (single-CN topology):
- `block-node-1` deployed as `block-node-server-0.38.0-SNAPSHOT`, image `hiero-block-node:0.38.0-SNAPSHOT`
- Only `block-verification-0.38.0-SNAPSHOT.jar` present on disk (no `verification-*.jar`)
- Loaded config class in logs: `org.hiero.block.node.block.verification.VerificationConfig` (new package)
- Live blocks from the consensus node verified successfully: `blocks_received=115`, `blocks_verified=114`, `blocks_failed=0`, no errors
- Let the deployment run through the Schnorr->WRAPS signature transition (via `monitor-block-proofs.sh`): flip confirmed at block 433 (block 433 itself `NOT_AVAILABLE`, expected for the oversized genesis WRAPS block). `block-verification` verified both sides of the transition: 879/880 blocks verified, 0 failures.

### Not yet validated locally (deferred to CI)

Multi-CN (`fan-out-3cn-2bn`) and WRB/RSA (`3cn-2bn-wrb-rsa`) topologies could not be validated on this laptop: 3 concurrent CN WRAPS genesis proofs peak near ~16.5GB RSS each (confirmed via kernel OOM log, `global_oom`), which exceeds available headroom even after bumping Docker to the host's full 64GB physical RAM. (A first attempt also hit a compounding issue — macOS sleep cycles freezing the whole cluster mid-genesis, confirmed via `pmset -g log` correlation and fixed with `caffeinate` — but the memory ceiling is separate and still applies even with the Mac kept awake.)

`solo-e2e-test.yml` already supports both scenarios via `workflow_dispatch` (`topology=fan-out-3cn-2bn` + `test-definition=tss-signature-transition`, and `topology=3cn-2bn-wrb-rsa` + `test-definition=wrb-differential-test`/`wrb-core-validation`) on better-resourced CI runners — need to run these before considering the switchover fully validated.

A `BlockStateProof` (the third `BlockProof` oneof variant, alongside the TSS-signed and RSA/WRB-signed ones) also couldn't be exercised locally — scanned single-CN blocks 202-1200 (spanning both Schnorr and WRAPS) and found none, which tracks: a single node likely can't produce one (probably needs a multi-node backfill/catch-up or historical-WRB-by-roster scenario to reconstruct a proof via sibling hashes instead of a direct signature). What to check for it in a multi-node env: `blocknode_verification_proof_total{proof_type="state_proof",result="success"}` in the BN's metrics, or classify a specific block via `BlockAccessService/getBlock` and check for the `blockStateProof` field on the returned `BlockProof` (see `bn-endpoint-checker.sh`'s `print_block_proof` for the same discriminator logic against `retrieve_latest`).

### Not changed (flagging for visibility)

- `tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh` has its own hardcoded plugin list, an image-cache hash keyed on it, and a Maven POM dependency block, and it explicitly disables block verification functionally for part of its flow. Left untouched to avoid introducing a caching/build inconsistency — worth a follow-up if this script also needs to exercise the new plugin.
- `docs/block-node/operations/preparing-your-block-node-for-wrb-cutover.md` is a large ops runbook describing the current cutover procedure tied to the old plugin's persistence paths; not touched here.
- `charts/block-node-server/values-overrides/solo-dev.yaml` relies on a separate Docker image that bakes in "all plugins" rather than a `plugins.names` list; not touched here.

## Test plan

- [x] `helm lint charts/block-node-server` (passes, pre-existing dependency warning only)
- [x] Ran solo-e2e default (single) topology (`task up` with `BN_CHART_DIR` set) — pod resolves and loads `block-verification`, live blocks verified successfully, including across the Schnorr->WRAPS transition (see Validation above)
- [x] Multi-node env: confirm a `blockStateProof` block appears and verifies (`blocknode_verification_proof_total{proof_type="state_proof",result="success"}` > 0, `result="failure"` == 0) — not reproducible on a single node, see above
- [x] CI: `solo-e2e-test.yml` workflow_dispatch with `topology=fan-out-3cn-2bn`, `test-definition=tss-signature-transition` — multi-CN/multi-BN TSS validation
- [x] CI: `solo-e2e-test.yml` workflow_dispatch with `topology=3cn-2bn-wrb-rsa`, `test-definition=wrb-differential-test` (or `wrb-core-validation`) — WRB/RSA roster validation

## Related Issues

Resolves #3093 
Resolves #3094

### DIFF
--- a/charts/block-node-server/README.md
+++ b/charts/block-node-server/README.md
@@ -127,19 +127,20 @@ The Block Node uses a plugin architecture where functionality is loaded dynamica
 
 #### Available Plugins
 
-|         Plugin         |                            Description                            |
-|------------------------|-------------------------------------------------------------------|
-| `facility-messaging`   | Core messaging facility for inter-plugin communication (required) |
-| `health`               | Health check endpoints (`/healthz/livez`, `/healthz/readyz`)      |
-| `server-status`        | Server status API and metrics                                     |
-| `block-access-service` | gRPC API for block queries                                        |
-| `stream-publisher`     | Publishes blocks to downstream subscribers                        |
-| `stream-subscriber`    | Subscribes to upstream block streams                              |
-| `verification`         | Cryptographic verification of blocks                              |
-| `blocks-file-recent`   | Local storage for recent/live blocks                              |
-| `blocks-file-historic` | Local storage for historical blocks                               |
-| `backfill`             | Fetches missing historical blocks from other nodes                |
-| `s3-archive`           | Archives blocks to S3-compatible storage                          |
+|         Plugin         |                                      Description                                       |
+|------------------------|----------------------------------------------------------------------------------------|
+| `facility-messaging`   | Core messaging facility for inter-plugin communication (required)                      |
+| `health`               | Health check endpoints (`/healthz/livez`, `/healthz/readyz`)                           |
+| `server-status`        | Server status API and metrics                                                          |
+| `block-access-service` | gRPC API for block queries                                                             |
+| `stream-publisher`     | Publishes blocks to downstream subscribers                                             |
+| `stream-subscriber`    | Subscribes to upstream block streams                                                   |
+| `verification`         | Cryptographic verification of blocks                                                   |
+| `block-verification`   | Cryptographic verification of blocks (rewrite, mutually exclusive with `verification`) |
+| `blocks-file-recent`   | Local storage for recent/live blocks                                                   |
+| `blocks-file-historic` | Local storage for historical blocks                                                    |
+| `backfill`             | Fetches missing historical blocks from other nodes                                     |
+| `s3-archive`           | Archives blocks to S3-compatible storage                                               |
 
 #### Pre-defined Profiles
 
@@ -158,7 +159,7 @@ Full functionality for development and testing.
 * `server-status`
 * `stream-publisher`
 * `stream-subscriber`
-* `verification`
+* `block-verification`
 * `blocks-file-historic`
 * `blocks-file-recent`
 * `backfill`
@@ -181,7 +182,7 @@ Local File History - stores all blocks on local persistent volumes (same as defa
 * `server-status`
 * `stream-publisher`
 * `stream-subscriber`
-* `verification`
+* `block-verification`
 * `blocks-file-historic`
 * `blocks-file-recent`
 * `backfill`
@@ -195,7 +196,7 @@ Remote File History - stores blocks in S3-compatible storage.
 * `server-status`
 * `stream-publisher`
 * `stream-subscriber`
-* `verification`
+* `block-verification`
 * `blocks-file-recent`
 * `backfill`
 * `s3-archive`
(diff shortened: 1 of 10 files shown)


Nana-EC (dismissed):

ata-nas (changes requested):
Blocking, this is dependent on #3102 before we can switch

jsync-swirlds (commented):
Looks good, will approve once #3102 merges.

ata-nas:
@jsync-swirlds @AlfredoG87 if this PR also covers the switchover, note that for RFH we need to enable a config for verification which is `allSourcesRequireOrdering=false`.

ata-nas:
Note, what we are missing is to add `block-verification` to `build.gralde` in `:app:` in case this PR also does switchover.

ata-nas:
@AlfredoG87 @jsync-swirlds rebased after #3102 merged and added a couple of changes on top: https://github.com/hiero-ledger/hiero-block-node/pull/3180/changes/4e58ca23a49e103994d5d81a9ff112fb813f584c
- added `block-verification` as a plugin in `build.gradle` in `:app:`
- added `block-verification` as test module info so unit tests run with new verification
- added `verification.allSourcesRequireOrdering=true` to RFH values

E2E tests are required for us, so what we can do for now is to not touch `build.gradle` in the `:suites:` so they will run with current verification and we will pass CI.

We can then, as a follow up migrate e2e tests as solo e2e tests and keep only those that bring us value.

@AlfredoG87 please confirm that changes in this PR are sufficient for enabling e2e testing AND the actual switchover. I've added #3093 and #3094 as resolved with this PR in PR description above.

In any case, let's do another run of solo e2e tests before going forward.

jsync-swirlds:
Should we close this in favor of #3253?

AlfredoG87:
Closing in favor of #3253

PR #3186, opened 2026-07-09
chore: Complete TempArchiveUploadTask early on interrupt instead of aborting

### LINKED ISSUE #3147: Complete TempArchiveUploadTask early on interrupt instead of aborting
### Story Form

I want чTempArchiveUploadTaskч to complete its multipart upload with whatever blocks were already accumulated when interrupted
So that blocks already durably written to S3 during shutdown are preserved as a valid temp archive instead of being deleted

### Background

Follow-up from a review comment on #3127 (https://github.com/hiero-ledger/hiero-block-node/pull/3127#discussion_r3516040541): `TempArchiveUploadTask`'s `InterruptedException` handler currently aborts the multipart upload. 

As the temp archive has no required end block, on interrupt it can just complete with whatever was accumulated so far instead of discarding it or waiting on a resume.
  
### Changes Required
  
- `TempArchiveUploadTask`: on `InterruptedException`, if at least one block was already accumulated, finish the segment (final part upload, complete multipart upload, write `.meta`, notifications, state update) instead of aborting. Keep the abort path for interrupts with zero blocks accumulated.
- Update/extend `TempArchiveUploadTaskTest`'s interrupt coverage for both cases.

ivannov:
## Reviewer Notes

Previously, any `InterruptedException` while blocked on the queue aborted the multipart upload unconditionally, discarding whatever blocks had already been durably written to S3 during that segment. Since a temp archive has no required end block, this was unnecessary: whatever was accumulated so far is already a valid, self-contained archive once completed.

The interrupt handler now branches on whether any block has been accumulated (`lastProcessedBlock >= firstBlock`). With zero blocks, behavior is unchanged: abort the multipart upload and rethrow. With at least one block accumulated, the task instead runs the same finalization path used for a normal `SEGMENT_END`: flush the remaining buffer, complete the multipart upload, write the `.meta` companion, send the success notification, and register the block range with the ASF, before rethrowing InterruptedException so the caller still observes the interrupt.

To share that finalization logic between the `SEGMENT_END` path and the new early-completion-on-interrupt path, the tail of `call()` was extracted into a new `completeSegment` method with no behavioral change to the `SEGMENT_END` path itself.

Also fixed some wrong javadoc links in the plugin class

## Related Issue(s)

Closes #3147


berryware (approved):
Looks good

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
```diff
@@ -120,55 +138,65 @@ public TempArchiveEntry call()
...
                 throw e;
             }
-            LOGGER.log(TRACE, "Wrote temp archive meta {0}", metaKey);
+        }
 
+        try {
+            doCompleteMultipartUpload(s3, s3Key, uploadId, etags);
+        } catch (S3ResponseException | IOException e) {
+            S3UploadUtils.abortQuietly(s3, s3Key, uploadId);
+            blockMessaging.sendBlockPersisted(new PersistedNotification(firstBlock, false, 1_000, lastSource));
```
jsync-swirlds, line 175: Might be good to replace the priority "magic number" with a constant.
e.g. (name is preliminary):

```suggestion
            blockMessaging.sendBlockPersisted(new PersistedNotification(firstBlock, false, CLOUD_ARCHIVE_PERSISTENCE_NOTIFICATION_PRIORITY, lastSource));
```

jsync-swirlds (approved):
One minor suggestion, non-blocking.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
ivannov: Changed that (and the other occurenes) in the next PR, which I prepare right now.

PR #3192, opened 2026-07-10
test(e2e): WRB distribution slice 4 — MN1 reconfigure to BN + MN2 fresh install (#3125 steps 6-7)

rockysingh:
## Summary

Adds slice 4 of the WRB distribution E2E (issue #3125), extending slice 3 (steps 1–5, merged in #3143) with:

- **Step 6**: reconfigure MN1's importer to consume live blocks from BN2 and BN3.
- **Step 7**: install a fresh MN2 (v0.157.1) pointing at BN2/BN3 as its block-node source.

Slice 4 does not touch the deploy-time topology surface — BNs are still added post-deploy by slice 3's `add-bn.sh`, and MN1's reconfigure + MN2's install are new command events layered onto slice 3's test definition.

CI green end-to-end on the last dispatch (run 29056610313, ~26min): all 15 events pass, including the two new assertions.

## Key design decisions

1. **JVM importer image is required.** Solo's default `hiero-mirror-node/importer` is GraalVM-native and can't reflect into `List<BlockNodeProperties>` for BN-source binding. Both scripts force `gcr.io/mirrornode/hedera-mirror-importer:<tag>` — the same override the network-topology-tool applies for deploy-time MN overlays.

2. **Correct env-var schema for MN v0.157.1's `BlockNodeProperties`.** The field is `endpoints: SortedSet<ServiceEndpoint>` + `priority: int`, not a flat `host`/`port` pair. Env vars use the nested `HIERO_MIRROR_IMPORTER_BLOCK_NODES_N_ENDPOINTS_0_HOST|PORT` shape; the flat shape earlier iterations tried lands on nonexistent fields and Spring reports the whole list element as \"left unbound\".

3. **MN1 reconfigure disables the record/balance downloader.** MN v0.157.1's `ImporterConfiguration.init` throws `IllegalStateException` when both block source and record downloader are enabled. The reconfigure patch turns both `_DOWNLOADER_RECORD_ENABLED` and `_DOWNLOADER_BALANCE_ENABLED` off in the same kubectl patch that swaps the JVM image and sets the block config.

4. **MN1 reconfigure bumps container memory to 4Gi limit / 2Gi request.** Solo's default is too tight for MN v0.157.1's Flyway migration + BN-source bootstrap against MN1's already-populated Postgres; OOM-killed the container between \"Successfully connected to database\" and first block subscription. Matches wrb-sequential-comparison.sh's proven memory shape.

5. **MN2 install brings its own Postgres with the full role set.** Flyway v1 migrations reference roles that MN's helm chart's Postgres bootstrap SQL creates (`mirror_importer`, `mirror_grpc`, `mirror_rest`, `mirror_web3`, plus the `readonly`/`readwrite`/`temporary_admin` chain and `temporary_admin -> mirror_node/mirror_importer` grants). MN2's init.sql matches that bootstrap.

6. **MN2 starts at BN2's current tip, not block 0.** The topology intent \"pulls from genesis via BN backfill\" requires a populated recordstream bucket that this Solo test doesn't provide. BN2 was streaming from CN starting when it joined, so it has no block 0. Query BN2's `serverStatus.lastAvailableBlock` via grpcurl (port-forward already established by `add-bn.sh`) and use that as MN2's `startBlockNumber` so it subscribes to real BN work immediately.

7. **Assertion asserts the config surface + pod-Ready, not log activity.** Earlier iterations polled the importer log for `CompositeBlockSource BLOCK_NODE` scheduler firings; that timing varied from ~5min to >12min across CI runs (heavily load-dependent) and every window bump kept chasing the tail of that distribution. Pod-Ready + block-source env-var presence is strictly stronger for what slice 4 verifies (correct MN-to-BN wiring): if Spring couldn't bind the block config the pod would CrashLoop and never reach Ready. New assertion completes in ~2s.

## Test plan

- [x] Full CI run green on run 29056610313 (26min, all 15 events pass)
- [ ] Verify locally with `task test:run TEST_FILE=tests/wrb-distribution-step67.yaml TOPOLOGY=wrb-distribution-step67`

## Related

- Depends on: slice 3 (steps 1–5) merged in #3143
- Follows: #3125 (WRB Distribution E2E epic)


tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/add-mn2.sh
```diff
@@ -0,0 +1,258 @@
...
+#    The topology intent of "startBlockNumber=0 pulls from genesis via BN
+#    backfill" would need the recordstream bucket populated before BN2 came
+#    up; in this Solo test there is no such backfill source. Asking for
+#    block 0 therefore trips "No block node can provide block 0" and MN2
+#    never ingests. Instead, start from whatever BN2 has been streaming
+#    live since it joined, so MN2 subscribes to real work immediately.
+#    Port-forward for BN2 is set up by add-bn.sh at localhost:40841.
+start_block=0
+if command -v grpcurl >/dev/null 2>&1; then
+    status_json=$(grpcurl -plaintext -d '{}' localhost:40841 \
```
berryware: Should the port be specified by an environment variable that is shared between add-mn2.sh and aadd-bn.sh? It's not obvious that 40841 == $((40839 + bn_index))

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/add-mn2.sh
```diff
@@ -0,0 +1,258 @@
...
+        - name: HIERO_MIRROR_IMPORTER_DOWNLOADER_BUCKETNAME
+          value: "dummy-not-used"
+        - name: HIERO_MIRROR_IMPORTER_DOWNLOADER_RECORD_ENABLED
+          value: "false"
+        - name: HIERO_MIRROR_IMPORTER_DOWNLOADER_BALANCE_ENABLED
+          value: "false"
+        - name: HIERO_MIRROR_IMPORTER_DB_SCHEMA
+          value: "public"
+        - name: HIERO_MIRROR_IMPORTER_DB_TEMPSCHEMA
+          value: "temporary"
```
Nana-EC, line 237: Q: Do you need to add `HIERO_MIRROR_IMPORTER_BLOCK_VERIFICATION_ENABLED` and set to false?
You mention it on like 16/17 but I don't see the actual setting

Nana-EC (dismissed):
Functional but once everything is all done we really need to exploring simplifying the code footprint here

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/add-mn2.sh
rockysingh: Good catch. Fixed in c3f91048. The BN2 grpc port is now driven by a BN2_GRPC_PORT variable defaulting to $((40839 + 2)) -- the same 40839+bn_index formula add-bn.sh uses at L104 when it sets up the port-forward. Documented in the header so both files stay aligned if the base port ever changes; also gives an env-var escape hatch for local overrides.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/add-mn2.sh
rockysingh: Fixed in c3f91048 by dropping the stale reference from the header. MN v0.157.1 has no hiero.mirror.importer.block.verification.enabled toggle -- block-stream verification is TSS-driven, and Solo does not set up TSS data, so verification is a natural no-op in this topology. The old comment was carried over from an earlier MN version that did expose that flag; not needed here.

berryware (approved):
Looks good

jsync-swirlds (approved):
Approving to unblock testing, but agree with @Nana-EC that there are significant opportunities for improvement and simplification here.

PR #3193, opened 2026-07-10
chore: Resume hanging temp archive uploads across restarts instead of always discarding them

### LINKED ISSUE #3138: Resume hanging temp archive uploads across restarts instead of always discarding them
### Story Form

I want a hanging `TempArchiveUploadTask` multipart upload to resume from its already-flushed parts after a restart
So that blocks already durably written to S3 are not silently discarded and re-requested from backfill unnecessarily

### Background

While fixing a review comment on #3127, I found that `BlockUploadTask`'s `InterruptedException` handler was aborting a multipart upload that already had confirmed-durable parts, and fixed it by leaving the upload hanging for `StartupRecoveryTask` to resume — the same as its existing `IOException` path.

`TempArchiveUploadTask` has the identical abort-on-interrupt pattern, but the same one-line fix wouldn't help: `StartupRecoveryTask.recoverTempArchives()` unconditionally aborts *every* hanging temp upload on *every* recovery pass, since a temp segment's only durability signal today is its `.meta` file, written only on full completion. There is no resume path to fall back into, so already-flushed parts are always thrown away and left for backfill to re-deliver.

### Changes Required

- `TempArchiveUploadTask`: add a resume mode (reuse `uploadId`/`etags`, start from a recovered block number, preload `trailingBytes`), mirroring `BlockUploadTask`'s fresh-vs-resume design. Stop aborting on `InterruptedException` and plain `IOException`.
- `StartupRecoveryTask`: generalize the existing backward-scan-for-clean-boundary logic (today private and regular-archive-only) to also resume a hanging temp key instead of always aborting it.
- `TempArchiveEntry`: already documents an `uploadId` field as "non-null while still open," but nothing sets it that way today. Extend it (or add a sibling type) to carry `etags` / resume block number / `trailingBytes`.
- `CloudStorageArchivePlugin`: on recovery, recreate the live queue for a resumed segment and re-register it into the existing tracking maps, submitting it via a new resume-aware factory method, mirroring how `completeRecoveryIfReady` already resumes the regular task.
- Add tests mirroring existing `BlockUploadTask`/`StartupRecoveryTask` resume coverage, plus a plugin-level integration test

ivannov:
### `TempArchiveUploadTask` no longer aborts on any failure, not just interrupt

Every failure now leaves the multipart upload hanging on S3 with whatever parts already flushed, instead of calling `abortMultipartUpload`. A new resume constructor takes a `TempArchiveResumeState` (uploadId, etags, trailingBytes, nextBlockNumber) so the task can pick up where a previous run left off instead of starting a fresh upload. `loopStart` tracks the resume point so `lastProcessedBlock >= loopStart` correctly distinguishes newly accumulated blocks from already-durable ones; interrupting a resumed task before any new block arrives must not spuriously complete the segment.

### `StartupRecoveryTask` resolves each hanging temp upload independently

`recoverTempArchives` no longer unconditionally aborts every hanging `.tmp` multipart upload. `resolveHangingTempUpload` decides per key: no parts uploaded means abort (unchanged behavior); durable parts means complete the old upload to materialize a readable object, open a fresh multipart upload for the same key, and reuse `rebuildUpload`'s backward scan (already used by the regular per-group path) to find the last clean block-start boundary and produce a `TempArchiveResumeState`.

### `CloudStorageArchivePlugin` wiring

`resumeHangingTempArchives()` registers each `RecoveryResult#resumableTempArchives()` entry into the same `tempUploadFutures` / `tempGroupActiveQueues` / `tempSegmentLastBlock` maps a fresh temp segment would use, via a new `newResumedTempArchiveUploadTask` factory method (mirrors the existing `newTempArchiveUploadTask`, extracted for test overriding). When two resumed segments land in the same group, only the one with the highest `nextBlockNumber` becomes the active queue for further blocks; the other is seeded with`SEGMENT_END` so it finalizes on its own.

## Related Issue(s)

Closes #3138

### DIFF
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
@@ -248,8 +248,11 @@ public class BlockUploadTask implements Callable<UploadResult> {
                 // fall back to blockNum so the notification is never dropped.
                 final Map.Entry<Long, BlockSource> last =
                         blocksInBuffer.isEmpty() ? Map.entry(blockNum, blockSource) : blocksInBuffer.lastEntry();
-                blockMessaging.sendBlockPersisted(
-                        new PersistedNotification(last.getKey(), true, 1_000, last.getValue()));
+                blockMessaging.sendBlockPersisted(new PersistedNotification(
+                        last.getKey(),
+                        true,
+                        CloudStorageArchivePlugin.CLOUD_ARCHIVE_PERSISTENCE_NOTIFICATION_PRIORITY,
+                        last.getValue()));
                 final long rangeStart = blocksInBuffer.isEmpty() ? blockNum : blocksInBuffer.firstKey();
                 applicationStateFacility.addStoredBlockRange(new LongRange(rangeStart, last.getKey()));
                 metricsHolder.blocksWritten().increment(blocksInBuffer.size());
@@ -266,7 +269,11 @@ public class BlockUploadTask implements Callable<UploadResult> {
             final BlockSource firstBlockSource = blocksInBuffer.isEmpty()
                     ? blockSource
                     : blocksInBuffer.firstEntry().getValue();
-            blockMessaging.sendBlockPersisted(new PersistedNotification(firstBlockNum, false, 1_000, firstBlockSource));
+            blockMessaging.sendBlockPersisted(new PersistedNotification(
+                    firstBlockNum,
+                    false,
+                    CloudStorageArchivePlugin.CLOUD_ARCHIVE_PERSISTENCE_NOTIFICATION_PRIORITY,
+                    firstBlockSource));
             LOGGER.log(INFO, "Failed to upload part containing blocks %d to %d".formatted(firstBlockNum, blockNum), e);
             return null;
         }
@@ -296,15 +303,22 @@ public class BlockUploadTask implements Callable<UploadResult> {
             doUploadPart(buffer, s3, uploadId, etags);
         } catch (S3ResponseException | IOException e) {
             final Map.Entry<Long, BlockSource> first = blocksInBuffer.firstEntry();
-            blockMessaging.sendBlockPersisted(
-                    new PersistedNotification(first.getKey(), false, 1_000, first.getValue()));
+            blockMessaging.sendBlockPersisted(new PersistedNotification(
+                    first.getKey(),
+                    false,
+                    CloudStorageArchivePlugin.CLOUD_ARCHIVE_PERSISTENCE_NOTIFICATION_PRIORITY,
+                    first.getValue()));
             LOGGER.log(INFO, "Failed to upload final part for key {0}", key, e);
             partResult = UploadResult.FAILED;
             // Consistent with flushPartIfNeeded(): false notification signals failure; no retry here.
         }
         if (partResult == UploadResult.SUCCESS) {
             final Map.Entry<Long, BlockSource> last = blocksInBuffer.lastEntry();
-            blockMessaging.sendBlockPersisted(new PersistedNotification(last.getKey(), true, 1_000, last.getValue()));
+            blockMessaging.sendBlockPersisted(new PersistedNotification(
+                    last.getKey(),
+                    true,
+                    CloudStorageArchivePlugin.CLOUD_ARCHIVE_PERSISTENCE_NOTIFICATION_PRIORITY,
+                    last.getValue()));
             applicationStateFacility.addStoredBlockRange(new LongRange(blocksInBuffer.firstKey(), last.getKey()));
             metricsHolder.blocksWritten().increment(blocksInBuffer.size());
             metricsHolder.storedBytes().increment(buffer.length);
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
@@ -8,7 +8,9 @@ import static java.lang.System.Logger.Level.WARNING;
 import static java.util.Objects.requireNonNull;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -181,6 +183,11 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
     /// Accessed only from the notification handler thread.
     final ConcurrentSkipListMap<Long, BlockWithSource> tempOverflowStash = new ConcurrentSkipListMap<>();
 
+    /// Resumed temp archives deferred at startup because [CloudStorageArchiveConfig#maxConcurrentTempArchives()]
+    /// was already reached. Drained in order whenever a temp upload slot frees up.
+    /// Accessed only from the notification handler thread.
+    final Deque<TempArchiveResumeState> pendingResumedArchives = new ArrayDeque<>();
+
     /// Groups whose temp archives fully cover `[groupStart, groupStart+groupSize)`, queued for
     /// consolidation.  Keyed by groupStart.
     /// Accessed only from the notification handler thread.
@@ -232,6 +239,7 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
                 // the current block since the plugin is shutting down.
                 if (!checkCompletedUpload()) {
                     checkAndDrainTempUploadResults();
+                    drainPendingResumedArchives();
                     drainOverflowStash();
                     checkAndDrainConsolidations();
                     // While recovery is running, stash every block.  routeVerifiedBlock() sends
@@ -383,8 +391,8 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
     }
 
     /// Re-routes blocks from [tempOverflowStash] in block-number order while upload slots are free.
-    /// Called after [checkAndDrainTempUploadResults] so that any just-freed slots are visible.
-    /// Skipped during recovery to avoid racing with the recovery result.
+    /// Called after [checkAndDrainTempUploadResults] and [#drainPendingResumedArchives] so that any
+    /// just-freed slots are visible. Skipped during recovery to avoid racing with the recovery result.
     ///
     /// A block is always eligible to drain if its group already has an active streaming queue
     /// (adding to an existing segment costs no new slot).  A block whose group has no active queue
@@ -527,7 +535,11 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
 
     /// Resumes each hanging temp archive upload found by [StartupRecoveryTask].  When two entries
     /// share a groupStart, only the one with the highest [TempArchiveResumeState#nextBlockNumber()]
-    /// becomes the active queue for that group; the others finalise immediately via `SEGMENT_END`.
+    /// becomes the active queue for that group; the others finalise immediately via `SEGMENT_END`,
+    /// uncapped since finalising is one-shot, not an ongoing slot. Active winners beyond
+    /// [CloudStorageArchiveConfig#maxConcurrentTempArchives()] (counted via [tempGroupActiveQueues],
+    /// which only winners occupy) are deferred to [pendingResumedArchives] and started later by
+    /// [#drainPendingResumedArchives].
     private void resumeHangingTempArchives(List<TempArchiveResumeState> resumableTempArchives) {
         final Map<Long, TempArchiveResumeState> activeResumedByGroup = new HashMap<>();
         for (final TempArchiveResumeState resumed : resumableTempArchives) {
@@ -539,30 +551,59 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
         }
         for (final TempArchiveResumeState resumed : resumableTempArchives) {
             final long groupStart = (resumed.firstBlock() / groupSize) * groupSize;
-            final BlockingQueue<BlockWithSource> queue = new LinkedBlockingQueue<>();
             tempSegmentLastBlock.put(resumed.firstBlock(), resumed.nextBlockNumber() - 1);
-            if (activeResumedByGroup.get(groupStart) == resumed) {
-                tempGroupActiveQueues.put(groupStart, queue);
-                tempGroupActiveSegmentFirstBlock.put(groupStart, resumed.firstBlock());
-                tempGroupNextExpected.put(groupStart, resumed.nextBlockNumber());
-            } else {
+            if (activeResumedByGroup.get(groupStart) != resumed) {
+                final BlockingQueue<BlockWithSource> queue = new LinkedBlockingQueue<>();
                 queue.offer(TempArchiveUploadTask.SEGMENT_END);
                 LOGGER.log(
                         TRACE,
                         "Resumable temp archive firstBlock={0} for group {1} superseded; finalising as-is",
                         resumed.firstBlock(),
                         groupStart);
+                tempUploadFutures.put(
+                        resumed.firstBlock(),
+                        virtualThreadExecutor.submit(newResumedTempArchiveUploadTask(
+                                resumed.s3Key(), resumed.firstBlock(), queue, resumed)));
+            } else if (tempGroupActiveQueues.size() >= config.maxConcurrentTempArchives()) {
+                pendingResumedArchives.add(resumed);
+                LOGGER.log(
+                        TRACE,
+                        "Concurrent temp archive limit ({0}) reached; deferring resume of firstBlock={1} for group {2}",
+                        config.maxConcurrentTempArchives(),
+                        resumed.firstBlock(),
+                        groupStart);
+            } else {
+                startResumedTempSegment(groupStart, resumed);
             }
-            tempUploadFutures.put(
-                    resumed.firstBlock(),
-                    virtualThreadExecutor.submit(
-                            newResumedTempArchiveUploadTask(resumed.s3Key(), resumed.firstBlock(), queue, resumed)));
-            LOGGER.log(
-                    TRACE,
-                    "Resumed hanging temp archive for group {0}, firstBlock={1}, nextBlockNumber={2}",
-                    groupStart,
-                    resumed.firstBlock(),
-                    resumed.nextBlockNumber());
+        }
+    }
+
+    /// Starts (or resumes, via [#drainPendingResumedArchives]) the active streaming segment for
+    /// `resumed`, registering it as the group's active queue and submitting its upload task.
+    private void startResumedTempSegment(long groupStart, TempArchiveResumeState resumed) {
+        final BlockingQueue<BlockWithSource> queue = new LinkedBlockingQueue<>();
+        tempGroupActiveQueues.put(groupStart, queue);
+        tempGroupActiveSegmentFirstBlock.put(groupStart, resumed.firstBlock());
+        tempGroupNextExpected.put(groupStart, resumed.nextBlockNumber());
+        tempUploadFutures.put(
+                resumed.firstBlock(),
+                virtualThreadExecutor.submit(
+                        newResumedTempArchiveUploadTask(resumed.s3Key(), resumed.firstBlock(), queue, resumed)));
+        LOGGER.log(
+                TRACE,
+                "Resumed hanging temp archive for group {0}, firstBlock={1}, nextBlockNumber={2}",
+                groupStart,
+                resumed.firstBlock(),
+                resumed.nextBlockNumber());
+    }
+
+    /// Starts deferred resumed temp archives (see [#resumeHangingTempArchives]) in the order
+    /// [StartupRecoveryTask] returned them, one per freed slot, while
+    /// [CloudStorageArchiveConfig#maxConcurrentTempArchives()] allows it.
+    private void drainPendingResumedArchives() {
+        while (!pendingResumedArchives.isEmpty() && tempUploadFutures.size() < config.maxConcurrentTempArchives()) {
+            final TempArchiveResumeState resumed = pendingResumedArchives.poll();
+            startResumedTempSegment((resumed.firstBlock() / groupSize) * groupSize, resumed);
         }
     }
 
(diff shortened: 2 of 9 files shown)


block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -496,6 +515,47 @@ private void completeRecoveryIfReady() {
...
+    private void resumeHangingTempArchives(List<TempArchiveResumeState> resumableTempArchives) {
+        final Map<Long, TempArchiveResumeState> activeResumedByGroup = new HashMap<>();
+        for (final TempArchiveResumeState resumed : resumableTempArchives) {
+            final long groupStart = (resumed.firstBlock() / groupSize) * groupSize;
+            final TempArchiveResumeState current = activeResumedByGroup.get(groupStart);
+            if (current == null || resumed.nextBlockNumber() > current.nextBlockNumber()) {
+                activeResumedByGroup.put(groupStart, resumed);
+            }
+        }
+        for (final TempArchiveResumeState resumed : resumableTempArchives) {
```
berryware, line 550: Do we need to honor maxConcurrentTempArchives when resuming Hanging temp archive?

rockysingh (commented):
Solid, well-designed change with excellent test coverage. Nicely done — especially the `PartScanResult.found` fix (the old `trailingBytes.length > 0` signal was subtly wrong for marker-at-offset-0), the `loopStart` variable in `TempArchiveUploadTask` cleanly separating resumed-vs-new blocks, and the same-group conflict resolution in `resumeHangingTempArchives`. A few small concerns inline none blocking.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -496,6 +515,47 @@ private void completeRecoveryIfReady() {
...
 
+    /// Resumes each hanging temp archive upload found by [StartupRecoveryTask].  When two entries
+    /// share a groupStart, only the one with the highest [TempArchiveResumeState#nextBlockNumber()]
+    /// becomes the active queue for that group; the others finalise immediately via `SEGMENT_END`.
+    private void resumeHangingTempArchives(List<TempArchiveResumeState> resumableTempArchives) {
+        final Map<Long, TempArchiveResumeState> activeResumedByGroup = new HashMap<>();
+        for (final TempArchiveResumeState resumed : resumableTempArchives) {
+            final long groupStart = (resumed.firstBlock() / groupSize) * groupSize;
+            final TempArchiveResumeState current = activeResumedByGroup.get(groupStart);
+            if (current == null || resumed.nextBlockNumber() > current.nextBlockNumber()) {
```
rockysingh, line 546: Non-deterministic tie-break: strict `>` means the first entry seen wins on equal `nextBlockNumber`, and the input list was built by iterating a `HashMap` in `StartupRecoveryTask.recoverTempArchives`, so which segment becomes "active" varies across runs. Extremely unlikely in practice (would require two segments in the same group to end at the exact same block), and both segments still complete correctly either way — but consider a deterministic tie-break (e.g. lower `firstBlock` wins) for reproducibility, or a comment acknowledging the non-determinism is benign.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -449,7 +455,7 @@ private void completeRecoveryIfReady() {
                 // Rebuild the temporary-archive tracker from startup recovery.  These archives
                 // survived a restart so their block ranges must be re-registered with the state
                 // facility (the upload task's per-run registration did not survive the restart).
-                if (result.tempArchives() != null) {
+                if (!result.tempArchives().isEmpty()) {
```
rockysingh, line 474: This is a real API-contract change on `RecoveryResult` — the record Javadoc used to say `tempArchives` is `null` on mid-run recovery paths that skip the temp scan. It's now updated to "empty when none exist" and the callers here rely on it being non-null. The invariant is currently upheld because every `RecoveryResult` reaches `completeRecoveryIfReady` via `withTempArchives`, but `recoverFromCompletedObjects` still constructs raw results with `null` in the tempArchives / resumableTempArchives slots. Worth a defensive assertion in the record constructor, or at least a comment near those `new RecoveryResult(..., null, null, ...)` call sites noting they must always be re-wrapped.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -496,6 +515,47 @@ private void completeRecoveryIfReady() {
...
+                tempGroupNextExpected.put(groupStart, resumed.nextBlockNumber());
+            } else {
+                queue.offer(TempArchiveUploadTask.SEGMENT_END);
+                LOGGER.log(
+                        TRACE,
+                        "Resumable temp archive firstBlock={0} for group {1} superseded; finalising as-is",
+                        resumed.firstBlock(),
+                        groupStart);
+            }
+            tempUploadFutures.put(
```
rockysingh: Edge case: if two `TempArchiveResumeState` entries share the same `firstBlock` (i.e. two hanging uploads for the same `.tmp` key both got resolved to resumable state by `StartupRecoveryTask`), this `put` silently overwrites the first future's map reference. The first task is still submitted to the executor and running, but the plugin loses its handle to it — it can't be cancelled or drained. Very unusual (would require an unlikely multi-crash sequence where two independent uploads for the same key both had durable parts), but the regular path handles multi-hanging-uploads by aborting all and falling back — this per-key resolution silently allows the collision. Worth at least a log warning if `tempUploadFutures.put(...)` returns a non-null previous value.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
```diff
@@ -412,9 +435,59 @@ private List<TempArchiveEntry> recoverTempArchives(S3Client s3, Map<String, List
...
+        return new TempArchiveRecovery(completedEntries, resumableEntries);
+    }
+
+    /// Resolves one hanging temp-archive multipart upload: aborts it if no parts were ever
+    /// uploaded, otherwise completes it into a materialized readable object, starts a fresh upload
+    /// for the same key, and scans backwards via [#rebuildUpload] to find the last clean
+    /// block-start boundary. Adds a [TempArchiveResumeState] to `resumable` when a boundary is
+    /// found; otherwise aborts the new upload and deletes the intermediate object, mirroring
+    /// [#recoverFromSingleHangingUpload]'s gibberish-parts branch.
+    private void resolveHangingTempUpload(
```
rockysingh, line 482: Related to the map-overwrite risk on the plugin side: because this method is called once per (key, uploadId) pair and appends unconditionally to `resumable`, two hanging uploads for the same `.tmp` key both produce entries with the same `firstBlock`. When the plugin later keys `tempUploadFutures` by `firstBlock`, one future gets orphaned. Consider dedup here (keep the entry with the higher `nextBlockNumber`, abort the loser's new uploadId), so the invariant "one resumable entry per firstBlock" is enforced at the source.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Good observation, @berryware 

However, I don't think we need additional guard here. The reason for that is that at this point we are re-creating the tasks snapshot that existed during the last time Block Node was running. And there we had the bound on the running concurrent temp archives

A slight issue might occur if the administrator decided to reduce that number for some reason between the runs. But "locking" the configuration against changes (and allowing to change for some entries, but dis-allowing for others) is a completely different topic.

Maybe I should create a task for configuration locking? @Nana-EC

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: `resumableTempArchives` over which we iterate is a list, not a `HashMap`. And it is ordered lexicographically by the S3 client API by S3 object key. As our object keys start with pre-padded zeroes, this means that the list is ordered numerically as well.

This list comes from the `RecoveryResult` obtained by the `StartupRecoveryTask`.recoverTempArchives(). Nowhere in the path is there a hash map to scramble the result.

Since 

```java
if (current == null || resumed.nextBlockNumber() > current.nextBlockNumber()) 
```

keeps the first-seen entry on a tie, the current code already deterministically picks the lowest firstBlock as the winner

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Good catch! Changed all sites where we create the instance to pass empty lists instead of null

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
ivannov: Great catch, @rockysingh!

I introduced a check for duplicate .tmp files and a logic to pick only the one with higher `nextBlockNumber`. Also added a test.

berryware (dismissed):
Re-approving

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: Depending on the exact sequence of events, it's theoretically possible to substantially exceed the maximum concurrent temp archives in this process, it requires many such archives to be stored in the bucket, but that's entirely possible.

I'd recommend limiting how many we attempt to resume simultaneously, if it isn't unusually difficult.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
```diff
@@ -248,8 +248,11 @@ private byte[] flushPartIfNeeded(
                 // fall back to blockNum so the notification is never dropped.
                 final Map.Entry<Long, BlockSource> last =
                         blocksInBuffer.isEmpty() ? Map.entry(blockNum, blockSource) : blocksInBuffer.lastEntry();
-                blockMessaging.sendBlockPersisted(
-                        new PersistedNotification(last.getKey(), true, 1_000, last.getValue()));
+                blockMessaging.sendBlockPersisted(new PersistedNotification(
+                        last.getKey(),
+                        true,
+                        CloudStorageArchivePlugin.CLOUD_ARCHIVE_PERSISTENCE_NOTIFICATION_PRIORITY,
```
ata-nas, line 254: An interesting observation here.

In the past, before adding stored vs available blocks, we'd assumed that a persistence notification will only be sent by a block provider. This now turns out to be wrong.

Something we should think about is removing priority from the persisted notifications. I do not believe anyone is using these anywhere coming from the notification.

Priority is useful for the historical facility because block providers are ordered based on priority, which is indeed a much needed optimization. The historical block facility does not rely or use notifications, however. And with plans to soon revisit that facility, possibly removing it, we should rethink how priority is really useful going forward.

I do not think there is anything to do here, just raising awareness.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -59,6 +60,10 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
     /// The logger for this class.
     private static final System.Logger LOGGER = System.getLogger(CloudStorageArchivePlugin.class.getName());
 
+    /// Block-provider priority used in [org.hiero.block.node.spi.blockmessaging.PersistedNotification]s
```
ata-nas, line 65: Here, same as my previous comment, this class is not a block provider. A clear indication we need to rethink priority in notifications.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ata-nas, line 462: Is it possible that the `recoveryFuture` field be updated while we are executing this method? This will lead to NPEs or us calling resultNow on a different future, especially if multiple threads execute this method simultaneously.

Generally, in such situations, if multiple threads can use this, we should think about this being either volatile or atomic, based on needs, but also think about race conditions and also working with locally saved refs.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ata-nas: Will this `put` possibly overwrite an existing task? Do multiple threads access this map? How does this operation affect the whole plugin? This seems dangerous, but I do not have the whole context of the plugin.

ata-nas (commented):
First pass, need to spend more time here.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: As far as I can understand, there are two threads that access `recoveryFuture`. One of them is the one on which the start method is running and the other one is that running the `handleVerification` method (if they are separate threads anyway).

If there is a potential race condition between them, we need to make all fields volatile and not just this one. And maybe we need to change other plugins as well?

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: The first question:

> Will this put possibly overwrite an existing task?

That's exactly the same concern which @rockysingh had on the same line. I addressed it with introducing the `StartupRecoveryTask.keepBestResumeCandidate()` method (forgot to reply to Rocky's comment).

To the second question:

> Do multiple threads access this map?

It's only accessed by the thread sending notifications

ivannov: Sorry, @rockysingh , I should have replied after I submitted the previous commit.

Yes, you were right and I introduced `StartupRecoveryTask.keepBestResumeCandidate()`: it dedups by S3 key before returning `resumableTempArchives`, so two entries can never share a `firstBlock` by the time they reach this put.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: @jsync-swirlds I implemented a cap for resumable archives. If you ask me, it adds to the complexity of the plugin, which even now is quite complex. Please tell me whether I should keep it or just remove the commit

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
ivannov: How do you propose to address that?

BTW, removing altogether the persisted notifications with the current Block Node architecture is not a solution as this is a way to signal the backfill plugin to send blocks, which persistence failed

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: The added complexity seems reasonable to ensure the system remains within stable limits.

jsync-swirlds (dismissed):

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
```diff
@@ -117,23 +149,21 @@ public TempArchiveEntry call()
...
                 } catch (InterruptedException e) {
-                    if (lastProcessedBlock >= firstBlock) {
+                    if (lastProcessedBlock >= loopStart) {
                         LOGGER.log(
                                 INFO,
                                 "Temp archive upload task for key {0} interrupted with blocks [{1}, {2}]"
                                         + " accumulated; completing early instead of aborting",
                                 s3Key,
-                                firstBlock,
+                                loopStart,
```
Nana-EC, line 158: Q: In the completeSegment below are we leaking other exceptions?
What happens if completeSegment fails and having caught the `InterruptedException`
Should it be wrapped na try catch itself?

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -496,6 +523,80 @@ private void completeRecoveryIfReady() {
...
+                LOGGER.log(
+                        TRACE,
+                        "Resumable temp archive firstBlock={0} for group {1} superseded; finalising as-is",
+                        resumed.firstBlock(),
+                        groupStart);
+                tempUploadFutures.put(
+                        resumed.firstBlock(),
+                        virtualThreadExecutor.submit(newResumedTempArchiveUploadTask(
+                                resumed.s3Key(), resumed.firstBlock(), queue, resumed)));
+            } else if (tempGroupActiveQueues.size() >= config.maxConcurrentTempArchives()) {
```
Nana-EC: Claude claims there may be Inconsistent concurrency-limit metric between `resumeHangingTempArchives` and `drainPendingResumedArchives`

Apparently

> `resumeHangingTempArchives` gates deferred resumes using `tempGroupActiveQueues.size()`:
> 
> ```java
> } else if (tempGroupActiveQueues.size() >= config.maxConcurrentTempArchives()) {
>     pendingResumedArchives.add(resumed);
> ```
> 
> But `drainPendingResumedArchives` (and `drainOverflowStash`) use `tempUploadFutures.size()`:
> 
> ```java
> while (!pendingResumedArchives.isEmpty() && tempUploadFutures.size() < config.maxConcurrentTempArchives()) {
> ```
> 
> `tempUploadFutures.size() ≥ tempGroupActiveQueues.size()` always: futures persist until removed by `checkAndDrainTempUploadResults`; active queues are removed as soon as `SEGMENT_END` is placed. In the startup call from `completeRecoveryIfReady()` both maps start empty and the gate in `resumeHangingTempArchives` increments correctly as winners are registered — so there is no practical correctness defect. But superseded segments add a future to `tempUploadFutures` without touching `tempGroupActiveQueues`, meaning `drainPendingResumedArchives` could see a slightly larger count than `resumeHangingTempArchives` measured when deferring. These superseded futures complete almost immediately (SEGMENT_END already placed) and are removed before the next handler cycle, so the over-count is transient. Using `tempUploadFutures.size()` uniformly (as `drainOverflowStash` already does) would align the invariant and reduce confusion during future audits.

Something to clarify

ivannov: Good catch! Fixed by switching `resumeHangingTempArchives` to gate on `tempUploadFutures.size()`, same as the other invariants.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
ivannov: If completeSegment throws in the InterruptedException's catch block (I believe you ask about that code path), the new exception propagates instead of the InterruptedException. But nothing is swallowed - `completeSegment()` method's own catch blocks already send the `PersistedNotification(false)` before rethrowing. And then `checkAndDrainTempUploadResults()` treats any non-successful future completion (whichever exception it was) in the same way.

jsync-swirlds (dismissed):

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
ata-nas: This is, perhaps, a more detailed and thorough design discussion we need to have.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ata-nas: So, `start`, in theory, should be called before any calls to `handleVerification` or any other notification handling method is called, but that is not always true; up to the developer that did the plugin. Note that the `start` method CAN and often is registering the respective plugin as a notification or live rb handler. Once registered, we should expect calls to the handler methods at any time.

Does the plugin take this into consideration? If not, then we will possibly see issues.

Remember, our current sequence is `init -> onContextUpdate (context is loaded from disk already, but can change as a side effect from any of the init calls!) -> start`. At some point of this lifecycle we will register ourselves notification and/or live rb handler, if we register ourselves after we have finished our operations on the fields, we SHOULD theoretically be fine because of the `happens before` & `as-if-serial semantics`.

Note that in real life, in this specific situation, this race should be quite rare, but worth it to reason about it and see if is is possibly problematic for us.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ata-nas: Thanks for clarifying, taking a look.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Digging deeper in the code, I can see your point.

`start()` is called in parallel for all plugins. Block messaging facility is itself a plugin. So in theory, its start method may finish first (thus activating the consumer thread). So there can be a window of time when the block messaging facility can call `handleVerificattion()` before the cloud storage archive plugin's start method has returned (or more precisely before the future is initialized).

This, however, holds true for other plugins as well. Which makes it a general thing to tackle. I don't see it in the scope of the curent task

ata-nas: Yeah, not asking for code changes, just to acknowledge this, note and possibly take an action if needed, when needed.

Nana-EC (dismissed):

ata-nas (dismissed):

ata-nas (approved):

Nana-EC (approved):

jsync-swirlds (approved):

PR #3198, opened 2026-07-10
chore: Update lfh-values override for mainnet

Nana-EC:
## Reviewer Notes
Update lfh-values override for mainnet

## Related Issue(s)

### DIFF
--- a/charts/block-node-server/values-overrides/lfh-values.yaml
+++ b/charts/block-node-server/values-overrides/lfh-values.yaml
@@ -10,7 +10,7 @@ blockNode:
         # BLOCK_NODE_EARLIEST_MANAGED_BLOCK: "110000000"
         BACKFILL_GREEDY: "true"
         BACKFILL_BLOCK_NODE_SOURCES_PATH: "/opt/hiero/block-node/backfill/block-node-sources.json"
-        JAVA_OPTS: "-Xms192G -Xmx192G -XX:+ZGenerational -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/dump.hprof"
+        JAVA_OPTS: "-Xms192G -Xmx192G -XX:+ZGenerational"
         PROMETHEUS_ENDPOINT_ENABLED: true
         PROMETHEUS_ENDPOINT_PORT_NUMBER: "16007"
         # "https://<previewnet|testnet|mainnet>.mirrornode.hedera.com"
--- a/charts/block-node-server/values-overrides/lfh-values.yaml
+++ b/charts/block-node-server/values-overrides/lfh-values.yaml
@@ -10,7 +10,7 @@ blockNode:
         # BLOCK_NODE_EARLIEST_MANAGED_BLOCK: "110000000"
         BACKFILL_GREEDY: "true"
         BACKFILL_BLOCK_NODE_SOURCES_PATH: "/opt/hiero/block-node/backfill/block-node-sources.json"
-        JAVA_OPTS: "-Xms64G -Xmx64G -XX:+ZGenerational -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/dump.hprof"
+        JAVA_OPTS: "-Xms192G -Xmx192G -XX:+ZGenerational -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/dump.hprof"
         PROMETHEUS_ENDPOINT_ENABLED: true
         PROMETHEUS_ENDPOINT_PORT_NUMBER: "16007"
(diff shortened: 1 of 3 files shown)


AlfredoG87 (dismissed):
LGTM, question do we want also to lower the log level from current default, and explicitly set it to INFO level?

right now is TRACE by default, which is too verbose imo.

charts/block-node-server/values-overrides/lfh-values.yaml
```diff
@@ -1,7 +1,21 @@
...
+
 blockNode:
     config:
-        BLOCK_NODE_EARLIEST_MANAGED_BLOCK: "100000000"
-        MESSAGING_BLOCK_NOTIFICATION_QUEUE_SIZE: "512"
+        # BNs expect blocks from 0 by default, the following value supports receiving any block before end of 2026
+        # BLOCK_NODE_EARLIEST_MANAGED_BLOCK: "110000000"
+        BACKFILL_GREEDY: "true"
+        BACKFILL_BLOCK_NODE_SOURCES_PATH: "/opt/hiero/block-node/backfill/block-node-sources.json"
+        JAVA_OPTS: "-Xms64G -Xmx64G -XX:+ZGenerational -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp/dump.hprof"
```
jsync-swirlds: I do not think We want to recommend such low memory settings.
The nodes should have `256G` RAM, and `212G` assigned to the _container_, so 192G allocated to Java heap is appropriate.
Also, generational ZGC is default for Java 25.

We also shouldn't specify the heap dump values, as they're less than ideal for production use.

```suggestion
        JAVA_OPTS: "-Xms192G -Xmx192G"
```

charts/block-node-server/values-overrides/lfh-values.yaml
```diff
@@ -48,12 +62,26 @@ blockNode:
...
+    port: "40840"
+    metallb.io/address-pool: public-address-pool
 
 resources:
     requests:
         cpu: "24"
         memory: "256Gi"
+    limits:
+        cpu: "24"
+        memory: "256Gi"
```
jsync-swirlds: These should be 212G, otherwise we run the risk of the kernel killing the process because there is no memory left for the kernel itself.
That 46G of buffer leaves room for the kernel and modest filesystem cache (which should be ample for our needs).

If we determine FS cache needs to be larger at some point, we can reduce the container to 150G and Java to 128G, but that should be held in reserve until/unless we can prove it's necessary.

jsync-swirlds (commented):
Just a few resource values we may want to adjust.

charts/block-node-server/values-overrides/lfh-values.yaml
Nana-EC: Updated

charts/block-node-server/values-overrides/lfh-values.yaml
Nana-EC: Updated

Nana-EC: Let me confirm an item with Java, provisioner should have us covered but might we leave it in there for the case of direct helm installs that weren't based on Java 25 environments?

jsync-swirlds: Block Node is not supported for Java versions prior to 25, and the compiled binaries won't run on those older JVMs because we compile with Java 25 (newer JVM can run older class files, but older JVMs cannot run newer class files).

charts/block-node-server/values-overrides/lfh-values.yaml
```diff
@@ -1,11 +1,25 @@
...
+        # "https://<previewnet|testnet|mainnet>.mirrornode.hedera.com"
+        ROSTER_BOOTSTRAP_RSA_MIRROR_NODE_BASE_URL: "https://mainnet.mirrornode.hedera.com"
+        SERVER_PORT: "40840"
         logs:
             level: "INFO"
             loggingProperties:
-                org.hiero.block.level: "FINE"
+                org.hiero.block.level: "INFO"
                 java.util.logging.ConsoleHandler.level: "FINEST"
                 java.util.logging.FileHandler.level: "FINEST"
```
jsync-swirlds, line 24: For safety, in production, we might want to change these to `INFO` as well, just in case a library update results in writing a large quantity of trace logs.

jsync-swirlds (approved):

AlfredoG87 (approved):
LGTM

charts/block-node-server/values-overrides/lfh-values.yaml
Nana-EC: Will resolve in follow up

PR #3200, opened 2026-07-10
feat: Add load balancer chart support to config multiple ports

### LINKED ISSUE #3199: Add load balancer config support for multi plugin port feature
### Story Form

As a block node operator
I want to support multiple ports on my load balancer
So that I can expose plugins publicly on multiple ports

### Technical Notes

BN charts has support for unique pots per plugin, however, it doesn't expose this in the loadbalancer configs

Nana-EC:
## Reviewer Notes
The `block-node-server` chart already supports per-plugin port configuration via
`blockNode.ports.*`, and the ClusterIP Service automatically exposes all configured plugin
ports in-cluster. However, the `LoadBalancer` Service template was hardcoded to a single
port, leaving operators with no way to expose more than one plugin port through the external
LoadBalancer when running with per-plugin port config.

This PR extends the `loadBalancer` values block with two new optional keys —
`includePluginPorts` and `extraPorts` — and wires them into `service-loadbalancer.yaml`.
Both keys default to off/empty, so existing deployments are unaffected.

### `values.yaml` — two new `loadBalancer` keys

```yaml
loadBalancer:
  # When true, all non-null blockNode.ports entries are added to the LB Service in
  # addition to the primary port. Requires each plugin port to be distinct from
  # loadBalancer.port.
  includePluginPorts: false

  # Explicit additional ports to expose on the LB Service (beyond the primary port and
  # any plugin ports included via includePluginPorts).
  extraPorts: []
  # - name: my-extra
  #   port: 40845
  #   targetPort: 40845   # optional; defaults to port value
  #   protocol: TCP       # optional; defaults to TCP
```

### `templates/service-loadbalancer.yaml`

Two new blocks are appended after the primary port entry:

```yaml
ports:
  - name: {{ default "grpc" .Values.loadBalancer.portName }}
    port: {{ default .Values.service.port .Values.loadBalancer.port }}
    targetPort: {{ default "http" .Values.loadBalancer.targetPort }}
    protocol: {{ default "TCP" .Values.loadBalancer.protocol }}
  {{- if .Values.loadBalancer.includePluginPorts }}
  {{- include "hiero-block-node.pluginServicePorts" . | nindent 4 }}
  {{- end }}
  {{- range .Values.loadBalancer.extraPorts }}
  - name: {{ .name }}
    port: {{ .port }}
    targetPort: {{ default .port .targetPort }}
    protocol: {{ default "TCP" .protocol }}
  {{- end }}
```

`pluginServicePorts` is the same helper already used by `service.yaml`; it deduplicates
entries by port number internally.

### `ci/lb-multi-port-values.yaml` (new)

CI test values that exercise both paths simultaneously:

| Port name | Number | Via |
|-----------|--------|-----|
| grpc (primary) | 40840 | hardcoded primary LB port |
| health | 40940 | `includePluginPorts: true` |
| publisher | 40941 | `includePluginPorts: true` |
| subscriber | 40942 | `includePluginPorts: true` |
| block-access | 40943 | `includePluginPorts: true` |
| server-status | 40944 | `includePluginPorts: true` |
| extra-test | 40945 | `extraPorts` entry |

`SERVER_PORT: "40840"` is retained because `BlockNodeApp` unconditionally opens a primary
Helidon socket on that port regardless of plugin-specific port assignments. It is set to a
distinct value so the connectivity step can assert that no plugin accidentally falls back to
it (same pattern as `ci/multi-port-values.yaml`).

### `.github/workflows/helm-charts.yaml` — new `lb-multi-port-install-test` job

Independent job triggered by the same file-change predicate as `multi-port-install-test`.
Two verification steps:

1. **LB Service spec check** (`kubectl get service -o json | jq`) — asserts the LB Service
   has exactly seven ports and each expected port number is present. This is the unique
   assertion for this job: it proves `includePluginPorts` and `extraPorts` are wired into
   the Service manifest correctly.

2. **Connectivity check** — mirrors the `multi-port-install-test` pod port-forward pattern:
   verifies each plugin port responds on its configured port and confirms health is absent
   from the shared `SERVER_PORT` (40840).

### `README.md`

Replaces the "planned as a follow-on" note under _LoadBalancer and Per-Plugin Ports_ with
documented examples of all three usage patterns:

- Single port (existing behaviour, unchanged)
- `includePluginPorts: true` (auto-include all `blockNode.ports` entries)
- `extraPorts` (explicit additional ports not in `blockNode.ports`)

## Backward Compatibility

| Scenario | Impact |
|---|---|
| `loadBalancer.enabled: false` (default) | No change — template not rendered |
| `loadBalancer.enabled: true`, new keys absent | No change — both keys default to off/empty; LB Service renders with single port as before |
| `includePluginPorts: true`, no `blockNode.ports` set | No change — helper emits nothing when all ports are null |
| `extraPorts` with duplicate port number | Kubernetes will reject the Service; operators must ensure all port numbers are distinct (documented in README) |

## Testing

- `helm template` (default values): LB Service renders with single `grpc` port — backward
  compatibility confirmed
- `helm template` (`includePluginPorts: true` + one `extraPorts` entry + five plugin ports):
  LB Service renders with all seven expected ports
- CI job `lb-multi-port-install-test`: LB Service port-count assertion + per-plugin
  connectivity checks

## Related Issue(s)
Resolves #3199


.github/workflows/helm-charts.yaml
```diff
@@ -241,3 +241,190 @@ jobs:
...
+        uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc # v1.14.0
+
+      - name: Build Docker images
+        if: steps.install-needed.outputs.needed == 'true'
+        run: ${CGEXEC} ./gradlew :app:createDockerImage
+
+      - name: Load image into kind
+        if: steps.install-needed.outputs.needed == 'true'
+        run: |
+          APP_VERSION=$(cat version.txt)
```
jsync-swirlds, line 296: Is this intended to test "current main" which does not have a tag yet; or specific tags?

.github/workflows/helm-charts.yaml
```diff
@@ -241,3 +241,190 @@ jobs:
...
+          docker tag "block-node-server:${APP_VERSION}" "ghcr.io/hiero-ledger/hiero-block-node:${APP_VERSION}"
+          kind load docker-image "ghcr.io/hiero-ledger/hiero-block-node:${APP_VERSION}" --name chart-testing
```
jsync-swirlds, line 298: Are we sure we want to interact with the remote container registry?
If we can have kind pull from local, then we don't need to worry about two runners pushing different code to the same snapshot tag.

.github/workflows/helm-charts.yaml
```diff
@@ -241,3 +241,190 @@ jobs:
...
+          fi
+
+          # Confirm that the shared server.port (40840) is NOT serving health plugin traffic
+          echo "--- port 40840 (server.port — health plugin must NOT be here) ---"
+          if curl -sf --max-time 5 http://localhost:40840/healthz/livez 2>/dev/null; then
+            echo "FAIL: health plugin is still on 40840 — expected it on 40940 per test config"
+            failed=1
+          else
+            echo "OK (health plugin correctly on 40940, not on shared server.port 40840)"
+          fi
```
jsync-swirlds, line 427: Slight error in text.
replace `40940` (not used) with `40983` (health port)

```suggestion
          # Confirm that the shared server.port (40840) is NOT serving health plugin traffic
          echo "--- port 40840 (server.port — health plugin must NOT be here) ---"
          if curl -sf --max-time 5 http://localhost:40840/healthz/livez 2>/dev/null; then
            echo "FAIL: health plugin is still on 40840 — expected it on 40983 per test config"
            failed=1
          else
            echo "OK (health plugin correctly on 40983, not on shared server.port 40840)"
          fi
```

charts/block-node-server/ci/lb/lb-multi-port-values.yaml
```diff
@@ -0,0 +1,55 @@
...
+#   serverStatus           → 40944   (included via includePluginPorts)
+#   extra-test             → 40945   (included via extraPorts)
+
+loadBalancer:
+  enabled: true
+  includePluginPorts: true
+  extraPorts:
+    - name: extra-test
+      port: 40945
+      targetPort: 40944   # routes to serverStatus container port for connectivity verification
```
jsync-swirlds: You might be intending 40982, if the comment here is correct.

charts/block-node-server/values.yaml
```diff
@@ -63,6 +63,16 @@ loadBalancer:
     # - 203.0.113.0/24  # your clients
     # - 35.191.0.0/16   # GCP health check (example)
     # - 130.211.0.0/22  # GCP health check (example)
+  # When true, all non-null blockNode.ports entries are added to the LB Service in addition to
+  # the primary port. Requires each plugin port to be distinct from loadBalancer.port.
+  includePluginPorts: false
```
jsync-swirlds, line 68: I would expect the default to be `true` here...

jsync-swirlds (commented):
Looks good, a few suggestions, corrections, and questions.

.github/workflows/helm-charts.yaml
Nana-EC: Yes in the `main` branch case.
It's going off of whatever the version in branch is.
This is the same logic used for all the jobs in this file, `lint` included.

.github/workflows/helm-charts.yaml
Nana-EC: Yeah, let me revisit this logic, as noted above this is consistent with existing jobs so if there's a fix here we should do it for all the jobs in this file.
I like the idea of a local build instead. I think i use this approach in my other PR with provisioner so let me see if I can pull that logic, either in this PR or in follow up

.github/workflows/helm-charts.yaml
Nana-EC: Thanks, left over from changing the values

charts/block-node-server/ci/lb/lb-multi-port-values.yaml
Nana-EC: Yes, need to update after updating initial values

charts/block-node-server/values.yaml
Nana-EC: Good consideration, this would immediately break everyone and force multi port.
For testing I believe we want to leave solo and others to use 1 port, for production indeed we should default to true.

AlfredoG87: I think we should consider deprecating a singlePort in the near future, so that we can test in the closest similar conditions to how the BNs will be deployed on production environments.

.github/workflows/helm-charts.yaml
Nana-EC: Will address separately

.github/workflows/helm-charts.yaml
Nana-EC: Resolved

charts/block-node-server/ci/lb/lb-multi-port-values.yaml
Nana-EC: Resolved

AlfredoG87 (approved):
LGTM!

jsync-swirlds (approved):

PR #3209, opened 2026-07-13
fix: Expand backfill to add failed persistence with source of BlockSource.BACKFILL to on-demand backfill

### LINKED ISSUE #3208: Expand backfill to add failed persistence with source of ```BlockSource.BACKFILL``` to on-demand backfill
### Story Form

As a block node engineer
I want to make sure that blocks with a source of ```BlockSource.BACKFILL``` that fail persistence are added to on-demand backfill
So that these blocks eventually get backfilled.

### Technical Notes

The current [handlePersisted](https://github.com/hiero-ledger/hiero-block-node/blob/health-plugin-independent-webserver/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java#L533-L545) method is not managing failed persistence. 

Failure notifications should add the block to the on-demand backfill where it can be retried.

berryware:
### Refactored handlePersisted in BackfillPlugin.java
- It now checks notification.succeeded() instead of treating every BACKFILL-sourced notification as a success.
- On success: unchanged behavior (increments backfillBlocksBackfilled).
- On failure: logs at INFO, increments backfillFetchErrors, and re-submits the failed block directly to the live-tail scheduler as a single-block on-demand gap (bypassing the high-water-mark dedup that would otherwise silently drop the retry).
- pendingBackfillBlocks is decremented in both cases, since the block is no longer in-flight either way.
- Added a null-check on liveTailScheduler for the narrow window between init() registering the handler and start() creating the schedulers, matching similar guards elsewhere in the file.

### BackfillPluginTest.java
- Added an end-to-end integration test in  that induces a single persistence failure mid-way through an on-demand backfill run and confirms the block is retried and eventually persisted successfully, with exactly one failed notification recorded

## Related Issue(s)
Closes: #3208

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -512,11 +512,12 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
      * Submits a gap to the appropriate scheduler based on its type.
      *
      * @param gap the gap to submit
+     * @return true if the gap was accepted, false if the queue was full (gap discarded)
      */
-    private void submitGap(GapDetector.Gap gap) {
+    private boolean submitGap(GapDetector.Gap gap) {
         BackfillTaskScheduler scheduler =
                 (gap.type() == GapDetector.Type.HISTORICAL) ? historicalScheduler : liveTailScheduler;
-        scheduler.submit(gap);
+        return scheduler.submit(gap);
     }
 
     // Package-private for test visibility
@@ -545,7 +546,14 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
                     final String backfillPersistFailedMsg =
                             "Backfill persistence failed for block=[{0}], re-queuing for on-demand backfill";
                     LOGGER.log(INFO, backfillPersistFailedMsg, blockNumber);
-                    submitGap(new GapDetector.Gap(new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL));
+                    if (!submitGap(
+                            new GapDetector.Gap(new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL))) {
+                        // if the schedule queue is full mark this as a failure for visibility
+                        // This block will now be picked up during the next periodic autonomous scan of
+                        // detectAndScheduleGaps, which will see the block missing from stored ranges on its next tick
+                        // and re-detect/resubmit it as a gap
+                        metricsHolder.backfillPersistenceFailures().increment();
+                    }
                 } else {
                     final String liveTailSchedulerNullMsg =
                             "LiveScheduler unavailable. Persistence failed for block=[{0}], unable to requeue for on-demand backfill";
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -532,11 +532,22 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
     @Override
     public void handlePersisted(PersistedNotification notification) {
         if (notification.blockSource() == BlockSource.BACKFILL) {
-            // Add more detailed logging for persistence notifications
-            final String backfillPersistedMsg = "Received backfill persisted notification for block=[{0}]";
-            LOGGER.log(TRACE, backfillPersistedMsg, notification.blockNumber());
-
-            metricsHolder.backfillBlocksBackfilled().increment();
+            final long blockNumber = notification.blockNumber();
+            if (notification.succeeded()) {
+                final String backfillPersistedMsg = "Received backfill persisted notification for block=[{0}]";
+                LOGGER.log(TRACE, backfillPersistedMsg, blockNumber);
+                metricsHolder.backfillBlocksBackfilled().increment();
+            } else {
+                final String backfillPersistFailedMsg =
+                        "Backfill persistence failed for block=[{0}], re-queuing for on-demand backfill";
+                LOGGER.log(INFO, backfillPersistFailedMsg, blockNumber);
+                metricsHolder.backfillFetchErrors().increment();
+                // Submit directly rather than via scheduleGap: the block is very likely already below
+                // the live-tail high-water mark, and scheduleGap's dedup would otherwise drop the retry.
+                if (liveTailScheduler != null) {
+                    submitGap(new GapDetector.Gap(new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL));
+                }
+            }
             pendingBackfillBlocks.updateAndGet(v -> Math.max(0, v - 1));
         } else {
             final String nonBackfillPersistedMsg = "Received non-backfill persisted notification: [{0}]";
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -68,6 +68,8 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             MetricKey.of("backfill_fetch_errors", LongCounter.class).addCategory(METRICS_CATEGORY);
     public static final MetricKey<LongCounter> METRIC_BACKFILL_RETRIES =
             MetricKey.of("backfill_retries", LongCounter.class).addCategory(METRICS_CATEGORY);
+    public static final MetricKey<LongCounter> METRIC_BACKFILL_PERSISTENCE_FAILURES =
+            MetricKey.of("backfill_persistence_failure", LongCounter.class).addCategory(METRICS_CATEGORY);
 
     /** The logger for this class. */
     private final System.Logger LOGGER = System.getLogger(getClass().getName());
@@ -531,11 +533,9 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
      */
     @Override
     public void handlePersisted(PersistedNotification notification) {
-        if (notification.blockSource() == BlockSource.BACKFILL) {
+        if (notification != null && notification.blockSource() == BlockSource.BACKFILL) {
             final long blockNumber = notification.blockNumber();
             if (notification.succeeded()) {
-                final String backfillPersistedMsg = "Received backfill persisted notification for block=[{0}]";
-                LOGGER.log(TRACE, backfillPersistedMsg, blockNumber);
                 metricsHolder.backfillBlocksBackfilled().increment();
             } else {
                 final String backfillPersistFailedMsg =
@@ -549,9 +549,6 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
                 }
             }
             pendingBackfillBlocks.updateAndGet(v -> Math.max(0, v - 1));
-        } else {
-            final String nonBackfillPersistedMsg = "Received non-backfill persisted notification: [{0}]";
-            LOGGER.log(TRACE, nonBackfillPersistedMsg, notification);
         }
     }
 
@@ -641,7 +638,8 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             LongCounter.Measurement backfillFetchedBlocks,
             LongCounter.Measurement backfillBlocksBackfilled,
             LongCounter.Measurement backfillFetchErrors,
-            LongCounter.Measurement backfillRetries) {
+            LongCounter.Measurement backfillRetries,
+            LongCounter.Measurement backfillPersistenceFailures) {
 
         /**
          * Factory method to create a MetricsHolder with all metrics registered.
@@ -686,6 +684,10 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
                             .getOrCreateNotLabeled(),
                     metricRegistry
                             .register(LongCounter.builder(METRIC_BACKFILL_RETRIES)
+                                    .setDescription("Number of retries during the backfill process."))
+                            .getOrCreateNotLabeled(),
+                    metricRegistry
+                            .register(LongCounter.builder(METRIC_BACKFILL_PERSISTENCE_FAILURES)
                                     .setDescription("Number of retries during the backfill process."))
                             .getOrCreateNotLabeled());
         }
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -542,19 +542,18 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
                 // Submit directly rather than via scheduleGap: the block is very likely already below
                 // the live-tail high-water mark, and scheduleGap's dedup would otherwise drop the retry.
                 final long blockNumber = notification.blockNumber();
-                if (liveTailScheduler != null) {
+                if (liveTailScheduler != null
+                        && submitGap(new GapDetector.Gap(
+                                new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL))) {
                     final String backfillPersistFailedMsg =
                             "Backfill persistence failed for block=[{0}], re-queuing for on-demand backfill";
                     LOGGER.log(INFO, backfillPersistFailedMsg, blockNumber);
-                    if (!submitGap(
-                            new GapDetector.Gap(new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL))) {
-                        // if the schedule queue is full mark this as a failure for visibility
-                        // This block will now be picked up during the next periodic autonomous scan of
-                        // detectAndScheduleGaps, which will see the block missing from stored ranges on its next tick
-                        // and re-detect/resubmit it as a gap
-                        metricsHolder.backfillPersistenceFailures().increment();
-                    }
                 } else {
+                    // if the scheduler is null or schedule queue is full mark this as a failure for visibility
+                    // This block will now be picked up during the next periodic autonomous scan of
+                    // detectAndScheduleGaps, which will see the block missing from stored ranges on its next tick
+                    // and re-detect/resubmit it as a gap
+                    metricsHolder.backfillPersistenceFailures().increment();
                     final String liveTailSchedulerNullMsg =
                             "LiveScheduler unavailable. Persistence failed for block=[{0}], unable to requeue for on-demand backfill";
                     LOGGER.log(INFO, liveTailSchedulerNullMsg, blockNumber);
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -69,7 +69,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
     public static final MetricKey<LongCounter> METRIC_BACKFILL_RETRIES =
             MetricKey.of("backfill_retries", LongCounter.class).addCategory(METRICS_CATEGORY);
     public static final MetricKey<LongCounter> METRIC_BACKFILL_PERSISTENCE_FAILURES =
-            MetricKey.of("backfill_persistence_failure", LongCounter.class).addCategory(METRICS_CATEGORY);
+            MetricKey.of("backfill_persistence_failures", LongCounter.class).addCategory(METRICS_CATEGORY);
 
     /** The logger for this class. */
     private final System.Logger LOGGER = System.getLogger(getClass().getName());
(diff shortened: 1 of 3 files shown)


block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -532,11 +532,22 @@ LongRange computeChunk(
...
+                metricsHolder.backfillFetchErrors().increment();
+                // Submit directly rather than via scheduleGap: the block is very likely already below
+                // the live-tail high-water mark, and scheduleGap's dedup would otherwise drop the retry.
+                if (liveTailScheduler != null) {
+                    submitGap(new GapDetector.Gap(new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL));
+                }
+            }
             pendingBackfillBlocks.updateAndGet(v -> Math.max(0, v - 1));
         } else {
             final String nonBackfillPersistedMsg = "Received non-backfill persisted notification: [{0}]";
```
ata-nas: This log seems redundant. We should not be logging if we have received a non-backfill notification.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -532,11 +532,22 @@ LongRange computeChunk(
...
         if (notification.blockSource() == BlockSource.BACKFILL) {
-            // Add more detailed logging for persistence notifications
-            final String backfillPersistedMsg = "Received backfill persisted notification for block=[{0}]";
-            LOGGER.log(TRACE, backfillPersistedMsg, notification.blockNumber());
-
-            metricsHolder.backfillBlocksBackfilled().increment();
+            final long blockNumber = notification.blockNumber();
+            if (notification.succeeded()) {
+                final String backfillPersistedMsg = "Received backfill persisted notification for block=[{0}]";
+                LOGGER.log(TRACE, backfillPersistedMsg, blockNumber);
```
ata-nas: This log seems redundant, we have a metric that increments just below.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -532,11 +532,22 @@ LongRange computeChunk(
...
+                metricsHolder.backfillBlocksBackfilled().increment();
+            } else {
+                final String backfillPersistFailedMsg =
+                        "Backfill persistence failed for block=[{0}], re-queuing for on-demand backfill";
+                LOGGER.log(INFO, backfillPersistFailedMsg, blockNumber);
+                metricsHolder.backfillFetchErrors().increment();
```
ata-nas: This metric seems off. Why do we increment a fetch error? We just received a failed persistence.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -532,11 +532,22 @@ LongRange computeChunk(
...
+                // Submit directly rather than via scheduleGap: the block is very likely already below
+                // the live-tail high-water mark, and scheduleGap's dedup would otherwise drop the retry.
+                if (liveTailScheduler != null) {
+                    submitGap(new GapDetector.Gap(new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL));
```
ata-nas: Is the gap detector type the right one?

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -532,11 +532,22 @@ LongRange computeChunk(
     @Override
     public void handlePersisted(PersistedNotification notification) {
         if (notification.blockSource() == BlockSource.BACKFILL) {
```
ata-nas, line 537: Nit: we can check for null notification. It is always a good idea to be safe. Considered also for verification notifications.

ata-nas (commented):
Cleanup suggestions.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: added null check

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: removed the trace log.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: added a new metric.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: It is handling the same way as in: handleNewestBlockKnownToNetwork.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: Removed.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -531,16 +533,22 @@ LongRange computeChunk(
...
     @Override
     public void handlePersisted(PersistedNotification notification) {
-        if (notification.blockSource() == BlockSource.BACKFILL) {
-            // Add more detailed logging for persistence notifications
-            final String backfillPersistedMsg = "Received backfill persisted notification for block=[{0}]";
-            LOGGER.log(TRACE, backfillPersistedMsg, notification.blockNumber());
-
-            metricsHolder.backfillBlocksBackfilled().increment();
+        if (notification != null && notification.blockSource() == BlockSource.BACKFILL) {
+            final long blockNumber = notification.blockNumber();
```
Nana-EC: nit: move into `else`

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -531,16 +533,22 @@ LongRange computeChunk(
...
+            if (notification.succeeded()) {
+                metricsHolder.backfillBlocksBackfilled().increment();
+            } else {
+                final String backfillPersistFailedMsg =
```
Nana-EC: move into `liveTailScheduler` if

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -676,6 +685,10 @@ public static MetricsHolder createMetrics(
                     metricRegistry
                             .register(LongCounter.builder(METRIC_BACKFILL_RETRIES)
                                     .setDescription("Number of retries during the backfill process."))
+                            .getOrCreateNotLabeled(),
+                    metricRegistry
+                            .register(LongCounter.builder(METRIC_BACKFILL_PERSISTENCE_FAILURES)
+                                    .setDescription("Number of retries during the backfill process."))
```
Nana-EC: incorrect description

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -531,16 +533,22 @@ LongRange computeChunk(
...
-            metricsHolder.backfillBlocksBackfilled().increment();
+        if (notification != null && notification.blockSource() == BlockSource.BACKFILL) {
+            final long blockNumber = notification.blockNumber();
+            if (notification.succeeded()) {
+                metricsHolder.backfillBlocksBackfilled().increment();
+            } else {
+                final String backfillPersistFailedMsg =
+                        "Backfill persistence failed for block=[{0}], re-queuing for on-demand backfill";
+                LOGGER.log(INFO, backfillPersistFailedMsg, blockNumber);
+                metricsHolder.backfillPersistenceFailures.increment();
```
Nana-EC:
```suggestion
                metricsHolder.backfillPersistenceFailures().increment();
```

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillRunnerTest.java
```diff
@@ -68,6 +68,7 @@ void setUp() {
                 mockFetchedBlocksCounter, // backfillFetchedBlocks
                 mock(LongCounter.Measurement.class), // backfillBlocksBackfilled
                 mockFetchErrorsCounter, // backfillFetchErrors
+                mock(LongCounter.Measurement.class),
```
Nana-EC: Comment is now out of place. Next like is //backfillPersistenceFailures
```suggestion
                mock(LongCounter.Measurement.class), // backfillInFlightGauge
```

Nana-EC (dismissed):
Logic is in the right direction. Non-blocking suggestions on other areas

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillPluginTest.java
```diff
@@ -659,6 +660,76 @@ void testBackfillOnDemand() throws InterruptedException {
...
+        assertEquals(
+                21,
+                persisted.size(),
+                "Should have sent 20 successful notifications plus 1 failed notification for the retried block");
+        long failedCount = persisted.stream().filter(n -> !n.succeeded()).count();
+        assertEquals(1, failedCount, "Exactly one persisted notification should report failure");
+        assertTrue(
+                persisted.stream().anyMatch(n -> n.blockNumber() == targetBlock && n.succeeded()),
+                "The retried block should eventually be persisted successfully");
+    }
```
Nana-EC, line 735: Also might be valuable to check metric
```suggestion
        assertEquals(1L, getMetricValue(BackfillPlugin.METRIC_BACKFILL_PERSISTENCE_FAILURES), "single backfill persistence failure was captured");
    }
```

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
jsync-swirlds: ~I would recommend using the `HISTORICAL` gap detector.~
~live tail is intended to be specific to "greedy" backfill, which probably isn't what we expect here.~

~It doesn't matter much because it only affects the scheduler priority, and we're skipping over that to send this to `liveTail` directly.~

Corrections:
1. We probably _do_ want to use LIVE_TAIL here.
1. The gap detector type selects which scheduler is used by `submitGap`.
1. The historical scheduler runs at very low frequency, so we probably want the live scheduler.

AlfredoG87 (dismissed):
LGTM

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
rockysingh: This is the description for the new `METRIC_BACKFILL_PERSISTENCE_FAILURES` metric but reads `"Number of retries during the backfill process."` same text as `METRIC_BACKFILL_RETRIES` four lines above. Should be something like `"Number of persistence failures for backfilled blocks."` Otherwise could be interesting when debugging

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -531,16 +533,22 @@ LongRange computeChunk(
...
+                // Submit directly rather than via scheduleGap: the block is very likely already below
+                // the live-tail high-water mark, and scheduleGap's dedup would otherwise drop the retry.
+                if (liveTailScheduler != null) {
```
rockysingh: Silent drop: if `liveTailScheduler` is still `null` (the narrow init→start window), the failed block is metric-counted but the retry is silently discarded as we have no log, no other queue. Given the whole point of the fix is "failed persists must not vanish," consider at minimum a WARN log noting the drop, or defer the retry (e.g. buffer to a `pendingRetries` list drained on `start()`) so the block is genuinely not lost during the init window. As-is, the guard exchanges one silent-drop bug for a smaller silent-drop bug.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -531,16 +533,22 @@ LongRange computeChunk(
...
+                    submitGap(new GapDetector.Gap(new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL));
+                }
+            }
             pendingBackfillBlocks.updateAndGet(v -> Math.max(0, v - 1));
```
rockysingh, line 563: question: this decrement fires unconditionally, including after the failure branch. Presumably `submitGap` → `BackfillRunner` re-increments `pendingBackfillBlocks` when it picks up the retry (based on the `pendingBackfillBlocks` reference passed to the runner at line 281), so the net accounting balances. But there's a brief window where `pendingBackfillBlocks` can hit 0 with a retry still queued, which could make `determineStatus` briefly report "done" mid-retry. Worth a comment here confirming the runner re-increments, so future readers don't try to "fix" this as an off-by-one.

rockysingh (commented):
Good bug fix: pre-fix code was silently treating every BACKFILL-sourced `PersistedNotification` as success, so a failed persist would drop the block with no retry. The direct-`submitGap` bypass of `scheduleGap`'s dedup is reasonable and well-commented, and the end-to-end retry test is a strong addition. A few small findings inline: one clear bug (metric description copy-paste), one observability gap on the `liveTailScheduler != null` path, and a couple of nits. They are not blocker but worth considering.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -68,6 +68,8 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             MetricKey.of("backfill_fetch_errors", LongCounter.class).addCategory(METRICS_CATEGORY);
     public static final MetricKey<LongCounter> METRIC_BACKFILL_RETRIES =
             MetricKey.of("backfill_retries", LongCounter.class).addCategory(METRICS_CATEGORY);
+    public static final MetricKey<LongCounter> METRIC_BACKFILL_PERSISTENCE_FAILURES =
+            MetricKey.of("backfill_persistence_failure", LongCounter.class).addCategory(METRICS_CATEGORY);
```
rockysingh: Nit: metric name is singular (`backfill_persistence_failure`) while every other counter in this file is plural (`backfill_fetch_errors`, `backfill_retries`, `backfill_blocks_backfilled`). Consider `backfill_persistence_failures` for consistency easier to remember and matches the `backfillPersistenceFailures` field name.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -531,16 +533,22 @@ LongRange computeChunk(
      */
     @Override
     public void handlePersisted(PersistedNotification notification) {
-        if (notification.blockSource() == BlockSource.BACKFILL) {
-            // Add more detailed logging for persistence notifications
-            final String backfillPersistedMsg = "Received backfill persisted notification for block=[{0}]";
-            LOGGER.log(TRACE, backfillPersistedMsg, notification.blockNumber());
-
-            metricsHolder.backfillBlocksBackfilled().increment();
+        if (notification != null && notification.blockSource() == BlockSource.BACKFILL) {
```
rockysingh: Nit: the `notification != null` check is a departure from `handleVerification` right below (line 560), which trusts the framework contract and calls `notification.source()` unguarded. Either the guard is genuinely needed (in which case `handleVerification` has a latent NPE and should also be guarded), or it isn't and this should be removed for consistency. Would lean toward removing since the messaging facility shouldn't be delivering nulls.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: changed

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: removed

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: moved

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: moved all the logging as well

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: updated

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: added message

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: added a comment

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: updated

berryware: updated

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillPluginTest.java
berryware: updated

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillRunnerTest.java
berryware: moved

Nana-EC (dismissed):

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -532,15 +534,27 @@ LongRange computeChunk(
...
+            } else {
+                metricsHolder.backfillPersistenceFailures().increment();
+                // Submit directly rather than via scheduleGap: the block is very likely already below
+                // the live-tail high-water mark, and scheduleGap's dedup would otherwise drop the retry.
+                final long blockNumber = notification.blockNumber();
+                if (liveTailScheduler != null) {
+                    final String backfillPersistFailedMsg =
+                            "Backfill persistence failed for block=[{0}], re-queuing for on-demand backfill";
+                    LOGGER.log(INFO, backfillPersistFailedMsg, blockNumber);
+                    submitGap(new GapDetector.Gap(new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL));
```
ivannov: Not sure how possible scenario is that. But what if the queue of the live tail scheduler is full? We will not be notified (the result is already swallowed in `submitGap()` and this block will effectively be skipped, just at a different level than before this PR

berryware: Every other call site (scheduleGap from the autonomous scan and from handleNewestBlockKnownToNetwork) already ignores this return value the same way. It's the existing "discard under backpressure" design for this scheduler.

The failed block isn't lost forever though: the periodic autonomous scan (detectAndScheduleGaps, every scanInterval) will still see the block missing from stored ranges on its next tick and re-detect/resubmit it as a gap — that's the same safety net handleVerification's failure path already relies on. So worst case, a persistence failure under a full live-tail queue degrades from "retried immediately" to "retried on the next periodic scan," rather than being dropped permanently.

Added a comment.

ivannov (approved):

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
berryware: removed null check per Rocky's comment. It was not there before and none of the other handlers check for null. Going to keep it consistent.

jsync-swirlds (approved):

AlfredoG87 (approved):

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -68,6 +68,8 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             MetricKey.of("backfill_fetch_errors", LongCounter.class).addCategory(METRICS_CATEGORY);
     public static final MetricKey<LongCounter> METRIC_BACKFILL_RETRIES =
             MetricKey.of("backfill_retries", LongCounter.class).addCategory(METRICS_CATEGORY);
+    public static final MetricKey<LongCounter> METRIC_BACKFILL_PERSISTENCE_FAILURES =
+            MetricKey.of("backfill_persistence_failures", LongCounter.class).addCategory(METRICS_CATEGORY);
```
ata-nas, line 72: @AlfredoG87 @berryware @jsync-swirlds please note that we need to include this new metric into the dashboards so it is visible.

PR #3212, opened 2026-07-13
fix: Fixed Server Status handling of earliest managed block

### LINKED ISSUE #3203: bug: serverStatus reporting EMB for last_available_block
### Description
It was observed that a BN with some blocks was reporting a `last_available_block` that exactly matched its `BLOCK_NODE_EARLIEST_MANAGED_BLOCK`.
This is done in the update to factor in BLOCK_NODE_EARLIEST_MANAGED_BLOCK into the serverStatus logic, however it has unintended consequences for CN streaming decisions.

In our case the following _expected behavior_ was observed when calling `./bn-endpoint-checker.sh` in its endpoint
```
:large_green_circle: TCP reachable  (75.2 s)
  :large_green_circle: serverStatus  (630 ms)
     first_available_block:         131233
     last_available_block:          100000000
     only_latest_state:             false
  :package: serverStatusDetail  (482 ms)
     BN version:                    0.38.0
     stream_proto_version:          0.73.0-rc.1
     available_ranges:              1 range(s)
       range 1: blocks 131233 – 131350
```

### Steps to Reproduce
2. publish a set of blocks to the BN that start and end less than 100000000
3. Stop publishing
1. Set `BLOCK_NODE_EARLIEST_MANAGED_BLOCK=100000000`
3. Resume publishing
4. Query serverStatus and observe `last_available_block`
5. Check CN metrics and verify no block items being sent

### Additional context
Instead of current behavior we can replace with the following small change:
> While latest available block `<` EMB, we should return `-1` in the status API (Unknown block number) to indicate to publishers that the block node will accept the next block, whatever that is.
This will cause CN to publish the next available block.
Subscribers must check the "detail" status API for more information.

jsync-swirlds:
## Reviewer Notes
* When
   * The `earliestManagedBlock` is greater than the last available block
   * The last available block is not `-1`
   * The `earliestManagedBlock` is greater than `0`
* Then
  * ServerStatus should return `-1` to indicate "last block is unmanaged".

## Related Issue(s)
 Fixes #3203

### DIFF
--- a/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
+++ b/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
@@ -64,8 +64,12 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
         // later block as the last available block so publishers know they can stream from earliestManagedBlock
         // instead of assuming the node only wants older blocks between the oldest stored block and earliest managed
         // block.
-        if (highestAvailableBlock != UNKNOWN_BLOCK_NUMBER && highestAvailableBlock < earliestManagedBlock) {
-            highestAvailableBlock = earliestManagedBlock;
+        if (highestAvailableBlock != UNKNOWN_BLOCK_NUMBER
+                && highestAvailableBlock < earliestManagedBlock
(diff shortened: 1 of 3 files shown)


AlfredoG87 (approved):
LGTM

rockysingh (approved):
lgtm

PR #3216, opened 2026-07-14
feat: Modify Health Plugin to use a separate Helidon Webserver.

### LINKED ISSUE #3174: Health Plugin should have its own webServer and configuration
### Story Form

As a BlockNode Operator
I want to set different WebServer Config for Health Endpoints
So that we can apply different rules than the rest of the system

### Technical Notes

Health checks are private only accesible to the K8S for monitoring the health of the application, and they are HTTP1.1 and REST instead of the rest of the system which is gRPC and HTTP2, so we need effectively different rules and config and makes sense to have it on its own separate port explicitly and by default

jsync-swirlds:
## Reviewer Notes
* Moved webserver management into the ServiceBuilder
* Added a method to register a new http server
* Added a RecordingServiceBuilder test fixture.
* Removed the BlockNodePluginTest that existed only for interface coverage and
  did not test any actual logic, but created a great deal of
  unnecessary complexity.
* Adjusted the ServiceBuilder `registerHttpNewServer` to accept an array of
  services with path for each port, instead of only one. Some servers
  (including health plugin) need multiple paths on one port.
* Modified the Health plugin to register a new webserver on it's own
  port. This precludes the Health plugin using the default port, so
  the Health Plugin port configuration now has a non-null default.
* Added "common socket values" to "new server" parameters, updated
  Health Plugin to provide its own config for those values, and
  adjusted implementations accordingly.
* Reduced most socket releated resources in Health Plugin because
  it does not need large buffers, deep queues, or a large number of
  open connections.
* Updated tests and implementations to match changes.
* Resolved review feedback.
   * Small code updates
   * Updated suites and charts for the new separate health port
   * Modified E2E Suites to remove dependence on command unique to a
     proprietary runtime.
   * Also modified run-in-container use to allow for retry due to
     idle process connection compatibility issues.

## Related Issue(s)
 Closes #3174


block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
```diff
@@ -31,20 +31,14 @@
...
+        @Loggable @ConfigProperty(defaultValue = "8_388_608") @Min(32768) @Max(Integer.MAX_VALUE) int socketReceiveBufferSizeBytes,
+        @Loggable @ConfigProperty(defaultValue = "40840") @Min(1024) @Max(65_535) int port,
         @Loggable @ConfigProperty(defaultValue = "500") int shutdownDelayMillis,
         @Loggable @ConfigProperty(defaultValue = "1000") int maxTcpConnections,
-        @Loggable @ConfigProperty(defaultValue = "1")
-            @Min(1) @Max(60) int idleConnectionPeriodMinutes,
-        @Loggable @ConfigProperty(defaultValue = "2")
-            @Min(1) @Max(1_440) int idleConnectionTimeoutMinutes,
+        @Loggable @ConfigProperty(defaultValue = "1") @Min(1) @Max(60) int idleConnectionPeriodMinutes,
+        @Loggable @ConfigProperty(defaultValue = "2") @Min(1) @Max(1_440) int idleConnectionTimeoutMinutes,
```
ivannov: Shouldn't we restore the default values to 5 and 30 respectively as they were before @AlfredoG87 's last change now as we have health service's own web service config?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -340,79 +314,16 @@ protected final BlockNodeVersions versionInfo(final List<BlockNodePlugin> plugin
...
-                INFO,
-                "BlockNode Server listening on port(s): {0}",
-                allPorts.stream().map(String::valueOf).collect(Collectors.joining(", ")));
         // start the ApplicationStateFacility
         startApplicationStateFacility();
         // start the plugins
         startPlugins(loadedPlugins);
         // mark the server as started
         state.set(State.RUNNING);
+        serviceBuilder.startAll();
```
ata-nas, line 326: Should this be above setting the state to running?

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
```diff
@@ -2,77 +2,251 @@
...
         httpBuilders.computeIfAbsent(resolve(port), k -> HttpRouting.builder()).register(path, service);
     }
 
-    /**
-     * {@inheritDoc}
-     */
+    /// {@inheritDoc}
     @Override
-    public void registerGrpcService(@NonNull ServiceInterface service, @Nullable Integer port) {
+    public void registerGrpcService(@Nullable Integer port, @NonNull ServiceInterface service) {
```
ata-nas, line 62: Why do we allow a nullable port for default value?
Instead, we could override this method to only accept service interface, internally calling this with a specific value. I know this is from the past and is some extra work, but this could be misused by others.

Not pushing for this, but I think it is a very nice to have.

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
```diff
@@ -2,77 +2,251 @@
...
+    }
+
+    @Override
+    public WebServerResult registerHttpNewServer(
+            final TreeMap<Integer, ServiceWithPath[]> services,
+            final Http2Config http2Config,
+            final SocketOptions socketOptions,
+            final CommonSocketValues commonSocketValues) {
+        // build a single WebServer with named sockets for each port/path.
+        Map<Integer, HttpRouting.Builder> routeBuilders = new TreeMap<>();
```
ata-nas: Nit:
```suggestion
        TreeMap<Integer, HttpRouting.Builder> routeBuilders = new TreeMap<>();
```

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
```diff
@@ -74,13 +84,24 @@ public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
...
+        // This must not be null, and the Health Plugin cannot share the
+        // "general" webserver with the rest of the Block Node.
+        final Integer port = healthConfig.port();
+        // Construct a map from port to services
+        TreeMap<Integer, ServiceWithPath[]> servicesToRegister = new TreeMap<>();
+        ServiceWithPath healthService = new ServiceWithPath(
+                HEALTHZ_PATH,
+                httpRules -> httpRules.get(LIVEZ_PATH, this::handleLivez).get(READYZ_PATH, this::handleReadyz));
+        ServiceWithPath statusService =
+                new ServiceWithPath(STATUSZ_PATH, httpRules -> httpRules.get("*", this::handleStatusz));
```
ata-nas, line 101: Interesting, not entirely familiar with the mapping of helidon, but won't `*` conflict with livez and readyz? Or does this map as if livez or healthz go to specific method, anything else is statusz?

block-node/spi-plugins/build.gradle.kts
```diff
@@ -8,14 +8,14 @@ description = "Hiero Block Node Plugin SPI"
 tasks.withType<JavaCompile>().configureEach { options.compilerArgs.add("-Xlint:-exports") }
 
 mainModuleInfo {
+    runtimeOnly("com.hedera.pbj.grpc.helidon.config")
     runtimeOnly("com.swirlds.config.impl")
-    runtimeOnly("org.apache.logging.log4j.slf4j2.impl")
     runtimeOnly("io.helidon.logging.jul")
-    runtimeOnly("com.hedera.pbj.grpc.helidon.config")
+    runtimeOnly("org.apache.logging.log4j.slf4j2.impl")
```
ata-nas, line 14: This looks like an undesired addition.

block-node/spi-plugins/src/test/java/org/hiero/block/node/spi/BlockNodePluginTest.java
ata-nas, line 1: Thank you.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
```diff
@@ -16,21 +21,95 @@
  * and pass it explicitly; plugins that are happy sharing the default port simply pass {@code null}.
  */
```
ata-nas: We can transform to MD javadoc.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
```diff
@@ -16,21 +21,95 @@
...
+    /// @param serverCreated A webserver that was newly created.
+    /// @param portsEnabled The ports routed to this server.
+    public static record WebServerResult(WebServer serverCreated, Set<Integer> portsEnabled) {}
+
+    /// The socket values shared between socket options and http 2 config
+    /// @param backlogSize
+    /// @param writeQueueLength
+    /// @param maxTcpConnections
+    /// @param idleConnectionPeriodMinutes
+    /// @param idleConnectionTimeoutMinutes
```
ata-nas: Params have no explanation.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
```diff
@@ -16,21 +21,95 @@
...
+
+    /// Registers a new webserver configured with one or more HTTP services
+    /// attached to a set of ports.
+    ///
+    /// @param services A map from port number to a combination of path and
+    ///     one or more HTTP services. This webserver is provided for reference,
+    ///     but must not be started or stopped independently. This server will
+    ///     be started and stopped with all other servers via the [#startAll]
+    ///     and [#stopAll] methods.
+    /// @return a result containing a webserver and a set of all configured ports.
```
ata-nas, line 84: Nit: no `@param` for `commonSocketValues`

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
```diff
@@ -16,21 +21,95 @@
...
+    /// Registers a new webserver configured with one or more HTTP services
+    /// attached to a set of ports.
+    ///
+    /// @param services A map from port number to a combination of path and
+    ///     one or more HTTP services. This webserver is provided for reference,
+    ///     but must not be started or stopped independently. This server will
+    ///     be started and stopped with all other servers via the [#startAll]
+    ///     and [#stopAll] methods.
+    /// @param http2Config
+    /// @param socketOptions
```
ata-nas: Nit: no explanation on these params.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
```diff
@@ -167,4 +188,37 @@ private void recordRequest(@NonNull final String result) {
...
+                .maxConcurrentStreams(webServerHttp2Config.maxConcurrentStreams())
+                .maxEmptyFrames(webServerHttp2Config.maxEmptyFrames())
+                .maxFrameSize(webServerHttp2Config.maxFrameSize())
+                .maxHeaderListSize(webServerHttp2Config.maxHeaderListSize())
+                .maxRapidResets(webServerHttp2Config.maxRapidResets())
+                .rapidResetCheckPeriod(Duration.ofMillis(webServerHttp2Config.rapidResetCheckPeriod()))
+                .build();
+    }
+
+    private static SocketOptions getSocketOptions(final HealthConfig healthConfig, final Configuration configuration) {
```
ata-nas: `configuration` seems to be unused.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
```diff
@@ -74,13 +84,24 @@ public void init(BlockNodeContext context, ServiceBuilder serviceBuilder) {
...
+        // Construct a map from port to services
+        TreeMap<Integer, ServiceWithPath[]> servicesToRegister = new TreeMap<>();
+        ServiceWithPath healthService = new ServiceWithPath(
+                HEALTHZ_PATH,
+                httpRules -> httpRules.get(LIVEZ_PATH, this::handleLivez).get(READYZ_PATH, this::handleReadyz));
+        ServiceWithPath statusService =
+                new ServiceWithPath(STATUSZ_PATH, httpRules -> httpRules.get("*", this::handleStatusz));
+        servicesToRegister.put(port, new ServiceWithPath[] {healthService, statusService});
+        // register a new web server
+        serviceBuilder.registerHttpNewServer(servicesToRegister, webserverHttpConfig, socketConfig, commonSocketValues);
```
ata-nas, line 104: Do (should) we care about the result here?

block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
```diff
@@ -3,20 +3,46 @@
...
-}
+        @Loggable @ConfigProperty(defaultValue = "40983") Integer port,
+        @Loggable @ConfigProperty(defaultValue = "32768") @Min(32768) @Max(Integer.MAX_VALUE) int socketSendBufferSizeBytes,
+        @Loggable @ConfigProperty(defaultValue = "65536") @Min(32768) @Max(Integer.MAX_VALUE) int socketReceiveBufferSizeBytes,
+        @Loggable @ConfigProperty(defaultValue = "100") int maxTcpConnections,
+        @Loggable @ConfigProperty(defaultValue = "1") @Min(1) @Max(60) int idleConnectionPeriodMinutes,
+        @Loggable @ConfigProperty(defaultValue = "2") @Min(1) @Max(1_440) int idleConnectionTimeoutMinutes,
+        @Loggable @ConfigProperty(defaultValue = "true") boolean tcpNoDelay,
+        @Loggable @ConfigProperty(defaultValue = "1_024") int backlogSize,
+        @Loggable @ConfigProperty(defaultValue = "2_048") int writeQueueLength) {}
```
ata-nas: Nit: for improved readability, values over 1k can be delimited with `_`.
```suggestion
        @Loggable @ConfigProperty(defaultValue = "40_983") Integer port,
        @Loggable @ConfigProperty(defaultValue = "32_768") @Min(32768) @Max(Integer.MAX_VALUE) int socketSendBufferSizeBytes,
        @Loggable @ConfigProperty(defaultValue = "65_536") @Min(32768) @Max(Integer.MAX_VALUE) int socketReceiveBufferSizeBytes,
        @Loggable @ConfigProperty(defaultValue = "100") int maxTcpConnections,
        @Loggable @ConfigProperty(defaultValue = "1") @Min(1) @Max(60) int idleConnectionPeriodMinutes,
        @Loggable @ConfigProperty(defaultValue = "2") @Min(1) @Max(1_440) int idleConnectionTimeoutMinutes,
        @Loggable @ConfigProperty(defaultValue = "true") boolean tcpNoDelay,
        @Loggable @ConfigProperty(defaultValue = "1_024") int backlogSize,
        @Loggable @ConfigProperty(defaultValue = "2_048") int writeQueueLength) {}
```

block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
```diff
@@ -3,20 +3,46 @@
...
+///     on the server socket
+/// @param writeQueueLength the number of buffers queued for write operations
+// spotless:off - long annotations on record components must stay on one line
 @ConfigData("health")
 public record HealthConfig(
-        // spotless:off
-        @Loggable @ConfigProperty(defaultValue = ConfigProperty.NULL_DEFAULT_VALUE) Integer port) {
-        // spotless:on
-}
+        @Loggable @ConfigProperty(defaultValue = "40983") Integer port,
```
ata-nas: Documentation states that we enforce the range `1_024 - 65_535` for port number, but the health config has no min/max annotations.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/RecordingServiceBuilder.java
```diff
@@ -0,0 +1,341 @@
...
+ *       port of every such registration (a {@code null} port resolves to the default port).
+ *   <li>Every {@link #registerHttpNewServer} call (either overload) creates one <b>additional</b>
+ *       web server, numbered <b>#2, #3, …</b> in the order the calls are made (across both
+ *       overloads). Its port set is the key set of the supplied services map.
+ * </ul>
+ *
+ * <p>The default port is {@link #DEFAULT_GENERAL_PORT} = {@code 131072}, which is deliberately larger
+ * than the maximum valid TCP port (65535). Because this fixture never binds a socket, using an
+ * out-of-range sentinel keeps recorded ports from ever being confused with a real one.
+ */
```
ata-nas: Nit: md javadoc?

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
```diff
@@ -2,77 +2,251 @@
...
+    private final Http2Config http2Config;
+    private final SocketOptions socketOptions;
+    private WebServerResult generalWebserver;
+    private final LinkedHashSet<WebServerResult> additionalWebservers;
+
+    public ServiceBuilderImpl(
+            final ServerConfig serverConfig, final Http2Config http2Config, final SocketOptions socketOptions) {
+        this.serverConfig = serverConfig;
+        this.http2Config = http2Config;
+        this.socketOptions = socketOptions;
```
ata-nas, line 50: Nit: null checks to fail fast seems like a good idea.

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
```diff
@@ -2,77 +2,251 @@
...
+            Set<Integer> ports,
+            Map<Integer, HttpRouting.Builder> httpBuilders,
+            Http2Config http2Config,
+            SocketOptions socketOptions,
+            CommonSocketValues socketCommon,
+            PbjConfig pbjConfig) {
+        // Override the default message size in PBJ
+        final var grpcBuilders = grpcRoutingBuilders();
+        final Iterator<Integer> portIterator = ports.iterator();
+        final int primaryPort = portIterator.next();
```
ata-nas, line 169: Is primary port always the first one?

ata-nas (commented):
Nits, cleanup suggestions and some questions, nothing blocking.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
```diff
@@ -16,21 +21,95 @@
  * and pass it explicitly; plugins that are happy sharing the default port simply pass {@code null}.
  */
 public interface ServiceBuilder {
```
ivannov, line 21: Why do we need to distinguish between regular and additional servers? If it is because the regular server's config comes via constructor and the additional ones' via parameters, can't we abstract away this implementation detail? And simplify the API, which looks quite asymetric right now

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
```diff
@@ -4,7 +4,12 @@
...
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import io.helidon.common.socket.SocketOptions;
+import io.helidon.webserver.WebServer;
 import io.helidon.webserver.http.HttpService;
+import io.helidon.webserver.http2.Http2Config;
+import java.util.Set;
+import java.util.TreeMap;
 
 /**
```
ivannov: NIT: You changed the javadoc style of all the methods, but didn't change the class level one

block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
ivannov: The new properties should be documented in configuration.md

On a more general note (and not for this PR and task), as these properties mostly duplicate the values in the server config, can we think about _inheritance_ mechanism between plugin server configuration and the server config.

I don't mean Java level inheritance (which is not possible anyway). And yes, I know this contradicts with my earlier comment about the design decision of having a regular and additional servers. 

WDYT?

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
ivannov: The `final Configuration configuration` parameter seems unused

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/RecordingServiceBuilder.java
```diff
@@ -0,0 +1,341 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.stream.Stream;
+import org.hiero.block.node.spi.ServiceBuilder;
+
+/**
```
ivannov: NIT: this should be the markdown style javadoc

ivannov (commented):
First pass with questions for understanding and some NITs

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: Deliberately moved down, the order prevents some interesting races in the Kubernetes layer.

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
jsync-swirlds: Null is allowed because that's how a plugin requests to use the "default" port.
That's not new, it's how we setup multiport.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
jsync-swirlds: The star is a regex value.
Mapping is `/statusz/*` going to a method.

block-node/spi-plugins/build.gradle.kts
jsync-swirlds: Not added, just reordered from two lines above.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
jsync-swirlds: I tried to avoid changing all of the javadoc, only where I was modifying anyway.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
jsync-swirlds: It's not "regular" versus "additional"
It's the _one_ general server versus any "additional" servers.

The general server has a singular configuration.
Each additional server has unique configuration.

Ideally we won't have many additional servers, as they're expensive in terms of RAM and resources.

The functionality is asymmetric, hence the API reflects that.

block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
jsync-swirlds: These are intended to be unique and different from server config.
The entire purpose of having a separate server is to change these values. We do not want to use "server" configuration here at all.

We considered configuration inheritance when developing the configuration library (and it's been considered in many other configuration libraries). The assessment always, to my knowledge, ends with an absence of viable use cases.

jsync-swirlds: Adjusted, and rearranged.  Also, missing min/max annotations added and a few limits adjusted.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
jsync-swirlds: Not as much as I expected to.
The ServiceBuilder implementation is responsible for start/stop, and there isn't much else we'd do with the return.

In _tests_ it's super useful, but not so much here.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
jsync-swirlds: It was needed before I removed all use of ServerConfig.
Will remove.

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
jsync-swirlds: Fixed.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
jsync-swirlds: Fixed.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
jsync-swirlds: Fixed.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
jsync-swirlds: Fixed.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
jsync-swirlds: Fixed.

block-node/health/src/main/java/org/hiero/block/node/health/HealthConfig.java
jsync-swirlds: Fixed.

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
jsync-swirlds: There is very little value to preemptive NPE in general.  These will fail with better messages and within the same method if they're null.

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
jsync-swirlds: Yes, we construct it that way.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/ServerConfig.java
jsync-swirlds: Fixed.

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ServiceBuilder.java
jsync-swirlds: I tried to limit javadoc swaps.
I've updated this one.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
jsync-swirlds: Removed.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/RecordingServiceBuilder.java
jsync-swirlds: Fixed.

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/RecordingServiceBuilder.java
jsync-swirlds: Done.

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
ata-nas: Yeah, as noted this is from the past. Was wondering if we wanted to possibly improve this. As noted, non-blocking and optional.

block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
ata-nas: I see, thanks for the explanation.

block-node/spi-plugins/build.gradle.kts
ata-nas: I wonder if we can simply drop this. In some other module info (verification) it was also present, but after removing it, everything is working as expected and we see no gradle check issues.

ata-nas (dismissed):
LG, unblocking, but seems like some CI is failing.

block-node/spi-plugins/build.gradle.kts
jsync-swirlds: Unfortunately we've found it's required by some of the CN libraries we use.
We get away with removing it in some cases because it's brought in here.

jsync-swirlds:
Note, Chart tests failing due to internal API change because those tests load Block Node and plugins from different versions.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
jsync-swirlds, line 1: The fixes in this file (and other Suites changes) are to improve stability and compatibility because the suites kept failing in this PR due to that instability.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/grpc/positive/PositiveServerAvailabilityTests.java
jsync-swirlds, line 1: Changes here to ensure suites correctly handle the health plugin using a different port.

ata-nas (approved):
LG

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
```diff
@@ -2,77 +2,251 @@
...
+    private CommonSocketValues newValuesFromConfig(final ServerConfig cfg) {
+        return new CommonSocketValues(
+                cfg.backlogSize(),
+                cfg.writeQueueLength(),
+                cfg.maxTcpConnections(),
+                cfg.idleConnectionPeriodMinutes(),
+                cfg.idleConnectionTimeoutMinutes());
+    }
+
+    private void configureSocket(
```
berryware, line 207: Non-blocking. configureSocket and configureListenerSocket could be unified to one generic method.

berryware (approved):
Looks good

block-node/app/src/main/java/org/hiero/block/node/app/ServiceBuilderImpl.java
jsync-swirlds: Added to #3223 for follow up.

PR #3219, opened 2026-07-14
feat: sidecar integrity validation for WRB command (fixes #3196)

### LINKED ISSUE #3196: WRB pipeline embeds sidecar files without verifying their SHA-384 against the signed record_file_contents metadata
## Problem

The Wrapped Record Block (WRB) production and validation pipeline reads sidecar file bytes and embeds them into the wrapped Block **without ever comparing their SHA-384 against the hashes listed in the record file's signed `record_file_contents.sidecars[].hash` metadata**.

The record file's RSA signature covers the sidecar hash list, but nothing in the wrapping or validation code cross-checks the sidecar bytes themselves against that list. If sidecar bytes are swapped anywhere between the recordstream store and the produced WRB, RSA validation still passes and the swap goes undetected downstream.

## Evidence

**Wrap path — RSA check happens, sidecar hash check does not**

`tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/ToWrappedBlocksCommand.java`

- L597: `unparsed.parse()` produces a `ParsedRecordBlock`.
- L612-626: record file RSA signature verified.
- L734: `RecordBlockConverter.toBlock(effectiveBlock, ...)` embeds sidecars into the Block.
- `ParsedRecordBlock.validate()` (see below) is **never invoked** in this pipeline.

**Sidecars are handed straight into the wrapped Block**

`tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/model/parsed/RecordBlockConverter.java`

L115:

```java
final RecordFileItem recordFileItem =
    new RecordFileItem(recordFileTimestamp, recordStreamFile, recordBlock.sidecarFiles(), amendments);
```

`recordStreamFile` includes the signed `sidecars[].hash` list; `recordBlock.sidecarFiles()` is the raw bytes. Both go into the Block, no comparison between them.

**The check exists, but only in a code path that isn't called**

`tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/model/parsed/ParsedRecordBlock.java`, lines 145-163 (`validate()`):

```java
MessageDigest digest = sha384Digest();
for (SidecarFile sidecarFile : sidecarFiles) {
    SidecarFile.PROTOBUF.toBytes(sidecarFile).writeTo(digest);
    final byte[] sidecarHash = digest.digest();
    boolean found = false;
    for (var sidecarMetadata : recordFile.recordStreamFile().sidecars()) {
        if (Arrays.equals(sidecarMetadata.hashOrThrow().hash().toByteArray(), sidecarHash)) {
            found = true;
            break;
        }
    }
    if (!found) {
        throw new ValidationException(\"Sidecar file hash not found...\");
    }
}
```

**A second audit confirmed no production caller anywhere**

`grep -rnE \"\\.validate\\(\\)\" tools-and-tests/tools/src/main/java/` returns exactly one hit — the TODO comment string in `LiveSequential.java:622` itself. `DownloadLive2.java`, `LiveDownloader.java`, `DownloadDayLiveImpl.java`, `DownloadDayImplV2.java` — none of them invoke the sidecar hash check. `validateBlockHashes` (which the live-download paths do call) verifies the record-file hash chain, not sidecars.

**Validation suite skips sidecar integrity entirely**

Searched `tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/` for `sidecar` — zero results. Required-items, structure, chain-continuity, and record-file signature checks all present; sidecar-hash verification is not.

**Pre-existing TODO already flags this — but its own claim is stale**

`tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java` around L622-624:

```
// TODO: download-live2 also calls UnparsedRecordBlockV6.validate() which verifies sidecar
// SHA-384 hashes against the record file metadata. This pipeline only does MD5 at download
// time + RSA verification in the wrap thread. Consider adding sidecar hash verification.
```

The TODO's claim that `download-live2` calls `.validate()` is inaccurate — `DownloadLive2.java` has no `.validate()` invocation. The TODO should be removed as part of this work.

**BN side is not compensating either**

Searched `block-node/` for `sidecar` — the only hits are in `BadBlockDumper.java` referring to a diagnostic JSON meta-file, unrelated to record-file sidecars. No sidecar-integrity verification on the BN side.

## Impact

- The record file's RSA signature ties the *manifest* (including sidecar hash list) to the network. It does **not** tie the sidecar bytes we actually embed to that manifest.
- Any bytes-vs-hash divergence at production time is silently baked into the WRB, ships to the BN, and downstream consumers accept it.
- Existing WRBs already carry both the signed sidecar hash list (inside `recordStreamFile`) and the raw sidecar bytes, so retroactive verification is possible — no re-wrap is required, only a validation-side check.
### Re-wrap from genesis is *not* required

Adding the check is a read-side operation over already-produced artifacts (one SHA-384 per sidecar per WRB, on demand). Re-wrapping — re-reading source record files, re-parsing sidecars, re-signing, re-emitting Blocks — would only be part of the response to an **incident finding** (i.e. running the validator against historical WRBs reveals a meaningful fraction have byte-vs-hash divergence). Shipping this fix does not itself require reproducing any WRB.

## Proposed fix

Extract a focused static helper that operates directly on the wrapped Block, then invoke it at every point that produces or consumes a WRB. No re-wrap of existing WRBs needed — they already carry everything the check requires.

**1. Extract `SidecarIntegrityValidator.validate(Block wrapped)`**

Pull the loop currently sitting dead inside `ParsedRecordBlock.validate()` into a dedicated static helper that takes a wrapped `Block` and walks its `RecordFileItem.sidecarFiles()` against `RecordFileItem.recordStreamFile().sidecars()[].hash`. Throw on any mismatch or any sidecar bytes with no corresponding hash entry.

The wrapped Block is the honest boundary — that's the artifact that leaves this pipeline and ships to BNs/consumers, so verifying integrity on the wrapped shape (rather than on the intermediate `ParsedRecordBlock`) is what matters. The Block already contains both the signed hash list and the raw sidecar bytes, so the helper is fully self-contained.

**2. Backfill the check into every producer path**

- **`ToWrappedBlocksCommand`** — call `SidecarIntegrityValidator.validate(wrapped)` right after the RSA signature check (~L626) and before the WRB is written to disk / streamed. Fail loud on production of a bad WRB.
- **`LiveSequential`** — call it inside the wrap thread on each produced Block, delete the stale TODO at L622.

**3. Backfill the check into the validation suite**

Under `tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/`, add a `SidecarIntegrityCheck` that hands each read-back Block to the same helper. Retroactive coverage for existing WRBs stored on disk, streamed from BNs, or in the recordstream.

**4. Fold the dead code path into shared code**

`ParsedRecordBlock.validate()` can either delegate to the same helper on the pre-wrap inputs, or be deleted if nothing else depends on it. Also remove the stale TODO in `LiveSequential.java` L622-624 since the check now has a real home.

## Test plan

- Unit test: build a Block with valid sidecar+hash pair → validator passes.
- Unit test: build a Block whose sidecar bytes differ from the signed hash list → validator throws with a clear diagnostic (block number, sidecar index).
- Unit test: build a Block whose sidecar has no matching hash entry in the metadata list → validator throws.
- Wire the validator into `ToWrappedBlocksCommand` and `LiveSequential` and confirm existing integration tests still pass (i.e. the check is a no-op on legitimate inputs from the recordstream).
- Add the validator to the validation suite and run against a representative WRB corpus to confirm no false positives against production data.

## Related

- `ParsedRecordBlock.validate()` — L145-163, existing implementation of the check (to be lifted into the shared helper)
- TODO in `LiveSequential.java` L622-624 — to be removed as part of this work
- Discussion in Slack (Joseph Sinclair) that surfaced the claim

rockysingh:
## Summary

Closes issue #3196. The WRB (Wrapped Record Block) pipeline was reading sidecar file bytes and embedding them into wrapped blocks **without ever comparing their SHA-384 against the hashes listed in the record file's signed `record_file_contents.sidecars[].hash` metadata**. The RSA signature covers the manifest but not the sidecar bytes; a byte-vs-hash divergence between the recordstream store and the produced WRB was silently baked in and accepted downstream.

This PR closes the gap at both the producer side (wrap-time enforcement) and the consumer side (read-time verification, no re-wrap required for historical archives). It also adds a targeted `blocks find-block` inspection command used during the mainnet investigation of real failures.

## What's new

**Producer-side (nothing bad ships from now on):**
- `SidecarIntegrityValidation.validateSidecars(sidecarFiles, sidecarMetadatas, blockNumber)` shared static helper.
- Wired into `ToWrappedBlocksCommand` right before `RecordBlockConverter.toBlock(...)`. On mismatch the wrap aborts, address book is saved, graceful shutdown fires.
- Wired into `LiveSequential`'s wrap thread at the same point. Also deleted the stale L622-624 TODO, audit confirmed its claim about `download-live2` calling the check was inaccurate.

**Consumer-side (retroactive coverage for archives already on disk / BNs):**
- Registered as a stateless parallel validation in `ValidateBlocksCommand`, gated behind `--skip-sidecar-integrity` for parity with the other skip flags.
- New dedicated `blocks validate-sidecars` subcommand with a parallel decomp + validate pipeline, progress bar, and per-failure diagnostics.

**Investigation tooling:**
- New `blocks find-block <blockNumber> <paths...>` subcommand. Locates a block inside WRB zip archives (or directories), prints block header, signed sidecar metadata, and each embedded sidecar's SHA-384. Also prints each `TransactionSidecarRecord`'s internal `consensus_timestamp` so investigators can tell whether an orphan sidecar's content belongs to the flagged block or was misattributed from a different block period. Supports `--extract-sidecars <dir>` for offline byte-level diffs and `--json` for full block dumps.

## Diagnostic classifier

The failure message now classifies each discrepancy so an investigator can tell three modes apart:

- **TAMPERED or EXTRA**: a sidecar with no matching hash in the signed list. Bytes altered or a rogue sidecar was attached.
- **MISSING**: a signed hash with no matching sidecar bytes in the block. Sidecar was dropped somewhere in the pipeline.

Example diagnostic:
```
Block 12345 sidecar integrity failed:
  sidecars in block:    3
  signed hash entries:  3
  discrepancies:
    - sidecar #1 SHA-384 abc123... -> no matching hash in signed metadata (TAMPERED or EXTRA)
    - signed hash #2 SHA-384 def456... -> no matching sidecar file in block (MISSING)
```

## Standalone commands UX

```
java -jar tools-all.jar blocks validate-sidecars [--threads N] [--prefetch N]
                                                 [--fail-fast] [--no-progress]
                                                 <files-or-dirs-or-zips>...

java -jar tools-all.jar blocks find-block <blockNumber> <files-or-dirs-or-zips>...
                                          [--extract-sidecars <dir>] [--json]
```

- Expands whole-zip archives before iterating so both single-block and container zips work.
- `validate-sidecars`: parallel worker pool (default: `cores - 1`), bounded queue, live progress bar with rate + ETA.
- Every failure prints as it happens; summary lists all failing block numbers so investigators don't have to grep.
- Exit codes: 0 all-pass, 1 sidecar failures, 2 I/O errors.

## Why no re-wrap is required

Every existing WRB already carries both sides of the comparison the validator performs: the signed hash list inside `recordStreamFile.sidecars[]` and the raw sidecar bytes inside `sidecarFileContents`. Running the validator against a historical archive gives a truthful answer today, no reproduction needed.

Re-wrapping would only enter the picture as an incident response if the validator surfaces bad blocks, see the incident-triage flow in issue #3196's discussion.

## Tests

11 unit tests in `SidecarIntegrityValidationTest`:
- Empty sidecar list passes
- No RecordFile item passes
- Single matching sidecar passes
- Unmatched sidecar hash fails with classifier
- Sidecar with no `recordFileContents` fails
- Shared helper: empty / single-match / multi-match all pass
- Shared helper: extra unmatched metadata flagged as MISSING
- Shared helper: metadata entries with no hash field are skipped
- Shared helper: MISSING failure mode surfaces with correct diagnostic counts

## Real-world validation

Command has been run against mainnet, testnet, and previewnet WRB archives. Testnet and previewnet: clean. Mainnet (97,630,000 blocks): 3 blocks flagged (52,333,943 / 58,446,161 / 62,113,066), all with the same "EXTRA sidecar" pattern (1 sidecar in block, 0 signed hash entries).

The `find-block` command was subsequently used to inspect each WRB. Root cause confirmed end-to-end: a small minority of CN nodes (1 or 2 per block) wrote orphan sidecar files under a record file's timestamp that contained transactions from a much later block (drift of 33 min to 13 hours). Consensus record files across the majority declared no sidecar; the orphan bucket files were nevertheless swept into the WRB by the wrap-time file matcher. All three failures fully explained. Wrap-side follow-up will consult the signed manifest before accepting sidecar files.

## Test plan

- [x] Local unit tests green (`./gradlew :tools:test --tests "*SidecarIntegrityValidationTest*"`)
- [x] Local module + spotless checks green
- [x] Shadow jar smoke-tested; commands visible in `blocks --help`, `blocks validate-sidecars --help`, `blocks find-block --help`
- [x] End-to-end run against mainnet (97.6M blocks), testnet, and previewnet WRB archives
- [x] `find-block` verified against real mainnet WRBs; SHA-384 output matches bucket orphan sidecars

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/FindBlockCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/FindBlockCommand.java
@@ -1,0 +1,325 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.block.tools.blocks;
+
+import static org.hiero.block.tools.utils.Sha384.sha384Digest;
+
+import com.hedera.hapi.block.stream.Block;
+import com.hedera.hapi.block.stream.BlockItem;
+import com.hedera.hapi.block.stream.RecordFileItem;
+import com.hedera.hapi.node.base.Timestamp;
+import com.hedera.hapi.streams.SidecarFile;
+import com.hedera.hapi.streams.SidecarMetadata;
+import com.hedera.hapi.streams.TransactionSidecarRecord;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.security.MessageDigest;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+import org.hiero.block.tools.blocks.model.BlockZipsUtilities;
+import org.hiero.block.tools.blocks.model.BlockZipsUtilities.BlockSource;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Help.Ansi;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.Parameters;
+
+/**
+ * Locate a specific block by number inside one or more WRB zip archives (or directories of them)
+ * and print an inspection report.
+ *
+ * <p>Primary use case is investigating sidecar integrity failures reported by
+ * {@code blocks validate-sidecars}. For each match the command prints:
+ *
+ * <ul>
+ *   <li>Where the block was found (zip path + entry name)</li>
+ *   <li>Block header summary (block number, timestamp, HAPI version)</li>
+ *   <li>RecordFileItem summary (creation time, amendment count)</li>
+ *   <li>Signed sidecar metadata from {@code record_file_contents.sidecars[]}</li>
+ *   <li>Embedded raw sidecars from {@code sidecar_file_contents[]}, with each sidecar's
+ *       computed SHA-384, whether it matches any signed metadata entry, and the internal
+ *       {@code consensus_timestamp} of each contained {@link TransactionSidecarRecord}
+ *       (which reveals which block period the sidecar bytes actually belong to when
+ *       investigating a mis-attribution)</li>
+ * </ul>
+ *
+ * <p>Optional {@code --extract-sidecars <dir>} writes each embedded sidecar's serialized bytes
+ * to disk so they can be diffed against bucket files or fed back through PBJ.
+ * Optional {@code --json} dumps the full parsed block as JSON.
+ *
+ * <p>Example:
+ * <pre>{@code
+ * java -jar tools-all.jar blocks find-block 62113066 /path/to/mainnet-wrbs
+ * }</pre>
+ */
+@Command(
+        name = "find-block",
+        description = "Locate a block by number inside WRB zip archives and print its contents with sidecar analysis",
+        mixinStandardHelpOptions = true)
+public class FindBlockCommand implements Runnable {
+
+    /** Maximum protobuf parse size in bytes (120 MB); matches ConvertToJson. */
+    private static final int MAX_BLOCK_SIZE_BYTES = 120 * 1024 * 1024;
+
+    @SuppressWarnings("unused")
+    @Parameters(index = "0", description = "Block number to locate")
+    private long blockNumber;
+
+    @SuppressWarnings("unused")
+    @Parameters(index = "1..*", description = "Block files, directories, or zip archives to search")
+    private Path[] paths;
+
+    @Option(
+            names = {"--extract-sidecars"},
+            description =
+                    "If set, write each embedded sidecar's raw serialized bytes to this directory as <block>_<idx>.rcd")
+    private Path extractSidecarsDir;
+
+    @Option(
+            names = {"--json"},
+            description = "Print the full parsed block as JSON to stdout")
+    private boolean printJson = false;
+
+    @Option(
+            names = {"--stop-on-first"},
+            description = "Stop after the first match instead of iterating every hit across the inputs")
+    private boolean stopOnFirst = true;
+
+    @Override
+    public void run() {
+        if (paths == null || paths.length == 0) {
+            System.err.println("Error: at least one file, directory, or zip must be given");
+            return;
+        }
+
+        final AtomicLong corruptZipCount = new AtomicLong(0);
+        final List<BlockSource> discovered = BlockZipsUtilities.findBlockSources(paths, corruptZipCount);
+        if (discovered.isEmpty()) {
+            System.err.println("Error: no block sources discovered in the given inputs");
+            return;
+        }
+
+        System.out.println(Ansi.AUTO.string("@|yellow Finding block " + blockNumber + "|@"));
+        System.out.println("  Sources discovered: " + discovered.size());
+        if (corruptZipCount.get() > 0) {
+            System.out.println("  Corrupt zips skipped: " + corruptZipCount.get());
+        }
+        System.out.println("  Expanding zip archives...");
+        final List<BlockSource> sources = BlockZipsUtilities.expandWholeZipSources(discovered);
+        System.out.println("  Total blocks visible: " + sources.size());
+        System.out.println();
+
+        final List<BlockSource> matches =
+                sources.stream().filter(s -> s.blockNumber() == blockNumber).toList();
+
+        if (matches.isEmpty()) {
+            System.err.println(Ansi.AUTO.string("@|red Block " + blockNumber + " not found in the given inputs.|@"));
+            System.exit(1);
+        }
+
+        System.out.println(Ansi.AUTO.string("@|green Found " + matches.size() + " match(es).|@"));
+        System.out.println();
+
+        int shown = 0;
+        for (final BlockSource source : matches) {
+            shown++;
+            printMatch(source, shown, matches.size());
+            if (stopOnFirst) {
+                if (matches.size() > 1) {
+                    System.out.println(Ansi.AUTO.string("@|yellow (--stop-on-first set; " + (matches.size() - 1)
+                            + " additional match(es) not shown; pass --stop-on-first=false to see all)|@"));
+                }
+                break;
+            }
+        }
+    }
+
+    private void printMatch(final BlockSource source, final int matchIndex, final int totalMatches) {
+        System.out.println("================================================================");
+        System.out.println("Match " + matchIndex + " of " + totalMatches);
+        System.out.println("  File:  " + source.filePath());
+        if (source.zipEntryName() != null) {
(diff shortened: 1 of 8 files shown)


jsync-swirlds (dismissed):

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/FindBlockCommand.java
```diff
@@ -0,0 +1,325 @@
...
+ * }</pre>
+ */
+@Command(
+        name = "find-block",
+        description = "Locate a block by number inside WRB zip archives and print its contents with sidecar analysis",
+        mixinStandardHelpOptions = true)
+public class FindBlockCommand implements Runnable {
+
+    /** Maximum protobuf parse size in bytes (120 MB); matches ConvertToJson. */
+    private static final int MAX_BLOCK_SIZE_BYTES = 120 * 1024 * 1024;
```
berryware: Appears to be unused

rockysingh: Good catch. Removed the constant in f4a5ed20d, plus the equally-dead `UNUSED` ArrayList placeholder and its orphaned `java.util.ArrayList` import while I was in there. The effective parse limit is already applied inside `BlockZipsUtilities.decompressAndParse` so nothing was relying on this.

rockysingh: good find fixed.

jsync-swirlds (approved):

berryware (approved):
Looks Good

PR #3222, opened 2026-07-15
fix: Cloud Storage Expanded plugin: blocks are permanently discarded when a single S3 upload attempt fails

### LINKED ISSUE #3207: Cloud Storage Expanded plugin: blocks are permanently discarded when a single S3 upload attempt fails
### Story Form

As the Cloud Storage Expanded plugin
I want a failed single-block S3 upload to be retried later instead of dropped after one attempt
So that blocks are not permanently and silently missing from S3-compatible storage due to transient upload errors

### Background

Observed on previewnet:

```
Block 1228455: S3 upload failed
org.hiero.block.node.cloud.storage.expanded.UploadException: Failed to complete multipart upload: key=0000/0000/0000/1228/455.blk.zstd, uploadId=...
Caused by: com.hedera.bucky.S3ResponseException: ...
Response status code: 404
Response body: <Error><Code>NoSuchUpload</Code><Message>The requested upload was not found.</Message></Error>
```

`SingleBlockStoreTask.call()` makes exactly one upload attempt per block beyond bucky's own internal transport-level retries. 404 responses such as `NoSuchUpload` are not in that retriable set, so they fail immediately on the first attempt. Both this class of non-retriable failures and an exhausted-retries failures surface identically as `UploadResult` status `S3_ERROR`/`IO_ERROR` -- the caller cannot and does not distinguish them.

`ExpandedCloudStoragePlugin.publishResult()` then sends `PersistedNotification(blockNumber, succeeded=false, ...)`, increments the `cloud_expanded_total_upload_failures` metric, and logs at INFO. That is the entire failure path today:  there is no retry, no dead-letter queue, and no local/ephemeral fallback storage. The block is permanently absent from this S3 archive unless an operator notices the log line and intervenes manually.

### Changes Required

- On any upload failure (`S3_ERROR` / `IO_ERROR`, whether retriable-exhausted or not), stage the compressed block bytes to a local directory (configurable path) keyed by block number, instead of discarding them after `publishResult`.
- Add a background retry mechanism (e.g. a periodic task) that re-attempts uploads for staged blocks with backoff, removing the staged file on success.
- Decide whether `PersistedNotification(succeeded=false)` should still fire immediately on first failure (current behavior) with a follow-up corrective signal once a retry succeeds, or be deferred until retries are exhausted.
- Bound local retry storage (max disk usage, max block age/attempts) and add metrics/alerting for "blocks pending local retry" and "blocks that exhausted retries and were ultimately dropped," so this failure mode is never silent again.
- On plugin `start()`, scan the staging directory for blocks queued before a restart so in-flight retries are not lost across restarts.
- Add tests covering: non-retriable failure (e.g. 404), exhausted-retries failure, eventual-retry-success, and local-storage-bound-exceeded paths.

ivannov:
### Deferred failure notification instead of immediate `succeeded=false`

`publishResult()` no longer sends `PersistedNotification(succeeded=false)` on the first upload failure. When retry is enabled and staging succeeds, no notification is sent at all until the block either recovers or exhausts its retries; only a compression error, or a failure where staging itself was rejected, still reports immediately. This was a deliberate choice over the alternative (fire succeeded=false immediately, then a corrective succeeded=true later): BackfillPlugin re-fetches the block from a peer on succeeded=false.

### `RetryStagingManager`: on-disk queue keyed by block number

Each failed upload's already-compressed bytes are written as a `{blockNumber}.blk.zstd` blob plus a `{blockNumber}.meta.properties` sidecar (objectKey, storageClass, blockSource, attempt count, first-staged time, next-eligible time) under a configurable directory. `loadExisting()` rebuilds the in-memory index from these files on `start()`, so staged blocks survive a process restart; any sidecar or blob whose counterpart is missing, or whose filename or contents cannot be parsed, is logged and deleted rather than aborting startup.

### Backoff and exhaustion bounds

`recordFailure()` grows the retry delay exponentially (`min(baseBackoff << (attempts-1), maxBackoff)`) and drops a block once it exceeds `retryMaxAttempts` or `retryMaxAgeHours`, whichever comes first. At the chosen defaults (30s base, 900s cap, 20 attempts, 6h age), attempt count is the bound that actually binds in practice (~4h of cumulative backoff).

### Background retry tick runs independently of the live-ordering machinery

`retryStagedBlocks()` (invoked on a dedicated scheduled executor from `ThreadPoolManager`) and `RetryUploadTask` do not go through `completionService`/`pendingPublish`, which exist specifically to keep the live block stream delivered in monotonically increasing order. Retries are out-of-band corrections for blocks that already passed verification, so that ordering guarantee does not apply to them. A `retryInFlight` set guards against the tick submitting a second concurrent attempt for a block whose previous retry has not finished yet.

## Related Issue(s)

Closes #3207


block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -186,6 +226,27 @@ public void start() {
...
         metricsHolder = Objects.requireNonNull(MetricsHolder.createMetrics(metricRegistry));
+        if (config.retryEnabled()) {
+            try {
+                Files.createDirectories(config.retryStagingDirectoryPath());
+            } catch (final IOException e) {
+                LOGGER.log(
+                        WARNING,
+                        "Failed to create retry staging directory {0}; retry feature may not work",
+                        config.retryStagingDirectoryPath(),
+                        e);
```
ata-nas: We are unable to create the root, but still move forward saying things may not work. Is it more correct to continue and simply not using the retry staging, or marking unhealthy? I do not think we want to startup in a state where "we may not work".

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -123,6 +145,22 @@ public class ExpandedCloudStoragePlugin implements BlockNodePlugin, BlockNotific
...
+    /// Constructed in {@link #init} from config.
+    private RetryStagingManager stagingManager;
+
+    /// Scheduled tick that scans {@link #stagingManager} for blocks due for another retry attempt.
+    /// `null` when {@link ExpandedCloudStorageConfig#retryEnabled()} is `false`.
+    private ScheduledExecutorService retryScheduler;
+
+    /// Block numbers with a retry attempt currently in flight, so the scheduled tick does not
+    /// submit a second concurrent retry for the same block before the first one completes.
+    private final Set<Long> retryInFlight = ConcurrentHashMap.newKeySet();
```
ata-nas: Maybe a concurrent skip list set would be better? Also, for performance reasons, it is better to not declare interfaces, but actual implementations that we instantiate.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/PayloadIterator.java
ata-nas, line 1: We really need to change the bucky API to include a single payload iterator, or to simply accept singe byte array for such uploads. This is quite inconvenient and error prone. I've seen this in a few places.
We can address this some other time, as the change is not small and unrelated to this issue.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
```diff
@@ -0,0 +1,400 @@
...
+        @NonNull
+        static StagedEntry fromProperties(
+                @NonNull final Properties props, final long blockNumber, @NonNull final Path source)
+                throws IOException {
+            final String objectKey = requireProperty(props, KEY_OBJECT_KEY, source);
+            final String storageClass = requireProperty(props, KEY_STORAGE_CLASS, source);
+            final BlockSource blockSource = parseBlockSource(requireProperty(props, KEY_BLOCK_SOURCE, source), source);
+            final int attempts = parseIntProperty(props, KEY_ATTEMPTS, source);
+            final long firstStagedEpochMs = parseLongProperty(props, KEY_FIRST_STAGED_EPOCH_MS, source);
+            final long nextEligibleEpochMs = parseLongProperty(props, KEY_NEXT_ELIGIBLE_EPOCH_MS, source);
```
ata-nas: We can proto serialize these instead. Seems like much less of a hassle.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
```diff
@@ -0,0 +1,400 @@
...
+                LOGGER.log(DEBUG, "Wrote retry staging files for block {0}.", key);
+                return entry;
+            } catch (final IOException e) {
+                LOGGER.log(INFO, "Failed to stage block {0} for retry; dropping.", key, e);
+                silentDelete(blobPath);
+                silentDelete(metaPath);
+                return null;
+            }
+        });
+        return staged.containsKey(blockNumber);
```
ata-nas: Race condition: Is it possible that 2 threads stage an entry with the same key? If so, compute if absent is safe, but the contains key check can change by the time we exit from compute if absent. The compute if absent will return a result we can compare.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
```diff
@@ -0,0 +1,400 @@
...
+                } catch (final NumberFormatException e) {
+                    LOGGER.log(
+                            DEBUG,
+                            "Retry sidecar {0} has a non-numeric block number in its filename; deleting.",
+                            metaPath);
+                    silentDelete(metaPath);
+                    continue;
+                }
+                final Path blobPath = blobPath(blockNumber);
+                if (!Files.isRegularFile(blobPath)) {
```
ata-nas: Note that isRegularFile check also returns false if the file simply does not exist. Something we should be mindful of.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
```diff
@@ -0,0 +1,400 @@
...
+                silentDelete(blobPath);
+                silentDelete(metaPath);
+                return null;
+            }
+        });
+        return staged.containsKey(blockNumber);
+    }
+
+    /// Scans {@link ExpandedCloudStorageConfig#retryStagingDirectoryPath()} and rebuilds the in-memory
+    /// index from any pre-existing sidecar/blob pairs. Called once from `start()` so blocks staged
```
ata-nas: What do "blob" and "sidecar" mean in the context of this staging manager?

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
```diff
@@ -0,0 +1,400 @@
...
+    /// before a restart are not lost. Any sidecar or blob file whose counterpart is missing is deleted
+    /// as an orphan.
+    void loadExisting() {
```
ata-nas: Lots and lots of individual try-catches in this method and in other places. This is generally a code smell.
Either have a try-catch that will be global for the whole method, or else, smaller try-catch scopes most likely need to be individual methods that return a result.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
```diff
@@ -0,0 +1,400 @@
...
+        // Computed atomically per block number (via ConcurrentHashMap#compute) so a concurrent stage()
+        // or unstage() for the same block cannot interleave with this read-modify-write. compute()
+        // returns the new mapping (or null once removed), which already tells us the outcome: null means
+        // the block was dropped (unknown block or retry limits exceeded), so no separate outcome variable
+        // is needed.
+        final StagedEntry result = staged.compute(blockNumber, (key, previous) -> {
+            if (previous == null) {
+                // Should not happen in practice, but treat an unknown block as already exhausted.
+                return null;
+            }
```
ata-nas: Judging by this comment, this event should not happen. Should we throw here, indicating that we have a bug we need to investigate instead? At least a WARNING log seems reasonable.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ata-nas, line 1: There seem to be some places where we could lose an upload's data being persisted here. I wonder, should we in this case simply fail with a failed persistence notification so that backfill will be able to request the block again?

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryUploadTask.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /// Re-uploads the staged compressed bytes to S3.
+    ///
+    /// @return the upload result (never `null`)
+    @Override
+    public UploadResult call() {
+        final long uploadStartNs = System.nanoTime();
+        try {
+            s3Client.uploadFile(objectKey, storageClass, new PayloadIterator(compressedBytes), CONTENT_TYPE);
+            LOGGER.log(TRACE, "Block {0}: retry upload succeeded for {1}", blockNumber, objectKey);
```
ata-nas, line 66: This TRACE log seems redundant. I believe we will increment a metric on success?

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryUploadTask.java
```diff
@@ -0,0 +1,83 @@
...
+                    false);
+        } catch (final UploadException e) {
+            LOGGER.log(DEBUG, "Block {0}: S3 retry upload failed", blockNumber, e);
+            return new UploadResult(
+                    blockNumber, UploadStatus.S3_ERROR, 0L, blockSource, System.nanoTime() - uploadStartNs, false);
+        } catch (final IOException e) {
+            LOGGER.log(DEBUG, "Block {0}: I/O error during retry upload", blockNumber, e);
+            return new UploadResult(
+                    blockNumber, UploadStatus.IO_ERROR, 0L, blockSource, System.nanoTime() - uploadStartNs, false);
+        }
```
ata-nas, line 105: Should we catch a runtime exception wildcard here and return upload status unknown error? This will still indicate the bug, so we will see it, but then the system can continue, we can consider this.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/SingleBlockStoreTask.java
```diff
@@ -135,40 +147,29 @@ public UploadResult call() {
...
-        public byte[] next() {
-            if (delivered) throw new NoSuchElementException();
-            delivered = true;
-            return payload;
-        }
+    /// Stages the compressed bytes for background retry if compression completed before the upload
+    /// failed. Returns `false` without staging when `compressed` is `null` (the `IOException` came from
+    /// compression itself, not the upload) or when {@link RetryStagingManager#stage} rejects the bytes.
+    private boolean stageForRetry(final byte[] compressed) {
+        return compressed != null
```
ata-nas: We can also check for compressed being an empty array. If it is, no need to schedule IMO.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/SingleBlockStoreTask.java
```diff
@@ -135,40 +147,29 @@ public UploadResult call() {
...
             LOGGER.log(WARNING, msg, e);
+            final boolean staged = stageForRetry(compressed);
             return new UploadResult(
-                    blockNumber, UploadStatus.S3_ERROR, 0L, blockSource, System.nanoTime() - uploadStartNs);
+                    blockNumber, UploadStatus.S3_ERROR, 0L, blockSource, System.nanoTime() - uploadStartNs, staged);
         } catch (final IOException e) {
             final String msg = "Block " + blockNumber + ": I/O error during upload";
             LOGGER.log(WARNING, msg, e);
+            final boolean staged = stageForRetry(compressed);
             return new UploadResult(
```
ata-nas, line 183: What happens when we fail to stage a failed upload? Do we fire a failed persistence notification? I think that should be the case so backfill can get the block again.

ata-nas (commented):
First pass. Mainly questions.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ivannov: Great catch, returning the result returned from `computeIfAbsent` instead of ignoring it

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ivannov: Added a WARNING log

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: So I was wondering how to treat this. In the persistence plugins we shutdown the block node in this case, but I thought it is very harsh.

So if we can't create the staging directory and the retry is enabled, when we try to store a block, we'll fail. `stage()` will return false and eventually we will fire the false persistence notification. Which will trigger the backfill.

So this is basically like retry enabled false.

Maybe I should change the warning log message. I would definitely not shutdown the block node, as retry is not a vital feature of this plugin

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/SingleBlockStoreTask.java
ivannov: If `staged` is `false`, the last parameter of the `UploadResult` will be false. Then in `ExpandedCloudStoragePlugin`'s `publishResult()` method we will hit the final `else` block, where we send persisted notification with `false`. So the backfill will trigger getting the block again in that case (as you suggested)

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ivannov: It's in the class javadoc:

> Each staged block is written as a `{blockNumber}.blk.zstd` blob plus a `{blockNumber}.meta.properties` sidecar under {@link ExpandedCloudStorageConfig#retryStagingDirectoryPath()}

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: The operations that we perform upon this field are add, contains and remove. They do not imply ordering. So I think that a hash-based structure is better here.

As for using the implementation, I would push back as well. It's a better style to define interfaces and I don't see any particular performance gains in the oposite

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ivannov: We had the same discussion about another object which we used for persisting data internal to the plugin, I think it was for the meta sidecars of the cloud storage archive plugin's temporary archives. And there it was agreed that proto is too much and we'll stick to manual parsing and constructing the JSON.

The only difference here is that we are using key-value pairs (Properties) and not JSON. I would stick to this self contained implementation rather than going to all the additional work with proto

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ivannov: There are actually two inner try-catches and one outer. The inner try-catches are intentional, they let the loop skip one bad file and continue with the rest. The outer try catch is for hard failure of the `Files.list()` method

So I wouldn't classify these as smells and clutter the class with a couple of one-line methods that would hide the try-catch inside.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryUploadTask.java
ivannov: Maybe it is redundant, I just followed the existing pattern in `SingleBlockStoreTask` (for consistency).

There is a metric that is incremented: `cloud_expanded_retry_success_total`. It's in the plugin class and is incremented in `processRetryResult`

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryUploadTask.java
ivannov: Catching RuntimeException and reporting an unkown error. As well as a warning a log to note for the bug

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ivannov: In all cases the sidecar does not have its own blob. So we should delete it. Added a comment to clarify that

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/SingleBlockStoreTask.java
ivannov: Implemented the proposed optimization

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/PayloadIterator.java
ivannov: You are completely right. This iterator should be Bucky internal implementation detail for cases when we have the whole content in memory. Added this task there: https://github.com/hashgraph/bucky/issues/54 and a TODO here

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ata-nas: Defining things as interfaces instead of actual runtime class does have a performance impact. This is something we've been careful about as of recently as it was pointed out by Cliff Click. As a matter of fact you will see in verification and publisher, we've started using the runtime implementation instead of declaring higher level.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ata-nas: That is ok, not against it. Mainly exploring options.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -316,34 +385,147 @@ private void processCompletedFuture(final Future<SingleBlockStoreTask.UploadResu
...
-            // BlockNodeApp shutdown (loadedPlugins order), so an upload that finishes during
-            // this plugin's final drain (see stop()) can find messaging already stopped.
-            // Nothing downstream can consume the notification at that point, so log and
-            // continue rather than letting this escape stop() and abort the app's shutdown
-            // sequence for every plugin still left to stop.
+        if (result.succeeded()) {
+            sendPersistedNotification(result.blockNumber(), true, result.blockSource());
+            metricsHolder.uploadsTotal().increment();
+            metricsHolder.uploadBytesTotal().increment(result.bytesUploaded());
+        } else if (result.stagedForRetry()) {
```
berryware, line 416: Should stagingManager.unstage(result.blockNumber()) be called here when result.succeeded?

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -316,34 +385,147 @@ private void processCompletedFuture(final Future<SingleBlockStoreTask.UploadResu
...
+    }
+
+    /// Applies the outcome of one background retry attempt: on success, clears staging and
+    /// publishes the deferred `succeeded=true` notification; on failure, records another attempt
+    /// and — only once retries are exhausted — publishes the deferred `succeeded=false`
+    /// notification.
+    private void processRetryResult(final StagedEntry entry, final SingleBlockStoreTask.UploadResult result) {
+        if (result.succeeded()) {
+            stagingManager.unstage(entry.blockNumber());
+            blockMessaging.sendBlockPersisted(
```
berryware: Can the sendPersistedNotification() helper be used here and below?

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
jsync-swirlds: A lot depends on code complexity and use. If the data is temporary, internal only, and never used by any other system it might be worth creating the code manually.
If the code complexity is higher than using protobuf definitions, then we'd prefer to use protobuf.

Another consideration is that while we can _manually_ create text formats, the only binary format we should is is protocol buffers, and if we want JSON parsing or automatic creation, then we must use PBJ JSON support, not add a third-party JSON library.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
jsync-swirlds: Using interfaces is OK in code that is not performance critical; part of that comes down to how often we're making calls through the interface; which imposes vcall overhead.

If this is only interacted with for a few hundred calls, the interface is OK.

`LinkedHashMap` is a good choice for a hash based map (which is not nearly as efficient as most believe) if we need to iterate over the map efficiently.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
jsync-swirlds: Retry and staging are definitely vital, but we do not want to be shutting down the node from a plugin (that method absolutely needs to go away).

This log is probably about the best we can do within reason.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
```diff
@@ -0,0 +1,403 @@
...
+        final String name = path.getFileName().toString();
+        return Long.parseLong(name.substring(0, name.length() - suffix.length()));
+    }
+
+    private static void silentDelete(@NonNull final Path path) {
+        try {
+            Files.deleteIfExists(path);
+        } catch (final IOException ignored) {
+            // best-effort cleanup
+        }
```
ata-nas: Instead of `silent`, we can consider `safe` delete. If we have issues deleting, it might be worth knowing why. Not sure, but something to consider.

jsync-swirlds: I agree, we should log something here at `DEBUG` level, if nothing else.
Probably not the exception (stack traces are a lot of detail), but at least block number and `ignored.toString` would be helpful.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: Yes, that's quite possible. A duplicate VerificationNotification for an already-staged block could succeed and leave a stale staged entry behind. Added the unstage() call

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: Yes, I switched both to use the helper. Thanks!

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ivannov: Added a DEBUG log with the path

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
ivannov: The solution with the properties file is internal-only, no cross-system use and no third party JSON library is used. Please confirm whether you are OK with that

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: `retryInFlight` is only used for add / remove / check, it is never iterated. So I think an ordered structure is not the most performant solution here

As for the call volume - I don't think it is over the hundreds. So I would leave the cleaner solution with the interface

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
jsync-swirlds: That should be fine.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryStagingManager.java
jsync-swirlds: No change needed here, but I want to note a couple things:

Code smell doesn't mean "wrong", it should indicate "look closely here; be certain it's necessary".
Even if it is the right way to implement this particular method, exception blocks that are tightly wrapped around code inside a method are always worth careful consideration. If possible, replacement with something other than exception handling is much better.

charts/block-node-server/values.yaml
```diff
@@ -259,6 +259,19 @@ blockNode:
...
+      enabled: true
+      # name of the volume within the statefulset
+      volumeName: "cloud-storage-expanded-retry-staging"
+      # should match CLOUD_STORAGE_EXPANDED_RETRY_STAGING_DIRECTORY_PATH, leave as is for default.
+      mountPath: "/opt/hiero/block-node/cloud-storage-expanded/retry-staging"
+      size: 5Gi
+      # Optionally add a storage class name if needed
+      # storageClass: "your-storage-class"
+      # Optionally specify an existing PVC name instead of creating via volumeClaimTemplate
+      # existingClaim: "my-cloud-storage-expanded-retry-staging-pvc"
```
jsync-swirlds: So, we have maintained a design goal for the cloud archive server to not have any local disk.
What led us to violate that invariant and create local storage?

We can hold blocks in memory for a bit (perhaps up to 10 block times) to retry, but we shouldn't try add local storage to the expanded cloud plugin thinking that will enable more or more reliable buffering.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStorageConfig.java
```diff
@@ -58,7 +70,11 @@ public record ExpandedCloudStorageConfig(
         @Loggable @ConfigProperty(defaultValue = "") String regionName,
         @ConfigProperty(defaultValue = "") String accessKey,
         @ConfigProperty(defaultValue = "") String secretKey,
-        @Loggable @ConfigProperty(defaultValue = "60") @Min(1) int uploadTimeoutSeconds) {
+        @Loggable @ConfigProperty(defaultValue = "60") @Min(1) int uploadTimeoutSeconds,
+        @Loggable @ConfigProperty(defaultValue = "true") boolean retryEnabled,
+        @Loggable @ConfigProperty(defaultValue = "10") @Min(1) int retryIntervalSeconds,
+        @Loggable @ConfigProperty(defaultValue = "60") @Min(1) int retryMaxAgeSeconds,
+        @Loggable @ConfigProperty(defaultValue = "200") @Min(1) int retryMaxPendingBlocks) {
```
jsync-swirlds: Default should probably be smaller, perhaps 30.

ivannov: Changed to 30

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -308,36 +381,149 @@ private void processCompletedFuture(final Future<SingleBlockStoreTask.UploadResu
...
+    /// plugin's final drain (see {@link #stop()}) can find messaging already stopped. Nothing
+    /// downstream can consume the notification at that point, so log and continue rather than
+    /// letting this escape {@link #stop()} and abort the app's shutdown sequence for every plugin
+    /// still left to stop.
+    private void sendPersistedNotification(final long blockNumber, final boolean succeeded, final BlockSource source) {
+        try {
+            blockMessaging.sendBlockPersisted(new PersistedNotification(blockNumber, succeeded, 0, source));
+        } catch (final RejectedExecutionException e) {
+            LOGGER.log(
+                    WARNING,
```
jsync-swirlds: This is more likely INFO; the system would have to be shutting down, and this is expected.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -308,36 +381,149 @@ private void processCompletedFuture(final Future<SingleBlockStoreTask.UploadResu
...
+                        entry.blockNumber());
+            }
+        } else {
+            final RetryOutcome outcome = retryBuffer.recordFailure(entry.blockNumber());
+            if (outcome == RetryOutcome.EXHAUSTED) {
+                sendPersistedNotification(entry.blockNumber(), false, entry.blockSource());
+                metricsHolder.retryExhaustedTotal().increment();
+                metricsHolder.uploadFailuresTotal().increment();
+                LOGGER.log(
+                        WARNING,
```
jsync-swirlds: Warning generally means "the node may be unable to continue processing".
A single upload failure seems more akin to "Something is not right, but it is likely recoverable", which would be `INFO` level.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryBuffer.java
```diff
@@ -0,0 +1,185 @@
...
+        if (!buffered.containsKey(blockNumber) && buffered.size() >= config.retryMaxPendingBlocks()) {
+            LOGGER.log(DEBUG, "Retry buffer full; not buffering block {0}.", blockNumber);
+            return false;
+        }
+        // computeIfAbsent leaves an already-buffered block untouched and serializes this write
+        // against a concurrent recordFailure()/unstage() for the same block number.
+        final BufferedEntry result = buffered.computeIfAbsent(blockNumber, key -> {
+            final long now = System.currentTimeMillis();
+            return new BufferedEntry(key, compressedBytes, objectKey, storageClass, blockSource, 1, now, now);
+        });
```
jsync-swirlds: Rather than taking a "use the pre-existing value" approach, we might wish to prefer the _last_ offered bytes for a block, rather than the _first_; on the assumption that if we accepted the same block more than once; the later arrival may have some form of update or correction.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryBuffer.java
```diff
@@ -0,0 +1,185 @@
...
+            @NonNull final String storageClass,
+            @NonNull final BlockSource blockSource) {
+        if (!config.retryEnabled() || closed) {
+            return false;
+        }
+        // A duplicate for an already-buffered block must not count against the cap.
+        if (!buffered.containsKey(blockNumber) && buffered.size() >= config.retryMaxPendingBlocks()) {
+            LOGGER.log(DEBUG, "Retry buffer full; not buffering block {0}.", blockNumber);
+            return false;
+        }
```
jsync-swirlds: Instead of rejecting the new value, perhaps we should replace the old value?

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryBuffer.java
```diff
@@ -0,0 +1,185 @@
...
+    List<BufferedEntry> dueForRetry(@NonNull final Instant now) {
+        final long nowMs = now.toEpochMilli();
+        final List<BufferedEntry> due = new ArrayList<>();
+        for (final BufferedEntry entry : buffered.values()) {
+            if (entry.nextEligibleEpochMs() <= nowMs) {
+                due.add(entry);
+            }
+        }
+        return due;
+    }
```
jsync-swirlds: If we're exclusively comparing millisecond values, why use an Instant?

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryBuffer.java
```diff
@@ -0,0 +1,185 @@
...
+    /// Removes and returns every buffered entry. Call {@link #close()} first so a task racing
+    /// this can't add an entry that gets silently wiped by `clear()` without ever being returned.
+    ///
+    /// @return all entries that were buffered at the moment of the call
+    @NonNull
+    List<BufferedEntry> drainAll() {
+        final List<BufferedEntry> drained = new ArrayList<>(buffered.values());
+        buffered.clear();
+        return drained;
+    }
```
jsync-swirlds, line 239: Should we check `closed` at the start of this method, just in case, and return `null` if called without closing the buffer first?

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -308,36 +381,149 @@ private void processCompletedFuture(final Future<SingleBlockStoreTask.UploadResu
...
+                            s3Client,
+                            entry.objectKey(),
+                            entry.storageClass(),
+                            entry.blockSource())
+                    .call();
+            processRetryResult(entry, result);
+        } finally {
+            retryInFlight.remove(entry.blockNumber());
+        }
+    }
```
jsync-swirlds, line 543: OK, so this is weird.
Why are we _Executing_ a lambda, that just calls a local method that misuses a Callable by directly executing `.call`?

The entire purpose of the `Callable` interface is that the object of that class can be directly passed to the `ExecutorService.submit` method.

Core issues here:
1. Executor.execute often directly executes Runnables, it _may not_ create new threads. Here, we have an `ExecutorService`, but we're calling `execute` instead of `submit`.
   - We should call `submit` instead of `execute`.
1. We're directly calling `Callable.call` instead of just passing the `Callable` to `submit`.
   - Line 451 should be something like `Future<UploadResult> retryTask = virtualThreadExecutor.submit(new RetryUploadTask(...));`
1. We're not managing the Future(s) for tasks; we should be recording the Futures in a list and have a separate cleaning/management task that collects results and processes them.
   - `retryStagedBlocks` is already scheduled, so we can call a method to process all active Futures at the start of that call, _then_ submit new Futures (current code).

jsync-swirlds (commented):
Not done yet.
Sending items found so far.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: Addressed all three concerns / issues

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryBuffer.java
ivannov: Changed to evict the oldest value

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryBuffer.java
ivannov: A block that comes to the buffer replaces an existing one, preserving number of retries and age

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/RetryBuffer.java
ivannov: Changed the parameter to long

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -123,6 +145,31 @@ public class ExpandedCloudStoragePlugin implements BlockNodePlugin, BlockNotific
...
+    private final Set<Long> retryInFlight = ConcurrentHashMap.newKeySet();
+
+    /// `CompletionService` for async retry-upload tasks, separate from {@link #completionService}
+    /// since retries don't participate in {@link #pendingPublish} ordering.
+    private CompletionService<SingleBlockStoreTask.UploadResult> retryCompletionService;
+
+    /// Tracks which block number each outstanding retry {@link Future} belongs to, so
+    /// {@link #retryInFlight} can be cleared once that future is drained, regardless of outcome.
+    private final Map<Future<SingleBlockStoreTask.UploadResult>, Long> retryFutureBlockNumbers =
+            new ConcurrentHashMap<>();
```
jsync-swirlds: We shouldn't have multiple retries for a single block number, right?
Would it make sense to flip key and value here so we have:

```suggestion
    /// Tracks which block number each outstanding retry {@link Future} belongs to, so
    /// {@link #retryInFlight} can be cleared once that future is drained, regardless of outcome.
    private final Map<Long, Future<SingleBlockStoreTask.UploadResult>> retryFutureBlockNumbers =
            new ConcurrentHashMap<>();
```

Then we could use the same map for both retryInFlight (the Map ensures no duplicate keys) and futures (null for in flight but no current task).

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -308,36 +392,196 @@ private void processCompletedFuture(final Future<SingleBlockStoreTask.UploadResu
...
+    /// {@link #processCompletedFuture}.
+    private void processCompletedRetryFuture(final Future<SingleBlockStoreTask.UploadResult> completed) {
+        final Long blockNumber = retryFutureBlockNumbers.remove(completed);
+        try {
+            if (completed.isCancelled()) {
+                LOGGER.log(TRACE, "Retry upload task was cancelled during shutdown.");
+                return;
+            }
+            try {
+                processRetryResult(completed.get());
```
jsync-swirlds: If we extract the `get()` call and call `processRetryResult` outside the try, we can better ensure that failed persistence notifications get sent correctly (for instance, if we get an exception and perhaps exceed a retry limit).

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -308,36 +392,196 @@ private void processCompletedFuture(final Future<SingleBlockStoreTask.UploadResu
...
+            } catch (final InterruptedException e) {
+                Thread.currentThread().interrupt();
+            } catch (final ExecutionException e) {
+                metricsHolder.uploadFailuresTotal().increment();
+                LOGGER.log(WARNING, "Unexpected exception in retry upload task", e.getCause());
```
jsync-swirlds: This, again, is probably INFO.
It might help to re-read the [logging guidelines](https://github.com/hiero-ledger/hiero-block-node/blob/main/docs/logging-guidelines.md) for the block node.

jsync-swirlds (dismissed):
Much improved.
Just a few non-blocking suggestions.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ata-nas: +1 here, I think having the block number be the key is semantically easier for anyone who reads this.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -143,6 +190,8 @@ public ExpandedCloudStoragePlugin() {}
     public void init(@NonNull final BlockNodeContext context, @NonNull final ServiceBuilder serviceBuilder) {
         this.config = context.configuration().getConfigData(ExpandedCloudStorageConfig.class);
         this.blockMessaging = context.blockMessaging();
+        this.threadPoolManager = context.threadPoolManager();
+        this.retryBuffer = new RetryBuffer(config);
```
ata-nas: Do we need these if we have disabled retry? Feels like we should not instantiate things that will not be used.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -178,7 +227,18 @@ public void start() {
         }
         virtualThreadExecutor = Executors.newVirtualThreadPerTaskExecutor();
         completionService = new ExecutorCompletionService<>(virtualThreadExecutor);
+        retryCompletionService = new ExecutorCompletionService<>(virtualThreadExecutor);
```
ata-nas: Do we need this if we have disabled retry? Feels like we should not instantiate things that will not be used.

ata-nas (commented):
Posting just a couple of comments as this is being updated as I review. Will circle back for final review.

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: Good catch. Changed so that the service is created only when the retry is enabled

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: Guarded this creation as well

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: We handle all the limits (retry and timeout) in case of exceptions in the `RetryUploadTask`. Just in any case, if an `Error` escapes, I've added handling to the `ExecutionException` catch clause

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: OK, so I see two concerns here:

* The key-value setup of the second structre (`retryFutureBlockNumbers`) should be flipped
* We have a couple of structures: `retryInFlight` and `retryFutureBlockNumbers`. The first one is used only for the dedup

The reason to have both structures initially was that I was aiming at O(1) in both of the following situations:

* Removing a future when it was completed: here I remove it from the map keyed by that future
* Checking whether there is a task running for retrying a certain block (i.e. dedup) before starting it: here I looked up the set

So I removed the set and left the map. 

* If it is keyed by future, this means that removing an entry from the map will be O(1) and dedup - O(N)
* If it is keyed by number, this means that dedup for adding an entry to the map will be O(1) and removal - O(N)

If we look how often removing and dedup is called:

* removing is called upon every `handleVerification()` where we try to drain the completed tasks (if any)
* dedup is called called only on the scheduled method (`retryStagedBlock()`)

Based on the above analysis, I will:

* remove the dedup set
* leave the orientation of the map

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: Changed to INFO

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
jsync-swirlds: My point was more that we could handle all of the notification cases in one place rather than several locations; but perhaps that should be a follow-up...

ivannov: At the moment that place is the `RetryUploadTask`. There we don't handle only `Throwable`, which we do in the plugin with my latest change. I can move that there as well. But we will have to still catch the `ExecutionException` in the plugin

jsync-swirlds: This should probably be follow-up then, as I am probably not communicating the concepts clearly.

jsync-swirlds (dismissed):

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -300,44 +387,211 @@ private void processCompletedFuture(final Future<SingleBlockStoreTask.UploadResu
                 // here means something more severe (e.g. an Error) escaped the task — count it as
                 // a failure and log the root cause.
                 metricsHolder.uploadFailuresTotal().increment();
-                LOGGER.log(WARNING, "Unexpected exception in upload task", e.getCause());
+                LOGGER.log(INFO, "Unexpected exception in upload task", e.getCause());
```
Nana-EC, line 390: Q: should this not remain as a `WARNING` so it doesn't get lost?

docs/design/persistence/cloud-storage-expanded.md
```diff
@@ -316,21 +400,29 @@ classDiagram
         -virtualThreadExecutor: ExecutorService
         -pendingPublish: ConcurrentSkipListMap
         -metricsHolder: MetricsHolder
+        -threadPoolManager: ThreadPoolManager
+        -retryBuffer: RetryBuffer
+        -retryScheduler: ScheduledExecutorService
+        -retryInFlight: Set~Long~
```
Nana-EC: Outdated, should be `-retryFutureBlockNumbers: Map~Future, Long~`

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
```diff
@@ -300,44 +387,211 @@ private void processCompletedFuture(final Future<SingleBlockStoreTask.UploadResu
...
+        try {
+            processRetryResult(completed.get());
+        } catch (final InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (final ExecutionException e) {
+            metricsHolder.uploadFailuresTotal().increment();
+            LOGGER.log(INFO, "Unexpected exception in retry upload task", e.getCause());
+            // Still counts as an attempt
+            if (blockNumber != null && retryBuffer.recordFailure(blockNumber) == RetryOutcome.EXHAUSTED) {
+                sendPersistedNotification(blockNumber, false, BlockSource.UNKNOWN);
```
Nana-EC: Why `BlockSource.UNKNOWN`?
It's an RFH so should it be `BACKFILL` or are we noting UNKNOWN because the plugin can be deployed by the operator in any configuration in a Hiero network?

Nana-EC (dismissed):
Non blocking items that can be addressed in follow up.
Also PR description needs to be updated to match latest implementation

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
jsync-swirlds: WARNING is reserved for cases where the block node is unhealthy, not recoverable events when connecting to remote services.
INFO is not lost; that is the floor for production logs.

The clearest question to ask here is "should this event wake the on-call engineer", warning logs will; info logs will not.

Levels are [documented here](https://github.com/hiero-ledger/hiero-block-node/blob/main/docs/logging-guidelines.md)

docs/design/persistence/cloud-storage-expanded.md
jsync-swirlds: Why are we using strikethrough for angle brackets?
```suggestion
        -retryFutureBlockNumbers: `Map<Future, Long>`
```

ivannov: Thanks for the catch @Nana-EC! Updated the documentation.

@jsync-swirlds this is part of the mermaid diagram of the `ExpandedCloudStoragePlugin` class. And this is the syntax for angle brackets in mermaid

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
ivannov: Changed the `RetryTask` to also return the source of the block. Let's not hardcode that

jsync-swirlds (approved):

block-node/cloud-storage-expanded/src/main/java/org/hiero/block/node/cloud/storage/expanded/ExpandedCloudStoragePlugin.java
Nana-EC: I meant lost in the sea of INFO logs, but that highlights a different issue which Fredy is fixing in his PR.

Nana-EC (approved):

PR #3225, opened 2026-07-15
fix: drop orphan sidecars at wrap time via signed manifest check (fixes #3224)

### LINKED ISSUE #3224: wrap: filter candidate sidecar files against record file's signed sidecar manifest
## Problem

The WRB wrap pipeline (`ToWrappedBlocksCommand`, `LiveSequential`, and shared plumbing via `findMostCommonSidecars` / `ParsedRecordBlock.parse`) associates sidecar `.rcd.gz` files with a record file **purely by filename pattern**. It never consults the record file's signed `record_file_contents.sidecars[]` manifest.

The manifest is the authoritative list. It's what the CN's RSA signature covers. When a source-side artifact leaves an orphan sidecar file next to a record file (a stray sidecar written by one node while the consensus majority declared none), the wrap process picks it up anyway and embeds its bytes into the WRB.

## Evidence

The mainnet integrity check added by PR #3219 (issue #3196) surfaced this in three real blocks across ~97.6M scanned:

| Block         | Dissenting node(s)   | Sidecar internal ts drift |
|---------------|----------------------|---------------------------|
| 52,333,943    | 0.0.7                | +13.3 hours               |
| 58,446,161    | 0.0.23               | +73 minutes               |
| 62,113,066    | 0.0.33 (+ 0.0.34*)   | +33 minutes               |

*On block 62,113,066 nodes 33 and 34 each wrote a different orphan sidecar; wrap consumed node 33's.

For every one of these blocks:
- 26+ majority nodes wrote byte-identical record files with **zero** entries in `sidecars[]`
- 1-2 dissenting nodes wrote an orphan sidecar file whose internal `TransactionSidecarRecord.consensus_timestamp` belongs to a much later block
- Wrap swept the orphan into the WRB despite the record file's signed manifest declaring nothing

Full investigation notes: [captured on Rocky's desktop, summarized in PR #3219 description]

## Proposed fix

In the wrap path (both `ToWrappedBlocksCommand` and `LiveSequential`), before embedding candidate sidecar files into the `RecordFileItem`:

1. Compute SHA-384 of each candidate sidecar's serialized bytes.
2. Cross-check against the record file's signed `sidecars[]` metadata list.
3. If a candidate sidecar's hash does not appear in the signed manifest, drop it and log a warning naming the block, the sidecar filename, and the offending hash.
4. If the signed manifest has an entry with no matching candidate, that's still a `MISSING` failure and should abort (existing `SidecarIntegrityValidation` behavior).

Estimated scope: ~10 lines in `findMostCommonSidecars` / `ParsedRecordBlock.parse`, plus a small refactor so the manifest is available at the point of sidecar candidate selection.

## Why this is separate from #3196

Issue #3196 / PR #3219 is a **detection** effort. It flags divergence between the signed manifest and the embedded sidecar bytes. It does not modify wrap-side selection logic; wrap still embeds whatever sidecar files match the filename pattern, then the check aborts the wrap if the result is inconsistent.

This issue is a **prevention** effort. Even with detection in place, historical WRBs already carry the bogus sidecars, and any future orphan (from the same or a similar CN behavior) would still halt the wrap process. Filtering candidates against the signed manifest before embedding means the wrap simply produces a manifest-consistent WRB and continues, only halting on the truly ambiguous case (signed manifest declares a sidecar but no candidate file matches its hash).

## Acceptance criteria

- [ ] Wrap paths (`ToWrappedBlocksCommand` and `LiveSequential`) filter candidate sidecar files by SHA-384 against the record file's signed `sidecars[]` list before embedding
- [ ] Dropped candidates are logged with block number, filename, and SHA-384
- [ ] Existing `SidecarIntegrityValidation` still fires for the MISSING case (signed hash with no candidate)
- [ ] Unit tests cover: (a) matching candidate embedded, (b) orphan candidate dropped and warning logged, (c) missing candidate for signed hash still aborts wrap
- [ ] End-to-end validation on the three known-affected mainnet block ranges shows re-wrap succeeds and produces WRBs whose validation passes

## Related

- Issue #3196: sidecar integrity validation (detection layer, closed by PR #3219)
- PR #3219: adds `SidecarIntegrityValidation` and `blocks validate-sidecars` / `blocks find-block` tooling used to identify these three blocks
- Follow-up ticket also needed for a `--allow-sidecar-mismatch` flag on the wrap CLI so operators can proceed past known-bad historical blocks during backfill without re-wrapping

rockysingh:
## Summary

Fixes #3224. The wrap pipeline picks up any file matching the sidecar filename pattern (`<timestamp>_NN.rcd*`) next to the record file and embeds its bytes into the WRB, without checking whether the record file's signed `sidecars[]` manifest actually declares it. This change treats the manifest as authoritative: candidate sidecars whose SHA-384 does not appear in the signed list are dropped with a warning before the RecordFileItem is constructed.

## What changed

- `ParsedRecordBlock.parse` now runs a filter pass over the parsed sidecar list against `recordStreamFile.sidecars[]`, keeping only entries whose SHA-384 matches a signed metadata hash.
- New `filterOrphanSidecars(List<SidecarFile>, List<SidecarMetadata>)` helper. Wraps a second overload that accepts an injectable `Consumer<String>` warning sink so tests can capture drop diagnostics without racing on `System.setErr` under parallel test forks.
- 7 unit tests in `ParsedRecordBlockFilterTest` covering: empty input short-circuit, all-match retention, single-orphan drop, all-orphan drop, metadata without hash field, extra signed hash with no candidate, and preservation of retained-list ordering.

## Motivation

Empirical background from the sidecar integrity investigation on mainnet (~97.6M blocks): three historical WRBs turned out to contain orphan sidecar bytes from 1-2 dissenting CN nodes for blocks that had zero contract activity. The dissenters' sidecar files remained in the bucket unreferenced by any signed manifest; the wrap-time filename matcher swept them in anyway. Detail in issue #3224.

The wrap-side integrity check (added by PR #3219) catches this at wrap time by aborting the wrap. That's the right safety net, but the correct default is to not embed the orphan in the first place. This PR is that default.

## Scope note (historical remediation is out of scope)

The 3 known mainnet occurrences are locked into the canonical chain hash for their block ranges (the entire serialized `RecordFileItem`, including `sidecar_file_contents`, is hashed into the WRB block hash via `BlockStreamBlockHasher`; see `case BLOCK_HEADER, RECORD_FILE, ...` in `hashBlockInternal`). Re-wrapping any of them would cascade through every subsequent block's `previous_block_hash` and effectively fork mainnet from the network's perspective. Those WRBs are canonical mainnet; this PR is prevention-only, not remediation.

## Related

- Issue #3196 / PR #3219: sidecar integrity validation (detection side)
- Investigation notes summarize the three mainnet observations that motivated this fix

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/model/parsed/ParsedRecordBlock.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/model/parsed/ParsedRecordBlock.java
@@ -5,6 +5,7 @@ import static org.hiero.block.tools.utils.Sha384.sha384Digest;
 
 import com.hedera.hapi.node.base.NodeAddressBook;
 import com.hedera.hapi.streams.SidecarFile;
+import com.hedera.hapi.streams.SidecarMetadata;
 import com.hedera.pbj.runtime.ParseException;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.hedera.pbj.runtime.io.stream.WritableStreamingData;
@@ -13,6 +14,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.zip.GZIPOutputStream;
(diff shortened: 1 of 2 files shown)


jsync-swirlds (dismissed):

tools-and-tests/tools/src/main/java/org/hiero/block/tools/records/model/parsed/ParsedRecordBlock.java
```diff
@@ -48,40 +50,120 @@ public Instant blockTime() {
...
+    private static boolean containsHash(final List<byte[]> hashes, final byte[] target) {
+        for (final byte[] h : hashes) {
+            if (Arrays.equals(h, target)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static String hex(final byte[] bytes) {
```
berryware: HexFormat.of().formatHex(bytes) is used elsewhere. Do we need this method?

rockysingh: Good call. Fixed in 2946252dd, inlined `HexFormat.of().formatHex(hash)` and dropped the custom `hex(byte[])` helper. Same lowercase output, and it's now consistent with how `UnparsedRecordBlockV6`, `MissingBlockFiller`, `MirrorNodeAddressBook`, `SigFileUtils`, etc. already do it.

berryware (approved):
Looks good

jsync-swirlds (approved):

Issue #2381, opened 2026-03-17
Update usage of configuration API


jsync-swirlds:
### Story Form

As a Block Node engineer
I want to use configuration APIs effectively and cleanly
So that I can manage operations during testing and ensure operators have effective means to manage individual deployments.

### Technical Notes

The config API defines several updated usage patterns that currently differ from Block Node usage. We need to make changes to migrate to the recommended usage.
This includes using the standard translations from environment variable to configuration and using the standard test configuration mechanisms rather than specialized test fixtures.

Many (perhaps most) of the sub-optimal usage is found in tools and simulator codebases. These must be included in the changes under this issue.

There are several issues already filed to accomplish portions of this task:
#2370
#1209
#2371
#1432

Possibly obsolete, related, issues
#1090
#1404

The related issues should be reviewed and closed in association with this task.
The possibly related issues should be reviewed and closed if obsolete following completion of this task.

PR #3232, opened 2026-07-16
chore: Don't register config extensions manually

### LINKED ISSUE #2370: Don't register config extensions manually
### Story Form

As a block node developer
I want to avoid registering configuration extensions manually
So that I rely on the `ConfigurationBuilder` auto discovery mechanism

### Technical Notes

From the [code review comment](https://github.com/hiero-ledger/hiero-block-node/pull/2351/changes#r2927273083):

When you use `.autoDiscoverExtensions()`, you can completely avoid manually registering your custom config records (calling `withConfigDataType()` and `withConfigDataTypes()`), if you create an extension of `ConfigurationExtension` (per module providing custom config records) and register it in `module-info.java` as a service provider.

That will allow the config library to discover it and avoid knowing explicitly about all config extensions.

See an example of how it's done [here](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hiero-observability/openmetrics-httpserver/src/main/java/module-info.java).

ivannov:
### `ConfigurationExtension` auto-discovery for all 16 config-bearing modules

Every module with a `@ConfigData` record now ships a `<Module>ConfigExtension implements ConfigurationExtension` plus a `provides com.swirlds.config.api.ConfigurationExtension with ...;` directive in its module-info, mirroring the existing SimulatorConfigExtension pattern. `app-config` gets a single `AppConfigExtension` covering the four app-level records (ServerConfig, WebServerHttp2Config, NodeConfig, ApplicationStateConfig) that were previously wired in by hand in BlockNodeApp. This is purely additive: `BlockNodePlugin.configDataTypes()` overrides are untouched since PluginTestBase still depends on them.

### `BlockNodeApp` retires manual config-type aggregation

`BlockNodeApp`'s constructor no longer builds `allConfigDataTypes` by walking `loadedPlugins` and calling `.configDataTypes()` on each, and no longer calls `.withConfigDataTypes(...)` on the `ConfigurationBuilder`; `.autoDiscoverExtensions()` now does that job via the extensions added above. `AutomaticEnvironmentVariableConfigSource`
  (which still needs an explicit type list to derive env var names) is now populated from `serviceLoader.loadServices(ConfigurationExtension.class)` instead, so both consumers read from the same source of truth.

## Related Issue(s)

Closes #2370
Relates to #2381

### DIFF
--- a/block-node/app/src/main/java/module-info.java
+++ b/block-node/app/src/main/java/module-info.java
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
+import com.swirlds.config.api.ConfigurationExtension;
 import org.hiero.block.node.spi.BlockNodePlugin;
 import org.hiero.block.node.spi.blockmessaging.BlockMessagingFacility;
 import org.hiero.block.node.spi.historicalblocks.BlockProviderPlugin;
@@ -6,11 +7,6 @@ import org.hiero.block.node.spi.historicalblocks.HistoricalBlockFacility;
 
 module org.hiero.block.node.app {
     exports org.hiero.block.node.app;
-    // it is expected the app module will never export any packages, only use others
-    uses HistoricalBlockFacility;
-    uses BlockMessagingFacility;
-    uses BlockNodePlugin;
-    uses BlockProviderPlugin;
 
     requires com.hedera.pbj.grpc.helidon.config;
     requires com.hedera.pbj.grpc.helidon;
@@ -30,4 +26,11 @@ module org.hiero.block.node.app {
     requires java.logging; // javax.annotation.processing.Generated
     requires static transitive com.github.spotbugs.annotations;
     requires static java.compiler;
+
+    uses BlockMessagingFacility;
+    uses BlockNodePlugin;
+    uses BlockProviderPlugin;
+    uses ConfigurationExtension;
+    // it is expected the app module will never export any packages, only use others
+    uses HistoricalBlockFacility;
 }
--- a/block-node/app/src/main/java/module-info.java
+++ b/block-node/app/src/main/java/module-info.java
@@ -31,6 +31,5 @@ module org.hiero.block.node.app {
     uses BlockNodePlugin;
     uses BlockProviderPlugin;
     uses ConfigurationExtension;
-    // it is expected the app module will never export any packages, only use others
     uses HistoricalBlockFacility;
 }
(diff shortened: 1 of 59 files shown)


block-node/app/src/main/java/module-info.java
```diff
@@ -30,4 +26,11 @@
     requires java.logging; // javax.annotation.processing.Generated
     requires static transitive com.github.spotbugs.annotations;
     requires static java.compiler;
+
+    uses BlockMessagingFacility;
+    uses BlockNodePlugin;
+    uses BlockProviderPlugin;
+    uses ConfigurationExtension;
+    // it is expected the app module will never export any packages, only use others
```
jsync-swirlds: This probably belongs a few lines above; spotless reordered somewhat blindly.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -206,24 +206,23 @@ public BlockNodeApp(final ServiceLoaderFunction serviceLoader, final boolean sho
...
-        allConfigDataTypes.add(ApplicationStateConfig.class);
-        loadedPlugins.forEach(plugin -> allConfigDataTypes.addAll(plugin.configDataTypes()));
+        // Collect all the config data types registered by every module's ConfigurationExtension; this is the same
+        // source of truth that autoDiscoverExtensions() below resolves against, so the env var source below sees
+        // exactly the same set of config data types as get registered with the configuration.
+        final List<Class<? extends Record>> allConfigDataTypes = serviceLoader
+                .loadServices(ConfigurationExtension.class)
+                .flatMap(extension -> extension.getConfigDataTypes().stream())
+                .distinct()
+                .toList();
```
jsync-swirlds: We should remove this too.
I know that test code depends on it, but we can move it into that test code.

jsync-swirlds (dismissed):
A couple of suggestions, non-blocking.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: You are right, we don't need this here. So I removed it (along with the `configDataTypes()` method on the `BlockNodePlugin` interface. But couldn't find a way to do without this piece of code in `AutomaticEnvironmentVariableConfigSource` in the `app-config` module. Declaring a config extension lead to a deadlock. Didn't find clue how to not have this code there

jsync-swirlds: Let's file a follow-up issue to resolve that, then.
It's doable, but probably requires a slightly larger change to the test fixtures (among other places).

jsync-swirlds (approved):
It would still be good to use the built-in environment variable source from the upstream library at some point.

This looks good so far.

ata-nas (approved):
Nice improvement.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ivannov: Created #3250 to track this

PR #3235, opened 2026-07-16
docs: catalog every block validation in blocks-commands.md (fixes #3234)

### LINKED ISSUE #3234: docs: document all block validation types in blocks-commands.md
## Problem

The current \`tools-and-tests/tools/docs/blocks-commands.md\` documentation for \`blocks validate\` only lists three high-level check categories (hash chain, signatures, balances). Under the hood the command actually runs 13+ distinct validations, each with its own purpose, skip flag, and failure semantics. Operators running the tool have no reference doc that maps a failure to the check that produced it, or a way to know which validations are on by default vs. opt-out.

## Missing coverage

Individual validations that are not currently documented:

| Validation class | Documented today | Skip flag |
|-------|:----------------:|-----------|
| \`RequiredItemsValidation\` | no | \`--skip-required-items\` |
| \`BlockStructureValidation\` | no | \`--skip-required-items\` |
| \`SignatureValidation\` | partial | \`--skip-signatures\` |
| \`AddressBookUpdateValidation\` | no | always on |
| \`NodeStakeUpdateValidation\` | no | always on |
| \`TssEnablementValidation\` | no | always on |
| \`BalanceCheckpointValidation\` | yes | \`--validate-balances=false\` |
| \`BlockChainValidation\` | partial | always on |
| \`HistoricalBlockTreeValidation\` | no | always on |
| \`HbarSupplyValidation\` | partial | \`--skip-supply\` |
| \`HashRegistryValidation\` | no | always on |
| \`StreamingMerkleTreeValidation\` | no | always on |
| \`JumpstartValidation\` | no | always on |

Additionally, \`blocks validate-sidecars\` and \`blocks find-block\` (introduced by PR #3219) need their own subsections. Those will ship with #3219 directly; this issue is scoped to the \`blocks validate\` deep dive.

## Proposed doc structure

Add a new subsection to \`blocks-commands.md\` under \`blocks validate\` titled "Validations run inside \`blocks validate\`". For each validation include:

1. What it checks (one to two sentences, taken from the class's \`description()\` return value plus a clarifying line)
2. Type: parallel (stateless, per-block) vs. sequential (needs prior blocks / state)
3. Default on/off and its skip flag if applicable
4. What a failure looks like (example diagnostic, if easily reproducible)
5. When you might legitimately skip it (e.g., skip-supply for a network with known transfer-list imbalances)

A table at the top summarizing all 13+ validations would let operators grep for a failure name and jump to the section.

## Acceptance criteria

- [ ] Every current \`BlockValidation\` implementer in \`tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/\` has a documented entry
- [ ] Summary table lists all validations, their type, and their skip flag
- [ ] Each entry has: purpose (1-2 sentences), example failure output, when-to-skip guidance if applicable
- [ ] Cross-links from the top-level tools README point to the new section
- [ ] No stale references (removing anything the code has renamed)

## Related

- PR #3219 (sidecar integrity validation) will add \`SidecarIntegrityValidation\` to the pipeline plus new \`blocks validate-sidecars\` / \`blocks find-block\` subcommands. Those docs land with #3219; this issue is orthogonal and covers the deep dive on the \`blocks validate\` catalog.

rockysingh:
## Summary

Fixes #3234. The `blocks validate` command orchestrates 14 individual validation classes under the hood; the existing docs listed only a handful of high-level checks (hash chain, signatures, balances). This PR adds a comprehensive "Validations in detail" section to `tools-and-tests/tools/docs/blocks-commands.md` so operators have a single reference that maps each failure mode to the class that produced it.

## What's in the update

- **Summary table** listing every validation with its type (parallel / sequential / end-of-run), whether it requires starting from block 0, and how to skip it.
- **Per-validation subsections** with: what the check does, example failure output where useful, and when an operator might legitimately disable it.
- Refresh of the top-of-section bullet list to reflect current behavior and link into the new deep dive.
- Explicit call-out of **fungible-token balance validation** inside `BalanceCheckpointValidation` (the code has always validated both HBAR and fungible-token balances per account via `RunningAccountsState.applyFungibleTokenChange` and `BalanceCheckpointValidator.validateCheckpoint(blockNumber, hbarBalances, tokenBalances)`; the docs previously only mentioned HBAR).

## Coverage

Sections added for:

- `RequiredItemsValidation`
- `BlockStructureValidation`
- `SignatureValidation`
- `SidecarIntegrityValidation`
- `AddressBookUpdateValidation`
- `NodeStakeUpdateValidation`
- `TssEnablementValidation`
- `BlockChainValidation`
- `HistoricalBlockTreeValidation`
- `HbarSupplyValidation`
- `BalanceCheckpointValidation`  (HBAR + fungible tokens + NFT ownership tracked)
- `HashRegistryValidation`
- `StreamingMerkleTreeValidation`
- `JumpstartValidation`

## Note on scope

The `SidecarIntegrityValidation` deep-dive is now included here so the `blocks validate` catalog is complete on merge, no matter which of #3235 / #3219 lands first. The standalone `blocks validate-sidecars` and `blocks find-block` subcommand sections still ship with PR #3219.

## Test plan

- [x] Doc-only change, no code touched
- [x] Table markdown renders correctly in GitHub preview
- [x] All in-page anchor links (`#requireditemsvalidation`, etc.) resolve
- [x] Descriptions cross-checked against each class's `description()` return string and top-of-file javadoc so the doc stays truthful to the code
- [x] Fungible-token coverage cross-checked against `BalanceCheckpointValidator.validateCheckpoint(...)` and `RunningAccountsState` javadoc


AlfredoG87 (dismissed):
LGTM

rockysingh:
@Nana-EC good catch, my earlier docs wording was misleading. On closer look at `RunningAccountsState.compare(AllAccountBalances)` (lines 197-215):

```java
for (final TokenUnitBalance tokenUnitBalance : snapshotEntry.tokenUnitBalances()) {
    final long tokenNum = tokenUnitBalance.tokenIdOrThrow().tokenNum();
    final long expectedBalance = tokenUnitBalance.balance();
    ...
    final long fungible = account.fungibleBalances.getIfAbsent(tokenNum, 0L);
    final LongHashSet nftSet = account.nftSerials.get(tokenNum);
    final int nftCount = nftSet == null ? 0 : nftSet.size();
    actualBalance = fungible + nftCount;
    ...
}
```

NFT ownership **is** validated at every checkpoint.  The *count* of NFT serial numbers each account holds per token is compared against the checkpoint's per-token `balance` field, so a lost or extra NFT for any account would fail the checkpoint.

What's NOT compared at checkpoint is the *identity* of individual serial numbers. The reason is that Hedera's on-disk balance snapshot format (`AllAccountBalances` protobuf, which is what `fetchBalanceCheckpoints` extracts from the bucket) only records a per-token `balance` scalar per account — it doesn't carry a list of the specific serials each account owns. So even though the tools track per-serial ownership continuously (correctly — every NFT mint / transfer / wipe / burn is applied to the account-serial set), there's no equivalent field in the checkpoint file to compare each serial against.

So this isn't really deferred: it's a data availability constraint upstream. If Hedera later exposes per-serial ownership in the balance snapshot schema, we can wire the comparison in trivially (the data is already in `account.nftSerials`). Until then, count-parity per account/token is the strongest assertion the checkpoint format allows.

Fixing the docs to reflect this now. The earlier phrasing "NFT ownership is retained for future checkpoint schemas" undersold what actually runs today. Will push in a moment.

AlfredoG87 (dismissed):
LGTM

jsync-swirlds (dismissed):

tools-and-tests/tools/docs/blocks-commands.md
```diff
@@ -15,6 +15,53 @@ The `blocks` command contains utilities for working with Block Stream files (.bl
...
+  -o /path/to/wrappedBlocks \
+  -b /path/to/metadata/block_times.bin \
+  -d /path/to/metadata/day_blocks.json
+
+# 2. Validate the resulting archive end-to-end
+java -jar tools-<VERSION>-all.jar blocks validate /path/to/wrappedBlocks
+
+# 3. Start live-sequential from the wrap watermark
+java -jar tools-<VERSION>-all.jar days live-sequential \
+  --wrapped-blocks-dir /path/to/wrappedBlocks
```
berryware: I'm not seeing this option in live-sequential.

tools-and-tests/tools/docs/blocks-commands.md
```diff
@@ -15,6 +15,53 @@ The `blocks` command contains utilities for working with Block Stream files (.bl
...
+|-------|------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 1     | `blocks wrap -i <compressedDays> -o <wrappedBlocks>` | `wrappedBlocks/*.zip` (WRB batches), `addressBookHistory.json`, `nodeStakeHistory.json`, `blockStreamBlockHashes.bin`, `streamingMerkleTree.bin`, `wrap-commit.bin`, `jumpstart.bin`               |
+| 2     | `blocks validate <wrappedBlocks>`                    | Confirms the chain is intact end-to-end; updates checkpoint state under `<wrappedBlocks>/validateCheckpoint/` (address book, node stakes, TSS enablement, running balances, jumpstart consistency) |
+| 3     | `days live-sequential …`                             | Picks up from the wrap watermark; continues downloading, wrapping, and validating live blocks from the CN in real time (see [days-commands.md](days-commands.md))                                  |
+
+### Why the order matters
```
berryware, line 30: Is this ordering accurate? Do we still need the validate step?

tools-and-tests/tools/docs/blocks-commands.md
rockysingh: The actual option on `days live-sequential` is `--wrap-output-dir` (see `LiveSequential.java:167`). Fixed in 21fcf64ae and updated both the example command and the pipeline-table entry to use the correct name.

tools-and-tests/tools/docs/blocks-commands.md
rockysingh: I checked the code and `validate` is strongly recommended but NOT strictly required. On startup `live-sequential` calls `v.load(checkpointDir)` only if `validateCheckpoint/validateProgress.json` exists (`LiveSequential.loadValidationCheckpoints`, around line 1052), and later code drops any validation whose `requiresGenesisStart()` is true when no checkpoint was loaded. So without a preceding `validate`, `live-sequential` still enforces hash chain / signatures / sidecar / required-items, but silently drops HBAR supply, historical block tree, jumpstart, hash registry, and streaming merkle tree end-of-run checks. This is not ideal though as we are dropping this type of validation.

Softened the wording in 21fcf64ae and added an explicit trade-off table listing all 8 affected checks so operators can make an informed choice. Also added a pitfall about a `validateCheckpoint` that's ahead of the wrap watermark (which is the fallback path in the code).

Nana-EC (approved):

jsync-swirlds (approved):

berryware (approved):

PR #3245, opened 2026-07-20
chore: Preserve the real exception when a Future fails in CloudStorageArchivePlugin instead of a causeless IllegalStateException

### LINKED ISSUE #3239: Preserve the real exception when a Future fails in CloudStorageArchivePlugin instead of a causeless IllegalStateException
### Story Form

I want to log the real exception when a submitted task's Future completes exceptionally
So that operators can diagnose upload/recovery/consolidation failures from the log instead of seeing a generic, causeless IllegalStateException

### Background

`CloudStorageArchivePlugin` has four call sites that call `Future.resultNow()` once a background task is done. The JDK's default `Future.resultNow()` throws a causeless `IllegalStateException` ("Task completed with exception") when the task failed: it does not attach the original exception as the cause. Every existing `catch (IllegalStateException e)` around these calls therefore logs a generic exception instead of the real
`S3ResponseException` / `IOException` / `S3ClientInitializationException` the task actually threw, which makes production failures hard to diagnose from logs alone.

### Changes Required

- At each of the four call sites  replace `Future.resultNow()` with `Future.get()` wrapped in a local `try`/`catch`: on success, keep the existing happy-path handling unchanged; on `catch (ExecutionException e)`, use `e.getCause()` to obtain the real `Throwable` the task threw and log it in place of the causeless `IllegalStateException`
- Add a `catch (InterruptedException e)` at each site that restores the thread's interrupt flag (`Thread.currentThread().interrupt()`) and takes no further action

ivannov:
### Four Future call sites switched from resultNow() to get()

`checkCompletedUpload()`, `checkAndDrainTempUploadResults()`, `checkAndDrainConsolidations()`, and `completeRecoveryIfReady()` all consumed an already-done `Future` via `resultNow()`, which throws a causeless `IllegalStateException` ("Task completed with exception") on failure and discards the real thrown exception. Each site now calls `get()` in a local try/catch: `catch (ExecutionException e)` logs `e.getCause()`, the actual `S3ResponseException` / `IOException` / `S3ClientInitializationException` the task threw, and `catch (InterruptedException e)` restores the interrupt flag and takes no further action, matching this plugin's behavior prior to an earlier refactor (26195e56) that had switched it to `resultNow()`.

### Test coverage added via a shared TestLogHandler

`TestLogHandler` (test fixtures) now also captures `LogRecord#getThrown()`, not just messages. The three affected test nest classes attach/detach it per test and assert the real exception type (and, where relevant, message or status code) was logged.

## Related Issue(s)

Closes #3239

### DIFF
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
@@ -247,30 +247,39 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
             } else {
                 logInvalidOrFailedNotification(notification);
             }
-        } catch (IllegalStateException e) {
-            LOGGER.log(WARNING, "Could not complete uploading blocks to cloud archive storage", e);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (ExecutionException e) {
             metricsHolder.failedTasks().increment();
+            // Capture before triggerMidRunRecovery() runs, since it nulls currentUploadFuture itself.
+            final boolean fromUploadTask = currentUploadFuture != null;
             triggerMidRunRecovery();
-            // routeVerifiedBlock() was never reached (the exception bypassed it), so stash the
-            // triggering block manually — it is safe to use blocksStash directly because
-            // triggerMidRunRecovery() already cleared currentUploadFuture.
-            if (notification != null && notification.block() != null) {
+            if (fromUploadTask) {
+                // routeVerifiedBlock() was never reached (the exception bypassed it), so stash the
+                // triggering block manually — it is safe to use blocksStash directly because
+                // triggerMidRunRecovery() already set currentGroupStart to -1.
+                LOGGER.log(WARNING, "Block upload task failed", e.getCause());
                 blocksStash.put(
                         notification.blockNumber(), new BlockWithSource(notification.block(), notification.source()));
+            } else {
+                // Exception from the recovery task itself; recoveryFuture is already null (cleared in
+                // completeRecoveryIfReady() before this was thrown), so the triggerMidRunRecovery() call
+                // above submits a retry.
+                LOGGER.log(WARNING, "Could not complete uploading blocks to cloud archive storage", e.getCause());
             }
         }
     }
 
     /// Checks whether the active [BlockUploadTask] has finished and cleans up its state.
     /// Returns `true` if the task was cancelled, signalling the caller to skip further processing.
-    private boolean checkCompletedUpload() {
+    private boolean checkCompletedUpload() throws ExecutionException, InterruptedException {
         boolean cancelled = false;
         if (currentUploadFuture != null && currentUploadFuture.isDone()) {
             if (currentUploadFuture.isCancelled()) {
                 LOGGER.log(TRACE, "Block upload task was cancelled");
                 cancelled = true;
             } else {
-                final UploadResult uploadResult = currentUploadFuture.resultNow();
+                final UploadResult uploadResult = currentUploadFuture.get();
                 if (uploadResult == UploadResult.FAILED) {
                     LOGGER.log(WARNING, "Block upload task failed");
                     metricsHolder.failedTasks().increment();
@@ -293,7 +302,7 @@ public class CloudStorageArchivePlugin implements BlockNodePlugin, BlockNotifica
     /// [TempArchiveUploadTask] calls `ApplicationStateFacility.addStoredBlockRange()` once after
     /// both the multipart upload and the meta file write succeed, so no additional range
     /// registration is needed here.
-    private void checkAndDrainTempUploadResults() {
+    private void checkAndDrainTempUploadResults() throws InterruptedException {
         final Iterator<Map.Entry<Long, Future<TempArchiveEntry>>> it =
                 tempUploadFutures.entrySet().iterator();
         while (it.hasNext()) {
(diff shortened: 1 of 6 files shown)


block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -247,30 +247,39 @@ public void handleVerification(VerificationNotification notification) {
             } else {
                 logInvalidOrFailedNotification(notification);
             }
-        } catch (IllegalStateException e) {
-            LOGGER.log(WARNING, "Could not complete uploading blocks to cloud archive storage", e);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (ExecutionException e) {
```
ata-nas, line 252: Have we fired a failed notification at some point?

ata-nas, line 252: We do not catch CancellationException, which is thrown if we call get, but the task is canceled. We should be aware when calling get. Either a check if the future is canceled, or a catch of the exception. This should be considered everywhere we call get on any future.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -247,30 +247,39 @@ public void handleVerification(VerificationNotification notification) {
...
             }
         }
     }
 
     /// Checks whether the active [BlockUploadTask] has finished and cleans up its state.
     /// Returns `true` if the task was cancelled, signalling the caller to skip further processing.
-    private boolean checkCompletedUpload() {
+    private boolean checkCompletedUpload() throws ExecutionException, InterruptedException {
         boolean cancelled = false;
         if (currentUploadFuture != null && currentUploadFuture.isDone()) {
```
ata-nas, line 277: As mentioned elsewhere, multiple threads executing this method or making any operations that change non-volatile/atomic filed values, and also operating on futures, are problematic. My fear is that we assume handling of these futures is biased towards expecting only a single thread at a time will operate on these.

AlfredoG87 (dismissed):
LGTM

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: We should pass in the future to check rather than having different methods for each.
Accessing futures is pretty thread-safe (it would be pretty silly if it were not), however.

ata-nas: I am not worried about thread safety on accessing the futures results themselves, rather than multiple threads calling the method which access and write to non volatile fields and doing operations based on results from the futures. The code seems to assume that onle one thread at a time will call these methods. Not blocking, just noting. We have other prs we can look at this.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Good point. Out of the four call sites of `Future.get()`, three of them were had a `Future.isCancelled()` guard. Only 'completeRecoveryIfReady` call site didn't. So I added one there too.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: @ata-nas as I replied in another PR, where you raised the same issue, here is my understanding. For the most part of the lifetime of this and other futures, they are only called by the message thread (i.e. by `handleNotifications`) method. We spawn quite a few tasks in the plugin, but none of them has access to the future objects. The only case outside this thread is in `stop()`, but before that the plugin unregisters as notification handler, so it can't expect to keep receiving notifications. So I don't see why we should make them volatile. We can of course, it's just adding a key word, but is it worth it?

@jsync-swirlds if I understood you correctly, you are proposing something that we discussed offline while defining the task. You asked me to explore the possibility of a generic helper method that would wrap calling the get() and the exception handling and return a record containing either the result or the exception.

I tried this and it occurred that the actual handling (what to log, what metric to bump, whether to call `triggerMidRunRecovery()` / `checkGroupCoverage()`, what to do on future being cancelled, whether the exception propagates to `handleVerification()` or is handled locally) differs at each call site, so a shared helper doesn't remove much duplication. That's why I discarded this implementation option.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Here is the full picture:

* `BlockUploadTask`: per-block / per-part failures inside `flushPartIfNeeded` / `uploadFinalPart` already send their own false notification and return `FAILED` instead of throwing: those never reach this catch. Only two calls in call() that can throw are: 
  * `createMultipartUpload()`: it is called before any block is buffered for uploading at all, Upon exception catch, the block which triggered the execution of the block upload task, is stashed
  * `completeMultipartUpload()`: here we have success notifications sent for all the blocks, we just failed to complete the upload
* `ConsolidationTask`: correctly sends nothing on failure, the source .tmp archives were already notified when created, and they're only deleted after a successful merge
* `StartupRecoveryTask`: never persists any block itself (read-only S3 scan), so no notification should be sent
* `TempArchiveUploadTask`: there are three places where we catch exception (and then re-rethrow it). In two of them persisted notification was send, in the third - not. So I changed that (not sure why I made it like that) and closed the gap

So, about `createMultipartUpload` and `completeMultipartUpload`: I should investigate deeper the implications (especially for the latter). But I think it goes beyond the scope of this PR.

AlfredoG87 (approved):

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: I still believe we can implement the helper approach and clean this code up a lot, but I'll explore that separately as a future task.

My comment can be considered non-blocking.

ata-nas: @ivannov the code is quite a lot and I do not have the full context. My concern is general, not specific. Just noting on possible undesired consequences in certain cases, which I am unaware if are true. My point is to raise attention to you, and in case any of my concerns are valid, we have to do something about that.

Note that even if you unregister on stop before accessing the fields from another thread, that does not mean that messaging thread is not in the middle of accessing them. This is a technicality, but can happen. In those cases, is there a possibility we end up with an undesirable end result?

Note that this is non-blocking. Note also that I agree with @jsync-swirlds's proposal as well.

berryware (approved):
Looks good

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: Please do file an issue to correct the behavior with create/complete multipart.
We should notify failure if the upload fails and cannot be stashed or otherwise stored in the bucket somewhere.

jsync-swirlds (approved):

PR #3246, opened 2026-07-20
feat: First Ordered Block Config in Verification

### LINKED ISSUE #3247: First Ordered Block Config in Verification
### Story Form

As a Block Node Operator
I want to set a first ordered block value for verification
So that all blocks below that pass verification immediately without blocking for ordering

### Technical Notes

- We want to add a configurable value to allow us to not impose strict ordering for blocks lower than our config value
- All blocks equal to or higher than that value will be strictly ordered

ata-nas:
* Added the `firstOrderedBlock` config in verification
* Ordering manager is now aware of this configuration
* Added documentation
* Added tests

Resolves #3247

### DIFF
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/ResultOrderingManager.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/ResultOrderingManager.java
@@ -64,7 +64,9 @@ public final class ResultOrderingManager implements Function<BlockVerificationRe
 
     private boolean shouldPark(final BlockVerificationResult verificationResult, final long nextExpectedBlock) {
         // spotless:off
-        return (verificationResult.blockNumber() > nextExpectedBlock)
+        final long verifiedBlockNumber = verificationResult.blockNumber();
+        return (verifiedBlockNumber >= verificationConfig.firstOrderedBlock())
+            && (verifiedBlockNumber > nextExpectedBlock)
             && (verificationResult.source() == BlockSource.PUBLISHER || verificationConfig.allSourcesRequireOrdering());
         // spotless:on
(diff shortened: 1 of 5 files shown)


block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/ResultOrderingManager.java
```diff
@@ -64,7 +64,9 @@ private void checkValidNextExpectedBlock(final AtomicLong lastVerifiedBlock, fin
 
     private boolean shouldPark(final BlockVerificationResult verificationResult, final long nextExpectedBlock) {
         // spotless:off
-        return (verificationResult.blockNumber() > nextExpectedBlock)
+        final long verifiedBlockNumber = verificationResult.blockNumber();
+        return (verifiedBlockNumber >= verificationConfig.firstOrderedBlock())
```
ata-nas, line 68: Having the first check be `(verifiedBlockNumber >= verificationConfig.firstOrderedBlock())` means that if we fail it, we short-circuit the other AND clauses. This means that we are NOT enforcing ordering EVEN IF source is publisher. Should that be so?

cc: @jsync-swirlds @AlfredoG87

jsync-swirlds: Yes, the point of the first-ordered configuration is to suppress ordering regardless of source; otherwise we could end up with CN publishing data that gets "stuck" because we do not expect to receive a predecessor block from any source, or we do not receive a backfilled predecessor quickly enough.

jsync-swirlds (approved):

ivannov (approved):

Issue #3195, opened 2026-07-10
Proto Ordinal Streaming Hasher Mapping in `BlockHasher`


ata-nas:
### Story Form

As a Block Hasher
I want to have proto ordinal based mapping of item type to corresponding hasher
So that I am forward compatible

### Technical Notes

- The `BlockHasher` receives block items and based on the item type, they data needs to go into a specific streaming hasher.
- We want to add a default branch and have ordinal based mapping to the correct streaming hasher to support forward compatibility.

The mapping is described below:
````java
 /// A single item within a block stream.
 ///
 /// Each item in the block stream SHALL be self-contained and independent,
 /// with the following constraints applicable to the _unfiltered_ stream.
 /// - A block SHALL start with a `header`.
 /// - A block SHALL end with a `block_proof`.
 /// - A `block_header` SHALL be followed by an `event_header`.
 /// - An `event_header` SHALL be followed by one or more
 ///   `event_transaction` items.
 /// - An `event_transaction` SHOULD be followed by a `transaction_result`.
 /// - A `transaction_result` MAY be followed by one or more `transaction_output`s.
 /// - A `transaction_result` (or a `transaction_output`, if present) MAY be
 ///     followed by one or more `state_changes`.
 ///
 /// This forms the following required sequence for each block, which is then
 /// repeated within the block stream, indefinitely.  Note that there is no
 /// container structure in the stream, the indentation below is only to
 /// highlight repeated subsequences.<br/>
 /// The order of items within each block below is REQUIRED and SHALL NOT change.
 ///
 /// ```text
 /// header
 ///   repeated {
 ///     start_event
 ///     repeated {
 ///       event_transaction
 ///       transaction_result
 ///       (optional) transaction_output
 ///       (optional) repeated state_changes
 ///       (optional) filtered_single_item
 ///     }
 ///     block_footer
 ///     repeated block_proof
 ///   }
 /// ```
 ///
 /// A filtered stream may exclude some items above, depending on filter
 /// criteria. A filtered item is replaced with a merkle path and hash value
 /// to maintain block stream verifiability.
 ///
 /// A BlockItem SHALL be individually and directly processed to create the
 /// item hash.<br/>
 /// Items to be hashed MUST NOT be contained within another item.<br/>
 /// Items which might be filtered out of the stream MUST NOT be
 /// contained in other items.
 ///
 /// ### Forward Compatibility
 /// In order to maximize forward compatibility, and minimize the need to
[...]

PR #3251, opened 2026-07-21
docs: Follow Up on Design Documentation and Plugin Javadoc

### LINKED ISSUE #2879: Follow Up on Design Documentation and Plugin Javadoc
## Context

PR #2701 updates `docs/design/block-verification.md` with open review conversations marked non-blocking by the team. This ticket captures the design decisions and document corrections that need follow-up before the doc can be considered stable.

---

### Design decisions

- [ ] **Session architecture — fork a verifier per block proof type** — @jsync-swirlds proposed that a session handles only the hashing process (matched by HAPI version), then forks a _verifier_ per block proof present after the block. A separate _result handler_ gathers the verifier results and decides what notifications to emit. This cleanly separates: hashing → session, proof cryptography → verifier, thread management → session handler, notification policy → result handler. Decide whether to adopt this model and update §3 (session design) accordingly.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3198290919

- [ ] **Failure metrics — single counter with labels vs separate counters per reason** — @ivannov raised whether failure reasons (bad block proof, parse failure, missing mandatory items, etc.) should each have their own metric or be combined in one counter with a `reason` label. Preferring one counter with labels (with future label expansion) is consistent with how other plugins approach this. Decide and document the metric scheme in §7 (metrics).
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3240940072

- [ ] **Boolean `isInformational` field vs `INFO_` prefix on `FailureType`** — @AlfredoG87 and @ata-nas discussed whether informational failure types should be marked with a naming prefix (e.g. `INFO_MISSING_RSA_KEYS`) or a boolean field on the notification. A boolean field is more explicit and avoids name-pattern coupling; a prefix is visible in any enum switch. Decide on the convention and apply it consistently to the `FailureType` enum definition in the doc.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277575875

- [ ] **`isInformational`/success-details in `VerificationNotification`** — @ivannov asked whether the notification should carry a boolean for informational (non-fatal) failures. Decide whether `VerificationNotification` carries an `isInformational` flag, and if so document its semantics alongside the existing `FailureType` field.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3240957466

### Document corrections

- [ ] **Clarify public key usage — sign vs verify** — @ivannov: "do we use the public keys to sign stuff or to verify signatures?" The document should explicitly state that public keys are used to _verify_ signatures (not to produce them); the block node is a verifier, not a signer.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3240818381

- [ ] **Review whether the ASF persistence section belongs in this doc** — @ivannov: if the Application State Facility is responsible for managing and persisting all back-hasher data, the section covering that in the verification design doc may be redundant or misplaced. Confirm scope and either remove the section or add a cross-reference to the ASF design.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3240911816

- [ ] **Terminology: "unverified" → "verification is considered failed"** — @ivannov: "unverified sounds as if the block's verification did not start at all." Replace uses of _unverified_ with _verification failed_ (or equivalent) to avoid ambiguity with blocks that have not yet been processed.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3240953806

- [ ] **Apply code suggestion — "Combined Schnorr signature"** — @AlfredoG87 left an unapplied suggestion to replace "a (future) TSS signature" with "a Combined Schnorr signature" since TSS produces a combined Schnorr signature rather than a collection of individual Schnorr signatures (as seen in the RSA/WRB case).
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277387892

- [ ] **Mention `backfillNotification` vs `BlockItemsHandler` in the data input section** — @AlfredoG87: the doc should explicitly note that verification uses `backfillNotification` for backfilled blocks rather than the `BlockItemsHandler`, so the distinction is clear to future contributors.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277469907

### Strategic / separate ticket candidate

- [ ] **Old state-proof verification session — create a deprecation plan** — @AlfredoG87, @ata-nas, and @jsync-swirlds agreed this path is increasingly a priority to remove. It is currently used only in unit tests and the simulator (not in production paths: solo-e2e and BNCE run with TSS; WRBs use RSA). A deprecation plan with a target milestone should be tracked as a separate issue.
  - https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277418111

---

## Threads answered in conversation — mark resolved

The following threads have satisfactory responses in the comment chain but were not marked resolved. Nevertheless review them and possibly make amends to the documentation:

| Thread | Summary | Answered by |
|--------|---------|-------------|
| [r3212259341](https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3212259341) | Ordering applies to notification emission, not internal processing | @Nana-EC ("You say this well at the end of Goal 3") |
| [r3277465347](https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277465347) | Document describes interim stage, not final form | @jsync-swirlds |
| [r3277509446](https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277509446) | Stalled/incomplete blocks handled by manager cancellation + failed notification | @jsync-swirlds |
| [r3277516740](https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277516740) | Publisher stall detection is independent of verification ordering | @jsync-swirlds |
| [r3277560130](https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277560130) | Success details — policy is fail on any failed proof; never strip proofs | @jsync-swirlds |
| [r3277567080](https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277567080) | CANCELLED covers stalled blocks regardless of detection mechanism | @jsync-swirlds |
| [r3277570277](https://github.com/hiero-ledger/hiero-block-node/pull/2701#discussion_r3277570277) | Missing RSA keys and missing TSS parameters already added as separate failure types | @ata-nas |

ata-nas:
* Updated design docs for verification
* Updated javadocs for verification
* New design doc for block stream forward compatibility
* Added a couple of try-catches to improve stability

Resolves #2879

### DIFF
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
@@ -145,16 +145,21 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
     /// This is also an important assumption for setting the last verified block initially.
     @Override
     public void onContextUpdate(final BlockNodeContext updatedContext) {
-        if (updatedContext != null) {
-            verificationDataProvider.safeUpdateTssData(updatedContext.tssData(), false);
-            safeUpdateLastVerified(updatedContext.storedBlocks());
+        try {
+            if (updatedContext != null) {
+                verificationDataProvider.safeUpdateTssData(updatedContext.tssData(), false);
+                updateLastVerifiedBlock(updatedContext.storedBlocks());
+            }
+        } catch (final RuntimeException e) {
+            LOGGER.log(WARNING, "onContextUpdate failed", e);
         }
     }
 
-    /// Safe update the last verified block.
+    /// Update the last verified block.
     /// If the latest update from application state has a high watermark for stored blocks
     /// higher than what we last verified, we want to roll forward.
-    private void safeUpdateLastVerified(final List<BlockRange> storedBlocks) {
+    /// We expect updates here to happen very infrequently.
+    private void updateLastVerifiedBlock(final List<BlockRange> storedBlocks) {
         if (storedBlocks != null && !storedBlocks.isEmpty()) {
             final BlockRange lastRange = storedBlocks.getLast();
             final long highestStoredBlock = lastRange.rangeEnd();
@@ -163,6 +168,7 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
                 lastVerifiedBlock.compareAndSet(localLastVerified, highestStoredBlock);
                 localLastVerified = lastVerifiedBlock.get();
             }
+            LOGGER.log(INFO, "onContextUpdate received, updated last verified block to {0}", lastVerifiedBlock.get());
         }
     }
 
@@ -258,12 +264,21 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
     /// @param notification a [PersistedNotification] received as an event.
     @Override
     public void handlePersisted(final PersistedNotification notification) {
-        if (notification != null && !notification.succeeded()) {
-            recentlyVerifiedBlocks.remove(notification.blockNumber());
+        try {
+            if (notification != null && !notification.succeeded()) {
+                recentlyVerifiedBlocks.remove(notification.blockNumber());
+            }
+        } catch (final RuntimeException e) {
+            LOGGER.log(WARNING, "Failed to handle persisted notification in verification ", e);
         }
     }
 
-    /// Send a notification to messaging.
+    /// Send a failure notification to messaging. Any failure to send is logged
+    /// and swallowed; this method never throws.
+    ///
+    /// @param blockNumber the number of the block the failure is for
+    /// @param blockSource the source of the block
+    /// @param sessionFailureType the failure type to report
     private void safeSendNotification(
             final long blockNumber, final BlockSource blockSource, final SessionFailureType sessionFailureType) {
         try {
@@ -282,6 +297,13 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
         }
     }
 
+    /// Validate the start of a block. When the supplied items mark the start of a
+    /// new block, the first item must be a block header and the header's number
+    /// must match the block number announced with the items. Items that do not
+    /// mark the start of a block are always considered valid here.
+    ///
+    /// @param blockItems the block items to validate
+    /// @return `true` if the items are valid to process, `false` otherwise
     private boolean validateStartOfBlock(final BlockItems blockItems) {
         try {
             final boolean result;
@@ -305,6 +327,9 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
         }
     }
 
+    /// Resolve the local hostname, used as the Block Node identity in bad block dumps.
+    ///
+    /// @return the local hostname, or `"unknown"` when it cannot be resolved
     private String resolveHostname() {
         try {
             return InetAddress.getLocalHost().getHostName();
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
@@ -151,7 +151,7 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
                 updateLastVerifiedBlock(updatedContext.storedBlocks());
             }
         } catch (final RuntimeException e) {
-            LOGGER.log(WARNING, "onContextUpdate failed", e);
+            LOGGER.log(INFO, "onContextUpdate failed", e);
         }
     }
 
@@ -164,11 +164,15 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
             final BlockRange lastRange = storedBlocks.getLast();
             final long highestStoredBlock = lastRange.rangeEnd();
             long localLastVerified = lastVerifiedBlock.get();
+            boolean updateHappened = false;
             while (highestStoredBlock > localLastVerified) {
-                lastVerifiedBlock.compareAndSet(localLastVerified, highestStoredBlock);
+                updateHappened = lastVerifiedBlock.compareAndSet(localLastVerified, highestStoredBlock);
                 localLastVerified = lastVerifiedBlock.get();
             }
-            LOGGER.log(INFO, "onContextUpdate received, updated last verified block to {0}", lastVerifiedBlock.get());
+            if (updateHappened) {
+                final String message = "onContextUpdate received, updated last verified block to {0}";
+                LOGGER.log(INFO, message, lastVerifiedBlock.get());
+            }
         }
     }
 
@@ -269,7 +273,7 @@ public final class VerificationServicePlugin implements BlockNodePlugin, BlockIt
                 recentlyVerifiedBlocks.remove(notification.blockNumber());
             }
         } catch (final RuntimeException e) {
-            LOGGER.log(WARNING, "Failed to handle persisted notification in verification ", e);
+            LOGGER.log(INFO, "Failed to handle persisted notification in verification ", e);
         }
     }
 
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
@@ -24,18 +24,36 @@ import org.hiero.block.node.block.verification.session.SessionFailureType;
 
 /// State proof verifier.
 public final class StateProofVerifier implements ProofVerifier {
+    /// Logger for the verifier.
     private static final Logger LOGGER = System.getLogger(StateProofVerifier.class.getName());
+    /// Cancellation flag shared with the owning session.
     private final AtomicBoolean isCanceled;
+    /// Metrics for proof verification results.
     private final ProofVerificationMetrics proofVerificationMetrics;
+    /// The number of the block being verified, used for logging.
     private final long blockNumber;
+    /// The state proof to verify.
     private final StateProof stateProof;
+    /// The computed root hash of the block being verified.
     private final Bytes rootHash;
+    /// Provider of the TSS data required for the final signature verification.
     private final VerificationDataProvider verificationDataProvider;
+    /// Pending merge checkpoints, keyed by join point index. See [MergeCheckpoint].
     private final HashMap<Integer, MergeCheckpoint> checkpoints;
+    /// Set to `true` once a hash leaf matching the computed block root hash has been seen.
     private boolean foundComputedRootHashMatch = false;
+    /// The reconstructed signed block root, set once the final join has been reached.
     private byte[] signedBlockRoot = null;
 
     /// Constructor.
+    ///
+    /// @param isCanceled cancellation flag shared with the owning session, must not be null
+    /// @param proofVerificationMetrics metrics for proof verification results, must not be null
+    /// @param blockNumber the number of the block being verified
+    /// @param stateProof the state proof to verify, must not be null
+    /// @param rootHash the computed root hash of the block being verified, must not be null
+    /// @param verificationDataProvider provider of the TSS data for the final signature
+    ///     verification, must not be null
     public StateProofVerifier(
             final AtomicBoolean isCanceled,
             final ProofVerificationMetrics proofVerificationMetrics,
@@ -52,7 +70,32 @@ public final class StateProofVerifier implements ProofVerifier {
         this.checkpoints = new HashMap<>();
     }
 
-    /// todo(2879) add documentation
+    /// {@inheritDoc}
+    /// ---
+    /// Verifies a state proof (an indirect proof) for the block.
+    ///
+    /// A state proof carries a list of [MerklePath]s: *leaves* (paths with
+    /// content) and *join points* (paths without content). The verifier walks
+    /// every leaf, merges its siblings, and follows its next path index through
+    /// the join points, merging partial results via [MergeCheckpoint]s, until the
+    /// *signed block root* is reconstructed at the final join (next index below
+    /// zero). Along the way it checks that one of the hash leaves matches the
+    /// computed block root hash, which ties the proof to the block content.
+    ///
+    /// The proof is rejected when the structure is malformed: fewer than three
+    /// paths, a non-leaf first path, an out-of-range next index, a leaf pointing
+    /// to another leaf, a duplicate checkpoint, unvisited paths, or leftover
+    /// checkpoints after the walk.
+    ///
+    /// Once reconstructed, the signed block root is verified with a [TSSVerifier]
+    /// against the signature carried by the proof. The success or failure state
+    /// proof series of the proof verification metrics is incremented accordingly
+    /// before returning.
+    ///
+    /// @return `null` if the proof verifies, [SessionFailureType#CANCELLED] when
+    ///     the session was cancelled mid-walk, or
+    ///     [SessionFailureType#BAD_BLOCK_PROOF] when the proof is malformed, does
+    ///     not match the computed root hash, or its signature does not verify
     @Override
     public SessionFailureType verify() {
         final SessionFailureType result;
@@ -259,6 +302,7 @@ public final class StateProofVerifier implements ProofVerifier {
         }
     }
 
+    /// Returns `true` if the owning session has been cancelled or the current thread interrupted.
     private boolean isCanceled() {
         return isCanceled.get() || Thread.currentThread().isInterrupted();
     }
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
@@ -85,12 +85,15 @@ public final class StateProofVerifier implements ProofVerifier {
     /// The proof is rejected when the structure is malformed: fewer than three
     /// paths, a non-leaf first path, an out-of-range next index, a leaf pointing
     /// to another leaf, a duplicate checkpoint, unvisited paths, or leftover
-    /// checkpoints after the walk.
+    /// checkpoints after the walk. Note that the non-leaf first path check may be
+    /// relaxed in the future: a state proof creator might order the paths
+    /// slightly differently and still produce a valid state proof.
     ///
-    /// Once reconstructed, the signed block root is verified with a [TSSVerifier]
-    /// against the signature carried by the proof. The success or failure state
-    /// proof series of the proof verification metrics is incremented accordingly
-    /// before returning.
+    /// Once reconstructed, the signed block root is verified against the
+    /// signature carried by the proof. A signature verifier is required for this
+    /// step; which verifier depends on the signature scheme in use. The success
+    /// or failure state proof series of the proof verification metrics is
+    /// incremented accordingly before returning.
     ///
     /// @return `null` if the proof verifies, [SessionFailureType#CANCELLED] when
     ///     the session was cancelled mid-walk, or
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
@@ -11,12 +11,21 @@ import org.hiero.block.node.block.verification.session.SessionFailureType;
 
 /// TSS based proof verifier.
 public final class TSSVerifier implements ProofVerifier {
+    /// Metrics for proof verification results.
     private final ProofVerificationMetrics proofVerificationMetrics;
+    /// The hash the signature must verify against, usually the computed block root hash.
     private final Bytes hashToVerify;
+    /// The TSS signature to verify.
     private final Bytes signature;
+    /// Provider of the TSS data (ledger id) required for verification.
     private final VerificationDataProvider verificationDataProvider;
 
     /// Constructor.
+    ///
+    /// @param proofVerificationMetrics metrics for proof verification results, must not be null
+    /// @param hashToVerify the hash the signature must verify against, must not be null
+    /// @param signature the TSS signature to verify, must not be null
+    /// @param verificationDataProvider provider of the TSS data required for verification, must not be null
     public TSSVerifier(
             final ProofVerificationMetrics proofVerificationMetrics,
             final Bytes hashToVerify,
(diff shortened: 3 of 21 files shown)


docs/design/block-stream-forward-compatibility.md
ata-nas, line 1: This design doc is created based on the description of #3195.

docs/design/block-stream-forward-compatibility.md
```diff
@@ -0,0 +1,405 @@
...
+
+## Design
+
+### The structure of the block stream
+
+Each item in the block stream **SHALL** be self-contained and independent, with
+the following constraints applicable to the *unfiltered* stream:
+
+- A block **SHALL** start with a `header`.
+- A block **SHALL** end with a `state_proof`.
```
ata-nas: Is this correct? Based on #3195 description. Do we mean `block_proof` here?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -52,7 +70,32 @@ public StateProofVerifier(
...
+    ///
+    /// Once reconstructed, the signed block root is verified with a [TSSVerifier]
+    /// against the signature carried by the proof. The success or failure state
+    /// proof series of the proof verification metrics is incremented accordingly
+    /// before returning.
+    ///
+    /// @return `null` if the proof verifies, [SessionFailureType#CANCELLED] when
+    ///     the session was cancelled mid-walk, or
+    ///     [SessionFailureType#BAD_BLOCK_PROOF] when the proof is malformed, does
+    ///     not match the computed root hash, or its signature does not verify
```
ata-nas, line 101: Is this correct?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
```diff
@@ -28,7 +37,22 @@ public TSSVerifier(
...
+    /// post-genesis (WRAPS) paths. Signatures without a recognized proof suffix
+    /// are rejected by the library.
+    ///
+    /// The success or failure TSS series of the proof verification metrics is
+    /// incremented accordingly before returning.
+    ///
+    /// @return `null` if the signature verifies,
+    ///     [SessionFailureType#MISSING_VERIFICATION_DATA] when no TSS data is
+    ///     available, or [SessionFailureType#BAD_BLOCK_PROOF] when the signature
+    ///     does not verify or is malformed
```
ata-nas, line 57: Is this correct?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -145,16 +145,21 @@ public void stop() {
...
-        if (updatedContext != null) {
-            verificationDataProvider.safeUpdateTssData(updatedContext.tssData(), false);
-            safeUpdateLastVerified(updatedContext.storedBlocks());
+        try {
+            if (updatedContext != null) {
+                verificationDataProvider.safeUpdateTssData(updatedContext.tssData(), false);
+                updateLastVerifiedBlock(updatedContext.storedBlocks());
+            }
+        } catch (final RuntimeException e) {
+            LOGGER.log(WARNING, "onContextUpdate failed", e);
```
ata-nas: Added try-catch so we do not propagate exceptions here.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -258,12 +264,21 @@ public void handleBackfilled(final BackfilledBlockNotification notification) {
...
     @Override
     public void handlePersisted(final PersistedNotification notification) {
-        if (notification != null && !notification.succeeded()) {
-            recentlyVerifiedBlocks.remove(notification.blockNumber());
+        try {
+            if (notification != null && !notification.succeeded()) {
+                recentlyVerifiedBlocks.remove(notification.blockNumber());
+            }
+        } catch (final RuntimeException e) {
+            LOGGER.log(WARNING, "Failed to handle persisted notification in verification ", e);
```
ata-nas: Added try-catch so we do not propagate exceptions here.

jsync-swirlds: We should generally propagate the exception; catch-to-log, particularly for runtime exception, is not a good policy at any point below the top-level of a module where decisions of plugin health can be made.

docs/design/block-stream-forward-compatibility.md
jsync-swirlds: This should say `block_proof`; not sure why it says `state_proof`; there might be an error in the proto "javadoc" style comments.

docs/design/block-stream-forward-compatibility.md
```diff
@@ -0,0 +1,405 @@
...
+| 6                   | `state changes`                                      | Yes                 |
+| 7                   | `trace data`                                         | Yes                 |
+| 8                   | Extension 0                                          | Not yet             |
+| 9                   | Extension 1                                          | Not yet             |
+| 10                  | Extension 2                                          | Not yet             |
+| 11                  | Extension 3                                          | Not yet             |
+| 12                  | Extension 4                                          | Not yet             |
+| 13                  | Extension 5                                          | Not yet             |
+| 14                  | Extension 6                                          | Not yet             |
+| 15                  | Extension 7                                          | Not yet             |
```
jsync-swirlds: These are all forward compatible, these "extension" subtrees are already defined, with the names given here.

docs/design/block-stream-forward-compatibility.md
```diff
@@ -0,0 +1,405 @@
...
+category = N modulo 20        (N is the actual field number)
+```
+
+The resulting category number means:
+
+| Category (N mod 20) |                       Meaning                        | Forward compatible? |
+|---------------------|------------------------------------------------------|---------------------|
+| 0                   | Not hashed (not part of the block proof merkle tree) | Yes                 |
+| 1                   | Requires specific handling                           | No                  |
+| 2                   | Requires specific handling                           | No                  |
```
jsync-swirlds: This is partially forward-compatible in that it clearly expresses a field that is not compatible if the field schema is not separately known.

docs/design/block-stream-forward-compatibility.md
```diff
@@ -0,0 +1,405 @@
...
+  and it contributes nothing to the block hash. Also *fully forward compatible*.
+- **Categories 1 and 2** are for items that need specific handling that the rule
+  alone cannot express. These are *not forward compatible* and require an update.
+- **Categories 3 to 7** are the five subtrees that already exist. A future item
+  here is hashed into that subtree exactly like any other item, with no code
+  change. This is *full forward compatibility*.
+- **Categories 8 to 15** are `extension categories`, reserved for subtrees that do
+  not exist yet. The category is known, but there is nowhere to fold the item into
+  the block root hash. See
+  [Where forward compatibility ends](#where-forward-compatibility-ends).
```
jsync-swirlds: These subtrees do exist.

docs/design/block-stream-forward-compatibility.md
```diff
@@ -0,0 +1,405 @@
...
+
+### Where Forward Compatibility Ends
+
+Forward compatibility is complete for categories 3 to 7 (the existing subtrees)
+and for categories 0 and 19 (not hashed). It **cannot** be complete for
+`extension categories` (8 to 15). The block root hash is built from a fixed set of
+category leaves, so a subtree that has not been added yet has no leaf, and an item
+in an extension category cannot be added to the hash without an upgrade that adds
+the subtree and extends the tree. The same is true for the "requires specific
+handling" categories (1 and 2) and the reserved categories (16 to 18).
```
jsync-swirlds:
```suggestion
Forward compatibility is complete for categories 3 to 15 (the defined subtrees)
and for categories 0 and 19 (not hashed). It **cannot** be complete for
`reserved categories` (16 to 18). The block root hash is built from a fixed set of
category leaves that cannot ever be extended, so an item
in a reserved category cannot be added to the hash without an upgrade that defines the correct subtree.
The "requires specific handling" categories (1 and 2) have defined subtrees for each specific field,
but may also require specific preprocessing or other handling, so while compatibility
is version-dependent, the exact processing needed is always either known or the
block cannot be processed. There is never ambiguity for categories 1 and 2.
```

docs/design/block-stream-forward-compatibility.md
```diff
@@ -0,0 +1,405 @@
...
+
+```mermaid
+flowchart TD
+    A[Block item received] --> B{Is this a known item type?}
+    B -- yes --> C["Use its established category (from the table above)"]
+    B -- no --> D["category = field number modulo 20"]
+    D --> E{Which category?}
+    E -- "3 to 7: existing subtree" --> F["Hash into that subtree, fully forward compatible"]
+    E -- "0 or 19: not hashed" --> G["Read and ignore, fully forward compatible"]
+    E -- "8 to 15: extension" --> H["Refuse the block, upgrade required"]
```
jsync-swirlds:
```suggestion
    E -- "8 to 15: extension" --> H["Include the item in the correct extension subtree"]
```

docs/design/block-stream-forward-compatibility.md
```diff
@@ -0,0 +1,405 @@
...
+        TDATA[trace_data]
+        BF[block_footer]
+        BP[block_proof]
+    end
+    subgraph Subtrees["Category subtrees"]
+        CHT[Consensus headers subtree]
+        INT[Inputs subtree]
+        OUTT[Outputs subtree]
+        SCT[State changes subtree]
+        TDT[Trace data subtree]
```
jsync-swirlds, line 322: This is missing the 8 subtrees on the "right side" of the merkle mountain top tree.

docs/design/block-stream-forward-compatibility.md
```diff
@@ -0,0 +1,405 @@
...
+   category and produces the same block root hash as before this change.
+2. **New item, existing category.** A block containing a simulated future item
+   whose `field number modulo 20` is 3 to 7 hashes to the same value a node would
+   produce if it had always known that item type.
+3. **New item, not hashed.** A block containing a simulated future item whose
+   `field number modulo 20` is 0 or 19 verifies successfully, with the item left
+   out of every category.
+4. **Extension category refused.** A block containing a simulated future item
+   whose `field number modulo 20` is 8 to 15 (or 1, 2, or 16 to 18) is refused
+   with the designated failure. It is never accepted and never silently dropped.
```
jsync-swirlds: This needs to be corrected for proper handling of subtrees 8-15.

jsync-swirlds (commented):
There are some corrections needed, particularly because subtrees 8-15 are already defined, not "undefined" or "to be added".
The merkle mountain top is not upgradeable, that's why we pre-defined 8 subtrees for future expansion.

docs/design/block-stream-forward-compatibility.md
ata-nas: Yeah, this is taken from the description of the issue, but made no sense to me. Will update.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
jsync-swirlds: TSSVerifier (line 90) isn't permanently true; it may be replaced or supplemented when we enable PQC.  Perhaps just note that a _verifier_ is required without being specific.

non-leaf first path might be a warning rather than "bad proof", as a state proof creator might get the order slightly wrong and still have a valid state proof.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
jsync-swirlds: Yes, we might note that the WRAPS verification key is _also_ required, but we set that once on startup and don't supply it for every call.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
jsync-swirlds: Same as noted below, I think we're better off letting the exception propagate.

docs/design/block-stream-forward-compatibility.md
jsync-swirlds: Description was copied direct from CN proto comments, so there's probably an error in that comment.

jsync-swirlds (commented):
Looks like this is mostly javadoc improvements, which is a good update.
There are a couple items noted in markdown and the small code changes.

docs/design/block-stream-forward-compatibility.md
ata-nas: Added.

docs/design/block-stream-forward-compatibility.md
ata-nas: Updated.

docs/design/block-stream-forward-compatibility.md
ata-nas: Updated.

docs/design/block-stream-forward-compatibility.md
ata-nas: Updated.

docs/design/block-stream-forward-compatibility.md
ata-nas: Updated.

docs/design/block-stream-forward-compatibility.md
ata-nas: Updated.

docs/design/block-stream-forward-compatibility.md
ata-nas: Updated. The reason I've put "not yet" is because the code does not support it yet. But this is a design doc, so should not reflect (at the most just note) the current state of the code. Hence also the mention of rejection if we see these.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: This is the top level of the module. I've added try-catch here and in the on context handling as we've discussed (I believe we also documented exception handling practices) that plugins should not propagate exceptions.

jsync-swirlds: Ahh, missed that this is a notification method.
It's odd to have Verification responding to persistence, to be fair, given that persistence responds to Verification, so we set ourselves up for another notification queue deadlock.

Couldn't we just keep recently verified blocks as a pure recency check (so it is self-contained and we reduce messaging overheads) rather than adding more complexity by trying to remove entries based on persistence or other signals?

We might have discussed this; but seeing it fresh it looks less than ideal (and adds more gating subscriptions to the overtaxed notification queue).

We probably need to discuss changes in messaging to relieve some of that pressure.

jsync-swirlds (dismissed):

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: So, we can rethink this. 
The reason we handle persistence notifications is to remove recently verified blocks if they fail. If persistence fails on block 5, and it was recently verified, subsequent failure of block 5 should not be treated as informational. That was what was discussed.

In any case, publisher and backfill have to nevertheless handle failed verification and request the failed block again, so the block will not be skipped and a gap be introduced. Verification is very reliant on no gaps being introduced because of strict ordering at high water mark.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
ata-nas: Noted as per your recommendation. Note that the key is either present or startup or not, based on tss bootrstrap file being present or not. If not, then tss data, along with setting the WRAPS key, will be initialized on block 0. I've added that note as well.

3d0364944076edba26abcb166cfaeead678c554d

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
ata-nas: Noted that in the future, we can drop the strict check for first path being a leaf (we do that now and reject non-leaf first path). Noted also that a signature verifier is required, not being specific about TSSVerifier.

3d0364944076edba26abcb166cfaeead678c554d

ata-nas:
@jsync-swirlds thanks for your review here, addressed final feedback.
Did you also take a look at the changed verification design doc: https://github.com/hiero-ledger/hiero-block-node/pull/3251/changes#diff-6aef8bed378e75c4d784f04975fb23b5f274d46d4d3b255628cd6a1903abd064

Did not see you comment on that, but maybe you have no objections or feedback to add.

jsync-swirlds (dismissed):

jsync-swirlds:
> @jsync-swirlds thanks for your review here, addressed final feedback. Did you also take a look at the changed verification design doc: https://github.com/hiero-ledger/hiero-block-node/pull/3251/changes#diff-6aef8bed378e75c4d784f04975fb23b5f274d46d4d3b255628cd6a1903abd064
> 
> Did not see you comment on that, but maybe you have no objections or feedback to add.

I generally read every file, at least I try to.  If I did not have any feedback on a file, it is because I did not see anything needing change before merging this PR.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -163,6 +168,7 @@ private void safeUpdateLastVerified(final List<BlockRange> storedBlocks) {
                 lastVerifiedBlock.compareAndSet(localLastVerified, highestStoredBlock);
                 localLastVerified = lastVerifiedBlock.get();
             }
+            LOGGER.log(INFO, "onContextUpdate received, updated last verified block to {0}", lastVerifiedBlock.get());
```
Nana-EC: I believe this fires on every `onContextUpdate` call that supplies at least one `BlockRange`, regardless of whether `lastVerifiedBlock` was actually advanced.
Should we guard this to only log on updates?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
Nana-EC: Q: DO we have test coverage for this path?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
Nana-EC: Q: DO we have test coverage for this path?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
Nana-EC, line 212: With the new logs you added the previous log level here becomes inconsistent. Should this be at a warning level?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -145,16 +145,21 @@ public void stop() {
...
         }
     }
 
-    /// Safe update the last verified block.
+    /// Update the last verified block.
     /// If the latest update from application state has a high watermark for stored blocks
     /// higher than what we last verified, we want to roll forward.
-    private void safeUpdateLastVerified(final List<BlockRange> storedBlocks) {
+    /// We expect updates here to happen very infrequently.
+    private void updateLastVerifiedBlock(final List<BlockRange> storedBlocks) {
```
Nana-EC, line 162: Q: Was the naming convention here to assume callers that this method doesn't throw?
If so you may want to confirm that consumer of it that had that assumption now have the right handling logic.

Nana-EC (dismissed):
LG, non blocking test comments and a suggestion for validation based on renaming behaviour change.

AlfredoG87 (dismissed):
LGTM

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: You are right, this log is only intended to happen if we've made an update here. Will change as this will be very frequent otherwise.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: Yes, the convention of the `safe` prefix is that we catch all runtime exceptions (that are reasonable) and if an exception happens here, it indicates an issue that must be investigated. This is not hte case here, however, hence the rename. The caller of this will catch runtime exceptions as it is top level plugin handler method.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: This should be an infomrational log, we do not expect to ever see this, but if we do, we want visibility.
If by other logs you mean in persistence and on context update, you are right and a good catch. Those are also informational, will change.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: ✅ Updated.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: This is the top level catch all for a handler method. It is probably not suitable to have just for coverage sake.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: This is the top level catch all for a handler method. It is probably not suitable to have just for coverage sake.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
ata-nas: ✅ Done.

ata-nas:
> > @jsync-swirlds thanks for your review here, addressed final feedback. Did you also take a look at the changed verification design doc: https://github.com/hiero-ledger/hiero-block-node/pull/3251/changes#diff-6aef8bed378e75c4d784f04975fb23b5f274d46d4d3b255628cd6a1903abd064
> > Did not see you comment on that, but maybe you have no objections or feedback to add.
> 
> I generally read every file, at least I try to. If I did not have any feedback on a file, it is because I did not see anything needing change before merging this PR.

Thanks, just wanted to confirm, glad to hear no feedback, just did not expect it 😆

jsync-swirlds (approved):

Nana-EC (approved):

PR #3253, opened 2026-07-22
feat(verification): complete block-verification switchover with real TSS signing for tests

AlfredoG87:
## Summary

- Switches the default block-verification plugin from `verification` to `block-verification` in charts, solo-e2e, app build, and suites runtime
- Adds a new `tools-and-tests/block-signing` test library (`org.hiero.block.signing`) that produces real, verifiable RSA and TSS block proofs, replacing the non-functional `SHA384(blockHash)` placeholder
- Wires the simulator and suites to use real TSS signatures and self-provisioning, completing the full E2E path under the new verifier
- Fixes two correctness bugs discovered during E2E validation under real signatures

## Changes

**`block-signing` (new module under `tools-and-tests/`)**
- `BlockSigner` seam: `signBlockProof(blockNumber, blockHash) -> BlockProof` + `verificationMaterial()`
- `RsaBlockSigner`: `SHA384withRSA` WRB record-file proofs; emits `RangedAddressBookHistory` with hex DER X.509 RSA public keys
- `TssBlockSigner`: TSS/hinTS threshold signatures via the genesis Schnorr-aggregate path (no WRAPS SNARK artifacts required); single-node roster; emits matching `TssData` and a `genesisLedgerIdSignedTransaction()` for in-stream self-provisioning from block 0. `createDeterministic()` produces a fixed-seed instance so multiple publishers in one JVM share one roster
- `VerificationMaterial` holder with in-process accessors and JSON bootstrap-file serializers

**`block-verification`**
- `NativeVerificationLibraryLock`: serializes all hinTS/WRAPS native-library calls behind a single process-wide lock. Verification sessions run concurrently on virtual threads but the native lib is a non-thread-safe JVM singleton; concurrent `verifyTSS`/`setAddressBook`/`setCurrentKey` calls raced under load and produced sporadic `BAD_BLOCK_PROOF`. Guards added to `TSSVerifier` and `VerificationDataProvider` (`StateProofVerifier` is covered transitively)
- `SignerTssVerificationTest`: end-to-end through the real `TSSVerifier` + `VerificationDataProvider` over the real `BlockHasher` root hash of a locally built block
- `SignerGenesisProvisioningTest`: proves block-0 ledger-id publication self-provisions the `VerificationDataProvider` in-stream with no bootstrap file

**simulator**
- `CraftBlockStreamManager` creates a shared `TssBlockSigner` (deterministic) and signs real TSS block proofs; block 0 carries a `LedgerIdPublicationHandler` (genesis publication) so the node self-provisions from any publisher's stream
- Multiple publishers share the same roster so duplicate blocks are byte-identical and detectable

**suites**
- `BlockItemBuilderUtils`: shared `TssBlockSigner` produces verifiable proofs; `provisionTssBootstrap()` writes the signer's roster to a temp bootstrap file and points the in-process `BlockNodeApp` at it via `app.state.tssBootstrapFilePath`
- `BlockNodeAPITests`, `BlockNodeCloudStorageTests`, `BlockNodeApiRegressionTest` call `provisionTssBootstrap()` before starting the in-process app
- Re-enabled the two RESEND regression tests (previously `@Disabled`); they pass with the native-library lock in place

**common**
- `HashingUtilities.computeV6SignedPayload(Bytes)`: `BlockHasher` now delegates to it so the signer and verifier share one definition of the version 6 signed payload

**build / charts / docs**
- `block-node/app/build.gradle.kts`: bundled plugins and runtime now reference `:block-verification`
- `tools-and-tests/suites/build.gradle.kts`: runtime dependency swapped to `block-verification`
- Charts (`values.yaml`, `plugin-profile-{all,lfh,rfh}.yaml`, `README.md`) and `docs/block-node/configuration.md`: default `plugins.names` updated to `block-verification`
- `solo-e2e-test`: `solo-deploy-network.sh`, `wrb-differential-test.yaml` use `block-verification`; new `--bn-chart-dir` / `BN_CHART_DIR` opt-in lets solo-e2e install the Block Node chart from this checkout to exercise unreleased chart defaults
- `gradle/modules.properties` and `hiero-dependency-versions/build.gradle.kts`: map and version-constrain `com.hedera.cryptography.hints`

## Test plan

- [x] `block-signing` unit tests: `RsaBlockSignerTest`, `TssBlockSignerTest` (verify output with exact verifier crypto ops + negative controls)
- [x] `block-verification` unit tests: `SignerTssVerificationTest`, `SignerGenesisProvisioningTest`, full suite (596 tests)
- [x] Simulator unit tests: 195 tests including `BlockProofHandlerTest`
- [x] `qualityCheck` clean across all modified modules
- [x] `runAPISuites` (in-process e2e): 16/16 pass under real TSS signatures with the new verifier
- [x] `runSuites` (TestContainers + simulator): BUILD SUCCESSFUL, 0 failures
- [x] Solo-e2e `task up` with `BN_CHART_DIR` set: 115 blocks received, 114 verified, 0 failed with `block-verification` jar


ata-nas:
@AlfredoG87 amazing job here!

I suspect we are dropping #3180 in favor of this one? It seems you have managed to produce signatures with the simulator, so in that case, we should be in no need of remote datasource blocks for e2e tests.

Please note we need to add `VERIFICATION_ALL_SOURCES_REQUIRE_ORDERING: "false"` in `rfh-values.yaml` as a BN config.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
```diff
@@ -39,8 +40,15 @@ public SessionFailureType verify() {
...
-                if (!TSS.verifyTSS(
-                        tssData.ledgerId().toByteArray(), signature.toByteArray(), hashToVerify.toByteArray())) {
+                final boolean verified;
+                // Serialize the native call: the hinTS/WRAPS library is a JVM-wide singleton and is
+                // not thread-safe, so it must never run concurrently with another verify or a TSS
+                // data update (see NativeVerificationLibraryLock).
+                synchronized (NativeVerificationLibraryLock.LOCK) {
+                    verified = TSS.verifyTSS(
+                            tssData.ledgerId().toByteArray(), signature.toByteArray(), hashToVerify.toByteArray());
+                }
```
ata-nas: Interesting observation here. I was under the impression that only `TSS.setAddressBook` and `WRAPSVerificationKey.setCurrentKey` are not thread safe. If so, it is ok to have it as it was set up before. If `TSS.verifyTSS` is not thread safe, we do need to serialize these calls. Are we sure we are not observing a race between setting address book + current key and blocks that are being verified async, do not yet see the update? If that is the case, this is kind of expected, we should explore.

We have to be extra careful, however, as any introduction of locks is dangerous.

tools-and-tests/block-signing/src/main/java/module-info.java
```diff
@@ -0,0 +1,17 @@
...
+/// app module graph; it exists only to drive verification from unit, integration, and E2E tests.
+module org.hiero.block.signing {
+    exports org.hiero.block.signing;
+
+    requires transitive com.hedera.pbj.runtime;
+    requires transitive org.hiero.block.protobuf.pbj;
+    requires com.hedera.cryptography.hints;
+    requires com.hedera.cryptography.wraps;
+    requires org.hiero.block.common;
+    requires static com.github.spotbugs.annotations;
```
ata-nas, line 16: `requires static com.github.spotbugs.annotations;` seems redundant.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/config/data/BlockGeneratorConfig.java
```diff
@@ -28,19 +29,28 @@
...
+
         @Loggable @ConfigProperty(defaultValue = "10") int maxTransactionsPerEvent,
         @Loggable @ConfigProperty(defaultValue = "") String folderRootPath,
         @Loggable @ConfigProperty(defaultValue = "36") int paddedLength,
         @Loggable @ConfigProperty(defaultValue = ".blk.gz") String fileExtension,
         // Optional block number range for the BlockAsFileLargeDataSets manager
-        @Loggable @ConfigProperty(defaultValue = "0") @Min(0) int startBlockNumber,
+        @Loggable @ConfigProperty(defaultValue = "0") @Min(0)
+        int startBlockNumber,
+
```
ata-nas: We can disable spotless for this config class here to improve readability.

ata-nas (commented):
LG, just a few nits.

Please note we need to add `VERIFICATION_ALL_SOURCES_REQUIRE_ORDERING: "false"` in `rfh-values.yaml` as a BN config.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/NativeVerificationLibraryLock.java
```diff
@@ -0,0 +1,27 @@
...
+/// producing sporadic, load-dependent `BAD_BLOCK_PROOF` failures for signatures
+/// that are individually valid.
+///
+/// Every native call in this module (`TSS.verifyTSS`, `TSS.setAddressBook`,
+/// `WRAPSVerificationKey.setCurrentKey`) must hold this single lock. This
+/// serializes verification throughput, an acceptable trade-off given the library
+/// constraint: correctness takes precedence over throughput here.
+public final class NativeVerificationLibraryLock {
+    /// The shared monitor guarding every hinTS/WRAPS native call in this module.
+    public static final Object LOCK = new Object();
```
jsync-swirlds: So, why use a Monitor lock instead of a ReentrantLock (which is significantly safer against common deadlock possibilities)?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationDataProvider.java
```diff
@@ -112,9 +112,15 @@ private void updateTssData(final TssData updatedTssData, final boolean sendUpdat
...
                             weights[i] = contribution.weight();
                         }
-                        TSS.setAddressBook(publicKeys, weights, nodeIds);
-                        WRAPSVerificationKey.setCurrentKey(
-                                updatedTssData.wrapsVerificationKey().toByteArray());
+                        // Serialize the native calls: the hinTS/WRAPS library is a JVM-wide singleton
+                        // and is not thread-safe. The address book and WRAPS key mutate the same shared
+                        // native state that concurrent verifyTSS calls read, so both updates must happen
+                        // together under the shared lock (see NativeVerificationLibraryLock).
+                        synchronized (NativeVerificationLibraryLock.LOCK) {
```
jsync-swirlds: Instead of synchronized, why not use more modern lock support?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/NativeVerificationLibraryLock.java
```diff
@@ -0,0 +1,27 @@
...
+/// executor — so without serialization a verify can race another verify or an
+/// address-book / key update, corrupting the library's shared native state and
+/// producing sporadic, load-dependent `BAD_BLOCK_PROOF` failures for signatures
+/// that are individually valid.
+///
+/// Every native call in this module (`TSS.verifyTSS`, `TSS.setAddressBook`,
+/// `WRAPSVerificationKey.setCurrentKey`) must hold this single lock. This
+/// serializes verification throughput, an acceptable trade-off given the library
+/// constraint: correctness takes precedence over throughput here.
+public final class NativeVerificationLibraryLock {
```
jsync-swirlds: If we're going to have a class for this, why not put methods in this class to isolate the lock management to this class instead of having to wrap calls in other classes in synchronized or lock management?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
jsync-swirlds: Locks in _hot_ code (which this is) are super costly; so we should be very careful with that.
If we can eliminate this specific lock, it would be very good.

Even just using more efficient locks (e.g. ReentrantLock instead of synchronized) would be good.

tools-and-tests/block-signing/src/main/java/org/hiero/block/signing/TssBlockSigner.java
```diff
@@ -0,0 +1,412 @@
...
+    private static final long[] NODE_IDS = {NODE_ID};
+
+    /// Length of a compressed WRAPS proof (the settled-path `abProof`).
+    private static final int COMPRESSED_WRAPS_PROOF_LENGTH = 704;
+
+    /// Classpath resource holding the pre-computed genesis WRAPS proof for the deterministic roster.
+    private static final String WRAPS_PROOF_RESOURCE = "/org/hiero/block/signing/genesis-wraps-proof.bin";
+
+    private static final HintsLibraryBridge HINTS = HintsLibraryBridge.getInstance();
+    private static final WRAPSLibraryBridge WRAPS = WRAPSLibraryBridge.getInstance();
```
jsync-swirlds, line 71: Do we need to protect these two instances with locks as well?

tools-and-tests/block-signing/src/main/java/module-info.java
jsync-swirlds: That's where we get `@Nullable` and `@NonNull` and related.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/config/data/BlockGeneratorConfig.java
jsync-swirlds: Yes, please add that disable and bring this file back to line-per-value.
The extra split out spotless demands on these makes them much harder to read quickly.

Note, that only applies to the record declaration, not the actual functional code below.

Second note, is this "config" ever actually loaded by the configuration library, or just created with the "builder" below?  If it's only ever created with the builder, we can remove all of the "config" annotations and simplify this without disabling spotless.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/utils/BlockItemBuilderUtils.java
```diff
@@ -40,9 +45,36 @@ public final class BlockItemBuilderUtils {
...
+    /// Shared signer used to produce verifiable block proofs. A single instance per JVM keeps the
+    /// hinTS native singleton state stable, and its roster is what [#provisionTssBootstrap] hands to
+    /// the block node.
+    private static volatile TssBlockSigner blockSigner;
+
     // Required to quiet warnings.
     private BlockItemBuilderUtils() {}
 
+    /// Returns the shared [TssBlockSigner], creating it on first use.
+    public static synchronized TssBlockSigner blockSigner() {
```
jsync-swirlds: We do generally try to avoid `synchronized` and use alternatives such as `ReentrantLock` instead.  It's OK here in tools-and-tests, but would still be good if we can avoid it without excess effort.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -106,6 +106,24 @@ public static MessageDigest sha384DigestOrThrow() {
         }
     }
 
+    /**
+     * Computes the version 6 record-file signed payload: {@code SHA-384(int32be(6) || recordFileContents)}.
+     * <p>
+     * This is the exact payload the consensus node signs (and the block node verifies) for a version 6
+     * record-file (WRB) block proof. The {@code int32be(6)} prefix is the four bytes
+     * {@code 0x00 0x00 0x00 0x06}. Keeping this in one place guarantees the signer and the verifier
```
jsync-swirlds: Slight clarification:

```suggestion
     * {@code 0x00 0x00 0x00 0x06}. Keeping this in one place helps to ensure the test signer and the verifier
```

jsync-swirlds (commented):
Thanks for working on this, it unblocks a lot.

There are a few details that can be improved but directionally this is good.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
ata-nas: I am not convinced we need a lock at all. In testing I've done, never seen an issue running multiple racing sessions that call tss verify. What I am almost certain we are observing is sessions calling tss verify before tss data is there or in the middle of update. That is something I've shared a few times and is expected. If the bn starts up with tss data available, we should see no issues. I think tests send block 0 which also has the ledger publication, along with more blocks. Those other blocks race and do not see the tss data initialized yet before they call tss verify. We have to explore of that is the issue. If so, then we can drop all locks.

ata-nas: P.s. in unit tests I've done, I make sure block 0 passes so tss data is there, then I run multiple blocks simultaneously sync.

AlfredoG87: Explored this. The `BAD_BLOCK_PROOF` failures showed up during steady-state streaming of resent / gap-fill / multi-publisher blocks — well after genesis. In these tests the address book and WRAPS key are installed once at block 0 and never change again, so those verifies were not racing a concurrent `setAddressBook` / `setCurrentKey`. The only concurrency present was `verifyTSS` against `verifyTSS`. That points to `verifyTSS` itself not being safe to run concurrently, so we do need to serialize the verifies — a read/write lock that allowed concurrent verifies would reintroduce the failures.

We kept full mutual exclusion but moved it off `synchronized`: it now lives behind a single `ReentrantLock` owned by `NativeVerificationLibrary`. Acquisition is uncontended-cheap and negligible next to the native verify it guards, and contention only occurs when verifies genuinely overlap, which the library requires us to serialize anyway.

tools-and-tests/block-signing/src/main/java/module-info.java
AlfredoG87: Kept. As noted, that's where `@NonNull` / `@Nullable` come from, and they're used across `BlockSigner`, `RsaBlockSigner`, `TssBlockSigner`, and `VerificationMaterial`. `requires static` (compile-only) is the correct scope.

tools-and-tests/simulator/src/main/java/org/hiero/block/simulator/config/data/BlockGeneratorConfig.java
AlfredoG87: Done, wrapped the record header in `// spotless:off` / `// spotless:on` and put each component back on one line. On the second note: this record is loaded by the config library (`@ConfigData("generator")`, and `generator.tssProofType` maps from `GENERATOR_TSS_PROOF_TYPE` in `SimulatorMappedConfigSourceInitializer`), so the `@ConfigProperty` annotations stay.

✅ Done!

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/NativeVerificationLibraryLock.java
AlfredoG87: Done. Replaced the exposed `public static final Object LOCK` monitor with a private `ReentrantLock` owned by the class, renamed it `NativeVerificationLibraryLock` -> `NativeVerificationLibrary`, and moved the native calls behind two methods — `verifyTss(...)` and `updateVerificationKeys(...)`. Callers (`TSSVerifier`, `VerificationDataProvider`) no longer wrap anything in `synchronized` or manage a lock; all lock management is isolated in the one class.

✅ Done!

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationDataProvider.java
AlfredoG87: Done via the same change — the `synchronized` block is gone; it now calls `NativeVerificationLibrary.updateVerificationKeys(...)`.

✅ Done!

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/NativeVerificationLibraryLock.java
AlfredoG87: ✅  Done!

tools-and-tests/suites/src/main/java/org/hiero/block/suites/utils/BlockItemBuilderUtils.java
AlfredoG87: Changed to Reentrant Lock 

✅ Done!

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
AlfredoG87: ✅  Done!

AlfredoG87:
@ata-nas 

> I suspect we are dropping https://github.com/hiero-ledger/hiero-block-node/pull/3180 in favor of this one? It seems you have managed to produce signatures with the simulator, so in that case, we should be in no need of remote datasource blocks for e2e tests.

Not really, we still need that feature, but this buys us time for now.

tools-and-tests/block-signing/src/main/java/org/hiero/block/signing/TssBlockSigner.java
AlfredoG87: ✅ Done!

Added a shared static `ReentrantLock` in `TssBlockSigner` guarding the key ceremony (all factory paths) and `signBlockProof`, so the native singletons are never touched concurrently.

charts/block-node-server/values-overrides/rfh-values.yaml
```diff
@@ -5,6 +5,8 @@ blockNode:
         # Change the following Value with the next expected Block to be
         # published by the Network.
         # BLOCK_NODE_EARLIEST_MANAGED_BLOCK: "0"
+        # Only enforce strict next-block ordering for the publisher source.
+        VERIFICATION_ALL_SOURCES_REQUIRE_ORDERING: "false"
```
jsync-swirlds, line 9: Not sure what added this, but RFH doesn't have a publisher, so this setting effectively removes verification ordering entirely. I think we might want to retain that ordering in RFH to reduce temporary tar file use.

jsync-swirlds (dismissed):

charts/block-node-server/values-overrides/rfh-values.yaml
ata-nas: Being able to set this to false in RHF is exactly why we added this config in the first place. We did not want to limit and impose order on RFH.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/TSSVerifier.java
ata-nas: Thanks for exploring and confirming this! Also, thank you for introducing a reentrant lock.

ata-nas (dismissed):
Great job here @AlfredoG87!

jsync-swirlds (approved):

Nana-EC (approved):

Issue #2276, opened 2026-02-26
`Files Historic Plugin` Sends Persisted Notifications on Staged Blocks and not on Zips


ivannov:
### Story Form

As a Files Historic Plugin
I want to Send a Persisted Notification when I stage a block
So that I will notify when I have a block safely stored in my staging area

### Technical Notes

- We currently send Persisted Notification when we zip a whole batch with the end of the batch as a block number
- We want to send these notifications every time we stage a block instead
- We want this feature to be configurable, meaning if we will or will not send Persisted Notifications on every staged block

PR #3256, opened 2026-07-22
fix: Edge case persistence exceptions do not push a persistence notification

### LINKED ISSUE #3252: Edge case persistence exceptions do not push a persistence notification
### Description

In some corner cases, persistence failure notifications are not sent to consensus nodes.
Specifically: unchecked exceptions and IOExceptions are an area that don't always trigger a failed notification.

We need to investigate further and close this gap for files recent and historical

### Steps to Reproduce

1. Observe a scenario that cases an unchecked exception or an IO exception

### Additional context

_No response_

ivannov:
### `BlockFileHistoricPlugin`: every early-return path in `writeBlockToStagingPath` now sends a failed notification

Previously, a null/empty block, a missing block header, or a block-number mismatch caused the method to silently return without notifying anyone. Each of those checks now sends a new persisted notification returning

### `createDirectoryOrFail` and `writeBlockOrFail` no longer let exceptions escape uncaught

`createDirectoryOrFail` used to throw `UncheckedIOException` on a failed `Files.createDirectories`; it now returns a boolean, logs the failure, and sends a failed notification itself. The caller only proceeds to `writeBlockOrFail` if directory creation succeeded. `writeBlockOrFail`'s existing `IOException` catch now also sends a failed notification (it previously only logged). The whole call chain in `writeBlockToStagingPath` is additionally wrapped in a `catch (RuntimeException e)` so any unchecked exception we did not anticipate still results in a failed notification rather than being swallowed or crashing the handler thread. Same approach was applied to `BlockFileRecentPlugin`.

## Related Issue(s)

Closes #3252

### DIFF
--- a/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
+++ b/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
@@ -18,7 +18,6 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import java.io.BufferedOutputStream;
 import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.nio.file.FileVisitResult;
 import java.nio.file.FileVisitor;
 import java.nio.file.Files;
@@ -324,7 +323,7 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
     @Override
     public void handleVerification(VerificationNotification notification) {
         if (notification != null && notification.success()) {
-            writeBlockToStagingPath(notification.block(), notification.blockNumber());
+            writeBlockToStagingPath(notification.block(), notification.blockNumber(), notification.source());
         }
         try {
             attemptZipping();
@@ -335,14 +334,23 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
     }
 
     // ==== Private Methods ============================================================================================
-    private void writeBlockToStagingPath(final BlockUnparsed block, final long blockNumber) {
+    private void sendBlockNotification(final long blockNumber, final boolean succeeded, final BlockSource source) {
+        context.blockMessaging()
+                .sendBlockPersisted(new PersistedNotification(
+                        blockNumber, succeeded, defaultPriority(), source == null ? BlockSource.UNKNOWN : source));
+    }
+
+    private void writeBlockToStagingPath(final BlockUnparsed block, final long blockNumber, final BlockSource source) {
         if (block == null || block.blockItems() == null || block.blockItems().isEmpty()) {
+            LOGGER.log(WARNING, "Block {0} is null or has no block items, cannot write to staging path", blockNumber);
+            sendBlockNotification(blockNumber, false, source);
             return;
         }
 
         BlockItemUnparsed firstItem = block.blockItems().getFirst();
         if (!firstItem.hasBlockHeader()) {
             LOGGER.log(WARNING, "Block {0} has no block header, cannot write to staging path", blockNumber);
+            sendBlockNotification(blockNumber, false, source);
             return;
         }
 
@@ -354,13 +362,21 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
                     "Block number mismatch between notification {0} and block header {1}, not writing block",
                     blockNumber,
                     headerNumber);
+            sendBlockNotification(blockNumber, false, source);
             return;
         }
 
-        final Path verifiedBlockPath = BlockFile.nestedDirectoriesBlockFilePath(
-                stagingPath, blockNumber, config.compression(), config.maxFilesPerDir());
-        createDirectoryOrFail(verifiedBlockPath);
-        writeBlockOrFail(block, blockNumber, verifiedBlockPath);
+        try {
+            final Path verifiedBlockPath = BlockFile.nestedDirectoriesBlockFilePath(
+                    stagingPath, blockNumber, config.compression(), config.maxFilesPerDir());
+            if (createDirectoryOrFail(verifiedBlockPath, blockNumber, source)) {
+                writeBlockOrFail(block, blockNumber, source, verifiedBlockPath);
+            }
+        } catch (final RuntimeException e) {
+            final String message = "Failed to persist block %d due to %s".formatted(blockNumber, e);
+            LOGGER.log(WARNING, message, e);
+            sendBlockNotification(blockNumber, false, source);
+        }
     }
 
     private BlockHeader getBlockHeader(final BlockUnparsed block) {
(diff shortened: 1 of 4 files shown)


block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -397,6 +417,7 @@ private void writeBlockOrFail(final BlockUnparsed block, final long blockNumber,
         } catch (final IOException e) {
             final String message = "Failed to write file for block %d due to %s".formatted(blockNumber, e);
             LOGGER.log(WARNING, message, e);
+            sendBlockNotification(blockNumber, false, source);
```
jsync-swirlds: Instead of sending the notification here, why not just return a boolean for success/fail and let the caller send notification?
Much like `createDirectoryOrFail` above?

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -366,14 +366,22 @@ private void writeBlockToLivePath(final BlockUnparsed block, final long blockNum
                 } else {
                     final Path verifiedBlockPath = RecentBlockPath.computeBlockPath(config, blockNumber)
                             .path();
-                    createDirectoryOrFail(verifiedBlockPath);
-                    writeBlockOrFail(block, blockNumber, effectiveSource, verifiedBlockPath);
+                    try {
+                        if (createDirectoryOrFail(verifiedBlockPath, blockNumber, effectiveSource)) {
+                            writeBlockOrFail(block, blockNumber, effectiveSource, verifiedBlockPath);
+                        }
```
jsync-swirlds: else case here should send failure notification.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -416,15 +424,18 @@ private void writeBlockOrFail(
...
         try {
             // create parent directory if it does not exist
             Files.createDirectories(verifiedBlockPath.getParent());
+            return true;
         } catch (final IOException e) {
             final String message = "Failed to create directories for path %s due to %s"
                     .formatted(verifiedBlockPath.toAbsolutePath(), e);
             LOGGER.log(WARNING, message, e);
-            throw new UncheckedIOException(e);
+            sendBlockNotification(blockNumber, false, source);
```
jsync-swirlds: Don't send notification here, return false and the _caller_ should send notification.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -366,14 +366,22 @@ private void writeBlockToLivePath(final BlockUnparsed block, final long blockNum
                 } else {
                     final Path verifiedBlockPath = RecentBlockPath.computeBlockPath(config, blockNumber)
                             .path();
-                    createDirectoryOrFail(verifiedBlockPath);
-                    writeBlockOrFail(block, blockNumber, effectiveSource, verifiedBlockPath);
+                    try {
+                        if (createDirectoryOrFail(verifiedBlockPath, blockNumber, effectiveSource)) {
+                            writeBlockOrFail(block, blockNumber, effectiveSource, verifiedBlockPath);
```
jsync-swirlds: I believe we would be better off having `writeBlockOrFail` return true/false and send notification based on that return here, rather than expecting the method to send notification.

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -354,13 +362,25 @@ private void writeBlockToStagingPath(final BlockUnparsed block, final long block
...
-                stagingPath, blockNumber, config.compression(), config.maxFilesPerDir());
-        createDirectoryOrFail(verifiedBlockPath);
-        writeBlockOrFail(block, blockNumber, verifiedBlockPath);
+        try {
+            final Path verifiedBlockPath = BlockFile.nestedDirectoriesBlockFilePath(
+                    stagingPath, blockNumber, config.compression(), config.maxFilesPerDir());
+            if (createDirectoryOrFail(verifiedBlockPath)) {
+                if (!writeBlockOrFail(block, blockNumber, verifiedBlockPath)) {
+                    sendBlockNotification(blockNumber, false, source);
+                }
```
jsync-swirlds, line 375: Do we never send success notifications?
I would have expected something like the following (and I might be incorrect, in which case perhaps a comment would help to explain why):

```suggestion
                final boolean successful = writeBlockOrFail(block, blockNumber, verifiedBlockPath);
                sendBlockNotification(blockNumber, successful, source);
```

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -366,14 +366,28 @@ private void writeBlockToLivePath(final BlockUnparsed block, final long blockNum
...
                             .path();
-                    createDirectoryOrFail(verifiedBlockPath);
-                    writeBlockOrFail(block, blockNumber, effectiveSource, verifiedBlockPath);
+                    try {
+                        if (createDirectoryOrFail(verifiedBlockPath)) {
+                            if (writeBlockOrFail(block, blockNumber, verifiedBlockPath)) {
+                                sendBlockNotification(blockNumber, true, effectiveSource);
+                            } else {
+                                sendBlockNotification(blockNumber, false, effectiveSource);
+                            }
```
jsync-swirlds: Very minor: this could be simplified

```suggestion
                            final boolean success = writeBlockOrFail(block, blockNumber, verifiedBlockPath);
                            sendBlockNotification(blockNumber, success, effectiveSource);
```

jsync-swirlds (dismissed):
Looks good, two small potential improvements, which are non-blocking.

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
ivannov: As discussed today, the historic plugin sends success notification only after a zip is created successfully. There is this task: https://github.com/hiero-ledger/hiero-block-node/issues/2276 which changes that. I plan to start working on it as soon as I can

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
ivannov: Simplified. Will do the same for the historic plugin once I go on implementing https://github.com/hiero-ledger/hiero-block-node/issues/2276

jsync-swirlds (approved):

AlfredoG87 (approved):
LGTM

PR #3259, opened 2026-07-22
fix: Remove Duplicate Entries of `recentlyVerifiedBlocks` on Failed Persistence

### LINKED ISSUE #3258: Remove Duplicate Entries of `recentlyVerifiedBlocks` on Failed Persistence
### Description

- recentlyVerifiedBlocks was updated from a set to queue
- duplicates are now allowed
- logic to remove a block number has to now support duplicates

ata-nas:
* Removing all duplicates from the recentlyVerifiedBlock queue
* Added test

Fixes #3258


ata-nas:
After a discussion with @jsync-swirlds we decided that we want to preserve the current behavior. Passing successfully multiple times for a single block in rapid succession, means that we will have multiple entries for that block in the recently verified queue. If a persistence failure occurs, we want to remove the first occurrence only, because we know the block was propagated downstream at least once more. In that case, we still want an informational subsequent verification failure for the same block, as long as when the verification failure happened, the recently verified blocks queue has an entry for that block.

PR #3263, opened 2026-07-22
test: WRB distribution E2E test slice 4, steps 8-9

berryware:
## Reviewer Notes
[issue #3125](https://github.com/hiero-ledger/hiero-block-node/issues/3125) (WRB distribution E2E test, epic #2954), continuing the cumulative Solo-based test suite from steps 1-7 (already merged).

- **[reconfigure-cn-to-push-bn3.sh](tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/reconfigure-cn-to-push-bn3.sh)** — writes a fresh `block-nodes.json` onto all 3 CN pods (`network-node1-0`/`node2-0`/`node3-0`) pointing at BN3, matching the existing deploy-time config format (config-file-only, no restart needed).
- **[assert-cn-publishing-to-bn3.sh](tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/assert-cn-publishing-to-bn3.sh)** — verifies each CN's config references BN3. Only checks the config surface, not actual block flow, since BN3's EMB=100M (set in step 5) means it intentionally won't persist low block numbers yet.

- **New Java subcommand [`PushWrappedBlocksCommand`](tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/PushWrappedBlocksCommand.java)** (`blocks push`) — reads already-wrapped blocks from a local directory and streams any BN1 doesn't have yet via the existing `LiveBlockPushClient` (from PR #3124). Reuses `BlockStreamBlockHashRegistry` (watermark) and `BlockReader` — no changes needed to existing commands. Registered in `BlocksCommand.java`, with a new unit test `PushWrappedBlocksCommandTest.java` (3 tests, all passing).
- **[start-live-push.sh](tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/start-live-push.sh) / [stop-live-push.sh](tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/stop-live-push.sh)** — background loop invoking `blocks push` against BN1 every 30s (first run = historical backfill, later runs = live continuation), mirroring the existing `start-live-wrap.sh` pattern.
- **[assert-live-push-produced-new-blocks.sh](tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/assert-live-push-produced-new-blocks.sh)** — confirms BN1 has a `lastAvailableBlock` (historical landed) and the push loop kept iterating (live continues).

- **[wrb-distribution-step89.yaml](tools-and-tests/scripts/solo-e2e-test/tests/wrb-distribution-step89.yaml)** + **[topology](tools-and-tests/scripts/solo-e2e-test/topologies/wrb-distribution-step89.yaml)** — cumulative copy of step67 plus the new step 8/9 events (21 events total).
- Registered `wrb-distribution-step89` in the [CI workflow](.github/workflows/solo-e2e-test.yml) dropdowns.

- `./gradlew :tools:build` — full module build, 1524 tests passing.
- Manual smoke test of the built CLI (`blocks push --help`, empty-dir fast path).
- `task test:validate` on the new YAML — schema/topology validation passes.

### DIFF
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/bulk-load-historical-to-bn1.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/bulk-load-historical-to-bn1.sh
@@ -1,0 +1,121 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#
+# WRB Distribution E2E (#3125 slice 5 — step 9, historical half) —
+# bulk-load-historical-to-bn1.
+#
+# Historical WRBs must NOT go through BN1's live publish/gRPC stream: that
+# pipeline is a single-slot-per-block-number race (see
+# LiveStreamPublisherManager.resolveActionForHeader's nextUnstreamedBlockNumber
+# CAS) meant for freshly-produced blocks, and it silently SKIPs anything that
+# loses the race rather than persisting it — which is exactly what happened
+# when this slice first tried pushing block 0 to BN1 over gRPC. The wrb-cli's
+# own `blocks bulk-load` subcommand (PR #3038) exists for precisely this case:
+# it copies wrapped block zips directly into a Block Node's historic-storage
+# directory, bypassing live-stream verification entirely; BlockFileHistoricPlugin
+# picks them up on the BN's next startup scan. This script:
+#
+#   1. Runs `blocks bulk-load` locally (source=wrappedBlocks, dest=a local
+#      staging dir) — reuses the CLI's own file selection (zips only) and
+#      resumability bookkeeping instead of hand-rolling a copy.
+#   2. Streams the staged result into BN1's pod via `tar | kubectl exec` (more
+#      reliable than `kubectl cp`'s directory-copy semantics) at
+#      /opt/hiero/block-node/data/historic — the chart's actual configured
+#      archive.mountPath (charts/block-node-server/values.yaml), NOT
+#      /opt/hiero/block-node/data/archive (a stale path used by the test
+#      runner's unrelated clear-block-storage helper).
+#   3. Rolls BN1's StatefulSet pod so BlockFileHistoricPlugin re-scans on
+#      startup and picks up the newly-copied files. Safe here because nothing
+#      else is writing to BN1's storage at this point in the test (CN was
+#      redirected to BN3-only in step 8; MN1/MN2 only ever read from BN2/BN3).
+#      /opt/hiero/block-node/data is PVC-backed (volumeClaimTemplates in the
+#      chart), so the copied files survive the pod restart.
+#   4. Re-establishes the grpc/metrics port-forwards the restart tears down
+#      (same convention as add-bn.sh).
+#
+# Reads:
+#   NAMESPACE          (default "solo-network")
+#   CONTEXT            (default "kind-solo-cluster")
+#   BN1_GRPC_PORT      (default 40840 — matches add-bn.sh's port-forward convention)
+#   BN1_METRICS_PORT   (default 16007)
+#   READY_TIMEOUT      (default 300)
+
+set -euo pipefail
+
+ENV_FILE="${ENV_FILE:-/tmp/wrb-distribution-step12.env}"
+if [[ -f "${ENV_FILE}" ]]; then
+    # shellcheck disable=SC1090
+    source "${ENV_FILE}"
+fi
+
+: "${NAMESPACE:=solo-network}"
+: "${CONTEXT:=kind-solo-cluster}"
+BN1_GRPC_PORT="${BN1_GRPC_PORT:-$((40839 + 1))}"
+BN1_METRICS_PORT="${BN1_METRICS_PORT:-$((16006 + 1))}"
+READY_TIMEOUT="${READY_TIMEOUT:-300}"
+
+: "${WRB_DIST_WORK_DIR:?WRB_DIST_WORK_DIR must be set (written by install-and-run-wrb-cli.sh)}"
+: "${CLI_LIB:?CLI_LIB must be set (written by install-and-run-wrb-cli.sh)}"
+
+log() { echo "[wrb-dist-bulk-load-bn1] $*"; }
+fail() { echo "[wrb-dist-bulk-load-bn1] ERROR: $*" >&2; exit 1; }
+
+wrapped_dir="${WRB_DIST_WORK_DIR}/wrappedBlocks"
+staging_dir="${WRB_DIST_WORK_DIR}/bn1-bulk-load-staging"
+[[ -d "${wrapped_dir}" ]] || fail "Wrapped dir missing: ${wrapped_dir}"
+
+HISTORIC_MOUNT_PATH="/opt/hiero/block-node/data/historic"
+pod="block-node-1-0"
+
+log "Staging wrapped blocks via 'blocks bulk-load' (${wrapped_dir} -> ${staging_dir})..."
+java -cp "${CLI_LIB}/*" \
+    org.hiero.block.tools.BlockStreamTool blocks bulk-load \
+        --source "${wrapped_dir}" \
+        --dest "${staging_dir}" \
+    || fail "'blocks bulk-load' staging failed"
+
+if [[ -z "$(find "${staging_dir}" -name '*.zip' -print -quit 2>/dev/null)" ]]; then
+    fail "No .zip files staged in ${staging_dir}; nothing to bulk-load onto BN1"
+fi
+
+log "Streaming staged blocks into ${pod}:${HISTORIC_MOUNT_PATH}..."
+kubectl --context "${CONTEXT}" --namespace "${NAMESPACE}" \
+    exec "${pod}" -c block-node-server -- mkdir -p "${HISTORIC_MOUNT_PATH}" \
+    || fail "Failed to ensure ${HISTORIC_MOUNT_PATH} exists on ${pod}"
+tar -C "${staging_dir}" -cf - . | kubectl --context "${CONTEXT}" --namespace "${NAMESPACE}" \
+    exec -i "${pod}" -c block-node-server -- tar xf - -C "${HISTORIC_MOUNT_PATH}" \
+    || fail "Failed to stream staged blocks into ${pod}"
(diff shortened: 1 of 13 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/PushWrappedBlocksCommand.java
```diff
@@ -0,0 +1,135 @@
...
+        try (final LiveBlockPushClient pushClient = new LiveBlockPushClient(
+                bnHost, bnPort, pushQueueCapacity, LiveBlockPushClient.loadDefaultWebConfig())) {
+            pushClient.start();
+            final long bnLastAvailable = pushClient.queryLastAvailableBlock();
+            if (bnLastAvailable < 0) {
+                // -1 is also returned when the query itself failed (BN unreachable, RPC error), not just
+                // when the BN is genuinely empty; queryLastAvailableBlock() can't tell these apart. If
+                // it's really the latter, the pushes below will fail to connect and the ack check at the
+                // end of this method will catch it - this is just to make that ambiguity diagnosable.
+                System.out.println(
```
Nana-EC, line 81: Q: Should there be an exit here? If no BNs are present why continue?

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/PushWrappedBlocksCommand.java
```diff
@@ -0,0 +1,135 @@
...
+            localHighest = registry.highestBlockNumberStored();
+        }
+        if (localHighest < 0) {
+            System.out.println("[push] No wrapped blocks found in " + inputDir + "; nothing to push.");
+            return 0;
+        }
+
+        try (final LiveBlockPushClient pushClient = new LiveBlockPushClient(
+                bnHost, bnPort, pushQueueCapacity, LiveBlockPushClient.loadDefaultWebConfig())) {
+            pushClient.start();
```
Nana-EC: Q: Is this in the right position? Feels like it should be after the 2 ifs below to ensure we don't start the client and open a session if it's a no-op.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/PushWrappedBlocksCommand.java
```diff
@@ -0,0 +1,135 @@
...
+                            + "; stopping this run early (" + e.getMessage() + ")");
+                    break;
+                }
+                pushClient.pushBlock(blockNumber, block);
+                highestPushed = blockNumber;
+                pushedCount++;
+            }
+
+            if (pushedCount == 0) {
+                pushClient.shutdown();
```
Nana-EC: I don't think the shutdowns are needed. They'll get called when close() is called on the resource.

Nana-EC (dismissed):
Some suggestion. Non-blocking and can be address in follow up to move forward

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/start-live-push.sh
```diff
@@ -0,0 +1,125 @@
...
+
+if [[ -f "${PID_FILE}" ]] && kill -0 "$(cat "${PID_FILE}")" 2>/dev/null; then
+    log "Live push already running (pid $(cat "${PID_FILE}")); leaving it in place."
+    exit 0
+fi
+
+wrapped_dir="${WRB_DIST_WORK_DIR}/wrappedBlocks"
+[[ -d "${wrapped_dir}" ]] || { echo "Missing prerequisite: ${wrapped_dir}" >&2; exit 1; }
+
+printf 'initial_push_ok_count=0\n' > "${STATE_FILE}"
```
rockysingh: Hardcoding `initial_push_ok_count=0` works because line 121 truncates `${LOG_FILE}`, but couples the two. The short-circuit on lines 71-74 (if worker already running, exit 0) skips both this write and the truncate, so a re-run in a debug session leaves stale initial state and the assert compares against 0 instead of the actual current count.

Idempotent alternative:
```suggestion
initial_ok=$( grep -cE '\] push OK' "${LOG_FILE}" 2>/dev/null || echo 0 )
printf 'initial_push_ok_count=%s\n' "${initial_ok}" > "${STATE_FILE}"
```
Works regardless of whether the worker or log file is fresh.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/bulk-load-historical-to-bn1.sh
```diff
@@ -0,0 +1,121 @@
...
+
+pf_log_dir="${TMPDIR:-/tmp}/wrb-dist-add-bn-pf"
+mkdir -p "${pf_log_dir}"
+nohup setsid kubectl --context "${CONTEXT}" --namespace "${NAMESPACE}" \
+    port-forward svc/block-node-1 "${BN1_GRPC_PORT}:40840" \
+    >"${pf_log_dir}/block-node-1-grpc.log" 2>&1 </dev/null &
+nohup setsid kubectl --context "${CONTEXT}" --namespace "${NAMESPACE}" \
+    port-forward svc/block-node-1 "${BN1_METRICS_PORT}:16007" \
+    >"${pf_log_dir}/block-node-1-metrics.log" 2>&1 </dev/null &
+sleep 2
```
rockysingh: Fixed 2s sleep after re-establishing the port-forwards is fine on local kind but a plausible CI flake under runner load — if kubectl takes longer than 2s to bind the local port, the next test step racing against `localhost:${BN1_GRPC_PORT}` would see a connection refused.

Safer to poll until the port is actually accepting connections, e.g.
```bash
for _ in $(seq 1 30); do
    nc -z localhost "${BN1_GRPC_PORT}" && break
    sleep 1
done
```
Or equivalent with `bash</dev/tcp/localhost/${BN1_GRPC_PORT}` if `nc` isn't guaranteed on the runner.

tools-and-tests/tools/src/test/java/org/hiero/block/tools/blocks/PushWrappedBlocksCommandTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Test
+    @DisplayName("a wrapped block establishes the local watermark used to detect what's pushable")
+    void writingBlockEstablishesWatermark(@TempDir final Path tempDir) throws Exception {
+        writeWrappedBlock(tempDir, 0L);
+        try (final BlockStreamBlockHashRegistry registry =
+                new BlockStreamBlockHashRegistry(tempDir.resolve("blockStreamBlockHashes.bin"))) {
+            assertEquals(0L, registry.highestBlockNumberStored());
+        }
+    }
```
rockysingh, line 91: Small gap worth closing: no test exercises `runCommand()` with a wrapped block present + BN unreachable. The whole live-push failure-counter path in `start-live-push.sh` (rc != 0 → `consecutive_failures++`) depends on the command returning non-zero in that scenario, and today nothing asserts that.

One ~10-line addition:
```java
@Test
@DisplayName("present block + unreachable BN returns non-zero")
@Timeout(30)
void presentBlockUnreachableBnReturnsNonZero(@TempDir final Path tempDir) throws Exception {
    writeWrappedBlock(tempDir, 0L);
    assertNotEquals(0, runCommand(tempDir));
}
```

Optional drive-by while you're in this area: add a one-line comment above the second `pushClient.shutdown()` in `PushWrappedBlocksCommand.java:122` noting it's deliberate (drains ACKs before the `lastAcked() < highestPushed` check on line 127) — currently reads as redundant with try-with-resources.

rockysingh (commented):
Solid PR overall

tools-and-tests/tools/src/test/java/org/hiero/block/tools/blocks/PushWrappedBlocksCommandTest.java
berryware: Added the test. It is a 2min test.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/bulk-load-historical-to-bn1.sh
berryware: Changed to poll for the port-forwards.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/start-live-push.sh
berryware: start-live-push.sh:79 now snapshots the log's actual current push OK count instead of hardcoding 0;

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/PushWrappedBlocksCommand.java
berryware: the pushedCount == 0 branch now relies on try-with-resources' automatic close(). Left the shutdown() at line 123 untouched, since that one's still load-bearing for the ack-completeness check right after it.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/PushWrappedBlocksCommand.java
berryware: start() repositioned (PushWrappedBlocksCommand.java:94) — now called right before the push loop, after both no-op checks, so the worker thread is only spun up when there's actually something to push

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/PushWrappedBlocksCommand.java
berryware: It may be empty or it maybe unreachable at this point it is ambiguous. Defers the decision to the ack-completeness check.

rockysingh (dismissed):

Nana-EC (approved):

rockysingh (approved):
lgtm

PR #3269, opened 2026-07-23
feat: Add a next expected block field to Server Status API

### LINKED ISSUE #3260: Publishers need a field in Server Status for "next expected block".
### Story Form

As a Publisher
I want to know what block a block node expects to receive next
So that I do not send blocks that will result in `
EndStream(DUPLICATE)` or `NodeBehindPublisher` responses

### Technical Notes

The following are suggestions.
- Add a new field to the base serverStatus API.
   - _Possible_ name `next_expected_block`.
- Update publisher and serverStatus to ensure the next expected block is reported to server status on change.
   - Publisher plugin will need to provide this update frequently. Application State Facility is likely the preferred component to hold the current value.

jsync-swirlds:
## Reviewer Notes
- Added the new field to the API definition in protocol buffer files
- Added support for the new field in the Application State interface
- Added support for the new field in the App implementation
- Updated test fixtures with new methods
- Added a method to combine update of the next expected block with the "next unstreamed" CAS check in the Publisher plugin.
   - Replaced all existing CAS updates with the new method.
- Added unit tests for new feature
- Updated unit tests and test fixtures as needed

## Related Issue(s)
 Resolves #3260.

### DIFF
--- a/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
+++ b/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
@@ -13,6 +13,7 @@ import org.hiero.block.api.ServerStatusDetailResponse;
 import org.hiero.block.api.ServerStatusRequest;
 import org.hiero.block.api.ServerStatusResponse;
 import org.hiero.block.node.app.config.node.NodeConfig;
+import org.hiero.block.node.spi.ApplicationStateFacility;
 import org.hiero.block.node.spi.BlockNodeContext;
 import org.hiero.block.node.spi.BlockNodePlugin;
 import org.hiero.block.node.spi.ServiceBuilder;
@@ -57,33 +58,27 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
     public ServerStatusResponse serverStatus(@NonNull final ServerStatusRequest request) {
         requestStatusCounter.increment();
 
+        final ApplicationStateFacility stateFacility = blockNodeContext.applicationStateFacility();
         final ServerStatusResponse.Builder serverStatusResponseBuilder = ServerStatusResponse.newBuilder();
         final long firstAvailableBlock = blockProvider.availableBlocks().min();
         long highestAvailableBlock = blockProvider.availableBlocks().max();
-        // If this node is configured to manage blocks starting later than what it currently holds, report that
-        // later block as the last available block so publishers know they can stream from earliestManagedBlock
-        // instead of assuming the node only wants older blocks between the oldest stored block and earliest managed
-        // block.
-        if (highestAvailableBlock != UNKNOWN_BLOCK_NUMBER
-                && highestAvailableBlock < earliestManagedBlock
-                && earliestManagedBlock > 0) {
-            // Fix for bug #3203. CN will stream next block if we return UNKNOWN_BLOCK_NUMBER,
-            //   but if we return a future block here, CN will not send any data.
-            highestAvailableBlock = UNKNOWN_BLOCK_NUMBER;
+        long nextExpectedBlock = stateFacility.nextExpectedBlock();
+        if (nextExpectedBlock < earliestManagedBlock) {
+            nextExpectedBlock = UNKNOWN_BLOCK_NUMBER;
         }
 
-        // TODO(#579) Should get from state config or status, which would be provided by the context from responsible
-        // facility
-        boolean onlyLatestState = false;
+        // TODO(#579) Should get from state config or status, which would be provided by
+        //     calls to the responsible plugin.
+        boolean onlyLatestState = true;
 
-        // TODO(#1139) Should get construct a block node version object from application config, which would be provided
-        // by
-        // the context from responsible facility
+        // TODO(#1139) Should construct a block node version object from application config,
+        //     which would be provided from the application state
 
         // Build the response
         ServerStatusResponse response = serverStatusResponseBuilder
                 .firstAvailableBlock(firstAvailableBlock)
                 .lastAvailableBlock(highestAvailableBlock)
+                .nextExpectedBlock(nextExpectedBlock)
                 .onlyLatestState(onlyLatestState)
                 .build();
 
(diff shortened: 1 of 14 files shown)


block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -57,33 +58,27 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
...
+        if (nextExpectedBlock < earliestManagedBlock) {
+            nextExpectedBlock = UNKNOWN_BLOCK_NUMBER;
         }
 
-        // TODO(#579) Should get from state config or status, which would be provided by the context from responsible
-        // facility
-        boolean onlyLatestState = false;
+        // TODO(#579) Should get from state config or status, which would be provided by
+        //     calls to the responsible plugin.
+        boolean onlyLatestState = true;
```
Nana-EC, line 72: I think this is false because we as of yet don't support state. True might imply we support state.
We probably should have left this field out altogether until we were ready.

Reminder for me to dust of PR 2903 for the beta state management plugin

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ApplicationStateFacility.java
```diff
@@ -90,6 +86,16 @@ default NodeAddressBook getAddressBookForBlock(long blockNum) {
      */
     NetworkData backfillSources();
 
+    /**
+     * The next block expected from publishers. This is approximate, but should
+     * be less than the latest available block, in most situations, but may
```
Nana-EC: Q: Why less than, shouldn't it be equal to or greater than latest available block?
For example if we last acknowledged 100, shouldn't `nextExpectedBlock=101` and `lastAvailableBlock=100`?

It may even be higher if we have connected publishers streaming 101 and 102, in that case `lastAvailableBlock=103` would be correct, no?

protobuf-sources/src/main/proto/block-node/api/node_service.proto
```diff
@@ -104,6 +104,16 @@ message ServerStatusResponse {
      * in any `StateSnapshotRequest` sent to this Block-Node.
      */
     bool only_latest_state = 3;
+
+    /**
+     * The next block that this Block-Node expects to receive from a Publisher.
+     *
+     * This value MAY contain `-1` to indicate this Block-Node will accept
```
Nana-EC: -1 would be incorrect for proto no?
Should be `18446744073709551615`
```suggestion
     * This value MAY contain `18446744073709551615` (uint64_max) to indicate this Block-Node will accept
```

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -384,6 +389,71 @@ void testGetActionERROR(final BlockAction action) {
...
+                        new LiveStreamPublisherManager(embContext, generateManagerMetrics());
+                final TestResponsePipeline<PublishStreamResponse> embPipeline = new TestResponsePipeline<>();
+                embManager.addHandler(embPipeline, sharedHandlerMetrics, null);
+                final long embHandlerId = 0L; // first handler registered on the fresh manager
+
+                // Block 50 is below earliestManagedBlock (100); right after (re)start it is accepted
+                // via the pre-EMB path, and the published next expected block must be -1.
+                assertThat(embManager.getActionForBlock(50L, null, embHandlerId))
+                        .isEqualTo(BlockAction.ACCEPT);
+                assertThat(testApplicationState.nextExpectedBlock()).isEqualTo(-1L);
```
Nana-EC, line 454: This might give a false confidence since you can't really tell teh difference between an initialzied valid and the case of stored blocks being less than EMB.

Might want to set up the test in a way that it start with non -1 (i.e. stream a valid block), before calling `embManager.getActionForBlock(...)`

Nana-EC (commented):
Some suggestions and clarification requests

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
jsync-swirlds: False says that we can provide current and prior state values (i.e. snapshots); I changed to true which says we don't provide older snapshots, only the current value.  The narrower option seemed more appropriate for "we don't have state services".

block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/ApplicationStateFacility.java
jsync-swirlds: Last available is the last persisted block; it has already completed verification and been persisted.

You're correct, next expected should be _greater_ than last available.

Will fix.

protobuf-sources/src/main/proto/block-node/api/node_service.proto
jsync-swirlds: Bad habits.

Will fix.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
jsync-swirlds: We do (in effect) stream a valid block and verify that it was accepted right before this line.
calling `getActionForBlock` sets the next unstreamed value.
I added an extra assertion to show that nextUnstreamed was incremented and the application state value did not change.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
Nana-EC: Change to true is fine, no-op either-ways with no state APIs yet

Nana-EC (approved):
LG

AlfredoG87 (approved):

PR #3274, opened 2026-07-23
feat: Files Historic Plugin Sends Persisted Notifications on Staged Blocks and not on Zips

### LINKED ISSUE #2276: `Files Historic Plugin` Sends Persisted Notifications on Staged Blocks and not on Zips

ivannov:
### New `stagedBlockNotificationsEnabled` config flag on `FilesHistoricConfig`

Adds a `stagedBlockNotificationsEnabled` boolean (default `false`) that controls whether `BlockFileHistoricPlugin` fires a `PersistedNotification` for every block as soon as it is staged to disk, instead of the legacy behavior of one notification per completed zip batch. Default is disabled so existing deployments see no behavior change.

### `BlockFileHistoricPlugin` wiring

In the staging path, a successful `writeBlockOrFail` now sends a per-block success notification when the flag is on (previously nothing was sent on the success path here, only on failure). In the zip-completion path, the batch-level notification is now skipped entirely when per-staged-block notifications are enabled, since each block in the batch was already notified individually as it was staged.

## Related Issue(s)

Closes #2276

### DIFF
--- a/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
+++ b/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
@@ -372,6 +372,8 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
             if (createDirectoryOrFail(verifiedBlockPath)) {
                 if (writeBlockOrFail(block, blockNumber, verifiedBlockPath)) {
                     if (config.stagedBlockNotificationsEnabled()) {
+                        // block is durably staged, but not yet accessible via block(long) until its
+                        // batch is zipped; see FilesHistoricConfig.stagedBlockNotificationsEnabled
                         sendBlockNotification(blockNumber, true, source);
                     }
                 } else {
--- a/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
+++ b/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
@@ -370,7 +370,11 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
             final Path verifiedBlockPath = BlockFile.nestedDirectoriesBlockFilePath(
                     stagingPath, blockNumber, config.compression(), config.maxFilesPerDir());
             if (createDirectoryOrFail(verifiedBlockPath)) {
-                if (!writeBlockOrFail(block, blockNumber, verifiedBlockPath)) {
+                if (writeBlockOrFail(block, blockNumber, verifiedBlockPath)) {
+                    if (config.stagedBlockNotificationsEnabled()) {
+                        sendBlockNotification(blockNumber, true, source);
+                    }
+                } else {
                     sendBlockNotification(blockNumber, false, source);
                 }
             } else {
@@ -633,9 +637,11 @@ public final class BlockFileHistoricPlugin implements BlockProviderPlugin, Block
                     plugin.totalZipFiles.incrementAndGet();
                     final String successMessage = "Successfully moved batch of blocks[{0} -> {1}] to zip file.";
                     plugin.LOGGER.log(DEBUG, successMessage, batchFirstBlockNumber, batchLastBlockNumber);
-                    // now all the blocks are in the zip file and accessible, send notification
-                    // @todo is this needed? Does anything actually care when a zip file is completed?
-                    plugin.sendBlockNotification(batchLastBlockNumber, true, BlockSource.HISTORY);
+                    // now all the blocks are in the zip file and accessible, send notification, unless
+                    // notifications were already sent per block as each one was staged
+                    if (!plugin.config.stagedBlockNotificationsEnabled()) {
+                        plugin.sendBlockNotification(batchLastBlockNumber, true, BlockSource.HISTORY);
+                    }
                     plugin.cleanup();
                 }
             } catch (final IOException e) {
--- a/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
+++ b/block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
@@ -24,6 +24,10 @@ import org.hiero.block.node.base.Loggable;
  * {@link #powersOfTenPerZipFileContents} is set to 3, then this means that 5 zips will be retained and these zips
  * contain 10^3 blocks, i.e. 5_000 blocks effectively retained. If set to 0 (zero), blocks will be retained
  * indefinitely.
+ * @param stagedBlockNotificationsEnabled whether a Persisted Notification is sent for every block as soon as it
+ * is staged, rather than once per zip batch (using the last block number of the batch). Defaults to disabled, so
+ * that by default only a single Persisted Notification is sent per successfully archived zip batch, matching the
+ * plugin's legacy behavior.
  */
 @ConfigData("files.historic")
 public record FilesHistoricConfig(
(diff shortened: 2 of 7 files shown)


block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -633,9 +637,11 @@ public void run() {
                     plugin.totalZipFiles.incrementAndGet();
                     final String successMessage = "Successfully moved batch of blocks[{0} -> {1}] to zip file.";
                     plugin.LOGGER.log(DEBUG, successMessage, batchFirstBlockNumber, batchLastBlockNumber);
-                    // now all the blocks are in the zip file and accessible, send notification
-                    // @todo is this needed? Does anything actually care when a zip file is completed?
-                    plugin.sendBlockNotification(batchLastBlockNumber, true, BlockSource.HISTORY);
+                    // now all the blocks are in the zip file and accessible, send notification, unless
+                    // notifications were already sent per block as each one was staged
+                    if (!plugin.config.stagedBlockNotificationsEnabled()) {
+                        plugin.sendBlockNotification(batchLastBlockNumber, true, BlockSource.HISTORY);
```
Nana-EC, line 645: Q: If we were only sending success notification on the batch archive how come there was no upstream impact? How were we acknowledging if no persistence success was being sent for individual blocks? Was the logic based on this line, so if we notify success for block x then we assume all blocks below x are persisted?

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
```diff
@@ -32,6 +36,7 @@ public record FilesHistoricConfig(
         @Loggable @ConfigProperty(defaultValue = "ZSTD") CompressionType compression,
         @Loggable @ConfigProperty(defaultValue = "4") @Min(1) @Max(6) int powersOfTenPerZipFileContents,
         @Loggable @ConfigProperty(defaultValue = "0") @Min(0) long blockRetentionThreshold,
-        @Loggable @ConfigProperty(defaultValue = "3") @Min(1) int maxFilesPerDir) {
+        @Loggable @ConfigProperty(defaultValue = "3") @Min(1) int maxFilesPerDir,
+        @Loggable @ConfigProperty(defaultValue = "false") boolean stagedBlockNotificationsEnabled) {
```
Nana-EC, line 41: This requires that operators take a manual step to capitalize on this.
Either that or we need to update the default in our charts.
What's the reason for leaving is false by default?

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
```diff
@@ -370,7 +370,11 @@ private void writeBlockToStagingPath(final BlockUnparsed block, final long block
             final Path verifiedBlockPath = BlockFile.nestedDirectoriesBlockFilePath(
                     stagingPath, blockNumber, config.compression(), config.maxFilesPerDir());
             if (createDirectoryOrFail(verifiedBlockPath)) {
-                if (!writeBlockOrFail(block, blockNumber, verifiedBlockPath)) {
+                if (writeBlockOrFail(block, blockNumber, verifiedBlockPath)) {
+                    if (config.stagedBlockNotificationsEnabled()) {
+                        sendBlockNotification(blockNumber, true, source);
```
Nana-EC, line 377: One impact here is that we'll send a notification of persistence but the block won't yet be available to callers until archive.
This is where the difference between available and stored is important.
This may be unclear to all callers so it's worth documenting either in the Javadocs for the new config and the notification itself

block-node/blocks-file-historic/src/test/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPluginTest.java
```diff
@@ -887,25 +903,64 @@ void testZipRangeHappyPathNotificationSent() throws IOException {
...
+            for (int i = 0; i < 10; i++) {
+                assertThat(sentPersistedNotifications.get(i))
+                        .returns(true, PersistedNotification::succeeded)
+                        .returns((long) i, PersistedNotification::blockNumber)
+                        .returns(toTest.defaultPriority(), PersistedNotification::blockProviderPriority);
+            }
+            // execute serially to ensure all tasks are completed, no additional notification is sent for the
+            // completed zip batch, as each block was already notified individually when staged
+            pluginExecutor.executeSerially();
+            assertThat(blockMessaging.getSentPersistedNotifications()).hasSize(10);
```
Nana-EC, line 963: Should complete teh test with a call to confirm accessibility.
That is before the zipping `assertThat(toTest.block(i)).isNull();` and after zipping  `assertThat(toTest.block(i)).isNotNull();`

Nana-EC (commented):
Nice add, some questions and suggestions

block-node/blocks-file-historic/src/test/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPluginTest.java
ivannov: Added block(i) accessibility assertions (null before zip, non-null after)

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
ivannov: I changed it back and forth several times during developing this. At the end I decided to keep it disabled by default to preserve today's behavior for existing deployments (single notification per archived batch) rather than silently changing behavior - I wanted to avoid surprises.

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
ivannov: I added:

* More detailed javadoc
* A caveat section in the documentation
* An inline comment in the code

As for the available blocks returned by the ASF (at the end they are returned by this class), they are only updated upon successful zipping. The notifications don't affect them

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
ivannov: Yes, this is what we also do in the `CloudStorageArchivePlugin`. When we send notifications, we don't send them for all the blocks in the range, but only for the last one. The assumption is that the zip batches here are always a contiguous range and the block number is the high water mark for that range.

jsync-swirlds: The protocol for publishers defines the block in each acknowledgement as "fully inclusive", meaning if we acknowledge block 45987, that means every block prior to 45987 is _also_ acknowledged.
By doing this we avoid possibly sending hundreds or thousands of successful persistence notifications in a batch, but can send single notifications to cover large groups of blocks and can transform those into single Acknowledgements if appropriate.

As a practical matter, the time needed to gather and zip blocks means we'll almost always see this post-zip notification and update internal bookkeeping only because later blocks were already acknowledged to all publishers.

Note also, Notification _does not_ impact blocks stored or available, those are direct plugin queries rather than having ASF track blocks stored and available via notifications.

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/FilesHistoricConfig.java
jsync-swirlds: Is this for success, failure, or both?

If this only affects _success_ notification, then i think `false` is what was discussed in slack. The reasoning being that current deployments all include files-recent, which handles the per-block successful persistence notifications, but we always want to publish persistence failures.

ivannov: This config is only for success: whether each block sends success notification or whether it is sent upon zipping. There is (even before this PR) an immediate notification upon failure

jsync-swirlds (approved):

ata-nas (approved):
LG, only thing I want to note and bring to attention is that we use `BlockSource.HISTORY` as source for these notifications. Plugins, that handle this, should be aware. This source appears to be used in successful notifications only.

PR #3156, opened 2026-07-06
ci(sync): use GitHub API commits to satisfy required_signatures ruleset

deshmukhpranali:
## Problem

Every bot-created sync PR to hiero-docs is blocked with **"Commits must have verified signatures"** because the hiero-docs `master` branch enforces `required_signatures` on all commits. The bot can't GPG-sign commits, requiring a manual GPG-sign amend before every merge — this is not scalable.

## Solution

Replace the manual `git push` + `gh pr create` approach with the `peter-evans/create-pull-request` action, which commits via the **GitHub API**. GitHub automatically signs and verifies API-created commits, satisfying the `required_signatures` ruleset without:
- A bot GPG key
- Any secrets
- Any manual intervention

## What changes

The "Copy docs" step is unchanged. The "Commit and open PR" step is replaced with a single `peter-evans/create-pull-request` step that:
- Creates a GitHub-API commit (automatically **Verified**)
- Opens or updates the `sync/block-node-docs` PR against `hiero-docs` master
- Is idempotent — updates the existing PR if already open

## Result

Future sync PRs will have verified commits and can be squash-merged directly without any manual signing step.

Closes #2737 (continuous task - removes the manual signing blocker from the sync pipeline)

PR #3277, opened 2026-07-23
fix(ci): run git config inside hiero-docs working directory

deshmukhpranali:
## Reviewer Notes

The Import GPG Key action's git_commit_gpgsign, git_tag_gpgsign, and git_user_signingkey inputs internally run `git config` commands. With two repos checked out as subdirectories (block-node/ and hiero-docs/), the runner workspace root is not a git repo, causing `fatal: not in a git directory` on every sync run since #3156 merged.

Fix: remove the git_* inputs from the action (key import only) and configure git manually in a `run:` step scoped to `working-directory: hiero-docs`. GPG signing is preserved via `git commit -S`.

### DIFF
--- a/.github/workflows/sync-docs-to-gitbook.yaml
+++ b/.github/workflows/sync-docs-to-gitbook.yaml
@@ -48,24 +48,32 @@ jobs:
         id: gpg_importer
         uses: step-security/ghaction-import-gpg@c0b4a334b9b72bfaaebec86ef022a78a5b17f828 # v7.0.0
         with:
-          git_commit_gpgsign: true
-          git_tag_gpgsign: true
-          git_user_signingkey: true
           gpg_private_key: ${{ secrets.GPG_KEY_CONTENTS }}
           passphrase: ${{ secrets.GPG_KEY_PASSPHRASE }}
 
-      - name: Open PR with GPG-signed commit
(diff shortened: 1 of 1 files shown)


jsync-swirlds (approved):

.github/workflows/sync-docs-to-gitbook.yaml
```diff
@@ -48,24 +48,32 @@ jobs:
...
+          git add block-node/
+          if git diff --staged --quiet; then
+            echo "No block-node doc changes to sync."
+            exit 0
+          fi
+          BRANCH="sync/block-node-docs"
+          git checkout -b "$BRANCH"
+          git commit -S --signoff -m "chore: sync block-node docs from hiero-block-node@${GITHUB_SHA::7}"
+          git push --force --set-upstream origin "$BRANCH"
+          if gh pr view "$BRANCH" --repo hiero-ledger/hiero-docs >/dev/null 2>&1; then
```
Nana-EC, line 72: FYI this assumes 1 open PR per branch i.e. no stale PRs that were discontinued.
If a branch ever has a closed PR it'll get picked up and the logic will fall into the echo and not the pr create.
Likely not an issue but might be worth an improvement follow up that uses a `gh pr list` and filters for open explicitly on that branch

Nana-EC (approved):
Works in most cases, noted a non-blocking improvement to avoid the edge case in future

PR #3287, opened 2026-07-24
fix(ci): use sha512 checksum for git-cliff install.

Nana-EC:
## Summary

- git-cliff releases publish `.sha512` per-asset checksum files, not `.sha256`. The `Install git-cliff` step in `release-automation.yaml` tried to download and verify `${TARBALL}.sha256`, which does not exist in any git-cliff release, causing an immediate exit-1 after the tarball download succeeded.
- Fixes the failed run: https://github.com/hiero-ledger/hiero-block-node/actions/runs/30125095647/job/89586623980

## Changes

- Download `${TARBALL}.sha512` instead of `${TARBALL}.sha256`
- Verify with `sha512sum --check` instead of `sha256sum --check`
- Clean up `${TARBALL}.sha512` in the `rm` step
- Update the inline comment hint (`grep sha512`)

## Related Issue(s)

Fixes regression introduced in #3275.

### DIFF
--- a/.github/workflows/release-automation.yaml
+++ b/.github/workflows/release-automation.yaml
@@ -40,7 +40,7 @@ jobs:
       RELEASE_NOTES_FILENAME: release_notes
       # Bump this to pick up a new git-cliff release; verify the new SHA via:
       #   gh release view v<VERSION> --repo orhun/git-cliff --json assets \
-      #     | jq '.assets[].name' | grep sha256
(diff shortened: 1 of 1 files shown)


berryware (dismissed):
LG

berryware (approved):
Looks even better

AlfredoG87 (approved):

PR #3288, opened 2026-07-24
fix(ci): use sha512 checksum and make git-cliff steps non-blocking

Nana-EC:
## Summary

- git-cliff releases publish `.sha512` per-asset checksum files, not `.sha256`. The `Install git-cliff` step tried to download `${TARBALL}.sha256` which doesn't exist in any git-cliff release, causing an immediate exit-1 after a successful tarball download.
- Both cliff steps are now `continue-on-error: true` so a future version mismatch or network hiccup cannot block a release. A guard step ensures the `bodyFile` reference always resolves to a valid (possibly empty) file.
- Fixes the failed run: https://github.com/hiero-ledger/hiero-block-node/actions/runs/30125095647/job/89586623980

## Changes

- `sha256` → `sha512` in the download pattern, checksum command, `rm` cleanup, and inline comment hint
- `continue-on-error: true` on `Install git-cliff` (with `id: install_cliff`) and `Generate Release Notes`
- `Generate Release Notes` skipped (`if: steps.install_cliff.outcome == 'success'`) when cliff isn't installed
- New `Ensure Release Notes File Exists` step (`touch`) always runs so `Create Github Release` always has a valid `bodyFile`

## Related Issue(s)

Fixes regression introduced in #3275. Companion fix for `release/0.39` in #3287.

### DIFF
--- a/.github/workflows/release-automation.yaml
+++ b/.github/workflows/release-automation.yaml
@@ -40,7 +40,7 @@ jobs:
       RELEASE_NOTES_FILENAME: release_notes
       # Bump this to pick up a new git-cliff release; verify the new SHA via:
       #   gh release view v<VERSION> --repo orhun/git-cliff --json assets \
-      #     | jq '.assets[].name' | grep sha256
+      #     | jq '.assets[].name' | grep sha512
       GIT_CLIFF_VERSION: "2.8.0"
     outputs:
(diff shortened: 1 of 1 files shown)


berryware (approved):
LG

Nana-EC:
Closing as wrapping into #3295

PR #3289, opened 2026-07-25
feat: Update network data with live roster data for RSA address book and proximate block for EndOfStream

### LINKED ISSUE #2650: Add a field to EndStream that can indicate which block is most proximate to the end stream event
### Story Form

As a Publisher
I want to know which block is most proximate to an EndStream event
So that I can better understand which block is most relevant to the EndStream.

### Technical Notes

Publishers have reported significant difficulty with understanding the cause for EndStream events because the only block reported is the last known verified and persisted block. For some events (particularly bad block proof) the publisher can greatly improve debugging and remediation if we can provide a more proximate block number (either the block that caused the end stream or the last received block).

This new field should be added to the EndStream API in a backward compatible manner.

jsync-swirlds:
## Reviewer Notes
* Add RSA address book endpoints to publishers
   * Adjust ApplicationStateFacility `updateAddressBookHistory` to also add endpoints for _Current_ address book to publishers.
   * This is a _partial_ implementation due to limits in current TSS roster data and absent state management.
* Add new field to EndOfStream message for the proximate block
   * Added a new field to the message definition for the block proximate to a causative event.
* Removed unused `default` implementations in application state interface.
* Extracted many ApplicationState methods from BlockNodeApp to a new ApplicationStateUtility class
* Added methods for processing address book data into network connections
* Added methods to merge address book update data into the known publishers set of network connections
* Added tests for new methods and missing tests for existing methods.
* Added a Comparator class for ranged address book, and test for that class

## Related Issue(s)
 Resolves #2650
 Partially addresses 3082


block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -43,6 +48,8 @@ public record ApplicationStateConfig(
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/known-publishers.json") Path knownPublishersFilePath,
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/inbound-partners.json") Path inboundPartnersFilePath,
         @Loggable @ConfigProperty(defaultValue = "/opt/hiero/block-node/application-state/outbound-partners.json") Path outboundPartnersFilePath,
+        @Loggable @ConfigProperty(defaultValue = "publisher") String publisherInboundCategory,
```
ata-nas, line 51: Are these categories limited? We can consider an enum here as enums are easier to work with.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -0,0 +1,328 @@
...
+    /// "Newer" is defined as:
+    /// <ul>
+    ///   <li>The current history is null or empty (always accept the first history).</li>
+    ///   <li>The incoming history's last era starts at a higher block number.</li>
+    ///   <li>Equal last-era start blocks but more total eras (defensive; should not occur in
+    ///       normal operation).</li>
+    /// </ul>
+    static boolean isNewerHistory(RangedAddressBookHistory incoming, RangedAddressBookHistory current) {
+        if (current == null || current.addressBooks().isEmpty()) return true;
+        if (incoming.addressBooks().isEmpty()) return false;
```
ata-nas: Null check for `incoming` as well?

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -0,0 +1,328 @@
...
+    /// Loads a [NetworkData] document from the given JSON file.
+    /// Missing files and parse failures are logged and yield
+    /// [NetworkData#DEFAULT] (an empty set) rather than throwing, mirroring
+    /// the lenient handling used for other optional application-state files.
+    ///
+    /// @param path the JSON file to read
+    /// @return the parsed NetworkData, or [NetworkData#DEFAULT] if
+    ///     absent or unreadable
+    static NetworkData loadNetworkData(final Path path) {
+        if (path == null || !Files.exists(path)) {
```
ata-nas, line 116: Instead of `exists`, we can do `isRegularFile`. This check will fail if the path does not exist, but also will fail if it is inaccessible or otherwise, not a regular file. If we proceed, I would also recommend chaning the below log to `Network data file not present or inaccessible...`

This is optional, however, as we do handle parse and IO exceptions below, which would also catch a file we cannot access.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -0,0 +1,328 @@
...
+            return NetworkData.DEFAULT;
+        }
+    }
+
+    /// Validates that the NodeAddressBook has at least one entry with a non-blank RSA\_PubKey.
+    ///
+    /// @param book the address book to validate
+    /// @param source human-readable source name for error messages
+    /// @throws IllegalStateException if the book is empty or has no usable entries
+    static void validateAddressBook(final NodeAddressBook book, final String source) {
```
ata-nas, line 135: Should we add a null check to `book`?

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -0,0 +1,328 @@
...
+                    "RSA address book from %s contains no entries — cannot verify WRB proofs".formatted(source));
+        }
+        final long declared = book.nodeAddress().stream()
+                .filter(a -> !a.rsaPubKey().isBlank())
+                .count();
+        final String noValidKeyMessage =
+                "RSA address book from %s has %s entries but none have a valid RSA_PubKey".formatted(source, declared);
+        if (declared == 0) {
+            throw new IllegalStateException(noValidKeyMessage);
+        }
```
ata-nas, line 147: Instead of throwing, should we consider to return a boolean result for this method?

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -0,0 +1,328 @@
...
+            }
+        }
+        return unique;
+    }
+
+    private static int compareNetworkConnection(NetworkConnection left, NetworkConnection right) {
+        int result = 0;
+        if (left == null || right == null) {
+            result = left == null ? right == null ? 0 : -1 : 1;
+        }
```
ata-nas: Should we also add an object identity check as well? Returning 0 if they are the same object? I do not think we need to do expensive comparison in that scenario.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -0,0 +1,328 @@
...
+            result = compareComparable(left.certificate(), right.certificate());
+        }
+        return result;
+    }
+
+    private static <T extends Comparable<T>> int compareComparable(final T left, final T right) {
+        int result = 0;
+        if (left == null || right == null) {
+            result = left == null ? right == null ? 0 : -1 : 1;
+        }
```
ata-nas: Should we also add an object identity check here?

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -0,0 +1,328 @@
...
+
+    private static int compareConnectionReference(final ConnectionReference left, final ConnectionReference right) {
+        int result = 0;
+        final String leftAddress = left.address();
+        final String rightAddress = right.address();
+        final String leftPort = left.port();
+        final String rightPort = right.port();
+        if (leftAddress == null || rightAddress == null) {
+            result = leftAddress == null ? rightAddress == null ? 0 : -1 : 1;
+        }
```
ata-nas: Should we also add an object identity check here?

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -0,0 +1,328 @@
...
+    /// address is the dotted-decimal form of `ipAddressV4` (see [#toIpV4String]) or, if the node
+    /// publishes a `domainName` instead, that domain name. An endpoint that sets both
+    /// `ipAddressV4` and `domainName`, sets neither, or carries a malformed `ipAddressV4` is
+    /// invalid and skipped.
+    ///
+    /// @param book the address book to derive connections from; must not be null
+    /// @return a new, mutable list of derived connections (possibly empty)
+    static List<NetworkConnection> publisherConnectionsFrom(final NodeAddressBook book) {
+        final List<NetworkConnection> connections = new LinkedList<>();
+        for (final NodeAddress node : book.nodeAddress()) {
```
ata-nas, line 292: Should we add a null check for `book` here? Perhaps returning an empty list if `book` is null?

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -0,0 +1,328 @@
...
+
+    /// Translates a 4-byte, big-endian IPv4 address into dotted-decimal string form.
+    /// For example the bytes `7F 00 00 01` become `127.0.0.1`.
+    ///
+    /// @param ipAddressV4 the raw 4-byte, big-endian address
+    /// @return the dotted-decimal string, or null if the input is not exactly four bytes
+    static String toIpV4String(final Bytes ipAddressV4) {
+        if (ipAddressV4.length() != 4) {
+            return null;
+        }
```
ata-nas, line 332: Should we also add a null check here?

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
ata-nas, line 1: I see quite a few null checks and identity checks opportunities here, but also some opportunities to return a result instead of throwing for validations. I think these are good additions that will improve stability and also allow.

+1 for this class, not only does it help with readability and reduced complexity within the block node app class, but also will be easier when app state is extracted as a standalone plugin later.

block-node/app/src/main/java/org/hiero/block/node/app/RangedAddressBookComparator.java
```diff
@@ -0,0 +1,35 @@
...
+///
+/// All block-number comparisons are performed as _unsigned_ comparisons. The open-ended sentinel
+/// `endBlock == -1` is therefore the largest possible unsigned value and naturally sorts to the
+/// end, so an open-ended era wins its `startBlock` tie against any bounded era.
+public final class RangedAddressBookComparator implements Comparator<RangedNodeAddressBook> {
+
+    @Override
+    public int compare(final RangedNodeAddressBook left, final RangedNodeAddressBook right) {
+        // 1. A null address book sinks to the front so it is never selected as the newest era.
+        final int byUsable = Boolean.compare(left.addressBook() != null, right.addressBook() != null);
```
ata-nas: Should we add null and object identity checks for `left` and `right`

block-node/facility-messaging/src/test/java/org/hiero/block/server/messaging/DynamicBlockItemTest.java
```diff
@@ -511,27 +512,14 @@ public void onTooFarBehindError() {
      * still see all events without any being silently dropped.
      */
     @Test
-    void testNoBackpressureHandlerReceivesAllItemsWhenFastEnough() throws Exception {
+    void testNoBackpressureHandlerReceivesAllItemsWhenFastEnough() throws InterruptedException {
```
ata-nas, line 515: We can consider to add `@Timeout` here or globally for this test suite.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -221,7 +221,7 @@ public void onNext(@NonNull final PublishStreamRequestUnparsed request) {
                 // If we reach here, it means that the handler was interrupted or
                 // an unexpected error occurred. We should log the error and shut down.
                 LOGGER.log(INFO, "[%2$s] Error processing request: %1$s".formatted(e, correlationIdPrefix), e);
-                endStreamWithCode(Code.ERROR, true);
+                endStreamWithCode(Code.ERROR, true, 0L);
```
ata-nas, line 224: Is `0L` the correct value here?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -354,7 +354,7 @@ public void shutdown() {
         for (final Long nextKey : handlers.keySet()) {
             final PublisherHandler value = handlers.remove(nextKey);
             if (value != null) {
-                value.endStreamWithCode(Code.SUCCESS, false);
+                value.endStreamWithCode(Code.SUCCESS, false, 0L);
```
ata-nas, line 357: Is `0L` the correct value here?

protobuf-sources/src/main/proto/block-node/api/block_stream_publish_service.proto
```diff
@@ -350,6 +350,25 @@ message PublishStreamResponse {
...
+         *   <li>For a bad block proof, this will be the Block Node's best
+         *     effort to report the number of the block that failed.</li>
+         *   <li>For a success, this will be the last block number observed by
+         *     the current handler.</li>
+         *   <li>For a timeout this will be the Block Node's best effort
+         *     to report the number of the block that timed out.</li>
+         * </ul>
+         * This value is a _best effort_ attempt and is not guaranteed to be
+         * correct, or even to be present. If the value cannot be filled in
+         * correctly, the value `0` will be provided.
```
ata-nas, line 368: If the value cannot be filled or provided for any reason, shall we consider `-1L`? This is a flag that clearly indicates "I could not provide this value". `0L` is a valid block number value and could be misleading.

ata-nas (commented):
Nice additions here. Some cleanup suggestions and questions.

ata-nas:
https://github.com/hiero-ledger/hiero-block-node/actions/runs/30146391453/job/89649031312?pr=3289

Note this test seems to be flaky.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: This is pre-existing code that was merely moved from one place to another.
I'd prefer not to make any changes to that code in this PR.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: Also pre-existing code.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: Also pre-existing code.

It would be helpful if you could gather all of these suggestions in a separate issue.  That way we separate a simple move from functional improvements.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: A lot of the null checks are unnecessary, as the inputs cannot be null.
I'll take a look at some of them, however.
Identity checks are dicey, they should never succeed, but might be worthwhile.

The changes with exceptions are in _pre-existing_ code that we do not want to modify here; we're just _moving_ the code with no change in this PR.

Those should definitely be captured in a separate issue, however.

block-node/app/src/main/java/org/hiero/block/node/app/RangedAddressBookComparator.java
jsync-swirlds: _Perhaps_; it's not entirely clear it these _can_ be null, but worth checking.

protobuf-sources/src/main/proto/block-node/api/block_stream_publish_service.proto
jsync-swirlds: `0L` wont ever be "valid" for this particular field as we would have to be working on a later block in order for it to be needed.
Said another way, if block `0` had an issue, it's the first block ever sent, no block has been acknowledged or received any other response, and there is no chance of being confused as to which block is most proximate to the issue.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Yes, `0L` is correct.  It's unambiguous (yes, block 0 is valid, but if that had an issue it would not need the "proximate" block and "last persisted" is sufficient) and _much_ cheaper to send than `uint64_max`.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
jsync-swirlds: Yes, it is correct.

block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
jsync-swirlds: Categories are not limited and the actual values are driven by Solo Provisioner, not the Block Node.
That said, there is a place in the Utility that should be using this value, so I'll fix that.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: incoming is never null, it's managed at the caller.
Wouldn't hurt, however, so will add.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: Identity is extremely unlikely; but will add.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: I don't believe identical objects are possible here, but will add.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: I don't believe identical objects are possible here, but will add.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: Book isn't null here, the caller ensures that.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
jsync-swirlds: the IP address isn't null here, the caller ensures that.

block-node/app/src/main/java/org/hiero/block/node/app/RangedAddressBookComparator.java
jsync-swirlds: The inputs _should_ never be null, but I added the check just in case.

jsync-swirlds:
> https://github.com/hiero-ledger/hiero-block-node/actions/runs/30146391453/job/89649031312?pr=3289
> 
> Note this test seems to be flaky.

Yes, the API tests continue to be unreliable and difficult to maintain in a working state.
They're failing on all PRs pretty consistently.

ata-nas (approved):
@jsync-swirlds this needs a rebase since a recent test fix now unblocks CI

jsync-swirlds:
> @jsync-swirlds this needs a rebase since a recent test fix now unblocks CI

CI wasn't blocked, the only blocking item was an unresolved conversation and missing a second review.

AlfredoG87 (approved):
LGTM

PR #3290, opened 2026-07-27
feat: Implemented step 10 of [#3125](https://github.com/hiero-ledger/hiero-block-node/issues/3125)

berryware:
*"Use the solo-provisioner to reconfigure BN2 and BN3 to set BN1 as a backfill source"* — following the exact slicing pattern established by the prior steps (step12/step345/step67), as slice 5:

- [reconfigure-bn-backfill.sh](tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/reconfigure-bn-backfill.sh) — reconfigures an already-deployed BN (added post-hoc via `add-bn.sh`) to backfill from another BN. Since Solo has no `block node update` command, it does by hand what the Helm chart would render: creates the `<bn>-block-node-sources` ConfigMap, merges `BACKFILL_BLOCK_NODE_SOURCES_PATH`/`BACKFILL_GREEDY` into the existing `<bn>-config` ConfigMap, then patches the StatefulSet's pod template to add the volume + volumeMount the chart only renders when `blockNode.backfill.sources` is set at install time.
- [assert-bn-backfill-source.sh](tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/assert-bn-backfill-source.sh) — verifies both the env var and the mounted sources file content on each target BN.
- [topologies/wrb-distribution-step10.yaml](tools-and-tests/scripts/solo-e2e-test/topologies/wrb-distribution-step10.yaml) and [tests/wrb-distribution-step10.yaml](tools-and-tests/scripts/solo-e2e-test/tests/wrb-distribution-step10.yaml) — same deploy-time shape as step67, extended with steps 1-7 plus the new step 10 reconfigure/assert events.
- Wired `wrb-distribution-step10` into the [solo-e2e-test.yml](.github/workflows/solo-e2e-test.yml) topology dropdown and test-definition description.

The test definition validated successfully via `solo-test-runner.sh --validate`. I couldn't run it end-to-end against a live cluster (would require standing up a Solo/Kind network), so the actual `kubectl patch` behavior against a real StatefulSet is unverified — worth a dry run before merging.

## Related Issue(s)

### DIFF
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -308,6 +308,14 @@ jobs:
           # Export RELAY_VERSION so Solo CLI reads the resolved semver (not 'latest')
           RELAY_RESOLVED=$(echo "$OUTPUT" | grep "^relay_version=" | cut -d= -f2)
           echo "RELAY_VERSION=${RELAY_RESOLVED}" >> "${GITHUB_ENV}"
+          # Export MN_VERSION so the wrb-distribution scripts' JVM importer swap
+          # (reconfigure-mn1-to-bn-sources.sh / add-mn2.sh) always matches
+          # whatever version MN1 was actually installed with — Flyway rejects
+          # rolling the importer back to an older migration set than what's
+          # already applied to MN1's Postgres, so this must never drift from
+          # the resolved mn_version below.
+          MN_RESOLVED=$(echo "$OUTPUT" | grep "^mn_version=" | cut -d= -f2)
+          echo "MN_VERSION=${MN_RESOLVED}" >> "${GITHUB_ENV}"
 
       # Install Solo CLI (npm by default, or build an approved fork/branch when solo-source=git)
       - name: Install Solo CLI
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -82,7 +82,7 @@ on:
         type: string
         default: "5"
       test-definition:
-        description: "Test definitions (comma-separated): smoke-test, basic-load, node-restart-resilience, full-history-backfill, tss-signature-transition, wrb-cli-hash-validation, wrb-cli-wrapping-validation, wrb-differential-test, wrb-core-validation, wrb-distribution-step12, wrb-distribution-step345, chaos-foundation-smoke, latency-cn-to-cn, latency-cn-to-bn, latency-bn-to-bn, latency-all-three"
+        description: "Test definitions (comma-separated): smoke-test, basic-load, node-restart-resilience, full-history-backfill, tss-signature-transition, wrb-cli-hash-validation, wrb-cli-wrapping-validation, wrb-differential-test, wrb-core-validation, rsa-roster-verification,  wrb-distribution-step12, wrb-distribution-step345, wrb-distribution-step67, wrb-distribution-step10, chaos-foundation-smoke, latency-cn-to-cn, latency-cn-to-bn, latency-bn-to-bn, latency-all-three"
         type: string
         default: "none"
     outputs:
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -36,7 +36,7 @@ on:
       solo-version:
         description: "Solo CLI Version"
         type: string
-        default: "0.83.0"
+        default: "0.84.0"
       solo-source:
         description: "Solo install source: 'npm' or 'git' (build an approved fork/branch)"
         type: string
@@ -144,7 +144,7 @@ on:
       solo-version:
         description: "Solo CLI Version"
         required: false
-        default: "0.83.0"
+        default: "0.84.0"
       solo-source:
         description: "Solo install source"
         required: false
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -116,6 +116,7 @@ on:
           - wrb-distribution-step12 # WRB Distribution E2E (#3125) — slice 1: 3 CN v0.75 + MN1 v0.157.1, no BN
           - wrb-distribution-step345 # WRB Distribution E2E (#3125) — slice 3: adds BN1/BN2/BN3 with EMB
           - wrb-distribution-step67 # WRB Distribution E2E (#3125) — slice 4: MN1 → BN2/BN3 + MN2 from genesis
+          - wrb-distribution-step10 # WRB Distribution E2E (#3125) — slice 5: BN2/BN3 reconfigured to backfill from BN1
       block-node-version:
         description: "Block Node version ('latest', 'rc', 'main', or tag like 'v0.29.0')"
         required: false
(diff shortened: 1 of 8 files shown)


tools-and-tests/scripts/solo-e2e-test/tests/wrb-distribution-step10.yaml
```diff
@@ -0,0 +1,191 @@
...
+
+  - id: reconfigure-bn3-backfill
+    type: command
+    description: "Step 10: reconfigure BN3 to set BN1 as a backfill source"
+    delay: 430
+    timeout: 300
+    args:
+      script: "${SCRIPT_DIR}/wrb-distribution/reconfigure-bn-backfill.sh 3"
+
+  - id: assert-bn-backfill-source
```
Nana-EC, line 151: You might hit a timing issue here.
I believe the reconfigure above will result in a rolling restart. In that case 10 seconds between tasks will be too short.

You could either extend i to be 90 seconds (just more certainty) or the more reliable option would be updating `assert-bn-backfill-source.sh` to check if the pods are ready.
Either a rollout status of the stateful set or a readiness check of the pods

.github/workflows/solo-e2e-test.yml
Nana-EC: nit: Should this have `wrb-distribution-step67` and `wrb-distribution-step10`?

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/reconfigure-bn-backfill.sh
```diff
@@ -0,0 +1,136 @@
...
+
+target_index="${1:?reconfigure-bn-backfill.sh: target BN index required (2|3)}"
+
+log() { echo "[wrb-dist-bn-backfill] $*"; }
+fail() { echo "[wrb-dist-bn-backfill] ERROR: $*" >&2; exit 1; }
+
+target_bn="block-node-${target_index}"
+source_bn="block-node-${SOURCE_BN_INDEX}"
+source_dns="${source_bn}.${NAMESPACE}.svc.cluster.local"
+source_port=40840
```
Nana-EC: Hard coded port. This would break in custom-port tests. Maybe add it as an input so that you pull from values.yaml to get the right value

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/assert-bn-backfill-source.sh
```diff
@@ -0,0 +1,73 @@
...
+        log "${target_bn}: BACKFILL_BLOCK_NODE_SOURCES_PATH='${actual_path:-<unset>}', expected '${expected_path}'"
+        failures=$(( failures + 1 ))
+        continue
+    fi
+
+    sources_content=$(kubectl --context "${CLUSTER_REFERENCE}" --namespace "${NAMESPACE}" \
+        exec "${pod}" -c block-node-server -- \
+        cat "${expected_path}" 2>/dev/null || echo "")
+
+    if [[ "${sources_content}" != *"${source_dns}"* ]]; then
```
Nana-EC: DNS verification should include port also

Nana-EC (commented):
Some suggestions

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/reconfigure-bn-backfill.sh
```diff
@@ -0,0 +1,136 @@
...
+cat > "${statefulset_patch}" <<EOF
+spec:
+  template:
+    spec:
+      volumes:
+        - name: block-node-sources
+          configMap:
+            name: ${sources_configmap}
+      containers:
+      - name: block-node-server
```
rockysingh, line 137: The strategic-merge behaviour that makes this a safe append-not-clobber depends on `spec.template.spec.volumes[]` and `spec.template.spec.containers[].volumeMounts[]` both being keyed on `name`. Which they are that's the deliberate design. But it also silently assumes the container is literally named `block-node-server`: if the chart ever renames it, the patch just no-ops on the volumeMount (the volume itself still lands via the `volumes[]` list) and there's no error to catch it.

Cheap ways to make the assumption explicit:

- One-line comment above the patch calling out the `name: block-node-server` dependency on the chart, OR
- Precondition check that the container exists before patching:

```bash
actual_container=$(kubectl --context "${CLUSTER_REFERENCE}" --namespace "${NAMESPACE}" \
    get statefulset "${target_bn}" \
    -o jsonpath='{.spec.template.spec.containers[0].name}')
[[ "${actual_container}" == "block-node-server" ]] \
    || fail "expected container named 'block-node-server' in ${target_bn}, got '${actual_container}'"
```

Catches a future chart rename immediately instead of at the next backfill run failure. This is a non blocker but worth consideirng to make it robust.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/assert-bn-backfill-source.sh
rockysingh: Substring match on `source_dns` non-blocking today (slice has max 3 BNs) but if the test ever scales past 9 BNs, `block-node-1.` prefix-matches `block-node-10.` and this silently accepts a wrong source. Future-proof with `jq -e`:

```suggestion
    if ! echo "${sources_content}" | jq -e --arg dns "${source_dns}" '.nodes[] | select(.address == $dns)' >/dev/null 2>&1; then
```

Same semantics for today's 3-BN case; safe against 10+ BNs and against source_dns appearing inside some other field's string.

rockysingh (commented):
Two non-blocking items worth folding in PR or follow up.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/assert-bn-backfill-source.sh
berryware: replaced the substring DNS check with a jq -e exact match on address + port, resolving both the prefix-collision risk and the missing port check in one change.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/reconfigure-bn-backfill.sh
berryware: added a precondition check that fails loudly if the StatefulSet's container isn't named block-node-server before patching, instead of silently no-op'ing the volumeMount.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/assert-bn-backfill-source.sh
berryware: replaced the substring DNS check with a jq -e exact match on address + port, resolving both the prefix-collision risk and the missing port check in one change.

tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/reconfigure-bn-backfill.sh
berryware: Changed to read from the BN's ConfigMap.

.github/workflows/solo-e2e-test.yml
berryware: sync'd this up. Step 10 was done separately from the prior.

tools-and-tests/scripts/solo-e2e-test/tests/wrb-distribution-step10.yaml
berryware: Added a comment about the timing

Nana-EC (approved):

rockysingh (approved):

PR #3299, opened 2026-07-28
feat: Proto Ordinal Streaming Hasher Mapping in BlockHasher

### LINKED ISSUE #3195: Proto Ordinal Streaming Hasher Mapping in `BlockHasher`

ata-nas:
* Add forward compatible item handling to BlockHasher: category is field number modulo 20 for fields 20 and above
  * Hash categories 3 to 7 into the existing subtrees and categories 8 to 15 into the eight extension subtrees
  * Ignore not hashed categories 0 and 19
  * Refuse categories 1, 2, 16 to 18 and unknown fields below 20 with the new UNSUPPORTED_ITEM_TYPE failure type
* Rework item processing into an exhaustive switch expression, so new item types fail compilation until handled
* Add UNSUPPORTED_STREAM_FORMAT failure type for encodings that parse but are not a processable block stream
  * Reject oneof violations with it: a known item type plus unknown fields, or multiple unknown fields
  * Reject mandatory once per block items (header, footer, record file) appearing more than once with it
* Refuse redacted and filtered items with UNSUPPORTED_ITEM_TYPE instead of UNABLE_TO_PARSE, which now strictly means a wire level parse failure
* Extend HashingUtilities.computeFinalBlockHash with the extension subtree roots per HIP-1424
  * Exclude absent extension leaves with single child 0x01 promotion, keeping current block hashes unchanged
* Add metrics counting future items hashed, not hashed and refused
* Add tests
  * Independent HIP-1424 reference implementation verifying agreement for every category, extension and refusal case
  * Not hashed items leave the root unchanged, future items round trip byte identical
  * Failure type mappings and the new failure scenarios for UNSUPPORTED_STREAM_FORMAT

Resolves #3195

**NOTE** a visualization of the tree structure can be seen here:
https://raw.githubusercontent.com/hiero-ledger/hiero-improvement-proposals/d1bf49d33473b9dbe4f6c018e460f802f293aa3c/assets/hip-1424/expanded-merkle-mountain-top.svg

Also, hip 1424 can be used as reference and a newly created forward compatibility doc here #3251

### DIFF
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
@@ -46,6 +46,14 @@ import org.hiero.block.node.spi.blockmessaging.BlockSource;
 public final class BlockHasher implements Supplier<HashingResult> {
     /// The time to park between polls of the block items deque when no data is available.
     private static final long DATA_BUSY_WAIT_TIME_NANOS = TimeUnit.MICROSECONDS.toNanos(200);
+    /// The first `BlockItem` field number governed by the block stream forward compatibility
+    /// numbering rule. Field numbers below this value belong to the first release; an unknown
+    /// field below it is reserved for item types that require specific handling.
+    private static final int FIRST_FORWARD_COMPATIBLE_FIELD_NUMBER = 20;
+    /// The modulus of the forward compatibility numbering rule: for a field numbered
+    /// [#FIRST_FORWARD_COMPATIBLE_FIELD_NUMBER] or above, the hashing category is the field
+    /// number modulo this value.
+    private static final int CATEGORY_MODULUS = 20;
     /// The number of the block being hashed.
     private final long blockNumber;
     /// The source of the block, carried through to the result and failures.
@@ -70,6 +78,26 @@ public final class BlockHasher implements Supplier<HashingResult> {
     private final NaiveStreamingTreeHasher stateChangesHasher;
     /// The tree hasher for trace data item hashes.
     private final NaiveStreamingTreeHasher traceDataHasher;
+    /// Extension 0 subtree hasher, leaf position 9 of the fixed 16 leaf block root tree
+    /// ("Merkle Mountain Top" in HIP-1424). The eight extension leaf positions (9 to 16) are
+    /// permanently defined in the tree; each hasher below is bound to its position. A hasher is
+    /// created when the block contains an item of the corresponding extension category, and a
+    /// null hasher means the leaf carries no data in this block, so it is excluded from the tree.
+    private NaiveStreamingTreeHasher extensionHasherZero;
+    /// Extension 1 subtree hasher, leaf position 10. See [#extensionHasherZero].
+    private NaiveStreamingTreeHasher extensionHasherOne;
+    /// Extension 2 subtree hasher, leaf position 11. See [#extensionHasherZero].
+    private NaiveStreamingTreeHasher extensionHasherTwo;
+    /// Extension 3 subtree hasher, leaf position 12. See [#extensionHasherZero].
+    private NaiveStreamingTreeHasher extensionHasherThree;
+    /// Extension 4 subtree hasher, leaf position 13. See [#extensionHasherZero].
+    private NaiveStreamingTreeHasher extensionHasherFour;
+    /// Extension 5 subtree hasher, leaf position 14. See [#extensionHasherZero].
+    private NaiveStreamingTreeHasher extensionHasherFive;
+    /// Extension 6 subtree hasher, leaf position 15. See [#extensionHasherZero].
+    private NaiveStreamingTreeHasher extensionHasherSix;
+    /// Extension 7 subtree hasher, leaf position 16. See [#extensionHasherZero].
+    private NaiveStreamingTreeHasher extensionHasherSeven;
     /// Provider of the verification data, used to publish TSS data found in block 0.
     private final VerificationDataProvider verificationDataProvider;
     /// The parsed block header, set when the header item is seen.
@@ -138,86 +166,9 @@ public final class BlockHasher implements Supplier<HashingResult> {
                     } else {
                         this.accumulatedBlockItems.addAll(currentBlockItems);
                         for (final BlockItemUnparsed item : currentBlockItems) {
-                            final BlockItemUnparsed.ItemOneOfType kind =
-                                    item.item().kind();
-                            switch (kind) {
-                                case BLOCK_HEADER -> {
-                                    if (this.blockHeader == null) {
-                                        this.blockHeader = standardParse(BlockHeader.PROTOBUF, item.blockHeader());
-                                        this.hapiProtoVersion = this.blockHeader.hapiProtoVersion();
-                                        if (this.hapiProtoVersion == null) {
-                                            throw new VerificationSessionFailedException(
-                                                    blockNumber,
-                                                    SessionFailureType.MISSING_MANDATORY_FIELD,
-                                                    blockSource);
-                                        } else {
-                                            outputTreeHasher.addLeaf(getBlockItemHash(item));
-                                        }
-                                    } else {
-                                        throw new VerificationSessionFailedException(
-                                                blockNumber, SessionFailureType.UNABLE_TO_PARSE, blockSource);
-                                    }
-                                }
-                                case ROUND_HEADER, EVENT_HEADER ->
-                                    consensusHeaderHasher.addLeaf(getBlockItemHash(item));
-                                case SIGNED_TRANSACTION -> {
-                                    inputTreeHasher.addLeaf(getBlockItemHash(item));
-                                    if (blockItemsRecord.blockNumber() == 0 && !verificationDataProvider.hasTssData()) {
-                                        final LedgerIdPublicationTransactionBody publication =
-                                                findLedgerIdPublication(item.signedTransaction());
-                                        if (publication != null) {
-                                            // publish TSS Data
-                                            final TssData tssData =
-                                                    VerificationHelper.extractTssData(publication, blockNumber);
-                                            verificationDataProvider.safeUpdateTssData(tssData, true);
-                                        }
-                                    }
-                                }
-                                case TRANSACTION_RESULT, TRANSACTION_OUTPUT ->
-                                    outputTreeHasher.addLeaf(getBlockItemHash(item));
-                                case STATE_CHANGES -> stateChangesHasher.addLeaf(getBlockItemHash(item));
-                                case TRACE_DATA -> traceDataHasher.addLeaf(getBlockItemHash(item));
-                                case RECORD_FILE -> {
-                                    if (this.rawRecordFileItemProtoBytes == null) {
-                                        this.rawRecordFileItemProtoBytes = item.recordFileOrThrow();
-                                        outputTreeHasher.addLeaf(getBlockItemHash(item));
-                                    } else {
-                                        throw new VerificationSessionFailedException(
-                                                blockNumber, SessionFailureType.UNABLE_TO_PARSE, blockSource);
-                                    }
-                                }
-                                case BLOCK_FOOTER -> {
-                                    if (this.blockFooter == null) {
-                                        this.blockFooter = standardParse(BlockFooter.PROTOBUF, item.blockFooter());
-                                    } else {
-                                        throw new VerificationSessionFailedException(
-                                                blockNumber, SessionFailureType.UNABLE_TO_PARSE, blockSource);
-                                    }
-                                }
-                                case BLOCK_PROOF -> {
-                                    final BlockProof blockProof = standardParse(BlockProof.PROTOBUF, item.blockProof());
-                                    blockProofs.add(blockProof);
-                                }
-                                case REDACTED_ITEM, FILTERED_SINGLE_ITEM -> {
-                                    // not permitted currently, fail the hashing.
-                                    throw new VerificationSessionFailedException(
-                                            blockNumber, SessionFailureType.UNABLE_TO_PARSE, blockSource);
-                                }
-                                case UNSET -> {
-                                    throw new VerificationSessionFailedException(
-                                            blockNumber, SessionFailureType.UNKNOWN_ERROR, blockSource);
-                                }
-                                default -> {
-                                    // @todo(3195) add field-number based sorting here.
-                                    throw new VerificationSessionFailedException(
-                                            blockNumber, SessionFailureType.UNKNOWN_ERROR, blockSource);
-                                }
-                            }
-                            final List<UnknownField> itemUnknownFields = item.getUnknownFields();
-                            if (!itemUnknownFields.isEmpty()) {
-                                // @todo(3195) add field-number based sorting here.
-                                throw new VerificationSessionFailedException(
-                                        blockNumber, SessionFailureType.UNABLE_TO_PARSE, blockSource);
+                            final SessionFailureType failure = processItem(item, blockItemsRecord.blockNumber());
+                            if (failure != null) {
+                                throw new VerificationSessionFailedException(blockNumber, failure, blockSource);
                             }
                         }
                     }
@@ -244,6 +195,234 @@ public final class BlockHasher implements Supplier<HashingResult> {
         return isCanceled.get() || Thread.currentThread().isInterrupted();
     }
 
+    /// Processes a single block item: hashes it into the correct subtree and captures block
+    /// level data (header, footer, proofs) along the way.
+    ///
+    /// The switch below is deliberately an exhaustive switch expression with no default branch.
+    /// When a new item type is added to the `BlockItem` schema, compilation fails here until the
+    /// new type is given an explicit handling decision. An item type that is unknown to the
+    /// compiled schema altogether surfaces as `UNSET` with the data preserved as an unknown
+    /// field, and is handled by the forward compatibility numbering rule in
+    /// [#processFutureItem].
+    /// @param item the block item to process
+    /// @param itemsBlockNumber the block number carried by the current block items record
+    /// @return null on success, or the failure type when the block must be refused
+    /// @throws ParseException if a known item fails to parse
+    private SessionFailureType processItem(final BlockItemUnparsed item, final long itemsBlockNumber)
+            throws ParseException {
+        final BlockItemUnparsed.ItemOneOfType kind = item.item().kind();
+        final List<UnknownField> unknownFields = item.getUnknownFields();
+        final SessionFailureType failure;
+        if (kind != BlockItemUnparsed.ItemOneOfType.UNSET && !unknownFields.isEmpty()) {
+            // A BlockItem is a protobuf oneof, so a valid item carries exactly one field. A known
+            // item type alongside unknown fields parses fine, but is not a processable stream.
+            failure = SessionFailureType.UNSUPPORTED_STREAM_FORMAT;
+        } else {
+            failure = switch (kind) {
+                case BLOCK_HEADER -> {
+                    if (this.blockHeader != null) {
+                        // a mandatory once per block item appearing more than once is a valid
+                        // encoding, but not a processable stream
+                        yield SessionFailureType.UNSUPPORTED_STREAM_FORMAT;
+                    } else {
+                        this.blockHeader = standardParse(BlockHeader.PROTOBUF, item.blockHeader());
+                        this.hapiProtoVersion = this.blockHeader.hapiProtoVersion();
+                        if (this.hapiProtoVersion == null) {
+                            yield SessionFailureType.MISSING_MANDATORY_FIELD;
+                        } else {
+                            outputTreeHasher.addLeaf(getBlockItemHash(item));
+                            yield null;
+                        }
+                    }
+                }
+                case ROUND_HEADER, EVENT_HEADER -> {
+                    consensusHeaderHasher.addLeaf(getBlockItemHash(item));
+                    yield null;
+                }
+                case SIGNED_TRANSACTION -> {
+                    inputTreeHasher.addLeaf(getBlockItemHash(item));
+                    if (itemsBlockNumber == 0 && !verificationDataProvider.hasTssData()) {
+                        final LedgerIdPublicationTransactionBody publication =
+                                findLedgerIdPublication(item.signedTransaction());
+                        if (publication != null) {
+                            // publish TSS Data
+                            final TssData tssData = VerificationHelper.extractTssData(publication, blockNumber);
+                            verificationDataProvider.safeUpdateTssData(tssData, true);
+                        }
+                    }
+                    yield null;
+                }
+                case TRANSACTION_RESULT, TRANSACTION_OUTPUT -> {
+                    outputTreeHasher.addLeaf(getBlockItemHash(item));
+                    yield null;
+                }
+                case STATE_CHANGES -> {
+                    stateChangesHasher.addLeaf(getBlockItemHash(item));
+                    yield null;
+                }
+                case TRACE_DATA -> {
+                    traceDataHasher.addLeaf(getBlockItemHash(item));
+                    yield null;
+                }
+                case RECORD_FILE -> {
+                    if (this.rawRecordFileItemProtoBytes != null) {
+                        // a mandatory once per block item appearing more than once is a valid
+                        // encoding, but not a processable stream
+                        yield SessionFailureType.UNSUPPORTED_STREAM_FORMAT;
+                    } else {
+                        this.rawRecordFileItemProtoBytes = item.recordFileOrThrow();
+                        outputTreeHasher.addLeaf(getBlockItemHash(item));
+                        yield null;
+                    }
+                }
+                case BLOCK_FOOTER -> {
+                    if (this.blockFooter != null) {
+                        // a mandatory once per block item appearing more than once is a valid
+                        // encoding, but not a processable stream
+                        yield SessionFailureType.UNSUPPORTED_STREAM_FORMAT;
+                    } else {
+                        this.blockFooter = standardParse(BlockFooter.PROTOBUF, item.blockFooter());
+                        yield null;
+                    }
+                }
+                case BLOCK_PROOF -> {
+                    blockProofs.add(standardParse(BlockProof.PROTOBUF, item.blockProof()));
+                    yield null;
+                }
+                // item types this version of the node cannot process currently
+                case REDACTED_ITEM, FILTERED_SINGLE_ITEM -> SessionFailureType.UNSUPPORTED_ITEM_TYPE;
+                case UNSET -> processFutureItem(item, unknownFields);
+            };
+        }
+        return failure;
+    }
+
+    /// Handles an item whose type is unknown to the compiled schema, applying the block stream
+    /// forward compatibility numbering rule: for a field numbered 20 or above, the hashing
+    /// category is the field number modulo 20. Categories that map to a defined subtree are
+    /// hashed like any other item, not-hashed categories are read and ignored, and everything
+    /// else refuses the block, because guessing could produce a hash that disagrees with an
+    /// upgraded node.
+    /// @param item the block item carrying the unknown field
+    /// @param unknownFields the unknown fields of the item
+    /// @return null on success, or the failure type when the block must be refused
+    private SessionFailureType processFutureItem(final BlockItemUnparsed item, final List<UnknownField> unknownFields) {
+        final SessionFailureType failure;
+        if (unknownFields.isEmpty()) {
+            // an item with no field at all, nothing valid to process
+            failure = SessionFailureType.UNKNOWN_ERROR;
+        } else if (unknownFields.size() > 1) {
+            // a BlockItem is a oneof, so a valid item carries exactly one field; more than one
+            // unknown field parses fine, but is not a processable stream
+            failure = SessionFailureType.UNSUPPORTED_STREAM_FORMAT;
+        } else {
+            final int fieldNumber = unknownFields.getFirst().field();
+            if (fieldNumber < FIRST_FORWARD_COMPATIBLE_FIELD_NUMBER) {
+                // an unknown field below 20 is a first release field reserved for item types that
+                // require specific handling this version does not know
+                hashingMetrics.futureItemsRefused().increment();
+                failure = SessionFailureType.UNSUPPORTED_ITEM_TYPE;
+            } else {
+                final int category = fieldNumber % CATEGORY_MODULUS;
+                failure = switch (category) {
+                    case 0, 19 -> {
+                        // not part of the block proof merkle tree, read and ignore
+                        hashingMetrics.futureItemsNotHashed().increment();
+                        yield null;
+                    }
+                    case 1, 2 -> {
+                        // requires specific handling
+                        hashingMetrics.futureItemsRefused().increment();
+                        yield SessionFailureType.UNSUPPORTED_ITEM_TYPE;
+                    }
+                    case 3 -> hashFutureItem(item, consensusHeaderHasher);
+                    case 4 -> hashFutureItem(item, inputTreeHasher);
+                    case 5 -> hashFutureItem(item, outputTreeHasher);
+                    case 6 -> hashFutureItem(item, stateChangesHasher);
+                    case 7 -> hashFutureItem(item, traceDataHasher);
+                    case 8 -> {
+                        extensionHasherZero = createIfNull(extensionHasherZero);
+                        yield hashFutureItem(item, extensionHasherZero);
+                    }
+                    case 9 -> {
+                        extensionHasherOne = createIfNull(extensionHasherOne);
+                        yield hashFutureItem(item, extensionHasherOne);
+                    }
+                    case 10 -> {
+                        extensionHasherTwo = createIfNull(extensionHasherTwo);
+                        yield hashFutureItem(item, extensionHasherTwo);
+                    }
+                    case 11 -> {
+                        extensionHasherThree = createIfNull(extensionHasherThree);
+                        yield hashFutureItem(item, extensionHasherThree);
+                    }
+                    case 12 -> {
+                        extensionHasherFour = createIfNull(extensionHasherFour);
+                        yield hashFutureItem(item, extensionHasherFour);
+                    }
+                    case 13 -> {
+                        extensionHasherFive = createIfNull(extensionHasherFive);
+                        yield hashFutureItem(item, extensionHasherFive);
+                    }
+                    case 14 -> {
+                        extensionHasherSix = createIfNull(extensionHasherSix);
+                        yield hashFutureItem(item, extensionHasherSix);
+                    }
+                    case 15 -> {
+                        extensionHasherSeven = createIfNull(extensionHasherSeven);
+                        yield hashFutureItem(item, extensionHasherSeven);
+                    }
+                    // categories 16 to 18 are reserved with no subtree in the block root tree
+                    case 16, 17, 18 -> {
+                        // categories 16 to 18 are reserved with no subtree in the block root tree
+                        hashingMetrics.futureItemsRefused().increment();
+                        yield SessionFailureType.UNSUPPORTED_ITEM_TYPE;
+                    }
+                    default -> SessionFailureType.UNKNOWN_ERROR;
+                };
+            }
+        }
+        return failure;
+    }
+
+    /// Hashes a future item into the given subtree hasher.
+    /// @param item the block item to hash
+    /// @param hasher the subtree hasher the item's category maps to
+    /// @return always null, the item was hashed successfully
+    private SessionFailureType hashFutureItem(final BlockItemUnparsed item, final NaiveStreamingTreeHasher hasher) {
+        hasher.addLeaf(getBlockItemHash(item));
+        hashingMetrics.futureItemsHashed().increment();
+        return null;
+    }
+
+    /// Returns the given extension subtree hasher, or a new one when it does not exist yet.
+    /// @param hasher the current extension subtree hasher, or null when not yet created
+    /// @return the given hasher, or a new one when the given hasher is null
+    private static NaiveStreamingTreeHasher createIfNull(final NaiveStreamingTreeHasher hasher) {
+        final NaiveStreamingTreeHasher result;
+        if (hasher == null) {
+            result = new NaiveStreamingTreeHasher();
+        } else {
+            result = hasher;
+        }
+        return result;
+    }
+
+    /// Returns the root hash of the given extension subtree hasher, or null when the hasher was
+    /// never created, meaning the corresponding leaf of the block root tree carries no data in
+    /// this block.
+    /// @param hasher the extension subtree hasher, or null when never created
+    /// @return the root hash of the subtree, or null when the hasher is null
+    private static byte[] extensionSubtreeRoot(final NaiveStreamingTreeHasher hasher) {
+        final byte[] root;
+        if (hasher == null) {
+            root = null;
+        } else {
+            root = hasher.rootHash().join().toByteArray();
+        }
+        return root;
+    }
+
     /// Finish the hashing operation.
     /// This method will finalize the hashing process. Root hash will be calculated and
     /// a [HashingResult] will be returned.
@@ -268,7 +447,15 @@ public final class BlockHasher implements Supplier<HashingResult> {
                         outputTreeHasher,
                         consensusHeaderHasher,
                         stateChangesHasher,
-                        traceDataHasher);
+                        traceDataHasher,
+                        extensionSubtreeRoot(extensionHasherZero),
+                        extensionSubtreeRoot(extensionHasherOne),
+                        extensionSubtreeRoot(extensionHasherTwo),
+                        extensionSubtreeRoot(extensionHasherThree),
+                        extensionSubtreeRoot(extensionHasherFour),
+                        extensionSubtreeRoot(extensionHasherFive),
+                        extensionSubtreeRoot(extensionHasherSix),
+                        extensionSubtreeRoot(extensionHasherSeven));
                 final BlockUnparsed block = BlockUnparsed.newBuilder()
                         .blockItems(accumulatedBlockItems)
                         .build();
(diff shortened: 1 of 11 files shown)


block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -216,6 +156,177 @@ private boolean isCanceled() {
...
+                        hashingMetrics.futureItemsNotHashed().increment();
+                        yield null;
+                    }
+                    case 3 -> hashFutureItem(item, consensusHeaderHasher);
+                    case 4 -> hashFutureItem(item, inputTreeHasher);
+                    case 5 -> hashFutureItem(item, outputTreeHasher);
+                    case 6 -> hashFutureItem(item, stateChangesHasher);
+                    case 7 -> hashFutureItem(item, traceDataHasher);
+                    case 8, 9, 10, 11, 12, 13, 14, 15 ->
+                        hashFutureItem(item, extensionHasher(category - FIRST_EXTENSION_CATEGORY));
```
jsync-swirlds: Instead of treating these differently, we might prefer to have a field for each extension hasher because those that are actually used will be named for some purpose. It's a lot easier to change a field name than to change the method of accessing the hasher.

```suggestion
                    case 8 -> hashFutureItem(item, createIfNull(extensionHasherOne));
                    case 9 -> hashFutureItem(item, createIfNull(extensionHasherTwo));
                    case 10 -> hashFutureItem(item, createIfNull(extensionHasherThree));
                    case 11 -> hashFutureItem(item, createIfNull(extensionHasherFour));
                    case 12 -> hashFutureItem(item, createIfNull(extensionHasherFive));
                    case 13 -> hashFutureItem(item, createIfNull(extensionHasherSix));
                    case 14 -> hashFutureItem(item, createIfNull(extensionHasherSeven));
                    case 15 -> hashFutureItem(item, createIfNull(extensionHasherEight));
```

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -216,6 +156,177 @@ private boolean isCanceled() {
...
+    private NaiveStreamingTreeHasher extensionHasher(final int extensionIndex) {
+        final NaiveStreamingTreeHasher hasher;
+        if (extensionHashers[extensionIndex] == null) {
+            hasher = new NaiveStreamingTreeHasher();
+            extensionHashers[extensionIndex] = hasher;
+        } else {
+            hasher = extensionHashers[extensionIndex];
+        }
+        return hasher;
+    }
```
jsync-swirlds: This method seems overly tied to the current not-yet-assigned state rather than being built on the basis of a fixed and known 16 leaf merkle mountain top.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -283,7 +283,18 @@ public static ByteBuffer getBlockItemHash(@NonNull BlockItemUnparsed blockItemUn
     }
 
     /**
-     * Computes the final block hash from the given block footer, timestamp and tree hashers.
+     * The number of extension subtree leaf positions (9 to 16) in the block root fixed size tree
+     * ("Merkle Mountain Top" in HIP-1424). These positions are reserved for future block item
+     * categories; a position carries a leaf only when the block contains at least one item of the
+     * corresponding extension category.
+     */
+    public static final int EXTENSION_SUBTREE_COUNT = 8;
```
jsync-swirlds: There's no point having a count, the count cannot ever change without invalidating every block in history; something Hiero is extremely unlikely to do.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -305,6 +316,57 @@ public static Bytes computeFinalBlockHash(
...
+            @NonNull final Timestamp blockTimestamp,
+            @NonNull final Bytes previousBlockHash,
+            @NonNull final Bytes rootHashOfAllPreviousBlockHashes,
+            @NonNull final Bytes startOfBlockStateRootHash,
+            @NonNull final StreamingTreeHasher inputTreeHasher,
+            @NonNull final StreamingTreeHasher outputTreeHasher,
+            @NonNull final StreamingTreeHasher consensusHeaderHasher,
+            @NonNull final StreamingTreeHasher stateChangesHasher,
+            @NonNull final StreamingTreeHasher traceDataHasher,
+            @NonNull final byte[][] extensionSubtreeRoots) {
```
jsync-swirlds: I would recommend having a parameter for each additional subtree, rather than using an array of arrays.  We can pass null for each unused subtree.

Note, the existing hashers are wrong in subtle ways, because those should _also_ be null if the subtree isn't used.  There is no guarantee that every subtree will have data; for instance, the Trace Data will _often_ be null and State Changes may occasionally be null.

We should double check that with Consensus Node; particularly because forward compatibility requires all subtrees to be optional, but CN might have the same "subtrees are always present" assumptions.

We should also check how WRBs are hashed, because WRBs have only one subtree filled in; they never add anything to input, consensus headers, state changes, or trace data (everything is output).

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -314,6 +376,11 @@ public static Bytes computeFinalBlockHash(
         Objects.requireNonNull(consensusHeaderHasher);
         Objects.requireNonNull(stateChangesHasher);
         Objects.requireNonNull(traceDataHasher);
+        Objects.requireNonNull(extensionSubtreeRoots);
```
jsync-swirlds: I'd recommend removing the "require non null" and _handle_ null by either returning empty bytes (so the block proof fails validation) or correctly constructing the "merkle mountain top" with only the subtrees that actually have data.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -337,13 +404,58 @@ public static Bytes computeFinalBlockHash(
...
+            node = null;
+        } else if (left == null) {
+            node = hashInternalNodeSingleChild(right);
+        } else if (right == null) {
+            node = hashInternalNodeSingleChild(left);
+        } else {
+            node = hashInternalNode(left, right);
+        }
+        return node;
+    }
```
jsync-swirlds, line 495: I think this should be used for all of the subtrees, rather than just the extensions.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -216,6 +167,221 @@ private boolean isCanceled() {
...
+    /// @param item the block item to process
+    /// @param itemsBlockNumber the block number carried by the current block items record
+    /// @return null on success, or the failure type when the block must be refused
+    /// @throws ParseException if a known item fails to parse
+    private SessionFailureType processItem(final BlockItemUnparsed item, final long itemsBlockNumber)
+            throws ParseException {
+        final BlockItemUnparsed.ItemOneOfType kind = item.item().kind();
+        final List<UnknownField> unknownFields = item.getUnknownFields();
+        final SessionFailureType failure;
+        if (kind != BlockItemUnparsed.ItemOneOfType.UNSET && !unknownFields.isEmpty()) {
```
ata-nas, line 216: Note that we expect that when we have an oneOf UNSET, we must have an unknown field. Also, we expect that if we DO NOT have oneOf UNSET, we must also not have any unknown fields. If that is not the case, we reject. We have to confirm this, but this is my understanding.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -216,6 +167,221 @@ private boolean isCanceled() {
...
+    /// upgraded node.
+    /// @param item the block item carrying the unknown field
+    /// @param unknownFields the unknown fields of the item
+    /// @return null on success, or the failure type when the block must be refused
+    private SessionFailureType processFutureItem(final BlockItemUnparsed item, final List<UnknownField> unknownFields) {
+        final SessionFailureType failure;
+        if (unknownFields.isEmpty()) {
+            // an item with no field at all, nothing valid to process
+            failure = SessionFailureType.UNKNOWN_ERROR;
+        } else if (unknownFields.size() > 1) {
```
ata-nas, line 314: We expect that we have exactly one item in the unknown fields, otherwise we reject. This is my understanding, but must confirm.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -216,6 +167,221 @@ private boolean isCanceled() {
...
+    /// else refuses the block, because guessing could produce a hash that disagrees with an
+    /// upgraded node.
+    /// @param item the block item carrying the unknown field
+    /// @param unknownFields the unknown fields of the item
+    /// @return null on success, or the failure type when the block must be refused
+    private SessionFailureType processFutureItem(final BlockItemUnparsed item, final List<UnknownField> unknownFields) {
+        final SessionFailureType failure;
+        if (unknownFields.isEmpty()) {
+            // an item with no field at all, nothing valid to process
+            failure = SessionFailureType.UNKNOWN_ERROR;
```
ata-nas, line 313: If we expect to have an unknown field but we don't, maybe we can return UNABLE_TO_PARSE instead of UNKNOWN_ERROR. Seems more fitting.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -216,6 +167,221 @@ private boolean isCanceled() {
...
+        } else if (unknownFields.size() > 1) {
+            // a BlockItem is a oneof, more than one field is malformed
+            failure = SessionFailureType.UNABLE_TO_PARSE;
+        } else {
+            final int fieldNumber = unknownFields.getFirst().field();
+            if (fieldNumber < FIRST_FORWARD_COMPATIBLE_FIELD_NUMBER) {
+                // an unknown field below 20 is a first release field reserved for item types that
+                // require specific handling this version does not know
+                hashingMetrics.futureItemsRefused().increment();
+                failure = SessionFailureType.UNSUPPORTED_ITEM_TYPE;
```
ata-nas, line 324: Is this the correct handling here? For an ordinal below 20, we should have explicit oneOf mapping. This is my understanding. But this could confilct with an extension, for instance we receive an extension, not explicitly defined as an item oneOf with ordinal 12. That is valid and should be accepted. Maybe we should extend the proto definition to also explicitly support the extensions.

I must note that it seems there is a mismatch between the block_item.proto's ordinal placement and what is defined for forward compatibility:

Here is today's definition of block item:

```proto
message BlockItem {
    // Reserved for future items that require separate handling for block hash purposes.
    reserved 13,14,15,16,17,18,20;

    oneof item {
        /**
         * An header for the block, marking the start of a new block.
         */
        com.hedera.hapi.block.stream.output.BlockHeader block_header = 1;

        /**
         * An header emitted at the start of a new network "event".
         * <p>
         * This item SHALL contain the properties relevant to a single
         * gossip event.
         */
        com.hedera.hapi.block.stream.input.EventHeader event_header = 2;

        /**
         * An header emitted at the start of a new consensus "round".
         * <p>
         * This item SHALL contain the properties relevant to a single
         * consensus round.
         */
        com.hedera.hapi.block.stream.input.RoundHeader round_header = 3;

        /**
         * A single transaction.
         * <p>
         * This item SHALL contain the serialized bytes of a
         * single proto.SignedTransaction.<br/>
         * This item MUST NOT contain data for more than one
         * `SignedTransaction`.
         */
        bytes signed_transaction = 4;

        /**
         * The result of running a transaction.
         * <p>
         * This item SHALL be present immediately after an
         * `signed_transaction` item.<br/>
         * This item MAY be redacted in some circumstances, and SHALL be
         * replaced with a `filtered_item` if removed.
         */
        com.hedera.hapi.block.stream.output.TransactionResult transaction_result = 5;

        /**
         * A transaction output.
         * <p>
         * This item MAY not be present if a transaction does not produce
         * an output.<br/>
         * If a transaction does produce an output that is not reflected
         * in state changes, then this item MUST be present after the
         * `transaction_result` for that transaction.
         */
        com.hedera.hapi.block.stream.output.TransactionOutput transaction_output = 6;

        /**
         * A set of state changes.
         * <p>
         * All changes to values in network state SHALL be described by
         * stream items of this type.<br/>
         * The source of these state changes SHALL be described by the
         * `reason` enumeration.
         */
        com.hedera.hapi.block.stream.output.StateChanges state_changes = 7;

        /**
         * A block item intentionally filtered from the stream.
         */
        FilteredSingleItem filtered_single_item = 8;

        /**
         * A signed block proof.<br/>
         * The signed merkle proof for this block. This will validate
         * a "virtual" merkle tree containing the previous block "virtual"
         * root, an "input" subtree, an "output" subtree, and
         * a "state changes" subtree.
         * <p>
         * This item is not part of the block stream hash chain/tree, and
         * MUST follow after the end of a block.
         */
        BlockProof block_proof = 9;

        /**
         * A record file and associated data.
         * <p>
         * This MUST contain a single Record file and associated Sidecar files.
         * If this item is present, special treatment is
         * REQUIRED for this block.
         * <ul>
         *   <li>The block SHALL have a `BlockHeader`.</li>
         *   <li>The block SHALL contain _exactly one_ `RecordFileItem`.</li>
         *   <li>The block MAY contain one or more StateChanges items.</li>
         *   <li>The block SHALL have a `BlockFooter`.</li>
         *   <li>The block SHALL have a `SignedRecordFileProof` type
         *       `BlockProof`.</li>
         *   <li>The block SHALL NOT contain any content item other than a
         *       block header, single `RecordFileItem`, optional StateChanges,
         *       block footer, and one or more block proofs.</li>
         *   <li>The content of the `RecordFileItem` can be validated using
         *       the `SignedRecordFileProof` according to
         *       the process used for Record Files prior to the creation
         *       of Block Stream.<br/>
         *       Additional Block Proof types may be available as well.</li>
         * </ul>
         */
        RecordFileItem record_file = 10;

        /**
         * A trace data.
         * <p>
         * Any informational trace data MAY be described by
         * stream items of this type.<br/>
         */
        com.hedera.hapi.block.stream.trace.TraceData trace_data = 11;

        /**
         * Hashes at the end of each block.
         * <p>
         * This footer SHALL signal the end of information to include
         * in a block's hashed content.<br/>
         * One or more Block Proof items SHALL follow this item.
         */
        com.hedera.hapi.block.stream.output.BlockFooter block_footer = 12;

        /**
         * A block item intentionally redacted from the stream.
         */
        RedactedItem redacted_item = 19;
    }
}
```

Which seems to mismatch with design doc ordinals definitions:

```md
| Category (N mod 20) |                       Meaning                        | Forward compatible? |
|---------------------|------------------------------------------------------|---------------------|
| 0                   | Not hashed (not part of the block proof merkle tree) | Yes                 |
| 1                   | Requires specific handling                           | Partially           |
| 2                   | Requires specific handling                           | Partially           |
| 3                   | `consensus headers`                                  | Yes                 |
| 4                   | `inputs`                                             | Yes                 |
| 5                   | `outputs`                                            | Yes                 |
| 6                   | `state changes`                                      | Yes                 |
| 7                   | `trace data`                                         | Yes                 |
| 8                   | Extension 0                                          | Yes                 |
| 9                   | Extension 1                                          | Yes                 |
| 10                  | Extension 2                                          | Yes                 |
| 11                  | Extension 3                                          | Yes                 |
| 12                  | Extension 4                                          | Yes                 |
| 13                  | Extension 5                                          | Yes                 |
| 14                  | Extension 6                                          | Yes                 |
| 15                  | Extension 7                                          | Yes                 |
| 16 to 18            | Reserved for future use                              | No                  |
| 19                  | Not hashed (not part of the block proof merkle tree) | Yes                 |
```

ata-nas (commented):
Reviewer's reference.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Changed as per your recommendation.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Changed as per your recommendation, hashers are now standalone fields.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
ata-nas: Removed as recommended, hashers are standalone fields and are all explicitly passed to the root hash calculation utility.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
ata-nas: Changed as per your recommendation, but only for extension hashers. Note that I will pre calculate the roots and pass a nullable byte[] rather than the streaming hasher, the existing hashers are passed as objects and calculate the root inside. We can consolidate both and pre-calcualte the hashes for them as well. Even though a CompletableFuture is returned when calling `rootHash` of the `NaiveStreamingTreeHasher`, it returns a completed future, meaning we do not actually run it async. Not sure why a completable future wrapper even exists, but that is another story.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
ata-nas: Removed, all 8 streaming hashers produce nullable byte[] results that are passed explicitly as arguments.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
ata-nas: Agreed, however after a discussion, for this PR, we will only add support for the extensions.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
jsync-swirlds: Let's file an issue to revise this so we either pass byte arrays for all hashers, or pass hashers for all subtrees (I lean toward the latter, also nullable).
That issue should probably also include recommendation to revisit use of the CompletableFuture.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
jsync-swirlds: Agreed.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: We explicitly call out all ordinals below 20 as reserved because the initial fields are not necessarily compliant with the field number mapping.

This is deliberate, and we do not want to use field number mapping for any field with number < 20.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: I prefer unknown error for this case; we successfully parsed the content, but it doesn't make sense, so something unexpected is happening.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: That's not accurate, unknown fields is any number of fields.  We might have data from a new release with several new fields added all at once if they're added outside the oneOf.
I'm not actually certain we'll even get the `UNSET` result, only that unknown fields map _will_ contain all of the otherwise unknown fields.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: That may not be valid.
I'm not sure PBJ creates an unset entry, it may just put all unknown fields in the map and not create a regular field entry for any of them.

I think we need to test this.

ata-nas: By default, constructing a new item object sets the kind to UNSET. Not sure how a parser would parse an unknown field, however.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
Nana-EC: It seems we're currently using `UNABLE_TO_PARSE` in a broad fashion to imply "the message structure is not what we can process" even when the wire encoding is valid.
That's probably where the consideration came from.

I agree `UNKNOWN_ERROR` is better here but only because there's no a clear type for "valid encoding but unambiguous in meaning".

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/hasher/BlockHasherTest.java
```diff
@@ -621,6 +627,379 @@ void testHeaderNotFirstItem() throws ParseException {
...
+                    switch (category) {
+                        case 0, 19 -> {
+                            // not hashed
+                        }
+                        case 3 -> consensusLeaves.add(leaf);
+                        case 4 -> inputLeaves.add(leaf);
+                        case 5 -> outputLeaves.add(leaf);
+                        case 6 -> stateChangesLeaves.add(leaf);
+                        case 7 -> traceLeaves.add(leaf);
+                        default -> extensionLeaves.get(category - 8).add(leaf);
```
Nana-EC, line 914: This may be problematic and result in `IndexOutOfBoundsException` for categories 1,2, 16 and 18.
Should we be more explicit here with the known cases.
Maybe
```suggestion
                        case 8, 9, 10, 11, 12, 13, 14, 15 -> extensionLeaves.get(category - 8).add(leaf);
default -> throw new IllegalArgumentException("Unsupported category in reference: " + category)
```

Nana-EC (commented):
LG. One test case consideration gap.
Also may be worth a new error case that covers the gpa of parseable but ambiguous.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: When parsing it's up to PBJ, and my guess (again, we have to check) is that PBJ just places the field in `unknownFields` and _doesn't represent the field elsewhere_.

I could be mistaken, of course, so we need to test this.

We should have a regression test for exactly these things anyway.
Perhaps use a modified BlockItem proto that's only compiled for testing, that adds fields in the range from 10,020-10,040; fills in values for some, serializes the result, then passes to this code to parse and validate the various handling cases.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: We could add something like `UNSUPPORTED_STREAM_FORMAT` to represent "we can parse this, but it is not a processable Block Stream for other reasons."

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/hasher/BlockHasherTest.java
jsync-swirlds: We may also want to be specific about 1,2, and 16-18 as unsupported.

Then again, this code is creating a _reference_ hash; I assume we know exactly what is going in to this code and whether it will have any of those other fields.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -216,6 +167,221 @@ private boolean isCanceled() {
...
+        final SessionFailureType failure;
+        if (unknownFields.isEmpty()) {
+            // an item with no field at all, nothing valid to process
+            failure = SessionFailureType.UNKNOWN_ERROR;
+        } else if (unknownFields.size() > 1) {
+            // a BlockItem is a oneof, more than one field is malformed
+            failure = SessionFailureType.UNABLE_TO_PARSE;
+        } else {
+            final int fieldNumber = unknownFields.getFirst().field();
+            if (fieldNumber < FIRST_FORWARD_COMPATIBLE_FIELD_NUMBER) {
```
jsync-swirlds, line 320: FYI.
This is where we ensure we don't apply ordinal rules to fields below 20.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: So, we can use a modified proto file for tests only, but I believe I've managed to cover such tests without such file.

Take a look at `ForwardCompatibilityTests`. I am building raw`Bytes` with the `fieldBytes` which then get parsed via `standardParse(BlockItemUnparsed.PROTOBUF, ...)` to construct a `BlockItemUnparsed`.

The result of doing that is that we have a oneOf UNSET + the unknown fields populated with the generated `Bytes`:
image

Let me know if these tests cover what you list above. A note that I will make is that these tests use a reference implementation so not to use the actual production logic. If that changes and it produces a different hash, we would like a failure to happen, otherwise tests would assert `new_prod_logic.call == new_prod_logic.call` instead of `ref_impl.call == new_prod_logic.call`

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: I can add the `UNSUPPORTED_STREAM_FORMAT ` suggestion, let me know.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Noted I also thought that this mismatch should not be an issue as we have to explicitly handle cases < 20, so if we have an unknown field < 20, we have to reject. That is my understanding.

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/hasher/BlockHasherTest.java
ata-nas: So what's the action here? Be specific about 1-2 && 16-18 as unsupported, or also be specific about the `default` branch as well?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: So the UNSET type should be present, it is the default one if oneOf is not set explicitly. This is true even if the new fields are added outside of the oneOf.

As far as multiple unknown fields, yes, this could happen. Tried that with a generated new proto definition:
image

So not sure what would be the correct action here.

ata-nas: I think in general, each item should be standalone, if someone constructs an item with multiple unknowns, because they have been declared outside the oneof, yes that can happen and will parse, but isn't that wrong to do in the first place? In the pic above, shouldn't we simply have received 3 items each UNSET + a single unknown field?

ata-nas: image

This also passes.

Note both this and test above are not part of this PR, we can choose to add the FutureBlockItem proto file created for testing and add more tests with that.

So in this case, if we run it through the hasher, we should expect a failure, as we cannot have one of our pre-defined oneOfs with an unknown field. This test does not asser that however, just showing parser behavior.

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/hasher/BlockHasherTest.java
jsync-swirlds: Not sure there is one; we can be specific if you believe it makes the code clearer, or not be specific if you believe that creating the reference hash does not require that extra specificity.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: Yes.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: That sounds like a good option.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: 1. We do need to add something like the FutureBlockItem proto file as something compiled _only_ in test fixtures and used for tests.
1. It _is acceptable_ to add multiple fields, and we should allow for it.  We don't know what can/will change in the future, and do not want to break if we can still correctly handle the change.
   - example, a new update adds fields 20, 39, 40, and 59 to the proto, all outside the oneof.  This should work, be accepted, and all 4 new fields be left out of the hash; even if all 4 are present on every block item.
   - example, a new update adds fields 28, 29, 30, and 31 to the proto, all outside the oneof, and each is sometimes present.  This should work, and each field is assigned to the appropriate "extension" subtree, when present, according to the field number rules.

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/hasher/BlockHasherTest.java
Nana-EC: Make it specific low bar for us to do now, high cost if something slips through

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: About point 2, if I understand correctly, say we receive an item that is UNSET and has 3 unknown fields.
In that case, we do nothing for the UNSET item itself, we do not add it to the hash, but we have to iterate over the 3 unknown fields, resolve the place of each of them, then get the bytes of the unknown fields and hash them under the correct hasher. Is that correct?

Then, what happens if we have say a BLOCK_HEADER which also has 3 unknown fields? Is that possible? If so, then how do we hash the header itself? The current logic will transform the item (which includes the 3 unknown fields) and add that to output hasher, and then we go through all the unknowns, resolving their placement and hashing the bytes they hold under the respective hasher?

Something seems way off, I am probably misunderstanding something.

The current branch with the current changes only expects 1 unknown field for an UNSET item and we simply transform the UNSET item which holds the unknown field into bytes using the `BlockItemUnparsed.PROTOBUF.toBytes(item)` method and add that result to the respective hasher. But that does not and cannot support us having multiple unknown fields, as we will either have to hash the whole item transformed to bytes (I think there is something off here), or not hash the hole item transformed to bytes, but only hash the value of the unknown fields (and there are some wire types and some prefixes which I have to dig into as well, it seems).

For the known items, we do exactly the same operation, we do `BlockItemUnparsed.PROTOBUF.toBytes(item)` and the resulting bytes get added to the respective hasher. This was always the case from the past.

Let's talk about this.

jsync-swirlds: Hopefully short answer:
1. Yes, with forward compatibility we need to handle unknown fields for any BlockItem; if we had a Header with 3 unknown fields, we would hash the header, and hash (or reject, based on field number) each of the unknown fields as well.
1. If we have a oneOf that is unset, we should check that the field number is in the unknown fields; and handle both "present in unknown" and "extant but not present in unknown" (if we can do so).

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
jsync-swirlds: Part of why I want the proto file is that I want to have a test that verifies PBJ-generated bytes that are PBJ-parsed work the way we expect, even as PBJ changes in future versions.  It might work today and break in a future PBJ update, and I'd prefer we catch that with a test.

We can add that test as an addition to the test(s) you already have.

ata-nas: I have those changes locally, will push them.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
ata-nas: @jsync-swirlds I created #3372, will also tag inside the code here.

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/hasher/BlockHasherTest.java
ata-nas: ✅ Done. All possible branches 0-19 incl. are specified. Default branch returns UNKNOWN_ERROR. Proto ordinals below 20 on unknown fields are automatically rejected.

ata-nas:
@jsync-swirlds @Nana-EC note that here: 268e317333724b510da692ba36b0213c9bd39c4a
I've added some improvements to state proof visited path accuracy + tests.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: :white_check_mark: Added the unsupported format

ata-nas:
@Nana-EC @jsync-swirlds just to confirm for this PR:

1. We do not change the pre-defined hashers, they will have 0 prefix if missing
2. For extension hashers
   - we will also hash the proto message and add 0 prefix to be the same as predefined ones
   - we will not add to the root hash if they are missing (0x0 hash)
3. Forward compatibility support
   - we will reject items that are != UNSET && have unknown fields
   - we will reject more than one unknowns
   - we will reject unknowns with ordinals < 20
Follow ups will address changes.
Can we confirm?

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/StateProofVerifier.java
```diff
@@ -204,8 +208,8 @@ private SessionFailureType followNextPath(
                 return SessionFailureType.BAD_BLOCK_PROOF;
             } else {
                 final MerklePath nextPath = allPaths.get(currentJoinPointIndex);
-                visited[currentJoinPointIndex] = true;
```
ata-nas, line 207: @jsync-swirlds should we move `visited[currentJoinPointIndex] = true;` above the `if (isJoinPoint(nextPath))` check like it was before?

ata-nas: reverted as discussed + added a trace log here: 1c849b379db376c90e4d9cf2098da60c28a3d2c4

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: @jsync-swirlds prefer to add in follow up. Created #3376

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: As discussed, we will address this in a follow up.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -337,13 +416,81 @@ public static Bytes computeFinalBlockHash(
...
+            final byte[] rootSix,
+            final byte[] rootSeven) {
+        final byte[] depth5Node5 = combineOptionalNodes(rootZero, rootOne);
+        final byte[] depth5Node6 = combineOptionalNodes(rootTwo, rootThree);
+        final byte[] depth5Node7 = combineOptionalNodes(rootFour, rootFive);
+        final byte[] depth5Node8 = combineOptionalNodes(rootSix, rootSeven);
+        final byte[] depth4Node3 = combineOptionalNodes(depth5Node5, depth5Node6);
+        final byte[] depth4Node4 = combineOptionalNodes(depth5Node7, depth5Node8);
+        return combineOptionalNodes(depth4Node3, depth4Node4);
+    }
```
jsync-swirlds, line 473: This probably belongs in another PR based on recent discussions.
There is a missing item here that if a given "leaf" node is non-empty (say rootTwo), then all "leaf" nodes "left" of that node (rootZero, rootOne) must be present (empty leaf hash or actual entries).

jsync-swirlds (approved):

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
ata-nas: Yes, this shall be a follow up.

rockysingh (approved):

PR #3301, opened 2026-07-28
fix(ci): GA automation should release the latest RC version, if one exists, else the release branch

### LINKED ISSUE #3300: GA release automation should release the latest tested RC tag if one exists.
### Story Form

As a BlockNode engineer I want to make sure that GA releases take the latest tested code from a rc tag so that we know we are releasing what was tested.

### Technical Notes

if any commit sneaks onto the release branch between the last RC and GA (a stray hotfix, a bad cherry-pick, etc.), GA would silently ship that extra code untested, rather than the exact bits that were validated as the RC.

berryware:
## Reviewer Notes

Summary of changes to [release-automation.yaml](.github/workflows/release-automation.yaml):

- **`Compute Release Version`**: now exports `SOURCE_TAG=v$CURRENT` when the version on the dispatch ref is an rc/alpha (this is exactly the tag a prior run of the workflow created). Stays unset if there was never an rc/alpha (snapshot).
- **`Create and Switch to Release Branch`**: for GA with a resolved `SOURCE_TAG`, fetches and checks out that tag detached instead of the release branch tip. Falls back to the existing `git checkout ${RELEASE_BRANCH}` behavior in every other case (non-GA, or GA with no prior rc/alpha).
- **`Commit and Tag`**: now skipped for GA (`if: release_type != 'GA'`), unchanged for rc/alpha/custom.
- **New `Commit and Tag (GA)`**: manually commits the version bump and creates the signed annotated tag, then pushes only `refs/tags/<tag>` — it never pushes/updates the release branch ref, whether GA was built from the rc tag or (fallback) from the branch tip.

Validated the YAML parses correctly and the step list still resolves as expected. 

## Related Issue(s)
 Closes #3300

### DIFF
--- a/.github/workflows/release-automation.yaml
+++ b/.github/workflows/release-automation.yaml
@@ -85,6 +85,17 @@ jobs:
               exit 1
             fi
 
+            # For a GA release, CURRENT is the last rc/alpha tagged by a previous
+            # run of this workflow (e.g. v0.38.0-rc2). GA must be built from that
+            # exact tag, not from whatever the release branch currently points
+            # to, so a stray commit landing on the branch after the last rc
+            # can't sneak into the GA release untested. If no rc/alpha ever
+            # existed for this version (KIND=snapshot), SOURCE_TAG stays unset
+            # and GA falls back to building from the release branch tip.
+            if [[ "$KIND" == "rc" || "$KIND" == "alpha" ]]; then
+              echo "SOURCE_TAG=v$CURRENT" >> "$GITHUB_ENV"
+            fi
+
             case "$INPUT_RELEASE_TYPE" in
               rc)
                 [[ "$KIND" == "rc" ]] \
@@ -157,6 +168,8 @@ jobs:
         uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.2.0
 
       - name: Create and Switch to Release Branch
+        env:
+          INPUT_RELEASE_TYPE: ${{ github.event.inputs.release_type }}
         run: |
           git fetch origin
           if ! git ls-remote --exit-code --heads --quiet origin refs/heads/${RELEASE_BRANCH}; then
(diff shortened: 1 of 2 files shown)


.github/workflows/release-automation.yaml
Nana-EC, line 279: For completeness this would also need to be changed.
Other in the normal flow the commit to take a a release branch from `0.x.y-rcz` to `0.x.y` itself would be orphaned.

.github/workflows/release-automation.yaml
```diff
@@ -192,6 +211,22 @@ jobs:
           commit_user_email: ${{ steps.gpg_importer.outputs.email }}
           tagging_message: ${{ env.RELEASE_TAG }}
 
+      - name: Commit and Tag (GA)
+        if: ${{ github.event.inputs.release_type == 'GA' }}
+        env:
```
Nana-EC, line 216: Q: Does the `git-auto-commit-action` not support the goal here?
Curious of the diff in approach

.github/workflows/release-automation.yaml
```diff
@@ -192,6 +211,22 @@ jobs:
...
+          GIT_AUTHOR_EMAIL: ${{ steps.gpg_importer.outputs.email }}
+          GIT_COMMITTER_NAME: ${{ steps.gpg_importer.outputs.name }}
+          GIT_COMMITTER_EMAIL: ${{ steps.gpg_importer.outputs.email }}
+        run: |
+          # GA is built on a detached HEAD (see "Create and Switch to Release
+          # Branch") so, unlike the rc/alpha path above, only the tag is
+          # pushed here — the release branch ref is intentionally left alone.
+          git add -A
+          git commit --no-verify --signoff -m "Bump versions for ${RELEASE_TAG}"
+          git tag -a "${RELEASE_TAG}" -m "${RELEASE_TAG}"
```
Nana-EC: Q: should this be a signed tag instead of an annotated tag?
```suggestion
          git tag -s "${RELEASE_TAG}" -m "${RELEASE_TAG}"
```

Nana-EC (commented):
Good catch, some questions

jsync-swirlds (approved):

Nana-EC (approved):

.github/workflows/release-automation.yaml
berryware: t doesn't cleanly. It's built to commit-and-push onto a branch; running it against a detached HEAD requires its create_branch: true + branch: inputs, which creates and pushes a new branch to origin. That's not what we want (we explicitly don't want to touch/create any branch ref for GA), so hand-rolled git commands were the only way to get "commit + tag + push only the tag."

PR #3302, opened 2026-07-28
fix: Superseded Verification Session must not Report a Cancellation

### LINKED ISSUE #3303: Superseded Verification Session must not Report a Cancellation
### Story Form

As a Verification Plugin
I want to not report a cancellation of an active publisher session when the same block is restarted
So that I do not ask for a resend I do not need

### Technical Notes

- Restarting a block on publisher should not report cancelation of current session that is received from publisher, assuming the restarted block is the same as the one we have a session for

ata-nas:
* Cancel an active publisher session silently when a new session for the same block supersedes it
  * A stale session left by a publisher severing mid-block raced with the resent block: its CANCELLED failure was treated downstream as a real verification failure, re-adding the in-flight block to blocksToResend and suppressing all further acknowledgements
  * The superseding session owns the fate of the block: it reports success or failure itself, and blockIsEnding schedules a resend if its stream dies
* Keep the CANCELLED failure report when the canceled session is for a different block, so the abandoned block is scheduled for resend and is not lost downstream
* Suppress by failure type, not exception type, so a cancellation surfacing as either CancellationException or a CANCELLED session failure is covered
* Add tests
  * Superseded same block session reports no verdict, only the success of the superseding session
  * Canceled session for a different block still reports CANCELLED
  * Update the existing cancellation test to interrupt with a different block

Resolves #3303

### DIFF
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
@@ -106,7 +106,17 @@ public final class BlockSessionHandler {
         BlockVerificationSession local = activePublisherSession.get();
         if (blockItems.isStartOfNewBlock()) {
             if (local != null) {
-                local.cancel();
+                if (local.sessionKey().blockNumber() == blockItems.blockNumber()) {
+                    // The new session supersedes the active one for the same block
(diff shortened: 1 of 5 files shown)


ata-nas:
Closing, we will not go with this change.
The change couples verification with publisher, which is something we do not want to do.
Verification should simply report the failure, others should take an action, rather than verification not reporting it.

PR #3308, opened 2026-07-28
feat(logging): status heartbeat, INFO-noise refinement, and logging docs

### LINKED ISSUE #3278: chore: log level refinement — reduce INFO noise and surface block progression milestones
## Summary

Current INFO-level logging produces too much noise from per-block construction events while hiding genuinely useful operational signals that live at DEBUG. The result is operators either drown in per-block chatter or miss important block progression and persistence milestones unless they enable DEBUG globally.

This issue collects concrete suggestions across three categories:

---

## 1. INFO logs that are too verbose — should be DEBUG

These fire every block (or on every probe/HTTP call) and inflate INFO output without conveying useful operator-level signals:

| File | Line | Current message | Why demote |
|------|------|----------------|------------|
| `block-node/verification/.../session/impl/DummyVerificationSession.java` | 47 | `Created DummyVerificationSession for block {0}` | Fires every block in non-cryptographic mode. Per-block session construction is an implementation detail, not an operator event. |
| `block-node/verification/.../session/impl/ExtendedMerkleTreeSession.java` | 176 | `Created ExtendedMerkleTreeSession for block {0}` | Same pattern — fires per block during live streaming and backfill. Session creation is internal bookkeeping. |
| `block-node/health/.../HealthServicePlugin.java` | 121, 143 | `Responded code 503 (Service is not running) to liveness check` / `readiness check` | During startup races or probe bursts a 503 can be expected and transient. Logging at INFO on every probe response floods logs during brief unavailability windows. Reserve for WARN on persistent failures. |
| `block-node/verification/.../VerificationServicePlugin.java` | 353 | `Received block items before a block header, ignoring.` | Expected at startup/reconnect (noted in the code comment). Logging at INFO makes it appear alarming; the comment already says this is normal. |

---

## 2. INFO logs that are good but could be expanded

These are at the right level but missing context that would make them genuinely actionable:

| File | Line | Current message | Suggested expansion |
|------|------|----------------|---------------------|
| `block-node/verification/.../VerificationServicePlugin.java` | 230 | `Loaded TSS parameters from file: {0}` | Add the count of TSS key shares or the era/block range covered so operators can verify the correct parameter set loaded. e.g. `Loaded TSS parameters from file: {0} (shares={1}, era={2})` |
| `block-node/stream-publisher/.../LiveStreamPublisherManager.java` | 310 | `Handler {0} is ending mid-block {1}` | Include the reason/cause. A handler dropping mid-block is significant; knowing whether it was a network error, timeout, or protocol violation helps triage. e.g. `Handler {0} ending mid-block {1}: reason={2}` |
| `block-node/stream-subscriber/.../BlockStreamSubscriberSession.java` | 898 | `Handler for {0} has fallen behind.` | Include how far behind — current send position vs latest available block. e.g. `Subscriber {0} has fallen behind: sendingBlock={1} latestBlock={2} lag={3}` |
| `block-node/blocks-file-historic/.../BlockFileHistoricPlugin.java` | 573 | `Could not get staged files for blocks {0} to {1}.` | Add how many blocks were in the batch and whether any were partially available, so operators can distinguish a full miss from a partial staging failure. |
| `block-node/backfill/.../BackfillPlugin.java` | ~623 | high-water mark reset message | Include old and new high-water mark values so gap-tracking state changes are traceable. e.g. `Backfill high-water mark reset from {0} to {1}` |

---

## 3. DEBUG logs that are good candidates for INFO

These capture meaningful block progression and persistence milestones that operators should see without enabling DEBUG:

| File | Line | Current message | Why promote |
|------|------|----------------|-------------|
| `block-node/verification/.../VerificationServicePlugin.java` | 364 | `COMPLETED_MESSAGE` — block verification result (success/failure) | **Most important.** Block verification outcome is the core operational signal of a block node. Operators should see pass/fail at INFO without debug mode. Currently only the `WARNING` path for failures surfaces; successes are completely invisible. |
| `block-node/verification/.../VerificationServicePlugin.java` | 344 | `Started new block verification session for block number {0}` | Pairs with the completion message to bracket a block's lifecycle. Promoting both gives a clear per-block start→end progression at INFO. |
| `block-node/blocks-file-recent/.../BlockFileRecentPlugin.java` | 411 | `Wrote verified block {0} to file {1}` | Persistence success is a key durability milestone. Operators troubleshooting disk/storage issues need this at INFO. |
| `block-node/blocks-file-historic/.../BlockFileHistoricPlugin.java` | 635 | `Successfully moved batch of blocks [{0} -> {1}] to zip file.` | Archive batch completion is a meaningful storage lifecycle event — the "blocks are now cold" signal. INFO-worthy. |
| `block-node/stream-publisher/.../LiveStreamPublisherManager.java` | 208, 238 | `Added new handler {0}` / `Removed handler {0}` | Publisher connect/disconnect events affect liveness directly. Operators need to see publisher churn at INFO to detect connection instability without debug mode. |
| `block-node/backfill/.../BackfillPlugin.java` | ~369 | `detectedGapMsg` — gap detected (type + range) | Gap detection is an operational alert — the node identified missing blocks. INFO is appropriate so operators catch gaps without debug. |
| `block-node/backfill/.../BackfillPlugin.java` | ~497 | `submittingGapMsg` — gap submitted (type + range) | Pairs with gap detection to show corrective action was taken. Together they give a complete gap→repair story at INFO. |

---

## Acceptance Criteria

- [ ] All log level changes in section 1 are applied (INFO → DEBUG)
- [ ] All log messages in section 2 are expanded with the suggested additional context
- [ ] All log level changes in section 3 are applied (DEBUG → INFO)
- [ ] No test breakage from log level changes (check any tests that assert on log output)
- [ ] A BN operator can follow block ingestion, verification, and persistence milestones through INFO logs alone without enabling DEBUG globally

AlfredoG87:
Fixes #3278

## Context

Production Block Nodes enable INFO and above only, so INFO/WARNING/ERROR quality determines whether operators can run the node. Issue #3278 asked to reduce INFO noise and surface block progression. Investigation showed the code had drifted from the project's own `docs/logging-guidelines.md`, and the single largest signal/noise defect — caught-exception logs emitted at INFO, indistinguishable from routine chatter — was not in the issue at all. This PR brings INFO+ back in line with the guidelines, adds one principled progression signal for log-only operators, and documents logging end to end.

## Governing policy (metrics for progress, logs for problems)

Per the guidelines and the maintainer discussion on #3278: per-item success/progress is a metrics concern, not per-item INFO logs; success is never INFO. Log-only operators are served by (1) a single periodic aggregate INFO heartbeat and (2) documented per-package DEBUG on demand.

## What changed

Server status heartbeat
- New periodic INFO line from `ServerStatusServicePlugin` (`Status heartbeat: oldestBlock=… newestBlock=… nextExpected=…`), emitted from data the plugin already holds — no new cross-service dependency.
- New `server.status.heartbeatPeriodSeconds` config (default 300 s = 5 min, ≤0 disables).

Log-level corrections (the main #3278 work)
- ~35 caught-exception logs moved off INFO: WARNING for recoverable failures, ERROR for scheduler-creation and the backfill uncaught-thread handler (now consistent with s3-archive / roster-bootstrap-rsa), DEBUG for two transient cases (upstream status fetch, temp-archive meta race).
- Client-controllable invalid-request log demoted to DEBUG.
- Several sites also had a lost stack trace (the exception was passed as a format argument, not the `Throwable`); fixed to log the exception properly.

Documentation
- `docs/design/architecture/Logging-Architecture.md` — design doc: framework, filesystem/config layout, level policy, patterns, decisions & alternatives discarded.
- `docs/block-node/logging.md` — operator/developer reference: format, logs-vs-metrics, per-class/package/global level how-to with examples.
- `docs/block-node/troubleshooting.md` §1.1 expanded into a playbook; gap-fills in `configuration.md`, `README.md`, and `logging-guidelines.md`.

## Notes for reviewers

- The `block-node/verification` module is the obsolete pre-#3253 module (only `gradle/aggregation` references it); its per-block INFO lines flagged in #3278 are dead code and were left untouched. Live verification logging is in `block-verification`.
- A few WARNING vs DEBUG/ERROR calls are judgment calls that affect on-call paging — please sanity-check: `BlockNodeApp` logging-config bootstrap fallback, `RosterBootstrapTssPlugin` parse-failed, `BackfillFetcher` upstream status fetch (set to DEBUG), `BackfillPlugin` scheduler-creation (set to ERROR), `TempArchiveUploadTask` meta-write race (set to DEBUG), and the two `VerificationServicePlugin` handler failures.

## Testing

- [x] All Test Pass
- [x] New Heartbeat test passes
- [x] Collect and inspect a raw piece of logs from local build
- [x] Collect and inspect a raw piece of logs from CI build

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,8 +72,9 @@ k6-out/
 # LLM agent files
 .agent/**
 agent/**
-# Ignore only local, per-developer Claude settings; shared .claude/ files (settings.json,
-# commands, skills) remain committable.
-.claude/settings.local.json
-.agents.local/
+# Ignore per-developer "*.local" settings files and directories (e.g. .claude/settings.local.json,
+# .agents.local/), for both files and directories at any depth. Shared .claude/ files
+# (settings.json, commands, skills) remain committable.
+*.local
+*.local.*
 
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,8 @@ k6-out/
 # LLM agent files
 .agent/**
 agent/**
-.claude/
+# Ignore only local, per-developer Claude settings; shared .claude/ files (settings.json,
+# commands, skills) remain committable.
+.claude/settings.local.json
 .agents.local/
 
--- a/.gitignore
+++ b/.gitignore
@@ -72,4 +72,6 @@ k6-out/
 # LLM agent files
 .agent/**
 agent/**
+.claude/
+.agents.local/
 
--- a/block-node/app/docker/ci-logging.properties
+++ b/block-node/app/docker/ci-logging.properties
@@ -13,8 +13,8 @@
 # Available Levels are (from most verbose to least verbose):
 # ALL FINEST FINER FINE CONFIG INFO WARNING SEVERE OFF
 .level=INFO
-# Block Node components log at INFO in production; raise a package/class here to debug
-org.hiero.block.level=INFO
+# CI keeps Block Node components at FINE (DEBUG) so flaky/transient test failures stay debuggable
+org.hiero.block.level=FINE
 # Helidon loggers
 io.helidon.level=INFO
 io.helidon.webserver.level=INFO
--- a/block-node/app/docker/ci-logging.properties
+++ b/block-node/app/docker/ci-logging.properties
@@ -13,8 +13,8 @@
 # Available Levels are (from most verbose to least verbose):
 # ALL FINEST FINER FINE CONFIG INFO WARNING SEVERE OFF
 .level=INFO
-# logs for Block Node components FINE (DEBUG) temporarily
-org.hiero.block.level=FINE
+# Block Node components log at INFO in production; raise a package/class here to debug
+org.hiero.block.level=INFO
 # Helidon loggers
 io.helidon.level=INFO
 io.helidon.webserver.level=INFO
--- a/block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
@@ -122,7 +122,7 @@ final class ApplicationStateUtility {
                     NetworkData.JSON, Bytes.wrap(Files.readAllBytes(path)), MAX_APP_STATE_MESSAGE_SIZE_BYTES);
             return data;
         } catch (ParseException | IOException e) {
-            LOGGER.log(WARNING, "Failed to read network data file %s.".formatted(path), e);
+            LOGGER.log(INFO, "Failed to read network data file %s.".formatted(path), e);
             return NetworkData.DEFAULT;
         }
     }
--- a/block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
@@ -2,7 +2,6 @@
 package org.hiero.block.node.app;
 
 import static java.lang.System.Logger.Level.DEBUG;
-import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.WARNING;
 import static org.hiero.block.node.base.ParseHelper.standardParse;
 
@@ -122,7 +121,7 @@ final class ApplicationStateUtility {
                     NetworkData.JSON, Bytes.wrap(Files.readAllBytes(path)), MAX_APP_STATE_MESSAGE_SIZE_BYTES);
             return data;
         } catch (ParseException | IOException e) {
-            LOGGER.log(INFO, "Failed to read network data file %s.".formatted(path), e);
+            LOGGER.log(WARNING, "Failed to read network data file %s.".formatted(path), e);
             return NetworkData.DEFAULT;
         }
     }
@@ -164,7 +163,7 @@ final class ApplicationStateUtility {
                 kf.generatePublic(new X509EncodedKeySpec(keyBytes));
                 usable++;
             } catch (InvalidKeySpecException | IllegalArgumentException e) {
-                LOGGER.log(INFO, "Malformed RSA_PubKey for node {0} — skipped: {1}", addr.nodeId(), e);
+                LOGGER.log(WARNING, "Malformed RSA_PubKey for node %s — skipped".formatted(addr.nodeId()), e);
             }
         }
         if (usable == 0) {
--- a/block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
@@ -122,7 +122,7 @@ final class ApplicationStateUtility {
                     NetworkData.JSON, Bytes.wrap(Files.readAllBytes(path)), MAX_APP_STATE_MESSAGE_SIZE_BYTES);
             return data;
         } catch (ParseException | IOException e) {
-            LOGGER.log(INFO, "Failed to read network data file %s.".formatted(path), e);
+            LOGGER.log(WARNING, "Failed to read network data file %s.".formatted(path), e);
             return NetworkData.DEFAULT;
         }
     }
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -168,7 +168,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
                     LOGGER.log(INFO, "No logging configuration found");
                 }
             } catch (IOException e) {
-                LOGGER.log(INFO, "Failed to load logging configuration", e);
+                LOGGER.log(WARNING, "Failed to load logging configuration", e);
             }
             CleanColorfulFormatter.makeLoggingColorful();
             LOGGER.log(DEBUG, "Using default logging configuration");
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -192,7 +192,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
                     LOGGER.log(INFO, "No logging configuration found");
                 }
             } catch (IOException e) {
-                LOGGER.log(INFO, "Failed to load logging configuration", e);
+                LOGGER.log(WARNING, "Failed to load logging configuration", e);
             }
             CleanColorfulFormatter.makeLoggingColorful();
             LOGGER.log(DEBUG, "Using default logging configuration");
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -192,7 +192,7 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
                     LOGGER.log(INFO, "No logging configuration found");
                 }
             } catch (IOException e) {
-                LOGGER.log(WARNING, "Failed to load logging configuration", e);
+                LOGGER.log(INFO, "Failed to load logging configuration", e);
             }
             CleanColorfulFormatter.makeLoggingColorful();
             LOGGER.log(DEBUG, "Using default logging configuration");
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
@@ -124,6 +124,8 @@ public class BackfillFetcher implements PriorityHealthBasedStrategy.NodeHealthPr
      */
     public List<LongRange> getNewAvailableRanges(long latestStoredBlockNumber) {
         final List<LongRange> peerRanges = new ArrayList<>();
+        final int totalNodes = blockNodeSource.nodes().size();
+        int reachedNodeCount = 0;
 
         for (BlockNodeSourceConfig node : blockNodeSource.nodes()) {
             if (isInBackoff(node)) {
@@ -144,12 +146,21 @@ public class BackfillFetcher implements PriorityHealthBasedStrategy.NodeHealthPr
                         peerRanges.add(new LongRange(start, range.end()));
                     }
                 }
+                reachedNodeCount++;
             } catch (RuntimeException e) {
                 final String failedToGetStatusMsg = "Failed to get status from node [%s:%d]: %s"
                         .formatted(node.address(), node.port(), e.getMessage());
                 LOGGER.log(DEBUG, failedToGetStatusMsg, e);
                 markFailure(node);
             }
+        }
+
+        // Individual unreachable/failed nodes are DEBUG, but no reachable source at all means backfill
+        // cannot progress: surface that aggregate condition once at WARNING for on-call visibility.
+        if (totalNodes > 0 && reachedNodeCount == 0) {
+            final String allSourcesUnreachableMsg =
+                    "All [{0}] backfill source(s) are unreachable; cannot discover peer ranges";
+            LOGGER.log(WARNING, allSourcesUnreachableMsg, totalNodes);
         }
 
         final List<LongRange> mergedRanges = LongRange.mergeContiguousRanges(peerRanges);
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
@@ -145,7 +145,7 @@ public class BackfillFetcher implements PriorityHealthBasedStrategy.NodeHealthPr
             } catch (RuntimeException e) {
                 final String failedToGetStatusMsg = "Failed to get status from node [%s:%d]: %s"
                         .formatted(node.address(), node.port(), e.getMessage());
-                LOGGER.log(INFO, failedToGetStatusMsg, e);
+                LOGGER.log(DEBUG, failedToGetStatusMsg, e);
                 markFailure(node);
             }
         }
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
@@ -147,7 +147,7 @@ public class BackfillFetcher implements PriorityHealthBasedStrategy.NodeHealthPr
             } catch (RuntimeException e) {
                 final String failedToGetStatusMsg = "Failed to get status from node [%s:%d]: %s"
                         .formatted(node.address(), node.port(), e.getMessage());
-                LOGGER.log(INFO, failedToGetStatusMsg, e);
+                LOGGER.log(DEBUG, failedToGetStatusMsg, e);
                 markFailure(node);
             }
         }
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
@@ -122,6 +122,8 @@ public class BackfillFetcher implements PriorityHealthBasedStrategy.NodeHealthPr
      */
     public List<LongRange> getNewAvailableRanges(long latestStoredBlockNumber) {
         final List<LongRange> peerRanges = new ArrayList<>();
+        final int totalNodes = blockNodeSource.nodes().size();
+        int reachedNodeCount = 0;
 
         for (BlockNodeSourceConfig node : blockNodeSource.nodes()) {
             if (isInBackoff(node)) {
@@ -142,12 +144,21 @@ public class BackfillFetcher implements PriorityHealthBasedStrategy.NodeHealthPr
                         peerRanges.add(new LongRange(start, range.end()));
                     }
                 }
+                reachedNodeCount++;
             } catch (RuntimeException e) {
                 final String failedToGetStatusMsg = "Failed to get status from node [%s:%d]: %s"
                         .formatted(node.address(), node.port(), e.getMessage());
                 LOGGER.log(DEBUG, failedToGetStatusMsg, e);
                 markFailure(node);
             }
+        }
+
+        // Individual unreachable/failed nodes are DEBUG, but no reachable source at all means backfill
+        // cannot progress: surface that aggregate condition once at WARNING for on-call visibility.
+        if (totalNodes > 0 && reachedNodeCount == 0) {
+            final String allSourcesUnreachableMsg =
+                    "All [{0}] backfill source(s) are unreachable; cannot discover peer ranges";
+            LOGGER.log(WARNING, allSourcesUnreachableMsg, totalNodes);
         }
 
         final List<LongRange> mergedRanges = LongRange.mergeContiguousRanges(peerRanges);
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
@@ -3,7 +3,6 @@ package org.hiero.block.node.backfill;
 
 import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.TRACE;
-import static java.lang.System.Logger.Level.WARNING;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.concurrent.ConcurrentHashMap;
@@ -160,8 +159,11 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
                 final String receivedConfirmationMsg = "Received persistence confirmation for block [{0}]";
                 LOGGER.log(TRACE, receivedConfirmationMsg, blockNumber);
             } else {
+                // The persistence failure originates in (and is logged by) the persistence layer; this is
+                // only the backfill awaiter observing the notification, so keep it at DEBUG to avoid a
+                // duplicate on-call signal for an event already reported at its source.
                 final String persistenceFailedMsg = "Block [{0}] persistence failed";
-                LOGGER.log(WARNING, persistenceFailedMsg, blockNumber);
+                LOGGER.log(DEBUG, persistenceFailedMsg, blockNumber);
             }
             pending.complete(notification.succeeded());
         }
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
@@ -2,8 +2,8 @@
 package org.hiero.block.node.backfill;
 
 import static java.lang.System.Logger.Level.DEBUG;
-import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.TRACE;
+import static java.lang.System.Logger.Level.WARNING;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.concurrent.ConcurrentHashMap;
@@ -110,7 +110,7 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
                 LOGGER.log(TRACE, receivedConfirmationMsg, blockNumber);
             } else {
                 final String persistenceFailedMsg = "Block [{0}] persistence failed";
-                LOGGER.log(INFO, persistenceFailedMsg, blockNumber);
+                LOGGER.log(WARNING, persistenceFailedMsg, blockNumber);
             }
             latch.countDown();
         }
@@ -136,7 +136,7 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
             CountDownLatch latch = pendingBlocks.get(blockNumber);
             if (latch != null) {
                 final String verificationFailedMsg = "Block [{0}] verification failed, releasing latch";
-                LOGGER.log(INFO, verificationFailedMsg, blockNumber);
+                LOGGER.log(WARNING, verificationFailedMsg, blockNumber);
                 latch.countDown();
             }
         }
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
@@ -2,8 +2,8 @@
 package org.hiero.block.node.backfill;
 
 import static java.lang.System.Logger.Level.DEBUG;
-import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.TRACE;
+import static java.lang.System.Logger.Level.WARNING;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.concurrent.ConcurrentHashMap;
@@ -161,7 +161,7 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
                 LOGGER.log(TRACE, receivedConfirmationMsg, blockNumber);
             } else {
                 final String persistenceFailedMsg = "Block [{0}] persistence failed";
-                LOGGER.log(INFO, persistenceFailedMsg, blockNumber);
+                LOGGER.log(WARNING, persistenceFailedMsg, blockNumber);
             }
             pending.complete(notification.succeeded());
         }
@@ -187,8 +187,9 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
             long blockNumber = notification.blockNumber();
             Pending pending = pendingBlocks.get(blockNumber);
             if (pending != null) {
+                // Echo of a failure already logged at its origin; keep at DEBUG to avoid a duplicate signal.
                 final String verificationFailedMsg = "Block [{0}] verification failed, marking block as not persisted";
-                LOGGER.log(INFO, verificationFailedMsg, blockNumber);
+                LOGGER.log(DEBUG, verificationFailedMsg, blockNumber);
                 pending.complete(false);
             }
         }
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -277,6 +277,8 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             GapProcessor gapProcessor = new GapProcessor(runner, schedulerName);
             return new BackfillTaskScheduler(executor, gapProcessor, queueCapacity, fetcher, persistenceAwaiter);
         } catch (RuntimeException e) {
+            // TODO: a scheduler-creation failure leaves the plugin degraded; mark it unhealthy via the
+            // health facility once that is available.
             final String createSchedulerFailedMsg = "Failed to create scheduler: [%s]".formatted(e.getMessage());
             LOGGER.log(WARNING, createSchedulerFailedMsg, e);
             return null;
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -276,7 +276,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             return new BackfillTaskScheduler(executor, gapProcessor, queueCapacity, fetcher, persistenceAwaiter);
         } catch (RuntimeException e) {
             final String createSchedulerFailedMsg = "Failed to create scheduler: [%s]".formatted(e.getMessage());
-            LOGGER.log(ERROR, createSchedulerFailedMsg, e);
+            LOGGER.log(WARNING, createSchedulerFailedMsg, e);
             return null;
         }
     }
@@ -294,14 +294,14 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             try {
                 historicalScheduler.close();
             } catch (RuntimeException e) {
-                LOGGER.log(WARNING, "Error closing historicalScheduler", e);
+                LOGGER.log(DEBUG, "Error closing historicalScheduler", e);
             }
         }
         if (liveTailScheduler != null) {
             try {
                 liveTailScheduler.close();
             } catch (RuntimeException e) {
-                LOGGER.log(WARNING, "Error closing liveTailScheduler", e);
+                LOGGER.log(DEBUG, "Error closing liveTailScheduler", e);
             }
         }
 
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -2,6 +2,7 @@
 package org.hiero.block.node.backfill;
 
 import static java.lang.System.Logger.Level.DEBUG;
+import static java.lang.System.Logger.Level.ERROR;
 import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.TRACE;
 import static java.lang.System.Logger.Level.WARNING;
@@ -220,7 +221,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
         // Create platform thread executors via threadPoolManager.
         // Platform threads required because Helidon WebClient HTTP/2 has issues with virtual threads in containers.
         Thread.UncaughtExceptionHandler handler =
-                (thread, e) -> LOGGER.log(INFO, "Uncaught exception in thread: " + thread.getName(), e);
+                (thread, e) -> LOGGER.log(ERROR, "Uncaught exception in thread: " + thread.getName(), e);
 
         autonomousExecutor =
                 context.threadPoolManager().createSingleThreadScheduledExecutor("BackfillPluginRunner", handler);
@@ -278,7 +279,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             return new BackfillTaskScheduler(executor, gapProcessor, queueCapacity, fetcher, persistenceAwaiter);
         } catch (RuntimeException e) {
             final String createSchedulerFailedMsg = "Failed to create scheduler: [%s]".formatted(e.getMessage());
-            LOGGER.log(INFO, createSchedulerFailedMsg, e);
+            LOGGER.log(ERROR, createSchedulerFailedMsg, e);
             return null;
         }
     }
@@ -296,14 +297,14 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             try {
                 historicalScheduler.close();
             } catch (RuntimeException e) {
-                LOGGER.log(INFO, "Error closing historicalScheduler: " + e.getMessage(), e);
+                LOGGER.log(WARNING, "Error closing historicalScheduler", e);
             }
         }
         if (liveTailScheduler != null) {
             try {
                 liveTailScheduler.close();
             } catch (RuntimeException e) {
-                LOGGER.log(INFO, "Error closing liveTailScheduler: " + e.getMessage(), e);
+                LOGGER.log(WARNING, "Error closing liveTailScheduler", e);
             }
         }
         if (autonomousFetcher != null) {
@@ -548,7 +549,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
                                 new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL))) {
                     final String backfillPersistFailedMsg =
                             "Backfill persistence failed for block=[{0}], re-queuing for on-demand backfill";
-                    LOGGER.log(INFO, backfillPersistFailedMsg, blockNumber);
+                    LOGGER.log(WARNING, backfillPersistFailedMsg, blockNumber);
                 } else {
                     // The live-tail scheduler is unavailable (plugin not yet started) or its queue is full -
                     // most likely because a mass failure (e.g. MISSING_VERIFICATION_DATA while the node is
@@ -580,7 +581,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             LOGGER.log(TRACE, verificationNotificationMsg, notification.blockNumber());
             if (!notification.success()) {
                 final String blockVerificationFailedMsg = "Block verification failed, block=[{0}]";
-                LOGGER.log(INFO, blockVerificationFailedMsg, notification.blockNumber());
+                LOGGER.log(WARNING, blockVerificationFailedMsg, notification.blockNumber());
                 metricsHolder.backfillFetchErrors().increment();
                 pendingBackfillBlocks.updateAndGet(v -> Math.max(0, v - 1));
                 // If a block verification fails, we will backfill it again later on the next gap detection run.
@@ -640,7 +641,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             } catch (ParseException | InterruptedException e) {
                 Thread.currentThread().interrupt();
                 final String errorExecutingGapMsg = "Error executing gap=[%s]".formatted(gap);
-                LOGGER.log(INFO, errorExecutingGapMsg, e);
+                LOGGER.log(WARNING, errorExecutingGapMsg, e);
             }
         }
     }
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -279,7 +279,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             return new BackfillTaskScheduler(executor, gapProcessor, queueCapacity, fetcher, persistenceAwaiter);
         } catch (RuntimeException e) {
             final String createSchedulerFailedMsg = "Failed to create scheduler: [%s]".formatted(e.getMessage());
-            LOGGER.log(ERROR, createSchedulerFailedMsg, e);
+            LOGGER.log(WARNING, createSchedulerFailedMsg, e);
             return null;
         }
     }
@@ -297,14 +297,14 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
             try {
                 historicalScheduler.close();
             } catch (RuntimeException e) {
-                LOGGER.log(WARNING, "Error closing historicalScheduler", e);
+                LOGGER.log(DEBUG, "Error closing historicalScheduler", e);
             }
         }
         if (liveTailScheduler != null) {
             try {
                 liveTailScheduler.close();
             } catch (RuntimeException e) {
-                LOGGER.log(WARNING, "Error closing liveTailScheduler", e);
+                LOGGER.log(DEBUG, "Error closing liveTailScheduler", e);
             }
         }
         if (autonomousFetcher != null) {
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -158,7 +158,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
         // Validate block node sources configuration
         final String sourcesPath = backfillConfiguration.blockNodeSourcesPath();
         if (sourcesPath == null || sourcesPath.isBlank()) {
-            LOGGER.log(DEBUG, "No block node sources path configured, backfill will not run");
+            LOGGER.log(INFO, "No block node sources path configured, backfill will not run");
             return;
         }
 
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -155,7 +155,7 @@ public class BackfillPlugin implements BlockNodePlugin, BlockNotificationHandler
         // Validate block node sources configuration
         final String sourcesPath = backfillConfiguration.blockNodeSourcesPath();
         if (sourcesPath == null || sourcesPath.isBlank()) {
-            LOGGER.log(DEBUG, "No block node sources path configured, backfill will not run");
+            LOGGER.log(INFO, "No block node sources path configured, backfill will not run");
             return;
         }
 
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
@@ -2,6 +2,7 @@
 package org.hiero.block.node.backfill;
 
 import static java.lang.System.Logger.Level.DEBUG;
+import static java.lang.System.Logger.Level.ERROR;
 import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.TRACE;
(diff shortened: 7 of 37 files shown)


block-node/app/docker/ci-logging.properties
```diff
@@ -13,8 +13,8 @@
 # Available Levels are (from most verbose to least verbose):
 # ALL FINEST FINER FINE CONFIG INFO WARNING SEVERE OFF
 .level=INFO
-# logs for Block Node components FINE (DEBUG) temporarily
-org.hiero.block.level=FINE
```
ata-nas, line 17: Do we want to change the CI properties? Those are usually used for running tests.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -275,7 +276,7 @@ private BackfillTaskScheduler createScheduler(ExecutorService executor, int queu
             return new BackfillTaskScheduler(executor, gapProcessor, queueCapacity, fetcher, persistenceAwaiter);
         } catch (RuntimeException e) {
             final String createSchedulerFailedMsg = "Failed to create scheduler: [%s]".formatted(e.getMessage());
-            LOGGER.log(INFO, createSchedulerFailedMsg, e);
+            LOGGER.log(ERROR, createSchedulerFailedMsg, e);
```
ata-nas: Error means something critical happened and we are shutting down. Is that true here?

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -293,14 +294,14 @@ public void stop() {
             try {
                 historicalScheduler.close();
             } catch (RuntimeException e) {
-                LOGGER.log(INFO, "Error closing historicalScheduler: " + e.getMessage(), e);
+                LOGGER.log(WARNING, "Error closing historicalScheduler", e);
```
ata-nas: Is that really a WARN? We are shutting down and we failed to shutdown properly.

ata-nas (commented):
Will get back to this to finish my review, but just a couple of notes  for now.

In general, I am not sure we are complying with the logging guidelines in some places (another review to come).

block-node/app/docker/ci-logging.properties
Nana-EC: Agreed, i don't think we should change CI. This would mean a developer is unable to see the details needed to deeply troubleshoot. I hit an issue like this recently and I had to change the log to FINE+ to unearth the deeper issue. That works sometimes but with transient or flaky issues we lose the opportunity to debug.
CI should be left at FINE at least.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
```diff
@@ -145,7 +145,7 @@ public List<LongRange> getNewAvailableRanges(long latestStoredBlockNumber) {
             } catch (RuntimeException e) {
                 final String failedToGetStatusMsg = "Failed to get status from node [%s:%d]: %s"
                         .formatted(node.address(), node.port(), e.getMessage());
-                LOGGER.log(INFO, failedToGetStatusMsg, e);
+                LOGGER.log(DEBUG, failedToGetStatusMsg, e);
```
Nana-EC, line 153: Individually this makes sense. But we should maybe have a WARNING log for when all backfill sources are unreachable.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -168,6 +176,63 @@ public void init(@NonNull final BlockNodeContext context, @NonNull final Service
...
+        }
+    }
+
+    /// Runs one heartbeat, guarding against exceptions so a single failure does not cancel the
+    /// fixed-rate schedule.
+    private void runHeartbeat() {
+        try {
+            logStatusHeartbeat();
+        } catch (final RuntimeException e) {
+            LOGGER.log(DEBUG, "Failed to emit server status heartbeat", e);
```
Nana-EC: Should this be warning?
If not operators have no indication whether their node stopped progressing or if the heartbeat itself is broken.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -168,6 +176,63 @@ public void init(@NonNull final BlockNodeContext context, @NonNull final Service
...
+                .threadPoolManager()
+                .createSingleThreadScheduledExecutor("server-status-heartbeat", this::onHeartbeatThreadException);
+        heartbeatExecutor.scheduleAtFixedRate(this::runHeartbeat, periodSeconds, periodSeconds, TimeUnit.SECONDS);
+        LOGGER.log(DEBUG, "Server status heartbeat scheduled every {0}s", periodSeconds);
+    }
+
+    @Override
+    public void stop() {
+        if (heartbeatExecutor != null) {
+            heartbeatExecutor.shutdownNow();
```
Nana-EC, line 224: Q: Should we not awaitTermination before doing a `shutdownNow()`?
Otherwise the heartbeat may still be executing `logStatusHeartbeat()`

Nana-EC (commented):
Nice work, this will help a lot.
Some suggestions

block-node/app/docker/ci-logging.properties
AlfredoG87: Reverted ✅ 

CI stays at `org.hiero.block.level=FINE`. Only the production files (`docker/logging.properties`, `app/src/main/resources/logging.properties`) move to INFO. Comment updated to say CI keeps FINE so flaky/transient test failures stay debuggable

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Agreed, ERROR was too strong. Changed to WARNING. ✅

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Agreed. updated to DEBUG ✅

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
AlfredoG87: Added. Individual unreachable/failed nodes stay at DEBUG; when a discovery cycle reaches zero sources we now emit one aggregate WARNING ("All [N] backfill source(s) are unreachable; cannot discover peer ranges").

Covered by a new BackfillFetcherTest

✅ Done!

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
AlfredoG87: ✅ Done, good catch agreed. Made it a WARN

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
AlfredoG87: ✅ Agreed and updated

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
AlfredoG87: Changed stop() to `shutdown()` + `awaitTermination(5s)`, falling back to `shutdownNow()` on timeout/interrupt. An in-flight logStatusHeartbeat() now completes instead of being interrupted mid-emit.

✅ Correct and Done!

Nana-EC (dismissed):

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -168,7 +168,7 @@ public BlockNodeApp(final ServiceLoaderFunction serviceLoader, final boolean sho
                     LOGGER.log(INFO, "No logging configuration found");
                 }
             } catch (IOException e) {
-                LOGGER.log(INFO, "Failed to load logging configuration", e);
+                LOGGER.log(WARNING, "Failed to load logging configuration", e);
```
ata-nas: Why would that be a WARNING? It seems like we have failed to load logging config, but is that destructive?

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -122,7 +121,7 @@ static NetworkData loadNetworkData(final Path path) {
                     NetworkData.JSON, Bytes.wrap(Files.readAllBytes(path)), MAX_APP_STATE_MESSAGE_SIZE_BYTES);
             return data;
         } catch (ParseException | IOException e) {
-            LOGGER.log(INFO, "Failed to read network data file %s.".formatted(path), e);
+            LOGGER.log(WARNING, "Failed to read network data file %s.".formatted(path), e);
```
ata-nas: Is this destructive to the system? Is it recoverable? If so, why changed to WARNING?

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
```diff
@@ -164,7 +163,7 @@ static void validateAddressBook(final NodeAddressBook book, final String source)
                 kf.generatePublic(new X509EncodedKeySpec(keyBytes));
                 usable++;
             } catch (InvalidKeySpecException | IllegalArgumentException e) {
-                LOGGER.log(INFO, "Malformed RSA_PubKey for node {0} — skipped: {1}", addr.nodeId(), e);
+                LOGGER.log(WARNING, "Malformed RSA_PubKey for node %s — skipped".formatted(addr.nodeId()), e);
```
ata-nas: Is this destructive to the system? Is it recoverable? If so, why changed to WARNING?

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
```diff
@@ -142,14 +144,23 @@ public List<LongRange> getNewAvailableRanges(long latestStoredBlockNumber) {
...
                 markFailure(node);
             }
         }
 
+        // Individual unreachable/failed nodes are DEBUG, but no reachable source at all means backfill
+        // cannot progress: surface that aggregate condition once at WARNING for on-call visibility.
+        if (totalNodes > 0 && reachedNodeCount == 0) {
+            final String allSourcesUnreachableMsg =
+                    "All [{0}] backfill source(s) are unreachable; cannot discover peer ranges";
+            LOGGER.log(WARNING, allSourcesUnreachableMsg, totalNodes);
```
ata-nas, line 163: Is this destructive to the system? Is it recoverable? If so, why changed to WARNING?

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
```diff
@@ -110,7 +110,7 @@ public void handlePersisted(@NonNull PersistedNotification notification) {
                 LOGGER.log(TRACE, receivedConfirmationMsg, blockNumber);
             } else {
                 final String persistenceFailedMsg = "Block [{0}] persistence failed";
-                LOGGER.log(INFO, persistenceFailedMsg, blockNumber);
+                LOGGER.log(WARNING, persistenceFailedMsg, blockNumber);
```
ata-nas: Why are we logging that persistence failed in the backfill plugin? I feel like this is redundant and we can remove. We already know we have a failed persistence, we just receive and log a notification here.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
```diff
@@ -136,7 +136,7 @@ public void handleVerification(@NonNull VerificationNotification notification) {
             CountDownLatch latch = pendingBlocks.get(blockNumber);
             if (latch != null) {
                 final String verificationFailedMsg = "Block [{0}] verification failed, releasing latch";
-                LOGGER.log(INFO, verificationFailedMsg, blockNumber);
+                LOGGER.log(WARNING, verificationFailedMsg, blockNumber);
```
ata-nas: Same here, why do we log an event that did not originate from us? If verification failed, we know about it, I think we can remove this log.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -217,7 +218,7 @@ public void start() {
         // Create platform thread executors via threadPoolManager.
         // Platform threads required because Helidon WebClient HTTP/2 has issues with virtual threads in containers.
         Thread.UncaughtExceptionHandler handler =
-                (thread, e) -> LOGGER.log(INFO, "Uncaught exception in thread: " + thread.getName(), e);
+                (thread, e) -> LOGGER.log(ERROR, "Uncaught exception in thread: " + thread.getName(), e);
```
ata-nas: ERROR is catastrophic, are we witnessing a catastrophic event if an async task here fails? I do not think so. INFO seems appropriate as this seems to be something recoverable.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -538,15 +539,15 @@ && submitGap(new GapDetector.Gap(
                                 new LongRange(blockNumber, blockNumber), GapDetector.Type.LIVE_TAIL))) {
                     final String backfillPersistFailedMsg =
                             "Backfill persistence failed for block=[{0}], re-queuing for on-demand backfill";
-                    LOGGER.log(INFO, backfillPersistFailedMsg, blockNumber);
+                    LOGGER.log(WARNING, backfillPersistFailedMsg, blockNumber);
```
ata-nas: Why WARNING? If backfill failed, I think this is a very easily recoverable event, we just try again.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -565,7 +566,7 @@ public void handleVerification(VerificationNotification notification) {
             LOGGER.log(TRACE, verificationNotificationMsg, notification.blockNumber());
             if (!notification.success()) {
                 final String blockVerificationFailedMsg = "Block verification failed, block=[{0}]";
-                LOGGER.log(INFO, blockVerificationFailedMsg, notification.blockNumber());
+                LOGGER.log(WARNING, blockVerificationFailedMsg, notification.blockNumber());
```
ata-nas: Why are we logging a failed verification here? If verification failed, we already know about it. I feel like we should completely remove this log, but also similar ones we have.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -625,7 +626,7 @@ public void accept(GapDetector.Gap gap) {
             } catch (ParseException | InterruptedException e) {
```
ata-nas: Couldn't help but notice, we set the interrupted flag when we hit a parse exception?! This is way off, if this is a task that would run async, this would confuse the future into thinking the task was cancelled instead of failing exceptionally.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -625,7 +626,7 @@ public void accept(GapDetector.Gap gap) {
...
                 Thread.currentThread().interrupt();
                 final String errorExecutingGapMsg = "Error executing gap=[%s]".formatted(gap);
-                LOGGER.log(INFO, errorExecutingGapMsg, e);
+                LOGGER.log(WARNING, errorExecutingGapMsg, e);
```
ata-nas: Why WARNING? If we failed to execute this gap, don't we just try again? This seems like a recoverable event.

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillFetcherTest.java
```diff
@@ -269,6 +271,27 @@ void returnsNullWhenAheadOfPeers() throws Exception {
...
+            julLogger.addHandler(logHandler);
+            try {
+                assertTrue(fetcher.getNewAvailableRanges(0L).isEmpty());
+            } finally {
+                julLogger.removeHandler(logHandler);
+            }
+
+            assertEquals(1, logHandler.countContaining("backfill source(s) are unreachable"));
+        }
+
```
ata-nas, line 294: Personal opinion, I am against logging tests, we assert that we log a certain level based on some event? This is very fragile and will fail all tests if we want to change a log leve.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/CompletableVerificationSession.java
```diff
@@ -221,7 +222,7 @@ public boolean complete() {
             final String message =
                     "Exception occurred during completion of verification session with id %d for block %d with source %s"
                             .formatted(sessionKey.uniqueId(), blockNumber, blockSource);
-            LOGGER.log(INFO, message, e);
+            LOGGER.log(WARNING, message, e);
```
ata-nas: This is a recoverable event, I do not understand why WARNING here. There is nothing for operator to do.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -209,7 +209,7 @@ public void handleBlockItemsReceived(final BlockItems blockItems) {
                 LOGGER.log(INFO, "Received null block items on live items ring buffer");
             }
         } catch (final RuntimeException e) {
-            LOGGER.log(INFO, "Failed to handle live block items in verification ", e);
+            LOGGER.log(WARNING, "Failed to handle live block items in verification ", e);
```
ata-nas: I disagree with WARNING here, this is a recoverable event and not destructive to the system.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
```diff
@@ -246,7 +246,7 @@ public void handleBackfilled(final BackfilledBlockNotification notification) {
                 LOGGER.log(INFO, "Received invalid backfill notification: {0}", notification);
             }
         } catch (final RuntimeException e) {
-            LOGGER.log(INFO, "Failed to handle backfill notification in verification ", e);
+            LOGGER.log(WARNING, "Failed to handle backfill notification in verification ", e);
```
ata-nas: I disagree with WARNING here, this is a recoverable event and not destructive to the system.

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
ata-nas, line 1: WARNING are probably acceptable here, as operator could take an action, but not sure if we actually want to do WARNING for most of these. In the cases where it is detrimental to the system, we can log a WARNING.

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessor.java
```diff
@@ -165,7 +164,7 @@ public void close() {
         } catch (final RuntimeException | IOException e) {
             final String message = FAILED_TO_DELETE_LINK_MESSAGE.formatted(
                     blockNumber, absoluteZipFilePath, blockPathData.blockFileName());
-            LOGGER.log(INFO, message, e);
+            LOGGER.log(WARNING, message, e);
```
ata-nas: This is not detrimental to the system, but an operator can take an action here. Not sure if switch to WARNING is necessary.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
```diff
@@ -175,7 +174,7 @@ public void close() {
         } catch (final RuntimeException | IOException e) {
             final String message = "Failed to delete accessor link for block: %d, path: %s"
                     .formatted(blockNumber, absolutePathToBlock);
-            LOGGER.log(INFO, message, e);
+            LOGGER.log(WARNING, message, e);
```
ata-nas: Not detrimental, an operator can take some action, not sure if switch is necessary.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -185,7 +185,7 @@ public void init(final BlockNodeContext context, final ServiceBuilder serviceBui
                     totalBytesStored.addAndGet(Files.size(blockFilePath));
                 }
             } catch (IOException e) {
-                LOGGER.log(INFO, "Failed to get size of block file for block " + blockNumber, e);
+                LOGGER.log(WARNING, "Failed to get size of block file for block %s".formatted(blockNumber), e);
```
ata-nas: I think this is purely informational. Not getting the size means we do not have much accuracy in metrics.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
```diff
@@ -484,7 +484,7 @@ private void deleteBlockCleanly(final long blockNumber, final Path blockFilePath
             blocksDeletedCounter.increment();
             totalBytesStored.addAndGet(-fileSize);
         } catch (final IOException e) {
-            LOGGER.log(INFO, DELETE_MESSAGE.formatted("Failure", blockFilePath), e);
+            LOGGER.log(WARNING, DELETE_MESSAGE.formatted("Failure", blockFilePath), e);
```
ata-nas: Not destructive if we cannot delete, not opposed to this change, maybe we can think about this.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
```diff
@@ -267,7 +267,8 @@ private byte[] flushPartIfNeeded(
                     ? blockSource
                     : blocksInBuffer.firstEntry().getValue();
             blockMessaging.sendBlockPersisted(new PersistedNotification(firstBlockNum, false, 1_000, firstBlockSource));
-            LOGGER.log(INFO, "Failed to upload part containing blocks %d to %d".formatted(firstBlockNum, blockNum), e);
+            LOGGER.log(
+                    WARNING, "Failed to upload part containing blocks %d to %d".formatted(firstBlockNum, blockNum), e);
```
ata-nas: Is this not recoverable? If so, this feels like a purely informational log, especially considering that cloud providers will fail to upload for any reason very often compared to local storage.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
```diff
@@ -298,7 +299,7 @@ private UploadResult uploadFinalPart(
             final Map.Entry<Long, BlockSource> first = blocksInBuffer.firstEntry();
             blockMessaging.sendBlockPersisted(
                     new PersistedNotification(first.getKey(), false, 1_000, first.getValue()));
-            LOGGER.log(INFO, "Failed to upload final part for key {0}", key, e);
+            LOGGER.log(WARNING, "Failed to upload final part for key %s".formatted(key), e);
```
ata-nas: Is this not recoverable? If so, this feels like a purely informational log, especially considering that cloud providers will fail to upload for any reason very often compared to local storage.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ConsolidationTask.java
```diff
@@ -76,7 +76,7 @@ public UploadResult call() throws S3ClientInitializationException, S3ResponseExc
                 }
                 doCompleteMultipartUpload(s3, finalKey, uploadId, etags);
             } catch (S3ResponseException | IOException e) {
-                LOGGER.log(INFO, "Consolidation failed for key {0}", finalKey, e);
+                LOGGER.log(WARNING, "Consolidation failed for key %s".formatted(finalKey), e);
```
ata-nas: Is this not recoverable? If so, this feels like a purely informational log.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
```diff
@@ -132,7 +134,10 @@ public TempArchiveEntry call()
                     }
                     throw e;
                 } catch (S3ResponseException | IOException e) {
-                    LOGGER.log(INFO, "Failed to accumulate block {0} for temp archive {1}", currentBlock, s3Key, e);
+                    LOGGER.log(
+                            WARNING,
+                            "Failed to accumulate block %s for temp archive %s".formatted(currentBlock, s3Key),
+                            e);
```
ata-nas: Is this not recoverable? If so, this feels like a purely informational log.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
```diff
@@ -164,7 +169,7 @@ private TempArchiveEntry completeSegment(
             } catch (S3ResponseException | IOException e) {
                 S3UploadUtils.abortQuietly(s3, s3Key, uploadId);
                 blockMessaging.sendBlockPersisted(new PersistedNotification(firstBlock, false, 1_000, lastSource));
-                LOGGER.log(INFO, "Failed to upload final temp archive part for key {0}", s3Key, e);
+                LOGGER.log(WARNING, "Failed to upload final temp archive part for key %s".formatted(s3Key), e);
```
ata-nas: Is this not recoverable? If so, this feels like a purely informational log. Cloud is expected to fail quite often opposed to local storage.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
```diff
@@ -174,7 +179,7 @@ private TempArchiveEntry completeSegment(
         } catch (S3ResponseException | IOException e) {
             S3UploadUtils.abortQuietly(s3, s3Key, uploadId);
             blockMessaging.sendBlockPersisted(new PersistedNotification(firstBlock, false, 1_000, lastSource));
-            LOGGER.log(INFO, "Failed to complete temp archive multipart upload for key {0}", s3Key, e);
+            LOGGER.log(WARNING, "Failed to complete temp archive multipart upload for key %s".formatted(s3Key), e);
```
ata-nas: Is this not recoverable? If so, this feels like a purely informational log. Cloud is expected to fail quite often opposed to local storage.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -168,6 +177,74 @@ public void init(@NonNull final BlockNodeContext context, @NonNull final Service
...
+        }
+    }
+
+    /// Emits the single periodic `INFO` status line: the available block range and next expected
+    /// block. This is the one progression signal available to operators at `INFO` without
+    /// enabling `DEBUG`; per-block progress is intentionally left to metrics.
+    void logStatusHeartbeat() {
+        if (!LOGGER.isLoggable(INFO)) {
+            return;
+        }
```
ata-nas, line 245: Do we want a heartbeat at INFO?

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -168,6 +177,74 @@ public void init(@NonNull final BlockNodeContext context, @NonNull final Service
...
+                INFO,
+                "Status heartbeat: oldestBlock={0} newestBlock={1} nextExpected={2}",
+                availableBlocks.min(),
+                availableBlocks.max(),
+                nextExpectedBlock);
+    }
+
+    /// Logs an uncaught exception escaping the heartbeat thread at `ERROR`.
+    private void onHeartbeatThreadException(final Thread thread, final Throwable throwable) {
+        LOGGER.log(ERROR, "Uncaught exception in server-status heartbeat thread", throwable);
```
ata-nas: ERROR is catastrophic, are we shutting down the system unrecoverably if this exception handler is used? If not, maybe a WARNING?

block-node/server-status/src/main/java/module-info.java
```diff
@@ -17,6 +17,7 @@
     requires org.hiero.block.node.app.config;
     requires org.hiero.block.node.base;
     requires com.github.spotbugs.annotations;
+    requires java.logging;
```
ata-nas, line 20: This might be added by mistake?

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
```diff
@@ -249,6 +251,71 @@ void shouldLeaveLastAvailableBlockUnknownWhenNoBlocksAvailable() throws ParseExc
...
+        final Logger julLogger = Logger.getLogger(ServerStatusServicePlugin.class.getName());
+        julLogger.addHandler(logHandler);
+        try {
+            // start() (run during setup) scheduled the heartbeat on the test executor; fire it now.
+            ((ScheduledBlockingExecutor) testThreadPoolManager.scheduledExecutor()).executeSerially();
+        } finally {
+            julLogger.removeHandler(logHandler);
+        }
+        assertTrue(logHandler.countContaining("Status heartbeat") >= 1);
+    }
```
ata-nas, line 287: Again, I have a negative opinion on such tests, what are we testing? That for log x we have INFO level and for log y we have WARNING? This, I believe, adds code we have to maintain, will break on the smallest change and adds no important coverage.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -818,7 +818,7 @@ else if (localFuture.isDone() && !localFuture.isCancelled()) {
                     LOGGER.log(
                             TRACE, "Interrupted while waiting for publisher unavailability timeout task to complete");
                 } catch (final ExecutionException e) {
-                    LOGGER.log(INFO, "Publisher unavailability timeout task completed exceptionally", e);
+                    LOGGER.log(WARNING, "Publisher unavailability timeout task completed exceptionally", e);
```
ata-nas: This could be migrated to WARNING, if we failed to notify that publishers are not available. Not sure, however.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -1038,7 +1038,7 @@ private void checkLogAndRestartForwarderTask() {
                 final String formatString = "Queue forwarder task completed normally after forwarding %d blocks.";
                 LOGGER.log(DEBUG, formatString.formatted(blocksForwarded));
             } catch (CancellationException | ExecutionException e) {
-                LOGGER.log(INFO, "Queue forwarder task completed exceptionally.", e);
+                LOGGER.log(WARNING, "Queue forwarder task completed exceptionally.", e);
```
ata-nas: This is recoverable I think, the next block should be able to spawn a new one. Don't think we want WARNING.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
```diff
@@ -220,7 +220,7 @@ public void onNext(@NonNull final PublishStreamRequestUnparsed request) {
             } catch (final RuntimeException e) {
                 // If we reach here, it means that the handler was interrupted or
                 // an unexpected error occurred. We should log the error and shut down.
-                LOGGER.log(INFO, "[%2$s] Error processing request: %1$s".formatted(e, correlationIdPrefix), e);
+                LOGGER.log(WARNING, "[%2$s] Error processing request: %1$s".formatted(e, correlationIdPrefix), e);
```
ata-nas: Why WARNING? This is not destructive to the system. Something happened, we could not process the request, we are shutting down the stream, not the node or the plugin.

docs/block-node/operations/production-guide/steady-state-operations.md
```diff
@@ -56,12 +56,19 @@ kubectl -n block-node logs $BN --previous      # post-crash container logs
...
+Production runs at `INFO`, which is deliberately low-volume. Block ingest and verification
+progress is tracked by [metrics](../../metrics.md) and the periodic status heartbeat — **not**
+by per-block INFO logs. Use logs to confirm startup/config and to find problems:
+
+|               Grep tag                |                       Indicates                       |
+|---------------------------------------|-------------------------------------------------------|
+| `Status heartbeat`                    | Liveness and block progression (`newestBlock` rising) |
+| `Started BlockNode Server`            | Startup completed                                     |
+| `VerificationServicePlugin` (WARNING) | Signature or proof verification failures              |
+| `SEVERE` / `WARNING`                  | Any failure needing attention                         |
```
ata-nas, line 68: This is correct, ERROR and WARNING should require immediate action, complying with our logging guidelines definition. This contradicts most of the level changes to logs done in this PR, in my opinion.

docs/design/architecture/Logging-Architecture.md
```diff
@@ -0,0 +1,124 @@
...
+Levels use `System.Logger.Level`, mapped to JUL as: `TRACE→FINEST`, `DEBUG→FINE`,
+`INFO→INFO`, `WARNING→WARNING`, `ERROR→SEVERE`. **Production enables INFO and above only.**
+
+|  Level  |                                         Use it for                                         | Frequency  |    Audience    |
+|---------|--------------------------------------------------------------------------------------------|------------|----------------|
+| ERROR   | System entering a failure state; needs immediate attention (data-loss risk, cannot serve). | Rare       | On-call, paged |
+| WARNING | Unhealthy but not failing; recoverable; may signal a defect or attack.                     | Uncommon   | On-call        |
+| INFO    | Low-frequency, operationally-significant state changes an operator needs in production.    | Infrequent | Operators      |
+| DEBUG   | Detail for diagnosing hard problems in test — or, on demand, targeted in prod.             | High       | Developers     |
+| TRACE   | Very detailed, per-item developer/unit-test diagnostics.                                   | Very high  | Developers     |
```
ata-nas, line 68: Do some of these not contradict our logging guidelines we have defined? I think so. If we have decided to move away from those and change our guidelines, I've missed that, otherwise, I have to be sure this is what we want.

.gitignore
```diff
@@ -72,4 +72,6 @@ k6-out/
 # LLM agent files
 .agent/**
 agent/**
+.claude/
```
ata-nas: @Nana-EC noted that soon we may want to use the `.claude` directory for shared files.

ata-nas (commented):
Quite a lot of questions.

It seems we are moving away from the logging guidelines we have created completely. When did we make such a decision? I am not aware if so. If we have not, however, I think we have to reconsider quite a few things.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
AlfredoG87: Changed to INFO; falls back to the bundled default config, not on-call actionable.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
AlfredoG87: Changed to INFO; falls back to `NetworkData.DEFAULT` (empty set), recoverable.

block-node/app/src/main/java/org/hiero/block/node/app/ApplicationStateUtility.java
AlfredoG87: Now DEBUG for the malformed key and INFO for the no-usable-key path; no WARNING here.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillFetcher.java
AlfredoG87: Kept at WARNING intentionally: this is the aggregate "all sources unreachable" signal. Per-node failures stay at DEBUG; only a full cycle with zero reachable sources warns. Backfill blind to every peer is genuinely unhealthy.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
AlfredoG87: Lowered to DEBUG; the persistence failure is already logged at its origin, so no duplicate signal.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
AlfredoG87: Lowered to DEBUG; the verification failure is already logged at its origin, so no duplicate signal.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Set to INFO; not escalated.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Changed to INFO; the block is re-queued and re-detected, recoverable.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Lowered to DEBUG; already logged at origin.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Fixed: split the catch so `interrupt()` only fires for `InterruptedException`; a `ParseException` no longer marks the thread interrupted.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Changed to INFO; part of the interrupt fix above (two catches, both INFO).

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillFetcherTest.java
AlfredoG87: It asserts the message is emitted, not the level (`TestLogHandler` does not expose level), so it will not break on a level change. Renamed accordingly.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/CompletableVerificationSession.java
AlfredoG87: Changed to INFO.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
AlfredoG87: Changed to INFO.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java
AlfredoG87: Changed to INFO.

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/BlockFileHistoricPlugin.java
AlfredoG87: Changed to INFO; recoverable file operations (parse skip, delete/rename/count), not detrimental to the node.

block-node/blocks-file-historic/src/main/java/org/hiero/block/node/blocks/files/historic/ZipBlockAccessor.java
AlfredoG87: Changed to INFO; not detrimental to the node.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileBlockAccessor.java
AlfredoG87: Changed to INFO; recoverable and informational.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
AlfredoG87: Changed to INFO; informational.

block-node/blocks-file-recent/src/main/java/org/hiero/block/node/blocks/files/recent/BlockFileRecentPlugin.java
AlfredoG87: Changed to INFO; recoverable.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
AlfredoG87: Now logged at INFO; no WARNING remains.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
AlfredoG87: Now logged at INFO; no WARNING remains.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ConsolidationTask.java
AlfredoG87: Changed to INFO.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
AlfredoG87: Now logged at INFO; no WARNING remains.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
AlfredoG87: Now logged at INFO; no WARNING remains.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/TempArchiveUploadTask.java
AlfredoG87: Now logged at INFO; no WARNING remains.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
AlfredoG87: Kept at INFO by design: the single low-frequency (5 min) progression signal for log-only operators; per-block progress stays in metrics.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
AlfredoG87: Lowered from ERROR to WARNING; a dead heartbeat thread degrades observability but is not a node failure.

block-node/server-status/src/main/java/module-info.java
AlfredoG87: Not a mistake: the tests use `java.util.logging` (TestLogHandler), so the module must `requires java.logging`. Main code uses `System.Logger` only.

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
AlfredoG87: The test asserts the heartbeat message is emitted, not its level (`TestLogHandler` does not expose level), so it will not break on a level change. Renamed accordingly.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
AlfredoG87: Changed to INFO; recoverable, kept out of the on-call band.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
AlfredoG87: Changed to INFO; recoverable, the next block spawns a new task.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java
AlfredoG87: Changed to INFO; recoverable (the stream shuts down, not the node).

docs/block-node/operations/production-guide/steady-state-operations.md
AlfredoG87: Now consistent: after the level changes, WARNING/ERROR again mean "needs attention," matching this row.

docs/design/architecture/Logging-Architecture.md
AlfredoG87: Added an "Open discrepancy" section: by best practice these recovered failures are WARNINGs, but this project scopes WARNING to on-call paging, so they sit at INFO for now. Documented as a deliberate deviation and a follow-up to amend the guidelines.

.gitignore
AlfredoG87: Narrowed to `.claude/settings.local.json`; shared `.claude/` files stay committable.

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillFetcherTest.java
ata-nas: Still, I am against this as a practice.

block-node/server-status/src/test/java/org/hiero/block/node/server/status/ServerStatusServicePluginTest.java
ata-nas: I am still against this practice to assert logs.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
```diff
@@ -278,7 +278,7 @@ private BackfillTaskScheduler createScheduler(ExecutorService executor, int queu
             return new BackfillTaskScheduler(executor, gapProcessor, queueCapacity, fetcher, persistenceAwaiter);
         } catch (RuntimeException e) {
             final String createSchedulerFailedMsg = "Failed to create scheduler: [%s]".formatted(e.getMessage());
-            LOGGER.log(INFO, createSchedulerFailedMsg, e);
+            LOGGER.log(WARNING, createSchedulerFailedMsg, e);
```
ata-nas, line 283: Note, if we cannot create the scheduler this means that the plugin is unhealthy. When we have the health facility in place, we can properly mark as unhealthy.

docs/block-node/operations/production-guide/steady-state-operations.md
```diff
@@ -56,12 +56,19 @@ kubectl -n block-node logs $BN --previous      # post-crash container logs
...
-| `VerificationServicePlugin` | Signature or proof validation failures |
+Production runs at `INFO`, which is deliberately low-volume. Block ingest and verification
+progress is tracked by [metrics](../../metrics.md) and the periodic status heartbeat — **not**
+by per-block INFO logs. Use logs to confirm startup/config and to find problems:
+
+|               Grep tag                |                       Indicates                       |
+|---------------------------------------|-------------------------------------------------------|
+| `Status heartbeat`                    | Liveness and block progression (`newestBlock` rising) |
+| `Started BlockNode Server`            | Startup completed                                     |
+| `VerificationServicePlugin` (WARNING) | Signature or proof verification failures              |
```
ata-nas: I would summarize to verification session failure instead. We can fail on block parsing for various reasons, fail because we do not see mandatory items/data, fail because we are cancelled etc. etc.

```suggestion
| `VerificationServicePlugin` (WARNING) | Verification session failures                                |
```

.gitignore
```diff
@@ -72,4 +72,8 @@ k6-out/
 # LLM agent files
 .agent/**
 agent/**
+# Ignore only local, per-developer Claude settings; shared .claude/ files (settings.json,
+# commands, skills) remain committable.
+.claude/settings.local.json
```
ata-nas: Does this work as well? I am not sure so we have to double check if this works for directories and files alike.
```suggestion
*local*
**/*local*
```
Ignore anything with local in its name (we should include both directories and file names as well).

ata-nas (commented):
Much improved, happy to reach consensus, some final suggestions, but they are not blocking. Also, great job on identifying and fixing exceptional logs that do not print the stacktrace because the appropriate logger method signature was not used.

docs/block-node/operations/production-guide/steady-state-operations.md
AlfredoG87: Applied; the row now reads "Verification session failures".

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java
AlfredoG87: Kept at WARNING; added a TODO to mark the plugin unhealthy via the health facility once it is available.

.gitignore
AlfredoG87: Yes, it works for both: a gitignore pattern with no trailing slash matches files and directories at any depth, so `**/*local*` is redundant. I scoped it to `*.local` / `*.local.*` rather than bare `*local*` so it does not also swallow legitimate names, e.g. bare `*local*` matches the tracked `local-node-block-node-integration.md` doc (and substrings like `locales`). Verified with `git check-ignore`: `settings.local.json` and `.agents.local/` are ignored while that doc and `settings.json` stay tracked.

ata-nas (approved):

Nana-EC (approved):

PR #3313, opened 2026-07-29
chore: solo chaos review followups

### LINKED ISSUE #3312: chore: follow-up on non-blocking review comments from #2798 (test: Add solo-chaos latency considerations to e2e test framework)
## Story Form

As a maintainer
I want all non-blocking review feedback from #2798 to be tracked

## Technical Notes

Follow-up items from unresolved review comments on #2798.

- [ ] **`tools-and-tests/scripts/solo-e2e-test/scripts/chaos-cleanup.sh` line 59 — `kubectl delete --all` and `-l selector` are mutually exclusive** (`berryware`): Passing both `--all` and `${selector_arg[@]}` (which expands to `-l <selector>`) to `kubectl delete` causes a hard error: _"error: setting 'all' parameter but found a non empty selector."_ The `--all` flag must be dropped when a label selector is specified; the delete command needs branching on whether `LABEL_SELECTOR` is set.

- [ ] **`tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh` line 553 — `yq '.bidirectional // true'` always returns `true` for boolean `false`** (`berryware`): The `//` operator in yq treats `false` as a falsy/null value and replaces it with the default `true`. All current scenarios set `bidirectional: true` so the bug is masked, but any future scenario with `bidirectional: false` will silently be treated as bidirectional. Fix: use `yq 'if has("bidirectional") then .bidirectional else true end'` or read the raw string and default explicitly.

- [ ] **`tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh` line 538 — `chaos_resource_name` truncation can produce a trailing `-`** (`berryware`): `chaos_slug` strips leading/trailing dashes from the full string, but `chaos_resource_name` then slices the result to 63 characters with `${n:0:63}`, which can leave a trailing `-` if the combined slug is longer than 63 chars. `kubectl apply` rejects names ending with `-`. Strip trailing dashes after the slice: `echo "${n:0:63}" | sed 's/-*$//'`.

- [ ] **`tools-and-tests/scripts/solo-e2e-test/scripts/test/fixtures/sample-logs-with-backfill.txt` — fixture contains obsolete log lines from the removed `VerificationServicePlugin`** (`jsync-swirlds`): The sample log fixture references `org.hiero.block.node.verification.VerificationServicePlugin`, a class from the plugin we renamed/replaced. The fixture will mislead future test authors about what real log output looks like. Either update to reflect current class names or remove the fixture if it is no longer exercised.

- [ ] **`tools-and-tests/scripts/solo-e2e-test/scripts/test/test-chaos-assertions.sh` line 4 — no Taskfile task to invoke the unit tests** (`berryware`): The bash unit tests in `scripts/test/` can only be run manually (`bash test-chaos-assertions.sh`). Add a Taskfile task (e.g. `test:chaos-scripts`) so CI and contributors can run them without knowing the file path.

- [ ] **`tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh` line 622 — `envsubst` not verified in the `check` task** (`berryware`): `envsubst` is now a required dependency (used in `execute_inject_latency` to render the NetworkChaos manifest template). The `Taskfile.yml` `check` task should assert `command -v envsubst` alongside the other tool checks so users get a clear error early rather than a confusing failure mid-test.

Nana-EC:
## Summary

Addresses all non-blocking review comments captured in #3312 from PR #2798 (solo-chaos latency framework).

## Changes

**Bug fixes (`solo-test-runner.sh`)**
- `yq '.bidirectional // true'` silently coerces boolean `false` to `true` because yq treats falsy values as null. Fixed with `if has("bidirectional") then .bidirectional else true end`.
- `chaos_resource_name` slices the combined slug to 63 chars with `${n:0:63}`, which can leave a trailing `-` that `kubectl apply` rejects. Fixed by stripping trailing dashes after the slice.

**Already fixed at merge:** `chaos-cleanup.sh` `--all` / `-l` mutual exclusion (item 1 in #3312) was addressed in the PR before it merged — the branching logic was already present on main.

**Cleanup (`Taskfile.yml`, test fixture)**
- Added `envsubst` to the `check` prerequisite task — it is required by `execute_inject_latency` to render `network-latency.yaml.tmpl` but was previously unvalidated.
- Added `test:scripts` Taskfile task to run `test-chaos-assertions.sh` without a live cluster (was previously manual-only).
- Updated `sample-logs-with-backfill.txt` fixture: class name was `org.hiero.block.node.verification.VerificationServicePlugin` (old module path); corrected to `org.hiero.block.node.block.verification.VerificationServicePlugin`.

## Test Plan

- [ ] `task test:scripts` runs all 20 bash unit tests with 0 failures
- [ ] Verified `bidirectional: false` in a scenario YAML now correctly sets `direction="to"` instead of `"both"`
- [ ] Verified resource names longer than 63 chars are truncated without a trailing `-`

Closes #3312

### DIFF
--- a/tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
+++ b/tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
@@ -97,6 +97,7 @@ tasks:
         command -v kind >/dev/null 2>&1 || { echo "ERROR: kind not found"; exit 1; }
         command -v solo >/dev/null 2>&1 || { echo "ERROR: solo not found. Install with: npm i @hashgraph/solo -g (or 'task solo:install' for a custom fork/branch build)"; exit 1; }
         command -v yq >/dev/null 2>&1 || { echo "ERROR: yq not found. Install from: https://github.com/mikefarah/yq"; exit 1; }
+        command -v envsubst >/dev/null 2>&1 || { echo "ERROR: envsubst not found. Install gettext: brew install gettext (macOS) or apt-get install gettext (Linux)"; exit 1; }
         echo "All prerequisites found!"
         echo ""
         echo "Versions:"
@@ -536,6 +537,11 @@ tasks:
         ls -la "{{.ROOT_DIR}}/tests/"*.yaml 2>/dev/null || echo "No test files found in tests/"
         echo ""
         echo "Run a test with: task test:run TEST_FILE=tests/<name>.yaml"
+
+  test:scripts:
+    desc: Run bash unit tests for chaos/assertion scripts (no cluster needed)
+    cmds:
+      - 'bash "{{.SCRIPTS_DIR}}/test/test-chaos-assertions.sh"'
 
   # ============================================================================
   # TCK-SDK Tests (matches consensus-node approach: uses latest tags)
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
@@ -538,7 +538,7 @@ function chaos_slug {
 function chaos_resource_name {
     local scenario_name="$1"
     local n="$(chaos_slug "${TEST_NAME}")--$(chaos_slug "${scenario_name}")"
-    echo "${n:0:63}"
+    echo "${n:0:63}" | sed 's/-*$//'
 }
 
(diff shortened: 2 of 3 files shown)


Nana-EC:
Closing, will address in 3387

PR #3314, opened 2026-07-29
fix: replace Files.createLink with atomic Files.move in persistBlockRanges()

### LINKED ISSUE #3315: bug: persistBlockRanges() uses Files.createLink which silently fails on overlay filesystems (Docker/CI)
## Story Form

As a block node operator running in a containerised environment
I want block-ranges.json to be persisted reliably
So that block range data survives restarts and is not silently lost in CI or Docker deployments

## Problem

`BlockNodeApp.persistBlockRanges()` uses `Files.createLink()` to atomically replace `block-ranges.json`.
Hard links are not supported on overlay filesystems (Docker, kind, CI runners) — `Files.createLink()` throws
`UnsupportedOperationException` in those environments.

Because `UnsupportedOperationException` is **not** a subtype of `IOException`, the existing
`catch (IOException e)` block did not catch it. The result:

- The exception propagated unchecked and was silently swallowed by the caller.
- The `.tmp` file was left on disk (`block-ranges.json.tmp` with no corresponding `block-ranges.json`).
- `block-ranges.json` was never written, so subsequent restarts loaded stale or missing block range data.

This manifested in CI as a `java.nio.file.NoSuchFileException` referencing the `.tmp` file and caused
flaky failures in the E2E node-operator workflow:

- **`Assert empty state after reset`** — after a provisioner reset, the BN restarted and loaded
  stale block range data from the retained application-state PVC, reporting `firstAvailableBlock: 0 /
  lastAvailableBlock: 4` even though no block files existed. The assertion against `uint64-max` failed.
- **`api-tests`** — `NoSuchFileException: /opt/hiero/block-node/application-state/block-ranges.json.tmp`
  appeared in logs, indicating the hard-link write was attempted and left the `.tmp` file behind.

## Root cause

```java
// BlockNodeApp.java — persistBlockRanges() (before fix)
Files.write(tmp, json.toByteArray());
Files.deleteIfExists(filePath);
Files.createLink(filePath, tmp);          // throws UnsupportedOperationException on overlay FS
Files.deleteIfExists(tmp);                // never reached; .tmp left on disk
// catch (IOException e) { ... }         // does NOT catch UnsupportedOperationException
```

## Fix (landed in #3314)

Replace the hard-link sequence with the write-to-tmp + `Files.move(ATOMIC_MOVE)` + non-atomic fallback
pattern already used in `persistNodeAddressBookHistory()`:

```java
try {
    Files.move(tmp, filePath, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE);
} catch (AtomicMoveNotSupportedException e) {
    Files.move(tmp, filePath, StandardCopyOption.REPLACE_EXISTING);
}
```

This works on local filesystems (atomic), overlay filesystems (non-atomic fallback), and keeps all
error paths inside `IOException`.

## Flaky tests to track

- [ ] `E2E Node Operator Lifecycle Test / Assert empty state after reset` — was failing because stale
      `block-ranges.json` survived provisioner reset. Fixed in the same pass by: (a) the write fix above
      and (b) a pre-reset `kubectl exec` step to remove `block-ranges.json` before reset, and (c)
      switching the empty-state assertion to `nextExpectedBlock = uint64-max` (the reliable field since 0.39+).
- [ ] Any test that exercises `persistBlockRanges()` on a non-local filesystem should be verified
      green after #3314 merges.

## References

- Fix PR: #3314
- E2E workflow failures: https://github.com/hiero-ledger/hiero-block-node/actions/runs/30420967501/job/90477500046

Nana-EC:
### Problem

`BlockNodeApp.persistBlockRanges()` used `Files.createLink()` to atomically
replace `block-ranges.json`. Hard-linking works on local filesystems but throws
`UnsupportedOperationException` on overlay filesystems (Docker, kind/CI).
Because `UnsupportedOperationException` is **not** a subtype of `IOException`,
the existing `catch (IOException e)` block did not catch it — the exception
propagated unchecked, the `.tmp` file was left on disk, and the write silently
failed in CI with:

```
java.nio.file.NoSuchFileException: /opt/hiero/block-node/application-state/block-ranges.json.tmp
```

The same bug meant `block-ranges.json` was never actually updated after the
failed link, so subsequent restarts loaded stale block range data.

### Fix

Replace the `deleteIfExists + createLink + deleteIfExists` sequence with the
write-to-tmp + `Files.move(ATOMIC_MOVE)` + non-atomic fallback pattern already
used in `persistNodeAddressBookHistory()`. This works on all filesystem types:

- On local filesystems (ext4, APFS) the atomic move succeeds.
- On overlay filesystems (Docker, kind) `AtomicMoveNotSupportedException` is
  caught and a plain `REPLACE_EXISTING` move is used instead.
- Any remaining `IOException` is caught and logged at WARNING, as before.

The log message for the outer catch is also updated from `%s`-style formatting
(which produces a literal `%s` in some JUL implementations) to `{0}:{1}`
parameter substitution consistent with the rest of the class.

Fixes #3315

### DIFF
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -684,11 +684,17 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
             final Path tmp = filePath.resolveSibling(filePath.getFileName() + ".tmp");
             final Bytes json = BlockRangesState.JSON.toBytes(toBlockRangesState());
             Files.write(tmp, json.toByteArray());
-            Files.deleteIfExists(filePath);
-            Files.createLink(filePath, tmp);
-            Files.deleteIfExists(tmp);
+            try {
+                Files.move(tmp, filePath, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE);
+            } catch (AtomicMoveNotSupportedException e) {
+                LOGGER.log(
+                        INFO,
+                        "Atomic move not supported on this filesystem for {0}; falling back to non-atomic replace",
+                        filePath);
+                Files.move(tmp, filePath, StandardCopyOption.REPLACE_EXISTING);
+            }
         } catch (IOException e) {
-            LOGGER.log(WARNING, "Failed to persist block ranges to %s".formatted(filePath), e);
+            LOGGER.log(WARNING, "Failed to persist block ranges to {0}: {1}", filePath, e.getMessage());
         }
     }
 
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -688,9 +688,8 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
             Files.createLink(filePath, tmp);
             Files.deleteIfExists(tmp);
         } catch (IOException | UnsupportedOperationException e) {
-            // UnsupportedOperationException is thrown by Files.createLink() on filesystems that don't
-            // support hard links; it is not a subtype of IOException, so it must be caught explicitly
-            // or it silently escapes this method, leaving the .tmp file orphaned on disk.
+            // UnsupportedOperationException (hard links unsupported) is not an IOException subtype,
+            // so it must be caught here or it silently escapes, leaving the .tmp file orphaned.
             LOGGER.log(WARNING, "Failed to persist block ranges to {0}: {1}", filePath, e.getMessage());
         }
     }
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -684,16 +684,13 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
             final Path tmp = filePath.resolveSibling(filePath.getFileName() + ".tmp");
             final Bytes json = BlockRangesState.JSON.toBytes(toBlockRangesState());
             Files.write(tmp, json.toByteArray());
-            try {
-                Files.move(tmp, filePath, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE);
-            } catch (AtomicMoveNotSupportedException e) {
-                LOGGER.log(
-                        INFO,
-                        "Atomic move not supported on this filesystem for {0}; falling back to non-atomic replace",
-                        filePath);
-                Files.move(tmp, filePath, StandardCopyOption.REPLACE_EXISTING);
-            }
-        } catch (IOException e) {
+            Files.deleteIfExists(filePath);
+            Files.createLink(filePath, tmp);
+            Files.deleteIfExists(tmp);
+        } catch (IOException | UnsupportedOperationException e) {
+            // UnsupportedOperationException is thrown by Files.createLink() on filesystems that don't
+            // support hard links; it is not a subtype of IOException, so it must be caught explicitly
+            // or it silently escapes this method, leaving the .tmp file orphaned on disk.
             LOGGER.log(WARNING, "Failed to persist block ranges to {0}: {1}", filePath, e.getMessage());
         }
     }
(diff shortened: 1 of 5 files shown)


jsync-swirlds:
We originally used create link because of the exception handling _required_ for atomic move, which adds expensive processing (stack walks are efficient, but still require a lot of processing compared to non-exception processing).
Given that we should never see deployments using overlay filesystems (you're not supposed to write to the image filesystem in an OCI container, why aren't we using volume claims in these tests?), we believed (in discussion) using a hard link is more efficient and _fails when it should fail_.

Rather than changing the code to allow writing to OCI container overlay filesystem, the tests should (in my opinion) mount a local or temp filesystem in the expected places.

Nana-EC:
> We originally used create link because of the exception handling _required_ for atomic move, which adds expensive processing (stack walks are efficient, but still require a lot of processing compared to non-exception processing). Given that we should never see deployments using overlay filesystems (you're not supposed to write to the image filesystem in an OCI container, why aren't we using volume claims in these tests?), we believed (in discussion) using a hard link is more efficient and _fails when it should fail_.
> 
> Rather than changing the code to allow writing to OCI container overlay filesystem, the tests should (in my opinion) mount a local or temp filesystem in the expected places.

A few things here.
There are volume claims, `application-state` is already a PVC in the solo-e2e (@AlfredoG87 to confirm). I did some digging and I believe the issue here is the E2E kind cluster is created with no `--config` hence no `extraMounts`. Without those, `kind` uses its default `local-path-provisioner` StorageClass, which  backs every PV with a `hostPath` directory (inside the kind node container's own overlay2 filesystem). So a PVC object exists, but the bytes still land on overlay storage in this specific CI setup.

So the better approach would be in line with your thinking and add `extraMounts` to the kind config so PVs are backed by a real host directory instead of the node's overlay layer.

That's the CI test side, back to the product side question this unearths - createLink vs atomic more. There is a crash safety gap with  `deleteIfExists(filePath) → createLink(...)`. I think that's a bigger concern than how `try/catch`impacts perf and the stack. The perf concern is only present in environments where this write isn't supported. In that case it's an operational blocker (which would be unearthed in live deployment testing) and not a perf concern .

I think we should do the CI change but I also think we should revisit the preference for createLink over the atomic move.

jsync-swirlds:
I don not believe there is a difference in crash safety.
We are missing an exception handling element in the create link path (we still need to handle UncheckedIOExecption, which is highly unfortunate but the NIO library requires it), but that isn't a difference in safety, it's a missing exception handling case in the current code.

The primary difference (when accounting for resilience requirements) is that the unchecked exception handling can happen much further up the stack in the "create link" case because we can report a failed persistence rather than catching very close and retrying.

Nana-EC:
> I don not believe there is a difference in crash safety. We are missing an exception handling element in the create link path (we still need to handle UncheckedIOExecption, which is highly unfortunate but the NIO library requires it), but that isn't a difference in safety, it's a missing exception handling case in the current code.
> 
> The primary difference (when accounting for resilience requirements) is that the unchecked exception handling can happen much further up the stack in the "create link" case because we can report a failed persistence rather than catching very close and retrying.

Did some more digging and this seems to highlight an inconsistency on the BaseSuite setup of file system bound directories.
I've revert the product change, added the missing exception and am exploring the test changes only

jsync-swirlds (dismissed):

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -712,8 +712,10 @@ private void persistBlockRanges() {
             Files.deleteIfExists(filePath);
             Files.createLink(filePath, tmp);
             Files.deleteIfExists(tmp);
-        } catch (IOException e) {
-            LOGGER.log(WARNING, "Failed to persist block ranges to %s".formatted(filePath), e);
+        } catch (IOException | UnsupportedOperationException e) {
```
ata-nas, line 690: Is it possible that other exception types be thrown? Perhaps `UncheckedIOException`? I do agree that we should probably not catch a wildcard, just wondering if there are some other "expected" runtime exceptions we should be aware of.

ata-nas (dismissed):
LG, note that your branch seems out of date, a rebase should probably fix the failing CI.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: Based on the `Files.*()` calls made I don't think `UncheckedIOException` is needed. `IOException` & `UnsupportedOperationException` seem correctly scoped.
Let me know if you see a gap

jsync-swirlds (dismissed):

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: Just checking with you, haven't checked for gaps, but different FSs might have different implementation.

ata-nas (approved):

jsync-swirlds (approved):

PR #3316, opened 2026-07-29
fix: /statusz/inbound and /statusz/outbound 404 due to double mount-prefix check

alex-au:
### Problem

`curl http://<host>:40983/statusz/inbound` returns 404 (`Unknown statusz subpath`).
Only the malformed `curl http://<host>:40983/statusz/statusz/inbound` actually
returns data.

`HealthServicePlugin` registers the `/statusz` service as a wildcard route via
`register(STATUSZ_PATH, httpRules -> httpRules.get("*", this::handleStatusz))`.
Helidon strips the `/statusz` mount prefix before invoking that wildcard
handler, so inside `handleStatusz`, `req.path().path()` is already relative to
the mount (e.g. `/inbound`). But `INBOUND_PATH`/`OUTBOUND_PATH` were defined as
`STATUSZ_PATH + "/inbound"` / `STATUSZ_PATH + "/outbound"` — i.e. they still
included the prefix Helidon had already stripped — so `path.endsWith(INBOUND_PATH)`
never matched a real request, and only a doubled path (which leaves one
`/statusz` behind after stripping) happened to satisfy the check.

### Fix

- Redefine `INBOUND_PATH`/`OUTBOUND_PATH` as mount-relative (`/inbound`,
  `/outbound`), with a comment on why they must not repeat `/statusz`.
- `HealthServiceTest`'s stubbed `TestServerRequest` paths were set to
  `/statusz/inbound` etc. — the literal value the buggy code expected — so the
  test passed by construction without ever exercising Helidon's real routing.
  Updated them to the mount-relative paths Helidon actually delivers.
- `HealthConnectionCloseTest` is the one test that boots a real Helidon
  `WebServer` and hits `/statusz/inbound` over a live socket, but its
  assertion deliberately skipped the status code, deferring that check to
  `HealthServiceTest` (which, per above, couldn't actually verify it). Added
  the missing `assertEquals(200, response.statusCode(), ...)`.

### Review guide

No JDK is available in the environment I used, so I verified with Docker
(`eclipse-temurin:25-jdk`), running `:health:test` for both affected classes:

- With this fix: `HealthServiceTest` 13/13 and `HealthConnectionCloseTest`
  3/3 pass.
- With only the production fix reverted (tests kept as in this PR):
  `HealthConnectionCloseTest#statuszSendsConnectionClose` fails with a real
  `HTTP/1.1 404 Not Found` from the live socket, reproducing the reported
  `curl` symptom end-to-end; `HealthServiceTest`'s inbound/outbound tests fail
  the same way.

```
./gradlew :health:test --tests "org.hiero.block.node.health.HealthServiceTest" \
  --tests "org.hiero.block.node.health.HealthConnectionCloseTest"
```

## Reviewer Notes

No behavior change for `/healthz/livez` or `/healthz/readyz` — those routes
are registered directly (not via a wildcard + manual path dispatch) and were
never affected.

## Related Issue(s)

None filed — reported directly via a live `curl` repro against a running
node.

### DIFF
--- a/block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
+++ b/block-node/health/src/main/java/org/hiero/block/node/health/HealthServicePlugin.java
@@ -43,8 +43,8 @@ public class HealthServicePlugin implements BlockNodePlugin {
     protected static final String LIVEZ_PATH = "/livez";
     protected static final String READYZ_PATH = "/readyz";
     protected static final String STATUSZ_PATH = "/statusz";
-    protected static final String INBOUND_PATH = STATUSZ_PATH + "/inbound";
-    protected static final String OUTBOUND_PATH = STATUSZ_PATH + "/outbound";
+    protected static final String INBOUND_PATH = "/inbound";
+    protected static final String OUTBOUND_PATH = "/outbound";
 
     /// Counter for health-check HTTP requests, labeled by [#LABEL_ENDPOINT] and [#LABEL_RESULT].
     /// Kubernetes probes flatlining this counter while the pod stays up is an early signal that the
--- a/block-node/health/src/test/java/org/hiero/block/node/health/HealthConnectionCloseTest.java
+++ b/block-node/health/src/test/java/org/hiero/block/node/health/HealthConnectionCloseTest.java
@@ -110,10 +110,13 @@ class HealthConnectionCloseTest
     @Test
     @DisplayName("statusz responses also send Connection: close on the wire")
     void statuszSendsConnectionClose() throws IOException {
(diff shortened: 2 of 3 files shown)


jsync-swirlds (approved):

jsync-swirlds:
DCO is failing; you need to run `git commit --amend --signoff` to add a DCO line in the commit message.

alex-au:
> DCO is failing; you need to run `git commit --amend --signoff` to add a DCO line in the commit message.

Thanks Joseph, the commit is now signed and DCO passed.

Nana-EC (approved):

PR #3320, opened 2026-07-29
feat: warn on sidecar records not tied to a parent-record-file transaction

### LINKED ISSUE #3319: validate + wrap: warn on sidecar records not tied to a transaction in the parent record file
## Problem

`SidecarIntegrityValidation` proves the sidecar bytes embedded in a WRB match the SHA-384 hash the CN signed in the record file's `sidecars[]` manifest. `SignatureValidation` proves the CN produced those record file bytes. Composed, they prove: "we have the exact sidecar bytes the CN produced, referenced by the exact hash the CN signed."

**None of those existing checks prove that the sidecar bytes were produced *for this block*.**

A hypothetical CN-side producer bug could:

1. Write a `SidecarFile` with `TransactionSidecarRecord` entries whose `consensus_timestamp` fields point at transactions from a different block period
2. Enter a matching SHA-384 hash for those bytes in *this* block's record file `sidecars[]` manifest
3. Sign the record file with that manifest entry

Under that bug: `SignatureValidation` verifies (the manifest is signed and correct), `SidecarIntegrityValidation` verifies (the hash matches), yet the sidecar content is semantically unrelated to the block it's attached to.

## Evidence this class of bug is real

The mainnet investigation in #3196 surfaced three WRBs (52,333,943 / 58,446,161 / 62,113,066) where the embedded sidecar's internal `consensus_timestamp` values were **33 minutes, 73 minutes, and 13 hours** after the block window they were named against. `SidecarIntegrityValidation` caught them via a different failure mode (the manifest was empty, so the sidecar was flagged as `TAMPERED_OR_EXTRA`), but a variant of that CN-side bug where the manifest hash *did* match would slip through both existing checks.

Ref: sidecar internal timestamps documented in the PR #3219 investigation write-up.

## Proposed check

For every `TransactionSidecarRecord` inside every embedded `SidecarFile`, verify its `consensus_timestamp` appears as the `consensus_timestamp` of at least one `RecordStreamItem` in the block's `RecordStreamFile.recordStreamItems[]`.

```
sidecarTs = record.consensus_timestamp
requires: sidecarTs is the consensus_timestamp of at least one
          RecordStreamItem in the block's RecordStreamFile.recordStreamItems
```

## Severity and integration decisions

- **Severity: warn-only, do NOT fail the block.** Logged at WARN level with a diagnostic naming block number, sidecar index, record index, and offending timestamp. This keeps the check as a soft defense-in-depth signal without blocking a validate/wrap run on a class of anomaly we don't yet have a remediation flow for. If a mismatch surfaces at scale we can revisit and promote it to a hard failure.
- **Runs alongside `SidecarIntegrityValidation`.** Same call sites (validate pipeline + wrap-time pre-embed check) with the same input data (parsed `RecordFileItem` + its `SidecarFile[]`). No extra parse work — the sidecar records and record stream items are already in memory by the time the existing sidecar check runs.
- **Wire into `LiveSequential`'s wrap-time path too.** Currently `SidecarIntegrityValidation.validateSidecars(...)` is called at wrap time by both `ToWrappedBlocksCommand` (per PR #3196) and `LiveSequential` (per PR #3196). The new timestamp warning must fire at the same two sites so live wraps surface the same signal as one-shot wraps. Because it's warn-only, no wrap abort happens.
- **Produce a consolidated list of affected blocks.** Per-block WARN lines get lost in a 97 M block run's stdout. Track every block that warned in an in-memory list and emit both a summary at end-of-run (block-number list, matching what `blocks validate-sidecars` already does for its failure list) and a machine-readable output file (`sidecar-timestamp-mismatches.txt` or similar in the output directory) so ops can grep / diff it programmatically.

### Implementation sketch (~50 lines)

1. Build a `Set<Timestamp>` from `RecordStreamFile.recordStreamItems[].record.consensusTimestamp` once per block (cached alongside the existing sidecar parsing).
2. Walk each `SidecarFile.sidecarRecords[].consensusTimestamp` and check set-membership.
3. On any miss, emit a WARN-level log line naming block number, sidecar index, record index, and the offending timestamp. Do NOT throw.
4. Record the (blockNumber, sidecarIndex, recordIndex, offendingTimestamp) tuple in a thread-safe accumulator so the run's finalize step can emit the consolidated list.
5. Continue to the next record — record all mismatches per block in a single log burst so investigators see the full picture.
6. On finalize, write:
   - A summary line to stderr: `Timestamp-mismatch warning fired on N block(s): [list of block numbers]`
   - A per-tuple text file (`sidecar-timestamp-mismatches.txt`) into the validation output directory, one line per (block, sidecar, record, offendingTimestamp) tuple, so downstream tooling can consume it

Data is already parsed into memory by the existing validation pipeline; this is a scan pass with no extra I/O.

### Where it lives

Fold into `SidecarIntegrityValidation` as an additional check performed inline with the existing hash-match pass. Rationale:

- Same inputs (parsed sidecars + record file contents), same call sites, same wire-up. One class, two levels of severity (existing hard-fail for hash mismatch; new WARN for timestamp mismatch).
- Keeps the wrap-time integration surface simple: `LiveSequential` and `ToWrappedBlocksCommand` already invoke `SidecarIntegrityValidation.validateSidecars(...)` — no new integration point needed.
- The affected-blocks accumulator + finalize output live on `SidecarIntegrityValidation`'s existing lifecycle (matching the `finalize(long blocksValidated, long durableWatermark)` pattern used by other `BlockValidation` implementations).

If keeping the two checks in one class starts to muddy diagnostics, split later into a sibling class. Start folded.

## Why the other 15 items from Steven's list weren't chosen

Analysis in the thread on Slack (short version):

- **5 items** (record file internal invariants: tx ordering, tx window, tx presence, version, metadata signature) are redundant with `SignatureValidation` — if the CN signature verifies, the record file's internal content is what the CN produced.
- **3 items** (sidecar hash correspondence, no orphans) are redundant with `SidecarIntegrityValidation`.
- **2 items** (contiguous block numbers, block timestamp monotonicity) are redundant with the hash chain.
- **2 items** (filename conventions) don't apply to WRBs.
- **3 items** (`SidecarMetadata.id` contiguity, `types` matches sidecar records, `migration=true` block pattern) are defense-in-depth against classes of bug that can't occur if the primary checks pass, or have unclear failure modes.

This issue is scoped to #14 only — the single item that gives genuinely new coverage.

## Acceptance criteria

- [ ] `SidecarIntegrityValidation` extended with a second pass that scans every `TransactionSidecarRecord.consensusTimestamp` against the record file's transaction timestamps
- [ ] Mismatches emit a WARN-level log; **no `ValidationException` thrown for this class of mismatch**
- [ ] Same check fires at all three sites `SidecarIntegrityValidation.validateSidecars(...)` is already called: `ValidateBlocksCommand` (read-back), `ToWrappedBlocksCommand` (one-shot wrap), and `LiveSequential` (live wrap)
- [ ] Thread-safe accumulator collects every warning tuple across the run
- [ ] Finalize step emits:
  - stderr summary: `Timestamp-mismatch warning fired on N block(s): [<block-number-list>]`
  - Machine-readable file `sidecar-timestamp-mismatches.txt` (one line per (block, sidecar, record, offendingTimestamp) tuple) in the run's output directory
- [ ] Unit tests cover:
  - Sidecar with all timestamps matching record file → pass silently (no WARN emitted, no accumulator entry)
  - Sidecar with one timestamp not in record file → check completes, WARN emitted, no exception, one accumulator entry
  - Sidecar with `sidecar_records` empty → pass trivially (no timestamps to check)
  - Block with no sidecars → pass trivially
  - `RecordStreamFile.recordStreamItems` empty (unusual but legal) → any non-empty sidecar warns for every record, all captured in accumulator
  - Sidecar hash-mismatch AND timestamp-mismatch → hash-mismatch still throws (existing behavior preserved), timestamp WARN still emitted for the parts that could be checked
  - Finalize with zero mismatches → summary file NOT created (or created empty, TBD by reviewer preference)
  - Finalize with multiple mismatches across multiple blocks → summary file lists all tuples, ordered by block number
- [ ] Docs updated in `tools-and-tests/tools/docs/blocks-commands.md`: expand the `SidecarIntegrityValidation` subsection to describe both severity levels (hard-fail for hash, WARN for timestamp) and the affected-blocks output file
- [ ] Runs cleanly against a full mainnet WRB archive: warns on the same three known blocks (52,333,943 / 58,446,161 / 62,113,066) as `SidecarIntegrityValidation`'s hash check catches, `sidecar-timestamp-mismatches.txt` contains their tuples, no other unexpected warnings

## Related

- #3196: sidecar integrity validation (established the composed checks this issue extends)
- #3224: wrap-side orphan filter (prevention of the specific bug variant observed on mainnet)
- PR #3235: validation catalog docs (update this check's subsection when it lands)

rockysingh:
## Summary

Fixes #3319.

This PR extends `SidecarIntegrityValidation` with an additional validation that verifies every `TransactionSidecarRecord.consensus_timestamp` contained within an embedded `SidecarFile` exists as the `consensus_timestamp` of a `RecordStreamItem` in the parent `RecordStreamFile`.

This check is **warn-only** and does **not** throw a `ValidationException`. The existing combination of `SignatureValidation` and the sidecar SHA-384 verification already guarantees that the sidecar bytes are exactly what the Consensus Node (CN) produced. A timestamp mismatch therefore indicates a CN-side consistency issue (for example, a sidecar attached to the wrong record file), rather than corruption or tampering.

This provides an additional defense-in-depth signal for identifying and triaging CN anomalies without rejecting otherwise valid data.

This is the class of issue observed in the three mainnet orphan-sidecar investigations (blocks **52,333,943**, **58,446,161**, and **62,113,066**), where the embedded sidecar record timestamps were **33 minutes**, **73 minutes**, and **13 hours** outside their parent record stream file windows. Those cases were already detected by the existing integrity checks, but this validation surfaces the inconsistency directly and provides more actionable diagnostics.

## Validation behavior

| Validation | Trigger | Severity | Reporting |
|------------|---------|----------|-----------|
| **Existing:** Sidecar SHA-384 matches signed manifest | Hash mismatch or missing sidecar | Throws `ValidationException` (`TAMPERED_OR_EXTRA` or `MISSING`) | Exception |
| **New:** Sidecar record `consensus_timestamp` exists in parent `RecordStreamFile` | Any sidecar record timestamp missing from the parent record stream | WARN only (no exception) | WARN log, validation accumulator, final summary, and `sidecar-timestamp-mismatches.txt` |

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SidecarIntegrityValidation.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SidecarIntegrityValidation.java
@@ -6,15 +6,29 @@ import static org.hiero.block.tools.blocks.validation.ProtobufParsingConstants.M
 import static org.hiero.block.tools.utils.Sha384.sha384Digest;
 
 import com.hedera.hapi.block.stream.RecordFileItem;
+import com.hedera.hapi.node.base.Timestamp;
 import com.hedera.hapi.streams.RecordStreamFile;
+import com.hedera.hapi.streams.RecordStreamItem;
 import com.hedera.hapi.streams.SidecarFile;
 import com.hedera.hapi.streams.SidecarMetadata;
+import com.hedera.hapi.streams.TransactionSidecarRecord;
 import com.hedera.pbj.runtime.ParseException;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Queue;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.function.Consumer;
 import org.hiero.block.internal.BlockUnparsed;
 import org.hiero.block.tools.records.model.parsed.ValidationException;
 
@@ -32,18 +46,47 @@ import org.hiero.block.tools.records.model.parsed.ValidationException;
  * ({@code RecordFileItem.recordFileContents.sidecars[]}) are embedded at wrap time, so
  * historical blocks can be validated retroactively without re-wrapping.
  *
- * <p>Behaviour:
+ * <p>Two severity levels are enforced:
+ *
  * <ul>
- *   <li>Blocks with no sidecars pass trivially.</li>
- *   <li>Blocks with sidecars but no {@code recordFileContents} fail: no signed hash list to
- *       check against.</li>
- *   <li>Every sidecar must have a matching hash in the signed list; any sidecar without a match
- *       fails the block with a diagnostic that names the sidecar index and computed hash.</li>
+ *   <li><b>Hash mismatch — hard fail.</b> Every sidecar's SHA-384 must appear in the signed
+ *       hash list, and every signed hash must have a matching sidecar file. Failure throws a
+ *       {@link ValidationException} classifying the discrepancy (TAMPERED_OR_EXTRA / MISSING).</li>
+ *   <li><b>Sidecar {@link TransactionSidecarRecord#consensusTimestamp() consensus_timestamp}
+ *       does not tie back to a transaction in the parent RecordStreamFile — warn only.</b>
+ *       For each record in each sidecar we check that its {@code consensus_timestamp} appears
+ *       as the {@code consensus_timestamp} of at least one {@code RecordStreamItem} in the
+ *       record file. Mismatches log a WARN line but do not throw, since composition of the
+ *       hash check + {@code SignatureValidation} already proves the sidecar bytes are what
+ *       the CN produced — a timestamp miss only signals a CN-side "wrong sidecar attached to
+ *       this block" bug that we track for triage rather than block on. Every warning is
+ *       recorded in an in-memory accumulator so {@link #finalize(long, long)} can emit a
+ *       consolidated affected-blocks summary and {@link #save(Path)} can write a
+ *       machine-readable {@code sidecar-timestamp-mismatches.txt} file. See issue #3319.</li>
  * </ul>
  */
 public final class SidecarIntegrityValidation implements BlockValidation {
 
     private static final int MAX_DEPTH = 512;
+
+    /** Filename of the machine-readable mismatch report written by {@link #save(Path)}. */
+    static final String MISMATCH_REPORT_FILE_NAME = "sidecar-timestamp-mismatches.txt";
+
+    /**
+     * One recorded timestamp-mismatch. Emitted as a WARN line at check time and appended to the
+     * instance's accumulator for later reporting.
+     *
+     * @param blockNumber        block whose embedded sidecar contained the offending record
+     * @param sidecarIndex       zero-based index into {@code RecordFileItem.sidecarFileContents}
+     * @param recordIndex        zero-based index into {@code SidecarFile.sidecarRecords}
+     * @param timestampSeconds   the record's {@code consensus_timestamp} seconds field
+     * @param timestampNanos     the record's {@code consensus_timestamp} nanos field
+     */
+    public record TimestampMismatch(
+            long blockNumber, int sidecarIndex, int recordIndex, long timestampSeconds, int timestampNanos) {}
+
+    /** Thread-safe accumulator populated by {@link #validate} across parallel worker threads. */
+    private final Queue<TimestampMismatch> timestampMismatches = new ConcurrentLinkedQueue<>();
 
     @Override
     public String name() {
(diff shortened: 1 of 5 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SidecarIntegrityValidation.java
```diff
@@ -178,6 +253,98 @@ public static void validateSidecars(
...
+    /**
+     * Emit a consolidated affected-blocks summary to {@code stderr} at end of run. No-op if no
+     * timestamp mismatches were recorded. Does not throw — timestamp mismatches are warn-only.
+     */
+    @Override
+    public void finalize(final long totalBlocksValidated, final long lastBlockNumber) {
+        if (timestampMismatches.isEmpty()) {
+            return;
+        }
+        final Set<Long> distinctBlocks = new TreeSet<>();
```
Nana-EC, line 341: I don't think we have test coverage for after the timestampMismatches?
`finalize_emptyAccumulator_noOp` only covers the 

Claude suggestion below:
Two behaviours need tests:

1. **Block deduplication** — two `TimestampMismatch` entries for the same block number should produce `"fired on 1 block(s): [42]"`, not `"fired on 2 block(s)"`. The `TreeSet` dedup is correct in the production code, but without a test, a future refactor to stream/count without dedup would pass silently.

2. **Message format** — the summary message is consumed by ops scripts. The format `"[SidecarTimestamp] WARN Timestamp-mismatch warning fired on N block(s): [...]"` is undocumented contract; a test would lock it in.

**Suggested fix:** Add a test that populates the accumulator with entries across two distinct block numbers, redirects `System.err` to a `ByteArrayOutputStream`, calls `finalize(0, 0)`, and asserts the message appears with the correct deduplicated count and sorted block list.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SidecarIntegrityValidation.java
```diff
@@ -178,6 +253,98 @@ public static void validateSidecars(
...
+        if (timestampMismatches.isEmpty()) {
+            return;
+        }
+        Files.createDirectories(directory);
+        final Path outFile = directory.resolve(MISMATCH_REPORT_FILE_NAME);
+        final List<TimestampMismatch> sorted = new ArrayList<>(timestampMismatches);
+        sorted.sort(Comparator.comparingLong(TimestampMismatch::blockNumber)
+                .thenComparingInt(TimestampMismatch::sidecarIndex)
+                .thenComparingInt(TimestampMismatch::recordIndex));
+        try (BufferedWriter w = Files.newBufferedWriter(outFile)) {
```
Nana-EC: This is a direct write.
In other cases we've done write-to-temp + atomic `Files.move` pattern to account for potential JVM issue and avoid half written files

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SidecarIntegrityValidation.java
```diff
@@ -84,35 +127,60 @@ public void validate(final BlockUnparsed block, final long blockNumber) throws V
...
+    void record(final TimestampMismatch mismatch) {
+        timestampMismatches.offer(mismatch);
+    }
+
+    /**
+     * Legacy no-record-stream-file overload retained for wrap-time callers that already pass
+     * only the manifest list. Runs only the hash check; the timestamp cross-check is skipped
+     * because we have no {@code RecordStreamFile} to draw transaction timestamps from.
+     */
+    public static void validateSidecars(
```
Nana-EC: Are we keeping this intentionally or should we remove it given this skips the vlaidation.
WHich flow would call this?

Nana-EC (commented):
LG, a suggestion and clarification

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SidecarIntegrityValidation.java
rockysingh: Fixed in 080eef436. Switched to write-to-temp + `Files.move(ATOMIC_MOVE)` with an `AtomicMoveNotSupportedException` fallback for filesystems that don't support it (tmpfs, some NFS, jimfs in tests). Same pattern as `ConvertAddressBookHistoryCommand.atomicMoveWithFallback` from PR #3294. Temp file is cleaned up on error. New test `save_atomicWrite_noHalfWrittenFileOnFailure` asserts the temp file doesn't linger after a successful rename.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SidecarIntegrityValidation.java
rockysingh: Good callout. I've deleted the 3-arg overload in 080eef436. Confirmed no production caller was using it (both wrap-time sites `ToWrappedBlocksCommand` and `LiveSequential` already invoke the 5-arg version); only tests were still on the old shape, and I updated those to pass `null` for the RecordStreamFile + sink so hash-only coverage is explicit rather than silently going through a skip-the-timestamp-check escape hatch.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SidecarIntegrityValidation.java
rockysingh: Filled both gaps in 080eef436:

* `finalize_multipleMismatchesSameBlock_dedupToOneBlock` — inserts three tuples on block 42 + one on block 99 and asserts the summary says "fired on 2 block(s): [42, 99]", not "4". Guards the TreeSet dedup against a future refactor that streams-and-counts without dedup, exactly as you called out.
* `finalize_singleBlock_summaryFormatStable` — pins the exact `[SidecarTimestamp] WARN Timestamp-mismatch warning fired on N block(s): [...]` shape so ops scripts that grep for it don't break silently on a future reword.

Test coverage went 21 -> 24, all passing.

Nana-EC (approved):

jsync-swirlds (approved):

PR #3328, opened 2026-07-30
fix: CloudStorageArchivePlugin handleVerification NumberFormatException

### LINKED ISSUE #3282: CloudStorageArchivePlugin handleVerification NumberFormatException
### Description

The CloudStorageArchivePlugin is unable to handle malformed inputs e.g. "2026-05-01_21-32-38_0000000000000116700-0000000000000116799"

> WARNING [org.hiero.block.node.cloud.storage.archive.CloudStorageArchivePlugin handleVerification] Could not complete uploading blocks to cloud archive storage
> java.lang.NumberFormatException: For input string: "2026-05-01_21-32-38_0000000000000116700-0000000000000116799"
> 	at java.base/java.lang.NumberFormatException.forInputString(Unknown Source)

### Steps to Reproduce

1. Pass ArchiveKey.parse a key of format "2026-05-01_21-32-38_0000000000000116700-0000000000000116799" instead of form `0000/0000/0000/0001/23.tar`

### Additional context

_No response_

ata-nas:
* Make ArchiveKey.parse resilient to malformed keys: log debug and return an UNPARSEABLE sentinel instead of throwing
  * Detect every malformed condition with explicit checks (prefix mismatch, non numeric segments, wrong digit count, overflow), keeping a catch of RuntimeException only as a safety net for unexpected library exceptions
* Skip foreign objects during completed objects recovery and start fresh when the bucket contains only foreign objects
* Leave a hanging multipart upload untouched when its key is not a valid archive key and recover from completed archives instead
* Skip malformed temp archive keys during temp archive recovery
* Log caught exceptions with the throwable signature, guarded by a logger level check
* Add tests
  * ArchiveKey parse returns UNPARSEABLE for every malformed path, grouping level and prefix combination, plus a format and parse round trip across all combinations
  * Recovery skips a foreign object next to a valid archive, starts fresh with only foreign objects, and leaves a foreign hanging upload untouched

Resolves #3282

### DIFF
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.cloud.storage.archive;
 
+import static java.lang.System.Logger.Level.DEBUG;
+
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.ArrayList;
 import java.util.List;
@@ -12,6 +14,13 @@ import java.util.List;
 /// The last segment has its leading zeros stripped (standard integer formatting), so the
 /// round-trip `parse(format(groupStart, level), level) == groupStart` always holds.
 final class ArchiveKey {
+
+    private static final System.Logger LOGGER = System.getLogger(ArchiveKey.class.getName());
+
+    /// Sentinel returned by [#parse] for keys that do not conform to the archive key format
+    /// produced by [#format]. The bucket may legitimately contain foreign objects created by
+    /// other tools; their keys must not fail archiving, they are logged at debug and skipped.
+    static final long UNPARSEABLE = -1L;
 
     /// Width, in characters, of each `/`-delimited path segment produced by [format] and consumed
     /// by [parse] (and, for the segment count alone, by [StartupRecoveryTask#directoryDepth]).
@@ -54,6 +63,12 @@ final class ArchiveKey {
     /// the group-level trailing zeros are restored.  When `objectKeyPrefix` is non-empty, the
     /// prefix and its trailing `/` separator are removed before parsing.
     ///
+    /// A key that does not conform to the format produced by [format] (for example an object
+    /// created in the bucket by another tool, such as
+    /// `2026-05-01_21-32-38_0000000000000116700-0000000000000116799`, or a key outside the
+    /// configured prefix) is not an error: it is logged at debug level and [#UNPARSEABLE] is
+    /// returned so the caller can skip the key. This method never throws.
+    ///
     /// **Width of the last segment** for each grouping level:
     /// | level | last-segment width |
     /// |---|---|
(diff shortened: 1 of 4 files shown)


block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/StartupRecoveryTask.java
```diff
@@ -216,44 +227,51 @@ private void collectTarKeys(S3Client s3, String prefix, int remainingDepth, Stri
...
-                    key, uploadId, parts.stream().map(PartInfo::etag).toList());
-            LOGGER.log(TRACE, "Completed hanging upload to create temporary S3 object at key {0}", key);
-            final String newUploadId =
-                    s3.createMultipartUpload(key, config.storageClass().name(), S3UploadUtils.CONTENT_TYPE);
-            final List<String> newEtags = new ArrayList<>();
+        if (groupStart == ArchiveKey.UNPARSEABLE) {
+            // The hanging upload was not created by this plugin; leave it untouched and recover
+            // from the completed archives instead.
+            LOGGER.log(DEBUG, "Hanging upload key {0} is not a valid archive key, ignoring it", key);
+            result = recoverFromCompletedObjects(s3);
```
ata-nas: We should probably abort this case instead of just ignoring it.

ata-nas: Made this change.

ivannov (commented):
LGTM

ivannov (approved):

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
```diff
@@ -63,23 +78,90 @@ private ArchiveKey() {}
...
+            if (!allSegmentsNumeric(segments)) {
+                // format() only ever emits decimal digit segments; this is a foreign object
+                // sharing the bucket.
+                LOGGER.log(DEBUG, "Key {0} is not a valid archive key", key);
+                result = UNPARSEABLE;
+            } else {
+                // Restore leading zeros stripped by format().
+                final int lastSegmentWidth = lastSegmentWidth(groupingLevel);
+                segments[segments.length - 1] =
+                        String.format("%0" + lastSegmentWidth + "d", Long.parseLong(segments[segments.length - 1]));
```
jsync-swirlds, line 124: This is an odd combination of concatenation and format.
We should probably pick one or the other.

Recommendation: Do not change now; just note for follow-up then resolve this conversation.

```suggestion
                        "%%0%dd".formatted(lastSegmentWidth).formatted(Long.parseLong(segments[segments.length - 1]));
```

jsync-swirlds (approved):

Nana-EC (approved):

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/ArchiveKey.java
ata-nas: Thanks for noting the nit. This, being one-off, however, I'll let pass in favor of time.

PR #3329, opened 2026-07-30
docs: fix GitBook rendering issues across block-node docs

deshmukhpranali:
## Reviewer Notes

-  Convert < dl >/< dt >/< dd > Key terms section in network-ports-and-protocols.md to markdown bullets, GitBook strips HTML definition list formatting leaving unrendered text: 
image

- Remove duplicate Table of Contents sections from metrics.md and quickstart.md:
image

- Fix SVG image sizing in architecture-overview.md, data-flow.md, and plugins.md by replacing plain markdown image syntax with "< img width="100%" >"  to prevent small rendering in GitBook.

### DIFF
--- a/docs/block-node/metrics.md
+++ b/docs/block-node/metrics.md
@@ -1,26 +1,5 @@
 # Hiero Block Node Metrics
 
-## Table of Contents
-
-1. [Summary](#summary)
-2. [Purpose](#purpose)
-3. [Configuration](#configuration)
-4. [How to Access Metrics](#how-to-access-metrics)
-5. [Metrics](#metrics-by-plugin)
-   1. [app](#app)
-   2. [Block Access](#block-access)
-   3. [Block Messaging](#block-messaging)
-   4. [Publisher](#publisher)
-   5. [Subscriber](#subscriber)
-   6. [Verification](#verification)
-   7. [files.recent](#filesrecent)
-   8. [files.historic](#fileshistoric)
-   9. [cloud-storage-archive](#cloud-storage-archive)
-   10. [Server Status API](#server-status-api)
-   11. [Backfill](#backfill)
-   12. [roster-bootstrap-rsa](#roster-bootstrap-rsa)
-   13. [roster-bootstrap-tss](#roster-bootstrap-tss)
-
 ## Summary
 
 This document describes the metrics that are available in the system, its purpose, and how to use them.
--- a/docs/block-node/operations/network-ports-and-protocols.md
+++ b/docs/block-node/operations/network-ports-and-protocols.md
@@ -4,19 +4,21 @@ This document defines the network ports, traffic directions, and TLS posture of 
 
 ## Key terms
 
-<dl>
-<dt>Initiator</dt>
-<dd>The component that opens the TCP connection. For every gRPC flow in this document the initiator is the gRPC client; the Block Node accepts the connection on a listening port.</dd>
+**Initiator**
 
-<dt>Direction</dt>
-<dd>Relative to the Block Node. Inbound traffic terminates on a Block Node port; outbound traffic originates from the Block Node and terminates on a port elsewhere.</dd>
+The component that opens the TCP connection. For every gRPC flow in this document the initiator is the gRPC client; the Block Node accepts the connection on a listening port.
 
-<dt>TLS-in-process</dt>
-<dd>Whether the Block Node binary itself terminates TLS. The Block Node does not terminate TLS in-process for any port; TLS is terminated upstream by a Kubernetes Ingress, load balancer, or similar.</dd>
+**Direction**
 
-<dt>Production exposure</dt>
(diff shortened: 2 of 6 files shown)


docs/block-node/metrics.md
```diff
@@ -1,26 +1,5 @@
...
-   5. [Subscriber](#subscriber)
-   6. [Verification](#verification)
-   7. [files.recent](#filesrecent)
-   8. [files.historic](#fileshistoric)
-   9. [cloud-storage-archive](#cloud-storage-archive)
-   10. [Server Status API](#server-status-api)
-   11. [Backfill](#backfill)
-   12. [roster-bootstrap-rsa](#roster-bootstrap-rsa)
-   13. [roster-bootstrap-tss](#roster-bootstrap-tss)
-
```
jsync-swirlds, line 23: I have a concern here with removing the table of contents.
While Gitbook generates a ToC, Github and IDEs do not, so we remove a feature that's redundant in one view, but it's now completely absent from all other views, including the most frequently used.

Can we leave this in place?

jsync-swirlds: View in Github:

Image

docs/block-node/quickstart.md
```diff
@@ -1,14 +1,5 @@
...
 
-## Table of Contents
-
-1. [Configuration](#configuration)
-2. [Running locally](#running-locally)
-   1. [Build the Server](#build-the-server)
-   2. [Run the Server](#run-the-server)
-   3. [Run the Server with Debug](#run-the-server-with-debug)
-   4. [Stop the Server](#stop-the-server)
-
```
jsync-swirlds, line 11: Similar concern here, this leaves the document with no table of contents anywhere _except_ Gitbook, and my experience is that technical readers often view the source in Github more often than a separate documentation site.

docs/block-node/operations/network-ports-and-protocols.md
```diff
@@ -4,19 +4,10 @@ This document defines the network ports, traffic directions, and TLS posture of
...
-
-<dt>Production exposure</dt>
-<dd>Whether the port is intended to be reachable from outside the Kubernetes cluster in a production deployment. Internal-cluster ports are still subject to <code>NetworkPolicy</code> within the cluster.</dd>
-</dl>
+- **Initiator** - The component that opens the TCP connection. For every gRPC flow in this document the initiator is the gRPC client; the Block Node accepts the connection on a listening port.
+- **Direction** - Relative to the Block Node. Inbound traffic terminates on a Block Node port; outbound traffic originates from the Block Node and terminates on a port elsewhere.
+- **TLS-in-process** - Whether the Block Node binary itself terminates TLS. The Block Node does not terminate TLS in-process for any port; TLS is terminated upstream by a Kubernetes Ingress, load balancer, or similar.
+- **Production exposure** - Whether the port is intended to be reachable from outside the Kubernetes cluster in a production deployment. Internal-cluster ports are still subject to `NetworkPolicy` within the cluster.
```
jsync-swirlds: Unfortunately this change makes the content significantly less clear.
I'm not sure what drives the "make gitbook work" requirement, but Github renders these documents much better without this change:

Image

jsync-swirlds (commented):
A bit concerned about these changes that significantly damage usability in Github (which I believe is the primary viewing environment) to accept gitbook limitations.

docs/block-node/operations/network-ports-and-protocols.md
deshmukhpranali: Thanks for the feedback, @jsync-swirlds! You're right that the < dl> version is significantly clearer on GitHub - the indented definition style is much easier to read but GitBook strips < dl >/< dt >/< dd> tags entirely and renders the content as plain unformatted text on docs.hiero.org.

I've updated the PR to use a middle-ground format - bold the term as its own paragraph, followed by the definition - rather than the flat bullet list. It doesn't have the CSS indentation of < dd>, but the visual hierarchy (term clearly separated from its definition) should be preserved on both GitHub and GitBook.

docs/block-node/metrics.md
deshmukhpranali: @jsync-swirlds , Slack conversation link for your above concern:
https://swirldslabs.slack.com/archives/C08HECALMNF/p1785767854349789?thread_ts=1785766660.139129&cid=C08HECALMNF

docs/block-node/operations/network-ports-and-protocols.md
jsync-swirlds: I still lack clarity on why we selected gitbook for documentation (I thought we had a different toolset in use that has much wider markdown feature support); but that can be addressed separately.

jsync-swirlds (approved):

docs/block-node/operations/network-ports-and-protocols.md
Nana-EC: Other option is a proper md sub heading so it's also more defined

Nana-EC (approved):

Issue #2961, opened 2026-06-04
Docs & deployment: special-purpose WRB BN operator runbook


Nana-EC:
Part of epic #2954 — _Distribute historical & live WRBs and TSS to Tier-1 Block Nodes_.

**Prereq:** T1, T3, T4 (their configs/paths must exist to document accurately).

## Story Form

As a node operator deploying the special-purpose / Tier-1 nodes,
I want a runbook covering address-book conversion, configuration, and the TSS file move,
So that I can stand up and operate the WRB distribution path correctly.

## Technical Notes

- Document: running T1 backfill and enabling T2 live push; running the T3 conversion and configuring the T4 roster.
- TSS operator step (no code): move `/mnt/wrb-operations/wrappedBlocks/tss-bootstrap-roster.json` → `/opt/hiero/block-node/node/`; warn against shared-location writes. Optionally fold the move into a deploy script.
- Configuration guidance for sphere operators needing different address-book history; publish the special-purpose BN IP to Council Tier-1 operators.

## Design

`docs/design/wrb-streaming/sp-wrb-bn-design.md`

Nana-EC:
@deshmukhpranali once the epic is done @rockysingh can support you with content for improvement here

PR #3333, opened 2026-07-30
chore: add previewnet Tier 1 LFH deployment config 

### LINKED ISSUE #3323: Deploy Tier 0 Block Node to previewnet using solo provisioner.
### Story Form

As a **Block Node operator** validating the Solo Provisioner single-VM deployment shape on a live public network before applying it more broadly
I want to **deploy a Tier 1 Block Node to previewnet using `sudo solo-provisioner block node install -p previewnet`**
So that we have a reference previewnet Tier 1 endpoint (default plugin preset is "Tier 1 — Local Full History"), and so we've proven out the Solo Provisioner flow end-to-end on a live public network with the sizing and OS combinations the docs prescribe.

### Reference material

- **Deployment guide (canonical):** https://docs.hiero.org/block-node-overview/block-node-hardware-specifications/solo-weaver-single-node-k8s-deployment (Deploy with Solo Provisioner — Recommended)
- **Solo Provisioner tool:** https://github.com/hashgraph/solo-weaver (releases + install script)
- **Block Node chart / releases:** https://github.com/hiero-ledger/hiero-block-node/releases
- **BNCE infrastructure repo (informational):** https://github.com/hashgraph/block-node-infrastructure — this is the multi-VM BNCE cutover tooling. Single-VM Solo Provisioner is a different, simpler flow; this repo is only relevant if we later fold the Solo Provisioner install into a BNCE-style workflow, not for the first-cut previewnet deployment.

### Deployment recipe (from the canonical guide)

Concrete step summary, straight from the linked docs, with previewnet-specific parameters substituted in:

**1. GCP VM.**
- Machine type: **`e2-standard-16`** (16 vCPUs, ≥32 GB RAM) per the doc's "For `previewnet` or `testnet`" recommendation.
- Boot disk: **Debian GNU/Linux 12 (bookworm)** — Solo Provisioner is tested against it. Avoid the cloud-default image because that shifts over time.
- Firewall: default GCP rules allow only SSH (port 22). If we plan to reach the BN from outside the VM (e.g. for external subscriber testing), add a firewall rule permitting TCP `40840` on the VM's network tag.

**2. Install Solo Provisioner.**
```bash
gcloud compute ssh --zone <ZONE> <INSTANCE_NAME> --project <PROJECT_ID>
curl -sSL https://raw.githubusercontent.com/hashgraph/solo-weaver/main/install.sh | bash
sudo solo-provisioner -v   # confirms install
```

**3. Install the Block Node (interactive; accept defaults for the first pass).**
```bash
sudo solo-provisioner block node install -p previewnet
```
Defaults land as:
- Kubernetes namespace: `block-node`
- Helm release: `block-node`
- Storage: individual paths under `/mnt/fast-storage`
- **Plugin preset: `Tier 1 — Local Full History`** (blocks on local disk — this is exactly what the ticket is asking for)
- Chart: `oci://ghcr.io/hiero-ledger/hiero-block-node/block-node-server` at whatever version the current Solo Provisioner release bundles. Cross-check against the [hiero-block-node releases page](https://github.com/hiero-ledger/hiero-block-node/releases) before running to confirm we're picking up a version compatible with previewnet's Tier 0 upstream.

For a reproducible / scripted run, use `--non-interactive` after we've settled on the values:
```bash
sudo solo-provisioner block node install -p previewnet --non-interactive
```
Custom Helm values or config file via `--values` / `--config` if we need to override defaults (e.g. pin a specific chart version).

**4. Verify pod health.**
```bash
kubectl get pods -A
```
Expect the `block-node` namespace pods `Running` with all containers ready (e.g. `1/1` / `2/2`).

**5. Verify RPC accessibility.**
```bash
# Install grpcurl + fetch the protobuf artifact for the release
curl -L https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz -o grpcurl.tar.gz
sudo tar -xzf grpcurl.tar.gz -C /usr/local/bin grpcurl && rm grpcurl.tar.gz
curl -s https://api.github.com/repos/hiero-ledger/hiero-block-node/releases/latest \
  | grep "browser_download_url.*block-node-protobuf.*tgz" | cut -d : -f 2,3 | tr -d \" | wget -qi -
tar xzf block-node-protobuf-*.tgz  # produces block-node-protobuf-<VERSION>/

# From inside the VM, hit the BN's serverStatus:
grpcurl -plaintext -emit-defaults \
  -import-path block-node-protobuf-<VERSION> \
  -proto block-node/api/node_service.proto \
  -d '{}' <BN_INTERNAL_IP>:40840 org.hiero.block.api.BlockNodeService/serverStatus
```
Fresh install returns `firstAvailableBlock` / `lastAvailableBlock` = `18446744073709551615` (UINT64_MAX sentinel = "no blocks yet"). After the BN starts syncing from previewnet's Tier 0, these should advance.

### Prerequisites / blocked-by

- GCP project + billing enabled, with `gcloud` CLI authenticated and permissions to create Compute Engine VMs.
- `bn-eng-bnce-archiver-sa@block-node-dev.iam.gserviceaccount.com` IAM issue (missing `roles/serviceusage.serviceUsageConsumer`): **not a blocker for this deployment** — Solo Provisioner's single-VM install does not go through the requester-pays GCS path that hit the archiver SA. That IAM issue only bites the BNCE / wrb-cli flow. Confirmed by reading the deploy recipe end-to-end.
- Confirm the previewnet Tier 0 upstream (host + port) the Tier 1 will subscribe from. If Solo Provisioner's `previewnet` profile defaults don't already point at it, we'll need to override via `--values` / `--config` on the install step.

### Acceptance criteria

- [ ] `e2-standard-16` VM created in GCP, running Debian 12, reachable via `gcloud compute ssh`
- [ ] `sudo solo-provisioner -v` returns a version string on the VM
- [ ] `sudo solo-provisioner block node install -p previewnet` completes with `Completed successfully`; setup report + provisioner log land under `/opt/solo/weaver/logs/`
- [ ] `kubectl get pods -A` shows `block-node` namespace pods all `Running` with expected container readiness
- [ ] `grpcurl ... serverStatus` returns a valid response from the BN
- [ ] BN is wired to and receiving blocks from previewnet's Tier 0 upstream; `lastAvailableBlock` advances over a sustained observation window (say 10 min) rather than staying pegged at the `UINT64_MAX` sentinel
- [ ] Downstream sanity: at least one subscriber (mirror node, a script subscriber, or another BN) can connect to the previewnet Tier 1's Subscribe endpoint and consume blocks
- [ ] Firewall rule for TCP `40840` in place (if external subscribers need to reach the BN); or explicit decision to leave it VM-internal
- [ ] Cross-linked from the `hashgraph/block-node-infrastructure` runbook (or a new `previewnet/` directory there) so the next operator can repeat the deployment without re-reading this ticket

### Out of scope

- Multi-VM BNCE cutover workflow (that's `hashgraph/block-node-infrastructure`'s territory; Solo Provisioner's single-VM install is a different shape). If we later want a BNCE-style workflow that automates this Solo Provisioner install as part of a larger orchestration, that's a follow-up ticket.
- Mainnet Tier 1 — the linked docs page explicitly says Solo Provisioner on a single GCP VM is *not* the right shape for production Tier 1 mainnet and points at the Bare Metal Single Node Kubernetes Deployment guide instead. This ticket is scoped to previewnet only.

### Related

- Deployment guide: hiero.org Solo Provisioner single-node k8s (link above)
- Solo Provisioner v0.3.0+ deprecates the old `block node setup` subcommand — this ticket uses the current `block node install` per the guide
- Epic: #3125 (BNCE / WRB distribution) for the Tier 0 / Tier 1 topology context (previewnet Tier 1 here is a similar shape to what BNCE stands up on the multi-VM path)

rockysingh:
## Summary

Fixes #3323. Adds configuration + operational tooling for deploying a previewnet Tier 1 Local-Full-History Block Node via Solo Provisioner and seeding it with historical wrapped blocks. Config-only PR — smoke-test install was performed against the previewnet VM at `34.70.11.223` and confirmed working end-to-end (see "Validation" below); nothing in this PR touches deployed infrastructure automatically.

## What's in the PR

Five YAML files under `charts/block-node-server/values-overrides/` (following the existing `lfh-values.yaml` / `backfill.yaml` / `mini.yaml` naming convention), one operational script under `tools-and-tests/tools/scripts/`, and one operator runbook under `docs/block-node/operations/`.

**Main pair — for a properly-sized previewnet Tier 1 host**

1. `previewnet-lfh-provisioner-config.yaml` — Solo Provisioner config
   - `profile: previewnet`, chart pinned at `0.37.1`, namespace/release `block-node`
   - Storage: 1 TiB archive, 100 GiB live, 20 GiB app-state, 20 GiB logs, 5 GiB plugins under `/mnt/fast-storage`
   - Header documents the two preflight gotchas we hit (hedera UID/GID must be 2000, 3 TB disk minimum) with fixes
2. `previewnet-lfh-values.yaml` — Helm values override
   - JVM heap `16G`, resources `8 CPU / 20 Gi` (right-sized for `e2-standard-16`)
   - `ROSTER_BOOTSTRAP_RSA_MIRROR_NODE_BASE_URL` → previewnet mirror node
   - `loadBalancer.enabled: true` + `metallb.io/address-pool: public-address-pool` — Solo Provisioner installs MetalLB during a fresh install; the annotation picks up the pool it configures. Team direction: use LB for external subscribers, not `kubectl port-forward`.
   - Backfill / Prometheus / plugin ports / init-container storage-dir chown / PVC references mirror the mainnet reference (`lfh-values.yaml`)

**Smoke-test variants — for a smaller-disk / bigger-RAM validation host**

3. `previewnet-lfh-provisioner-config-smoketest.yaml` — same Solo Provisioner config with storage shrunk to fit ~1 TB total disk (600 GiB archive, 50 GiB live, smaller others; combined ~700 GiB footprint)
4. `previewnet-lfh-values-smoketest.yaml` — bigger JVM heap (`96G`) and resources (`16 CPU / 128 Gi`) because the actual smoke-test box had 256 GB RAM. Same LoadBalancer + MetalLB setup as the main values so the smoke-test variant differs only in sizing.

Headers on both smoke-test files explicitly mark them as smoke-test-only, call out the `--skip-hardware-checks` CLI flag they require, and warn against using them for a durable Tier 1.

**Operational script — for seeding the BN with historical blocks**

5. `tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh` — general-purpose backfill helper. Stages a local wrappedBlocks archive via `blocks bulk-load`, tar-streams it into the BN's pod at the chart's historic mount path, rolls the StatefulSet so `BlockFileHistoricPlugin` re-scans on startup and picks up the copied files.
   - Uses `blocks bulk-load` (PR #3038) not `blocks push` (PR #3263): the live publish/gRPC stream is a single-slot-per-block-number CAS meant for freshly-produced blocks and silently SKIPs anything that loses the race, so it's the wrong path for seeding an empty BN from history.
   - Target BN addressable by kubectl context / namespace / statefulset / pod / container — all defaults match Solo Provisioner's `block-node/block-node-block-node-server/block-node-server` shape.
   - Preflight (pod Ready check) before staging; `SKIP_ROLLOUT=true` and `KEEP_STAGING=true` escape hatches; closing log line emits the exact grpcurl verification command.

**Static-PV helper — for single-node clusters that lack a default StorageClass**

6. `charts/block-node-server/values-overrides/previewnet-lfh-static-pvs.yaml` — five pre-bound hostPath `PersistentVolume` objects with `claimRef`s already targeting the chart's `volumeClaimTemplate` PVC names (`*-storage-block-node-block-node-server-0`). Solo Provisioner's stock previewnet install does NOT ship a default `StorageClass`, so `persistence.*.create: true` (added by `ffbdcedb3`) would otherwise leave the chart-provisioned PVCs `Pending` forever. `kubectl apply -f` this file BEFORE `solo-provisioner install` and the PVCs bind immediately.

**Operator runbook — covers the T1 backfill portion of #2961**

7. `docs/block-node/operations/wrb-seed-runbook.md` — step-by-step runbook covering fresh install + historic-archive seeding via `backfill-wrb-to-bn.sh`, plus symptom/cause/workaround for the three operational footguns discovered during smoke-testing (StorageClass requirement, `subPath: archive-data` mismatch, `BlockFileHistoricPlugin` crash-loop on single corrupt zip). Each footgun links a pending upstream fix.

## How to use

**Prerequisite (single-node clusters without a default StorageClass):**

```bash
kubectl apply -f /path/to/previewnet-lfh-static-pvs.yaml
```

**Production shape** (proper `n2d-standard-16` or larger, ≥3 TB disk):

```bash
sudo solo-provisioner block node install \
  -p previewnet \
  --config /path/to/previewnet-lfh-provisioner-config.yaml \
  --values /path/to/previewnet-lfh-values.yaml \
  --non-interactive
```

**Smoke-test shape** (smaller disk, needs the hardware-check bypass):

```bash
sudo solo-provisioner block node install \
  -p previewnet \
  --config /path/to/previewnet-lfh-provisioner-config-smoketest.yaml \
  --values /path/to/previewnet-lfh-values-smoketest.yaml \
  --skip-hardware-checks \
  --non-interactive
```

**Backfill after install** (from the same host, or any host with `kubectl` pointed at the cluster):

```bash
tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh /path/to/wrappedBlocks
```

Full recipe (VM creation, Solo Provisioner install, verification via `kubectl` + `grpcurl serverStatus`) is in the linked hiero.org docs page referenced in #3323.

## Preflight gotchas discovered during validation

Documented in the main config's header:

- `hedera` user/group must be at UID/GID 2000. If the target VM was previously used for anything, a pre-existing `hedera` at another UID (we saw 1005) will fail preflight with `hedera owner user "hedera" has incorrect UID: expected 2000, got 1005`. Fix: `sudo usermod -u 2000 hedera && sudo groupmod -g 2000 hedera`, then re-chown anything that was owned by the old IDs.
- The `previewnet` profile requires ≥3 TB total disk at preflight. Use `--skip-hardware-checks` together with the smoke-test variants on smaller boxes.

## Known limitation — chart 0.37.1 LoadBalancer port surface

Chart 0.37.1's `service-loadbalancer.yaml` template hard-codes a single port entry (40840). Under the split-port setup we use, each API family lives on its own dedicated port (`server-status` on 40982, `publisher` on 40984, `subscriber` on 40980, `block-access` on 40981, `health` on 40983), and port 40840 alone serves plain HTTP rather than gRPC — external `grpcurl :40840 serverStatus` fails with `malformed header: missing HTTP content-type`.

For chart 0.37.1, external gRPC probing therefore requires `kubectl port-forward` against the ClusterIP Service on the plugin port. Documented in both values files' `loadBalancer:` block comments. Chart 0.38.0+ adds `loadBalancer.includePluginPorts: true` which fixes this by appending the plugin ports to the LB Service; revisit when we bump the chart pin.

## Validation

Ran the smoke-test path end-to-end against the previewnet VM at `34.70.11.223`:

- Solo Provisioner `Completed successfully` in 40s after hedera UID/GID + LoadBalancer flag alignment
- Pod comes up `1/1 Running`; StatefulSet ready
- Service exposes all 7 ports on ClusterIP (`server-status: 40982`, `publisher: 40984`, `subscriber: 40980`, `blockAccess: 40981`, `health: 40983`, `metrics: 16007`, `http: 40840`)
- MetalLB assigns `10.128.0.2` to the `-external` LoadBalancer Service (from `public-address-pool`)
- `serverStatus` gRPC (via kubectl port-forward on 40982) returns:
  ```json
  {
    "firstAvailableBlock": "18446744073709551615",
    "lastAvailableBlock": "18446744073709551615",
    "onlyLatestState": false,
    "nextExpectedBlock": "0"
  }
  ```
  UINT64_MAX sentinel = "no blocks stored yet" (expected on a fresh install); `nextExpectedBlock: 0` confirms the BN is ready to accept blocks from genesis.

## Validation — seeded end state (Aug 3)

Extended the smoke-test on `34.70.11.223` end-to-end through a historic-archive seed. Ran `backfill-wrb-to-bn.sh --install-and-seed` against a 173 GB WRB archive (171 files × ~10K blocks each), pointed the BN at the seeded PVCs, confirmed the historic plugin indexed and served the blocks.

Startup log after the seed:

```
2026-08-03 22:17:10.171+0000 INFO [org.hiero.block.node.app.BlockNodeApp start]
  Started BlockNode Server : State=RUNNING HistoricBlockRange=0->1699999
```

`serverStatus` gRPC (via `kubectl port-forward` on the ClusterIP Service, port 40982):

```json
{
  "firstAvailableBlock": "0",
  "lastAvailableBlock": "1699999",
  "onlyLatestState": false,
  "nextExpectedBlock": "0"
}
```

170 of 171 files parsed cleanly; one file at `000/000/000/000/17/00000.zip` (blocks 170000–179999) was rejected by the plugin as content-corrupt (valid ZIP structurally, but the plugin's block-parser can't read the entry). The full recipe for reproducing this end-to-end (including the three manual footgun workarounds we hit along the way) is in the new operator runbook: [wrb-seed-runbook.md](./docs/block-node/operations/wrb-seed-runbook.md).

## Operational footguns discovered during seeding

Three real issues hit during the seeded smoke test. All non-blocking for this PR (config-only) but documented in the operator runbook with symptom / cause / workaround / pending-upstream-fix for each. To be filed as follow-up bugs against the appropriate repos:

1. **`persistence.*.create: true` requires a default `StorageClass`** — worked around in this PR by shipping `previewnet-lfh-static-pvs.yaml`. Real fix pending against solo-weaver.
2. **`archive-storage` volumeMount applies `subPath: archive-data`** even when the values file explicitly sets `subPath: ""`, so `backfill-wrb-to-bn.sh`'s seeded files land one level above where the pod reads. Manual `mv` required. Real fix pending against this repo's chart template.
3. **`BlockFileHistoricPlugin` crash-loops on subsequent restart if the archive contains any single content-corrupt zip.** One bad file poisons `maxStoredBlockNumber` → `IllegalStateException` at `init:181`. Manual cleanup of `corrupted/` + empty leaf dir + stale state file required. Real fix pending against `BlockFileHistoricPlugin`.

## Reference material

- Deployment guide: https://docs.hiero.org/block-node-overview/block-node-hardware-specifications/solo-weaver-single-node-k8s-deployment
- Solo Provisioner tool: https://github.com/hashgraph/solo-weaver
- Mainnet reference this was derived from: `charts/block-node-server/values-overrides/lfh-values.yaml` + `lfh-provisioner-config.yaml`
- Reference bulk-load pattern the backfill script generalizes: `bulk-load-historical-to-bn1.sh` (on branch `3125-slice4-steps8-9`, not yet on main)

## Out of scope

- Pointing the BN at a previewnet Tier 0 upstream for live subscription — see #3323 for the operational follow-up
- Bumping the chart version pin from 0.37.1 to a version that supports `includePluginPorts` (would unlock external gRPC probing via the LB IP without kubectl port-forward)
- Multi-VM BNCE-style automation — see `hashgraph/block-node-infrastructure` if we later want to fold Solo Provisioner into a larger orchestration
- Mainnet Tier 1 — docs explicitly recommend a bare-metal shape rather than Solo Provisioner for prod mainnet; this config is scoped to previewnet only

### DIFF
--- a/docs/block-node/operations/wrb-seed-runbook.md
+++ b/docs/block-node/operations/wrb-seed-runbook.md
@@ -1,10 +1,10 @@
-# Previewnet Tier 1 LFH — Deployment + Historic Seed Runbook
+# WRB Seeding Runbook
 
 ## Overview
 
-This runbook covers deploying a Local-Full-History (LFH) Block Node to previewnet via Solo Provisioner and seeding it with a historical [Wrapped Record Block](../glossary.md#wrb-wrapped-record-block) (WRB) archive. Ties together the config from PR #3333 (previewnet Tier 1 values files) with the operational tooling from `backfill-wrb-to-bn.sh`.
+This runbook covers deploying a Local-Full-History (LFH) Block Node via Solo Provisioner and seeding it with a historical [Wrapped Record Block](../glossary.md#wrb-wrapped-record-block) (WRB) archive. The procedure applies to any environment (previewnet, testnet, mainnet) that uses the single-node Solo-Provisioner deployment shape; examples below use previewnet paths and values files, but substituting the equivalent `<env>-lfh-*.yaml` files and `<env>` profile targets the same steps at testnet or mainnet.
 
-**Audience**: Operators standing up a previewnet Tier 1 Block Node for the first time, or seeding an existing empty install with historical blocks.
+**Audience**: Operators standing up a Tier 1 Block Node for the first time, or seeding an existing empty install with historical blocks.
 
 **Scope**: Covers the T1 backfill portion of the WRB distribution operator runbook (part of #2961). Does not cover T2 live-push, T3 address-book conversion, or T4 roster/TSS configuration.
 
--- a/tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh
+++ b/tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh
@@ -1,0 +1,172 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#
+# backfill-wrb-to-bn.sh — Bulk-load a local wrapped-block archive into a
+# deployed Block Node's historic-storage directory, then roll the StatefulSet
+# so the BN's `BlockFileHistoricPlugin` re-scans and picks up the copied files.
+#
+# Use this instead of `blocks push` (gRPC publish stream) whenever you're
+# seeding an EMPTY BN from historical data. The live publish path is a
+# single-slot-per-block-number CAS meant for freshly-produced blocks and
+# silently SKIPs anything that loses the race, which will bite on any block
+# that a peer producer has already streamed (typical on real networks). The
+# `blocks bulk-load` subcommand exists for exactly this case: it copies WRB
+# zips directly into the BN's on-disk historic archive, bypassing live-stream
+# verification entirely.
+#
+# Works against any deployed Block Node addressable via kubectl -- Tier 0 or
+# Tier 1, mainnet / testnet / previewnet / a local kind cluster / etc. Defaults
+# match the shape produced by `sudo solo-provisioner block node install`
+# (namespace `block-node`, release `block-node`, chart mount path
+# `/opt/hiero/block-node/data/historic`); override via env vars for other
+# deployment shapes.
+#
+# Prerequisites:
+#   * `kubectl` on PATH, current context pointed at the target cluster
+#     (or override via BN_KUBE_CONTEXT)
+#   * `java` on PATH
+#   * A shaded wrb-cli jar (`tools-*-all.jar`) on disk -- point at it with CLI_JAR
+#   * A local directory of wrapped blocks (the output of `blocks wrap`)
+#
+# Usage:
+#   ./backfill-wrb-to-bn.sh <path-to-wrappedBlocks-dir>
+#
+# Environment overrides:
+#   BN_KUBE_CONTEXT      kubectl context           (default: current context)
+#   BN_NAMESPACE         Kubernetes namespace      (default: block-node)
+#   BN_STATEFULSET       StatefulSet name          (default: block-node-block-node-server)
+#   BN_POD               Pod name                  (default: <BN_STATEFULSET>-0)
+#   BN_CONTAINER         Container name in pod     (default: block-node-server)
+#   HISTORIC_MOUNT_PATH  On-pod historic dir       (default: /opt/hiero/block-node/data/historic)
+#   STAGING_DIR          Local staging dir         (default: /tmp/bn-backfill-<pid>)
+#   CLI_JAR              Path to wrb-cli jar       (default: hunts tools-*-all.jar)
+#   READY_TIMEOUT        Pod-ready wait (seconds)  (default: 300)
+#   SKIP_ROLLOUT         "true" to skip the pod    (default: false)
+#                        restart step
+#   KEEP_STAGING         "true" to keep the local  (default: false)
+#                        staging dir on success
+#
+# Example (previewnet Tier 1 deployed via Solo Provisioner):
+#   ./backfill-wrb-to-bn.sh ~/wrappedBlocks
+#
+# Example (custom deployment shape, non-standard namespace + container):
+#   BN_NAMESPACE=my-bn \
+#   BN_STATEFULSET=my-bn-server \
+#   BN_CONTAINER=bn \
+#   ./backfill-wrb-to-bn.sh /mnt/archive/wrappedBlocks
+
+set -euo pipefail
+
+# --- Config ---
+: "${BN_KUBE_CONTEXT:=}"                                              # empty -> use current
+: "${BN_NAMESPACE:=block-node}"
+: "${BN_STATEFULSET:=block-node-block-node-server}"
+: "${BN_POD:=${BN_STATEFULSET}-0}"
+: "${BN_CONTAINER:=block-node-server}"
+: "${HISTORIC_MOUNT_PATH:=/opt/hiero/block-node/data/historic}"
+: "${STAGING_DIR:=/tmp/bn-backfill-$$}"
+: "${READY_TIMEOUT:=300}"
+: "${SKIP_ROLLOUT:=false}"
+: "${KEEP_STAGING:=false}"
+
+# --- Logging ---
+log()  { echo "[backfill-wrb-to-bn] $*"; }
+fail() { echo "[backfill-wrb-to-bn] ERROR: $*" >&2; exit 1; }
+
+# --- Argument parsing ---
+[[ $# -ge 1 ]] || fail "usage: $0 <path-to-wrappedBlocks-dir>"
+WRB_SOURCE_DIR="$1"
+[[ -d "${WRB_SOURCE_DIR}" ]] || fail "wrappedBlocks dir does not exist: ${WRB_SOURCE_DIR}"
+
+# --- Locate the wrb-cli jar ---
+if [[ -z "${CLI_JAR:-}" ]]; then
+    # Prefer explicit location next to this script (built via ./gradlew :tools:shadowJar)
+    script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    candidate=$(ls "${script_dir}/../build/libs/"tools-*-all.jar 2>/dev/null | head -1 || true)
+    if [[ -z "${candidate}" ]]; then
+        candidate=$(command -v tools-cli || ls tools-*-all.jar 2>/dev/null | head -1 || true)
+    fi
+    [[ -n "${candidate}" ]] || fail "wrb-cli jar not found; set CLI_JAR=<path> to override"
+    CLI_JAR="${candidate}"
+fi
+[[ -f "${CLI_JAR}" ]] || fail "CLI_JAR does not exist: ${CLI_JAR}"
+
+# --- kubectl invocation prefix (context + namespace) ---
+kctl=(kubectl --namespace "${BN_NAMESPACE}")
+[[ -n "${BN_KUBE_CONTEXT}" ]] && kctl+=(--context "${BN_KUBE_CONTEXT}")
+
+log "Config:"
+log "  source WRB dir:        ${WRB_SOURCE_DIR}"
+log "  staging dir:           ${STAGING_DIR}"
+log "  CLI jar:               ${CLI_JAR}"
+log "  kubectl context:       ${BN_KUBE_CONTEXT:-<current>}"
+log "  namespace:             ${BN_NAMESPACE}"
+log "  target pod:            ${BN_POD} (container ${BN_CONTAINER})"
+log "  target statefulset:    ${BN_STATEFULSET}"
+log "  historic mount path:   ${HISTORIC_MOUNT_PATH}"
+
+# --- Preflight ---
+command -v kubectl >/dev/null || fail "kubectl not on PATH"
+command -v java >/dev/null || fail "java not on PATH"
+
+log "Preflight: verifying target pod ${BN_POD} is Ready..."
+"${kctl[@]}" get pod "${BN_POD}" >/dev/null || fail "pod ${BN_POD} not found in namespace ${BN_NAMESPACE}"
+"${kctl[@]}" wait --for=condition=Ready pod/"${BN_POD}" --timeout=60s \
+    || fail "pod ${BN_POD} not Ready within 60s; aborting before we stage anything"
+
+# --- Stage via `blocks bulk-load` ---
+log "Staging wrapped blocks via 'blocks bulk-load' (${WRB_SOURCE_DIR} -> ${STAGING_DIR})..."
+mkdir -p "${STAGING_DIR}"
+java -jar "${CLI_JAR}" blocks bulk-load \
+    --source "${WRB_SOURCE_DIR}" \
+    --dest   "${STAGING_DIR}" \
+    || fail "'blocks bulk-load' staging failed"
+
+if [[ -z "$(find "${STAGING_DIR}" -name '*.zip' -print -quit 2>/dev/null)" ]]; then
+    fail "no .zip files staged in ${STAGING_DIR}; nothing to backfill"
+fi
+staged_count=$(find "${STAGING_DIR}" -name '*.zip' | wc -l)
+log "Staged ${staged_count} zip file(s) into ${STAGING_DIR}."
+
+# --- Stream into the pod via tar | kubectl exec ---
+log "Streaming staged blocks into ${BN_POD}:${HISTORIC_MOUNT_PATH}..."
+"${kctl[@]}" exec "${BN_POD}" -c "${BN_CONTAINER}" -- mkdir -p "${HISTORIC_MOUNT_PATH}" \
+    || fail "failed to ensure ${HISTORIC_MOUNT_PATH} exists on ${BN_POD}"
+tar -C "${STAGING_DIR}" -cf - . | \
+    "${kctl[@]}" exec -i "${BN_POD}" -c "${BN_CONTAINER}" -- \
+    tar xf - -C "${HISTORIC_MOUNT_PATH}" \
+    || fail "failed to stream staged blocks into ${BN_POD}"
+log "Blocks copied onto ${BN_POD}'s persistent historic volume."
+
+# --- Roll the StatefulSet so BlockFileHistoricPlugin re-scans ---
+if [[ "${SKIP_ROLLOUT}" == "true" ]]; then
+    log "SKIP_ROLLOUT=true; not rolling the StatefulSet. The BN will pick up the new"
+    log "historic files on its next natural restart. If you don't want to wait, run:"
+    log "  kubectl rollout restart statefulset/${BN_STATEFULSET} -n ${BN_NAMESPACE}"
+else
+    log "Rolling statefulset/${BN_STATEFULSET} so BlockFileHistoricPlugin re-scans..."
+    "${kctl[@]}" rollout restart statefulset/"${BN_STATEFULSET}" \
+        || fail "rollout restart failed for statefulset/${BN_STATEFULSET}"
+    "${kctl[@]}" rollout status statefulset/"${BN_STATEFULSET}" --timeout="${READY_TIMEOUT}s" \
+        || fail "statefulset/${BN_STATEFULSET} rollout did not complete within ${READY_TIMEOUT}s"
+    "${kctl[@]}" wait --for=condition=Ready pod/"${BN_POD}" --timeout="${READY_TIMEOUT}s" \
+        || fail "${BN_POD} did not become Ready after restart"
+    log "${BN_POD} is Ready after rollout."
+fi
+
+# --- Cleanup ---
+if [[ "${KEEP_STAGING}" == "true" ]]; then
+    log "KEEP_STAGING=true; leaving staging dir at ${STAGING_DIR}"
+else
+    rm -rf "${STAGING_DIR}"
+    log "Removed staging dir ${STAGING_DIR}."
+fi
+
+log "Backfill complete. Verify the BN sees the new blocks with:"
+log "  kubectl port-forward -n ${BN_NAMESPACE} svc/${BN_STATEFULSET} 18082:40982 &"
+log "  grpcurl -plaintext -emit-defaults \\"
+log "    -import-path <path-to-block-node-protobuf> \\"
+log "    -proto block-node/api/node_service.proto \\"
+log "    -d '{}' localhost:18082 org.hiero.block.api.BlockNodeService/serverStatus"
+log "Expect: firstAvailableBlock / lastAvailableBlock now show real block numbers"
+log "(no longer UINT64_MAX sentinel 18446744073709551615)."
--- a/tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh
+++ b/tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh
@@ -1,172 +1,441 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: Apache-2.0
 #
-# backfill-wrb-to-bn.sh — Bulk-load a local wrapped-block archive into a
-# deployed Block Node's historic-storage directory, then roll the StatefulSet
-# so the BN's `BlockFileHistoricPlugin` re-scans and picks up the copied files.
+# backfill-wrb-to-bn.sh -- seed a Block Node deployment with a local
+# wrappedBlocks archive. Three modes:
 #
-# Use this instead of `blocks push` (gRPC publish stream) whenever you're
-# seeding an EMPTY BN from historical data. The live publish path is a
-# single-slot-per-block-number CAS meant for freshly-produced blocks and
-# silently SKIPs anything that loses the race, which will bite on any block
-# that a peer producer has already streamed (typical on real networks). The
-# `blocks bulk-load` subcommand exists for exactly this case: it copies WRB
-# zips directly into the BN's on-disk historic archive, bypassing live-stream
-# verification entirely.
+#   --install-and-seed
+#                   (recommended one-shot for a fresh box)
+#                   Runs `sudo solo-provisioner block node install` to
+#                   create the PVCs + everything, immediately scales the
+#                   StatefulSet to 0 so nothing writes to the archive
+#                   PVC yet, seeds via the pre-install flow, then scales
+#                   the StatefulSet back to 1. Operator ends up with a
+#                   BN serving pre-seeded historical blocks after one
+#                   script invocation. Requires --config and --values
+#                   files (passed through to solo-provisioner).
 #
-# Works against any deployed Block Node addressable via kubectl -- Tier 0 or
-# Tier 1, mainnet / testnet / previewnet / a local kind cluster / etc. Defaults
-# match the shape produced by `sudo solo-provisioner block node install`
-# (namespace `block-node`, release `block-node`, chart mount path
-# `/opt/hiero/block-node/data/historic`); override via env vars for other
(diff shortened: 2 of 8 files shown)


docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+
+---
+
+## Prerequisites
+
+- GCP VM (or equivalent) provisioned per [Solo Weaver Single-Node K8s Deployment](./solo-weaver-single-node-k8s-deployment.md)
+- `sudo solo-provisioner -v` returns a version string on the VM
+- `kubectl` on PATH, current context pointed at the target cluster
+- `java` on PATH (JDK 25+)
+- A shaded tools jar (`tools-*-all.jar`) staged on the VM
```
Nana-EC, line 19: Can be updated to the updated jar in the official releases

docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+- A local WRB archive directory to seed from
+- The 4 values/config files from PR #3333 staged on the VM:
+  - `previewnet-lfh-provisioner-config.yaml` (or the `-smoketest` variant)
```
Nana-EC, line 22: Instructions are env agnostic so should just be `config.yaml` or similar, just drop env .
Same with values

docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Prerequisites
+
+- GCP VM (or equivalent) provisioned per [Solo Weaver Single-Node K8s Deployment](./solo-weaver-single-node-k8s-deployment.md)
+- `sudo solo-provisioner -v` returns a version string on the VM
+- `kubectl` on PATH, current context pointed at the target cluster
+- `java` on PATH (JDK 25+)
+- A shaded tools jar (`tools-*-all.jar`) staged on the VM
+- A local WRB archive directory to seed from
+- The 4 values/config files from PR #3333 staged on the VM:
```
Nana-EC, line 21: Q: why do we need the smoketest configs?

docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+  - `previewnet-lfh-values.yaml` (or the `-smoketest` variant)
+- `previewnet-lfh-static-pvs.yaml` if the target cluster does not have a default `StorageClass` (see [Footgun 1](#footgun-1-persistencecreate-true-requires-a-default-storageclass) below)
+
+---
+
+## Fresh install + seed (recommended)
+
+### Step 1. Pre-provision static PVs (single-node clusters only)
+
+If the target cluster does not have a default `StorageClass` (Solo Provisioner's stock previewnet install does not), pre-apply the static PVs before running `solo-provisioner install`. Otherwise the chart's `volumeClaimTemplate` PVCs stay `Pending` forever and the BN pod cannot schedule.
```
Nana-EC, line 32: Q: is this a previewnet only issue? Did we open a bug for the team to resolve? Seems odd for the tool to assume infrastructure details.

docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+```bash
+kubectl get pv | grep pv-static
+```
+
+### Step 2. Install the Block Node
+
+**Production shape** (properly-sized VM, 3 TB+ disk):
+
+```bash
+sudo solo-provisioner block node install \
```
Nana-EC, line 49: Command shouldn't assume environment

docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+  -p previewnet \
+  --config previewnet-lfh-provisioner-config.yaml \
+  --values previewnet-lfh-values.yaml \
+  --non-interactive
+```
+
+**Smoke-test shape** (smaller-disk VM):
```
Nana-EC, line 56: Q: Is smoke test needed to support smaller VMs?

docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+If `RESTARTS > 0`, see [Footgun 3](#footgun-3-plugin-crash-loop-on-single-corrupt-zip).
+
+### Step 6. Verify blocks are served
+
+```bash
+kubectl port-forward -n block-node svc/block-node-block-node-server 18082:40982 &
+sleep 3
+grpcurl -plaintext -emit-defaults \
```
Nana-EC, line 126: Can also use `tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh` which handles all the tools and protobuf

docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+  localhost:18082 org.hiero.block.api.BlockNodeService/serverStatus
+```
+
+Expected shape after a successful seed:
+
+```json
+{
+  "firstAvailableBlock": "0",
+  "lastAvailableBlock": "<highest block number in the seeded archive>",
+  "onlyLatestState": false,
```
Nana-EC, line 139: nit: this changed to true
```suggestion
  "onlyLatestState": true,
```

docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+
+**Symptom**: after running `backfill-wrb-to-bn.sh` and starting the BN, `serverStatus` returns `firstAvailableBlock: "18446744073709551615"` (UINT64_MAX sentinel) even though the seed reported success. The startup log shows `HistoricBlockRange=` (empty).
+
+**Cause**: the chart's `archive-storage` volumeMount includes `subPath: archive-data`, so the pod's `/opt/hiero/block-node/data/historic` is backed by `<pv-hostpath>/archive-data/` (a subdirectory of the PV) rather than the PV root. The backfill script writes to the PV root, one level above where the pod actually reads.
+
+The values file explicitly sets `subPath: ""` for the archive volume, but the effective pod spec still shows `subPath: archive-data`. Either the chart template ignores the values-file override, or applies a hardcoded default that clobbers it.
+
+**Fix**: after the script completes and BEFORE bringing the BN back up, move the seeded files into the `archive-data/` subdir (Step 4 above).
+
+**Upstream fix pending**: chart template should either honor the values-file `subPath: ""` override, or the backfill script should be aware of the effective `subPath` and write to the correct location directly.
```
Nana-EC, line 178: Please open a follow up bug and bring to team, might be worth addressing

docs/block-node/operations/wrb-seed-runbook.md
```diff
@@ -0,0 +1,224 @@
...
+
+**Symptom**: after seeding, the pod initially starts fine (1/1 Ready, `HistoricBlockRange` populated in the startup log). On the NEXT restart (helm upgrade, pod delete, node reboot) it enters `CrashLoopBackOff` with this exception:
+
+```
+Exception in thread "main" java.lang.IllegalStateException:
+  First zipped block number [0] cannot be greater than the latest zipped block number [-1]
+    at BlockFileHistoricPlugin.init(BlockFileHistoricPlugin.java:181)
+```
+
+**Cause**: during the first successful startup, `BlockFileHistoricPlugin` scans the archive and detects any content-corrupt zip files (valid ZIP structure, but internal block data unreadable). Corrupt files are moved to a `corrupted/` quarantine subdirectory. The quarantine leaves three artifacts that trip up subsequent inits:
```
Nana-EC, line 190: Q: are we saying this is expected in all scenarios or that you observed this once based on a non clean setup?

tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh
```diff
@@ -0,0 +1,441 @@
...
+fail() { echo "[backfill-wrb-to-bn] ERROR: $*" >&2; exit 1; }
+
+# --- Argument parsing ---
+MODE="auto"
+WRB_SOURCE_DIR=""
+# --install-and-seed passthrough flags to solo-provisioner:
+SOLO_PROFILE=""
+SOLO_CONFIG=""
+SOLO_VALUES=""
+SOLO_EXTRA_FLAGS=(--skip-hardware-checks --non-interactive)
```
Nana-EC, line 133: I don't think the skip H/W checks should be default. If an operator expect san undersized setup then let them specific. Having it default might miss cases where an operator should be notified of undersized H/W

tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh
```diff
@@ -0,0 +1,441 @@
...
+    local replicas
+    replicas=$("${kctl[@]}" get statefulset "$BN_STATEFULSET" -o jsonpath='{.spec.replicas}' 2>/dev/null || echo "")
+    if [[ -n "$replicas" && "$replicas" != "0" ]]; then
+        fail "StatefulSet $BN_STATEFULSET has replicas=$replicas; scale to 0 first: kubectl scale statefulset/$BN_STATEFULSET --replicas=0 -n $BN_NAMESPACE"
+    fi
+
+    # 3. Locate PV host path (hostPath OR local)
+    local pv_host_path
+    pv_host_path=$(kubectl get pv "$pv_name" -o jsonpath='{.spec.hostPath.path}')
+    [[ -n "$pv_host_path" ]] || pv_host_path=$(kubectl get pv "$pv_name" -o jsonpath='{.spec.local.path}')
```
Nana-EC, line 224: You're skipping the use of the context-aware kubectl
```suggestion
    [[ -n "$pv_host_path" ]] || pv_host_path=$("${kctl[@]}" get pv "$pv_name" -o jsonpath='{.spec.local.path}')
```

Nana-EC (commented):
Good adds.
Left comments.

I'm not sure all the configs are needed. I'd expect just config and yaml, maybe the pvs.
Maybe put them under `charts/block-node-server/values-overrides/wrb-to-tier-0/*.yaml`

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Reworded prerequisite to point at the official [Block Node releases](https://github.com/hiero-ledger/hiero-block-node/releases) page.

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Dropped env prefixes; runbook now uses `<env>-lfh-*.yaml` placeholders throughout so the same doc covers previewnet, testnet, and mainnet.

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Added a Smoke-test variants section explaining these exist for smaller dev/CI VMs (reduced PV sizes and pod resource requests); not intended for real Tier 1 use.

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Not previewnet-only, any single-node Solo Provisioner install lacks a default StorageClass. Filed follow-up [#3669](https://github.com/hiero-ledger/hiero-block-node/issues/3669) to ship one by default and linked it from Footgun 1.

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Command now uses `-p <env> --config <env>-lfh-*.yaml` placeholders.

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Yes, smoke-test variant covers under-spec VMs (smaller PV sizes and reduced pod resource requests). Documented in the new Smoke-test variants section.

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Added a pointer to `bn-endpoint-checker.sh` as the bundled alternative and included it in the Reference section.

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Applied; corrected to `"onlyLatestState": true`.

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Filed [#3670](https://github.com/hiero-ledger/hiero-block-node/issues/3670) and linked it from Footgun 2.

docs/block-node/operations/wrb-seed-runbook.md
rockysingh: Clarified in the doc: observed once on a smoke-test seed where a single source zip was truncated. Not every seed hits it, but the recovery is deterministic once any file in the source archive is content-corrupt.

tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh
rockysingh: Agreed, removed `--skip-hardware-checks` from the default `SOLO_EXTRA_FLAGS`. Operators on undersized VMs pass `--solo-flag --skip-hardware-checks` explicitly.

tools-and-tests/tools/scripts/backfill-wrb-to-bn.sh
rockysingh: Applied, both jsonpath lookups now use `"${kctl[@]}"`.

rockysingh:
Thanks @Nana-EC. Kept the smoketest variants but the runbook now has a Smoke-test variants section explaining they exist for dev/CI VMs and are not intended for production Tier 1 (happy to drop them if the team prefers). On the subdir move, I can bundle `charts/block-node-server/values-overrides/wrb-to-tier-0/*.yaml` with the testnet + mainnet copies in a follow-up so all three envs move together in one PR.

PR #3334, opened 2026-07-30
fix(test): eliminate flakiness in backfill and slowdown E2E tests

### LINKED ISSUE #2266: test(e2e-suites): Flaky test for `Positive Single Subscriber Tests > Should validate 2ms slowdown after each block in a range`
### Description

Positive Single Subscriber Tests > Should validate 2ms slowdown after each block in a range FAILED
https://github.com/hiero-ledger/hiero-block-node/actions/runs/22390663558/job/64811554928?pr=2259

### Steps to Reproduce

This is a flaky test, it happens from time to time, specially when the CI is congested.

### Additional context

_No response_

### LINKED ISSUE #2285: test(e2e-suite): Positive Single Subscriber Tests > Should subscribe to stream and receive both historical and live FAILED
### Description

Potential Flaky Test:

Positive Single Subscriber Tests > Should subscribe to stream and receive both historical and live FAILED

https://github.com/hiero-ledger/hiero-block-node/actions/runs/22469465166/attempts/1?pr=2250

### Steps to Reproduce

_No response_

### Additional context

_No response_

Nana-EC:
## Summary

Four E2E test flakiness sources fixed in this PR.

### 1. `testAutonomousBackfill` — backfill scanner got only one chance within the poll window

The test already overrides `BACKFILL_INITIAL_DELAY=5000ms` but never overrode `BACKFILL_SCAN_INTERVAL`, which defaults to 60,000ms (the production value). With the existing 60s `awaitBackfilledBlocks` poll, any transient failure on the first scan (e.g. live-storage not yet reporting its range after a Docker restart) meant the second scan wouldn't fire until T+65s — one second after the poll had already expired.

**Fix:** add `BACKFILL_SCAN_INTERVAL=10000ms` (10s) alongside the existing `BACKFILL_INITIAL_DELAY=5000ms` in `BaseSuite.createContainer(config)`. This gives 6 scan opportunities within the 60s window, so any single bad scan is a retry away from success — not a test failure.

### 2. `shouldValidateSlowdownAfterEachBlock` — per-block interval measurement was racy

The test measured wall-clock time between successive `consumedBlocks()` increments and required each interval to be ≥ 2ms. This was fragile because:
- `System.currentTimeMillis()` has ~10–15ms resolution on Linux CI runners vs. a 2ms threshold
- The 1ms polling loop could observe two consecutive block increments in one wake-up, recording a ≈0ms interval for the second block
- The consumer uses two separate internal counters (`blocksConsumed` AtomicLong for slowdown gating vs. `LiveBlocksConsumed` metric used by `getStreamStatus()`), adding an independent desync path

**Fix:** replace the per-block loop with a total-duration measurement anchored to when block 1 is first observed. The 9 remaining blocks (2-10) each carry the 2ms delay, so the minimum total is 9 × 2ms = 18ms — well above any OS timer resolution — and the check is immune to the batch-detection race entirely.

### 3. `shouldSubscribeForHistoricalAndLiveBlocks` — consumer baseline chased a moving publisher target (closes #2285)

The same baseline-initialisation bug fixed in PR #3314 for `PositiveMultipleSubscribersTests` was missed in the single-subscriber counterpart. Setting `lastConsumedBlock = publisherSimulator.publishedBlocks()` as the initial threshold forced the consumer to beat a moving target (the publisher keeps publishing) within only 3 retries (3s), which fails under CI load.

**Fix:** reset baseline to 0 and extend retries to 10 (10s) — consistent with the multi-subscriber fix in PR #3314.

### 4. Publisher spin loops — NPE and CPU starvation on simulator failure

Two busy-wait loops in `shouldSubscribeForHistoricalBlocks` and `shouldValidateSlowdownAfterEachBlock` spun unconditionally on `publisherSimulator.getStreamStatus().publishedBlocks()`. When the simulator receives a gRPC `UNAVAILABLE` (e.g. block node container shutting down mid-test) it stops itself and `getStreamStatus()` returns null, causing an NPE at the call site. The spin also consumed 100% CPU, starving the simulator thread.

**Fix:** guard each loop with `publisherSimulator.isRunning()` so it exits cleanly on failure, null-check the returned `StreamStatus` to cover the TOCTOU window between the guard and the call, and add a `Thread.sleep(10)` to yield the CPU each iteration. Also raise `shouldSubscribeForHistoricalBlocks` consumer retries from 3 to 10 for consistency.

## Test plan
- [ ] `./gradlew :suites:runSuites` passes consistently with the block-node image
- [ ] No regressions in remaining `PositiveMultiplePublishersTests` or `PositiveSingleSubscriberTests` cases
- [ ] CI green

Closes #2266
Closes #2285

### DIFF
--- a/tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
+++ b/tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
@@ -321,6 +321,7 @@ public abstract class BaseSuite {
                 .withEnv("VERSION", blockNodeVersion)
                 .withEnv("BACKFILL_BLOCK_NODE_SOURCES_PATH", config.backfillSourcePath())
                 .withEnv("BACKFILL_INITIAL_DELAY", "5000") // 5 seconds
+                .withEnv("BACKFILL_SCAN_INTERVAL", "10000") // 10 seconds — shorter than the default 60s
                 .withEnv("SERVER_PORT", String.valueOf(blockNodePort))
                 .withEnv(config.envOverrides())
                 .withEnv(
--- a/tools-and-tests/suites/src/main/java/org/hiero/block/suites/subscriber/positive/PositiveSingleSubscriberTests.java
+++ b/tools-and-tests/suites/src/main/java/org/hiero/block/suites/subscriber/positive/PositiveSingleSubscriberTests.java
@@ -186,24 +186,30 @@ public class PositiveSingleSubscriberTests extends BaseSuite {
         final Future<?> consumerSimulatorThread = startSimulatorInThread(consumerSimulator);
         simulators.add(consumerSimulatorThread);
 
-        long previousBlockTime = System.currentTimeMillis();
-        boolean slowdownValidated = true;
-
-        for (long block = startBlock; block <= endBlock; block++) {
-            while (consumerSimulator.getStreamStatus().consumedBlocks() < block) {
-                Thread.sleep(1); // Wait for the block to be consumed
-            }
-            final long currentBlockTime = System.currentTimeMillis();
-            final long timeDifference = currentBlockTime - previousBlockTime;
-
-            if (timeDifference < expectedSlowdownMillis) {
-                slowdownValidated = false;
-                break;
-            }
-            previousBlockTime = currentBlockTime;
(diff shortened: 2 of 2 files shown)


AlfredoG87 (dismissed):
LGTM

jsync-swirlds (approved):

AlfredoG87 (approved):

PR #3336, opened 2026-07-31
chore: Follow up for remaining items from PR 3289.

### LINKED ISSUE #3307: chore: follow-up on non-blocking review comments from #3289 (live roster / address book)
## Story Form

As a maintainer
I want all non-blocking review feedback from #3289 to be tracked

## Technical Notes

Follow-up items from review comments on #3289 — either the one genuinely unresolved thread, or items the PR author (jsync-swirlds) explicitly asked to track in a separate issue because they touch pre-existing code outside the stated scope of #3289 ("those should definitely be captured in a separate issue").

### Correctness / robustness

- [x] **`ApplicationStateUtility.java` — use `Files.isRegularFile` instead of `Files.exists` when checking network-data path** (`ata-nas`): `isRegularFile` returns `false` for non-existent, non-regular, or inaccessible paths, giving a more precise guard than `exists`. If adopted, also update the surrounding log message to `"Network data file not present or inaccessible..."` to match the broader semantics. Deferred from #3289 because the code was only moved (not modified) in that PR.

- [x] **`ApplicationStateUtility.java` — add null guard for `book` in `loadNetworkData`** (`ata-nas`): The parse step could theoretically yield a null `AddressBook`; returning early rather than proceeding is safer. Pre-existing code; deferred from #3289.

### Design

- [x] **`ApplicationStateUtility.java` — validation method should return `boolean` instead of throwing** (`ata-nas`): The current method throws on invalid input; returning `false` (or `Optional`) would allow callers to handle failures without exception control-flow. Pre-existing code; PR author agreed this should be a separate issue.

### Test quality

- [x] **`DynamicBlockItemTest.java` line 515 — add `@Timeout` annotation** (`ata-nas`): The test at that line has no timeout guard; either annotate it with `@Timeout` or add a class-level `@Timeout` for the whole suite to prevent the test from hanging indefinitely in CI.

jsync-swirlds:
## Reviewer Notes
* Added todo items in block node app for issue 3321
* Updated `validateAddressBook`
   * Added null checks
   * Removed use of illegal state exception as a return value
   * Added a boolean return value
* Added Timeout to `DynamicBlockItemTest`

## Related Issue(s)
 Resolves #3307

### DIFF
--- a/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
+++ b/block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
@@ -43,6 +43,7 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.LockSupport;
 import java.util.logging.LogManager;
 import java.util.stream.Collectors;
 import org.hiero.block.api.BlockNodeVersions;
@@ -76,6 +77,13 @@ import org.hiero.block.node.spi.threading.ThreadPoolManager;
 import org.hiero.metrics.ObservableGauge;
 import org.hiero.metrics.core.MetricKey;
 import org.hiero.metrics.core.MetricRegistry;
+
+// @todo(3321) This class uses runtime exceptions (such as
+//     illegal state exception) as flow control in many methods. We need to
+//     rework the logic flow to remove all of those cases and move exception
+//     handling up to where it can be resolved rather than just logging
+//     warnings and continuing as though nothing failed or using the exception
+//     to choose logic branches or replace return values.
 
 /// Main class for the block node server
 public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
@@ -338,34 +346,30 @@ public class BlockNodeApp implements HealthFacility, ApplicationStateFacility {
     /// {@inheritDoc}
     @Override
     public void shutdown(String className, String reason) {
-        state.set(State.SHUTTING_DOWN);
-        LOGGER.log(INFO, "Shutting down, reason={0} class={1}", reason, className);
-        // stop the application state facility
-        stopApplicationStateFacility();
+        try {
+            state.set(State.SHUTTING_DOWN);
+            LOGGER.log(INFO, "Shutting down, reason={0} class={1}", reason, className);
+            // stop the application state facility
+            stopApplicationStateFacility();
 
-        // wait for the shutdown delay
-        try {
-            Thread.sleep(serverConfig.shutdownDelayMillis());
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-            LOGGER.log(INFO, "Shutdown interrupted");
-        }
-        serviceBuilder.stopAll();
-        // Stop all the facilities &  plugins
-        for (BlockNodePlugin plugin : loadedPlugins) {
-            LOGGER.log(INFO, "    {0}", plugin.name());
-            plugin.stop();
-        }
-        // Stop metrics
-        try {
+            // wait for the shutdown delay
+            LockSupport.parkNanos(serverConfig.shutdownDelayMillis() * 1_000_000L);
+            serviceBuilder.stopAll();
+            // Stop all the facilities &  plugins
+            for (BlockNodePlugin plugin : loadedPlugins) {
+                LOGGER.log(INFO, "\t{0}", plugin.name());
+                plugin.stop();
+            }
+            // Stop metrics
             blockNodeContext.metricRegistry().close();
             LOGGER.log(DEBUG, "Metric registry successfully closed.");
+            // finally exit
+            LOGGER.log(INFO, "System Exiting");
+            if (shouldExitJvmOnShutdown) System.exit(0);
         } catch (IOException e) {
-            LOGGER.log(DEBUG, "Could not properly close metric registry.", e);
+            LOGGER.log(INFO, "Could not properly shut down due to IO failure.", e);
+            if (shouldExitJvmOnShutdown) System.exit(0);
         }
-        // finally exit
-        LOGGER.log(INFO, "System Exiting");
-        if (shouldExitJvmOnShutdown) System.exit(0);
     }
 
     /// Main entrypoint for the block node server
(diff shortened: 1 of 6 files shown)


block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -338,34 +346,30 @@ public State blockNodeState() {
...
+                plugin.stop();
+            }
+            // Stop metrics
             blockNodeContext.metricRegistry().close();
             LOGGER.log(DEBUG, "Metric registry successfully closed.");
+            // finally exit
+            LOGGER.log(INFO, "System Exiting");
+            if (shouldExitJvmOnShutdown) System.exit(0);
         } catch (IOException e) {
             LOGGER.log(DEBUG, "Could not properly close metric registry.", e);
```
Nana-EC: With this change the log is misleading. Prior it was tightly wrapped around the metric registry close. Now any issue sin stopping ASF or plugins will have the incorrect message and will be hidden to the operator at the DEBUG level.
I think the message and level should be corrected

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -338,34 +346,30 @@ public State blockNodeState() {
...
-        }
-        // Stop metrics
         try {
+            state.set(State.SHUTTING_DOWN);
+            LOGGER.log(INFO, "Shutting down, reason={0} class={1}", reason, className);
+            // stop the application state facility
+            stopApplicationStateFacility();
+
+            // wait for the shutdown delay
+            LockSupport.parkNanos(serverConfig.shutdownDelayMillis() * 1_000_000L);
```
Nana-EC, line 356: I think Thread.sleep as we have before was the better choice.
Why the change?
In this case we want to ensure as much as possible that we give processes enough time to complete.
Since we don't again check that process have indeed stopped Thread.sleep would be my suggesion

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -338,34 +346,30 @@ public State blockNodeState() {
...
+            serviceBuilder.stopAll();
+            // Stop all the facilities &  plugins
+            for (BlockNodePlugin plugin : loadedPlugins) {
+                LOGGER.log(INFO, "\t{0}", plugin.name());
```
Nana-EC, line 360: Q: Why the change? 
Also this assumes log aggregators won't render this as a literal.

Nana-EC (commented):
Some suggestions around logging and sleep

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: It is still debug, because it's non-critical, the system is shutting down.
I can change the message.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: Why is thread sleep better than park? All sleep does is call park and throw interrupted exceptions if it ends early.
We didn't handle the interrupted exception (except to restore the interrupted flag), so this delay is more reliable (or at least equally reliable) because we still try to continue shut down.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: The `\t` never reaches anything, it's a tab character to the _compiler_. Java replaces the escape when the code is compiled.
It's added because a tab at the line start is cleaner and easier to recognize in log aggregation than multiple spaces.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: Message changed, and level increased to INFO to match other logs in the same method.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: I agree with using park instead of sleep here.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -338,34 +346,30 @@ public State blockNodeState() {
...
+                plugin.stop();
+            }
+            // Stop metrics
             blockNodeContext.metricRegistry().close();
             LOGGER.log(DEBUG, "Metric registry successfully closed.");
+            // finally exit
+            LOGGER.log(INFO, "System Exiting");
+            if (shouldExitJvmOnShutdown) System.exit(0);
```
ata-nas, line 368: Should this be in a `finally` block? Possibly exiting with other than 0 if an issue detected? This will also allow us to not have multiple `exit` calls like we have in the catch IOException block.

ata-nas (approved):
LG, too bad we have to move exception flow control up one level, but I agree there is quite a lot to rework, which is outside the scope of this follow up.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: I tried to avoid making major changes here; as this will all need further rework in future issues.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
Nana-EC: parkNanos does not guarantee that the time has fully elapsed.
I believe the JDK itself calls out that parkNanos has spurious-wakeup semantics where it could return at any time, for no reason, so in general if used it should be invoked within a loop that rechecks conditions on return.

jsync-swirlds: TL;DR - We don't care about the early return in this case, so it's safer and avoids a spurious exception to use parkNanos.

Sleep and parkNanos are _identical in effect_ if all you do with interrupted exception is catch it and continue.

Sleep _calls parkNanos_ and does not use a loop, it throws interrupted exception if the park returns early.

This is a common bit of FUD; using parkNanos is _not_ significantly different from calling Thread.sleep except that parkNanos returns quietly instead of throwing an exception.
Any pause can return early, the question is whether we _do not care_ about the early return, want an exception thrown for early return, or want a loop to detect and manage an early return.

Nana-EC (approved):

PR #3339, opened 2026-07-31
fix: fixed flakey tests in S3ArchivePluginTest.

jsync-swirlds:
## Reviewer Notes
* Fixed two minor flakey tests in S3ArchivePluginTest.

### DIFF
--- a/block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
+++ b/block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
@@ -32,10 +32,12 @@ import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
@@ -110,12 +112,13 @@ class S3ArchivePluginTest extends PluginTestBase<S3ArchivePlugin, ExecutorServic
     }
 
     @Test
+    @Timeout(value = 30, unit = TimeUnit.SECONDS)
     @DisplayName("ArchivePlugin should upload a tar file for single batch of blocks")
     void startWithSingleBatch(@TempDir Path tempDir) throws IOException, InterruptedException {
         // create 10 sample blocks, this should trigger the plugin to archive them
         sendBlocks(START_TIME, 0, 9);
-        // await archive task to complete
-        parkNanos(TASK_AWAIT_NANOS);
+        // await all archive uploads to complete (deterministic; no fixed sleep)
+        awaitArchiveUploadsComplete();
         // send another persisted notification to trigger the executor service
         // cleanup and ensure the task ran
         plugin.handlePersisted(new PersistedNotification(0L, true, 0, BlockSource.UNKNOWN));
@@ -175,12 +178,13 @@ class S3ArchivePluginTest extends PluginTestBase<S3ArchivePlugin, ExecutorServic
     }
 
     @Test
+    @Timeout(value = 30, unit = TimeUnit.SECONDS)
     @DisplayName("ArchivePlugin should upload multiple tar files for multiple batches of blocks")
     void doTenBatches() {
         // create 100 sample blocks, this should trigger the plugin to archive them
         sendBlocks(START_TIME, 0, 99);
-        // await archive task to complete
-        parkNanos(TASK_AWAIT_NANOS * 3);
+        // await all archive uploads to complete (deterministic; no fixed sleep)
+        awaitArchiveUploadsComplete();
         // send another persisted notification to trigger the executor service
         // cleanup and ensure the task ran
         plugin.handlePersisted(new PersistedNotification(0L, true, 0, BlockSource.UNKNOWN));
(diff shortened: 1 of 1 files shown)


block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
```diff
@@ -329,6 +333,25 @@ private void downloadFile(String objectName, Path destFile) {
...
+    /// completed. Deterministic; replaces a fixed sleep that flaked when uploads were slower than the
+    /// guessed wait.
+    private void awaitArchiveUploadsComplete() {
+        try {
+            testThreadPoolManager.executor().submit(() -> {}).get(20, TimeUnit.SECONDS);
+        } catch (final InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Interrupted while awaiting archive uploads", e);
+        } catch (final ExecutionException | TimeoutException e) {
+            throw new RuntimeException("Archive uploads did not complete in time", e);
```
Nana-EC, line 351: I would break these out for more accurate Exception message

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
```diff
@@ -329,6 +333,25 @@ private void downloadFile(String objectName, Path destFile) {
...
+    /// every prior upload — including the final "latest archived block" write to the bucket — has
+    /// completed. Deterministic; replaces a fixed sleep that flaked when uploads were slower than the
+    /// guessed wait.
+    private void awaitArchiveUploadsComplete() {
+        try {
+            testThreadPoolManager.executor().submit(() -> {}).get(20, TimeUnit.SECONDS);
+        } catch (final InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Interrupted while awaiting archive uploads", e);
+        } catch (final ExecutionException | TimeoutException e) {
```
Nana-EC: Also, are we sure `ExecutionException` is valid and reachable here?
`() -> {}` is a Runnable  that can't throw an exception and neither will the `Future.get()`

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
```diff
@@ -329,6 +333,25 @@ private void downloadFile(String objectName, Path destFile) {
...
+    ///
+    /// The plugin runs uploads on the single, test-owned executor (`testThreadPoolManager.executor()`)
+    /// in FIFO order, and all batch uploads for the blocks just sent are submitted synchronously while
+    /// those blocks are delivered. A no-op barrier task on that same executor therefore runs only after
+    /// every prior upload — including the final "latest archived block" write to the bucket — has
+    /// completed. Deterministic; replaces a fixed sleep that flaked when uploads were slower than the
+    /// guessed wait.
+    private void awaitArchiveUploadsComplete() {
+        try {
+            testThreadPoolManager.executor().submit(() -> {}).get(20, TimeUnit.SECONDS);
```
Nana-EC: 20 seconds should be an input param.
The current tests pass use a desired timeout, that should be preserved and allow for other tests to specify the wait

Nana-EC (commented):
Nice improvement, a few more tweaks necessary

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
jsync-swirlds: If we don't catch here, we must declare ExecutionException thrown from the method, because it's a checked exception and Future.get() declares it (Future.get() wraps all exceptions from the runnable or callable it represents, including runtime, in an ExecutionException when it's called).

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
jsync-swirlds: It's a pretty small difference, but can adjust.

Reminder, this class is part of a deprecated plugin we plan to remove soon.
This fix is just because the code was failing unit tests frequently enough to cause problems.

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
jsync-swirlds: This code waits for a defined event (the executor is serial, so this releases as soon as the tested actions run), which doesn't require a fixed timeout (and using the former very small timeout _is what made the test flakey_).

Re-adding the tight timeout here would just make the test flakey again.

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
jsync-swirlds: Done

berryware (dismissed):
Looks good.

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
```diff
@@ -329,6 +333,27 @@ private void downloadFile(String objectName, Path destFile) {
...
+    /// those blocks are delivered. A no-op barrier task on that same executor therefore runs only after
+    /// every prior upload — including the final "latest archived block" write to the bucket — has
+    /// completed. Deterministic; replaces a fixed sleep that flaked when uploads were slower than the
+    /// guessed wait.
+    private void awaitArchiveUploadsComplete() {
+        try (ExecutorService executor = testThreadPoolManager.executor()) {
+            executor.submit(() -> {}).get(20, TimeUnit.SECONDS);
+        } catch (final InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Interrupted while awaiting archive uploads", e);
```
ata-nas, line 349: Not sure we should throw here, if the test runner thread is interrupted, this most likely means that it is being shutdown due to `@Timeout`. If we throw, we might silence that, if a test times out, we will see it clearly in the console out later.

It might be better to not catch the InterruptedException here, let it propagate and let ti `throws` on tests that call it, jUnit will handle the rest.

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
ata-nas: Same here as above, we can simply propagate this I believe, jUnit should handle the rest.

ata-nas (commented):
A few suggestions that should improve failure conditions readability.

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
jsync-swirlds: Timeout throws an error, it doesn't interrupt.
I agree on not throwing the runtime exception, but the goal here is minimal change to make CI reliable because this code is already deprecated.

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
jsync-swirlds: I agree on not throwing the runtime exception, but the goal here is minimal change to make CI reliable because this code is already deprecated.
If we want to change how exceptions are managed, I'd prefer to make that a separate PR.

block-node/s3-archive/src/test/java/org/hiero/block/node/archive/s3/S3ArchivePluginTest.java
ata-nas: `@Timeout` does in fact interrupt the test runner, hence why when tests hang, we see an InterruptedException thrown from a method that is blocking. Later, this results in a `TimeoutException` thrown by junit itself.

Not pushing for it, we can go with the runtime exception. Up to you.

jsync-swirlds: Let's leave for now, particularly as this code is going away soon.

ata-nas (dismissed):

Nana-EC (dismissed):
My bar went down where I realized which plugin this was on.

ata-nas (dismissed):

ivannov (dismissed):

ata-nas (approved):

rockysingh (approved):

PR #3341, opened 2026-07-31
fix: Added changes to improve stability in baseSuite.

jsync-swirlds:
## Reviewer Notes
* Adjusted BaseSuite.java to guarantee launched-container teardown in @AfterEach, fix a base-channel leak, and make teardownBlockNodes more resilient and idempotent.
    * This should prevent the "first error kills all subsequent tests" issue.
* Added a time-limited teardownEnvironment wait in PositiveMultiplePublishersTests.

### DIFF
--- a/tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
+++ b/tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
@@ -21,6 +21,7 @@ import org.hiero.block.api.protoc.BlockAccessServiceGrpc;
 import org.hiero.block.api.protoc.BlockNodeServiceGrpc;
 import org.hiero.block.api.protoc.BlockStreamPublishServiceGrpc;
 import org.hiero.block.simulator.BlockStreamSimulatorApp;
+import org.hiero.block.simulator.exception.BlockSimulatorParsingException;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.testcontainers.containers.Container;
@@ -61,8 +62,12 @@ public abstract class BaseSuite {
     /** Executor service for managing threads */
     protected static ErrorLoggingExecutor executorService;
 
-    /** gRPC channel for connecting to Block Node */
-    protected static ManagedChannel channel;
+    /**
+     * Every gRPC channel opened for the base Block Node during {@link #setup()}. Tracked so that all
+     * of them are shut down in {@link #teardown()} — previously only the last-created channel was
+     * closed, leaking the other two on every test in the suite's shared JVM.
+     */
+    private static final List<ManagedChannel> baseChannels = new ArrayList<>();
 
     /** Map from port to gRPC channel */
     protected Map<Integer, ManagedChannel> channels = new LinkedHashMap<>();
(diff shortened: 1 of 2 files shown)


tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -126,37 +130,72 @@ public void setup() {
     @AfterEach
     public void teardown() throws InterruptedException {
         if (blockNodeContainer != null) {
-            blockNodeContainer.stop();
-            blockNodeContainer.close();
+            try {
+                blockNodeContainer.stop();
+                blockNodeContainer.close();
+            } catch (final RuntimeException e) {
+                // best-effort: keep tearing the rest down so nothing leaks
```
Nana-EC, line 140: The catch is useful but hiding the underlying reason will make troubleshooting harder.
You should log the exception message

Nana-EC (commented):
LG, would recommend one log to not make future troubleshooting harder

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
jsync-swirlds: Troubleshooting a shutdown exception in a test suite?
We've been seeing those for months with full stack trace and nobody's ever tried to fix it.

If you mean _just_ the exception name and message (message alone is _often_ empty); then that might be useful in rare cases I suppose.

Exception name added.

Note also, we don't have loggers in test code, so I just printed to stderr.

berryware (dismissed):
Looks good

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -126,37 +131,76 @@ public void setup() {
...
+     * so one channel's shutdown can never skip the teardown of the remaining channels or containers.
+     *
+     * @param channelToClose the channel to shut down; ignored if {@code null}
+     */
+    private static void shutdownChannelQuietly(final ManagedChannel channelToClose) {
+        try {
+            if (channelToClose == null) {
+                return;
+            }
+            channelToClose.shutdown().awaitTermination(5, TimeUnit.SECONDS);
```
ata-nas, line 198: Should we catch a TimeoutException? If we timeout, we still want to try others.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -380,7 +422,7 @@ protected Future<?> startSimulatorInThread(BlockStreamSimulatorApp blockStreamSi
         return executorService.submit(() -> {
             try {
                 blockStreamSimulatorAppInstance.start();
-            } catch (Exception e) {
+            } catch (InterruptedException | BlockSimulatorParsingException | IOException e) {
```
ata-nas, line 425: If we have an interrupted exception, shouldn't we also reset the interrupted flag?

ata-nas (commented):
Couple of questions, otherwise looks good.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
jsync-swirlds: Not in this case.
All this change does is replace catching Exception with the same exceptions it was already catching (minus runtime exceptions that buried things like NPE).
The goal is minimum necessary change; not making the test code perfect.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
jsync-swirlds: Timeout Exception is not thrown here, the awaitTermination method just completes normally if it times out.

ata-nas (dismissed):

Nana-EC (dismissed):

ata-nas (dismissed):

ata-nas (dismissed):

rockysingh (dismissed):

ata-nas (approved):

Nana-EC (approved):

PR #3342, opened 2026-07-31
fix: Make Positive Multiple Publisher Tests less flakey

jsync-swirlds:
## Reviewer Notes
* Made changes to improve determinism.
* Added wait methods useable by any test: awaitCondition, awaitPublishedBlocks, awaitStatusContaining, awaitBlocksAvailable, awaitLatestBlockAtLeast
   * This generalized awaitBackfilledBlocks.
* Replaced every enabled test's fixed Thread.sleep/tight-spin with an await on the exact precondition its assertions need
* Two were tricky. The ack test now awaits the broadcast-ack, and the duplicate-block test awaits the node to persist past the skip window.
* Changed the the chained block-number dereferences to avoid NPE issues (no more getFirst() on a not-yet-available block)
* Limted the two start-helper waits
* Raised the timeout ceiling  from 30 to 60s with 90s on the two await-heavy tests.

### DIFF
--- a/tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
+++ b/tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
@@ -16,12 +16,16 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import io.grpc.stub.StreamObserver;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.locks.LockSupport;
+import java.util.function.BooleanSupplier;
 import java.util.stream.Stream;
 import org.hiero.block.api.protoc.BlockAccessServiceGrpc.BlockAccessServiceBlockingStub;
 import org.hiero.block.api.protoc.BlockResponse;
@@ -48,7 +52,9 @@ import org.junit.jupiter.params.provider.MethodSource;
  */
 // TODO(#1665) Verify simulator's connection has been closed for multi-publisher scenarios
 @DisplayName("Positive Multiple Publishers Tests")
-@Timeout(30)
+// Ceiling only: every test below waits on an observed condition (see the await* helpers), so a
+// green run finishes in seconds; this bound just fails a genuinely stuck test with a clear message.
+@Timeout(60)
 public class PositiveMultiplePublishersTests extends BaseSuite {
 
     private final List<Future<?>> simulators = new ArrayList<>();
@@ -65,7 +71,7 @@ public class PositiveMultiplePublishersTests extends BaseSuite {
                 // Bound the stop-wait: a simulator that never reports stopped must not hang teardown.
                 final long deadline = System.currentTimeMillis() + 5_000;
                 while (simulator.isRunning() && System.currentTimeMillis() < deadline) {
-                    Thread.sleep(100);
+                    LockSupport.parkNanos(100_000_000);
                 }
             } catch (InterruptedException e) {
                 // Do nothing, this is not mandatory, we try to shut down  cleaner and graceful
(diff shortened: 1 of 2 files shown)


tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
```diff
@@ -545,27 +614,52 @@ public void testSettledWrapsAcknowledgements() throws IOException, InterruptedEx
...
+
+        // Snapshot the fast publisher's acknowledgements, then wait until the (slower) second publisher
+        // has been sent the same ones. The node broadcasts each acknowledgement to every connected
+        // publisher, so this converges deterministically instead of relying on a fixed sleep and a
+        // lucky sampling instant.
+        final List<String> firstAcknowledgements =
+                firstSimulator.getStreamStatus().lastKnownPublisherClientStatuses().stream()
+                        .filter(status -> status.toLowerCase().contains("acknowledgement"))
+                        .toList();
+        awaitCondition(
```
Nana-EC, line 638: There's a no-op when moving form teh firs tto second publisher.
We should check on teh firstAcknowledgemnts.
maybe something like

```suggestion
        assertFalse(
          firstAcknowledgements.isEmpty(),
          "Expected at least one acknowledgement in snapshot immediately after awaitStatusContaining");
        awaitCondition(
```

tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
```diff
@@ -192,31 +199,93 @@ public void testAutonomousBackfill() throws IOException, InterruptedException {
...
-                    allPresent = false;
-                    break;
-                }
-            }
-            if (allPresent) return;
-            if (System.currentTimeMillis() >= deadlineMs) {
-                throw new AssertionError("Backfill did not complete within 60 s; blocks still missing from node");
+        final long deadline = System.currentTimeMillis() + timeoutMs;
+        while (!condition.getAsBoolean()) {
+            if (System.currentTimeMillis() >= deadline) {
```
Nana-EC, line 220: You should probably invert the check and check for deadline first before condition.
In this case the condition is a call so in some cases with tighter timeouts you could cause unexpected behaviour

Nana-EC (commented):
Nice improvements, some suggestions

tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
jsync-swirlds: The whole change was to wait on conditions _instead of_ tight timeouts.
Tight timeouts are flakey; conditions are much more reliable and generally run faster than the timeout would have.

Granted these condition waits are still polling due to limitations in multi-process testing, but it's still an improvement over a blind wait.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
jsync-swirlds: We already waited until there is at least one acknowledgement for the first publisher on line 632.

I'm not sure what this extra assertion adds.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
```diff
@@ -65,7 +71,7 @@ void teardownEnvironment() {
                 // Bound the stop-wait: a simulator that never reports stopped must not hang teardown.
                 final long deadline = System.currentTimeMillis() + 5_000;
                 while (simulator.isRunning() && System.currentTimeMillis() < deadline) {
-                    Thread.sleep(100);
+                    LockSupport.parkNanos(100_000_000);
```
ata-nas, line 74: Nit: the lock support values can be static fields as some are reused. Also, for readability, we can consider to construct those fields as `TimeUnit.<UNIT>.toNanos(<duration>);`

ata-nas (dismissed):
Good improvements, just a nit from my side.

Nana-EC (dismissed):

ata-nas (dismissed):

tools-and-tests/suites/src/main/java/org/hiero/block/suites/publisher/positive/PositiveMultiplePublishersTests.java
jsync-swirlds: Will add to follow-up.

ata-nas (dismissed):

rockysingh (dismissed):

Nana-EC (dismissed):

Nana-EC (approved):

ata-nas (approved):

PR #3353, opened 2026-08-03
feat: per-pod chaos selector — asymmetric BN latency injection (#3348)

### LINKED ISSUE #3348: Chaos injection targets all pods of a type — no per-pod selector support
## Description

The `inject-latency` event in the solo E2E test runner applies chaos to **all pods of a given type** simultaneously. The label selector is:

```bash
network-node) echo "solo.hedera.com/type=network-node" ;;
block-node)   echo "block-node.hiero.com/type=block-node" ;;
```

`solo-test-runner.sh`, `chaos_label_selector()`, line ~524.

A test YAML entry like:

```yaml
- type: inject-latency
  args:
    source: { kind: block-node }
    target: { kind: block-node }
    latency: 1000ms
```

results in a NetworkChaos resource that selects every pod with `block-node.hiero.com/type=block-node`. In a `paired-3` or `7cn-3bn-distributed` topology this means BN1, BN2, and BN3 all receive the same latency rule. There is no mechanism to target a single BN pod.

**Impact:** It is impossible to create asymmetric chaos between individual BNs or CN/BN pairs. This means:
- The `backfill-triggered` assertion can never be satisfied (all BNs slow equally, no divergence)
- Geographic simulation (different latencies representing different data-centre locations) cannot be expressed
- Failure injection on a single node cannot be done via latency rules

## Steps to Reproduce

Inspect any `latency-*.yaml` test that uses `source: { kind: block-node }`. Every BN in the cluster gets the same netem rule applied.

## Possible Solution

Extend the `source`/`target` YAML schema with an optional `name` field:

```yaml
source: { kind: block-node, name: block-node-1 }
target: { kind: block-node, name: block-node-2 }
```

In `execute_inject_latency`, when `name` is present, add a second label match on `app.kubernetes.io/name=<name>` alongside the type label. Chaos Mesh `NetworkChaos` supports compound label selectors natively.

This is a small change to `solo-test-runner.sh` and the test YAML schema validation, but unblocks geographic simulation scenarios and the `backfill-triggered` use case.

Part of epic #3346

Nana-EC:
## Problem

Chaos tests injected identical latency across all block nodes simultaneously.
Because every BN was equally degraded, no divergence occurred — the tests were
verifying nothing meaningful about single-node resilience.

Four additional bugs meant chaos was silently never applied at all:

1. **`envsubst` missing on CI runner** — chaos manifest template substitution
   produced empty files; `kubectl apply` on an empty file returned no error.
2. **Loss field format** — test YAMLs specified `loss: "50%"` but Chaos Mesh
   requires a plain number string (`"50"`); the admission webhook rejected the
   manifest and the event was still marked PASS.
3. **Silent `kubectl apply` failure** — return code was not checked, so rejected
   manifests caused chaos to silently not apply.
4. **BackfillPlugin LIVE_TAIL (`greedy: true`)** — even when chaos did apply,
   backfill polled unchaosed peers every ~60s and pulled missing blocks before
   the mid-chaos snapshot, collapsing any spread to zero.

### Infrastructure fixes
- **`.github/workflows/solo-e2e-test.yml`** — install `gettext-base` so `envsubst`
  is available on the runner.
- **`solo-test-runner.sh` / `execute_inject_latency`** — strip trailing `%` from
  loss values; fail fast if `kubectl apply` returns non-zero.

### Per-pod instance selector
- **`network-latency.yaml.tmpl`** — added `${SOURCE_NAME_FILTER}` /
  `${TARGET_NAME_FILTER}` placeholder lines in each `labelSelectors` block.
  When set they inject `app.kubernetes.io/instance: <name>` at the correct
  indentation. When empty the line is blank — existing tests unaffected.
- **`solo-test-runner.sh`** — extracts optional `source.name` / `target.name`
  from event args; builds compound dryrun selector for pod-count validation;
  exports the two filter env vars.

### Mid-chaos measurement
- **`snapshot-block-heights`** event — queries `lastAvailableBlock` on every BN
  mid-chaos and writes a snapshot file.
- **`blocks-diverged`** assertion — reads the snapshot; fails if
  `max − min < min_spread`. Proves chaos produced detectable divergence.
- **`blocks-converged`** assertion — queries live heights at assertion time;
  fails if spread exceeds `tolerance_blocks`. Proves streams survived and nodes
  caught up.

### Topology change
- **`paired-3`**, **`7cn-3bn-distributed`** — all BNs set to `greedy: false`
  to prevent BackfillPlugin from pulling blocks over the unchaosed peer path
  and masking divergence. See #3379 for the bandwidth-throttle follow-on that
  restores `greedy: true`.

### Test definitions

**`latency-stress`** — survival test. One BN (BN1) receives high latency +
moderate loss. At this tier TCP throughput still exceeds the block data rate
so lag is constant (~2.5 blocks), not cumulative; divergence is not asserted.
Assertions confirm BN1 stays healthy, keeps receiving blocks, and converges
after chaos clears.

| Pod | CN↔BN latency | BN→BN peers |
|-----|--------------|-------------|
| BN1 | 5000 ms ± 1250 ms + 50% loss, bidirectional | 2500 ms ± 625 ms, direction=to |
| BN2/3 | — | — |
| CNs | 100 ms ± 20 ms (background) | — |

**`latency-severe`** — divergence test. Three-tier gradient; BN1 most degraded,
BN2 mid, BN3 reference. At 8000 ms + 75% correlated loss, burst-loss RTO chains
freeze the CN↔BN1 stream. Spread of ~39 blocks observed consistently.

| Pod | CN↔BN latency | BN→BN peers |
|-----|--------------|-------------|
| BN1 | 8000 ms ± 2000 ms + 75% loss (corr 75%), bidirectional | 4000 ms ± 1000 ms, direction=to |
| BN2 | 4000 ms ± 1000 ms + 40% loss (corr 50%), bidirectional | 2000 ms ± 500 ms, direction=to |
| BN3 | — (reference) | — |
| CNs | 1000 ms ± 250 ms (corr 75%), bidirectional | — |

Convergence tolerance is wide (150 blocks) at this tier because the CN↔BN1
stream is severed and BN1 does not reconnect within the test window.
`blocks-increasing` and `block-rate-floor` target BN3 only.

## Known limitation / follow-on

`latency-stress` has no block-divergence assertion because the current
tc-netem approach cannot produce cumulative lag when block payloads are
small (~2 KB). **#3379 / #3387** (bandwidth throttle via `tc tbf`) will cap
ingest below block production rate, creating sustained lag with `greedy: true`.

## Test results

| Run | Test | Outcome | Spread |
|-----|------|---------|--------|
| [31114793205](https://github.com/hiero-ledger/hiero-block-node/actions/runs/31114793205) | latency-severe | ✅ PASS | 39 |
| [31105361075](https://github.com/hiero-ledger/hiero-block-node/actions/runs/31105361075) | latency-stress | ✅ PASS | — (survival) |

Closes #3348

### DIFF
--- a/tools-and-tests/scripts/solo-e2e-test/tests/latency-severe.yaml
+++ b/tools-and-tests/scripts/solo-e2e-test/tests/latency-severe.yaml
@@ -182,22 +182,22 @@ assertions:
 
   - id: blocks-converged
     type: blocks-converged
-    description: "All BNs converge to within 5 blocks of each other after latency clears — proves CN↔BN streams survived across the full gradient and all nodes caught up"
+    description: "All BNs within 150 blocks of each other after latency clears — at severe tier the CN↔BN1 stream may die; tolerance is wide enough to detect a completely frozen node while accepting normal catchup lag"
     args:
-      tolerance_blocks: 5
+      tolerance_blocks: 150
 
   - id: blocks-increasing
     type: blocks-increasing
-    description: "Block production recovers on all BNs after latency clears"
-    target: all
+    description: "Block production continues on reference node (BN3) after latency clears — BN1 stream may not recover at this severity tier"
+    target: block-node-3
     args:
       wait_seconds: 60
       max_attempts: 5
 
   - id: block-rate-floor
     type: block-rate-floor
-    description: "Post-chaos block rate recovers to ≥ 0.1 blk/s on all BNs"
-    target: all
+    description: "Post-chaos block rate on reference node (BN3) is ≥ 0.1 blk/s — confirms network liveness even if BN1 stream was severed"
+    target: block-node-3
     args:
       min_rate_per_sec: 0.1
       window_seconds: 30
--- a/tools-and-tests/scripts/solo-e2e-test/tests/latency-severe.yaml
+++ b/tools-and-tests/scripts/solo-e2e-test/tests/latency-severe.yaml
@@ -2,27 +2,26 @@
 #
 # Severe ("breaking-point") CN↔BN Latency — Geo-Asymmetric Gradient Test
 #
-# Three-tier per-BN latency that forces real divergence between BNs so that
-# backfill fires reliably:
+# Three-tier per-BN latency across all BNs:
 #
 #   BN1  — "very far"  : CN↔BN1 800 ms ± 200 ms + 25% loss (corr 75%), BN1→BN 1000 ms (direction=to)
 #   BN2  — "mid-range" : CN↔BN2 400 ms ± 100 ms + 15% loss (corr 75%), BN2→BN  500 ms (direction=to)
-#   BN3  — "local"     : no CN↔BN or BN→BN chaos  (fastest; BN1/2 backfill from it)
+#   BN3  — "local"     : no CN↔BN or BN→BN chaos  (fastest)
 #   CNs  — global      : CN↔CN 600 ms ± 150 ms, corr 75%  (bidirectional)
 #
-# Packet loss on CN↔BN1/2 causes bursty correlated drops that break gRPC
-# streams, forcing reconnects during which BN1 (and BN2) miss blocks →
-# BackfillPlugin detects sequence gaps and fetches missing blocks from BN3.
+# BN→BN rules use direction=to (unidirectional) so the per-BN rules for BN1
+# and BN2 do not additively stack on the BN1↔BN2 path.
 #
-# BN→BN rules use direction=to (unidirectional) so the two per-BN rules do not
-# additively stack on the BN1↔BN2 path.
+# Intent: exercises the full asymmetric degradation profile — BN3 is fastest,
+# BN2 is degraded, BN1 is most degraded. Validates that all three BNs stay
+# healthy and continue receiving blocks despite severe asymmetric latency.
 #
-# Intent: BN3 accumulates blocks fastest; BN2 falls behind; BN1 falls furthest
-# behind. After chaos clears, BN1 (and possibly BN2) must backfill missing
-# blocks from BN3, exercising the BackfillPlugin gap-fill path.
+# NOTE: TCP on kind cluster loopback absorbs packet loss via retransmit; the
+# gRPC stream from CN never breaks, so no block-sequence gap forms in persistent
+# storage. BackfillPlugin gap-fill testing requires explicit storage gap injection
+# (see the dedicated backfill-recovery test, tracked in #3349).
 #
-# Requires topology with ≥ 3 BNs that have backfill peers configured
-# (e.g. paired-3, 7cn-3bn-distributed). The dryrun check will fail on 3cn-1bn.
+# Requires topology with ≥ 3 BNs (e.g. paired-3, 7cn-3bn-distributed).
 #
 # Pre-req:
 #   task up TOPOLOGY=paired-3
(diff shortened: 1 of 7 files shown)


tools-and-tests/scripts/solo-e2e-test/tests/latency-severe.yaml
```diff
@@ -38,99 +54,150 @@ events:
...
   - id: block-rate-floor
     type: block-rate-floor
-    description: "Post-chaos block rate recovers to ≥ 0.1 blk/s (recovery, not steady-state)"
-    target: all
+    description: "Post-chaos block rate on reference node (BN3) is ≥ 0.1 blk/s — confirms network liveness even if BN1 stream was severed"
+    target: block-node-3
     args:
       min_rate_per_sec: 0.1
-      window_seconds: 30
+      window_seconds: 30
```
jsync-swirlds, line 203: Missing newline at EOL.

tools-and-tests/scripts/solo-e2e-test/tests/latency-stress.yaml
```diff
@@ -38,60 +56,60 @@ events:
...
-      name: bn-to-bn
-      source: { kind: block-node }
-      target: { kind: block-node }
-      latency: 600ms
-      jitter: 150ms
+      name: cn-to-bn1
+      source: { kind: network-node }
+      target: { kind: block-node, name: block-node-1 }
+      latency: 5000ms
+      jitter: 1250ms
```
jsync-swirlds, line 78: This is about 10X the highest we see in any real network.
Are we sure we want this much?

tools-and-tests/scripts/solo-e2e-test/tests/latency-stress.yaml
```diff
@@ -38,60 +56,60 @@ events:
...
     args:
-      name: cn-to-bn
-      source: { kind: network-node }
+      name: bn1-to-bn
+      source: { kind: block-node, name: block-node-1 }
       target: { kind: block-node }
-      latency: 450ms
-      jitter: 120ms
+      latency: 2500ms
+      jitter: 625ms
```
jsync-swirlds, line 92: This is about 10X the longest _typical_ latency we've observed in any real network.

jsync-swirlds (commented):
The high latency options seem a bit too high (around 10X any real-world latency that doesn't involve deep-space radio links), in my opinion.

tools-and-tests/scripts/solo-e2e-test/tests/latency-stress.yaml
Nana-EC: With latency only this is how far I had to go to get things cracking.
I have a future PR where i'm looking to add bandwidth constraints that will likely help relax these.

tools-and-tests/scripts/solo-e2e-test/tests/latency-stress.yaml
Nana-EC: With latency only this is how far I had to go to get things cracking.
I have a future PR where i'm looking to add bandwidth constraints that will likely help relax these.

tools-and-tests/scripts/solo-e2e-test/tests/latency-stress.yaml
jsync-swirlds: I'm not sure why we need to force errors, typically I would expect to have a target value that represents a goal state, and verify that the system always works correctly up to that target.

Perhaps we just need better documentation explaining why we are injecting values this far outside of any realistic range.

jsync-swirlds (approved):
Approving to get testing in place, but the extreme latency involved here, compared to larger test environments, suggests we need to look at more than just latency to find the proper edges.

Nana-EC:
> Approving to get testing in place, but the extreme latency involved here, compared to larger test environments, suggests we need to look at more than just latency to find the proper edges.

Yep, already in motion - https://github.com/hiero-ledger/hiero-block-node/pull/3387
Will continue to explore additional items.
Also captured your thoughts in https://github.com/hiero-ledger/hiero-block-node/issues/3503

rockysingh (approved):

PR #3359, opened 2026-08-03
fix(docs): repair broken relative links, rename configuration overview, and add lychee CI

deshmukhpranali:
## Summary

Three related doc-infrastructure fixes in one PR.

### Broken relative links (5 files)

Pre-existing path bugs causing 404s on docs.hiero.org:

- **Cutover-Process.md** - 5 links: `../glossary.md#…` → `./glossary.md#…`
- **glossary.md** - 6 links: `./../../Block-Node-Types.md` → `../Block-Node-Types.md`
- **faq/operator-faq.md** - 32 links: missing `../` prefix throughout
- **faq/integration-faq.md** - 12 links: missing `../` prefix throughout
- **block-node/quickstart.md** - updated to point at renamed file (see below)

### Rename: `docs/configuration.md` → `docs/configuration-overview.md`

The hiero-docs hub maintains a GitBook card-table landing page at `block-node/configuration.md`.
The sync workflow copies `docs/` verbatim into `block-node/`, so any file with the same name in
both repos is silently overwritten on every sync - destroying the card table.

Renaming the prose overview to `configuration-overview.md` eliminates this collision: the sync
produces `block-node/configuration-overview.md` while `block-node/configuration.md` (the card
table) is left untouched. The detailed configuration reference at `docs/block-node/configuration.md`
is unchanged.

### Lychee link-check CI

Two jobs added in `.github/workflows/link-check.yml`:

- **Internal links** (blocking) - `lychee --offline` on all `docs/**/*.md` on every doc-touching PR
- **External links** (advisory) - checks live URLs without blocking merges

`.lychee.toml` excludes paths with deliberate source-code links that can't be validated offline
(`docs/design/`, `docs/test-plan/`, `docs/block-node/architecture/plugins.md`).

`CLAUDE.md` documents the link integrity rule and sync architecture for future contributors.


.github/workflows/link-check.yml
```diff
@@ -0,0 +1,63 @@
+# SPDX-License-Identifier: Apache-2.0
+name: Link Check
+
+on:
```
Nana-EC, line 4: Q: How heavy (resources and time) is this CI run? 
If it's not minor should we have this run weekly or only on release branch?
Docs aren't updated as often (after this big push of course) so i'm exploring if these should be on every push and PR

.github/workflows/link-check.yml
```diff
@@ -0,0 +1,63 @@
...
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Check internal links
+        uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
+        with:
+          args: >-
+            --config .lychee.toml
+            --offline
+            --no-progress
+            'docs/**/*.md'
```
Nana-EC, line 41: Q: there's some docs outside of `/docs`, should we cover those also or if this focused on the docs that end up on docs.hiero?

CLAUDE.md
```diff
@@ -0,0 +1,59 @@
+# hiero-block-node workspace guidance
```
Nana-EC: Please drop this Claude.md file.
I'm thinking on if and what should be in here. I can factor this into it. Thanks

.github/workflows/link-check.yml
deshmukhpranali: @Nana-EC the docs/**/*.md scope is intentional - those are the only files synced to docs.hiero.org. There are developer tool docs under tools-and-tests/tools/docs/, but they're not user-facing published docs, so we've kept them out of scope.

.github/workflows/link-check.yml
deshmukhpranali: @Nana-EC good point. The internal-links job runs lychee --offline (file-existence check only, no network, ~30 seconds) and is gated by paths: ["docs/**/*.md"], so it only fires on doc-touching PRs - that one is fine as-is.

The external-links job does make live HTTP requests and is the heavier one. Based on your feedback, I've updated the workflow: external-links now runs on a weekly schedule only instead of on every PR, and internal-links is skipped on the scheduled run.` workflow_dispatch` still triggers both manually if needed.

.github/workflows/link-check.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Check external links
+        uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
+        with:
+          args: >-
+            --config .lychee.toml
+            --no-progress
+            'docs/**/*.md'
+          fail: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jsync-swirlds: Is this necessary?  It's not present for "internal links", so I wonder if it's actually required here.

jsync-swirlds (commented):
Needs a rebase for CI fixes.

If the `printHelp` test fails again just rerun, that test is currently very flakey.

.github/workflows/link-check.yml
```diff
@@ -0,0 +1,65 @@
...
+            --config .lychee.toml
+            --offline
+            --no-progress
+            'docs/**/*.md'
+          fail: true
+
+  external-links:
+    name: External links (advisory)
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
```
rbarker-dev: Needs to use the block node self-hosted runners.

.github/workflows/link-check.yml
```diff
@@ -0,0 +1,65 @@
...
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  internal-links:
+    name: Internal links
+    if: github.event_name != 'schedule'
+    runs-on: ubuntu-latest
```
rbarker-dev: Needs to use the block node self-hosted runners.

rbarker-dev (changes requested):

rbarker-dev (approved):

jsync-swirlds (approved):

jsync-swirlds:
FYI, Lychee is not entirely compatible with CI infrastructure as-is.
The lychee build we're loading requires a glibc newer than the one we are able to run (due to upgrade constraints related to supporting customer O/S requirements).

Nana-EC (approved):
LG
Likely a follow up need but let's fix the current links

PR #3360, opened 2026-08-03
fix: Increase performance of serverStatus and serverStatusDetails gRpc calls.

### LINKED ISSUE #3004: todo: Change ServerStatusDetail to use BlockNodeContext.availableBlocks()
### Story Form

As a BlockNode engineer
I want to use the source of truth for the available blocks known to the BlockNode server
So that the block node server reports available blocks accurately.

### Technical Notes

`ServerStatusDetail` currently retrieves the available blocks from the `HistoricalBlockFacility`. The available blocks reported to the `ApplicationStateFacility` are not yet the complete picture. This is a placeholder to go back and update the `ServerStatusDetail` when the `ApplicationStateFacility` represents the source of truth.

### LINKED ISSUE #3306: ServerStatus Concurrent High Load Support
### Story Form

As a block node operator
I want to expose a serverStatus endpoint with high tolerance
So that it can manage multiple MNs at high TPS

### Technical Notes

ServerStatus is supposed to be light weight and quickly respond with a BNs status of available blocks.
Let's do some perf testing to see and refine support.
~20 MNs is a good start

berryware:
## Reviewer Notes

**`serverStatusDetail`:** Replaced the per-request loop that called `context.historicalBlockProvider().availableBlocks().streamRanges().toList()` and rebuilt a `List<BlockRange>` by hand. That path triggered `CombinedBlockRangeSet.reduceRanges()`, which reconstructs and re-merges a fresh `ConcurrentLongRangeSet` on every call. It now reads the already-merged, periodically-refreshed `context.availableBlocks()` snapshot — the same source already used for `storedRanges`, `tssData`, etc.  

One thing to flag: `context.availableBlocks()` is refreshed on the app-state scan interval (default 500ms), so `availableRanges` is now eventually-consistent (up to one scan interval stale) rather than live-at-request-time. Given every sibling field in this response was already snapshot-based, that's consistent with the existing design.

**`serverStatus`:** Captured `blockProvider.availableBlocks()` into a single local so `min()` and `max()` read from one reference instead of two separate calls. This is a consistency tidy-up, not a meaningful speedup — that call was already O(1)-ish.

## Related Issue(s)
Closes #3004.
Closes #3306.

### DIFF
--- a/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
+++ b/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
@@ -108,10 +108,11 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
         // serverStatus has the latest max available block. serverStatusDetail has an up to .5s old
         // snapshot. This will align the lastBlock range end with what serverStatus would report without
         // having to rebuild the entire available Blocks list.
+        final long liveMax = blockProvider.availableBlocks().max();
         List<BlockRange> fixedAvailable = !context.availableBlocks().isEmpty()
-                        && context.availableBlocks().getLast().rangeEnd()
-                                != blockProvider.availableBlocks().max()
-                ? fixAvailable(context.availableBlocks())
+                        && liveMax != UNKNOWN_BLOCK_NUMBER
+                        && context.availableBlocks().getLast().rangeEnd() != liveMax
+                ? fixAvailable(context.availableBlocks(), liveMax)
                 : context.availableBlocks();
 
         // Return detailed block node status information. Every field is read from the
@@ -130,18 +131,18 @@ public class ServerStatusServicePlugin implements BlockNodePlugin, BlockNodeServ
 
     /**
      * Returns a copy of {@code availableBlocks} whose last range end is aligned with the live
-     * {@code blockProvider.availableBlocks().max()}. A copy is returned because the supplied list is
-     * the shared, immutable context snapshot (see {@code ApplicationStateUtility.toBlockRange}); it
-     * must never be mutated in place.
+     * {@code liveMax}. A copy is returned because the supplied list is the shared, immutable context
+     * snapshot (see {@code ApplicationStateUtility.toBlockRange}); it must never be mutated in place.
      *
      * @param availableBlocks the context snapshot of available ranges (never empty)
+     * @param liveMax the live maximum available block number to align the last range end with
      * @return a new list with the last range end aligned to the live max
      */
-    private List<BlockRange> fixAvailable(final List<BlockRange> availableBlocks) {
+    private List<BlockRange> fixAvailable(final List<BlockRange> availableBlocks, final long liveMax) {
         final BlockRange lastBlockRange = availableBlocks.getLast();
         final BlockRange newLastBlockRange = BlockRange.newBuilder()
                 .rangeStart(lastBlockRange.rangeStart())
-                .rangeEnd(blockProvider.availableBlocks().max())
+                .rangeEnd(liveMax)
                 .build();
 
         final List<BlockRange> aligned = new ArrayList<>(availableBlocks);
--- a/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
+++ b/block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
@@ -5,6 +5,7 @@ import static java.lang.System.Logger.Level.TRACE;
 import static java.util.Objects.requireNonNull;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
 import java.util.List;
 import org.hiero.block.api.BlockNodeServiceInterface;
 import org.hiero.block.api.BlockRange;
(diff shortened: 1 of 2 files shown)


block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -101,39 +103,52 @@ public ServerStatusDetailResponse serverStatusDetail(@NonNull final ServerStatus
...
                 .nodeAddressBook(context.nodeAddressBook())
                 .rangedAddressBookHistory(context.rangedAddressBookHistory())
                 .build();
     }
 
+    /**
+     * Returns a copy of {@code availableBlocks} whose last range end is aligned with the live
+     * {@code blockProvider.availableBlocks().max()}. A copy is returned because the supplied list is
+     * the shared, immutable context snapshot (see {@code ApplicationStateUtility.toBlockRange}); it
+     * must never be mutated in place.
```
jsync-swirlds: That snapshot probably _can_ be modified in place; we just try to avoid it because too much code assumes it doesn't change.

jsync-swirlds (dismissed):
Just a minor comment, this is good work.

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -101,39 +103,52 @@ public ServerStatusDetailResponse serverStatusDetail(@NonNull final ServerStatus
...
-        return detailsBuilder
-                .availableRanges(blockRanges)
-                // @todo(3004) change to use context.availableBlocks() when that becomes the source of truth
-                // .availableRanges(context.availableBlocks())
+        final BlockNodeContext context = blockNodeContext;
+
+        // serverStatus has the latest max available block. serverStatusDetail has an up to .5s old
+        // snapshot. This will align the lastBlock range end with what serverStatus would report without
+        // having to rebuild the entire available Blocks list.
+        List<BlockRange> fixedAvailable = !context.availableBlocks().isEmpty()
```
Nana-EC, line 112: Might be worth capturing a `liveMax` once and guarding against accidentally setting the sentinel in the case the live provider becomes empty (node restart, pruning gap)
```suggestion
        final long liveMax = blockProvider.availableBlocks().max();
        List<BlockRange> fixedAvailable = !context.availableBlocks().isEmpty() && liveMax != UNKNOWN_BLOCK_NUMBER
```

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -101,39 +103,52 @@ public ServerStatusDetailResponse serverStatusDetail(@NonNull final ServerStatus
...
+    /**
+     * Returns a copy of {@code availableBlocks} whose last range end is aligned with the live
+     * {@code blockProvider.availableBlocks().max()}. A copy is returned because the supplied list is
+     * the shared, immutable context snapshot (see {@code ApplicationStateUtility.toBlockRange}); it
+     * must never be mutated in place.
+     *
+     * @param availableBlocks the context snapshot of available ranges (never empty)
+     * @return a new list with the last range end aligned to the live max
+     */
+    private List<BlockRange> fixAvailable(final List<BlockRange> availableBlocks) {
```
Nana-EC: Follow up to previous suggestion
```suggestion
    private List<BlockRange> fixAvailable(final List<BlockRange> availableBlocks, final long liveMax) {
```

Nana-EC: Claude suggested test case
```
@Test
@DisplayName("Should not corrupt last range when live provider is empty")
void shouldReturnSnapshotUnchangedWhenLiveProviderIsEmpty() throws ParseException {
    // Set context snapshot to a non-empty list
    replaceAvailableBlocks(List.of(new BlockRange(0L, 5L)));
    // Simulate live provider empty — set temporaryAvailableBlocks to empty set
    historicalBlockFacility.setTemporaryAvailableBlocks(new SimpleBlockRangeSet());

    toPluginPipe.onNext(ServerStatusRequest.PROTOBUF.toBytes(
            ServerStatusRequest.newBuilder().build()));
    assertEquals(1, fromPluginBytes.size());

    final ServerStatusDetailResponse response =
            standardParse(ServerStatusDetailResponse.PROTOBUF, fromPluginBytes.getFirst());

    // Snapshot returned unchanged — no rangeEnd = -1 corruption
    final List<BlockRange> ranges = response.availableRanges();
    assertEquals(1, ranges.size());
    assertEquals(0L, ranges.getFirst().rangeStart());
    assertEquals(5L, ranges.getFirst().rangeEnd());  // NOT -1
}
```

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
```diff
@@ -101,39 +103,52 @@ public ServerStatusDetailResponse serverStatusDetail(@NonNull final ServerStatus
...
+        final BlockRange lastBlockRange = availableBlocks.getLast();
+        final BlockRange newLastBlockRange = BlockRange.newBuilder()
+                .rangeStart(lastBlockRange.rangeStart())
+                .rangeEnd(blockProvider.availableBlocks().max())
```
Nana-EC: Use input and avoids double call (here and caller)
```suggestion
                .rangeEnd(liveMax)
```

Nana-EC (dismissed):
LG. Some suggestions

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: added

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: added

berryware: Added

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: added

block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java
berryware: added

jsync-swirlds (approved):

Nana-EC (approved):

PR #3362, opened 2026-08-03
chore(tools): add bn-block-capture and find-state-proof-block scripts

AlfredoG87:
Adds two operational scripts under `tools-and-tests/scripts/`:

- **bn-block-capture** — Python tooling to download blocks from a Block Node at TPS plateaus and store them as `<n>.blk.zstd` files. Includes a profiler, bulk downloader, and verifier. Uses `grpcurl` instead of native `grpcio` (Helidon rejects the Python HTTP/2 SETTINGS frame with GOAWAY).
- **find-state-proof-block.sh** — Bash script that scans a Block Node in parallel to find the first block carrying a `BlockStateProof`. Workers retry on transient failures with exponential backoff; the winning hit is re-verified before being reported.

### DIFF
--- a/tools-and-tests/scripts/bn-block-capture/README.md
+++ b/tools-and-tests/scripts/bn-block-capture/README.md
@@ -1,0 +1,136 @@
+# bn-block-capture
+
+Python tooling to capture block recordings from a Hiero Block Node and store them as standard `<n>.blk.zstd` files (binary `Block.SerializeToString()` zstd-compressed — bit-exact to what BNs write to disk).
+
+Designed for sampling blocks at known TPS plateaus (idle, 100, 1.5k, 7k, 10k, ...) and around the Schnorr -> WRAPS TSS signature transition.
+
+## Why grpcurl, not native Python gRPC
+
+Helidon-based Block Nodes reject Python `grpcio`'s HTTP/2 settings frame with `GOAWAY error code 1` immediately after the SETTINGS exchange. `grpcurl` (Go-based) works fine, so `bn_client.py` shells out to `grpcurl`, parses the JSON response, and re-serializes through the generated protobuf classes. The on-disk bytes are still identical to a native binary fetch.
+
+## Prerequisites
+
+- `python3` (>=3.8) with `venv`
+- [`grpcurl`](https://github.com/fullstorydev/grpcurl) on PATH (`brew install grpcurl`)
+- Run from a checkout of the Hiero Block Node repo (the script uses `../../../protobuf-sources/`)
+
+## Setup (one-time)
+
+```
+cd tools-and-tests/scripts/bn-block-capture
+./setup.sh
+```
+
+This creates `.venv/`, installs Python deps, and compiles all `.proto` files into `proto-py/`. Re-run after proto changes.
+
+## Scripts
+
+|     Script     |                                              Purpose                                              |
+|----------------|---------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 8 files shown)


tools-and-tests/scripts/find-state-proof-block.sh
```diff
@@ -0,0 +1,418 @@
...
+  local raw
+  raw=$(grpcurl -plaintext \
+    -import-path "${API_PROTO_DIR}" \
+    -import-path "${STREAM_PROTO_DIR}" \
+    -proto "block-node/api/block_access_service.proto" \
+    -max-msg-sz "${MAX_MSG_SIZE}" \
+    -d "{\"block_number\": ${block_number}}" \
+    "${BN_ENDPOINT}" org.hiero.block.api.BlockAccessService/getBlock 2>&1) || true
+
+  if echo "${raw}" | grep -q '"status": *"NOT_FOUND"\|"status": *"NOT_AVAILABLE"'; then
```
Nana-EC, line 173: Use the extended regex to increase support across platforms
```suggestion
  if echo "${raw}" | grep -qE '"status": *"NOT_FOUND"|"status": *"NOT_AVAILABLE"'; then
```

tools-and-tests/scripts/bn-block-capture/profiler.py
```diff
@@ -0,0 +1,226 @@
...
+            done += 1
+            if done % 10 == 0 or done == total:
+                print(f"  [{done}/{total}]", file=sys.stderr)
+    return results
+
+
+def compute_tps(rows: list) -> list:
+    """Annotate rows with tps (dt to next sample). Rows must be sorted by block_number."""
+    rows_sorted = sorted([r for r in rows if r.get("status") == "OK"], key=lambda r: r["block_number"])
+    by_num = {r["block_number"]: r for r in rows_sorted}
```
Nana-EC, line 72: unused, can be removed

tools-and-tests/scripts/bn-block-capture/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Troubleshooting
+
+- **`GOAWAY received; Error code: 1`** — you're hitting native `grpcio` instead of `grpcurl`. Make sure you're invoking the scripts in this folder, which always shell out to `grpcurl`.
+- **`malformed header: missing HTTP content-type`** — server doesn't speak gRPC reflection. Use the explicit `-proto` flag (which the scripts do).
+- **`NOT_AVAILABLE` on a few blocks in a 1,000-block run** — transient. Just re-run the same command; the downloader is resumable.
+- **Block 361 (or whichever the WRAPS transition is) fails with timeout** — that block is ~95 MB. Bump `--timeout 240` and run with `--parallel 1`.
+- **`platform' is not a package`** — `setup.sh` did not rename the generated `platform/` package to `_bn_platform/`. Re-run `./setup.sh`.
+
+See `.claude/commands/bn-backup.md` for the higher-level workflow doc that wraps this README.
```
Nana-EC, line 136: left over form your local

tools-and-tests/scripts/bn-block-capture/download.py
```diff
@@ -0,0 +1,119 @@
+"""Bulk-download blocks from a Block Node and store as ``<n>.blk.zstd``.
```
Nana-EC, line 1: Q: Does this tool also serve as a perf tool in pulling blocks at high load ?

AlfredoG87: I haven't tried in high-load for a perf. I've used it to fetch historical blocks, so I am confident it should support live stream while on high-load TPS.

PR #3363, opened 2026-08-04
chore: Consolidate Server Port Variable Definitions

### LINKED ISSUE #1404: Consolidate Server Port Variable Definitions
### Story Form

As a block node operator
I want to consolidate the number of location in the code where the ports are specified
So that there is a cleaner, efficient and common mechanism for port configuration for the server

### Technical Notes

In PR #805 a single change to set the new default port for Helidon and Prometheus resulted in a lot of file needing to be changed.
This covered configs at the JVM, docker and k8s levels. It was also made more complex by the Simulator runs.
Preferably we'd have them in 1 or very few places so operators can easily customize it.

A single environment variable location would be preferred

ivannov:
### New `PortsConfig` record consolidates four per-plugin port properties

`app-config` gains a single `PortsConfig` record with publisher, subscriber, blockAccess, and serverStatus fields, replacing the separate `BlockAccessConfig`, `ServerStatusConfig`, and the `port` fields on `PublisherConfig`/`SubscriberConfig`. The four plugin classes (`BlockAccessServicePlugin`, `ServerStatusServicePlugin`, `StreamPublisherPlugin`, `SubscriberServicePlugin`) now read their dedicated port from `PortsConfig` now. The health plugin is intentionally untouched: it runs its own dedicated web server and keeps `HEALTH_PORT` separate from this record.

### `LegacyPortsEnvironmentConfigSource` keeps the old per-plugin env vars working

New `ConfigSource` in `app-config` maps the pre-consolidation env var names (`BLOCK_ACCESS_PORT`, `SERVER_STATUS_PORT`, `PRODUCER_PORT`, `SUBSCRIBER_PORT`) onto the renamed `ports.*` properties, so operators who have not migrated their env vars yet see no behavior change. It sits at ordinal 299, one below `AutomaticEnvironmentVariableConfigSource`'s 300, so the new `PORTS_*` env var wins whenever both the old and new name are set for the same plugin. 

## Related Issue(s)

Closes #1404


jsync-swirlds:
I somehow missed that this was in the backlog.

We should not do this as it, once again, makes some plugins "special access" because they get ports configured system-wide.

Every plugin should be completely severable, and that includes that it should not have plugin-specific configuration in system-wide configuration.

jsync-swirlds (changes requested):
We shouldn't make this change and should close the parent issue instead.

This violates the plugin separation rules.

ivannov:
Based on the discussions in the comments, we decided not to go into this PR's direction

PR #3369, opened 2026-08-04
fix(backfill): dial server status on a dedicated source port

### LINKED ISSUE #3368: bug(backfill): Backfill using same port for subscribe and serverStatus APIs of other Source BNs
### Description

Backfill currently does not support Multi-port configuration

currently backfill configuration only supports 1 port (subscribe method) but when querying the serverStatus service to get the available ranges

### Steps to Reproduce

deploy a BN with different ports for subscribe and serverStatus Services.

Try to connect another BN to the multi-port for backfill, backfill will try to reach serverStatus at the only port that Backfill allows to configure.

### Additional context

_No response_

AlfredoG87:
## Problem

Production Block Nodes expose each service on its own port (e.g. `subscriber: 40980`, `serverStatus: 40982`). Backfill built a single gRPC channel per source from the one configured `port` and used it for both the subscribe RPC (block fetching) and the server-status RPC (range discovery). With split ports, the server-status call was dialed on the subscriber port (40980) instead of the server-status port (40982), so range discovery failed and the node never backfilled. CI, which runs every service on one port (40840), was unaffected and hid the issue.

## Change

- Added an optional `status_port` field to `BlockNodeSourceConfig`. When unset (`0`) it falls back to `port`, so existing sources and CI are unchanged.
- `BlockNodeClient` now builds a distinct channel per service: the subscribe RPC dials `port`, the server-status RPC dials `status_port`. Each service gets its own channel; only the port value falls back.
- The Helm chart renders `status_port` into `block-node-sources.json` when a source sets it.

getBlock/BlockAccess is intentionally out of scope: backfill only uses the subscribe and server-status services.

## Configuration

```json
{ "nodes": [ { "address": "bn1", "port": 40980, "status_port": 40982, "priority": 1 } ] }
```

`status_port` is optional; omit it (or set `0`) to keep dialing `port` for server status.

## Testing

- `BlockNodeClientTest` — asserts the server-status RPC routes to `status_port` (40982) while subscribe stays on `port` (40980), and that an unset `status_port` falls back to `port`.
- `BackfillMultiportSourceTests` — in-process API test that boots one `BlockNodeApp` with subscriber and server-status on different ports, publishes blocks 0-2, then points the real backfill `BlockNodeClient` at it and asserts `serverStatusDetail` (on the status port) reports the range and `getBatchOfBlocks` (on the subscribe port) fetches all three blocks.

Both pass; the API test runs in-process in ~6s.

Resolves #3368

### DIFF
--- a/block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
+++ b/block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
@@ -111,7 +111,11 @@ public class BlockNodeClient implements AutoCloseable {
             new Options(Optional.empty(), ServiceInterface.RequestOptions.APPLICATION_GRPC);
 
     private final PbjGrpcClientConfig grpcConfig;
-    private final WebClient webClient;
+    /** Channel for the subscribe RPC, dialing the source's `port`. Package-private for testing. */
+    final WebClient webClient;
+    /** Dedicated channel for the server status RPC, dialing `status_port` (or `port` when unset). Package-private for testing. */
+    final WebClient statusWebClient;
+
     private final int globalTimeoutMs;
     private BlockStreamSubscribeUnparsedClient blockStreamSubscribeUnparsedClient;
     private BlockNodeServiceInterface.BlockNodeServiceClient blockNodeServiceClient;
@@ -153,15 +157,33 @@ public class BlockNodeClient implements AutoCloseable {
                 maxProtobufMessageSizeBytes,
                 maxIncomingBufferSize);
 
-        webClient = WebClient.builder()
-                .baseUri(protocol + blockNodeConfig.address() + ":" + blockNodeConfig.port())
+        // The subscribe RPC dials the source's main port, while the server status RPC may live
+        // on a dedicated port. Fall back to the main port when no status port is configured (statusPort == 0).
+        final int subscribePort = blockNodeConfig.port();
+        final int statusPort = blockNodeConfig.statusPort() != 0 ? blockNodeConfig.statusPort() : subscribePort;
+
+        // Each service gets its own channel; only the port value falls back to `port` when unset.
+        webClient = buildWebClient(
+                protocol + blockNodeConfig.address() + ":" + subscribePort, tls, tuning, pollWaitMs, connectTimeoutMs);
+        statusWebClient = buildWebClient(
+                protocol + blockNodeConfig.address() + ":" + statusPort, tls, tuning, pollWaitMs, connectTimeoutMs);
+
+        initializeClient();
+    }
+
+    private WebClient buildWebClient(
+            @NonNull String baseUri,
+            @NonNull Tls tls,
+            @Nullable GrpcWebClientTuning tuning,
+            int pollWaitMs,
+            int connectTimeoutMs) {
+        return WebClient.builder()
+                .baseUri(baseUri)
                 .tls(tls)
                 .protocolConfigs(List.of(buildHttp2Config(tuning), buildGrpcConfig(pollWaitMs, tuning)))
                 .connectTimeout(Duration.ofMillis(connectTimeoutMs))
                 .keepAlive(true)
                 .build();
-
-        initializeClient();
     }
 
     private Http2ClientProtocolConfig buildHttp2Config(@Nullable GrpcWebClientTuning tuning) {
@@ -201,9 +223,11 @@ public class BlockNodeClient implements AutoCloseable {
 
     public void initializeClient() {
         try {
-            PbjGrpcClient pbjGrpcClient = new PbjGrpcClient(webClient, grpcConfig);
-            blockNodeServiceClient = new BlockNodeServiceInterface.BlockNodeServiceClient(pbjGrpcClient, OPTIONS);
-            blockStreamSubscribeUnparsedClient = new BlockStreamSubscribeUnparsedClient(pbjGrpcClient, globalTimeoutMs);
+            PbjGrpcClient subscribeGrpcClient = new PbjGrpcClient(webClient, grpcConfig);
+            PbjGrpcClient statusGrpcClient = new PbjGrpcClient(statusWebClient, grpcConfig);
+            blockNodeServiceClient = new BlockNodeServiceInterface.BlockNodeServiceClient(statusGrpcClient, OPTIONS);
+            blockStreamSubscribeUnparsedClient =
+                    new BlockStreamSubscribeUnparsedClient(subscribeGrpcClient, globalTimeoutMs);
             nodeReachable = true;
         } catch (IllegalArgumentException | IllegalStateException | UncheckedIOException ex) {
             LOGGER.log(Level.WARNING, "Failed to initialize gRPC client: %s".formatted(ex.getMessage()), ex);
--- a/block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
+++ b/block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
@@ -119,6 +119,7 @@ public class BlockNodeClient implements AutoCloseable {
     private final int globalTimeoutMs;
     private BlockStreamSubscribeUnparsedClient blockStreamSubscribeUnparsedClient;
     private BlockNodeServiceInterface.BlockNodeServiceClient blockNodeServiceClient;
+    private PbjGrpcClient subscribeGrpcClient;
     private boolean nodeReachable;
 
     /**
@@ -225,7 +226,7 @@ public class BlockNodeClient implements AutoCloseable {
 
     public void initializeClient() {
         try {
-            PbjGrpcClient subscribeGrpcClient = new PbjGrpcClient(webClient, grpcConfig);
+            subscribeGrpcClient = new PbjGrpcClient(webClient, grpcConfig);
             PbjGrpcClient statusGrpcClient = new PbjGrpcClient(statusWebClient, grpcConfig);
             blockNodeServiceClient = new BlockNodeServiceInterface.BlockNodeServiceClient(statusGrpcClient, OPTIONS);
             blockStreamSubscribeUnparsedClient =
@@ -239,8 +240,13 @@ public class BlockNodeClient implements AutoCloseable {
 
     @Override
     public void close() throws IOException {
+        // Close both channels: the status client owns the status channel; the subscribe channel is
+        // held directly since BlockStreamSubscribeUnparsedClient does not expose a close().
         if (blockNodeServiceClient != null) {
             blockNodeServiceClient.close();
+        }
+        if (subscribeGrpcClient != null) {
+            subscribeGrpcClient.close();
         }
     }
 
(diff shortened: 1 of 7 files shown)


block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
```diff
@@ -111,7 +111,11 @@ private record Options(Optional<String> authority, String contentType) implement
             new Options(Optional.empty(), ServiceInterface.RequestOptions.APPLICATION_GRPC);
 
     private final PbjGrpcClientConfig grpcConfig;
-    private final WebClient webClient;
+    /** Channel for the subscribe RPC, dialing the source's `port`. Package-private for testing. */
```
jsync-swirlds: If possible, I'd like to see us use `subscribePort` with `port` if unset; which would make subscribe and status consistent.

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
```diff
@@ -153,15 +157,33 @@ public BlockNodeClient(
...
+        // The subscribe RPC dials the source's main port, while the server status RPC may live
+        // on a dedicated port. Fall back to the main port when no status port is configured (statusPort == 0).
+        final int subscribePort = blockNodeConfig.port();
+        final int statusPort = blockNodeConfig.statusPort() != 0 ? blockNodeConfig.statusPort() : subscribePort;
+
+        // Each service gets its own channel; only the port value falls back to `port` when unset.
+        webClient = buildWebClient(
+                protocol + blockNodeConfig.address() + ":" + subscribePort, tls, tuning, pollWaitMs, connectTimeoutMs);
+        statusWebClient = buildWebClient(
+                protocol + blockNodeConfig.address() + ":" + statusPort, tls, tuning, pollWaitMs, connectTimeoutMs);
```
jsync-swirlds, line 172: For safety, we could consider the following adjustment:

```suggestion
        private final String subscribeUrl = "%s%s://%s:%d".formatted(protocol, enableTls ? "s" : "", blockNodeConfig.address(), subscribePort);
        private final String statusUrl = "%s%s://%s:%d".formatted(protocol, enableTls ? "s" : "", blockNodeConfig.address(), statusPort);
        webClient = buildWebClient(
                subscribeUrl, tls, tuning, pollWaitMs, connectTimeoutMs);
        statusWebClient = buildWebClient(
                statusUrl, tls, tuning, pollWaitMs, connectTimeoutMs);
```

Note, this would allow us to use `http` or `grpc` for the protocol; rather than `http://`, `https://`, `grpc://`, or `grpcs://`.

protobuf-sources/src/main/proto/internal/block_node_source.proto
```diff
@@ -53,6 +53,14 @@ message BlockNodeSourceConfig {
...
    */
   org.hiero.block.api.NetworkConnection.IpProtocol protocol = 8;
+
+  /**
+   * Optional dedicated port for the server status RPC (0 = use `port`).
+   * Some deployments expose server status on a different port than the
+   * block-stream/subscribe RPC; when set, status queries dial this port
+   * while subscribe/getBlock keep using `port`.
+   */
+  uint32 status_port = 9;
```
jsync-swirlds, line 63: I'd like to see both subscribe and status called out specifically, if we could.


```suggestion

  /**
   * Optional dedicated port for the server status RPC (0 = use `port`).
   * Some deployments expose server status on a different port than the
   * default; when set, status queries dial this port while other APIs
   * use different ports.
   */
  uint32 status_port = 9;

  /**
   * Optional dedicated port for the subscribe RPC (0 = use `port`).
   * Some deployments expose the subscribe API on a different port than the
   * default; when set, block stream subscriptions dial this port while
   * other APIs use different ports.
   */
  uint32 subscribe_port = 10;
```

jsync-swirlds (dismissed):
A few improvements that are non-blocking.

Unfortunately, the issues with GHA mean we may not see CI run for quite some time.

Nana-EC (dismissed):
Approving based on other reviews to unblock cherry pick needs.
I know the risk is low

rockysingh (dismissed):

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
AlfredoG87: I am trying to avoid breaking changes, but this is a good suggestion for a follow-up not a HotFix.

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
AlfredoG87: I am trying to avoid breaking changes, but this is a good suggestion for a follow-up not a HotFix.

protobuf-sources/src/main/proto/internal/block_node_source.proto
AlfredoG87: I am trying to avoid breaking changes, but this is a good suggestion for a follow-up not a HotFix.

jsync-swirlds: This is properly part of this change, not a separate breaking change.
It reduces the chance of further confusion.

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
jsync-swirlds: We're adding an optional port for status, adding an optional port for subscribe is not a "breaking change", it's the same thing we're doing for status.

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
jsync-swirlds: This is, by its nature non-breaking; but it is not critical and can be done in a separate PR.

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
```diff
@@ -201,9 +223,11 @@ private GrpcClientProtocolConfig buildGrpcConfig(int pollWaitTimeMs, @Nullable G
 
     public void initializeClient() {
         try {
-            PbjGrpcClient pbjGrpcClient = new PbjGrpcClient(webClient, grpcConfig);
-            blockNodeServiceClient = new BlockNodeServiceInterface.BlockNodeServiceClient(pbjGrpcClient, OPTIONS);
-            blockStreamSubscribeUnparsedClient = new BlockStreamSubscribeUnparsedClient(pbjGrpcClient, globalTimeoutMs);
+            PbjGrpcClient subscribeGrpcClient = new PbjGrpcClient(webClient, grpcConfig);
+            PbjGrpcClient statusGrpcClient = new PbjGrpcClient(statusWebClient, grpcConfig);
+            blockNodeServiceClient = new BlockNodeServiceInterface.BlockNodeServiceClient(statusGrpcClient, OPTIONS);
+            blockStreamSubscribeUnparsedClient =
```
Nana-EC, line 232: `blockStreamSubscribeUnparsedClient` is missing from the `close()`

block-node/base/src/test/java/org/hiero/block/node/base/client/BlockNodeClientTest.java
```diff
@@ -2,14 +2,54 @@
...
+                .statusPort(statusPort)
+                .build();
+        return new BlockNodeClient(
+                source, GLOBAL_TIMEOUT_MS, false, MAX_INCOMING_BUFFER_SIZE, MAX_PROTOBUF_MESSAGE_SIZE, null);
+    }
+
+    @Test
+    @DisplayName("Status RPC dials the dedicated status port when configured")
+    void statusPortRoutesToDedicatedChannel() {
+        final BlockNodeClient client = clientFor(40980, 40982);
```
Nana-EC: missing close. 
Maybe wrap in a resource try
```suggestion
        try (BlockNodeClient client = clientFor(40980, 40982)) {
```

tools-and-tests/suites/src/main/java/org/hiero/block/suites/backfill/BackfillMultiportSourceTests.java
```diff
@@ -0,0 +1,203 @@
...
+        // Provision the verifier with the roster that signs blocks built by BlockItemBuilderUtils.
+        BlockItemBuilderUtils.provisionTssBootstrap();
+    }
+
+    @AfterEach
+    void afterEach() {
+        if (app != null && app.blockNodeState() != State.SHUTTING_DOWN) {
+            app.shutdown("BackfillMultiportSourceTests", "test teardown");
+        }
+        System.clearProperty("server.status.port");
```
Nana-EC, line 111: property clear is better in beforeEach to avoid spill overs from crashes etc

charts/block-node-server/templates/configmap-block-node-sources.yaml
```diff
@@ -9,7 +9,7 @@ data:
       "nodes": [
         {{- range $index, $node := .Values.blockNode.backfill.sources }}
         {{- if $index }},{{ end }}
-        { "address": "{{ $node.address }}", "port": {{ $node.port }}, "priority": {{ $node.priority }} }
+        { "address": "{{ $node.address }}", "port": {{ $node.port }}, "priority": {{ $node.priority }}{{ if $node.statusPort }}, "status_port": {{ $node.statusPort }}{{ end }} }
```
Nana-EC: This would mean where default the value ends up omitting status_port?
Is is better to always have status _port and rather decide what value to set, default or statusPort is set?
I'm not sure.

Nana-EC (commented):
Some considerations

jsync-swirlds (dismissed):

charts/block-node-server/templates/configmap-block-node-sources.yaml
AlfredoG87: - The fallback (0/absent → use port) is defined in exactly one place: the proto contract + BlockNodeClient. Always-emitting in the chart (status_port: {{ .statusPort | default .port }}) would duplicate that fallback into 
  the Helm template — two sources of truth for the same rule, which drift over time.
  - Absence isn't ambiguous here — it's a documented, well-defined state (0 = use port), and it keeps the rendered block-node-sources.json minimal.
  - For operators who want it explicit, the values-override example already sets both ports literally, so they can be explicit when they choose to.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/backfill/BackfillMultiportSourceTests.java
AlfredoG87: Also great recommendation, thanks!

✅ Done!

block-node/base/src/test/java/org/hiero/block/node/base/client/BlockNodeClientTest.java
AlfredoG87: Great Suggestion, thanks again!!

✅ Done!

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
AlfredoG87: Great catch!! Thank you! 

✅ Done!

protobuf-sources/src/main/proto/internal/block_node_source.proto
AlfredoG87: I've addded it ✅

jsync-swirlds (approved):

Nana-EC (approved):

PR #3373, opened 2026-08-05
feat: Update publisher to use ApplicationStateFacility for block set ranges

### LINKED ISSUE #2822: Update publisher to use ApplicationStateFacility for block set ranges
### Story Form

As a BlockNode engineer I would like the ApplicationStateFacility to be the source of truth for all block range searches.

### Technical Notes

Publisher must, on startup, also use the ranged set of stored blocks coming from the application state facility in order to correctly determine which block to expect next.

ivannov:
### Publisher startup watermark now also considers ApplicationStateFacility stored blocks

`LiveStreamPublisherManager.initializeBlockNumbers` previously seeded `lastPersistedBlockNumber` only from `historicalBlockProvider().availableBlocks().max()`. That is wrong for a node with no local block provider (RHF), since `availableBlocks()` stays empty for the node's whole lifetime even though ASF has a persisted stored-block range. The watermark now takes `Math.max` of the historical provider's max and `context.storedBlocks()`'s last range end.

### `StreamPublisherPlugin.context` needs `onContextUpdate()` wired up, and `volatile`

`BlockNodeApp` calls `init()` with an empty-lists context, then delivers a rebuilt context carrying the ASF-synced stored blocks via `onContextUpdate()`, and only then calls `start()`. The plugin was not overriding `onContextUpdate()`, so `start()` was seeding the watermark from the stale, empty-lists context and silently dropping the ASF update. Field is marked `volatile` since `onContextUpdate()` runs on a separate thread from `start()`.

## Related Issue(s)

Closes #2822

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -751,10 +751,15 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         // streamed, but _only_ on startup. After that there should always be
         // a delta (next unstreamed must always be strictly greater than the current
         // streaming block number).
-        final List<BlockRange> storedBlocks = serverContext.storedBlocks();
-        final long latestKnownBlock = storedBlocks.isEmpty()
-                ? UNKNOWN_BLOCK_NUMBER
-                : storedBlocks.getLast().rangeEnd();
+        final long latestKnownBlock;
+        if (serverContext != null && serverContext.storedBlocks() != null) {
+            final List<BlockRange> storedBlocks = serverContext.storedBlocks();
+            latestKnownBlock = storedBlocks.isEmpty()
+                    ? UNKNOWN_BLOCK_NUMBER
+                    : storedBlocks.getLast().rangeEnd();
+        } else {
+            latestKnownBlock = UNKNOWN_BLOCK_NUMBER;
+        }
         // Always set the last persisted block number, even if there are no
         // known blocks.
         lastPersistedBlockNumber.set(latestKnownBlock);
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -49,6 +49,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
+import org.hiero.block.api.BlockRange;
 import org.hiero.block.api.PublishStreamResponse;
 import org.hiero.block.api.PublishStreamResponse.EndOfStream.Code;
 import org.hiero.block.internal.BlockItemSetUnparsed;
@@ -750,8 +751,12 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         // streamed, but _only_ on startup. After that there should always be
         // a delta (next unstreamed must always be strictly greater than the current
         // streaming block number).
-        final long latestKnownBlock =
-                serverContext.historicalBlockProvider().availableBlocks().max();
+        final List<BlockRange> storedBlocks = serverContext.storedBlocks();
+        final long highestStoredBlock = storedBlocks.isEmpty()
+                ? UNKNOWN_BLOCK_NUMBER
+                : storedBlocks.getLast().rangeEnd();
+        final long latestKnownBlock = Math.max(
+                serverContext.historicalBlockProvider().availableBlocks().max(), highestStoredBlock);
         // Always set the last persisted block number, even if there are no
         // known blocks.
         lastPersistedBlockNumber.set(latestKnownBlock);
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -752,11 +752,9 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         // a delta (next unstreamed must always be strictly greater than the current
         // streaming block number).
         final List<BlockRange> storedBlocks = serverContext.storedBlocks();
-        final long highestStoredBlock = storedBlocks.isEmpty()
+        final long latestKnownBlock = storedBlocks.isEmpty()
                 ? UNKNOWN_BLOCK_NUMBER
                 : storedBlocks.getLast().rangeEnd();
-        final long latestKnownBlock = Math.max(
-                serverContext.historicalBlockProvider().availableBlocks().max(), highestStoredBlock);
         // Always set the last persisted block number, even if there are no
         // known blocks.
         lastPersistedBlockNumber.set(latestKnownBlock);
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
@@ -105,7 +105,7 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
     private static final System.Logger LOGGER = System.getLogger(StreamPublisherPlugin.class.getName());
 
     /// The block node context, for access to core facilities.
-    private BlockNodeContext context;
+    private volatile BlockNodeContext context;
     /// The publisher block manager, which connects handlers to the messaging facility.
     private StreamPublisherManager publisherManager;
 
@@ -180,6 +180,15 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
     public void stop() {
         context.blockMessaging().unregisterBlockNotificationHandler(publisherManager);
         publisherManager.shutdown();
+    }
+
+    /**
+     * {@inheritDoc}
+     * This method is called on a separate thread. Make sure this.context is marked as `volatile`
+     */
+    @Override
+    public void onContextUpdate(final BlockNodeContext context) {
+        this.context = context;
     }
 
     /// This method is called when a new publisher handler is created.
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
@@ -105,7 +105,7 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
     private static final System.Logger LOGGER = System.getLogger(StreamPublisherPlugin.class.getName());
 
     /// The block node context, for access to core facilities.
-    private volatile BlockNodeContext context;
+    private BlockNodeContext context;
     /// The publisher block manager, which connects handlers to the messaging facility.
     private StreamPublisherManager publisherManager;
 
@@ -184,7 +184,6 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
 
     /**
      * {@inheritDoc}
-     * This method is called on a separate thread. Make sure this.context is marked as `volatile`
      */
     @Override
     public void onContextUpdate(final BlockNodeContext context) {
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
@@ -190,7 +190,9 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
      */
     @Override
     public void onContextUpdate(final BlockNodeContext context) {
-        this.context.set(Objects.requireNonNull(context));
+        if (context != null) {
+            this.context.set(context);
+        }
     }
 
     /// This method is called when a new publisher handler is created.
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
@@ -9,6 +9,7 @@ import com.hedera.pbj.runtime.grpc.Pipelines;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
 import org.hiero.block.api.BlockStreamPublishServiceInterface;
 import org.hiero.block.api.PublishStreamRequest;
 import org.hiero.block.api.PublishStreamResponse;
@@ -105,7 +106,7 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
     private static final System.Logger LOGGER = System.getLogger(StreamPublisherPlugin.class.getName());
 
     /// The block node context, for access to core facilities.
-    private BlockNodeContext context;
+    private final AtomicReference<BlockNodeContext> context = new AtomicReference<>();
     /// The publisher block manager, which connects handlers to the messaging facility.
     private StreamPublisherManager publisherManager;
 
@@ -136,7 +137,7 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
                 (BlockStreamPublishServiceMethod) method;
 
         final int maxMessageSize =
-                context.configuration().getConfigData(ServerConfig.class).maxMessageSizeBytes();
+                context.get().configuration().getConfigData(ServerConfig.class).maxMessageSizeBytes();
 
         final String rawCorrelationId = options.metadata().getOrDefault("hiero-correlation-id", "");
         final String correlationId = truncateCorrelationId(rawCorrelationId);
@@ -155,7 +156,7 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
 
     @Override
     public void init(@NonNull final BlockNodeContext context, @NonNull final ServiceBuilder serviceBuilder) {
-        this.context = Objects.requireNonNull(context);
+        this.context.set(Objects.requireNonNull(context));
         // register us as a service, we need to register the gRPC service in
         // the init method, otherwise the server will be started and we will not
         // have registered at all. A null port (the default) shares server.port.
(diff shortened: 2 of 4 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -750,8 +751,12 @@ private void initializeBlockNumbers(final BlockNodeContext serverContext) {
...
         // a delta (next unstreamed must always be strictly greater than the current
         // streaming block number).
-        final long latestKnownBlock =
-                serverContext.historicalBlockProvider().availableBlocks().max();
+        final List<BlockRange> storedBlocks = serverContext.storedBlocks();
+        final long highestStoredBlock = storedBlocks.isEmpty()
+                ? UNKNOWN_BLOCK_NUMBER
+                : storedBlocks.getLast().rangeEnd();
+        final long latestKnownBlock = Math.max(
+                serverContext.historicalBlockProvider().availableBlocks().max(), highestStoredBlock);
```
ata-nas: This seems wrong, why are we also consulting the available blocks from historical provider? Aren't these merged into the stored blocks?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -182,6 +182,15 @@ public void stop() {
...
 
+    /**
+     * {@inheritDoc}
+     * This method is called on a separate thread. Make sure this.context is marked as `volatile`
+     */
+    @Override
+    public void onContextUpdate(final BlockNodeContext context) {
+        this.context = context;
+    }
+
```
ata-nas, line 197: This seems unnecessary, the context is only used on start and init? Am I wrong? It is passed to the live stream manager, we should be careful how it uses the context.

Mind you, we should probably have the ability to see updates of latest stored in case they have rolled forward, similar to what we have in the new verification plugin. I think we are using the context update for the wrong purposes here.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -105,7 +105,7 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
     private static final System.Logger LOGGER = System.getLogger(StreamPublisherPlugin.class.getName());
 
     /// The block node context, for access to core facilities.
-    private BlockNodeContext context;
+    private volatile BlockNodeContext context;
```
ata-nas: Not sure why we need this to be volatile as no one will ever see this update post start.

ata-nas (commented):
This is deceptively simple and requires more things to reason about and do here.

I think the initial approach is inaccurate and misses quite a lot of things to do.

We have to more carefully evaluate what needs to change, who owns volatile values and how updates happen and affect this plugin. We have done, I believe, a good job of value ownership in the new plugin, we can take inspiration from there, but we must also take into consideration dynamically changing stored.max values, which we do not seem to evaluate for yet. This is tricky, however, as this plugin has a live stream manager which holds state that is quite difficult to reason about.

ata-nas:
@ivannov note failing CI, we do not support `feature:` pr title.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ivannov: Yes, you are right. I completely ignored the fact that this is called only once - upon start. And here stored really includes available. My mind was in the mood that this will be called regularly, so I wanted to have the fresh data from the historical block provider. But now I notice that the live stream publisher manager also handles persisted notifications

Simplified it...

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
ivannov: Don't we need it to be volatile in `start()` as well?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
ivannov: It has also been used in `open()`. I thought that there it needs updates as well (and don't know why we haven't spotted it so far)

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
jsync-swirlds: This `volatile` only affects the _reference_, not the entire object.
That means we're only protecting the _assignment to this variable_ from being cached on the CPU.

When `init()` is called (the only place we _assign_ this variable), the other threads in this plugin aren't running yet, so we shouldn't need it to be volatile.

The possible reason to make this `volatile` would be if we add `onContextUpdate`; but I'm not sure we actually need that.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
jsync-swirlds: After startup, we maintain the last persisted completely independent from ASF, context, or stored versus available blocks.

Last persisted (after startup) is based on _notifications_, not query to the ASF.

We probably don't need this context update.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -750,8 +751,10 @@ private void initializeBlockNumbers(final BlockNodeContext serverContext) {
         // streamed, but _only_ on startup. After that there should always be
         // a delta (next unstreamed must always be strictly greater than the current
         // streaming block number).
-        final long latestKnownBlock =
-                serverContext.historicalBlockProvider().availableBlocks().max();
+        final List<BlockRange> storedBlocks = serverContext.storedBlocks();
+        final long latestKnownBlock = storedBlocks.isEmpty()
+                ? UNKNOWN_BLOCK_NUMBER
+                : storedBlocks.getLast().rangeEnd();
```
jsync-swirlds: We should incorporate both options here, and do some preemptive null checks:

```suggestion
        final long latestKnownBlock;
        if (serverContext != null
                && serverContext.storedBlocks() != null
                && serverContext.historicalBlockProvider() != null
                && serverContext.historicalBlockProvider().availableBlocks() != null) {
            final List<BlockRange> storedBlocks = serverContext.storedBlocks();
            final long maxAvailableBlock = serverContext.historicalBlockProvider().availableBlocks().max();
            latestKnownBlock = storedBlocks.isEmpty()
                    ? maxAvailableBlock >= 0
                            ? maxAvailableBlock
                            : UNKNOWN_BLOCK_NUMBER
                    : storedBlocks.getLast().rangeEnd();
        } else {
            latestKnownBlock = UNKNOWN_BLOCK_NUMBER;
        }
```

ivannov: I agree about the preemptive checks. But about the available blocks, I got this comment from @ata-nas: https://github.com/hiero-ledger/hiero-block-node/pull/3373#discussion_r3720894927

And based on that I only left the stored blocks, which at the moment of constructing the manager (which is during plugin start) should contain the available blocks as well.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
ivannov: OK, after digging deeper. Each plugin (including the publisher) receives and instance of the block node context in the init method. There it contains empty lists for stored and available blocks.

But then the block node app loads TSS data, address books, stored and available blocks prior to starting the plugins. And in that process it builds a new instance of the context and sets there the stuff that it loads (including the stored and available blocks). So, if we don't update our context, the publisher plugin will possess the stale data for the stored blocks during its start method.

That is why we need this method. I agree that the context does not need to be volatile as we are using it only during start method. But keeping this method is the easiest way to "fix" the staleness of the stored blocks.

ata-nas: Okay, I got confused here. Currently, right after init and just before start is called on plugins, there will be a context update. We need that because otherwise we will never see the updated ranged sets.

On another note, @jsync-swirlds I think we do actually want to dynamically receive updates, because for any reason, the publisher might fall behind from the latest stored, maybe a plugin reports something to those sets, which is not necessary to also result in a persistence notification (soon with architectural reworks, we will have plugins initialize async, so reading of disk will not be guaranteed to finish before we get the context update before start is called, but also, we will probably drop this handler method). I question if this is the right time to do this change, publisher only runs with local history plugins, which only report available blocks. If we want cloud plugins to run with publisher, we have to listen for async updates which will roll forward the publisher.

Not sure we gain anything at the moment, but possibly make things more fragile.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: I kind of disagree with all these null checks. If we do not have a valid context, or no context at all, something is very broken.

Then, I do not understand why we want to fall back to available blocks. This defeats the whole purpose of this change and makes me think even more we should delay this change, as we are expecting architectural changes + ASF as a plugin, which changes quite a lot of things.

jsync-swirlds: Making ASF a plugin won't change any of the logic here, just the source of some data.

As far as not having context, I agree, but the right way to handle that is that it is detected in the start or init method and plugin health is updated properly (which also depends on future work). Here we just avoid throwing unnecessary NPEs when the issue should/must be handled elsewhere.

As to "don't use available", why would we not use available if stored is empty? The data sources are not the same, and there is no reason not to use available as a backstop. The point of the requested change is to _prefer_ stored, not ignore available completely.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
jsync-swirlds: 1. We clearly have an issue with context if we're replacing it after init; we made a lot of assumptions in that regard.
   * That's something to fix separately, however, so I see the value of including this, but we need to very strictly minimize what we do here.
1. Publisher never checks context after startup, and really should not, because context is far more likely to be out of date than Publisher; if the notifications for persistence are desynchronized we'll have problems far beyond the "latest available".
1. The "publisher is not run with cloud plugins" is, indeed, true _for production servers_, but the insistence on maintaining a "everything included" testing image means we have to handle the unexpected combination or we'll break in test environments.
   * Users may choose to setup a combined local-and-remote history block node (perhaps in a sphere environment), so these tests _are_ valid.

ata-nas: 1. just before start(), onContextUpdate() is called on every plugin. This includes every update we have for the app state from it's own initialization, but also in case some other plugin updated application state with some values during init()
2. True, we should not be checking context, but we should be aware of stored blocks updates. This is not about out of sync persistence notifications. THis is about plugin x reports something new, but does not fire a notification. This happens in every plugin on init for local storage, but happens async (bn app continues with the lifecycle) with cloud plugins.
3. I agree on the intentionality of the publisher not run with cloud plugins, but if we have to also support that, we have to dynamically listen for stored blocks update and be able to roll forward just like we do with verification.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: I agree with us not having health plugin yet to handle the missing/uninitialized context. I think we can go with the suggested for now, but we should probably add a WARNING log - "hey, something is wrong".

I am totally confused about why we want to use available if stored is empty? If we have something available and stored is empty, or stored mismatches (does not include the complete available subset), this is a bug with asf. Moreover, a mixture of storage type plugins where some of them are also providers, but some not, means that some will report to available, others to stored directly. If we cannot trust stored, we will have issues in such deployments I think. In my mind, the whole reason to switching to stored is to never use available. Sorry for my confusion, I just don't see it :shrug:

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
jsync-swirlds: Not sure if it was clear, but this roughly matches what I was suggesting in my most recent reply.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
jsync-swirlds: It appears we do need `onContextUpdate`, so `volatile` is likely required; we might even upgrade this to `AtomicReference` so a CAS can ensure we don't have an incorrect update.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I'm taking a "belt and suspenders" approach here to ensure we get a useable value in as many cases as possible.
We _should not_ have a value in available if nothing is in stored, but there is a host of timing and other conditions that might cause one or the other to be unavailable, so it makes sense to use every available option.

ata-nas: I am ok with the "belt and suspenders" so long as we log it as well, we do want to know how the BN has started up and where it took it's data from.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -166,22 +167,32 @@ public void init(@NonNull final BlockNodeContext context, @NonNull final Service
...
+        context.get().blockMessaging().unregisterBlockNotificationHandler(publisherManager);
         publisherManager.shutdown();
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onContextUpdate(final BlockNodeContext context) {
+        this.context.set(Objects.requireNonNull(context));
```
ata-nas: Not sure we want to throw here, rather simply do not make an update if context is null?

But then note that on every update, we will be doing this operation. Do we really want that? I am not sure about that. We really only want an update of the stored.max value. Note also this update will be very frequent, as context update happen basically on every block received and stored.

Let's discuss that separately as there are also different opinions on what update we should make. We have another example in the new verification plugin where we also listen for an update only for the stored.max

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
```diff
@@ -105,7 +106,7 @@ public final class StreamPublisherPlugin implements BlockNodePlugin, BlockStream
     private static final System.Logger LOGGER = System.getLogger(StreamPublisherPlugin.class.getName());
 
     /// The block node context, for access to core facilities.
-    private BlockNodeContext context;
+    private final AtomicReference<BlockNodeContext> context = new AtomicReference<>();
```
ata-nas, line 109: Having switched this to an atomic ref means that the context (actual reference) can at any point be `null`, but this is true also for the volatile field. This means, however, that if we now all of a sudden start to change the context, we have to be aware of that. Just pointing out.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
ata-nas: I think that points 2 and 3 differ slightly, perhaps we can discuss separately?

ata-nas (commented):
A few comments, not sure we land on the same page.
Perhaps we can discuss this separately?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Logs are costly.  We could log at trace level, but we'll only see that when debugging.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
jsync-swirlds: I agree, we significantly overuse forced NPE when we could just test for null and skip over it (perhaps with a trace log).

In this specific case I think we should just not update the local reference.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
jsync-swirlds: Agreed, but if we ensure we never set null, it will never be null, after it's set initially (which makes most use of the reference simpler).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
jsync-swirlds: I think the entire discussion is a bit moot because we realized that the context update is required for other reasons.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: This is a single log on startup. I do not see that as an issue.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
ata-nas: True, what I am questioning, is do we really need an atomic ref, when all we need is a single ref set on a field at startup which we must never update again. If we have an on context update we need at runtime, no need to replace the actual context, we just take the update we need and move forward.

jsync-swirlds (approved):

Nana-EC (approved):

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java
jsync-swirlds: I think I see the point.
There is good reason to keep the context updated here, but if we only needed a value or two, I agree we could manage without the atomic reference.

We might want to revisit this in later work (perhaps when we migrate to a true ASF plugin instead of the non-plugin implementation), but I think this works well enough for now.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: Within reason, perhaps.  We need to avoid anything above INFO unless something is truly failing (as defined in the logging guidelines), but a single line once-per-startup log is within reason.

Issue #3372, opened 2026-08-05
Revisit `HashingUtilities#computeFinalBlockHash` Method Signature and `StreamingTreeHasher`'s Use of `CompletableFuture`


ata-nas:
### Story Form

As a Block Node Engineer
I want to have clean and maintainable utilities
So that I can efficiently make use of them

### Technical Notes

- Based on [this comment](https://github.com/hiero-ledger/hiero-block-node/pull/3299#discussion_r3687595252), we want to revisit and re-evaluate our hashing utilities.
- We want to choose the best consistent use of the `computeFinalBlockHash`, either passing `byte[]` or streaming hashers.
- We want to re-evaluate the use of `CompletableFuture` as a returned result of `StreamingTreeHasher#rootHash`, as currently we return a completed future from the only implementation we have.

ata-nas:
@jsync-swirlds what parent would be appropriate for this one?

jsync-swirlds:
> [@jsync-swirlds](https://github.com/jsync-swirlds) what parent would be appropriate for this one?

Absent any other ideas, a follow up usually belongs to the same parent as the issue to which we are following up.

ata-nas:
I think we should close this as complete in #3378? 
@rockysingh @jsync-swirlds ?

rockysingh:
@ata-nas yes this can be closed.

Issue #3367, opened 2026-08-04
`org.hiero.block.common.hasher.HashingUtilities#getBlockItemHash` Adds `BlockItemUnparsed` Wrapper to Streaming Hasher Instead of Value


ata-nas:
### Description

The Block Hasher of verification iterates over all block items received for a block. For each item, we resolve the respective streaming hasher the raw value of the item must go to as see in [this image](https://raw.githubusercontent.com/hiero-ledger/hiero-improvement-proposals/d1bf49d33473b9dbe4f6c018e460f802f293aa3c/assets/hip-1424/expanded-merkle-mountain-top.svg). So, for instance, if we see a `ROUND_HEADER` block item oneOf type, we have to take the raw value of the item (bytes) and add that to the `consensusHeaderHasher`.

What seems to be done, however, is that instead of the raw bytes to be added to the hasher, the whole `BlockItemUnparsed` proto wrapper gets hashed and added:

<img width="509" height="82" alt="Image" src="https://github.com/user-attachments/assets/1dc699b7-6620-43dd-aa96-b007c92cd368" />

<img width="735" height="293" alt="Image" src="https://github.com/user-attachments/assets/8c3fb569-6c80-4201-9723-b7781a9ec845" />

We should be adding the raw bytes of the `BlockItemUnparsed` message's `item` field instead of the whole serialized proto message (which can contain unknown fields).

PR #3378, opened 2026-08-05
feat(hasher): always feed 16 leaves to the Merkle Mountain Top hasher

### LINKED ISSUE #3377: Block hashing: empty pre-defined hasher subtrees should propagate null, not EMPTY_TREE_HASH
## Description

The block root hash treats missing hasher subtrees inconsistently:

- **Pre-defined hashers (HIP-1424 categories 3-7)**: when empty, contribute `EMPTY_TREE_HASH` (`SHA-384(0x00)`) to the root.
- **Extension hashers (categories 8-15)**, added in #3299: when empty, propagate `null` and are excluded from the root entirely.

Two hashers in the same "empty this block" state produce different contributions. Not a property we want in a merkle root spec.

## Reference implementation: PR #3299

PR #3299 already introduces the null-propagation mechanics we want to apply to the pre-defined side. Reuse the same pattern:

- `combineOptionalNodes(left, right)` at `common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java:483` — returns `null` when both children are absent; single-child promotion with `0x01` prefix; both-children combine with `0x02` prefix.
- `combineExtensionSubtreeRoots(...)` at `HashingUtilities.java:457` — unrolled tree combine over the 8 extension leaves using `combineOptionalNodes` at every level.

Applying the same pattern to categories 3-7 gives a single, consistent algorithm across every subtree position: any empty subtree contributes `null` and is excluded from the root, regardless of whether it's pre-defined or extension.

## Proposed fix

- Change pre-defined hasher combination in `HashingUtilities.computeFinalBlockHash` to feed subtree roots through `combineOptionalNodes` instead of always combining `EMPTY_TREE_HASH`.
- When a pre-defined hasher had no items added, its subtree root is `null` (rather than `EMPTY_TREE_HASH`), and the null propagates up per #3299's existing pattern.

## Coordination

Breaking change to block hashes. Requires:

- `common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java` (verifier side)
- `tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java` (producer side, mirrored change)
- HIP-1424 spec update
- Test vector regeneration (including #3299's HIP-1424 reference test)
- Existing WRB archives across previewnet/testnet/mainnet were produced under the old algorithm and won't verify against updated verifiers. Needs coordinated re-wrap or a version/cutover mechanism.

## Related

- **PR #3299** (reference implementation for null propagation)
- Slack + meeting: [pgm-a3-finish-bn 2026-08-04](https://swirldslabs.slack.com/archives/C08HECALMNF/p1785886793737959)
- #3367 (wrapper-vs-field bytes, tracked separately)
- #3195 (the epic #3299 resolves)
- HIP-1424, HIP-1056

rockysingh:
Fixes #3377.

## Summary

Per the offline discussion and Jasper's / Joseph's follow-up direction: use one streaming Merkle hasher for the block-root "Merkle Mountain Top", fed all 16 leaves every time, with `EMPTY_TREE_HASH` (`SHA-384(0x00)`) for any absent subtree. The 16-leaf tree shape is fully stable across all presence patterns, so Merkle proof paths for any fixed position are independent of which other positions are populated.

## Algorithm

Both `HashingUtilities.computeFinalBlockHash` (block-node) and `BlockStreamBlockHasher.hashBlockInternal` (tools) now:

1. Assemble the 16 leaves in fixed positional order (see table below).
2. Feed all 16 into a single `NaiveStreamingTreeHasher` / `StreamingHasher` — the same streaming hasher used within each subtree.
3. Take `rootHash()` as the Mountain Top root.
4. Combine with the consensus-timestamp leaf via `hashInternalNode` (0x02 prefix) to get the block root hash.

Empty subtree hashers and absent extension slots contribute `EMPTY_TREE_HASH`. Absent state root also defaults to `EMPTY_TREE_HASH`.

## Fixed leaf positions

| # | Content | Fallback when absent |
|---|---|---|
| 0 | `previousBlockHash` | required (throws) |
| 1 | `rootHashOfAllPreviousBlockHashes` | required (throws) |
| 2 | `startOfBlockStateRootHash` | `EMPTY_TREE_HASH` |
| 3 | `consensusHeaderHasher.rootHash()` | `EMPTY_TREE_HASH` |
| 4 | `inputTreeHasher.rootHash()` | `EMPTY_TREE_HASH` |
| 5 | `outputTreeHasher.rootHash()` | `EMPTY_TREE_HASH` |
| 6 | `stateChangesHasher.rootHash()` | `EMPTY_TREE_HASH` |
| 7 | `traceDataHasher.rootHash()` | `EMPTY_TREE_HASH` |
| 8-15 | Extension subtree roots (Extension 0-7) | `EMPTY_TREE_HASH` |

## Related changes in this PR

- **Length guard on positions 0-1**: short values would silently be zero-padded by the streaming hasher's buffer, producing an incorrect root. `computeFinalBlockHash` now throws `IllegalArgumentException` if either is not exactly 48 bytes.
- **`BlockHasher` (block-node/block-verification)**: all 8 extension subtree hashers are now instantiated up-front instead of the lazy null-and-create pattern. `createIfNull` deleted; `extensionSubtreeRoot` simplified to a plain `rootHash()` call.
- **API cleanup** (per Joseph's Aug 10 review):
  - Removed the legacy 9-arg `computeFinalBlockHash` overload. All callers now pass explicit extension roots.
  - Removed the `padAbsent` helper — extension args are `@NonNull byte[]`; callers pass `EMPTY_TREE_HASH` for absent slots.
  - Removed the 9 `Objects.requireNonNull` calls; the `@NonNull` annotations carry the same intent.
- **Design docs updated**: `block-stream-forward-compatibility.md` and `block-verification.md` now describe the fixed 16-leaf shape and the `EMPTY_TREE_HASH` fallback for absent subtrees.

## Fixture regeneration

- **`CN_0_73_TSS_WRAPS`** (7 fixture blocks): regenerated from CN's `block-11.12.3` batch. Blocks 0-4 (Schnorr) replaced with new `.blk.gz` files; old `BLOCK_466`/`BLOCK_467` (transition + settled) replaced with `BLOCK_322`/`BLOCK_391` since the new batch tops out at 392. New expected root hashes updated in both `ResourceTestBlockBuilder` and `BlockUtils`, along with the three test callers that referenced the old constant names.
- **`CN_0_77_TSS_SCHNORR`** (`StateProof`, 5 blocks) and **`WRB/SOLO_4N`** (5 blocks): CN is regenerating these separately; fixture tests referencing them will continue to fail on the current expected hashes until those land.

## Test status

- `:common` `HashingUtilitiesTest` — 16 passing. Parametric coverage over pre-defined and extension presence patterns, minimum-content case (only positions 0-1), and the length-guard rejection.
- `:tools` `BlockStreamBlockHasherTest` — 16 passing. New `MerkleMountainTopShapeTests` verifies always-feed-16 shape via a distinct manual balanced reference (production uses `StreamingHasher`; reference uses `hashInternalNode`).
- `:block-verification` / `:verification` — WRAPS fixture-chain tests now pass with the regenerated blocks. StateProof and WRB fixture tests still fail on hardcoded expected hashes pending CN regen.

## Reviewer notes

- `hashInternalNodeSingleChild` is no longer called from the block-hash path but remains as public API (kept for callers outside the block-root computation).
- The @todo(3372) comment on `computeFinalBlockHash` is preserved — the signature review continues in #3372.

## Related

- Review meeting: [pgm-a3-finish-bn 2026-08-04](https://swirldslabs.slack.com/archives/C08HECALMNF/p1785886793737959)
- #3299 (extension hasher slots — merged)
- #3367 (wrapper-vs-field-bytes, decided separately)
- #3372 (BlockHasher signature review, ongoing)
- HIP-1424, HIP-1056


common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -315,29 +316,43 @@ public static Bytes computeFinalBlockHash(
...
-        final byte[] rootOfTraceData = traceDataHasher.rootHash().join().toByteArray();
+                emptyToNull(consensusHeaderHasher.rootHash().join().toByteArray());
+        final byte[] rootOfInputs =
+                emptyToNull(inputTreeHasher.rootHash().join().toByteArray());
+        final byte[] rootOfOutputs =
+                emptyToNull(outputTreeHasher.rootHash().join().toByteArray());
+        final byte[] rootOfStateChanges =
+                emptyToNull(stateChangesHasher.rootHash().join().toByteArray());
+        final byte[] rootOfTraceData =
+                emptyToNull(traceDataHasher.rootHash().join().toByteArray());
```
jsync-swirlds: Perhaps I missed something elsewhere, but we need to ensure that if any node is filled in, we _also_ fill in every node "left" of that node.

We cannot just skip leaves, we always fill left-to-right, and include the hash of `0x00` for missing subtree leaves that precede the "rightmost" tree that has actual content.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -346,4 +361,37 @@ public static Bytes computeFinalBlockHash(
...
+            node = null;
+        } else if (left == null) {
+            node = hashInternalNodeSingleChild(right);
+        } else if (right == null) {
+            node = hashInternalNodeSingleChild(left);
+        } else {
+            node = hashInternalNode(left, right);
+        }
+        return node;
+    }
```
jsync-swirlds, line 390: There is a subtle requirement here that the "left" value _cannot_ be null if the "right" value is non-null.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
```diff
@@ -192,35 +194,45 @@ private static BlockHashResult hashBlockInternal(BlockUnparsed block) throws Exc
         // combine all the merkle tree roots and other block data into final block hash
         // spotless:off
         // Code here won't be formatted by Spotless, Spotless makes it less readable
-        // Capture intermediate hashes before folding them into the tree
+        // Capture intermediate hashes before folding them into the tree.
+        // Each hasher whose subtree had no items this block returns EMPTY_TREE_HASH from its
+        // computeRootHash(); emptyToNull converts that sentinel to null so combineOptionalNodes
+        // propagates absence up the tree instead of contributing EMPTY_TREE_HASH as a real value
+        // at that leaf position (issue #3377, matching the block-node's HashingUtilities).
```
jsync-swirlds: Also, we cannot propagate a null if _anything_ to the "right" of that value is non-null.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/HashingUtils.java
```diff
@@ -110,4 +110,56 @@ public static byte[] hashInternalNode(
...
+     * </ul>
+     * Matches the block-node's
+     * {@code org.hiero.block.common.hasher.HashingUtilities#combineOptionalNodes}. Used to keep
+     * empty streaming-hasher subtrees out of the block root hash entirely, rather than
+     * contributing {@link #EMPTY_TREE_HASH} as a real value at their leaf position (issue #3377).
+     *
+     * @param digest the MessageDigest instance to use for hashing (should be SHA-384)
+     * @param left the left child hash, or {@code null} when absent
+     * @param right the right child hash, or {@code null} when absent
+     * @return the node hash, or {@code null} when both children are absent
```
jsync-swirlds: We cannot have null left of a non-null leaf.
It is not as clear-cut as "everything is optional" because if we don't fill in leaves left-to-right, then we move the subtrees around, and that makes stable hashing and state proofs nigh impossible.

jsync-swirlds (commented):
I think a requirement got missed in how this was presented.

Also, we already have the correct algorithm implemented in hashing utilities.  The entire point of this change is to ensure we use the _exact same streaming hasher algorithm_ for the Merkle Mountain Top that we do for each subtree.

We should not need a different combining process.

The only complex code change needed is to ensure subtrees are filled left-to-right, never skipping a subtree, and using the EMPTY hash for empty subtrees but _stopping_ at the last non-empty subtree (not including empty subtrees to the "right" of the last non-empty subtree).

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: Addressed in ae151ed2d. Dropped `emptyToNull` entirely. The rewrite feeds leaves left-to-right into a `NaiveStreamingTreeHasher` up to the rightmost non-`EMPTY_TREE_HASH` position; interior empties contribute `EMPTY_TREE_HASH` (fill-left-to-right invariant preserved), only trailing empties drop out.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: Addressed in ae151ed2d. `combineOptionalNodes` deleted from both block-node and tools HashingUtilities. The rewrite reuses `NaiveStreamingTreeHasher` for the Merkle Mountain Top, so no null-handling primitive at that level and the 'left cannot be null if right is non-null' invariant is enforced by construction (leaves are added positionally in order).

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
rockysingh: Addressed in ae151ed2d. Same rewrite mirrored on the tools side. `BlockStreamBlockHasher.hashBlockInternal` now feeds leaves left-to-right into `StreamingHasher` (using `EMPTY_TREE_HASH` for interior empties, skipping trailing empties), then combines with the timestamp leaf via `hashInternalNode`. No null propagation anywhere.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/HashingUtils.java
rockysingh: Addressed in ae151ed2d. This method was deleted from the tools `HashingUtils` in the rewrite. The tools side now reuses the existing `StreamingHasher` for the Merkle Mountain Top computation, same shape as the block-node side.

rockysingh:
Thanks for the correction @jsync-swirlds — the "reuse the streaming hasher we already have" reframe was the right one. I had over-engineered a separate combining process when the primitive we already needed was `NaiveStreamingTreeHasher` itself. Pushed the rewrite as ae151ed2d.

**What changed since the last pass:**

- Dropped `combineOptionalNodes(byte[], byte[])` on both sides (block-node `HashingUtilities` + tools `HashingUtils`)
- Dropped `emptyToNull(byte[])` on both sides — no longer needed
- Dropped the `hashInternalNodeSingleChild(depth3Node1)` reserved-subtree wrap at depth 2
- `computeFinalBlockHash` and `BlockStreamBlockHasher.hashBlockInternal` now: materialize the 8 leaves in fixed position, find rightmost non-`EMPTY_TREE_HASH`, feed positions `0..rightmost` into a fresh `NaiveStreamingTreeHasher` / `StreamingHasher` (using `EMPTY_TREE_HASH` for interior empties, skipping trailing empties entirely), take `rootHash()`, combine with the timestamp leaf via `hashInternalNode`

Your left-to-right invariant is preserved: interior empty positions (an empty hasher with any non-empty leaf to its right) contribute `EMPTY_TREE_HASH` and are hashed into the tree normally. Only trailing empty positions (those with nothing non-empty to their right) drop out and reshape the tree.

**All four of your inline points addressed** (see the inline reply threads for the per-comment resolution).

**Wider blast radius flagged:** removing the reserved-subtree wrap means every block hash changes under this algorithm, not just the empty-hasher edge case. This aligns hopefully.

Draft still. Happy to un-draft once you're happy with the algorithm and we settle the fixture-regen sequencing.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -315,35 +316,53 @@ public static Bytes computeFinalBlockHash(
...
+        final byte[][] mountainTopLeaves = new byte[][] {
+            previousBlockHash.toByteArray(),
+            rootHashOfAllPreviousBlockHashes.toByteArray(),
+            stateRootHash,
+            consensusHeaderHasher.rootHash().join().toByteArray(),
+            inputTreeHasher.rootHash().join().toByteArray(),
+            outputTreeHasher.rootHash().join().toByteArray(),
+            stateChangesHasher.rootHash().join().toByteArray(),
+            traceDataHasher.rootHash().join().toByteArray()
+        };
```
ata-nas: I think we should also be adding the extensions, right?

rockysingh: I will add it in now good spot.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -315,35 +405,54 @@ public static Bytes computeFinalBlockHash(
...
+        for (int i = 3; i < mountainTopLeaves.length; i++) {
+            if (!Arrays.equals(mountainTopLeaves[i], EMPTY_TREE_HASH)) {
+                rightmostIncluded = i;
+            }
+        }
+        final NaiveStreamingTreeHasher mountainTopHasher = new NaiveStreamingTreeHasher();
+        for (int i = 0; i <= rightmostIncluded; i++) {
+            mountainTopHasher.addLeaf(ByteBuffer.wrap(mountainTopLeaves[i]));
+        }
+        final byte[] mountainTopRoot = mountainTopHasher.rootHash().join().toByteArray();
```
jsync-swirlds: There is a missing item after this point.
The mountain top root will vary based on how many leaves are included.

The actual height at which we join the timestamp to create the root hash must always be height 4.
So we need to add a single-child parent for each missing height level.

```text
2 or less -> height 1 : add 3 single-child parents
3 or 4    -> height 2 : add 2 single-child parents
5-8       -> height 3 : add 1 single-child parent
9-16      -> height 4 : no added parents
```

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
```diff
@@ -189,42 +190,43 @@ private static BlockHashResult hashBlockInternal(BlockUnparsed block) throws Exc
...
+            }
+        }
+        final StreamingHasher mountainTopHasher = new StreamingHasher();
+        for (int i = 0; i <= rightmostIncluded; i++) {
+            mountainTopHasher.addNodeByHash(mountainTopLeaves[i]);
+        }
+        final byte[] mountainTopRoot = mountainTopHasher.computeRootHash();
+        // Combine the consensus-timestamp leaf with the mountain-top root. Both always present,
+        // so this is a plain 0x02-prefix internal-node combine.
+        final byte[] rootHash = hashInternalNode(digest, ctHash, mountainTopRoot);
```
jsync-swirlds: Isn't this somewhat duplicated from the HashingUtilities?

jsync-swirlds (commented):
A couple questions.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: I now always feed all 16 leaves to the streaming hasher, using EMPTY_TREE_HASH for empty pre-defined subtrees and null-normalized extensions. Because 16 is a power of two the resulting mountain top is a fully balanced height-4 tree by construction, so no explicit single-child promotion is needed as per height of 4 requirement is satisfied naturally.

This also converges with the direction Atanas's #3299 will be moving toward (fixed 16-leaf tree structure) and satisfies the same left-to-right invariant you flagged on his `combineExtensionSubtreeRoots` at #3299 (interior nulls contribute `EMPTY_TREE_HASH` bytes rather than propagating null via 0x01 single-child promotion).

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
rockysingh: Correct, it's mirrored intentionally.  I Could extract a shared `computeMountainTopRoot(byte[][] leaves)` helper in HashingUtilities that takes raw byte arrays instead of hasher instances, and have both sides call it. That would cut the duplication to zero for the mountain-top logic (leaving just the per-side per-item routing, which is genuinely different: tools' hashBlockItemDirect vs block-node's getBlockItemHash).

Happy to do the extraction as a follow-up commit here if you'd like or as a separate cleanup PR to keep this one focused on the algorithm change. Preference?

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
jsync-swirlds: This isn't how the streaming hasher is supposed to work.

We _must not_ feed all 16 leaves, only the left edge (this is critical, if we feed null or empty leaves, they affect the hash and it will be incorrect).

We have to have single child nodes, not two child nodes, we have to have _missing_ leaves, not "fill in everything" leaves.

We were _already_ doing the "fill in all leaves" path, and that is why we have to make the change.

Always feeding 16 leaves is the opposite of what we're supposed to be changing.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
jsync-swirlds: Please do not add extensions here.  Let's add that only in the proto ordinal code, because otherwise it's far too easy to do this completely backwards and try to feed all 16 leaves every time, instead of only feeding in _just enough_ leaves to cover the subtrees that are not empty.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -425,17 +430,9 @@ public static Bytes computeFinalBlockHash(
...
-        for (int i = 3; i < mountainTopLeaves.length; i++) {
-            if (!Arrays.equals(mountainTopLeaves[i], EMPTY_TREE_HASH)) {
-                rightmostIncluded = i;
-            }
-        }
         final NaiveStreamingTreeHasher mountainTopHasher = new NaiveStreamingTreeHasher();
-        for (int i = 0; i <= rightmostIncluded; i++) {
-            mountainTopHasher.addLeaf(ByteBuffer.wrap(mountainTopLeaves[i]));
+        for (final byte[] leaf : mountainTopLeaves) {
+            mountainTopHasher.addLeaf(ByteBuffer.wrap(leaf));
```
jsync-swirlds: This is backwards, the prior (where we calculate the rightmost to include and only feed those in) is correct.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
```diff
@@ -190,13 +189,12 @@ private static BlockHashResult hashBlockInternal(BlockUnparsed block) throws Exc
...
-        // but which have a non-empty leaf to their right) contribute EMPTY_TREE_HASH. Trailing
-        // empties (empty hashers at the right edge) are not fed to the hasher at all. See issue
-        // #3377; the block-node's HashingUtilities.computeFinalBlockHash applies the same
-        // pattern producer-side.
+        // Always feed all 16 leaves to the streaming hasher: 8 pre-defined slots + 8 extension
+        // slots. Empty pre-defined hasher subtrees contribute EMPTY_TREE_HASH via
+        // StreamingHasher.computeRootHash(); extension slots are all EMPTY_TREE_HASH here
+        // because tools does not yet classify block items into extension categories. This
+        // produces a fixed-shape height-4 balanced tree, byte-identical to what block-node's
+        // HashingUtilities.computeFinalBlockHash produces for the same block. See issue #3377.
```
jsync-swirlds: This is completely backwards.
The entire change Jasper requested was to use the streaming algorithm and _not_ include subtree leaves to the right of the rightmost non-empty subtree.

Always including 16 leaves is the opposite of what was requested.

jsync-swirlds (changes requested):
Not sure how we went from "we need to not include all 8 left-hand leaves, and instead only include just enough leaves to reach the left-most non-empty subtree, and _not hash_ the remaining leaves" to "include all 16 leaves all the time".

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: Reverted in v4. Positions 0-7 now use rightmost-scan; only leaves up to the rightmost non-empty subtree are fed, interior empties contribute EMPTY_TREE_HASH.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: Removed. Extensions stay in #3299 only; this PR touches the pre-defined side.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: Back to rightmost-scan in v4.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
rockysingh: Tools side matches now: rightmost-scan on 0-7, single-child promote to height 3.

rockysingh:
v4 pushed: rightmost-scan on positions 0-7, single-child promote to height 3, extensions untouched (#3299). Fixture verifier tests will diverge until consensus lands the matching change and blocks are regenerated.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: The extensions are in main I believe via the other PR.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -394,28 +395,44 @@ public static Bytes computeFinalBlockHash( // @todo(3372) re-evaluate this signa
...
+            previousBlockHash.toByteArray(),
+            rootHashOfAllPreviousBlockHashes.toByteArray(),
+            stateRootHash,
+            consensusHeaderHasher.rootHash().join().toByteArray(),
+            inputTreeHasher.rootHash().join().toByteArray(),
+            outputTreeHasher.rootHash().join().toByteArray(),
+            stateChangesHasher.rootHash().join().toByteArray(),
+            traceDataHasher.rootHash().join().toByteArray()
+        };
+        // Positions 0-2 always contribute so 2 is the floor for the rightmost scan.
```
jsync-swirlds: 0 and 1 are required, position (2) is not (for example, I do not think WRBs have a state root value, do they?)

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -394,28 +395,44 @@ public static Bytes computeFinalBlockHash( // @todo(3372) re-evaluate this signa
...
+        }
+        final NaiveStreamingTreeHasher preDefinedHasher = new NaiveStreamingTreeHasher();
+        for (int i = 0; i <= rightmostIncluded; i++) {
+            preDefinedHasher.addLeaf(ByteBuffer.wrap(preDefinedLeaves[i]));
+        }
+        byte[] depth3Node1 = preDefinedHasher.rootHash().join().toByteArray();
+        // Canonicalize to height 3 via single-child parents so the pre-defined side always joins
+        // the extension side at the same depth regardless of how many leaves were populated.
+        // Streaming-hasher output heights by leaf count: 3-4 leaves -> height 2, 5-8 -> height 3.
+        // Positions 0-2 always populated so we always have >= 3 leaves.
```
jsync-swirlds: Small clarification that will help when we add the extensions:

```suggestion
        // Canonicalize to height 3 via single-child parents so the subtrees internal root always joins
        // the timestamp leaf at the same depth regardless of how many leaves were populated.
        // Streaming-hasher output heights by leaf count: 3-4 leaves -> height 2, 5-8 -> height 3.
        // Positions 0-1 always populated so we always have >= 2 leaves.
```

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -394,28 +395,44 @@ public static Bytes computeFinalBlockHash( // @todo(3372) re-evaluate this signa
...
+        final int preDefinedHeight = rightmostIncluded < 4 ? 2 : 3;
```
jsync-swirlds: It doesn't hurt to account for the 2 leaf case, but not necessary either:

```suggestion
        final int preDefinedHeight = rightmostIncluded < 2 ? 1 : rightmostIncluded < 4 ? 2 : 3;
```

FYI, also, we'll be off-by-one here in all cases if we have extension tree content; but, again, that will be resolved in a future PR.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
jsync-swirlds: Just to be clear, this is wrong currently, we'll fix it in a separate PR.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
```diff
@@ -189,42 +190,45 @@ private static BlockHashResult hashBlockInternal(BlockUnparsed block) throws Exc
...
+            previousBlockHash,
+            blockFooter.rootHashOfAllBlockHashesTree().toByteArray(),
+            stateRootHash,
+            hashers.consensusHeaders.computeRootHash(),
+            hashers.inputItems.computeRootHash(),
+            oitHash,
+            hashers.stateChangeItems.computeRootHash(),
+            hashers.traceItems.computeRootHash()
+        };
+        int rightmostIncluded = 2;
```
jsync-swirlds: Given this is hashing only WRB, couldn't we start at subtree `5` (the "output items")?

jsync-swirlds (dismissed):
Much improved, a few small items.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
jsync-swirlds: They are, they're still the old approach, but we will update those in a separate PR.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
berryware, line 190: It looks like the stateRootHash in HashingUtilities.java and BlockStreamBlockHasher.java have different default values for stateRootHash. Is the default value ever actually assigned?

berryware (dismissed):
Added one comment on the stateRootHash default values not being the same. Approving in the event this is not an issue

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
jsync-swirlds: We should use the same value regardless. Any missing subtree should be the empty hash constant.

This value here is correct.

common/src/test/java/org/hiero/block/common/hasher/HashingUtilitiesTest.java
```diff
@@ -151,25 +229,54 @@ private static byte[] randomHash() {
...
-                inputs.stateChangesHasher().rootHash().join().toByteArray();
-        final byte[] traceRoot = inputs.traceDataHasher().rootHash().join().toByteArray();
-        final byte[] leftHalf = refNode(
-                refNode(
-                        refNode(
-                                inputs.previousBlockHash().toByteArray(),
-                                inputs.rootOfAllPreviousBlockHashes().toByteArray()),
-                        refNode(inputs.startOfBlockStateRootHash().toByteArray(), consensusRoot)),
-                refNode(refNode(inputsRoot, outputsRoot), refNode(stateChangesRoot, traceRoot)));
+        final byte[] stateRoot = inputs.startOfBlockStateRootHash().length() == 0
```
Nana-EC, line 288: This differs from what CLI is doing.
The test is matching that HashingUtilities is doing - new byte[HASH_SIZE].
Vs what the CLI does `EMPTY_TREE_HASH` i.e `SHA384(0x00)`

Nana-EC (commented):
Starting to go through.
Seen a diff on the fallback hash for stateRoot.
Still reviewing

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -394,28 +395,44 @@ public static Bytes computeFinalBlockHash( // @todo(3372) re-evaluate this signa
...
-
-        // Treat missing state root hash as zero hash, matching the CN convention
+        // Pre-defined side (leaf positions 0-7): feed only up to the rightmost non-empty leaf
+        // into a NaiveStreamingTreeHasher (same fold-up algorithm used within each subtree
+        // hasher). Trailing empty subtrees are not fed at all and drop out of the tree; interior
+        // empties contribute EMPTY_TREE_HASH so the left-to-right invariant holds. After the
+        // fold-up, wrap the result in single-child parents until height 3, so the pre-defined
+        // side always joins the extension side (depth3Node2) at a fixed height. See issue #3377.
         final byte[] stateRootHash =
                 startOfBlockStateRootHash.length() == 0 ? new byte[HASH_SIZE] : startOfBlockStateRootHash.toByteArray();
```
jsync-swirlds: This should use the empty hash sentinel:

```suggestion
                startOfBlockStateRootHash.length() == 0 ? EMPTY_TREE_HASH : startOfBlockStateRootHash.toByteArray();
```

jsync-swirlds (dismissed):

Nana-EC (dismissed):

tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/model/hashing/BlockStreamBlockHasher.java
rockysingh: This has been corrected.

jsync-swirlds (dismissed):

jsync-swirlds (dismissed):

common/src/test/java/org/hiero/block/common/hasher/HashingUtilitiesTest.java
rockysingh: This is done.

jsync-swirlds:
New update, based on discussions offline and an algorithm correction that Jasper determined needs to be done.

1. We will use the same streaming hasher algorithm/utility for each subtree, and _also_ for the "Merkle Mountain Top" tree.
1. We will feed _all 16 leaves_ to the "Merkle Mountain Top" streaming hasher every time.
1. If a subtree has no content, then it is replaced with the `EMPTY_TREE_HASH` which is `hash(0x00)`.

There will be no unary nodes, no sentinels, no omissions, and no modification to the root hash or addition of internal nodes.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
jsync-swirlds: We should remove this overload.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -394,103 +394,72 @@ public static Bytes computeFinalBlockHash( // @todo(3372) re-evaluate this signa
         Objects.requireNonNull(stateChangesHasher);
         Objects.requireNonNull(traceDataHasher);
```
jsync-swirlds: I would abandon the requireNonNulls; they'll be obvious enough without this.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -394,103 +394,72 @@ public static Bytes computeFinalBlockHash( // @todo(3372) re-evaluate this signa
...
+        mountainTopHasher.addLeaf(
+                ByteBuffer.wrap(traceDataHasher.rootHash().join().toByteArray()));
+        mountainTopHasher.addLeaf(ByteBuffer.wrap(padAbsent(extensionSubtreeRootZero)));
+        mountainTopHasher.addLeaf(ByteBuffer.wrap(padAbsent(extensionSubtreeRootOne)));
+        mountainTopHasher.addLeaf(ByteBuffer.wrap(padAbsent(extensionSubtreeRootTwo)));
+        mountainTopHasher.addLeaf(ByteBuffer.wrap(padAbsent(extensionSubtreeRootThree)));
+        mountainTopHasher.addLeaf(ByteBuffer.wrap(padAbsent(extensionSubtreeRootFour)));
+        mountainTopHasher.addLeaf(ByteBuffer.wrap(padAbsent(extensionSubtreeRootFive)));
+        mountainTopHasher.addLeaf(ByteBuffer.wrap(padAbsent(extensionSubtreeRootSix)));
+        mountainTopHasher.addLeaf(ByteBuffer.wrap(padAbsent(extensionSubtreeRootSeven)));
```
jsync-swirlds: Should treat all the leaves identically; so either all of them get `padAbsent` or none do.
If we remove the obsolete overload above, we won't need the `padAbsent` method at all.

jsync-swirlds (dismissed):
Looks good, just a few small improvements/cleanups.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: Done removed. All 5 callers now pass extension roots explicitly.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: Done

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
rockysingh: The padAbsent removed; extension args are @NonNull now.

block-node/app/src/testFixtures/resources/test-blocks/CN_11_12_TSS_WRAPS/0.blk.gz
jsync-swirlds, line 1: We have dynamic generation capability for generating valid blocks when needed in tests.
Could we replace these with new dynamically generated blocks?

block-node/verification/src/main/java/org/hiero/block/node/verification/session/impl/ExtendedMerkleTreeSession.java
jsync-swirlds, line 1: We need to make a note to remove this plugin soon.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -376,121 +320,69 @@ public static Bytes computeFinalBlockHash( // @todo(3372) re-evaluate this signa
...
-
-        // Treat missing state root hash as zero hash, matching the CN convention
+            @NonNull final byte[] extensionSubtreeRootZero,
+            @NonNull final byte[] extensionSubtreeRootOne,
+            @NonNull final byte[] extensionSubtreeRootTwo,
+            @NonNull final byte[] extensionSubtreeRootThree,
+            @NonNull final byte[] extensionSubtreeRootFour,
+            @NonNull final byte[] extensionSubtreeRootFive,
+            @NonNull final byte[] extensionSubtreeRootSix,
+            @NonNull final byte[] extensionSubtreeRootSeven) {
```
jsync-swirlds: I think I missed explaining this before.
These should probably be the streaming tree hashers; just like the other 8 leaves.

block-node/app/src/testFixtures/resources/test-blocks/CN_11_12_TSS_WRAPS/0.blk.gz
jsync-swirlds: To be entirely clear, the goal here is to not store block files in the code repo.

rockysingh: We good to follow this up in a seperate PR and we can remove the block files. At the moment we have block files.

jsync-swirlds: As long as we follow up relatively quickly. Every update to these files is another increment of additional data that every clone must download, forever, regardless of whether we later remove the files.

There are large projects on Github that outright cannot be cloned because they included too many binary files in the code repo that were later removed or replaced and the unused data takes long enough that github fails to complete the download.

I don't think we're anywhere near that volume, but we should still be careful.

jsync-swirlds (dismissed):

block-node/app/src/testFixtures/resources/test-blocks/CN_11_12_TSS_WRAPS/0.blk.gz
rockysingh: That is a good point.

jsync-swirlds (dismissed):

block-node/block-access-service/src/test/java/org/hiero/block/node/access/service/BlockAccessServicePluginTest.java
```diff
@@ -171,7 +171,7 @@ void testInvalidRequest() throws ParseException {
     @Test
     @DisplayName("TSS Wraps Transition Block (466) can be retrieved from BlockAccessService")
     void testGetTssWrapsLargeBlock() throws ParseException, IOException {
-        final BlockUtils.SampleBlockInfo info = BlockUtils.getSampleBlockInfo(BlockUtils.SAMPLE_BLOCKS.BLOCK_466);
+        final BlockUtils.SampleBlockInfo info = BlockUtils.getSampleBlockInfo(BlockUtils.SAMPLE_BLOCKS.BLOCK_322);
```
Nana-EC, line 174: You update the block here but the assertion at the end was not updated and still looks for block 466. If this is passing now something else is off.
Update should probably use `info.blockNumber()` instead of 466 or 322

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillPluginTest.java
```diff
@@ -773,7 +773,7 @@ void testFailedPersistenceBeforeStartIsHandledGracefully() {
     @DisplayName("On-Demand Backfill - TSS Wraps Transition Block (466)")
     void testBackfillOnDemandTssWrapsBlock() throws InterruptedException, IOException, ParseException {
         final BlockUtils.SampleBlockInfo tssBlockInfo =
-                BlockUtils.getSampleBlockInfo(BlockUtils.SAMPLE_BLOCKS.BLOCK_466);
+                BlockUtils.getSampleBlockInfo(BlockUtils.SAMPLE_BLOCKS.BLOCK_322);
```
Nana-EC, line 776: Same issue of missing test updates noted in block-node/block-access-service/src/test/java/org/hiero/block/node/access/service/BlockAccessServicePluginTest.java here also

Nana-EC (commented):
LG. Just inconsistencies in tests to watch out for and confirm

block-node/block-verification/src/test/java/org/hiero/block/node/block/verification/verifier/TSSVerifierTest.java
```diff
@@ -77,10 +77,10 @@ void testShouldVerifyTssWrapsBlock0BeforeSettled() throws IOException, ParseExce
     @DisplayName("verify() successful verification block 467 settled path")
```
ata-nas, line 77: Block is now 391 instead of 467

block-node/block-access-service/src/test/java/org/hiero/block/node/access/service/BlockAccessServicePluginTest.java
rockysingh: Fixed — assertion now uses `info.blockNumber()`.

block-node/backfill/src/test/java/org/hiero/block/node/backfill/BackfillPluginTest.java
rockysingh: Fixed — same pattern, uses `tssBlockInfo.blockNumber()` throughout.

Nana-EC (commented):
Change is solid.
Just test coverage breaks
Node operator CI - https://github.com/hiero-ledger/hiero-block-node/actions/runs/31533670002/job/93919509758?pr=3378 If you fix the link to the blocks it should pass and give you more coverage

The other 2 CI failures seems like a flaky test and a known docs issue with Lychee

rockysingh:
@Nana-EC This has been fixed.

jsync-swirlds (dismissed):

.github/workflows/e2e-node-operator.yaml
```diff
@@ -395,8 +409,13 @@ jobs:
           sed "s#oci://ghcr.io/hiero-ledger/hiero-block-node/block-node-server#oci://${CHART_REGISTRY}/hiero-ledger/hiero-block-node/block-node-server#" \
             tools-and-tests/scripts/e2e-node-operator/solo-ci-config.yaml > /tmp/solo-ci-config-snapshot.yaml
           # Render CI values for the snapshot phase (image tag + plugin profile); volumes use chart defaults.
+          # Force PLUGIN_NAMES empty for the snapshot phase: this skips the Maven resolve-plugins
```
Nana-EC: Why is this change needed?
I recently fixed this to support the dreaded Maven pull. This PR shouldn't require any improvements here? Did you hit an issue we need to resolve?

Nana-EC (dismissed):
LG
Let me know about node operator CI update.
Not sure why the Maven targeted fix is needed but I understand the need for both set's of blocks to manage the breaking change in verification

.github/workflows/e2e-node-operator.yaml
rockysingh: I hit a real issue with the snapshot phase’s Maven resolve was pulling `block-verification-0.41.0-SNAPSHOT.jar ` from Sonatype Snapshots, but that published JAR is stale relative to this branch. It’s compiled against the old `HashingUtilities` signature with` byte[]` extension args.

As a result, verification fails at runtime with a `NoSuchMethodError` against the new `StreamingTreeHasher` signature.

Setting plugins.names="" for the snapshot phase skips the init container, so the pod uses the fresh plugins already baked into the locally built solo-dev image.

Happy to revert this if your recent fix should have covered it. Do you know how a locally built snapshot is supposed to reach the Maven resolve step instead of it pulling the stale artifact from Sonatype?

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/ResourceTestBlockBuilder.java
```diff
@@ -43,38 +43,38 @@ public enum WRAPS implements ResourceBlock {
...
+                "a6d563cf11b04e5884958223169cbdd080f27425628584db76b0447982e869ebac08e9f8ecaa03d6b8aa225e8e188eb4",
                 4),
-        /// Transition block — first block with WRAPS signature (Schnorr to Wraps transition, oversized ~13MB).
-        BLOCK_466(
-                "CN_0_73_TSS_WRAPS/466.blk.gz",
-                "ad532f179da5abfc1f982a2a1dbc3d5c0c2e27b47126d559356a03fb656f81c862b3030ba13e1a98242e6390756d9c14",
-                466),
+        /// Transition block — first block with WRAPS signature (Schnorr to WRAPS transition).
+        BLOCK_322(
+                "CN_0_73_TSS_WRAPS/322.blk.gz",
```
ata-nas: Are these blocks 0.73 blocks? I think this is later versions now?

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/blocks/ResourceTestBlockBuilder.java
```diff
@@ -177,34 +177,40 @@ public NodeAddressBook loadAddressBook() throws IOException, ParseException {
...
     /// Sample blocks containing state proofs from a hapiTestWraps capture with Schnorr TSS signatures.
-    /// Every 4th block (0, 4, ...) is directly signed; blocks in between have state proofs.
-    /// Block 0 contains LedgerIdPublicationTransactionBody for TSS initialization.
+    /// Every 5th block (0, 5, ...) is directly signed with Schnorr; blocks in between carry
+    /// state proofs referencing the next signed block. Block 0 contains
+    /// LedgerIdPublicationTransactionBody for TSS initialization.
     public enum StateProof implements ResourceBlock {
         /// Genesis block — bootstraps TSS parameters and ledger ID. Direct Schnorr proof.
         BLOCK_0(
                 "CN_0_77_TSS_SCHNORR/0.blk.gz",
```
ata-nas: Are these blocks 0.77 blocks? If not, we can change the version semantics here.

block-node/app/src/testFixtures/resources/test-blocks/CN_11_12_TSS_WRAPS/0.blk.gz
ata-nas: I agree that we should migrate from these blocks, it has been on our radar. I also agree that we should tackle this separately.

common/src/main/java/org/hiero/block/common/hasher/HashingUtilities.java
```diff
@@ -376,121 +320,72 @@ public static Bytes computeFinalBlockHash( // @todo(3372) re-evaluate this signa
...
-            final byte[] rootSeven) {
-        final byte[] depth5Node5 = combineOptionalNodes(rootZero, rootOne);
-        final byte[] depth5Node6 = combineOptionalNodes(rootTwo, rootThree);
-        final byte[] depth5Node7 = combineOptionalNodes(rootFour, rootFive);
-        final byte[] depth5Node8 = combineOptionalNodes(rootSix, rootSeven);
-        final byte[] depth4Node3 = combineOptionalNodes(depth5Node5, depth5Node6);
-        final byte[] depth4Node4 = combineOptionalNodes(depth5Node7, depth5Node8);
-        return combineOptionalNodes(depth4Node3, depth4Node4);
+    private static ByteBuffer subtreeRootBuffer(final StreamingTreeHasher hasher) {
+        return ByteBuffer.wrap(hasher.rootHash().join().toByteArray());
```
ata-nas: As part of #3372, we wanted to revisit the `CompletableFuture` returned from the `rootHash` method. As of today, the naive hasher returns a completed future, so no actual async computation happens. We can decide to update that here, I think it is a good idea. If we ever want to add an async computation of the root hash, we can simply add an async version of the rootHash method.

Note this is optional for this PR.

ata-nas (commented):
Here just a few nits and cleanup suggestions. I like how this is turning out.

jsync-swirlds (dismissed):
It looks like we're starting to balloon this PR with added tests.

Can we cut this off, merge the working code, and add more tests in a separate PR?

jsync-swirlds (approved):

.github/workflows/e2e-node-operator.yaml
Nana-EC: NO need for a new push.
What's unclear is the current logic should already support this.
For the upgrade to snapshot it's not supposed to be pullign from maven for the snapshot version. It's supposed to use the solo-dev image that is built in CI and then the local cahrt to get the plugins, not pull from maven. Otherwise the test wouldn't work on confirming in PR changes don't break the lifecycle, so there's something off that we can look at in follow up

Nana-EC (approved):

PR #3409, opened 2026-08-11
test: Add Solo CI test for archiving coverage

### LINKED ISSUE #1711: Add Solo CI test for archiving coverage
### Story Form

As a block node engineer and performance engineer
I want to have easily repeatable tests that show cases block archiving scenarios
So that I can gain early confidence over a large real world Hiero network deployment

### Technical Notes

Use solo to deploy a network with the following networks distribution
- 2 CNs & 2 BNs
- CNs 1 streams to BN 1 primarily
- CNs 2 streams to BN 2 primarily
- BN 1 archives to a minio setup simulating S3
- BNs 1 to 2 backfill between each other as needed
- MN points to all BNs to pull blocks as needed
- MN pinger runs at about 10 TPS
- Eventually kill the RFH variant, allow some time to pass
- Replace BN 1 w a new BN 3 in the deployment with same configs
- Verify BN3 backfills from BN2 and that eventually archiving resumes

ivannov:
### New topology `2cn-2bn-archive` and test `archive-backfill.yaml` cover the full archiving + node-replacement scenario

The test deploys minio (as an S3 stand-in) and BN1/BN2, lets BN1 archive to it via cloud-storage-archive (TAR) and cloud-storage-expanded (.blk.zstd), kills BN1, deploys a replacement BN3 with the same archive config, redirects CN1 to it, and verifies BN3 backfills from BN2 and resumes archiving (both minio buckets gain files).

### New `plugin-profile-cloud.yaml` Helm values override

Adds a plugin list that layers `cloud-storage-archive` and `cloud-storage-expanded` onto the standard plugin set.

### `solo-deploy-minio.sh` uses the official MinIO chart

Bitnami's Docker Hub images are periodically rotated off, which made the chart unreliable in CI; the official chart's image lives on quay.io. Memory requests are capped to 512Mi/1Gi (chart default is 16Gi) since the Kind cluster cannot afford that. The script also seeds `minio-cloud-storage-creds`, the Secret both cloud plugins read their access/secret keys from.

## Related Issue(s)

Closes #1711


charts/block-node-server/values-overrides/plugin-profile-cloud.yaml
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: Apache-2.0
+# Plugin profile for block nodes with cloud storage archiving.
+# Enables cloud-storage-archive and cloud-storage-expanded alongside the standard plugin set.
+# blocks-file-historic is included because cloud-storage-archive reads blocks from local
+# persistence via BlockAccessor after receiving PersistedNotification.
+plugins:
+  names: "facility-messaging,block-access-service,health,server-status,stream-publisher,stream-subscriber,block-verification,blocks-file-historic,blocks-file-recent,backfill,cloud-storage-archive,cloud-storage-expanded"
```
jsync-swirlds: This is excessive, the cloud profile shouldn't have local archive plugins (cloud-storage-archive _does not_ read from local persistence).

```suggestion
  names: "facility-messaging,health,server-status,block-verification,backfill,cloud-storage-archive,cloud-storage-expanded"
```

tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-minio.sh
```diff
@@ -0,0 +1,96 @@
...
+# Options:
+#   --namespace NS        Kubernetes namespace (required)
+#   --release-name NAME   Helm release name (default: minio)
+#   --buckets LIST        Comma-separated bucket names to create on startup
+#                         (default: block-archive-tar,block-archive-expanded)
+
+set -eo pipefail
+
+NAMESPACE=""
+RELEASE_NAME="minio"
```
jsync-swirlds: Note, this is a risk item; minio is a defunct project and unsupported.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-network.sh
```diff
@@ -430,6 +430,62 @@ EOF
...
+    CLOUD_STORAGE_EXPANDED_ENDPOINT_URL: "http://${minio_service}:9000"
+    CLOUD_STORAGE_EXPANDED_REGION_NAME: "us-east-1"
+    CLOUD_STORAGE_EXPANDED_BUCKET_NAME: "block-archive-expanded"
+    CLOUD_STORAGE_EXPANDED_STORAGE_CLASS: "STANDARD"
+    CLOUD_STORAGE_EXPANDED_OBJECT_KEY_PREFIX: ""
+  persistence:
+    archive:
+      size: 10Gi
+    live:
+      size: 10Gi
```
jsync-swirlds: These should not be present, there is no local file storage.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
```diff
@@ -392,6 +398,46 @@ EOF
...
+    CLOUD_STORAGE_EXPANDED_ENDPOINT_URL: "http://${minio_service}:9000"
+    CLOUD_STORAGE_EXPANDED_REGION_NAME: "us-east-1"
+    CLOUD_STORAGE_EXPANDED_BUCKET_NAME: "block-archive-expanded"
+    CLOUD_STORAGE_EXPANDED_STORAGE_CLASS: "STANDARD"
+    CLOUD_STORAGE_EXPANDED_OBJECT_KEY_PREFIX: ""
+  persistence:
+    archive:
+      size: 10Gi
+    live:
+      size: 10Gi
```
jsync-swirlds: Here again, we should not have local storage plugins, so these should not be present.

charts/block-node-server/values-overrides/plugin-profile-cloud.yaml
ivannov: I removed the block access and the block file plugins. However, I will leave for the time being the publisher and subscriber as the task definition requires the block nodes to get blocks from consensus nodes and mirror nodes to get blocks from them.

Or maybe we should re-define the scenario in #1711?

tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-minio.sh
ivannov: Yes, right! Change to `rustfs`

tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-network.sh
ivannov: Removed

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
ivannov: Removed

charts/block-node-server/values-overrides/plugin-profile-cloud.yaml
jsync-swirlds: RFH doesn't receive blocks directly from CN; we run those only in Tier 2, so it uses backfill to read from the Tier 1 block nodes.

We also don't typically run subscriber for RFH because it doesn't have local storage to provide blocks if a client falls a little behind.

The issue might be out of date in that regard.

Nana-EC (commented):
Will circle back to review but take a look at the cloud-storage-expanded integration and E2E which used S3Mock.
This PR would move the repo to use 2 different tools.
We should first see if `S3Mock` does the job, if not does `RustFS`

ivannov:
> Will circle back to review but take a look at the cloud-storage-expanded integration and E2E which used S3Mock. This PR would move the repo to use 2 different tools. We should first see if `S3Mock` does the job, if not does `RustFS`

When I did the initial research, I asked three LLMs (Claude, Gemini and Chat GPT) for a replacement of Min IO. Adobe S3Mock wasn't there in some of the replies, the reason being that it is a test double, not a real S3 backend. That is why I left it out and the choice was between Garage, SeaweedFS and RustFS. I picked the latter as it was said to be closest to MinIO replacement, plus its license is Apache.

Now I asked again claude to assess whether we should have MockS3 for our unit / integration tests and RustFS for solo e2e, or either of them for both. Here is the reponse:

One backend: RustFS everywhere. Drop MinIO server image and drop Adobe S3Mock.

> Reasons:
>
> - S3Mock buys nothing here. Its only edge is `withInitialBuckets` ergonomics; it still needs Docker, so no docker-free speed win. And it is a mock — cloud-storage-archive exercises multipart upload, listMultipartUploads, abort; mock divergence there is silent risk. 
> - Same backend at unit and solo layer means a solo e2e S3 failure reproduces locally in a unit test.
> - Deletes a dependency plus its JPMS hack: `s3mock-testcontainers` has no module-info, so settings.gradle.kts:30 carries a synthetic module descriptor, gradle/modules.properties:19 an entry, and the version is hand-duplicated as a string constant in two test classes with a TODO: must match s3MockVersion (no way to share across classpath boundary). Plain `GenericContainer("rustfs/rustfs:<tag>")` erases all of it.

charts/block-node-server/values-overrides/plugin-profile-cloud.yaml
ata-nas: +1 here, cloud storage block nodes setup quite differently from local storage block nodes. I think what @jsync-swirlds suggests is valid.

tools-and-tests/scripts/solo-e2e-test/tests/archive-backfill.yaml
```diff
@@ -0,0 +1,208 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Archive Backfill Test (Issue #1711)
+#
+# Tests cloud-storage-archive (TAR) and cloud-storage-expanded (.blk.zstd) plugins
+# in a realistic multi-node deployment with node replacement and archiving resumption.
+#
+# Scenario:
```
Nana-EC, line 8: I think the scenario is a bit off to what we expect.
here's my suggestion

   1. BN1 archives to RustFS and does not receive from CN but backfills from BN2; BN2 receives from CN2 — verify both archive buckets fill
   2. Kill BN1 (archiving node); BN2 continues receiving from CN2
   3. Deploy BN3 with same RustFS archive config
   4. BN3 backfills from BN2 and resumes archiving to RustFS
   5. Verify both buckets have additional files (archiving resumed)

tools-and-tests/scripts/solo-e2e-test/topologies/2cn-2bn-archive.yaml
```diff
@@ -0,0 +1,50 @@
...
+  rustfs-1:
+    buckets: [block-archive-tar, block-archive-expanded]
+    port: 9000
+
+block_nodes:
+  block-node-1:
+    address: block-node-1
+    port: 40840
+    peers: [block-node-2]
+    greedy: false
```
Nana-EC:
```suggestion
    greedy: true
```

charts/block-node-server/values-overrides/plugin-profile-cloud.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
+# Plugin profile for block nodes with cloud storage archiving.
+# No local block-storage plugins: cloud-storage-archive/cloud-storage-expanded read block
+# bytes straight off VerificationNotification.
+plugins:
+  names: "facility-messaging,health,server-status,stream-publisher,stream-subscriber,block-verification,backfill,cloud-storage-archive,cloud-storage-expanded"
```
Nana-EC: Neither publish nor subscriber should be present
```suggestion
  names: "facility-messaging,health,server-status,block-verification,backfill,cloud-storage-archive,cloud-storage-expanded"
```

Nana-EC (commented):
A few things in the test setup needs to be tweaked

tools-and-tests/scripts/solo-e2e-test/tests/archive-backfill.yaml
jsync-swirlds: Also verify that BN3 continued from where BN1 left off before it was killed.

charts/block-node-server/values-overrides/plugin-profile-cloud.yaml
ivannov: Done. The profile is now: `facility-messaging,health,server-status,verification,backfill,cloud-storage-archive,cloud-storage-expanded`

tools-and-tests/scripts/solo-e2e-test/tests/archive-backfill.yaml
ivannov: The events and assertions were reworked to match. Added an `archive-contiguous` assertion on both buckets.

tools-and-tests/scripts/solo-e2e-test/topologies/2cn-2bn-archive.yaml
ivannov: Applied. BN1 is `greedy: true`, and BN3 is deployed with `greedy: true` as well

jsync-swirlds (dismissed):

jsync-swirlds (approved):

berryware (approved):
Looks good.

PR #3411, opened 2026-08-11
fix: BlockNodeClient Connection Cache should be private

berryware:
- added .shareConnectionCache(false) to the webclient builder so that a private, non-shared cache per client ensures eviction actually forces a fresh connection.

## Related Issue(s)

### DIFF
--- a/block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
+++ b/block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
@@ -186,6 +186,14 @@ public class BlockNodeClient implements AutoCloseable {
                 .protocolConfigs(List.of(buildHttp2Config(tuning), buildGrpcConfig(pollWaitMs, tuning)))
                 .connectTimeout(Duration.ofMillis(connectTimeoutMs))
                 .keepAlive(true)
+                // Helidon shares one JVM-wide HTTP/2 connection cache across all WebClient instances by
+                // default (shareConnectionCache=true). That defeats getNodeClient()'s evict-and-recreate
+                // logic: a "fresh" BlockNodeClient still points at the same shared cache, so if it holds a
+                // dead connection handler for this peer (e.g. the peer's server restarted), the new client
+                // reuses that stale entry instead of dialing a new connection — surfacing as a persistent
+                // SocketException: Socket closed on every retry. A private, non-shared cache per client
+                // ensures eviction actually forces a fresh connection.
+                .shareConnectionCache(false)
                 .build();
     }
 
--- a/block-node/base/src/test/java/org/hiero/block/node/base/client/BlockNodeClientTest.java
+++ b/block-node/base/src/test/java/org/hiero/block/node/base/client/BlockNodeClientTest.java
@@ -2,6 +2,7 @@
 package org.hiero.block.node.base.client;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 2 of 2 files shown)


ivannov (approved):

block-node/base/src/test/java/org/hiero/block/node/base/client/BlockNodeClientTest.java
```diff
@@ -74,6 +75,21 @@ void statusFallsBackToPortNotSubscribePort() throws IOException {
...
+    @Test
+    @DisplayName("WebClients do not share Helidon's JVM-wide HTTP/2 connection cache")
+    void connectionCacheIsNotShared() throws IOException {
+        // Helidon's HttpClientConfigBlueprint defaults shareConnectionCache to true, which would let a
+        // freshly-constructed client (e.g. after getNodeClient() evicts an unreachable peer) reuse a
+        // stale connection handler for the same host:port from the JVM-wide Http2ConnectionCache.SHARED
+        // singleton instead of dialing a new connection — surfacing as a persistent
+        // "SocketException: Socket closed" after the peer restarts. Each client must use its own
+        // private cache so eviction actually forces a fresh connection.
+        try (BlockNodeClient client = clientFor(40840, 0, 0)) {
```
Nana-EC, line 87: We're really just asserting the config is set.
Is this because it's difficult to test that it doesn't hold onto a stale connection?

Nana-EC (commented):
Good catch. Test case seems limited. Is it intentional?

block-node/base/src/test/java/org/hiero/block/node/base/client/BlockNodeClientTest.java
jsync-swirlds: We'd need to have an E2E test for that; unit tests shouldn't be launching "remote" services.
I agree this isn't testing anything meaningful; we probably just need a test with chaos mesh so we can open a connection, block it in the network layer (so it's a zombie connection), and verify the refresh reconnects.

berryware: The tests inside the JVM clean up the connections and close the sockets and the error does not happen. It is specific to using k8s to restart a Tier-0 BN and the Tier-1 BNs having stale connections. The E2E test will test this.

block-node/base/src/main/java/org/hiero/block/node/base/client/BlockNodeClient.java
```diff
@@ -186,6 +186,14 @@ private WebClient buildWebClient(
...
                 .connectTimeout(Duration.ofMillis(connectTimeoutMs))
                 .keepAlive(true)
+                // Helidon shares one JVM-wide HTTP/2 connection cache across all WebClient instances by
+                // default (shareConnectionCache=true). That defeats getNodeClient()'s evict-and-recreate
+                // logic: a "fresh" BlockNodeClient still points at the same shared cache, so if it holds a
+                // dead connection handler for this peer (e.g. the peer's server restarted), the new client
+                // reuses that stale entry instead of dialing a new connection — surfacing as a persistent
+                // SocketException: Socket closed on every retry. A private, non-shared cache per client
+                // ensures eviction actually forces a fresh connection.
+                .shareConnectionCache(false)
```
jsync-swirlds, line 196: This probably should be configured so we run it this way in _tests_ but not in production.
In production, non-shared connection cache has a substantial impact on memory and performance, particularly in high latency conditions.

berryware:
Closing as not needed. The E2E tests were not propagating the BN_VERSION correctly and the wrong BlockNode version was used. Causing the bad connection eviction fix to not get picked up. Once corrected, the connections were getting reset correctly.

Issue #3236, opened 2026-07-16
Dynamic File Recent Retention Policy


Nana-EC:
### Story Form

As a block node operator
I want to set a dynamic retention policy on files recent plugin
So that I avoid cases of running out of disk space based on a static block number retention policy

### Technical Notes

Dynamic nature can be based on age or size on disk of live folder

ivannov:
As discussed, the free space check will not be done automatically. Documentation was updated for the operators to introduce alerts of their own

PR #3412, opened 2026-08-11
docs: Dynamic File Recent Retention Policy

ivannov:
### Steady-state operations guide documents the disk-space gap left by count-based retention

`files.recent.blockRetentionThreshold` caps the `live` volume by block count, not by disk size, so a spike in block size or a threshold set too high for the provisioned volume can fill the disk before the count cap ever triggers. 

Since free disk space is not a metric that the Block Node report itself, the doc tells operators to alert on it externally, suggesting an alert threshold of 15% free space, and explains the failure mode so operators know why the alert matters.

## Related Issue(s)
  
Relates to #3236

### DIFF
--- a/docs/block-node/operations/production-guide/steady-state-operations.md
+++ b/docs/block-node/operations/production-guide/steady-state-operations.md
@@ -132,6 +132,22 @@ on-call contact provided at handoff.
 
 ---
 
+## Disk space alerting **[OPERATOR]**
+
+The `live` volume's retention policy is a fixed block-count cap
+(`files.recent.blockRetentionThreshold`), not a disk-size cap. If block sizes grow or the configured
+count is set too high for the provisioned volume, disk usage can reach 100% before the
+block-count cap is ever hit.
+
+There is no in-app metric for free disk space; it is host/volume-level, not something the
+Block Node process reports. Monitor it independently (e.g. `node_filesystem_avail_bytes`
+from node-exporter, or your platform's volume-usage metric for the `live` mount) and alert
+operators before the volume fills, for example when free space drops below 15%. Filling
+the `live` volume causes block writes to fail, which stalls ingestion until space is freed
+or the volume is expanded.
+
+---
+
 ## Related documentation
 
 - [Resetting and Upgrading the Block Node](../resetting-and-upgrading-the-block-node.md)
--- a/docs/block-node/operations/production-guide/steady-state-operations.md
+++ b/docs/block-node/operations/production-guide/steady-state-operations.md
@@ -134,17 +134,21 @@ on-call contact provided at handoff.
 
 ## Disk space alerting **[OPERATOR]**
 
-The `live` volume's retention policy is a fixed block-count cap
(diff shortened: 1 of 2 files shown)


docs/block-node/operations/production-guide/steady-state-operations.md
```diff
@@ -132,6 +132,22 @@ on-call contact provided at handoff.
...
+(`files.recent.blockRetentionThreshold`), not a disk-size cap. If block sizes grow or the configured
+count is set too high for the provisioned volume, disk usage can reach 100% before the
+block-count cap is ever hit.
+
+There is no in-app metric for free disk space; it is host/volume-level, not something the
+Block Node process reports. Monitor it independently (e.g. `node_filesystem_avail_bytes`
+from node-exporter, or your platform's volume-usage metric for the `live` mount) and alert
+operators before the volume fills, for example when free space drops below 15%. Filling
+the `live` volume causes block writes to fail, which stalls ingestion until space is freed
+or the volume is expanded.
```
Nana-EC: Good note.
Please add the specific mentioned in docs/block-node/metrics.md
See at the end our full set of recommended alerts are there. You should add to that also

Nana-EC (commented):
Add an explicit mention in the metrics.md

docs/block-node/operations/production-guide/steady-state-operations.md
```diff
@@ -132,6 +132,22 @@ on-call contact provided at handoff.
 
 ---
 
+## Disk space alerting **[OPERATOR]**
+
+The `live` volume's retention policy is a fixed block-count cap
```
ata-nas: We only seem to mention `live` here, but `historic` actually also has retention policy. By default, it is disabled. We can also describe `historic` interactions as well.

docs/block-node/operations/production-guide/steady-state-operations.md
```diff
@@ -132,6 +132,22 @@ on-call contact provided at handoff.
...
+(`files.recent.blockRetentionThreshold`), not a disk-size cap. If block sizes grow or the configured
+count is set too high for the provisioned volume, disk usage can reach 100% before the
+block-count cap is ever hit.
+
+There is no in-app metric for free disk space; it is host/volume-level, not something the
```
ata-nas, line 144: We actually do have a metric for how much space we have used. This is dynamically calculated, but should be viewed as a best-effort and not an ultimately reliable one.

I think we also have that metric for `historic` as well, but don't quote me on that.

ivannov: I would not mention that metric in the documentation, so that operators are not tempted to use it. In the recent commit I added the more precise metric that they should use

docs/block-node/operations/production-guide/steady-state-operations.md
ivannov: Included the historic plugin's directory as well

ata-nas (approved):

jsync-swirlds (approved):

PR #3413, opened 2026-08-11
fix(ci): switch lychee to musl binary to fix glibc incompatibility

deshmukhpranali:
## Reviewer Notes

The lychee-action downloads a GNU-linked binary that requires GLIBC_2.38 and GLIBC_2.39, which are not available on the hl-bn-lin-md self-hosted runner. Switch to the statically-linked musl binary (same lychee v0.24.2), which has no glibc dependency and runs on any Linux system.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -18,8 +18,27 @@ permissions:
   contents: read
 
 jobs:
+  setup-lychee:
+    name: Setup lychee
+    runs-on: hl-bn-lin-md
+    steps:
+      - name: Harden the runner
+        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/link-check.yml
```diff
@@ -55,11 +54,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+      - name: Install lychee
```
Nana-EC: Move the Lychee install to a single job, then have these two jobs depend on that. The job should check if lychee exists, if so no-op. If not install.
This way you install once only when needed

Nana-EC (commented):
Good call. I provided a suggestion

jsync-swirlds (approved):

Nana-EC (approved):

PR #3426, opened 2026-08-12
feat(tools): add block-access-port and fix ResourceExhausted in bn-endpoint-checker.sh

### LINKED ISSUE #3423: bug(tools): bn-endpoint-checker.sh --latest-block-proof fails with ResourceExhausted on large blocks
## Description

When running `bn-endpoint-checker.sh` with `--latest-block-proof` against a node that is serving large blocks, the `getBlock` call fails with:

```
🟠 latest block proof WARN (getBlock unavailable or no blocks stored)  (1.3 s)
   ERROR:
     Code: ResourceExhausted
     Message: grpc: received message larger than max (17478344 vs. 4194304)
```

The error originates on the **client side** (`grpcurl`), which enforces a hard 4 MiB (4,194,304 byte) limit on incoming gRPC messages by default. Blocks in production can exceed this limit (17+ MiB observed).

The script passes no `-max-msg-sz` flag to `grpcurl` in `grpc_call_block_proof`, so it inherits the 4 MiB default, making `--latest-block-proof` unreliable on any deployment that stores large blocks.

## Steps to Reproduce

1. Run the script with `--latest-block-proof` against a mainnet or previewnet node that is serving heavy transaction epochs:
   ```bash
   bn-endpoint-checker.sh --version 0.40.0 --latest-block-proof \
     node1.mainnet.blocknode.hashgraph-devops.com:40840
   ```
2. Observe the 🟠 warning with `ResourceExhausted / grpc: received message larger than max`.

## Additional context

- `grpcurl` exposes a `-max-msg-sz <bytes>` flag to override this limit.
- The `serverStatus`/`serverStatusDetail` calls in `grpc_call` are not affected because their responses are tiny.
- Only `grpc_call_block_proof` (which calls `BlockAccessService/getBlock`) is impacted.
- Suggested fix: raise the default to 64 MiB and optionally expose a `--max-block-sz BYTES` flag so operators can tune it without editing the script.

### LINKED ISSUE #3424: feat(tools): add --block-access-port flag to bn-endpoint-checker.sh for multi-port deployments
## Story Form

As a **block node operator or developer** running the `bn-endpoint-checker.sh` health-check script against a production or staging deployment,
I want to **specify a separate port for `BlockAccessService/getBlock` calls (`--block-access-port`)** independently of the port used for `BlockNodeService/serverStatus`,
So that **I can correctly target multi-port deployments where the server-status port (e.g. 40982) and the block-access port (e.g. 40981) differ**, without having to maintain separate invocations or manually switch ports.

## Technical Notes

In production deployments (`charts/block-node-server/values-overrides/lfh-values.yaml`) the block node exposes distinct ports per service:

| Service | Port | Helm key |
|---------|------|----------|
| Publisher gRPC | 40984 | `ports.publisher` |
| Subscriber gRPC | 40980 | `ports.subscriber` |
| **Block Access (getBlock)** | **40981** | `ports.blockAccess` |
| Health HTTP | 40983 | `ports.health` |
| Server Status HTTP | 40982 | `ports.serverStatus` |

`bn-endpoint-checker.sh` currently accepts endpoints as `<host:port>` positional arguments and sends **all** gRPC calls — `serverStatus`, `serverStatusDetail`, and `BlockAccessService/getBlock` (via `--latest-block-proof`) — to the **same** port.

This is incorrect for multi-port deployments: `serverStatus`/`serverStatusDetail` live on `BlockNodeService` (port 40982 or the LB port 40840), while `getBlock` lives on `BlockAccessService` (port 40981). Specifying the status port in the endpoint list allows status calls to succeed but silently routes `getBlock` to the wrong port.

### Proposed change

Add a `--block-access-port PORT` flag:

```bash
bn-endpoint-checker.sh --version 0.40.0 --latest-block-proof \
  --block-access-port 40981 \
  node1.mainnet.blocknode.hashgraph-devops.com:40982 \
  node2.mainnet.blocknode.hashgraph-devops.com:40982
```

When `--block-access-port` is set, `grpc_call_block_proof` uses `host:<block-access-port>` as its target while status calls continue using the positional `host:port`. When the flag is omitted, existing behaviour is preserved — all calls use the same endpoint.

Nana-EC:
## Summary

Fixes two issues observed when running `bn-endpoint-checker.sh --latest-block-proof` against production multi-port block node deployments.

---

### Commit 1 — feat: `--block-access-port` for multi-port deployments (closes #3424)

**Problem:** Production deployments expose `BlockNodeService` (serverStatus / serverStatusDetail) and `BlockAccessService` (getBlock) on **separate ports** (`40982` and `40981` in `lfh-values.yaml`). The script previously sent all gRPC calls — including `getBlock` — to the single port in the positional `host:port` argument, routing getBlock to the wrong service.

**Fix:** Add `--block-access-port PORT`. When set, `grpc_call_block_proof` targets `host:<block-access-port>` while status calls continue using the positional endpoint port. Backward-compatible — when omitted, all calls use the same port as before.

```bash
# Before: only one port, getBlock goes to wrong service in multi-port setups
bn-endpoint-checker.sh --version 0.40.0 --latest-block-proof \
  node1.example.com:40982

# After: status on 40982, getBlock on 40981
bn-endpoint-checker.sh --version 0.40.0 --latest-block-proof \
  --block-access-port 40981 \
  node1.example.com:40982
```

---

### Commit 2 — fix: ResourceExhausted on large blocks (closes #3423)

**Problem:** `grpcurl` has a hard 4 MiB (4,194,304 byte) limit on incoming gRPC response messages. Production blocks can exceed 17 MiB, causing `--latest-block-proof` to fail:

```
🟠 latest block proof WARN (getBlock unavailable or no blocks stored)  (1.3 s)
   ERROR:
     Code: ResourceExhausted
     Message: grpc: received message larger than max (17478344 vs. 4194304)
```

**Fix:** Pass `-max-msg-sz 67108864` (64 MiB) to grpcurl in `grpc_call_block_proof`. Add `--max-block-sz BYTES` so operators can tune the ceiling if blocks grow larger.

---

## Test plan

- [x] Run `--latest-block-proof --block-access-port 40981` against a multi-port deployment and confirm getBlock succeeds on port 40981
- [ ] Run without `--block-access-port` and confirm existing single-port behaviour is unchanged
- [x] Run against a node serving large blocks (>4 MiB) and confirm no ResourceExhausted error
- [ ] Run `--help` and confirm both new flags appear in usage
- [x] Confirm run summary prints `Block access port` and `Max block size` lines correctly

### DIFF
--- a/tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
+++ b/tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
@@ -60,14 +60,13 @@
 #                           the node exposes separate ports per service — the
 #                           production default has serverStatus on 40982 and
 #                           getBlock on 40981.
-#       --max-block-sz BYTES
-#                           Maximum gRPC response size (bytes) accepted from
+#       --max-block-sz MIB
+#                           Maximum gRPC response size in MiB accepted from
 #                           BlockAccessService/getBlock. grpcurl's built-in
-#                           default is 4 MiB (4194304), which is too small for
-#                           production blocks that can exceed 16 MiB, causing a
-#                           ResourceExhausted error. Defaults to 10485760
-#                           (10 MiB). Increase if blocks grow beyond that
-#                           threshold.
+#                           default is 4 MiB, which is too small for production
+#                           blocks (17+ MiB observed), causing a
+#                           ResourceExhausted error. Must be a positive integer
+#                           between 1 and 100 (inclusive). Defaults to 20.
 #   -h, --help              Print this help text and exit.
 #
 # ENVIRONMENT
@@ -228,7 +227,7 @@ USE_TLS=false            # When true, grpcurl uses TLS; default is plaintext.
 DETAILED_STATUS=false    # When true, also call serverStatusDetail per endpoint.
 LATEST_BLOCK_PROOF=false # When true, fetch the latest block and report proof type.
 BLOCK_ACCESS_PORT=""     # When set, used as the port for BlockAccessService/getBlock.
-BLOCK_MAX_MSG_SZ=10485760 # Max gRPC response bytes for getBlock (default 10 MiB).
+BLOCK_MAX_BLOCK_MIB=20   # Max getBlock response size in MiB (default 20; range 1–100).
 ENDPOINTS=()             # Positional <host:port> arguments collected here.
 
 while [[ $# -gt 0 ]]; do
@@ -250,7 +249,11 @@ while [[ $# -gt 0 ]]; do
       BLOCK_ACCESS_PORT="$2"; shift 2 ;;
     --max-block-sz)
       [[ -n "${2:-}" ]] || { log_err "--max-block-sz requires a value"; usage 2; }
-      BLOCK_MAX_MSG_SZ="$2"; shift 2 ;;
+      if ! [[ "${2}" =~ ^[0-9]+$ ]] || (( ${2} < 1 || ${2} > 100 )); then
+        log_err "--max-block-sz must be a positive integer between 1 and 100 (MiB)."
+        usage 2
+      fi
+      BLOCK_MAX_BLOCK_MIB="${2}"; shift 2 ;;
     -h|--help)
       usage 0 ;;
     -*)
@@ -540,7 +543,7 @@ grpc_call_block_proof() {
   local -a flags=(
     -import-path "${RESOLVED_PROTO_DIR}"
     -proto        "${BLOCK_ACCESS_PROTO}"
-    -max-msg-sz   "${BLOCK_MAX_MSG_SZ}"
+    -max-msg-sz   "$(( BLOCK_MAX_BLOCK_MIB * 1048576 ))"
     -d            '{"retrieve_latest": true}'
   )
   [[ "$USE_TLS" == "false" ]] && flags=("-plaintext" "${flags[@]}")
@@ -1007,7 +1010,7 @@ main() {
   log_info "  Detailed status   : ${detail_label}"
   log_info "  Block proof       : ${proof_label}"
   log_info "  Block access port : ${block_access_label}"
-  log_info "  Max block size    : ${BLOCK_MAX_MSG_SZ} bytes"
+  log_info "  Max block size    : ${BLOCK_MAX_BLOCK_MIB} MiB"
 
   # Iterate over every supplied endpoint. Failures are accumulated rather than
   # stopping immediately so the operator gets a complete picture of all nodes in
--- a/tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
+++ b/tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
@@ -52,6 +52,14 @@
 #                           Requires the proto package to contain the
 #                           block_access_service.proto file (same archive as the
 #                           node_service.proto already required).
+#       --block-access-port PORT
+#                           Port to use for BlockAccessService/getBlock calls
+#                           when --latest-block-proof is enabled. Defaults to
+#                           the same port as the positional endpoint argument.
+#                           Set this to the blockAccess port (e.g. 40981) when
+#                           the node exposes separate ports per service — the
+#                           production default has serverStatus on 40982 and
+#                           getBlock on 40981.
 #   -h, --help              Print this help text and exit.
 #
 # ENVIRONMENT
@@ -80,6 +88,12 @@
 #   bn-endpoint-checker.sh --version 0.32.0 --latest-block-proof \
 #     node1.mainnet.blocknode.hashgraph-devops.com:40840 \
 #     node2.mainnet.blocknode.hashgraph-devops.com:40840
+#
+#   # Multi-port production deployment: status on 40982, block-access on 40981
+#   bn-endpoint-checker.sh --version 0.40.0 --latest-block-proof \
+#     --block-access-port 40981 \
+#     lfh01.mainnet.blocknode.hashgraph-devops.com:40982 \
+#     lfh02.mainnet.blocknode.hashgraph-devops.com:40982
 #
 # REQUIREMENTS
 #   grpcurl   Auto-installed if missing (you will be prompted for the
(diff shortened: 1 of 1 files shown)


tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
```diff
@@ -52,6 +52,21 @@
...
+#                           the node exposes separate ports per service — the
+#                           production default has serverStatus on 40982 and
+#                           getBlock on 40981.
+#       --max-block-sz MIB
+#                           Maximum gRPC response size in MiB accepted from
+#                           BlockAccessService/getBlock. grpcurl's built-in
+#                           default is 4 MiB, which is too small for production
+#                           blocks (17+ MiB observed), causing a
+#                           ResourceExhausted error. Must be a positive integer
+#                           between 1 and 100 (inclusive). Defaults to 20.
```
jsync-swirlds, line 69: This configuration does not exist in any manner that is specific to the block access service.
The block access service has exactly one configuration item, and it is the port.

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
```diff
@@ -205,6 +226,8 @@ PROTO_DIR_ARG=""         # --proto-dir flag value (overrides PROTO_DIR env var).
 USE_TLS=false            # When true, grpcurl uses TLS; default is plaintext.
 DETAILED_STATUS=false    # When true, also call serverStatusDetail per endpoint.
 LATEST_BLOCK_PROOF=false # When true, fetch the latest block and report proof type.
+BLOCK_ACCESS_PORT=""     # When set, used as the port for BlockAccessService/getBlock.
+BLOCK_MAX_BLOCK_MIB=20   # Max getBlock response size in MiB (default 20; range 1–100).
```
jsync-swirlds, line 230: This is not related to `getBlock`, and there is no code that enables this value.

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
```diff
@@ -955,17 +997,20 @@ main() {
...
-  log_info "  Proto dir       : ${RESOLVED_PROTO_DIR}"
-  log_info "  Transport       : ${tls_label}"
-  log_info "  Detailed status : ${detail_label}"
-  log_info "  Block proof     : ${proof_label}"
+  log_info "  Proto dir         : ${RESOLVED_PROTO_DIR}"
+  log_info "  Transport         : ${tls_label}"
+  log_info "  Detailed status   : ${detail_label}"
+  log_info "  Block proof       : ${proof_label}"
+  log_info "  Block access port : ${block_access_label}"
+  log_info "  Max block size    : ${BLOCK_MAX_BLOCK_MIB} MiB"
```
jsync-swirlds, line 1013: This also does not have any implementation.

Why would we try to limit response size?  We already accepted, verified, and stored the block

jsync-swirlds (commented):
Looks like the AI invented a new configuration for no reason...

Nana-EC:
> Looks like the AI invented a new configuration for no reason...

Item 2 in the PR description highlight an issue where when a large block is queried to check the block proof type the grpcurl observes a ResourceExhausted failure.
The config is part of the flags that grpcurl uses

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
Nana-EC: `BLOCK_MAX_BLOCK_MIB` is added to the grpcurl `flags`.
The implementation to use flags is already present in the `grpc_call()`

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
Nana-EC: It is related and there is code that enabled it. It's for grpcurl no the BN API itself

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
Nana-EC: It's a grpcurl flag added to the `grpc_call_block_proof()`
[max-msg-sz](https://github.com/fullstorydev/grpcurl/blob/master/cmd/grpcurl/grpcurl.go#L143) is the config

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
```diff
@@ -221,6 +244,16 @@ while [[ $# -gt 0 ]]; do
       DETAILED_STATUS=true; shift ;;
     --latest-block-proof)
       LATEST_BLOCK_PROOF=true; shift ;;
+    --block-access-port)
+      [[ -n "${2:-}" ]] || { log_err "--block-access-port requires a value"; usage 2; }
```
berryware, line 248: Nit: Should this field have numeric validation like the others.

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
```diff
@@ -221,6 +244,16 @@ while [[ $# -gt 0 ]]; do
...
+      BLOCK_ACCESS_PORT="$2"; shift 2 ;;
+    --max-block-sz)
+      [[ -n "${2:-}" ]] || { log_err "--max-block-sz requires a value"; usage 2; }
+      if ! [[ "${2}" =~ ^[0-9]+$ ]] || (( ${2} < 1 || ${2} > 100 )); then
```
berryware, line 252: Nit: The regex allows for zero leading numbers which will get picked up as octal.

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
jsync-swirlds: Ahh, did not realize this is grpcurl, not BN.

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
jsync-swirlds: Ahh, did not realize this is grpcurl, not BN.

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
jsync-swirlds: Ahh, did not realize this is grpcurl, not BN.

jsync-swirlds (approved):

rockysingh (approved):

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
Nana-EC: Will review in https://github.com/hiero-ledger/hiero-block-node/issues/3497

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
Nana-EC: Will review in https://github.com/hiero-ledger/hiero-block-node/issues/3497

Issue #3436, opened 2026-08-13
Improve Publisher Manager Accuracy for Determining Which Handler Supplied Which Block


ata-nas:
### Story Form

As a Publisher Manager
I want to be able to accurately determine which block was supplied by which publisher
So that I can take appropriate action when I see a failure

### Technical Notes
- We recently tackled a flaky e2e test in #3391 and #3392
- After investigating, we've found out that the new verification plugin now reports session cancellations as a failure (failed verification notification)
  - This then gets picked up by publisher, but it only sees a current snapshot of which handler is supplying which block
  - So if p1 originally supplied b1, then b1 fails, there is a race between the notification arriving in publisher and being processed, and some other publisher, say p2, starting or is in the middle of supplying b1 again
  - It is possible, as we see with the test being flaky, that we are taking action on a publisher, that was not originally responsible for supplying the actual block that failed
- We want to figure out a new and accurate way of determining the actual handler that supplied the actual block that failed, so we can be accurate in taking an action specifically for that handler
- We also intend to make adjustments to Publisher handling for interrupted blocks.
  - Add a message, similar to end of block, to indicate an interrupted or abandoned block.
  - Adjust Verification to drop abandoned or interrupted blocks; possibly with a low-priority failure notification.
  - Adjust Publisher to tailor action more closely based on types of verification failure notifications.

ata-nas:
@Nana-EC @jsync-swirlds after some discussion, we want to keep the topic of sophisticated and accurate block source tracking because we want to have that option in the future if need arises, there could be other uses for it as well.

Note that this topic is probably one for the future and is on if-we-need-it basis.

Issue #2977, opened 2026-06-08
Utilize `FailureInfo` on Failed Verification


ata-nas:
### Story Form

As a Publisher Plugin
I want to utilize the `FailureType` received on failed verification
So that I can take the most optimal action

### Technical Notes

- Recently `FailureType`s were introduced (extended in #2869)
  - later superseded by a more complex `FailureInfo`
- We want to use these types to optimize
- An example is that informational failures do not require resend
- Update the publisher to take the best and most optimal action based on failure type

ata-nas:
@jsync-swirlds note that publisher must decide if to send a resend based on available information. Maybe we've canceled a session for block 3, which results in a failure, but someone else is already sending block 3 again, or the same publisher just restarted it.

PR #3437, opened 2026-08-13
fix: stream-publisher resend handler killed with BAD_BLOCK_PROOF before ACK can be sent

### LINKED ISSUE #3391: test(e2e-suites): BlockNodeApiRegressionTest RESEND/ACK latches intermittently time out under CI
### Description

`BlockNodeApiRegressionTest > stall-detected RESEND must persist stalled block and acknowledge in order` failed intermittently in CI, timing out waiting for the ACK that confirms block 2 was persisted after being resent to fill a stall-detected gap.

Observed in a run of PR #3295 (unrelated release-automation change — the test itself isn't touched by that PR):
https://github.com/hiero-ledger/hiero-block-node/actions/runs/31106024916/job/92631294648?pr=3295

```
org.opentest4j.AssertionFailedError: Timed out waiting for ACK(>=2) — block 2 persisted after RESEND ==> expected: <0> but was: <1>
    at app/org.hiero.block.node.suites@0.40.0-SNAPSHOT/org.hiero.block.suites.e2e.BlockNodeApiRegressionTest.awaitLatch(BlockNodeApiRegressionTest.java:486)
```

The test drives two publisher connections: publisher 1 stalls after a header-only block 2, publisher 2 pushes blocks 4–6 to trigger stall detection, receives a `RESEND_BLOCK(2)`, then sends the full block 2 to fill the gap. The failure is at the `awaitLatch` call waiting for the resulting `ACK(>=2)` — the latch never reached 0 within the fixed 60s timeout (`DEFAULT_AWAIT_TIMEOUT`, `BlockNodeApiRegressionTest.java:61`) even though the test ran for the full ~1m2s, i.e. the ACK for block 2 either never arrived or arrived after the deadline.

The surrounding log also shows some noisy secondary errors during test-teardown shutdown (`AccessDeniedException: /opt/hiero`, an `UploadException`/`S3ClientInitializationException` from the cloud-storage-expanded plugin about a blank input string, and a `SocketWriterException`/`RejectedExecutionException` chain) — unclear yet whether any of these are the actual root cause of the missed ACK or just downstream noise from the JVM shutting down after the test already failed. Worth checking whether the cloud-storage-expanded upload error is spurious test-fixture config (blank access/secret key in the E2E test environment) versus something that could be blocking the persistence path that the ACK depends on.

### Steps to Reproduce

Not yet reproduced locally — only seen once in CI so far. Re-running `BlockNodeApiRegressionTest.stallDetectedResendMustPersistStalledBlockAndAcknowledgeInOrder` under CI-like load (or with an artificially shortened `DEFAULT_AWAIT_TIMEOUT`) may help surface it more reliably.

### Additional context

`tools-and-tests/suites/src/main/java/org/hiero/block/suites/e2e/BlockNodeApiRegressionTest.java` — test method `stallDetectedResendMustPersistStalledBlockAndAcknowledgeInOrder` (line ~248), `awaitLatch` helper (line ~426).

### LINKED ISSUE #3392: stream-publisher resend handler killed with BAD_BLOCK_PROOF before ACK can be sent
### Description

Root-cause dig into the timeouts tracked in #3391. Traced the actual server-side event sequence from the raw CI logs of both failures (same PR run, two different CI attempts, two different tests in `BlockNodeApiRegressionTest`) and found a concrete, confirmed mechanism: **the handler that's asked to resend a block gets its stream torn down with `BAD_BLOCK_PROOF` moments after being asked to resend — before it ever gets a chance to have its resent block acknowledged.** That's why the client-side test times out waiting for the ACK: the server-side stream backing that client connection is already dead.

**Confirmed sequence (from `stall-detected RESEND must persist stalled block and acknowledge in order`, run 31106024916 job 92631294648):**
```
13:32:08.266  Handler 0 ending with code TIMEOUT                     (publisher 1's stalled connection times out)
13:32:08.266  Handler 1 is sending RESEND(2)                          (publisher 2 asked to resend block 2)
13:32:08.269  Handler 1 replies.onNext ... to send RESEND_BLOCK
13:32:08.277  Handler 0 issued onComplete/closeConnection
13:32:08.311  Handler 1 handling failed verification for block 2
13:32:08.311  Handler 1 ending with code BAD_BLOCK_PROOF              <-- publisher 2's connection killed
13:32:08.343  Wrote verified block 2 to file                          <-- block 2 verifies fine moments later
  ...(dead silence for ~60s, client-side timeout fires)...
```

**Confirmed sequence (from `publisher RESEND fills block gap left by severed connection`, same PR, retry attempt, job 92647232881):**
```
14:27:41.494  Handler 0 received EndStream with RESET
14:27:41.495  Handler 0 issued onComplete/closeConnection
14:27:41.541  Handler 1 is sending RESEND(3)
14:27:41.541  Handler 1 replies.onNext ... to send RESEND_BLOCK
14:27:41.545  Handler 1 handling failed verification for block 3
14:27:41.545  Handler 1 ending with code BAD_BLOCK_PROOF              <-- same pattern, different block/test
14:27:41.595  Wrote verified block 3 to file                          <-- again verifies fine shortly after
```

Same shape both times: the handler that was just asked to RESEND a block gets a `handleFailedVerification` call for that exact block number within tens of milliseconds, is killed with `BAD_BLOCK_PROOF`, and the block then verifies and persists successfully anyway — with nobody left to send the ACK to.

### Where this happens in code

- `PublisherHandler.handleFailedVerification(long blockNumber)` (`block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/PublisherHandler.java:302-317`): if `blockNumber` is in *this* handler's own `unacknowledgedStreamedBlocks` set, it removes it and immediately calls `endStreamWithCode(Code.BAD_BLOCK_PROOF, ...)` — no check on whether the failure actually originated from data *this* handler sent.
- `LiveStreamPublisherManager.handleVerification(VerificationNotification notification)` (`block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java:408-431`): on a failed verification, schedules a resend (`blocksToResend.add(blockNumber)`) and then **broadcasts to every active handler** (`for (final PublisherHandler handler : handlers.values())`), calling `handler.handleFailedVerification(blockNumber)` on each and stopping at the first one that returns `true`.
- `unacknowledgedStreamedBlocks` is populated in `PublisherHandler.handleEndOfBlock` (`PublisherHandler.java:539`), keyed only by block number, with no correlation back to *which specific submission* (i.e. which handler's actual bytes) the eventual verification result is for.

### Hypothesis (needs confirmation — this is where I'd point someone continuing the dig)

Verification appears to run asynchronously relative to the RESEND handoff. The most likely explanation: the failed-verification notification being delivered to Handler 1 is *stale* — it belongs to the original, incomplete/bad submission from the handler that just disconnected (Handler 0), but by the time it's delivered, ownership bookkeeping for that block number has already moved to Handler 1 (the one now resending it), and `handleVerification`'s all-handlers broadcast has no way to distinguish "this failure belongs to the submission that already left" from "this failure belongs to the handler currently registered for this block number." An alternative, not yet ruled out: it's a genuine (if transient) verification defect specifically on the first bytes of a post-RESEND resubmission, unrelated to the disconnected handler. Distinguishing these needs either a local reproduction with added tracing on notification identity (not just block number), or a targeted unit test around `LiveStreamPublisherManager.handleVerification` + concurrent handler resend.

### Steps to Reproduce

Not yet reproduced locally — seen twice in CI so far (see #3391), both times in this exact resend-then-verify-then-ACK window. A local repro would need to trigger stall-detection/severed-connection + RESEND under enough concurrency/timing pressure to hit the same race; the existing `BlockNodeApiRegressionTest` tests already exercise the right scenario, so tight-looping them locally (or under artificial delay/thread contention) may reproduce it.

### Additional context

Related: #3391 (tracks the test-visible symptom — timeouts in `BlockNodeApiRegressionTest`). This issue tracks the underlying server-side cause.

ata-nas:
* Ignore failed verification notifications of type CANCELLED in LiveStreamPublisherManager.handleVerification, only log them at INFO
  * A cancelled verification session cannot be reliably attributed to the publisher that supplied the block, so scheduling a resend and sending EndOfStream(BAD_BLOCK_PROOF) could kill the wrong publisher and cause the intermittent RESEND/ACK latch timeouts seen in CI
  * Accurate attribution is follow-up work captured by #3436, full FailureInfo handling by #2977 
* Add tests pinning the new behavior in LiveStreamPublisherManagerTest
  * CANCELLED failures produce no responses on any handler and update no metrics
  * CANCELLED failures do not schedule a resend, a later persisted notification for the same block is acknowledged normally
  * All failure types other than CANCELLED still take the BAD_BLOCK_PROOF path (parameterized over the FailureType enum)

Resolves #3391
Resolves #3392

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -64,6 +64,7 @@ import org.hiero.block.node.spi.blockmessaging.PersistedNotification;
 import org.hiero.block.node.spi.blockmessaging.PublisherStatusUpdateNotification;
 import org.hiero.block.node.spi.blockmessaging.PublisherStatusUpdateNotification.UpdateType;
 import org.hiero.block.node.spi.blockmessaging.VerificationNotification;
+import org.hiero.block.node.spi.blockmessaging.VerificationNotification.FailureType;
 import org.hiero.block.node.spi.threading.ThreadPoolManager;
 import org.hiero.metrics.LongCounter;
 import org.hiero.metrics.LongGauge;
@@ -416,15 +417,32 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
                 && blockNumber > lastPersistedBlockNumber.get()
                 && blockNumber < nextUnstreamedBlockNumber.get();
         if (shouldHandle) {
-            // Schedule a resend for the block before sending the bad block proof message
-            blocksToResend.add(blockNumber);
-            // Iterate over all handlers and attempt to send the
-            // bad block proof message.
-            for (final PublisherHandler handler : handlers.values()) {
-                if (handler.handleFailedVerification(blockNumber)) {
-                    // There will always be only one handler that will send the
-                    // bad block proof message. Once we have, we can break.
-                    break;
+            // A note for a flaky test fix for #3391 & #3392:
+            // After investigating, we have found out that the new verification
+            // plugin's report of canceled verifications is being picked up
+            // here, but there is no guarantee that we will accurately point to
+            // the actual publisher that has supplied the failed (canceled) block.
+            // This is true because we only see a snapshot of the current activity.
+            // If publisher 1 supplied block 5 that was later canceled, publisher 2
+            // might have started resending block 5, and the current state
+            // of the publisher manager will point a finger to publisher 2 for
+            // the failure of block 5.
+            // There is more work to be done so that we will be more accurate,
+            // this is captured by @todo(3436).
+            // Completely handling the FailureInfo is captured by @todo(2977).
+            if (notification.failureInfo().failureType() == FailureType.CANCELLED) {
+                LOGGER.log(INFO, "Session for block {0} was cancelled", blockNumber);
+            } else {
+                // Schedule a resend for the block before sending the bad block proof message
+                blocksToResend.add(blockNumber);
+                // Iterate over all handlers and attempt to send the
+                // bad block proof message.
+                for (final PublisherHandler handler : handlers.values()) {
+                    if (handler.handleFailedVerification(blockNumber)) {
+                        // There will always be only one handler that will send the
+                        // bad block proof message. Once we have, we can break.
+                        break;
+                    }
                 }
             }
         }
--- a/block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
+++ b/block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
@@ -1289,6 +1289,186 @@ class LiveStreamPublisherManagerTest {
                         .hasSize(1)
                         .containsExactly(block[block.length - 1]);
             }
+
+            // @todo(2977) this test might not be entirely true once we handle failure info
+            /// This test aims to assert that the
+            /// [LiveStreamPublisherManager#handleVerification(VerificationNotification)]
+            /// ignores a failed verification of type [FailureType#CANCELLED].
+            /// A cancelled verification session cannot be reliably attributed
+            /// to the publisher that supplied the block, so the manager must
+            /// only log the event. No responses of any kind are expected to be
+            /// sent, even by the handler that supplied the block, and no
+            /// metrics are expected to be updated.
+            @Test
+            @DisplayName("handleVerification() ignores CANCELLED failures, no responses of any kind sent")
+            void testHandleVerificationCancelledIgnored() {
+                // We need to send a request via the publisher handler first,
+                // This will properly update the internal state of the manager
+                // so we can assert correctly. We aim to increment the next
+                // unstreamed block number to 1L so we have a gap between
+                // latest persisted (which should be -1L) and next unstreamed.
+                // This is an expected condition during normal operation.
+                final TestBlock block = TestBlockBuilder.generateBlockWithNumber(0);
+                // Now we build the request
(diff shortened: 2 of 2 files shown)


ivannov (approved):

jsync-swirlds (approved):

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -416,15 +417,32 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+            // might have started resending block 5, and the current state
+            // of the publisher manager will point a finger to publisher 2 for
+            // the failure of block 5.
+            // There is more work to be done so that we will be more accurate,
+            // this is captured by @todo(3436).
+            // Completely handling the FailureInfo is captured by @todo(2977).
+            if (notification.failureInfo().failureType() == FailureType.CANCELLED) {
+                LOGGER.log(INFO, "Session for block {0} was cancelled", blockNumber);
+            } else {
+                // Schedule a resend for the block before sending the bad block proof message
```
Nana-EC, line 436: Curious if the order has impacts on external publisher handling?
Does the protocol describe this case and if a publisher should expect a resend then a bad_block_proof for the same block?

Nana-EC (approved):
LG
Curious on order and if we have expectations covered

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: Not sure I understand your question, perhaps we can discuss separately?

PR #3441, opened 2026-08-13
fix: Change RsaRosterBootstrapPlugin to use two HttpCLients when connecting to MirrorNode

### LINKED ISSUE #3440: RsaRosterBootstrapPlugin - HttpClient keepalive cannot be used across MirrorNode's java and js rest APIs
### Story Form

As a BlockNode engineer
I want to access the MirrorNode as efficiently as possible
So that the RsaRosterBootstrapPlugin works as performantly as possible.

### Technical Notes

On MirrorNode
/api/v1/blocks is a javascript rest API
/api/v1/network/nodes is a java rest api.

Both endpoints/services are accessed via the same HAProxy ingress. Once an HttpClient is pinned to the endpoint, it cannot be used to access the other endpoint.

The BlockNode creates one HttpClient with keepalive=true to access both enpoints. Once the HttpClient connects to one endpoint, it can only see that endpoint and throws a 404 on the second. 

Rather than setting keepalive=false, the recommended solution is to create 2 HttpClients with keepalive=true. One client for the js endpoint, and one client for the java enpoint.

berryware:
## Reviewer Notes
- Added two separate HttpClients, with keepAlive=true for /network/nodes (java api) and /blocks (js api).
- Added a test helper to reproduce the split api setup of MirrorNode.
- verified the test fails without the fix and passes with it.

## Related Issue(s)
Closes #3440


jsync-swirlds (approved):

Nana-EC (approved):
Nice catch

PR #3442, opened 2026-08-14
chore: Follow up on improvements raised for PR 3245

### LINKED ISSUE #3264: Follow up on improvements raised for PR 3245
### Story Form

N/A

### Technical Notes

Three tasks:
- Resolve possibly incorrect behavior in createMultipartUpload and completeMultipartUpload as noted [here](https://github.com/hiero-ledger/hiero-block-node/pull/3245#discussion_r3628688792).
- Add handling for CancellationException where potentially needed in CloudStorageArchivePlugin (as noted [here](https://github.com/hiero-ledger/hiero-block-node/pull/3245#discussion_r3623366486)).
- Thoroughly vet thread handling and thread safety for Upload methods in CloudStorageArchivePlugin as noted [here](https://github.com/hiero-ledger/hiero-block-node/pull/3245#discussion_r3623378739).
   - Also consider a consolidated helper method to simplify the multiple code sections that handle Futures.

ivannov:
## Reviewer Notes

Addressed three leftover comments from PR #3245 

* One more retry in `completeMultipartUpload` as it is cheaper than the recovery task and also it avoids re-retrieving the boundary component
* All 4 `Future.get()` call sites were already guard with isCancelled(), so comment 2 brought no code changes
* Added `FutureOutcome<T>` record and `awaitOutcome(Future<T>)` helper

## Related Issue(s)

Closes #3264

### DIFF
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
@@ -205,11 +205,18 @@ public class BlockUploadTask implements Callable<UploadResult> {
     private void completeMultipartUploadWithRetry(S3Client s3, String uploadId, List<String> etags)
             throws S3ResponseException, IOException {
         try {
-            s3.completeMultipartUpload(key, uploadId, etags);
+            doCompleteMultipartUpload(s3, uploadId, etags);
         } catch (S3ResponseException | IOException e) {
             LOGGER.log(DEBUG, "Failed to complete multipart upload for key {0}; retrying once", key, e);
-            s3.completeMultipartUpload(key, uploadId, etags);
+            doCompleteMultipartUpload(s3, uploadId, etags);
         }
+    }
+
+    /// Delegates to [S3Client#completeMultipartUpload].  Overridable so tests can inject completion
+    /// failures without a live S3 endpoint.
+    void doCompleteMultipartUpload(S3Client s3, String uploadId, List<String> etags)
+            throws S3ResponseException, IOException {
+        s3.completeMultipartUpload(key, uploadId, etags);
     }
 
     /// Takes one block from [blockQueue], encodes it as a tar entry, and appends it to `buffer`.
--- a/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
+++ b/block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.cloud.storage.archive;
 
+import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.INFO;
 import static java.lang.System.Logger.Level.TRACE;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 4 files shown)


block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -279,28 +279,52 @@ public void handleVerification(VerificationNotification notification) {
...
+    private static <T> FutureOutcome<T> awaitOutcome(Future<T> future) throws InterruptedException {
+        if (future.isCancelled()) {
+            return new FutureOutcome<>(null, null);
+        }
+        try {
+            return new FutureOutcome<>(future.get(), null);
+        } catch (ExecutionException e) {
+            return new FutureOutcome<>(null, e);
+        }
+    }
```
jsync-swirlds: Why not catch InterruptedException within the method and place that in the return value also?
It should never happen (because you check `isDone` before calling this method); but it's a lot cleaner to return the exception as a failure than to try to catch it in every caller.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
jsync-swirlds: Another place we could potentially avoid declaring the exception if we caught it in the awaitOutcome method.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -279,28 +279,52 @@ public void handleVerification(VerificationNotification notification) {
     private boolean checkCompletedUpload() throws ExecutionException, InterruptedException {
```
jsync-swirlds: Should be able to remove the throw for both exceptions here...

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
```diff
@@ -279,28 +279,52 @@ public void handleVerification(VerificationNotification notification) {
...
         boolean cancelled = false;
         if (currentUploadFuture != null && currentUploadFuture.isDone()) {
-            if (currentUploadFuture.isCancelled()) {
+            final FutureOutcome<UploadResult> outcome = awaitOutcome(currentUploadFuture);
+            if (outcome.isCancelled()) {
                 LOGGER.log(TRACE, "Block upload task was cancelled");
                 cancelled = true;
+            } else if (outcome.failure() != null) {
+                throw outcome.failure();
```
jsync-swirlds: Why are we throwing this instead of handling it?
The whole point of returning the exception is to avoid the high cost of a throw.

jsync-swirlds (commented):
We can better meet the spirit of the original comments by reducing unnecessary exception throws.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/BlockUploadTask.java
```diff
@@ -193,13 +194,25 @@ public UploadResult call()
...
+    /// [StartupRecoveryTask], which discards and re-derives the last block on this boundary.
+    private void completeMultipartUploadWithRetry(S3Client s3, String uploadId, List<String> etags)
+            throws S3ResponseException, IOException {
+        try {
+            s3.completeMultipartUpload(key, uploadId, etags);
+        } catch (S3ResponseException | IOException e) {
+            LOGGER.log(DEBUG, "Failed to complete multipart upload for key {0}; retrying once", key, e);
+            s3.completeMultipartUpload(key, uploadId, etags);
+        }
+    }
```
jsync-swirlds, line 213: We still need recovery as a backstop if the second attempt fails.
I assume that's already in place, however.

ivannov: Yes, we have. As stated in the PR message, this retry was introduced because it is cheaper than the recovery

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Done. `awaitOutcome` now catches `InterruptedException`, restores the interrupt flag, and returns it as a failure like any other

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Removed. No throws remains on this one and the other three methods that call `awaitOutcome`

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Agreed. The exception branch is handled in place and merged with the `UploadResult.FAILED` branch since both do the same thing. `handleVerification`'s whole try/catch is gone now.

block-node/cloud-storage-archive/src/main/java/org/hiero/block/node/cloud/storage/archive/CloudStorageArchivePlugin.java
ivannov: Removed

jsync-swirlds (approved):

ata-nas (approved):

PR #3443, opened 2026-08-14
fix: BackfillMultiportSourceTests.publishBlocks assumes 1 ACK per block, but acks are watermark-coalesced

### LINKED ISSUE #3401: test(e2e-suites): BackfillMultiportSourceTests.publishBlocks assumes 1 ACK per block, but acks are watermark-coalesced
### Description

`BackfillMultiportSourceTests > Backfill client fetches from a source with subscribe and serverStatus on different ports` failed in CI on an unrelated PR (#3396, a release-automation YAML-only change):

https://github.com/hiero-ledger/hiero-block-node/actions/runs/31136399137/job/92736618552

```
org.opentest4j.AssertionFailedError: Timed out waiting for publish acknowledgements ==> expected: <0> but was: <1>
    at app/org.hiero.block.node.suites@0.40.0-SNAPSHOT/org.hiero.block.suites.backfill.BackfillMultiportSourceTests.publishBlocks(BackfillMultiportSourceTests.java:180)
```

This is a **different root cause** from #3391/#3392 (the `BlockNodeApiRegressionTest` RESEND/ACK timeouts) — no stall detection, no `BAD_BLOCK_PROOF`, no severed connections involved. This test simply publishes 3 fresh blocks (0, 1, 2) on a brand-new `BlockNodeApp` and waits for 3 `onNext()` calls on the response stream.

Tracing the actual server log for the failing run: all 3 blocks verified and persisted correctly —

```
21:37:16.187  Wrote verified block 1
21:37:16.191  Handler 0 replies.onNext ... to send ACKNOWLEDGEMENT   <- ack #1
21:37:16.198  Wrote verified block 0
21:37:16.218  Wrote verified block 2
21:37:16.219  Handler 0 replies.onNext ... to send ACKNOWLEDGEMENT   <- ack #2 (only 2 total)
```

— but only **2** `ACKNOWLEDGEMENT` messages were ever sent for 3 persisted blocks. This isn't a server defect: `LiveStreamPublisherManager.handlePersisted` (`block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java:456-486`) is an intentional, documented watermark/cumulative design —

```java
handlers.values().parallelStream().unordered().forEach(handler -> {
    // _Important_, we only need the last persisted block number
    // all previous blocks are implicitly acknowledged.
    handler.sendAcknowledgement(lastPersistedBlockNumber.get());
});
```

If two blocks finish persisting close enough together that the second's `handlePersisted` notification arrives before the first's acknowledgement send completes, only **one** ack gets sent — for the higher block number, implicitly covering everything below it. Under CI load (or just occasionally, since this is a real race and not strictly load-dependent), 3 rapidly-persisted blocks can legitimately collapse into 2 or even 1 acknowledgement message.

`BackfillMultiportSourceTests.publishBlocks` (`tools-and-tests/suites/src/main/java/org/hiero/block/suites/backfill/BackfillMultiportSourceTests.java:160-183`) assumes 1:1 block-to-ack cardinality:

```java
final AtomicReference<CountDownLatch> ackLatch = observer.setAndGetOnNextLatch(LAST_BLOCK_NUMBER + 1);
...
ackLatch.get().await(AWAIT_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
assertEquals(0, ackLatch.get().getCount(), "Timed out waiting for publish acknowledgements");
```

That assumption is incompatible with the server's intentional coalescing — this test can flake any time persistence for 2+ blocks happens to overlap, independent of CI load.

### Fix

Wait for a single acknowledgement whose block number is `>= LAST_BLOCK_NUMBER`, not for a fixed count of `onNext()` calls — the same pattern already used correctly elsewhere in the suite, e.g. `BlockNodeApiRegressionTest`'s `setAndGetOnMatchLatch(response -> ... blockNumber() >= N)`.

### Other call sites worth auditing for the same risk

`ResponsePipelineUtils.setAndGetOnNextLatch(1)` (single expected response) is unaffected — the risk is specifically multi-count usages expecting one ack per block:
- `BlockNodeAPITests.java:481` — `setAndGetOnNextLatch(totalBlocks)`
- `BlockNodeAPITests.java:661` — `setAndGetOnNextLatch(totalBlocks)`
- `BlockNodeAPITests.java:312` / `:905` — `setAndGetOnNextLatch(2)`
- `BlockNodeE2EClient.java:131` — `setAndGetOnNextLatch(blockFiles.size())`

Haven't traced each of these individually to confirm they're all exposed the same way (some may count a different response type that doesn't coalesce) — flagging for someone closing this out to check, not asserting they're all broken.

### Additional context

Not reproduced locally yet — seen once in CI. Given the root cause is a timing race in ack coalescing (not resource exhaustion), it should be reproducible by adding a small artificial delay in the test's persistence path or by tight-looping the test under load.

ata-nas:
* Wait for a single acknowledgement covering the last published block instead of counting one response per block in BackfillMultiportSourceTests.publishBlocks, since the server coalesces acks to the highest persisted block
* Apply the same watermark wait in BlockNodeAPITests duplicateBlockSkipWindowAppliesWithinWindowAndEndsOutsideWindow and e2eSocketClosureTest
* Drop response count assertions that assumed one ack per block, keeping the deterministic last-response checks
* Assert the persisted block range via serverStatus after publishing in the duplicate skip window test so persistence is verified independently of acks

Resolves #3401

### DIFF
--- a/tools-and-tests/suites/src/main/java/org/hiero/block/suites/backfill/BackfillMultiportSourceTests.java
+++ b/tools-and-tests/suites/src/main/java/org/hiero/block/suites/backfill/BackfillMultiportSourceTests.java
@@ -156,14 +156,18 @@ public class BackfillMultiportSourceTests {
     }
 
     /// Publishes blocks 0..{@link #LAST_BLOCK_NUMBER}, chained by block hash, on the given port and
-    /// awaits an acknowledgement for each.
+    /// awaits an acknowledgement covering the last block. Acknowledgements are watermark-coalesced by
+    /// the server (an ack for block N implicitly acks all blocks below N), so fewer acks than blocks
+    /// may arrive; waiting for a fixed per-block count would flake (#3401).
     private void publishBlocks(final int port) throws InterruptedException {
         final BlockStreamPublishServiceClient publishClient =
                 new BlockStreamPublishServiceClient(createGrpcClientForPort(port), OPTIONS);
         final ResponsePipelineUtils<PublishStreamResponse> observer = new ResponsePipelineUtils<>();
         final Pipeline<? super PublishStreamRequest> stream = publishClient.publishBlockStream(observer);
 
-        final AtomicReference<CountDownLatch> ackLatch = observer.setAndGetOnNextLatch(LAST_BLOCK_NUMBER + 1);
+        final AtomicReference<CountDownLatch> ackLatch = observer.setAndGetOnMatchLatch(
+                response -> response.response().kind() == PublishStreamResponse.ResponseOneOfType.ACKNOWLEDGEMENT
+                        && response.acknowledgement().blockNumber() >= LAST_BLOCK_NUMBER);
         Bytes previousBlockHash = null;
         for (long blockNumber = 0; blockNumber <= LAST_BLOCK_NUMBER; blockNumber++) {
             final BlockItem[] items = BlockItemBuilderUtils.createSimpleBlockWithNumber(blockNumber, previousBlockHash);
@@ -177,7 +181,10 @@ public class BackfillMultiportSourceTests {
         }
 
         ackLatch.get().await(AWAIT_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
-        assertEquals(0, ackLatch.get().getCount(), "Timed out waiting for publish acknowledgements");
+        assertEquals(
+                0,
+                ackLatch.get().getCount(),
+                "Timed out waiting for a publish acknowledgement of block >= " + LAST_BLOCK_NUMBER);
         stream.closeConnection();
         publishClient.close();
     }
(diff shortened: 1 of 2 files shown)


ivannov:
Looks good. Before approving, one question. The task says that `BlockNodeE2EClient.java:131` is also affected. But this change does not touch it

I assume that this is because the client does not take part in the CI?

ata-nas:
> Looks good. Before approving, one question. The task says that `BlockNodeE2EClient.java:131` is also affected. But this change does not touch it
> 
> I assume that this is because the client does not take part in the CI?

Thanks for your comment. Yes, after investigating, the e2e client can also be updated in the same fashion.
After speaking with @Nana-EC who initially introduced that client, it is actually used, but it may be short-lived and not worth making changes to now. We have acknowledged this, but we will not make any changes to it now. In any case, if we ever see issues with it, we know what changes need to be made and we can quickly respond.

This PR will only deal with the e2e tests.

ivannov (approved):

Nana-EC (approved):

PR #3444, opened 2026-08-14
fix(tools): three CLI fixes so LiveSequential --push-enabled works against a real BN

### LINKED ISSUE #3374: LiveSequential --push-enabled never lands blocks on the BN; queryLastAvailableBlock fails and BN reports every push "ending mid-block"
## Description

Running `days live-sequential --push-enabled` against a freshly-seeded Block Node results in zero blocks ever being persisted. `serverStatus.lastAvailableBlock` stays pinned at the seed's tip forever. The CLI's push worker enters a rejection loop, the BN reports every in-flight block as "ending mid-block", and no block-persist success is logged on either side.

Discovered during previewnet Tier 1 live-push validation on `34.70.11.223` (follow-up work after PR #3333 seed validation).

### CLI side (`live-sequential.log`)

Initial startup query fails:
```
[push] queryLastAvailableBlock failed: com.hedera.pbj.runtime.grpc.GrpcException:
```
`LiveBlockPushClient.queryLastAvailableBlock()` is meant to fetch the BN's current tip so `pushSkipUpToInclusive` gets set to a real value. When it fails, `pushSkipUpToInclusive` stays at the initial `-1L`, so the push worker attempts to push every block wrap produces (including blocks the BN already has).

Then the rejection loop:
```
[push] BN sent EndOfStream (code=DUPLICATE_BLOCK, last persisted block=489866); resuming from block 489867
[push] stream failure: java.lang.RuntimeException: stream closed by remote; reconnecting after 250ms (0 in-flight)
[push] BN sent EndOfStream (code=DUPLICATE_BLOCK, last persisted block=489866); resuming from block 489867
[push] stream failure: java.lang.RuntimeException: stream closed by remote; reconnecting after 250ms (0 in-flight)
```
Repeats indefinitely, interspersed with SkipBlock rejections:
```
[push] BN sent SkipBlock: block 489863 is being sent by another publisher; skipping resend
[push] BN sent SkipBlock: block 489864 is being sent by another publisher; skipping resend
[push] BN sent SkipBlock: block 489865 is being sent by another publisher; skipping resend
[push] BN sent SkipBlock: block 489866 is being sent by another publisher; skipping resend
```
Note: `ps aux | grep tools-.*live-sequential` confirms only ONE CLI process is running. There is no other publisher.

### BN side (`kubectl logs`)

`LiveStreamPublisherManager` reports every in-flight block as "ending mid-block", one line per block, in fast bursts:
```
LiveStreamPublisherManager blockIsEnding - Handler 800 is ending mid-block 504,319
LiveStreamPublisherManager blockIsEnding - Block 504,319 will be resent due to handler 800 ending mid block
LiveStreamPublisherManager blockIsEnding - Handler 800 is ending mid-block 504,320
LiveStreamPublisherManager blockIsEnding - Block 504,320 will be resent due to handler 800 ending mid block
... (continues for ~20 consecutive blocks in <300ms)
```
Zero `persist` / `committed` / `accepted` log entries anywhere across many minutes of activity. `serverStatus.lastAvailableBlock` remains pinned at the seed's value indefinitely.

## Steps to Reproduce

1. Deploy a Block Node with seeded historic archive (e.g., per PR #3333's `wrb-seed-runbook.md`)
2. Confirm `serverStatus` returns a real `lastAvailableBlock` (e.g., `489866` for a freshly-seeded post-reset previewnet)
3. Run `LiveSequential` with live-push against that BN:
   ```bash
   java -jar tools-0.40.0-SNAPSHOT-all.jar --network previewnet days live-sequential \
     -l metadata/listingsByDay \
     -o compressedDays \
     --wrap-output-dir wrappedBlocks \
     --address-book wrappedBlocks/addressBookHistory.json \
     --push-enabled \
     --push-bn-host <cluster-ip-of-bn-service> \
     --push-bn-port 40984
   ```
4. Observe:
   - `queryLastAvailableBlock failed: GrpcException` in the CLI log
   - Push worker enters `EndOfStream(DUPLICATE_BLOCK)` + `SkipBlock(another publisher)` loop
   - BN's `serverStatus.lastAvailableBlock` never advances
   - BN logs show `Handler N ending mid-block X` for every block in bursts

Killing the CLI process, bouncing the BN pod (fresh handler state), and restarting the CLI reproduces the exact same behavior.

## Suspected root causes

Not confirmed which side is at fault; probably a mix. Candidates:

1. **`LiveBlockPushClient.queryLastAvailableBlock` port assumption** — the query uses `--push-bn-host:--push-bn-port` (in the repro above `<cluster-ip>:40984`), but `serverStatus` is served on the split-port setup's server-status port (`40982`), not the publisher port (`40984`). If that's the port mismatch, the query is guaranteed to fail against any BN using per-API split ports.

2. **`LiveBlockPushClient` fallback when `queryLastAvailableBlock` fails** — should either abort startup with a clear error, or fall back to `pushSkipUpToInclusive = 0` (push everything and let the BN's dedup catch dupes). Current behavior of falling to `-1L` combined with aggressive push cascades into the loop below.

3. **`LiveStreamPublisherManager` handler-state management** (`block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java`) — the CLI's rapid reconnect on EndOfStream creates overlapping handler state on the BN side, and the BN considers each reconnected stream a "competing publisher" for blocks that overlap with in-flight ones. That drives the SkipBlock rejections. On top of that, the BN reports "handler ending mid-block" for blocks whose BLOCK_PROOF never lands (stream drops before the block completes). None of this reflects a genuinely competing publisher — it's the CLI's own reconnect stream state being mis-attributed.

4. **CLI batch/stream lifecycle** — `LiveBlockPushClient.sendBlock` splits a block into batches and calls `session.requestPipeline.onNext(req)` per batch. The block boundary is implicit (BLOCK_PROOF item). If the stream drops (from EndOfStream) between batches of a block, the BN sees a partial block and flags it. Combined with the reconnect loop, no block ever completes.

## Additional context

- **Environment**: previewnet Tier 1 BN on `34.70.11.223` running chart 0.37.1
- **CLI version**: `tools-0.40.0-SNAPSHOT-all.jar` (built from `main` with the `--push-*` flags from commit `c173876b3` / PR #3124)
- **Seeded state**: 489,866 blocks from a fresh post-reset previewnet WRB archive (produced via `blocks wrap`, seeded via `backfill-wrb-to-bn.sh`)
- **BN log excerpt** with the handler-ending burst is available on the box; happy to attach if it helps

## Related

- PR #3333 (previewnet Tier 1 deployment config) — the seed we're pushing on top of
- #3364 (StorageClass gap), #3365 (subPath override), #3366 (plugin crash-loop on corrupt zip) — related deployment/operational bugs found during the same validation run

rockysingh:
Fixes #3374.

## Summary

Three CLI-side fixes that together make `LiveSequential --push-enabled` actually push blocks in steady state against a real Block Node (single-port or split-port). Each was independently observed on previewnet during validation.

## Fix 1 `LiveBlockPushClient`: fail-fast on serverStatus query failure + `--push-bn-status-port`

**Root cause:** `queryLastAvailableBlock()` returned `-1` on any failure (unreachable BN, port mismatch, gRPC error). The caller stored that sentinel into `pushSkipUpToInclusive`. The producer's `blockNum > pushSkipUpToInclusive` guard then treated `-1` as "push everything from block 0" — flooding a seeded BN with duplicates, triggering the `DUPLICATE_BLOCK` → reconnect → `SkipBlock("another publisher")` loop from the issue.

**Changes:**
- `queryLastAvailableBlock()` throws a new `QueryFailedException` (with the exact `host:statusPort` and a split-port hint) instead of returning `-1`.
- `LiveSequential` aborts startup on `QueryFailedException` with a clear message, and rejects a negative watermark as defense-in-depth.
- New `--push-bn-status-port` flag — the BN's `BlockNodeService` (which owns `serverStatus`) can live on a different port than `BlockStreamPublishService` in split-port deployments (previewnet: publish=40984, status=40982). Defaults to `--push-bn-port`. `LiveBlockPushClient` constructor now takes an optional `statusPort`; two `WebClient`s are cached, one per URI. Existing 4-arg constructor delegates with `statusPort = port`.

## Fix 2 `BlockTimeReader`: always decode `block_times.bin` relative to mainnet genesis

**Root cause:** `block_times.bin` always encodes values as nanoseconds since mainnet genesis via `instantToBlockTimeLong`, regardless of network. The non-mainnet decode path (`genesisInstant.plusNanos`) was adding a second network-genesis offset, shifting all block times forward by the gap between mainnet and the target network genesis (~4 years for testnet, ~6.6 years for previewnet). This is what caused the tool to think block 0 lived in the far future and refuse to process any day tars.

**Changes:** always use `blockTimeLongToInstant` / `instantToBlockTimeLong` for encode and decode. Removes the now-dead `useMainnetHelpers` flag.

## Fix 3 `LiveBlockPushClient`: emit mandatory `end_of_block` after each block's items

**Root cause:** the publish protocol requires a distinct `PublishStreamRequest.end_of_block` (`BlockEnd{blockNumber=N}`) message to commit block N. `sendBlock` was only sending `blockItems` batches. Without the explicit `end_of_block`, the BN's `PublisherHandler` stays "mid-block" for N; when the next block's `BlockHeader` arrives, `handleBlockItemsRequest:427` fires `publisherManager.blockIsEnding(N, handlerId)`, marks N for resend, and starts N+1. Every subsequent block repeats the pattern.

This surfaced *after* Fix 1 got the tool past the initial `DUPLICATE_BLOCK` cascade — steady-state push produced the `Handler N is ending mid-block X` burst pattern on previewnet with zero blocks ever persisted.

**Changes:** emit `PublishStreamRequest(end_of_block = BlockEnd{blockNumber = qb.blockNumber})` after all item batches for a block.

## Test plan

- [x] `./gradlew :tools:test --tests LiveBlockPushClientTest` — green. Existing `queryReturnsMinusOneOnUnreachable` case replaced by `queryThrowsOnUnreachable` asserting `QueryFailedException` + host:port in the message.
- [x] `./gradlew spotlessApply` clean.
- [x] Smoke-tested end-to-end on previewnet: Fix 1 aborted cleanly against wrong port (previously would have looped forever); after passing `--push-bn-status-port 40982` the tool reached steady-state push. Fix 2 unblocked wrap on testnet with corrected block-time decoding. Fix 3 identified by observing `Handler N ending mid-block X` bursts even after Fix 1 landed.
- [ ] Full CI green.

###

```

harpender@previewnet-wrapper:~$ kubectl logs -n block-node block-node-block-node-server-0 --since=10m 2>&1 | \
    grep -iE 'block.*complete|block.*persisted|block.*acked' | tail -20
2026-08-19 03:11:57.057+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,289 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.057+0000 FINE    [org.hiero.block.node.stream.publisher.LiveStreamPublisherManager closeBlock] Completed blocks 9,287
2026-08-19 03:11:57.059+0000 FINE    [org.hiero.block.node.stream.publisher.LiveStreamPublisherManager closeBlock] Completed blocks 9,288
2026-08-19 03:11:57.060+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,290 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.070+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,291 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.080+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,292 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.090+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,293 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.100+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,294 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.191+0000 FINE    [org.hiero.block.node.stream.publisher.LiveStreamPublisherManager closeBlock] Completed blocks 9,289
2026-08-19 03:11:57.194+0000 FINE    [org.hiero.block.node.stream.publisher.LiveStreamPublisherManager closeBlock] Completed blocks 9,290
2026-08-19 03:11:57.195+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,295 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.196+0000 FINE    [org.hiero.block.node.stream.publisher.LiveStreamPublisherManager closeBlock] Completed blocks 9,291
2026-08-19 03:11:57.198+0000 FINE    [org.hiero.block.node.stream.publisher.LiveStreamPublisherManager closeBlock] Completed blocks 9,292
2026-08-19 03:11:57.199+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,296 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.200+0000 FINE    [org.hiero.block.node.stream.publisher.LiveStreamPublisherManager closeBlock] Completed blocks 9,293
2026-08-19 03:11:57.206+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,297 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.216+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,298 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.226+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,299 succeeded=true source=PUBLISHER
2026-08-19 03:11:57.239+0000 FINE    [org.hiero.block.node.stream.publisher.LiveStreamPublisherManager closeBlock] Completed blocks 9,294
2026-08-19 03:11:57.245+0000 FINE    [org.hiero.block.node.messaging.BlockMessagingFacilityImpl lambda$sendBlockPersisted$0] Sending block persisted notification: block=1,083,300 succeeded=true source=PUBLISHER
harpender@previewnet-wrapper:~$

```

BN Status calls:

```

harpender@previewnet-wrapper:~$ ./bn-status.sh
Forwarding from 127.0.0.1:18082 -> 40982
Forwarding from [::1]:18082 -> 40982
Handling connection for 18082
{
  "lastAvailableBlock": "1083890",
  "onlyLatestState": true,
  "nextExpectedBlock": "1083891"
}
harpender@previewnet-wrapper:~$ ./bn-status.sh
Forwarding from 127.0.0.1:18082 -> 40982
Forwarding from [::1]:18082 -> 40982
Handling connection for 18082
{
  "lastAvailableBlock": "1084028",
  "onlyLatestState": true,
  "nextExpectedBlock": "1084030"
}
harpender@previewnet-wrapper:~$
```

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
@@ -205,6 +205,16 @@ public class LiveSequential implements Runnable {
     private int pushBnPort = 40840;
 
     @Option(
+            names = {"--push-bn-status-port"},
+            description = "Block Node BlockNodeService port for the one-time serverStatus query "
+                    + "used to establish the push-skip watermark. Defaults to --push-bn-port for the "
+                    + "common single-port deployment. Override when the BN uses per-API split ports "
+                    + "(publish and status on different ports); a wrong port here causes the startup "
+                    + "query to fail, and the tool aborts rather than pushing every block from 0 "
+                    + "(see #3374). Ignored unless --push-enabled is set. -1 means 'unset'.")
+    private int pushBnStatusPort = -1;
+
+    @Option(
             names = {"--push-queue-capacity"},
             description = "Bounded backpressure queue between the wrap thread and the push worker "
                     + "(default: 32). When full, the wrap thread blocks; nothing is dropped. "
@@ -360,12 +370,39 @@ public class LiveSequential implements Runnable {
             // side coordinating with the other. The push subsystem owns its own thread + bounded
             // queue (see LiveBlockPushClient javadoc).
             if (pushEnabled) {
+                final int effectiveStatusPort = pushBnStatusPort > 0 ? pushBnStatusPort : pushBnPort;
                 pushClient = new LiveBlockPushClient(
-                        pushBnHost, pushBnPort, pushQueueCapacity, LiveBlockPushClient.loadDefaultWebConfig());
-                pushSkipUpToInclusive = pushClient.queryLastAvailableBlock();
+                        pushBnHost,
+                        pushBnPort,
+                        effectiveStatusPort,
+                        pushQueueCapacity,
+                        LiveBlockPushClient.loadDefaultWebConfig());
+                // Fail-fast on query failure. Silently proceeding with pushSkipUpToInclusive=-1
+                // would push every block from 0 into a BN that already has thousands, driving the
+                // BN into the DUPLICATE_BLOCK/SkipBlock reconnect loop that motivated #3374.
+                try {
+                    pushSkipUpToInclusive = pushClient.queryLastAvailableBlock();
+                } catch (final LiveBlockPushClient.QueryFailedException e) {
+                    throw new IllegalStateException(
+                            "Cannot enable --push-enabled: " + e.getMessage()
+                                    + ". Remove --push-enabled or point --push-bn-status-port at the BlockNodeService port.",
+                            e);
+                }
+                if (pushSkipUpToInclusive < 0) {
+                    // Defense in depth: the BN should never advertise a negative watermark, but if
+                    // it does, the producer-side `blockNum > pushSkipUpToInclusive` guard would
+                    // still push everything. Abort with a diagnosable message.
+                    throw new IllegalStateException("BN returned lastAvailableBlock=" + pushSkipUpToInclusive
+                            + " which is not a valid watermark; aborting to avoid pushing every block from 0");
+                }
                 System.out.printf(
-                        "[live-sequential] Live push enabled: target=%s:%d queueCapacity=%d BN lastAvailableBlock=%d (blocks <= this are skipped from push)%n",
-                        pushBnHost, pushBnPort, pushQueueCapacity, pushSkipUpToInclusive);
+                        "[live-sequential] Live push enabled: publish=%s:%d status=%s:%d queueCapacity=%d BN lastAvailableBlock=%d (blocks <= this are skipped from push)%n",
+                        pushBnHost,
+                        pushBnPort,
+                        pushBnHost,
+                        effectiveStatusPort,
+                        pushQueueCapacity,
+                        pushSkipUpToInclusive);
                 pushClient.start();
             }
 
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
@@ -388,21 +388,21 @@ public class LiveSequential implements Runnable {
                                     + ". Remove --push-enabled or point --push-bn-status-port at the BlockNodeService port.",
                             e);
                 }
-                if (pushSkipUpToInclusive < 0) {
-                    // Defense in depth: the BN should never advertise a negative watermark, but if
-                    // it does, the producer-side `blockNum > pushSkipUpToInclusive` guard would
-                    // still push everything. Abort with a diagnosable message.
-                    throw new IllegalStateException("BN returned lastAvailableBlock=" + pushSkipUpToInclusive
-                            + " which is not a valid watermark; aborting to avoid pushing every block from 0");
-                }
+                // An empty BN advertises lastAvailableBlock as uint64 max, which arrives here as
+                // -1L. That's the correct sentinel for "push everything from block 0" — the
+                // producer-side guard `blockNum > pushSkipUpToInclusive` gives us exactly that.
+                final boolean bnEmpty = pushSkipUpToInclusive == -1L;
                 System.out.printf(
-                        "[live-sequential] Live push enabled: publish=%s:%d status=%s:%d queueCapacity=%d BN lastAvailableBlock=%d (blocks <= this are skipped from push)%n",
+                        "[live-sequential] Live push enabled: publish=%s:%d status=%s:%d queueCapacity=%d BN %s%n",
                         pushBnHost,
                         pushBnPort,
                         pushBnHost,
                         effectiveStatusPort,
                         pushQueueCapacity,
-                        pushSkipUpToInclusive);
+                        bnEmpty
+                                ? "is empty (will push from block 0)"
+                                : "lastAvailableBlock=" + pushSkipUpToInclusive
+                                        + " (blocks <= this are skipped from push)");
                 pushClient.start();
             }
 
--- a/tools-and-tests/tools/src/test/java/org/hiero/block/tools/push/LiveBlockPushClientTest.java
+++ b/tools-and-tests/tools/src/test/java/org/hiero/block/tools/push/LiveBlockPushClientTest.java
@@ -3,6 +3,7 @@ package org.hiero.block.tools.push;
 
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -28,9 +29,24 @@ class LiveBlockPushClientTest {
     /** Port 1 is reserved and effectively always refuses TCP connections — used as an unreachable target. */
     private static final int UNREACHABLE_PORT = 1;
 
+    /**
+     * A second reserved/unreachable port, distinct from {@link #UNREACHABLE_PORT}, so a split-port
+     * test can tell which of the two client URIs the code actually contacted.
+     */
+    private static final int UNREACHABLE_STATUS_PORT = 7;
+
     private static LiveBlockPushClient newClient(final int queueCapacity) {
         return new LiveBlockPushClient(
                 "127.0.0.1", UNREACHABLE_PORT, queueCapacity, LiveBlockPushClient.loadDefaultWebConfig());
+    }
+
+    private static LiveBlockPushClient newSplitPortClient(final int queueCapacity) {
+        return new LiveBlockPushClient(
+                "127.0.0.1",
+                UNREACHABLE_PORT,
+                UNREACHABLE_STATUS_PORT,
+                queueCapacity,
+                LiveBlockPushClient.loadDefaultWebConfig());
     }
 
     @Nested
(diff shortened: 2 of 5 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
```diff
@@ -360,12 +370,39 @@ public void run() {
...
+                // BN into the DUPLICATE_BLOCK/SkipBlock reconnect loop that motivated #3374.
+                try {
+                    pushSkipUpToInclusive = pushClient.queryLastAvailableBlock();
+                } catch (final LiveBlockPushClient.QueryFailedException e) {
+                    throw new IllegalStateException(
+                            "Cannot enable --push-enabled: " + e.getMessage()
+                                    + ". Remove --push-enabled or point --push-bn-status-port at the BlockNodeService port.",
+                            e);
+                }
+                if (pushSkipUpToInclusive < 0) {
```
Nana-EC: I'm not sure this guard is correct. 
A new Tier 0 BN with empty state would receive -1 (18446744073709551615) in which case it would be the correct thing to push from block 0

tools-and-tests/tools/src/test/java/org/hiero/block/tools/push/LiveBlockPushClientTest.java
```diff
@@ -98,12 +99,19 @@ void initialMetrics() {
     class UnreachableBn {
 
         @Test
-        @DisplayName("queryLastAvailableBlock() returns -1 when the BN is unreachable")
+        @DisplayName("queryLastAvailableBlock() throws QueryFailedException when the BN is unreachable")
         @Timeout(60)
-        void queryReturnsMinusOneOnUnreachable() {
+        void queryThrowsOnUnreachable() {
             try (final LiveBlockPushClient client = newClient(8)) {
```
Nana-EC, line 121: So this test uses the old webClient and not the buildStatusWebClient you just added. I don't think you're testing the same flow you just added so it wouldn't catch regressions in this area.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/LiveSequential.java
rockysingh: Good find. I will get this addressed.

tools-and-tests/tools/src/test/java/org/hiero/block/tools/push/LiveBlockPushClientTest.java
rockysingh: Good point will get this updated.

jsync-swirlds (approved):

Nana-EC (approved):

PR #3448, opened 2026-08-17
fix(release): alpha version derivation fails on first alpha for a new base version

### LINKED ISSUE #3447: fix(release): alpha version derivation fails on first alpha for a new base version
## Description

The **Compute Release Version** step in `release-automation.yaml` fails when no prior alpha tags exist for the current base version. This means the very first alpha release attempt for any new minor version (e.g. `v0.41.0-alpha1`) kills the job.

**Observed failure:** https://github.com/hiero-ledger/hiero-block-node/actions/runs/32034528518/job/95401738368

## Root Cause

```bash
HIGHEST_ALPHA=$(gh api "repos/${{ github.repository }}/tags" --paginate --jq '.[].name' \
  | grep -E "^v${BASE}-alpha[0-9]+$" \
  | sed -E "s/^v${BASE}-alpha//" \
  | sort -n | tail -1)
```

With `set -eo pipefail`, `grep` exits 1 when it finds no matching tags. That propagates through the pipeline and terminates the script before `${HIGHEST_ALPHA:-0}` can default to 0. The intended fallback (counter = 0 → alpha1) is never reached.

## Fix

Append `|| true` to the pipeline so a zero-match result is treated as an empty string, allowing the `${HIGHEST_ALPHA:-0}` default to take effect and produce `alpha1` as intended.

## Steps to Reproduce

1. Trigger an alpha release on `main` when no tags matching `v<current-base>-alpha*` exist yet.
2. The "Compute Release Version" step exits non-zero with a `grep` failure.

## Context

- Workflow: `.github/workflows/release-automation.yaml`
- Step: **Compute Release Version** (around line 84)
- Shell flag in effect: `set -eo pipefail`

Nana-EC:
## Summary

- Fixes `grep` pipefail killing the "Compute Release Version" step when no prior alpha tags exist for the current base version
- Adds `|| true` so a zero-match result produces an empty `HIGHEST_ALPHA`, allowing the `${HIGHEST_ALPHA:-0}` default to compute the correct `alpha1` counter

Closes #3447

## Root Cause

`set -eo pipefail` propagates `grep`'s exit code 1 (no matches found) through the pipeline, terminating the script before the `${HIGHEST_ALPHA:-0}` fallback is reached. Observed failure: https://github.com/hiero-ledger/hiero-block-node/actions/runs/32034528518/job/95401738368

## Test Plan

- [ ] Trigger an alpha release on main when no prior `v<base>-alpha*` tags exist → should compute `alpha1` instead of failing
- [ ] Trigger a second alpha release → should compute `alpha2` (existing path, still works)

### DIFF
--- a/.github/workflows/release-automation.yaml
+++ b/.github/workflows/release-automation.yaml
@@ -83,12 +83,13 @@ jobs:
 
             # grep exits 1 when no alpha tags exist yet (first alpha for this base
             # version). With set -eo pipefail that would kill the script before the
-            # ${HIGHEST_ALPHA:-0} fallback is reached. || true absorbs the non-zero
-            # exit so an empty result is treated as counter = 0 → alpha1.
+            # ${HIGHEST_ALPHA:-0} fallback is reached. || true is scoped to grep only
+            # so an empty result is treated as counter = 0 → alpha1, while real
+            # failures from gh api/sed/sort still fail the pipeline via pipefail.
             HIGHEST_ALPHA=$(gh api "repos/${{ github.repository }}/tags" --paginate --jq '.[].name' \
-              | grep -E "^v${BASE}-alpha[0-9]+$" \
+              | { grep -E "^v${BASE}-alpha[0-9]+$" || true; } \
               | sed -E "s/^v${BASE}-alpha//" \
-              | sort -n | tail -1 || true)
+              | sort -n | tail -1)
             COMPUTED_VERSION="${BASE}-alpha$(( ${HIGHEST_ALPHA:-0} + 1 ))"
           else
             # Read version.txt from the dispatching branch — no checkout needed yet.
--- a/.github/workflows/release-automation.yaml
+++ b/.github/workflows/release-automation.yaml
@@ -81,10 +81,14 @@ jobs:
               exit 1
             fi
 
+            # grep exits 1 when no alpha tags exist yet (first alpha for this base
+            # version). With set -eo pipefail that would kill the script before the
+            # ${HIGHEST_ALPHA:-0} fallback is reached. || true absorbs the non-zero
+            # exit so an empty result is treated as counter = 0 → alpha1.
             HIGHEST_ALPHA=$(gh api "repos/${{ github.repository }}/tags" --paginate --jq '.[].name' \
               | grep -E "^v${BASE}-alpha[0-9]+$" \
               | sed -E "s/^v${BASE}-alpha//" \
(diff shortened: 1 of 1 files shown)


AlfredoG87 (approved):

AlfredoG87 (dismissed):
LGTM

.github/workflows/release-automation.yaml
```diff
@@ -81,10 +81,14 @@ jobs:
               exit 1
             fi
 
+            # grep exits 1 when no alpha tags exist yet (first alpha for this base
+            # version). With set -eo pipefail that would kill the script before the
+            # ${HIGHEST_ALPHA:-0} fallback is reached. || true absorbs the non-zero
+            # exit so an empty result is treated as counter = 0 → alpha1.
             HIGHEST_ALPHA=$(gh api "repos/${{ github.repository }}/tags" --paginate --jq '.[].name' \
```
berryware, line 89: If it's the grep exit status you want to isolate, The true can go after the grep like this:
```
HIGHEST_ALPHA=$(gh api "repos/${{ github.repository }}/tags" --paginate --jq '.[].name' \
              | { grep -E "^v${BASE}-alpha[0-9]+$" || true; } \
              | sed -E "s/^v${BASE}-alpha//" \
              | sort -n | tail -1)
```
This will let the other errors fail fast

ata-nas: I think this is a good idea, cause we want to catch other failures.

ata-nas (commented):
I think that @berryware has a suggestion that may be worthwhile. We want to limit the `true` flag only for the grep, because we do not want to silence other failures.

.github/workflows/release-automation.yaml
ata-nas: On a second look, if grep fails, the rest is expected to fail as well. It might be better to leave the true at the end.

ata-nas (dismissed):
Approving, on a second look, if grep fails, the rest should also be expected to fail.

.github/workflows/release-automation.yaml
Nana-EC: @berryware good suggestion.
I adopted it.

ata-nas (approved):

ata-nas:
@Nana-EC your CI fails due to the flaky backfill tests, I've just merged the fix for those, you can rebase on main.

rockysingh (approved):

jsync-swirlds (approved):

PR #3486, opened 2026-08-18
test(simulator): lastKnownStatuses deques are not thread-safe, causing NPE flakes in e2e suites

### LINKED ISSUE #3446: test(simulator): lastKnownStatuses deques are not thread-safe, causing NPE flakes in e2e suites
## Observed failure

E2E Test Suites run on PR #3443 failed with:

```
NegativeSingleSubscriberTests > Should return correct status when subscription request has invalid end block FAILED
    java.lang.NullPointerException at NegativeSingleSubscriberTests.java:276
```

Failing job: https://github.com/hiero-ledger/hiero-block-node/actions/runs/32024963345/job/95372342688?pr=3443

## Root cause

Line 276 itself cannot NPE. Gradle prints exceptions in short format, showing only the first test-class frame; the real origin is inside `List.copyOf(lastKnownStatuses)`:

* `PublishStreamGrpcClientImpl`, `ConsumerStreamGrpcClientImpl` and `PublishStreamGrpcServerImpl` keep `lastKnownStatuses` in a plain, non-thread-safe `ArrayDeque`.
* gRPC callback threads mutate it concurrently: `PublishStreamObserver.onNext`/`onError` (`pollFirst` + `add`), and `init()` calls `clear()`.
* Test threads busy-spin on `getStreamStatus()`, which snapshots the deque via `List.copyOf`. A torn `ArrayDeque.toArray()` under concurrent mutation can contain a null slot, and `List.copyOf` throws `NullPointerException` on null elements.

The log timing confirms it: the NPE fired exactly while the observer thread was appending acknowledgement statuses for blocks 2 and 3.

## Fix

* Replace the shared `ArrayDeque`s with `ConcurrentLinkedDeque` in the three gRPC impl classes.
* Add a concurrency regression unit test.
* Harden the busy-wait loops in `NegativeSingleSubscriberTests` with a poll-sleep-timeout helper in `BaseSuite` so a dead simulator fails fast instead of spinning at 100% CPU until the CI job times out.

ata-nas:
* replace ArrayDeque with ConcurrentLinkedDeque for lastKnownStatuses in PublishStreamGrpcClientImpl, ConsumerStreamGrpcClientImpl and PublishStreamGrpcServerImpl
* add concurrency regression tests that snapshot statuses while gRPC callback threads mutate them
* add waitFor poll-sleep-timeout helper to BaseSuite
* replace busy-spin status polls in NegativeSingleSubscriberTests with bounded waitFor calls

Resolves #3446

### DIFF
--- a/tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
+++ b/tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.suites;
 
+import static org.junit.jupiter.api.Assertions.fail;
+
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import java.io.IOException;
@@ -9,6 +11,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.attribute.PosixFilePermissions;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
@@ -16,6 +19,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BooleanSupplier;
 import java.util.stream.Stream;
 import org.hiero.block.api.protoc.BlockAccessServiceGrpc;
 import org.hiero.block.api.protoc.BlockNodeServiceGrpc;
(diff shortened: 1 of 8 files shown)


tools-and-tests/suites/src/main/java/org/hiero/block/suites/BaseSuite.java
```diff
@@ -428,6 +432,32 @@ protected Future<?> startSimulatorInThread(BlockStreamSimulatorApp blockStreamSi
...
+    /**
+     * Polls the given condition until it becomes true, sleeping briefly between polls. Fails the
+     * calling test if the condition is not met within the timeout, so a dead counterpart (for
+     * example a simulator that never produced a status) fails fast instead of spinning until the
+     * CI job times out.
+     *
+     * @param condition the condition to poll
+     * @param timeout the maximum time to wait for the condition
+     */
+    protected static void waitFor(final BooleanSupplier condition, final Duration timeout) {
```
ivannov, line 444: The class `PositiveMultiplePublishersTest` has a method `awaitCondition()` which serves the same purpose as this one, but is slightly more flexible (contains configurable sleep period as we as a description). Maybe you could pull that up here and use that across tests?

ivannov (approved):
Just one small nit, you may ignore it, approving

AlfredoG87 (approved):
LGTM 👍

PR #3495, opened 2026-08-19
feat(tools): improve bn-endpoint-checker network diagnostics (#3489)

### LINKED ISSUE #3489: feat(tools): improve bn-endpoint-checker network diagnostics — drop TCP pre-check, add ping RTT and throughput reporting
## Summary

Three improvements to `bn-endpoint-checker.sh` that make network diagnostics more reliable and informative:

### 1 — Replace TCP pre-check with grpcurl `--connect-timeout`

The current `nc`-based TCP probe has two problems: it requires `nc` as a dependency, and without `timeout`/`gtimeout` (not built into macOS) it can hang for 75–150 s on DROP-firewalled nodes rather than the intended 3 s. The pre-check was originally a "fail fast" gate before gRPC, but grpcurl already provides `-connect-timeout` which does the same job more reliably. Removing the TCP check eliminates the `nc` dependency and the portability problem.

### 2 — Add ping RTT probe per endpoint

Before the gRPC calls, run `ping -c 3` to report min/avg/max RTT. This immediately tells the operator whether slowness is their network or the server — without needing a separate tool. Falls back gracefully when ICMP is blocked.

### 3 — Report effective throughput from block download

When `--latest-block-proof` is used and the block fetch succeeds, compute and display effective download throughput (response bytes / elapsed time). Gives a rough bandwidth indicator using data already being fetched — no `iperf3` or server-side setup needed.

## Context

Observed during testing against `lfh-values.yaml` multi-port environments:
- s01 TCP probe took 75 s (SYN retransmission backoff — no `timeout` binary on the test machine)
- s02/s03 hit 150 s TCP timeout when reset mid-test

Builds on `feat/3424-bn-endpoint-checker-block-access-port` (PR #3426).

### LINKED ISSUE #3497: chore: follow-up on non-blocking review comments from #3426 (feat(tools): add block-access-port and fix ResourceExhausted in bn-endpoint-checker.sh)
## Story Form

As a maintainer
I want all non-blocking review feedback from #3426 to be tracked

## Technical Notes

Follow-up items from unresolved review comments on #3426.

- [ ] **`tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh` line 252 — regex allows leading-zero numbers, enabling silent octal interpretation** (`berryware`): The argument-parsing regex for one of the new numeric flags (likely `--block-access-port` or `--max-block-sz`) permits leading zeros (e.g. `010`, `08`). In shells that interpret numeric literals as octal this causes unexpected values; `08` / `09` also fail as invalid octal. Tighten the regex to `^[1-9][0-9]*$` (or `^(0|[1-9][0-9]*)$` if zero is valid) so leading zeros are rejected at parse time.

- [ ] **`tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh` line 248 — `--block-access-port` argument lacks numeric validation** (`berryware`): Unlike the other new numeric flags, `--block-access-port` does not appear to validate that the supplied value is a positive integer. Add the same pattern-match guard used for the other port/size arguments so invalid input is caught with a clear error rather than passed to grpcurl.

Nana-EC:
## Summary
- Replaces `nc`-based TCP pre-check with grpcurl `--connect-timeout 5` — eliminates the `nc` dependency and the 75–150 s hang on DROP-firewalled nodes when `timeout`/`gtimeout` is absent (common on macOS)
- Adds ping RTT probe per endpoint to distinguish network latency from server-side slowness
- Reports effective download throughput from block fetch when `--latest-block-proof` is used

Closes #3489

## Follow-up commits
- **Fall back to TCP-connect RTT when ICMP is blocked**: many cloud/k8s firewalls drop ICMP by policy while leaving the service port open, which made the ping RTT probe go blank on otherwise-healthy nodes. When ping fails, the script now attempts a raw TCP connect via bash's `/dev/tcp` against the target port (bounded by a self-managed timeout — no `timeout` binary dependency), distinguishing a successful handshake, an immediate refusal (host up, port closed), and a timeout (SYN likely dropped by a firewall). Addresses review feedback from jsync.
- **Validate `--block-access-port` and tighten numeric flag regex**: fixes both follow-up items from #3497 — `--block-access-port` previously had no numeric validation (unlike the other port/size flags) so bad input silently flowed through to grpcurl instead of failing fast; `--max-block-sz`'s regex also allowed leading-zero values. Closes #3497.

Verified against BNCE endpoints: confirmed the ICMP-blocked/TCP-fallback path fires correctly with support for multi-port.

## Test plan
- [x] Run against a reachable node — should show ping RTT, green gRPC, and throughput line
- [x] Run against an unreachable host — should fail at serverStatus with connect-timeout error (no more 75–150 s nc hang)
- [x] Run with `--latest-block-proof` — should show `throughput (approx)` line
- [ ] Run on macOS without coreutils — should work without `nc`/`timeout`/`gtimeout`

### DIFF
--- a/tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
+++ b/tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
@@ -5,14 +5,13 @@
 # =============================================================================
 #
 # Checks one or more Block Node gRPC endpoints by:
-#   1. Verifying TCP reachability  (nc)
-#   2. Calling serverStatus        → reports first/last available block,
+#   1. Calling serverStatus        → reports first/last available block,
 #                                    only_latest_state flag
-#   3. Calling serverStatusDetail  → reports BN software version, stream proto
+#   2. Calling serverStatusDetail  → reports BN software version, stream proto
 #                                    version, available block ranges, registered
 #                                    plugins, and TSS data presence
 #                                    (only when --detailed-server-status is passed)
-#   4. Fetching the latest block   → reports the block proof type observed in the
+#   3. Fetching the latest block   → reports the block proof type observed in the
 #                                    most recent block: WRB/RSA (Phase 2a),
 #                                    TSS hinTS + WRAPS proof, or TSS hinTS +
 #                                    Aggregate Schnorr signature (Phase 2b)
@@ -105,7 +104,6 @@
 # REQUIREMENTS
 #   grpcurl   Auto-installed if missing (you will be prompted for the
 #             install location).
-#   nc        Netcat — used for the TCP reachability check.
 #   jq        JSON processor — used for pretty-printing gRPC responses.
 #   curl      Used to download grpcurl and proto archives when needed.
 #
@@ -125,8 +123,8 @@
 #   These are all present in the official hiero-block-node release archive.
 #
 # EXIT CODES
-#   0   All endpoints passed TCP and serverStatus checks.
-#   1   One or more endpoints failed (TCP unreachable or serverStatus error).
+#   0   All endpoints passed serverStatus checks.
+#   1   One or more endpoints failed (serverStatus error).
 #   2   Usage / configuration error (bad arguments, missing dependencies).
 # =============================================================================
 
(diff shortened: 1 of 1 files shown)


AlfredoG87 (approved):
LGTM

jsync-swirlds (approved):

PR #3498, opened 2026-08-19
chore(chart): clarify health port Services logic and update outdated values references

### LINKED ISSUE #3496: chore: follow-up on non-blocking review comments from #3470 (fix(chart): exclude health port from includePluginPorts on LoadBalancer)
## Story Form

As a maintainer
I want all non-blocking review feedback from #3470 to be tracked

## Technical Notes

Follow-up items from unresolved review comments on #3470.

- [ ] **`charts/block-node-server/templates/_helpers.tpl` line 126 — remove stale `service.extraPorts` reference from docstring** (`berryware`): The updated `pluginServicePorts` docstring says operators can add health explicitly "via `loadBalancer.extraPorts` or `service.extraPorts`", but `service.extraPorts` is not wired in `service.yaml` — only `loadBalancer.extraPorts` exists. Either add an `extraPorts` loop to `service.yaml` (so the reference is accurate) or drop "or `service.extraPorts`" from the comment.

- [ ] **`charts/block-node-server/templates/_helpers.tpl` line 131 — confirm / document that removing health from `pluginServicePorts` intentionally drops it from the ClusterIP Service too** (`berryware`): `pluginServicePorts` is used by both `service-loadbalancer.yaml` and `service.yaml`, so the health port (40983) is now absent from both the public LB and the ClusterIP Service. The PR docstring should be updated to make this explicit, and the values comment for `blockNode.ports.health` should note that 40983 is declared as a container port (so probes work) but is not added to any Service by default.

Nana-EC:
## Summary

- Docstring and comments cleanup following `berryware`'s unresolved review comments on #3470, tracked in #3496.
- Correction of two config errors in the LFH production values overrides.

### Health port docstring cleanup (closes #3496)

- **`_helpers.tpl`**: removes the non-existent `service.extraPorts` reference and explicitly states that health (40983) is excluded from **both** the ClusterIP and LB Services — `pluginServicePorts` is called from both `service.yaml` and `service-loadbalancer.yaml`.
- **`values.yaml` `includePluginPorts` comment**: corrects "ClusterIP-internal only" to "excluded from all Services".
- **`values.yaml` `blockNode.ports.health` comment**: adds a note that 40983 is a containerPort declaration (probes and hostPort bindings work) but is not added to any Service by default; directs operators to `loadBalancer.extraPorts` for explicit LB exposure.

### LFH production values corrections

- **`lfh-provisioner-config.yaml`**: bumps chart version from `0.37.1` to `0.40.1`.
- **`lfh-values.yaml`**: moves `metallb.io/address-pool` under the `annotations:` parent key (the bare key form is not a valid `loadBalancer` field and would be ignored by Helm); adds `includePluginPorts: true` so the plugin service ports are exposed on the LoadBalancer.

## Test plan

- [ ] `helm template` with `includePluginPorts: true` does not include a `health` port in either `service.yaml` or `service-loadbalancer.yaml` output
- [ ] `helm template` with health in `loadBalancer.extraPorts` includes it on the LB (opt-in still works)
- [ ] `helm template` with `lfh-values.yaml` renders `metallb.io/address-pool` under `annotations:` in the LB Service manifest

### DIFF
--- a/charts/block-node-server/templates/_helpers.tpl
+++ b/charts/block-node-server/templates/_helpers.tpl
@@ -119,11 +119,12 @@ Usage: include "hiero-block-node.pluginPortEnvVars" .
 
 {{/*
 Emit Service port entries for each non-null plugin port in blockNode.ports.
-The health port is intentionally excluded: probes go directly to the pod port
-(by number) and do not need a Service entry. Excluding it prevents health from
-appearing on a public LoadBalancer when includePluginPorts is true. Operators
-who want health in a Service (either ClusterIP or LB) can add it explicitly
-via loadBalancer.extraPorts or service.extraPorts.
+The health port is intentionally excluded from both the ClusterIP and the
+LoadBalancer Service: probes go directly to the pod port by number and do not
+need a Service entry, so advertising 40983 in any Service serves no functional
+purpose and would expose the /healthz endpoint on the public LB when
+includePluginPorts is true. Operators who want health in a Service for any
+reason can add it explicitly via loadBalancer.extraPorts.
 Multiple plugins that share the same port number emit only one Service port entry
 (Kubernetes rejects duplicate port numbers within a Service).
 Usage: include "hiero-block-node.pluginServicePorts" . | nindent 4
--- a/charts/block-node-server/values-overrides/lfh-provisioner-config.yaml
+++ b/charts/block-node-server/values-overrides/lfh-provisioner-config.yaml
@@ -7,7 +7,7 @@ blockNode:
     namespace: block-node
     release: block-node
     chart: oci://ghcr.io/hiero-ledger/hiero-block-node/block-node-server
-    version: "0.37.1"
+    version: "0.40.1"
     storage:
         applicationStatePath: /mnt/fast-storage/block-node/application-state
         applicationStateSize: 50Gi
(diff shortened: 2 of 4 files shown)


AlfredoG87 (approved):
LGTM

charts/block-node-server/templates/_helpers.tpl
```diff
@@ -119,11 +119,12 @@ Usage: include "hiero-block-node.pluginPortEnvVars" .
...
-(by number) and do not need a Service entry. Excluding it prevents health from
-appearing on a public LoadBalancer when includePluginPorts is true. Operators
-who want health in a Service (either ClusterIP or LB) can add it explicitly
-via loadBalancer.extraPorts or service.extraPorts.
+The health port is intentionally excluded from both the ClusterIP and the
+LoadBalancer Service: probes go directly to the pod port by number and do not
+need a Service entry, so advertising 40983 in any Service serves no functional
+purpose and would expose the /healthz endpoint on the public LB when
+includePluginPorts is true. Operators who want health in a Service for any
+reason can add it explicitly via loadBalancer.extraPorts.
```
jsync-swirlds, line 127: The health port is only properly accessed by the kublet and provisioner.
The operators _should not_ expose that port to anything outside the local host.

jsync-swirlds (approved):
One informational note.

PR #3501, opened 2026-08-19
test: Forward Compatibility Tests with Test Fixture Proto Definition File

### LINKED ISSUE #3376: Forward Compatibility Tests with Test Fixture Proto Definition File
### Story Form

As a BlockHasher
I want to have good tests for forward compatibility
So that I am confident in new logic, but also be able to catch regressions

### Technical Notes

- Based on [this comment](https://github.com/hiero-ledger/hiero-block-node/pull/3299#discussion_r3713779595) we want to add FutureBlockItem proto definition that will be a test fixture, compiled and used for forward compatibility tests

ata-nas:
* Add future_block_item.proto test fixture compiled with the PBJ compiler in a new block-verification testFixtures source set
* Add FutureBlockItem mirroring the BlockItem oneof with future fields covering every category of the numbering rule
* Add MultiFieldBlockItem without a oneof to serialize invalid multi field combinations
* Add OneofWithExtraFieldBlockItem combining a oneof with fields declared outside of it
* Add fixture proto forward compatibility tests hashing PBJ serialized future items parsed by the production schema
* Assert fixture codec bytes are identical to the hand rolled wire bytes and round trip byte for byte
* Cover all single field invariant violations: known type with unknown fields, multiple unknown fields, oneof value with outside fields, empty item
* Document the single field invariant, the first release field rule and the implemented metrics in the forward compatibility design doc
* Fix stale reference root hash javadoc about absent extension leaves

Resolves #3376

### DIFF
--- a/block-node/block-verification/build.gradle.kts
+++ b/block-node/block-verification/build.gradle.kts
@@ -1,11 +1,20 @@
 // SPDX-License-Identifier: Apache-2.0
-plugins { id("org.hiero.gradle.module.library") }
+plugins {
+    id("org.hiero.gradle.module.library")
+    id("org.hiero.gradle.feature.test-fixtures")
+    id("com.hedera.pbj.pbj-compiler")
+}
 
 description = "Hiero Block Node Verification Service"
 
 // Remove the following line to enable all 'javac' lint checks that we have turned on by default
 // and then fix the reported issues.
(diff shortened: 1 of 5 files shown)


AlfredoG87 (approved):
LGTM

jsync-swirlds (approved):

PR #3502, opened 2026-08-19
test: High Load Performance Test

### LINKED ISSUE #2126: High Load Performance Test
# High Load Performance Test

As a block node engineer
I want to run E2E tests with high transaction load
So that I can validate BN behavior under stress conditions

## Description

Create a high-load test that exercises the network with significantly higher throughput than current tests, targeting 5000+ TPS sustained load.

## Goals

- BN handles high transaction load without crash or OOM
- Verification keeps pace with block production (no growing backlog)
- Memory usage remains stable throughout test
- MN continues receiving and importing blocks

## Implementation Notes

**We might need the dedicated HW Runners for this Tests**

Can be implemented using existing test framework:
- Create new test definition (e.g., `high-load.yaml`) with higher NLG parameters
- Use existing `load-start`, `print-metrics`, `load-stop` event types
- Use existing assertions (`no-errors`, `blocks-increasing`, `node-healthy`)
- May need new assertion for MN blocks flowing

## Acceptance Criteria

- [ ] High-load test definition created
- [ ] Test runs successfully on all topologies
- [ ] Added to weekend/RC scheduler (resource intensive)
- [ ] Metrics captured at intervals throughout test

ivannov:
## Reviewer Notes

* Adds a few more assertions to the high-load solo e2e test
* Implements the new assertion types
* Adds tests for the two new assertion types
* Adds a task for running the local utilities tests
* Includes the high load tests in the CI flow

## Related Issue(s)

Closes #2126

### DIFF
--- a/tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
+++ b/tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
@@ -550,6 +550,17 @@ tasks:
         echo ""
         echo "Run a test with: task test:run TEST_FILE=tests/<name>.yaml"
 
+  test:unit:
+    desc: "Run unit tests for script helpers (scripts/test/test-*.sh)"
+    cmds:
+      - |
+        failed=0
+        for f in "{{.SCRIPTS_DIR}}/test/test-"*.sh; do
+          echo "--- $f"
+          bash "$f" || failed=1
+        done
+        exit $failed
+
   # ============================================================================
   # TCK-SDK Tests (matches consensus-node approach: uses latest tags)
   # ============================================================================
--- a/tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
+++ b/tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
@@ -550,17 +550,6 @@ tasks:
         echo ""
         echo "Run a test with: task test:run TEST_FILE=tests/<name>.yaml"
 
-  test:unit:
-    desc: "Run unit tests for script helpers (scripts/test/test-*.sh)"
-    cmds:
-      - |
-        failed=0
-        for f in "{{.SCRIPTS_DIR}}/test/test-"*.sh; do
-          echo "--- $f"
-          bash "$f" || failed=1
-        done
-        exit $failed
-
   # ============================================================================
   # TCK-SDK Tests (matches consensus-node approach: uses latest tags)
   # ============================================================================
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
@@ -1000,15 +1000,32 @@ function assert_block_available {
     fi
 }
 
+# Restart counts from before the test, keyed by node name. Pods on a long-lived
+# cluster may already carry restarts, so node-healthy asserts on the delta.
+declare -A BASELINE_RESTARTS=()
+
+function get_restart_count {
+    local target="$1" count
+    count=$(kctl get pods -n "${NAMESPACE}" -l "app.kubernetes.io/name=${target}" \
+        -o jsonpath='{.items[0].status.containerStatuses[0].restartCount}' 2>/dev/null)
+    echo "${count:-0}"
+}
+
+function snapshot_restart_counts {
+    local bn
+    for bn in $(get_all_block_nodes); do
+        BASELINE_RESTARTS["$bn"]=$(get_restart_count "$bn")
+    done
+}
+
 # Single node health check
 function assert_node_healthy_single {
     local target="$1"
     local status restart_count
     status=$(kctl get pods -n "${NAMESPACE}" -l "app.kubernetes.io/name=${target}" \
         -o jsonpath='{.items[0].status.phase}' 2>/dev/null)
-    restart_count=$(kctl get pods -n "${NAMESPACE}" -l "app.kubernetes.io/name=${target}" \
-        -o jsonpath='{.items[0].status.containerStatuses[0].restartCount}' 2>/dev/null)
-    restart_count="${restart_count:-0}"
+    # Nodes deployed mid-test have no baseline; they start from 0 by definition.
+    restart_count=$(( $(get_restart_count "$target") - ${BASELINE_RESTARTS[$target]:-0} ))
 
     if [[ -z "$status" ]]; then
         echo "${target}: Pod not found"
@@ -1024,7 +1041,7 @@ function assert_node_healthy_single {
         local term_reason
         term_reason=$(kctl get pods -n "${NAMESPACE}" -l "app.kubernetes.io/name=${target}" \
             -o jsonpath='{.items[0].status.containerStatuses[0].lastState.terminated.reason}' 2>/dev/null)
-        echo "${target}: Running but restarted ${restart_count}x (last: ${term_reason:-unknown})"
+        echo "${target}: Running but restarted ${restart_count}x during test (last: ${term_reason:-unknown})"
         return 1
     fi
 
@@ -1982,6 +1999,8 @@ function main {
         echo "ERROR: Namespace '${NAMESPACE}' not found"
         exit 1
     fi
+
+    snapshot_restart_counts
 
     echo "=== Executing Events ==="
     echo ""
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
@@ -834,7 +834,9 @@ function execute_event {
             local script
             script=$(echo "$args" | yq '.script // ""')
             echo "Executing: $script"
-            eval "$script"
+            # Subshell so an `exit` in the script fails just this event. Unguarded it
+            # would terminate run_events' pipeline and silently skip every later event.
+            ( eval "$script" )
             ;;
         clear-block-storage)
             [[ -z "$target" || "$target" == "null" ]] && target=$(echo "$args" | yq '.target // ""')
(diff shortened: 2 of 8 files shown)


Nana-EC (commented):
LG, needs a rebase and would like @AlfredoG87 to check

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
```diff
@@ -947,6 +958,14 @@ function assert_node_healthy_single {
         return 1
     fi
 
+    if [[ "$restart_count" -gt 0 ]]; then
```
berryware, line 1052: non-blocking. If you are running this on an existing cluster, restart_count may already be > 0 before you start. You may want to snapshot this value and check for changes.

berryware (dismissed):
Minor comment

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
ivannov: Good catch, fixed. The runner now snapshots each Block Node's restartCount in main before the events run, so that node-healthy asserts on the delta

Nana-EC (dismissed):
LG, need @AlfredoG87 to review

jsync-swirlds (dismissed):

tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
```diff
@@ -550,6 +550,17 @@ tasks:
         echo ""
         echo "Run a test with: task test:run TEST_FILE=tests/<name>.yaml"
 
+  test:unit:
```
AlfredoG87: [blocking] Duplicate `test:unit` key. There is already a `test:unit` task on main 15 lines above this one (line 538 on this branch) that calls `scripts/test/run-all.sh`. The header of `run-all.sh` even says `Usage: scripts/test/run-all.sh   (or: task test:unit)`.

go-task accepts the duplicate silently and this second one wins, I confirmed with `task --list`. Stricter YAML parsers reject duplicate keys outright.

The new version also loses `run-all.sh`'s failure count and summary line. Suggest just deleting lines 553-562, the existing task already picks up `test-mirror-assertions.sh` by glob. I ran `task test:unit` on your branch and it found all 73 tests.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
```diff
@@ -992,12 +1000,32 @@ function assert_block_available {
     fi
 }
 
+# Restart counts from before the test, keyed by node name. Pods on a long-lived
+# cluster may already carry restarts, so node-healthy asserts on the delta.
+declare -A BASELINE_RESTARTS=()
```
AlfredoG87: [blocking] `declare -A` is bash 4+. macOS system bash is 3.2 and it is first on PATH, so `#!/usr/bin/env bash` picks it up. On your branch:

```
$ /bin/bash scripts/solo-test-runner.sh --test tests/high-load.yaml --topology single --validate
scripts/solo-test-runner.sh: line 1005: declare: -A: invalid option
```

The script only has `set -o pipefail`, no `set -e`, so it keeps going with a regular array and both the write and the read fail on string subscripts:

```
$ /bin/bash -c 'declare -A B=() 2>/dev/null; t=block-node-1; B["$t"]=3; rc=$(( 5 - ${B[$t]:-0} )); echo "rc=$rc"'
bash: B["$t"]: bad array subscript
bash: B: bad array subscript
rc=5
```

So the baseline is never stored and `restart_count` falls back to the absolute pod restart count, which is exactly what the "snapshot the restart count" commit was meant to avoid. On a long lived local kind cluster any earlier restart makes `node-healthy` fail for all 21 test definitions that assert it, not just `high-load`. CI is fine since the runners have bash 5, so this will not show up in the checks.

Easiest fix is to drop the associative array, either a `name=value` string you grep, or write the baselines under `TMP_DIR`. Otherwise add a `BASH_VERSINFO` guard at the top so it fails loudly instead of silently.

tools-and-tests/scripts/solo-e2e-test/tests/high-load.yaml
```diff
@@ -77,3 +81,26 @@ assertions:
...
+    target: all
+
+  - id: mirror-not-lagging
+    type: mirror-lag
+    description: "Mirror Node kept pace with Block Node"
+    target: all
+    args:
+      max_blocks_behind: 30
+
+  - id: verification-queue-healthy
```
AlfredoG87: [suggestion] This one and `mirror-not-lagging` above run after the queue has already drained, so they will pass no matter what happened at peak.

Assertions only start once `run_events` finishes. Load stops at 510s, last event is at 570s, then `all-blocks-increasing` takes at least 60s and `mirror-blocks-flowing` another 60s before this runs. So the queue gets sampled around 3 minutes after the load stopped.

That means #2126's "verification keeps pace with block production, no growing backlog" is not really covered, and `max_blocks_behind: 30` gives the MN importer minutes to catch up before we look.

`metric-threshold` is assertion only right now, there is no event equivalent, so getting a during-load reading needs either a `command` event around t=300 or a new event type. The `blocks-increasing` and `mirror-blocks-flowing` assertions are still useful as post-load liveness checks, it is just these two thresholds that lose their meaning. Fine to scope out in a comment like you did for memory-stability, just want it to be a deliberate call.

tools-and-tests/scripts/solo-e2e-test/scripts/test/test-mirror-assertions.sh
```diff
@@ -0,0 +1,152 @@
...
+
+# Topology resolution used by assert_mirror_lag_single (reads mirror_nodes YAML).
+TOPOLOGIES_DIR="${SCRIPT_DIR}/../../topologies"
+TOPOLOGY="single"
+
+# shellcheck disable=SC1090
+source "${LIB}"
+
+# Mocks override lib definitions; must come AFTER source so they are not overwritten.
+function get_all_mirror_nodes  { echo "mirror-1"; }
```
AlfredoG87, line 25: [suggestion] These two mocks cover the parts most likely to drift. `get_all_mirror_nodes` parses the topology YAML and `get_mn_rest_port` encodes the port allocation from `solo-port-forward.sh`, so both are coupled to external files that can change without anyone noticing here.

Both work without a cluster, so they could be tested for real. `test-topology-decisions.sh` already loops over the actual `topologies/` dir the same way, and the port function is just arithmetic.

The 10 tests you did write are good, especially the `lag == max` boundary.

tools-and-tests/scripts/solo-e2e-test/tests/high-load.yaml
```diff
@@ -3,7 +3,11 @@
 # High Load Test
 #
 # Validates that all Block Nodes correctly receive and store blocks
-# under high load conditions (5000 TPS).
+# under high load conditions (5000 TPS cap) and that Mirror Node
+# keeps pace. Scheduled on: single, paired-3.
+#
+# Out of scope: memory-stability (BN /metrics exposes no heap gauge;
```
AlfredoG87, line 9: [question] Is there a follow-up issue for the memory check? #2126 lists "memory usage remains stable throughout test" as a goal and this PR closes it. The technical reason looks right, I could not find any JVM or heap gauge in `docs/block-node/metrics.md` either, just want a link so it does not get lost.

AlfredoG87 (commented):
Reviewed and ran the harness locally. Design looks good, the `fetch_bn_server_status` extraction and the graceful no-MN skip fit the existing patterns well, and `task test:unit` passes 73/73 including the 10 new ones.

Two blocking items, both easy fixes. Details inline.

1. `Taskfile.yml` now has two `test:unit` keys. The one you added shadows the one already on main.
2. `declare -A` does not work on bash 3.2 (macOS system bash), so the restart baseline never gets recorded and `node-healthy` breaks for every test that uses it.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
jsync-swirlds: Alternatively, we might just recommend that engineers using MacOS install a version of bash less than 20 years old by running `brew install bash` before building Block Node.

It's not good (in my opinion) to hamstring code intended mostly for CI environments (where modern bash is default) because a single vendor installs an ancient version that isn't used by default (default for MacOS is zsh).

tools-and-tests/scripts/solo-e2e-test/tests/high-load.yaml
jsync-swirlds: We gather node metrics in our production setups (though not directly in BN).
Is there a reason the test cannot use O/S measurements?

ivannov: This would bring additional requirements to our kind cluster

I will file an issue and link it here

tools-and-tests/scripts/solo-e2e-test/scripts/test/test-mirror-assertions.sh
ivannov: Leaving these mocked. `get_mn_rest_port` is `5550 + n`, so a real test just restates the formula.

`get_all_mirror_nodes` YAML parsing is already tested against the real `topologies/` directory by `test-topology-decisions.sh`.

tools-and-tests/scripts/solo-e2e-test/Taskfile.yml
ivannov: Removed redundant task

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
ivannov: Changed from the array to a name=count string.

Note that `scripts/wrb-distribution/assert-bn-emb.sh` still has a `declare -A`

tools-and-tests/scripts/solo-e2e-test/tests/high-load.yaml
ivannov: Here is the task: https://github.com/hiero-ledger/hiero-block-node/issues/3602

jsync-swirlds (dismissed):

ata-nas (dismissed):

AlfredoG87 (dismissed):
LGTM

AlfredoG87 (approved):

ata-nas (approved):

PR #3512, opened 2026-08-20
feat: Utilize `FailureInfo` on Failed Verification

### LINKED ISSUE #2977: Utilize `FailureInfo` on Failed Verification

ata-nas:
* Resolve the publisher action for failed verification with an exhaustive switch expression over `FailureType` so enum changes break compilation
* Return a `FailureActionResult` record carrying whether to schedule a resend and the end of stream code, null when the stream stays open
* Schedule a resend for every failure type except cancelled sessions, which cannot be attributed accurately until #3436
* Send EndOfStream `BAD_BLOCK_PROOF` only for an invalid block proof and `ERROR` for malformed block and node side failure types
* Keep the supplying stream open for possibly transient failure types
* Skip the resend for informational failures, the block was already verified successfully
* Pass the end of stream code through `PublisherHandler.handleFailedVerification`
* Update publisher manager and handler tests for the new classification
* Follow ups received as feedback will be addressed in #3562

Resolves #2977

### DIFF
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
@@ -146,6 +146,11 @@ public final class BlockSessionHandler {
         }
         // check if we have an active publisher session, if not, then disregard the items
         if (local != null) {
+            if (blockItems.isEndOfBlock()) {
+                // mark before offering so the session never observes the ending
+                // batch in its deque while still considered incomplete
+                local.markEndOfBlockReceived();
+            }
             local.getBlockItemsDeque().offer(blockItems);
         }
         if (blockItems.isEndOfBlock()) {
@@ -161,6 +166,8 @@ public final class BlockSessionHandler {
     private void processBackfilledItems(final BlockItems blockItems) {
         final BlockVerificationSession session = startNewSession(blockItems, BlockSource.BACKFILL);
         activateSession(session);
+        // a backfilled block always arrives complete in a single batch
+        session.markEndOfBlockReceived();
         session.getBlockItemsDeque().offer(blockItems);
     }
 
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -65,6 +65,7 @@ import org.hiero.block.node.spi.blockmessaging.PersistedNotification;
 import org.hiero.block.node.spi.blockmessaging.PublisherStatusUpdateNotification;
 import org.hiero.block.node.spi.blockmessaging.PublisherStatusUpdateNotification.UpdateType;
 import org.hiero.block.node.spi.blockmessaging.VerificationNotification;
+import org.hiero.block.node.spi.blockmessaging.VerificationNotification.FailureInfo;
 import org.hiero.block.node.spi.blockmessaging.VerificationNotification.FailureType;
 import org.hiero.block.node.spi.threading.ThreadPoolManager;
 import org.hiero.metrics.LongCounter;
@@ -403,9 +404,15 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     /// facility. Only in cases of failed verification, given that the block
     /// number of the failed block satisfies the conditions of it being
     /// higher than the last persisted and lower than the next unstreamed block,
-    /// we will attempt to handle that failure. That includes allowing
-    /// individual handlers to handle that block, but also scheduling the failed
-    /// block to be resent.
+    /// we will attempt to handle that failure. Failures are classified by
+    /// their failure type: a resend is scheduled for every type except
+    /// incomplete, and depending on the type the supplying handler's stream
+    /// is either left open or ended with the appropriate end of stream code.
+    /// An incomplete session means the block was never fully received, so it
+    /// is already handled elsewhere and no action is taken. An informational
+    /// failure, meaning the same block was already verified successfully
+    /// within reasonable recency, is handled the same as a standard one,
+    /// except that no resend is scheduled.
     @Override
     public void handleVerification(@NonNull final VerificationNotification notification) {
         final long blockNumber = notification.blockNumber();
@@ -418,35 +425,98 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
                 && blockNumber > lastPersistedBlockNumber.get()
                 && blockNumber < nextUnstreamedBlockNumber.get();
         if (shouldHandle) {
-            // A note for a flaky test fix for #3391 & #3392:
-            // After investigating, we have found out that the new verification
-            // plugin's report of canceled verifications is being picked up
-            // here, but there is no guarantee that we will accurately point to
-            // the actual publisher that has supplied the failed (canceled) block.
-            // This is true because we only see a snapshot of the current activity.
-            // If publisher 1 supplied block 5 that was later canceled, publisher 2
-            // might have started resending block 5, and the current state
-            // of the publisher manager will point a finger to publisher 2 for
-            // the failure of block 5.
-            // There is more work to be done so that we will be more accurate,
-            // this is captured by @todo(3436).
-            // Completely handling the FailureInfo is captured by @todo(2977).
-            if (notification.failureInfo().failureType() == FailureType.CANCELLED) {
-                LOGGER.log(INFO, "Session for block {0} was cancelled", blockNumber);
-            } else {
-                // Schedule a resend for the block before sending the bad block proof message
-                blocksToResend.add(blockNumber);
-                // Iterate over all handlers and attempt to send the
-                // bad block proof message.
-                for (final PublisherHandler handler : handlers.values()) {
-                    if (handler.handleFailedVerification(blockNumber)) {
-                        // There will always be only one handler that will send the
-                        // bad block proof message. Once we have, we can break.
-                        break;
-                    }
-                }
-            }
+            // A scheduled resend goes to all connected publishers, so we do
+            // not care which handler supplied the block that failed. Further
+            // improvements to the actions taken per failure type, like
+            // tailoring the response to the supplying handler, are captured
+            // by @todo(3436).
+            final FailureInfo failureInfo = notification.failureInfo();
+            final FailureActionResult actionResult = resolveFailureAction(blockNumber, failureInfo);
+            actionResult.handle(blockNumber, failureInfo, blocksToResend, handlers);
         }
+    }
+
+    /// Classifies a verification failure type, logs the failure and resolves
+    /// the operations the manager must perform for it. This is intentionally
+    /// an exhaustive switch expression with no default branch so that adding
+    /// a type to or removing a type from [FailureType] breaks compilation and
+    /// forces this policy to be revisited.
+    ///
+    /// @param blockNumber the block number that failed verification
+    /// @param failureInfo the failure information of the verification
+    /// @return the operations to perform for the failed block
+    private FailureActionResult resolveFailureAction(final long blockNumber, final FailureInfo failureInfo) {
+        return switch (failureInfo.failureType()) {
+            // The proof is proven invalid, a clear publisher fault.
+            case BAD_BLOCK_PROOF -> resendAndEndStream(blockNumber, failureInfo, WARNING, Code.BAD_BLOCK_PROOF);
+            // Possibly transient faults, the block is not proven bad,
+            // a retry may succeed, so the supplying stream stays open.
+            case MISSING_VERIFICATION_DATA, UNABLE_TO_PARSE -> resendOnly(blockNumber, failureInfo);
+            // The block is malformed, but the proof itself is not
+            // proven bad, so the stream ends with a generic error.
+            case MISSING_MANDATORY_FIELD, MISSING_MANDATORY_ITEM, UNSUPPORTED_STREAM_FORMAT ->
+                resendAndEndStream(blockNumber, failureInfo, INFO, Code.ERROR);
+            // Node side faults or capability limitations, the block is
+            // not proven bad, so the stream ends with a generic error.
+            case UNKNOWN_ERROR, UNRECOGNIZED_PROOF_TYPE, UNSUPPORTED_HAPI_VERSION, UNSUPPORTED_ITEM_TYPE ->
+                resendAndEndStream(blockNumber, failureInfo, INFO, Code.ERROR);
+            // The complete block was received, but the session was cancelled
+            // before producing a result (e.g. evicted from the active sessions
+            // buffer). The publishers consider the block delivered, so nothing
+            // else will supply it and a resend must be scheduled. Resends go
+            // to all publishers, attribution is irrelevant.
+            case CANCELLED -> resendOnly(blockNumber, failureInfo);
+            // The block was never fully received, so it is already handled:
+            // either a resend was already scheduled, or another source is
+            // supplying it. No action is needed.
+            case CANCELLED_INCOMPLETE -> logOnlyNoAction(blockNumber, failureInfo);
+        };
+    }
+
+    /// Logs the given verification failure at the given level and resolves a
+    /// resend for the failed block with the supplying stream ended with the
+    /// given code.
+    ///
+    /// @param blockNumber the block number that failed verification
+    /// @param failureInfo the failure information of the verification
+    /// @param logLevel the level to log the failure at
+    /// @param endStreamCode the end of stream code to send to the supplier
+    /// @return the operations to perform for the failed block
+    private FailureActionResult resendAndEndStream(
+            final long blockNumber,
+            final FailureInfo failureInfo,
+            final System.Logger.Level logLevel,
+            final Code endStreamCode) {
+        LOGGER.log(
+                logLevel,
+                "Verification of block {0} failed with {1}, ending the supplying stream with {2}",
+                blockNumber,
+                failureInfo,
+                endStreamCode);
+        return new FailureActionResult(true, endStreamCode);
+    }
+
+    /// Logs the given verification failure and resolves a resend for the
+    /// failed block with the supplying stream left open.
+    ///
+    /// @param blockNumber the block number that failed verification
+    /// @param failureInfo the failure information of the verification
+    /// @return the operations to perform for the failed block
+    private FailureActionResult resendOnly(final long blockNumber, final FailureInfo failureInfo) {
+        LOGGER.log(INFO, "Verification of block {0} failed with {1}", blockNumber, failureInfo);
+        return new FailureActionResult(true, null);
+    }
+
+    /// Logs the given verification failure and resolves no operations.
+    /// This is used for failures that require no further action from the
+    /// manager.
+    ///
+    /// @param blockNumber the block number whose session ended with failure
+    /// @param failureInfo the failure information of the verification
+    /// @return the operations to perform for the failed block
+    private FailureActionResult logOnlyNoAction(final long blockNumber, final FailureInfo failureInfo) {
+        LOGGER.log(INFO, "Session for block {0} failed with {1}, no action needed", blockNumber, failureInfo);
+        return new FailureActionResult(false, null);
     }
 
     /// {@inheritDoc}
@@ -1445,6 +1515,50 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
                 return true;
             } else {
                 return false;
+            }
+        }
+    }
+
+    /// The operations the publisher manager must perform for a verification
+    /// failure, as resolved by
+    /// [#resolveFailureAction(long,FailureInfo)].
+    ///
+    /// @param shouldResend true if a resend must be scheduled for the failed
+    /// block
+    /// @param endStreamCode the end of stream code to send to the supplying
+    /// handler, null if the stream stays open
+    private record FailureActionResult(boolean shouldResend, Code endStreamCode) {
+        /// Performs the resolved operations for the given failed block:
+        /// schedules a resend, unless the failure is informational, meaning
+        /// the same block was already verified successfully within reasonable
+        /// recency, and ends the supplying handler's stream when an end of
+        /// stream code is present. The resend is scheduled before the end of
+        /// stream message is sent.
+        ///
+        /// @param blockNumber    the block number that failed verification
+        /// @param failureInfo    the failure information of the verification
+        /// @param blocksToResend the manager's set of blocks scheduled to be resent
+        /// @param handlers       the manager's currently connected publisher handlers
+        private void handle(
+                final long blockNumber,
+                final FailureInfo failureInfo,
+                final ConcurrentSkipListSet<Long> blocksToResend,
+                final ConcurrentNavigableMap<Long, PublisherHandler> handlers) {
+            // An informational failure happened after the same block was
+            // already verified successfully within reasonable recency, so
+            // no resend is scheduled for it.
+            if (shouldResend && !failureInfo.isInformational()) {
+                // Schedule a resend for the block before sending the end of stream message
+                blocksToResend.add(blockNumber);
+            }
+            if (endStreamCode != null) {
+                for (final PublisherHandler handler : handlers.values()) {
+                    if (handler.handleFailedVerification(blockNumber, endStreamCode)) {
+                        // There will always be only one handler that will send the
+                        // end of stream message. Once we have, we can break.
+                        break;
+                    }
+                }
             }
         }
     }
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -405,14 +405,17 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     /// number of the failed block satisfies the conditions of it being
     /// higher than the last persisted and lower than the next unstreamed block,
     /// we will attempt to handle that failure. Failures are classified by
-    /// their failure type: a resend is scheduled for every type except
-    /// incomplete, and depending on the type the supplying handler's stream
-    /// is either left open or ended with the appropriate end of stream code.
-    /// An incomplete session means the block was never fully received, so it
-    /// is already handled elsewhere and no action is taken. An informational
-    /// failure, meaning the same block was already verified successfully
-    /// within reasonable recency, is handled the same as a standard one,
-    /// except that no resend is scheduled.
+    /// their failure type into three stream scopes: a proven publisher fault
+    /// ends only the supplying handler's stream, malformed blocks with no
+    /// indication of malice keep all streams open with a resend scheduled,
+    /// and Block Node faults or capability limitations end every publisher
+    /// stream and clear all block tracking, encouraging publishers to
+    /// connect to a healthy Block Node. An incomplete session means the
+    /// block was never fully received, so it is already handled elsewhere
+    /// and no action is taken. An informational failure, meaning the same
+    /// block was already verified successfully within reasonable recency,
+    /// is handled the same as a standard one, except that no resend is
+    /// scheduled.
     @Override
     public void handleVerification(@NonNull final VerificationNotification notification) {
         final long blockNumber = notification.blockNumber();
@@ -432,7 +435,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             // by @todo(3436).
             final FailureInfo failureInfo = notification.failureInfo();
             final FailureActionResult actionResult = resolveFailureAction(blockNumber, failureInfo);
-            actionResult.handle(blockNumber, failureInfo, blocksToResend, handlers);
+            actionResult.handle(blockNumber, failureInfo, this);
         }
     }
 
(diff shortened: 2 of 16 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -403,9 +404,15 @@ public PublisherConfig configuration() {
...
     /// number of the failed block satisfies the conditions of it being
     /// higher than the last persisted and lower than the next unstreamed block,
-    /// we will attempt to handle that failure. That includes allowing
-    /// individual handlers to handle that block, but also scheduling the failed
-    /// block to be resent.
+    /// we will attempt to handle that failure. Failures are classified by
+    /// their failure type: a resend is scheduled for every type, and
+    /// depending on the type the supplying handler's stream is either left
+    /// open or ended with the appropriate end of stream code. A cancelled
+    /// session also warrants a resend, but until @todo(3436) is in place it
```
jsync-swirlds: I'm not sure a cancelled stream should include a resend; because a cancel means the block wasn't finished (and it would have already been handled).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -430,25 +437,88 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+    /// @param blockNumber the block number that failed verification
+    /// @param failureInfo the failure information of the verification
+    /// @return the operations to perform for the failed block
+    private FailureActionResult resolveFailureAction(final long blockNumber, final FailureInfo failureInfo) {
+        return switch (failureInfo.failureType()) {
+            // The proof is proven invalid, a clear publisher fault.
+            case BAD_BLOCK_PROOF -> resendAndEndStream(blockNumber, failureInfo, WARNING, Code.BAD_BLOCK_PROOF);
+            // Possibly transient faults, the block is not proven bad,
+            // a retry may succeed, so the supplying stream stays open.
+            case MISSING_VERIFICATION_DATA, UNABLE_TO_PARSE -> resendOnly(blockNumber, failureInfo);
```
jsync-swirlds: Missing verification data is a clear _block node_ fault.
We should send error responses (to all publishers), not resends.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -430,25 +437,88 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+            // The block is malformed, but the proof itself is not
+            // proven bad, so the stream ends with a generic error.
+            case MISSING_MANDATORY_FIELD, MISSING_MANDATORY_ITEM, UNSUPPORTED_STREAM_FORMAT ->
+                resendAndEndStream(blockNumber, failureInfo, INFO, Code.ERROR);
```
jsync-swirlds: Not sure why we end the stream on these, there is no indication of malice, so we should request resend, but not close the stream.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -430,25 +437,88 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+            // Node side faults or capability limitations, the block is
+            // not proven bad, so the stream ends with a generic error.
+            case UNKNOWN_ERROR, UNRECOGNIZED_PROOF_TYPE, UNSUPPORTED_HAPI_VERSION, UNSUPPORTED_ITEM_TYPE ->
+                resendAndEndStream(blockNumber, failureInfo, INFO, Code.ERROR);
+            // A cancelled session also warrants a resend, but it
+            // cannot be reliably attributed to the publisher that
+            // supplied the block, so the resend cannot be scheduled
+            // accurately until @todo(3436) is in place.
+            // This branch should be short-lived.
+            case CANCELLED -> cancelledLogOnly(blockNumber);
```
jsync-swirlds: When we schedule resend, that goes to all publishers; we don't care about attribution. A cancel is already handled because it was never completed so either it was already scheduled for resend or was received from another source.

See [this comment](https://github.com/hiero-ledger/hiero-block-node/pull/3512#discussion_r3826093472) for a related concern, however.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -430,25 +437,88 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+    private FailureActionResult resendOnly(final long blockNumber, final FailureInfo failureInfo) {
+        LOGGER.log(INFO, "Verification of block {0} failed with {1}", blockNumber, failureInfo);
+        return new FailureActionResult(true, null);
+    }
+
+    /// Logs the cancelled verification session and resolves no operations.
+    ///
+    /// @param blockNumber the block number whose session was cancelled
+    /// @return the operations to perform for the failed block
+    private FailureActionResult cancelledLogOnly(final long blockNumber) {
```
jsync-swirlds: Probably should make this more generic.

```suggestion
    private FailureActionResult logOnlyNoAction(final long blockNumber) {
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -430,25 +437,88 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+    /// @return the operations to perform for the failed block
+    private FailureActionResult resendOnly(final long blockNumber, final FailureInfo failureInfo) {
+        LOGGER.log(INFO, "Verification of block {0} failed with {1}", blockNumber, failureInfo);
+        return new FailureActionResult(true, null);
+    }
+
+    /// Logs the cancelled verification session and resolves no operations.
+    ///
+    /// @param blockNumber the block number whose session was cancelled
+    /// @return the operations to perform for the failed block
```
jsync-swirlds:
```suggestion
    /// Logs the verification session but returns no action.
    /// This is typically used for cancelation or other failures that do
    /// not require further action.
    ///
    /// @param blockNumber the block number whose session ended with failure
    /// @return the action to perform for the failed block
```

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -1448,4 +1518,48 @@ public Boolean call() {
...
+            if (shouldResend && !failureInfo.isInformational()) {
+                // Schedule a resend for the block before sending the end of stream message
+                blocksToResend.add(blockNumber);
+            }
+            if (endStreamCode != null) {
+                for (final PublisherHandler handler : handlers.values()) {
+                    if (handler.handleFailedVerification(blockNumber, endStreamCode)) {
+                        // There will always be only one handler that will send the
+                        // end of stream message. Once we have, we can break.
+                        break;
```
jsync-swirlds: This is not always true, it's only true for a bad proof or other error that represents incorrect behavior from the publisher.
Some errors (particularly block-node sourced errors) should end all publisher streams.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -1424,11 +1424,12 @@ void testHandleVerificationCancelledIgnored() {
...
             }
 
-            // @todo(2977) this test might not be entirely true once we handle failure info
             /// This test aims to assert that the
             /// [LiveStreamPublisherManager#handleVerification(VerificationNotification)]
             /// does not schedule a resend for a block whose verification failed
-            /// with [FailureType#CANCELLED]. If a resend were scheduled, the
+            /// with [FailureType#CANCELLED]. A resend is also warranted for a
+            /// cancelled session, but it cannot be scheduled accurately until
+            /// @todo(3436) is in place. If a resend were scheduled, the
```
jsync-swirlds: I don't think cancelled verifications should result in resend.

The _possible_ exception (and we should probably add a different result code for this case) would be a verification cancelled _after_ the verification plugin received the end of block message.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
```diff
@@ -1539,6 +1548,291 @@ void testHandleVerificationNonCancelledFailureTypes(final FailureType failureTyp
...
+                toTest.handlePersisted(new PersistedNotification(block.number(), true, 0, BlockSource.PUBLISHER));
+                assertThat(toTest.getLatestBlockNumber())
+                        .as("latest known block must advance, no resend is pending for the block")
+                        .isEqualTo(block.number());
+            }
+
+            /// This test aims to assert that the
+            /// [LiveStreamPublisherManager#handleVerification(VerificationNotification)]
+            /// handles an informational failure of a malformed block or node
+            /// limitation type the same as a standard one, except that no
```
jsync-swirlds: For unknown error, we might choose to end the stream for all publishers, because something is broken in this block node...

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -430,25 +437,88 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+            // a retry may succeed, so the supplying stream stays open.
+            case MISSING_VERIFICATION_DATA, UNABLE_TO_PARSE -> resendOnly(blockNumber, failureInfo);
+            // The block is malformed, but the proof itself is not
+            // proven bad, so the stream ends with a generic error.
+            case MISSING_MANDATORY_FIELD, MISSING_MANDATORY_ITEM, UNSUPPORTED_STREAM_FORMAT ->
+                resendAndEndStream(blockNumber, failureInfo, INFO, Code.ERROR);
+            // Node side faults or capability limitations, the block is
+            // not proven bad, so the stream ends with a generic error.
+            case UNKNOWN_ERROR, UNRECOGNIZED_PROOF_TYPE, UNSUPPORTED_HAPI_VERSION, UNSUPPORTED_ITEM_TYPE ->
+                resendAndEndStream(blockNumber, failureInfo, INFO, Code.ERROR);
```
jsync-swirlds: Unknown error probably needs to result in all connected publishers receiving an end stream message.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -1448,4 +1518,48 @@ public Boolean call() {
...
+        /// the same block was already verified successfully within reasonable
+        /// recency, and ends the supplying handler's stream when an end of
+        /// stream code is present. The resend is scheduled before the end of
+        /// stream message is sent.
+        ///
+        /// @param blockNumber    the block number that failed verification
+        /// @param failureInfo    the failure information of the verification
+        /// @param blocksToResend the manager's set of blocks scheduled to be resent
+        /// @param handlers       the manager's currently connected publisher handlers
+        private void handle(
```
jsync-swirlds: We're cheating a bit here; this should be at least package-private, if not public.
Think how we'd have to declare these if this record was defined in its own java file.

jsync-swirlds (commented):
Good work, a few missing or unclear requirements.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: What you say is true, but not necessarily always.

It is possible for a block to be finished completely, but still be cancelled the verification session due to a full buffer. In those cases, from the perspective of the publisher, it has done its job and has not scheduled anything for resend because there was never a reason to do so. We need a way to detect if a block has actually finished or not (most probably a different failure type code, as I also see you have stated below), but this is something we should discuss later, as there are other items in #3436 that need attention.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: Good point, I treated it like "we do not have that data YET", but maybe I need to shift my perspective.

Note, however, that we've opted to also use this failure type on the frontend of the plugin. When we get a new block, we validate the start of the block. If it is invalid, we still send missing verification data. This is not BN's fault, however, and somewhat contradicts the above

image

Finally, you mentioned we should send responses to all publishers, something I do not do at all at the moment, I can add a value to the result object to do that. Do you envision any other places we want to close all publishers?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: These could be a network fluke or something else, but if we see any of these, we cannot verify the block. We have no idea if the block we received is valid, because we have not yet reached the point to do proof verification. Unsupported format is something we cannot really work with.

So yes, I do think you are right to say "there is no indication of malice", but we also cannot disprove it. I've chosen to go for closing streams, because most likely something is wrong on publisher side.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: If I look at your comment, where you state we can resolve this with a different error code (and I agree with that solution), then, cancel is actually the code where we want to schedule a resend, and the other code is where we do nothing.

Attribution does indeed not matter, I have to update the docs.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: Yes, we do not have an "end all" publishers handling for verification. Will address this.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
ata-nas: I think we should adda different result code (failure type) to differentiate, but in that case, cancel will be the code where we ask for a resend, and the new code will be where we do nothing. We want to be consistent all around, so let's leave CANCELLED to be the code where we treat it as "a session was cancelled after we have streamed a complete block, we need to schedule a resend".

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: Is that always true? Should we be so aggressive?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: This whole record and the handle method are a simple convenience. I view the whole thing as a "private method", just wrap data it needs for decision-making into a record-style format.

Thus, I did not view this as a separate component, rather I view it as a private convenience.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
```diff
@@ -146,6 +146,11 @@ private void processPublisherLiveItems(final BlockItems blockItems) {
         }
         // check if we have an active publisher session, if not, then disregard the items
         if (local != null) {
+            if (blockItems.isEndOfBlock()) {
+                // mark before offering so the session never observes the ending
+                // batch in its deque while still considered incomplete
+                local.markEndOfBlockReceived();
+            }
```
ivannov, line 153: Not sure whether there's a difference, but shouldn't both `markEndOgBlockReceived()` come after the block items are offered to the block items deque?

ata-nas: That is a good question, but I think we should first mark end of block received and then push the items.
This flag only affects what cancellation we will send to messaging in case this session gets cancelled. Regardless if we actually supply those items or not, if we've seen the end of this block, we want to mark that as complete ASAP.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: :white_check_mark: On another note, decided to tackle this now by adding a `CANCELLED_INCOMPLETE` failure type and support for it.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: :white_check_mark: added support for cancelled and incomplete sessions. This has been altered. Take a look.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
ata-nas: :white_check_mark: added another failure type and support for it.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
ata-nas: Is this always the case? Unknown errors could possibly happen for a single session, does that mean we close everybody?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: We have a failure we do not know how to handle.
We should encourage all publishers to connect to a BN that's not potentially failing.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I get that, but it's a lot more testable if we make the record and any internal methods it calls package-private.

That's merely a preference, however.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
jsync-swirlds: We have a failure we do not know how to handle.
We should encourage all publishers to connect to a BN that's not potentially failing.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: We do (or did) have code that sends an end of stream message to all active handlers.  We should be able to reuse that.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: We're mostly trying to be gentler with closing streams because it makes the CN go through a very expensive loop and rapid blacklist of the BN.
I think we'd prefer to avoid that cost for these specific cases.

If we see these repeatedly from the same source, that source will get an end of stream and potential penalty count, which I think is a reasonable balance.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: 1. There are a few cases where we should send responses to all publishers
   - Resend is always sent to everyone (possibly excepting a bad block source)
   - Any error that is a Block Node fault should go to all publishers
   - Any error indicating the BN is not currently able to accept _any_ block should go to all publishers.
1. When we fail to validate the _Start_ of the block, that isn't missing verification data; I think it's `MISSING_MANDATORY_ITEM` or perhaps something similar.
1. If we actually do not have verification data, there is no guarantee we'll ever have it; it's pretty safe to assume the failure is at least locally persistent.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
jsync-swirlds: It probably doesn't make a lot of difference in this case; it's an extreme corner case that the timing would make a substantive difference.

It's reasonable as it is.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: ✅ I see no value here, but these are now package private.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: ✅ Okay, I've added scopes of closure, closing all means we roll-back and reset, reusing the method we have in failed persistence.

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManagerTest.java
ata-nas: ✅ Okay, I've added scopes of closure, closing all means we roll-back and reset, reusing the method we have in failed persistence.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: ✅ We do have such roll-back logic, what I meant was that I do not support end all in verification, but I've added scopes of closure, closing all means we roll-back and reset, reusing the method we have in failed persistence. Please note that this changes some behavior of the publisher, but I think that is acceptable as long as it is acknowledged. My primary concern is that valid publishers, who stream valid blocks will all of a sudden be shut down because a single publisher supplied something that we cannot process.

Let's discuss this further, as this change in behavior is impactful and we need to be midnful of it and be sure this is what we actually want.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: ✅ Done.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: ✅ I agree with that, while we do not yet support penalties, we will have that in the future. This, however, means that we need #3436 as we need to track publisher activity with absolute certainty.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: ✅ Changed. For p.2, I do not remember why we went for missing verification data instead of missing mandatory item, but I remember we made that decision consciously. This is now migrate.

jsync-swirlds: Thanks, I don't recall the decision either, but I suspect one of us (likely me) was not thinking ahead far enough.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: I'm still not sold on the need to be absolutely certain of publisher identity (partly because I still believe it's only possible to be confused in a testing environment, not in a real-world network).
That aside, we already do have penalty enforcement for excessive call rate (exceeding publishing throttles), and can hook into that for other purposes as needed.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds: If any publisher somehow manages to make the BN encounter an unknown error (which should only happen due to bugs in the block node) or any other event indicating this block node is unstable or unable to accept any blocks, then I think we need to accept that well behaved publishers will be disconnected.
We're not punishing the publishers here, we're informing them that this Block Node is no longer reliable and they should consider streaming elsewhere for the stability of the network.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
jsync-swirlds, line 427: Minor note, we should have a test (might already exist) that verifies we do not respond to backfilled notifications here.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -418,37 +428,122 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
-                }
-            }
+            // A scheduled resend goes to all connected publishers, so we do
+            // not care which handler supplied the block that failed. Further
+            // improvements to the actions taken per failure type, like
+            // tailoring the response to the supplying handler, are captured
+            // by @todo(3436).
+            final FailureInfo failureInfo = notification.failureInfo();
+            final FailureActionResult actionResult = resolveFailureAction(blockNumber, failureInfo);
+            actionResult.handle(blockNumber, failureInfo, this);
```
jsync-swirlds, line 438: The result is created from failureInfo; why do we need to send it in again for the `handle` method?

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -418,37 +428,122 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+            // The complete block was received, but the session was cancelled
+            // before producing a result (e.g. evicted from the active sessions
+            // buffer). The publishers consider the block delivered, so nothing
+            // else will supply it and a resend must be scheduled. Resends go
+            // to all publishers, attribution is irrelevant.
+            case CANCELLED -> resendOnly(blockNumber, failureInfo);
+            // The block was never fully received, so it is already handled:
+            // either a resend was already scheduled, or another source is
+            // supplying it. No action is needed.
+            case CANCELLED_INCOMPLETE -> logOnlyNoAction(blockNumber, failureInfo);
```
jsync-swirlds, line 484: There is a lot of comment here that describe what the code is doing.
I think we can move that to the method Javadoc, and remove it here.
It is harming readability more than it's helping, in my opinion.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -418,37 +428,122 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+    ///
+    /// @param blockNumber the block number that failed verification
+    /// @param failureInfo the failure information of the verification
+    /// @param endStreamCode the end of stream code to send to the supplier
+    /// @return the operations to perform for the failed block
+    private FailureActionResult resendAndEndSupplierStream(
+            final long blockNumber, final FailureInfo failureInfo, final Code endStreamCode) {
+        LOGGER.log(
+                WARNING,
+                "Verification of block {0} failed with {1}, ending the supplying stream with {2}",
```
jsync-swirlds, line 499: If we introduce a variable for the message, this will be 2 lines instead of 6...

We have similar improvements in several other logging calls.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -418,37 +428,122 @@ public void handleVerification(@NonNull final VerificationNotification notificat
...
+    /// This is used for failures that require no further action from the
+    /// manager.
+    ///
+    /// @param blockNumber the block number whose session ended with failure
+    /// @param failureInfo the failure information of the verification
+    /// @return the operations to perform for the failed block
+    private FailureActionResult logOnlyNoAction(final long blockNumber, final FailureInfo failureInfo) {
+        LOGGER.log(INFO, "Session for block {0} failed with {1}, no action needed", blockNumber, failureInfo);
+        return new FailureActionResult(false, null, FailureActionResult.StreamScope.NO_STREAM_ACTION);
+    }
```
jsync-swirlds, line 545: All of these have a pattern; log something, and then return a failure action result.

Would it be cleaner, and more efficient, to push the log down to the `handle` method and remove the method overhead for these (i.e. create the failure action result directly in the switch above)?

Not blocking, but perhaps worth considering.

Note also, we use `action` everywhere except the javadoc, where we call it `operation`.  I think we should be consistent and use `action` in javadoc as well.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -1448,4 +1546,63 @@ public Boolean call() {
...
+
+    /// The operations the publisher manager must perform for a verification
+    /// failure, as resolved by [#resolveFailureAction(long,FailureInfo)].
+    ///
+    /// @param shouldResend true if a resend must be scheduled for the failed
+    /// block
+    /// @param endStreamCode the end of stream code to send when streams end,
+    /// null if the scope is [StreamScope#NO_STREAM_ACTION]
+    /// @param scope the scope of publisher streams to end
+    record FailureActionResult(boolean shouldResend, Code endStreamCode, StreamScope scope) {
```
jsync-swirlds, line 1558: Might be worth adding `failureInfo` and `blockNumber` as attributes here so they are not needed in `handle` below.

jsync-swirlds (approved):
Much improved.
A few additional suggestions, but none are blocking.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
```diff
@@ -161,6 +166,8 @@ private void processPublisherLiveItems(final BlockItems blockItems) {
     private void processBackfilledItems(final BlockItems blockItems) {
         final BlockVerificationSession session = startNewSession(blockItems, BlockSource.BACKFILL);
         activateSession(session);
+        // a backfilled block always arrives complete in a single batch
+        session.markEndOfBlockReceived();
```
Nana-EC, line 170: It's a narrow window but is there a concurrency concern that backfill on a different thread may pick up this session in between?
If so is it safer to move the `markEndOfBlockReceived()` to before `activateSession()`?

Nana-EC (approved):
LG, one minor consideration.
Non-blocking and can be followed up on if it indeed is possible

ata-nas:
@Nana-EC @jsync-swirlds thanks for your feedback! Since you both approve non blocking, I will proceed to merge this and have captured a follow up #3562 which will tackle your non-blocking suggestions.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
ata-nas: To be addressed in #3562.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: To be addressed in #3562.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: To be addressed in #3562.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: To be addressed in #3562.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: To be addressed in #3562.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: To be addressed in #3562.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: To be addressed in #3562.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
ata-nas: To be addressed in #3562.

Issue #3517, opened 2026-08-20
A Block Node with no publisher and `greedy: false` can never backfill itself


ivannov:
### Story Form

As a Block Node operator,
I want a Block Node that never receives a live stream to fill itself from its peers,
So that a redundant Block Node is actually redundant instead of sitting empty and silent for the lifetime of the deployment.

### Technical Notes

`determineEndCap` is called from the gap scan in `BackfillPlugin`, and its result is checked immediately:

```java
long endCap = determineEndCap(blockRanges);
if (endCap < 0 || startBound > endCap) {
    LOGGER.log(TRACE, "Nothing to backfill: startBound=[{0}] endCap=[{1}]", startBound, endCap);
    return;
}
```

The greedy branch sizes its window from `getPeerMaxAvailableBlock(storeMax)`, so it can look beyond an empty local store; the non-greedy branch has no peer query at all and can only cap at what is already stored. That is the asymmetry: `greedy: false` is a sensible "do not chase the live tail" setting for a Block Node that has a publisher, and a dead end for one that does not.

Nothing about the gap detection below the cap is broken - `BackfillFetcher.getAvailabilityForRange` resolves per-gap availability independently. The scan simply never gets that far, because the window is empty.

Config surface: `greedy` comes from the backfill plugin configuration and is set per Block Node in the topology file (`block_nodes.<id>.greedy`), so the topology change is one line per node. Peer wiring is already present in `fan-out-3cn-2bn.yaml` (`peers: [block-node-1]` / `[block-node-2]`), so block-node-2 has a source to pull from as soon as the scan window allows it.

Relevant paths:

- `block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java` - `determineEndCap` (line 430), the `endCap < 0` guard (line 360), `getPeerMaxAvailableBlock` (line 449)
- `tools-and-tests/scripts/solo-e2e-test/topologies/fan-out-3cn-2bn.yaml` - `greedy: false` on both Block Nodes, and the header comment describing fan-out

Verification of this task depends on the `relay_nodes` deploy blocker landing first, and on `fan-out-3cn-2bn` being in the matrix that runs (weekend and RC only today).

## Possible Solution

- Set `greedy: true` on the Block Nodes in `fan-out-3cn-2bn.yaml`: a Block Node with no publisher must be able to greedy-backfill, since the non-greedy scan window is derived from a store it cannot fill.
[...]

Issue #3516, opened 2026-08-20
Tag-triggered E2E runs start before the release artifacts are published


ivannov:
### Story Form

As a release manager,
I want a tag-triggered Solo E2E run to start only once the artifacts for that tag are published,
So that a red run on a release candidate means the release is broken, not that CI raced the publish job.

## Technical Notes

`solo-e2e-scheduler.yml` triggers on `push: tags: - "v*"`. On run 32322321036 (`v0.41.0-rc1`) the early matrix jobs failed because the artifacts for the tag did not exist yet:

```
INSTALLATION FAILED: ghcr.io/hiero-ledger/hiero-block-node/block-node-server:0.41.0-rc1: not found
[ERROR] Could not find artifact org.hiero.block-node:backfill:jar:0.41.0-rc1 in central
[ERROR] Could not find artifact org.hiero.block-node:backfill:jar:0.41.0-rc1 in sonatype-snapshots
```

Both the OCI image (used by the Solo Helm install) and the Maven artifacts (used by the plugin resolver) were affected. The `single` and `paired-3` jobs failed on the image at 01:58 and 02:03 UTC; the `3cn-1bn` job failed on the Maven artifacts at 02:24; jobs that started later in the same run got past that point, so the artifacts landed mid-run.

The failure is indistinguishable at a glance from a real deploy failure, which cost analysis time on top of the wasted runner time.

**Changes needed, either approach:**

1. Trigger on the release publish workflow completing (`workflow_run`) instead of on the tag push, so ordering is explicit; or
2. Keep the tag trigger and have the "Resolve component versions" step poll until the Block Node image and the Maven artifacts for the resolved version are retrievable, failing with a clear message after a bounded wait.

Option 1 is preferable if the publish workflow lives in this repo and reliably reports completion; option 2 is the smaller change and also covers manual dispatches that name a not-yet-published version.

Either way the error message when artifacts are genuinely missing should say so explicitly, rather than surfacing as a Helm install failure.

Issue #3515, opened 2026-08-20
`wrb-differential-test` is deployed with TSS verification but produces RSA/WRB blocks


ivannov:
### Story Form

As a test operator,
I want the `wrb-differential-test` topology deployed with the verification mode its own header comment describes,
So that the Mirror Node ingestion test exercises ingestion instead of timing out against a Block Node that rejected every block.

### Technical Notes

solo-deploy-network.sh` reads the marker in the same function that resolves the component sections:

```bash
VERIFICATION_MODE=$(yq '.verification_mode // "tss"' "${topology_file}" 2>/dev/null)
if [[ "${VERIFICATION_MODE}" == "rsa-wrb" ]]; then
  WRB_RSA="true"
  TSS_ENABLED="false"
fi
```

So the marker does two things at once: it selects WRB/RSA deployment values and it disables TSS. Absent the marker both stay at their TSS defaults, which is what happens today - the deploy summary line `Verification: tss` is the fastest confirmation of which path was taken.

Current state of the marker across topologies:

| Topology | `verification_mode` | Full-matrix result |
|---|---|---|
| `single-wrb-rsa` | `rsa-wrb` | Pass |
| `3cn-2bn-wrb-rsa` | `rsa-wrb` | Pass |
| `wrb-differential-test` | absent, defaults to `tss` | Fail, `verify_failed=84` |

`wrb-differential-test.yaml` is a single-Consensus-Node topology (`consensus_nodes: node1`) with one Block Node and an explicit plugin list that includes `block-verification`; nothing else in the file needs to change.

The failure counter surfaced by the `no-errors` assertion (`verify_failed`) is the metric to watch: it should be `0` after the fix, and the Block Node should report a real block range instead of the `UINT64_MAX` sentinel.

Only the topology file is expected to change. The two side observations below are Block Node behaviour rather than test configuration, and should be split out if they turn out to be real defects rather than consequences of the empty store.

## Possible Solution

- Add `verification_mode: rsa-wrb` to `topologies/wrb-differential-test.yaml`, matching what its own header comment already describes and what `single-wrb-rsa.yaml` does.
- Add a guard in `solo-deploy-network.sh`, or a check over `topologies/*.yaml`, so a single-Consensus-Node topology cannot silently be deployed with TSS verification.

ivannov:
## Description

[...]

Issue #3514, opened 2026-08-20
`block-available` accepts the `UINT64_MAX` no-blocks sentinel as a valid block range


ivannov:
### Story Form

As a test operator,
I want an assertion against a Block Node that holds no blocks to fail,
So that a run reporting "assertions passed" actually means blocks were ingested, and a broken run cannot report mostly-green.

### Technical Notes

Everything is in `tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh`:

- `assert_block_available_single` (around line 843) reads `firstAvailableBlock` / `lastAvailableBlock` from `BlockNodeService/serverStatus` via grpcurl and jq, then compares them with `-gt` / `-lt`. Its only "no blocks" check is `[[ "$first_block" == "null" ]]`, which the sentinel does not trigger because the field is present and populated.
- `assert_blocks_increasing_single` (around line 1112) obtains both baseline and current values from `get_block_count` and compares with `-gt`; sentinel-to-sentinel compares equal, so it reports "Blocks not increasing" rather than "no blocks", which is a misleading but at least red result.
- The dispatch cases for `block-available` and `blocks-increasing` are in `run_assertion` (around line 1332); no change is needed there.

`get_block_count` (line 1079) is the single funnel for `lastAvailableBlock` and already returns empty plus a non-zero exit for a missing or `null` value, so it is the natural place to reject the sentinel too - both `assert_blocks_increasing_single` call sites are covered at once. `assert_block_available_single` queries `serverStatus` itself and needs its own guard, since it also reads `firstAvailableBlock`.

The values reach bash as decimal strings from jq, so the overflow happens inside bash's `[[ ... -gt ... ]]` arithmetic evaluation, not in jq or grpcurl. Any comparison of a `uint64` field coming out of `serverStatus` has this hazard, so a shared guard helper is preferable to inline checks at each site.

The sentinel constant and the accompanying explanation already exist in `tools-and-tests/scripts/solo-e2e-test/scripts/wrb-distribution/assert-bn1-historical-backfill-landed.sh` (around line 55) and can be lifted into the shared library rather than redefined.

Fixture-based unit tests for assertions live under `tools-and-tests/scripts/solo-e2e-test/scripts/test/`, and the latency-aware assertion primitives they exercise are in `scripts/lib/chaos-assertions.sh` - the same place a shared sentinel guard would naturally sit.

## Possible Solution

[...]

PR #3523, opened 2026-08-21
feat(solo-e2e): backfill-during-live-stream test, CN local builds, and Solo E2E green-run fixes

AlfredoG87:
Latest Dry Run: https://github.com/hiero-ledger/hiero-block-node/actions/runs/32777895974

## Summary

Three related strands of Solo E2E work: the historical-backfill-while-live-streaming test from #3053, support for deploying a Consensus Node built from source (needed because no published CN tag yet carries the new block-root hashing), and four of the five sub-issues under the #3511 green-runs epic.

Validated end to end on a live `single` deployment: CN built from `hiero-consensus-node` `main` (`cec27bc6`), MN `v0.162.0-rc1`, BN `main`. 513 blocks ingested with **zero** verification failures, and the WRAPS signature transition test passed (Schnorr -> WRAPS at block 428).

## Issues

| Issue | Status |
|---|---|
| #3053 — Test and assertion for historical backfill while receiving live blocks | Closes |
| #3513 — `relay_nodes:` with only commented-out entries silently enables the relay | Closes |
| #3514 — `block-available` accepts the `UINT64_MAX` no-blocks sentinel | Closes |
| #3515 — `wrb-differential-test` deployed with TSS but produces RSA/WRB blocks | Closes |
| #3517 — Block Node with no publisher and `greedy: false` can never backfill | Closes (test-side) |
| #3511 — EPIC: Solo E2E Suite: Restore Green Runs | Advances, 4 of 5 sub-issues |
| #3516 — Tag-triggered E2E runs start before release artifacts are published | **Not addressed** — the only sub-issue needing `.github/workflows` changes |

## Backfill during live stream (#3053)

The hinge is `earliestManagedBlock`. On a store-less start the publisher treats it as its next expected block, so at the default `0` the Block Node answers every offer from the Consensus Node with `BlockNodeBehind` and live streaming only resumes once backfill has walked the entire chain. Raised above the chain head, the first offered block is accepted and everything below it becomes a `HISTORICAL` backfill gap — which is the condition the test needs.

New scripts under `scripts/backfill/`:

| Script | Purpose |
|---|---|
| `wait-for-network-height.sh` | Blocks until the network reaches `MIN_HEIGHT` so there is a recoverable gap |
| `configure-bn-live-tail-backfill.sh` | Raises `earliestManagedBlock` to `<chain head> + EMB_OFFSET` and throttles backfill while the BN is scaled to zero |
| `assert-backfill-during-live-stream.sh` | Samples `serverStatusDetail` plus `publisher_open_connections` and requires the historical range and the live range to both advance in the same window |

`clear-block-storage` now also removes `application-state/block-ranges.json`. That file sits on its own PVC and is reloaded into the in-memory range set at startup, which the backfill plugin reads as "blocks this node already has" — leaving it behind made a wiped node believe it still held the deleted blocks, so backfill found no gap. Only that file is removed; the sibling TSS and RSA bootstrap rosters in the same directory must survive, or every block fails `BAD_BLOCK_PROOF`.

## Consensus Node from source

Solo does not pull the Consensus Node as a container image. `solo consensus node setup` downloads a platform build zip from `builds.hedera.com/node/software/<vMAJOR.MINOR>/build-<tag>.zip` and unpacks it into the pod's `root-container`; the pod image itself comes from the `solo-deployment` chart and is not selected by `--release-tag`. Only tagged releases are published there, so `CN_VERSION=main` — which resolves to `0.79.0-SNAPSHOT` from the CN branch's `version.txt` — 404s at setup time, after the cluster and Block Nodes are already deployed.

`--local-build-path` short-circuits that download and uploads the local jars instead. This PR adds `--cn-local-build-path` to `solo-deploy-network.sh`, plumbs `CN_LOCAL_BUILD_PATH` through the Taskfile, and documents it in `.env.example`:

```bash
cd <cn-repo> && ./gradlew assemble
task up CN_LOCAL_BUILD_PATH=<cn-repo>/hedera-node/data
```

Both failure modes now fail before any cluster work: a `-SNAPSHOT` version with no build path, and a build path that has not been built. Previously the first surfaced as a 404 twenty minutes in.

Named CI tags such as `sdpt-pass-00380` are not a workaround — no build zip is published for them, and Solo's `Templates.prepareReleasePrefix` rejects any tag that is not dotted semver.

### `CN_VERSION` now defaults to `main`

> [!IMPORTANT]
> **This is a breaking change for local development.** After this merges, `task up` fails immediately unless `CN_LOCAL_BUILD_PATH` points at a built Consensus Node. Build one (`cd <cn-repo> && ./gradlew assemble`) or pin a released tag (`CN_VERSION=v0.79.0-alpha.1`) if you don't need the new hashing.

No published CN tag yet carries the fixed 16-slot block-root hashing rework, so a released tag cannot verify against a current Block Node and Mirror Node — `main` is the only version that actually works today. The default changes in `Taskfile.yml` and `.env.example` for local runs, and in `solo-e2e-scheduler.yml` for CI (see the next section, which also adds the CN build CI needs to make that default usable).

`resolve-versions.sh`'s internal `${1:-latest}` fallback stays as-is, since CI passes the version positionally and it is only reached when nothing was supplied at all.

## CI builds the Consensus Node when the version is a `-SNAPSHOT`

`solo-e2e-test.yml` now detects a `-SNAPSHOT` resolved CN version and, only in that case, checks out `hiero-ledger/hiero-consensus-node` at `main` (`fetch-depth: 1`), runs `./gradlew assemble --no-daemon`, verifies `hedera-node/data/{apps,lib}` exist, and passes the directory as `--cn-local-build-path`. Released tags are untouched and still fetch the published zip, so only SNAPSHOT runs pay the build cost.

`solo-e2e-scheduler.yml` gets two fixes:

- The final `else` branch hardcoded all three component versions, so a manual dispatch that set only `consensus-node-version` had it **silently discarded** — this is why run 32461677811 deployed CN `latest` despite being dispatched for `main`. It now honours any explicitly supplied version and falls back to `main` for CN and BN.
- `MN_VERSION` no longer defaults to `main`. Mirror Node publishes no `main` snapshots, so that default could never resolve; it is pinned to `v0.162.0-rc1`, the first release carrying the reworked block-root hashing that a `main` Consensus Node produces.

> [!NOTE]
> Scheduled daily and weekend runs now deploy CN `main` rather than `latest`, so they build the Consensus Node and take correspondingly longer. That is the point — with a released CN the Block Node rejects every block, so those runs carried no signal. It does mean the suite no longer exercises "BN `main` against a GA-stable Consensus Node"; worth adding back as a separate matrix entry once a released tag carries the hashing rework.

## Topology component detection (#3513)

`yq '.<section> | keys | length // 0'` errors with `Cannot get keys of !!null` whenever a section is null — a `relay_nodes:` key whose only child is commented out, or a missing key. With stderr discarded and no fallback the count came back as the empty string, so both branches of the comparison were false, control fell through to the legacy `components.<key> // true` default, and the component deployed anyway.

Six of the nine matrix jobs hit this for both `relay_nodes` and `explorer_nodes`. The blast radius is larger than a stray pod: `solo relay node add` blocks on relay readiness, and the relay's `GET :7546/health/readiness` probe has been failing on recent images, so the wedged relay aborted `task up` **before** `task port-forward` — presenting as "port forwards are broken" rather than "relay is broken".

The three duplicated blocks are now one `skip_component` helper using `yq '.<section> | length'`, which returns `0` for both `null` and `{}` without erroring. The legacy `components.<key>` flag must now say `true` explicitly.

## No-blocks sentinel (#3514)

An empty Block Node reports `firstAvailableBlock` and `lastAvailableBlock` as `UINT64_MAX`, not null and not 0 — 0 means "block 0 is available". That value wraps to -1 inside bash's `[[ ... -gt ... ]]`, so an assertion against a node holding nothing passed.

`is_valid_block_number` in `scripts/lib/chaos-assertions.sh` is now the single guard, rejecting the whole out-of-signed-64-bit-range class rather than one sentinel value, so a future sentinel change cannot silently reintroduce the overflow. Four call sites:

- `assert_block_available_single` — previously an exact-equality check only
- `get_block_count` — previously unguarded; since sentinel compares equal to sentinel, `blocks-increasing` reported "Blocks not increasing" for an empty node, a plausible-looking failure that hid the real one
- `solo-network-status.sh` — previously printed the sentinel as a range, reading as a healthy node holding 18 quintillion blocks; now "No blocks"
- `wrb-distribution/assert-bn1-historical-backfill-landed.sh` — previously redefined the constant; now sources the shared one

Expect this to turn currently-passing assertions red wherever a Block Node is empty. That is the correction, not a regression.

## Verification mode and backfill reachability (#3515, #3517)

`wrb-differential-test.yaml` gets the `verification_mode: rsa-wrb` marker its own header already describes. Without it the deploy defaulted to `tss` and the Block Node rejected every block (`verify_failed=84`), so the Mirror Node ingestion test timed out against an empty store instead of testing ingestion. The WRB overlay is passed to Helm after the per-BN overlay and its `plugins.names` is a strict superset of what the topology listed, so nothing is lost.

`fan-out-3cn-2bn.yaml` gets `greedy: true` on both Block Nodes. A Consensus Node's `block_nodes` entry is a priority list, not a broadcast list — each CN streams to its first entry and falls back only on unreachability — so in steady state all three CNs publish to BN1 and **BN2 has no publisher**. The non-greedy `determineEndCap` branch sizes its scan window from the local store and never queries peers, so on an empty store the window is empty and BN2 stays empty for the lifetime of the deployment. The header comment, which claimed "all CNs to both BNs", is corrected to describe priority failover.

## Tests

Two fixture-based suites, both runnable without a cluster:

- `scripts/test/test-chaos-assertions.sh` — 14 new cases for `is_valid_block_number` covering the sentinel, `INT64_MAX` boundaries, and malformed input, plus one that asserts the sentinel still wraps negative in bash so the guard's rationale cannot silently expire. **35 passed, 0 failed.**
- `scripts/test/test-topology-decisions.sh` (new) — every topology times every section, the four null-section shapes that caused #3513, the legacy `components` opt-in, and `verification_mode` validity. It extracts the real `skip_component` from the deploy script rather than copying it, so it tracks the implementation. **69 passed, 0 failed.**

No Java changed, so no module tests or `spotlessApply` were required.

## Deliberately not implemented

#3515 and #3517 both suggest a guard so that "a single-Consensus-Node topology cannot silently be deployed with TSS verification". **That guard would break `single.yaml`**, which is a 1-CN topology that deploys with TSS and verified 105+ TSS proofs with zero failures during this PR's validation run. The premise traces to `wrb-differential-test.yaml`'s own header claim that Solo's single-node setup cannot do TSS, which is stale. A name-based "wrb implies rsa-wrb" variant would also false-positive on the five `wrb-distribution-step*` topologies, which deploy with no Block Node at all.

Implemented instead: `verification_mode` must resolve to `tss` or `rsa-wrb`, so a typo like `rsa_wrb` fails loudly rather than silently defaulting to TSS.

## Follow-ups

- #3516 remains open — it needs `.github/workflows` changes, either a `workflow_run` trigger or an artifact-availability poll in version resolution.
- `solo-e2e-scheduler.yml` (around lines 212-215) defaults **both** `CN_VERSION` and `MN_VERSION` to `main` on the manual-dispatch-with-explicit-BN branch. That branch is broken for two independent reasons: CN `main` has no published artifact (it now fails fast rather than 404ing), and Mirror Node publishes no `main` snapshots at all. Only reachable via manual dispatch. Left alone here as a CI change; worth folding into #3516.
- The `block-ranges.json` fix in `clear-block-storage` is adjacent to #3517's theme but a distinct defect with no issue of its own; worth filing.
- `roster_entries_loaded` reports 0 on a TSS deployment while TSS proofs verify normally, so that metric appears to be RSA-path only. Harmless here, but misleading if used as a health signal.

### DIFF
--- a/.github/workflows/solo-e2e-scheduler.yml
+++ b/.github/workflows/solo-e2e-scheduler.yml
@@ -211,12 +211,21 @@ jobs:
             CN_VERSION="${{ inputs.consensus-node-version }}"
             CN_VERSION="${CN_VERSION:-main}"
             MN_VERSION="${{ inputs.mirror-node-version }}"
-            MN_VERSION="${MN_VERSION:-main}"
+            # Mirror Node publishes no 'main' snapshots, so this cannot default to main.
+            # Pinned to the first release carrying the reworked block-root hashing, which is
+            # what a 'main' Consensus Node produces.
+            MN_VERSION="${MN_VERSION:-v0.162.0-rc1}"
           else
-            # Scheduled runs (daily/weekend): use SNAPSHOT BN against GA stable CN/MN
-            BN_VERSION="main"
-            CN_VERSION="latest"
-            MN_VERSION="latest"
+            # Scheduled runs (daily/weekend) and manual dispatches that named no Block Node
+            # version. Honour any explicitly supplied version — this branch used to hardcode
+            # all three, so a manual dispatch setting only consensus-node-version had it
+            # silently discarded.
+            BN_VERSION="${{ inputs.block-node-version }}"
+            BN_VERSION="${BN_VERSION:-main}"
+            CN_VERSION="${{ inputs.consensus-node-version }}"
+            CN_VERSION="${CN_VERSION:-main}"
+            MN_VERSION="${{ inputs.mirror-node-version }}"
+            MN_VERSION="${MN_VERSION:-v0.162.0-rc1}"
           fi
 
           # Solo CLI version: use input if provided, otherwise default to latest stable
--- a/.github/workflows/solo-e2e-scheduler.yml
+++ b/.github/workflows/solo-e2e-scheduler.yml
@@ -348,7 +348,11 @@ jobs:
     needs: configure
     strategy:
       fail-fast: false
-      max-parallel: 1 # Run sequentially to avoid resource contention
+      # Each matrix entry runs on its own GitHub runner with its own Kind cluster, so
+      # parallel deployments do not contend for resources with each other. 3 caps how
+      # many runners (and, since a -SNAPSHOT Consensus Node is built per job, how many
+      # concurrent CN builds) a single scheduler run consumes.
+      max-parallel: 3
       matrix:
         include: ${{ fromJson(needs.configure.outputs.matrix) }}
     uses: ./.github/workflows/solo-e2e-test.yml
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -317,6 +317,43 @@ jobs:
           # the resolved mn_version below.
           MN_RESOLVED=$(echo "$OUTPUT" | grep "^mn_version=" | cut -d= -f2)
           echo "MN_VERSION=${MN_RESOLVED}" >> "${GITHUB_ENV}"
+          # Solo installs the Consensus Node from a build zip on builds.hedera.com and only
+          # tagged releases are published there. A -SNAPSHOT (what 'main' resolves to, from
+          # version.txt on the CN branch) has no artifact, so it must be built from source and
+          # handed to Solo via --local-build-path. Flag that for the steps below.
+          CN_RESOLVED=$(echo "$OUTPUT" | grep "^cn_version=" | cut -d= -f2)
+          if [[ "${CN_RESOLVED}" == *-SNAPSHOT ]]; then
+            echo "cn_needs_local_build=true" >> "${GITHUB_OUTPUT}"
+          else
+            echo "cn_needs_local_build=false" >> "${GITHUB_OUTPUT}"
+          fi
+
+      # Only for a -SNAPSHOT Consensus Node: build it from source, since no artifact exists.
+      - name: Checkout Consensus Node source
+        if: steps.versions.outputs.cn_needs_local_build == 'true'
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: hiero-ledger/hiero-consensus-node
+          ref: main
+          path: hiero-consensus-node
+          fetch-depth: 1
+
+      - name: Build Consensus Node from source
+        id: cn-build
+        if: steps.versions.outputs.cn_needs_local_build == 'true'
+        working-directory: hiero-consensus-node
+        run: |
+          echo "Building Consensus Node $(cat version.txt) at $(git rev-parse --short HEAD)"
+          ./gradlew assemble --no-daemon
+          # Solo wants the data directory itself and validates it holds apps/ and lib/.
+          CN_DATA_DIR="${PWD}/hedera-node/data"
+          for subdirectory in apps lib; do
+            if [[ ! -d "${CN_DATA_DIR}/${subdirectory}" ]]; then
+              echo "ERROR: ${CN_DATA_DIR}/${subdirectory} missing after assemble" >&2
+              exit 1
+            fi
+          done
+          echo "cn_local_build_path=${CN_DATA_DIR}" >> "${GITHUB_OUTPUT}"
 
       # Install Solo CLI (npm by default, or build an approved fork/branch when solo-source=git)
       - name: Install Solo CLI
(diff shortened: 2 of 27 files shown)


Nana-EC (dismissed):
Unblocking based on latest observed run.
Will circle back also

Nana-EC (approved):

jsync-swirlds (approved):

PR #3524, opened 2026-08-21
chore(metrics): Add new metrics for latency to dashboards

### LINKED ISSUE #1567: feat(metrics): Add new metrics for latency to dashboards
### Story Form

As a Block Node Operator
I want to see BN Latency related Metrics
So that I know how much time BN is adding latency to the MN to receive the Blockstream, and also, how long is taking to ACK a block back to CN.

### Technical Notes

Due to the the multiple nature of connections, many publishers and many consumers. and the inability to use metric labels on current metrics platforrm sdk, for now it would be good to only have the avg. latency of the fastest of them.

So we only observe and record latency time taken to ACK from the current or faster publisher.
And we only observer and record latency time added to the fastest LIVE consumer. measured from the moment we receive a batch and we re-send a batch the first LIVE consumer.

(the added BN Latency will be hard to measure since it involves a timer that starts on the publisherManager and is stoped and observed on the publisher plugin. but we might consider a SharedMetrics Pojo inside the BN Context that all the plugins have access to.

ivannov:
## Reviewer Notes
  
* Adds a per-subscriber metric for how long a live batch waits before being sent
* Recycles a fixed set of metric label slots, so series count follows peak concurrent subscribers

## Related Issue(s)

Closes #1567

### DIFF
--- a/block-node/app/docker/metrics/dashboards/performance-metrics.json
+++ b/block-node/app/docker/metrics/dashboards/performance-metrics.json
@@ -1616,6 +1616,178 @@
       },
       "fieldConfig": {
         "defaults": {
+          "decimals": 2,
+          "noValue": "No live traffic",
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "blue",
+                "value": null
+              },
+              {
+                "color": "green",
+                "value": 20000000
+              },
+              {
+                "color": "yellow",
+                "value": 50000000
+              },
+              {
+                "color": "red",
+                "value": 60000000
+              }
+            ]
+          },
+          "unit": "ns"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 6,
+        "w": 5,
+        "x": 0,
+        "y": 43
+      },
+      "id": 42,
+      "options": {
+        "colorMode": "value",
+        "graphMode": "area",
+        "justifyMode": "auto",
+        "orientation": "auto",
+        "percentChangeColorMode": "standard",
+        "reduceOptions": {
+          "calcs": [
+            "lastNotNull"
+          ],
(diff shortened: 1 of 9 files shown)


jsync-swirlds:
This change adds new allocations and overhead for _every message to every subscriber_ just to track a low value latency metric (how long does it take to _send_ data, which is mostly internal buffer copies).
This is time and resource costs for no real value.

We really need to review these old (over a year) issues before starting work, because most (including possibly this one) ask for things we either no longer need or should not implement.

jsync-swirlds (changes requested):
Dashboard adds are mostly OK, but subscriber metric add is not a good change, in my opinion.

ivannov:
> Dashboard adds are mostly OK, but subscriber metric add is not a good change, in my opinion.

So my current assumption was correct.  Every file in this PR exists to serve `subscriber_live_send_latency_ns`:

- Both `performance-metrics.json` copies: the only real additions are two panels querying `blocknode_subscriber_live_send_latency_ns_total` / `blocknode_subscriber_live_batches_sent_total`
- `docs/block-node/metrics.md`: two table rows, the explanation of the slot labelling, and one alert row, all describing the new metric
- `BlockStreamSubscriberSession`: the timestamp capture and counter increments
- `SubscriberServicePlugin`: the 64-slot label allocator plus overflow slot, which exists purely to label the metric
- Both test classes and the `getMetricValuesByLabel` test-fixture helper: only exercised by the new metric's tests

So maybe we should close the PR and the issue (as won't do)?

jsync-swirlds:
If the only change is adding the `live_send_latency_ns` metric in both code and dashboards, then I agree that we should just close this PR and the issue behind it.

ivannov:
As agreed with @jsync-swirlds , this metrics is not very useful, so closing the PR

PR #3525, opened 2026-08-21
fix: Await App State Update Before Assert

### LINKED ISSUE #3536: Await App State Update Before Assert
- After completing #3401, we found out that the test flaked for another simple reason
  - App State needs some time to update the available ranges, but none is given
- We need to give time to app state to actually make the update before we poll server status

ata-nas:
* Flaky test fix, need to await for app state update to happen to get accurate status detail

Fixes #3536

### DIFF
--- a/tools-and-tests/suites/src/main/java/org/hiero/block/suites/backfill/BackfillMultiportSourceTests.java
+++ b/tools-and-tests/suites/src/main/java/org/hiero/block/suites/backfill/BackfillMultiportSourceTests.java
@@ -25,6 +25,7 @@ import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.LockSupport;
 import org.hiero.block.api.BlockEnd;
 import org.hiero.block.api.BlockItemSet;
 import org.hiero.block.api.BlockStreamPublishServiceInterface.BlockStreamPublishServiceClient;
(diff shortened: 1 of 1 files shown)


ata-nas:
Another way to fix this is to keep pulling status detail until we see the values instead of a brief sleep. Test flakes because context update does not happen before we pull status detail and assert.

tools-and-tests/suites/src/main/java/org/hiero/block/suites/backfill/BackfillMultiportSourceTests.java
```diff
@@ -128,7 +129,7 @@ void backfillClientFetchesFromSplitPortSource() throws Exception {
         awaitRunning();
 
         publishBlocks(defaultPort);
-
+        LockSupport.parkNanos(TimeUnit.MILLISECONDS.toNanos(1_500));
```
jsync-swirlds: This is a long delay, can we do this without a fixed sleep; perhaps by waiting for some triggering event?

ata-nas: We can, I even posted a comment below. We can keep repeating the status detail call until we get a non empty response, limited by timing, some reasonable amount.

ata-nas: P.S. the delay is big because context updated are scheduled every second I believe

jsync-swirlds: If we can avoid a fixed delay, that is preferable.

ata-nas: ✅ I've moved to a retry solution.

ivannov (approved):

berryware (approved):
Looks good

Issue #3528, opened 2026-08-22
Spike: validate foundational assumptions for the API throttling design


Nana-EC:
## Summary

Before implementation begins, validate the assumptions the API throttling / admission-control design depends on. This is time-boxed investigation, not implementation work — findings should be captured as comments/updates on this issue and folded into the design doc before implementation issues are filed.

## Items to validate

- [x] **Streaming call lifecycle.** Confirm exactly which lifecycle signal(s) reliably indicate that a gRPC call has completed, for both unary calls and long-lived streaming calls, so a per-call resource (e.g. a concurrency permit) can be released exactly once regardless of how the call ends (normal completion, client cancellation, or timeout). This is foundational: the design's concurrency limiting depends on correctly detecting completion for every RPC shape, and unary and streaming calls do not necessarily expose that signal the same way.
- [ ] **Default rate/concurrency values.** The design will propose starting values for per-API rate and concurrency limits. Validate these against realistic traffic patterns and load tests before they ship as defaults.
- [ ] **Backend read-concurrency limit sizing.** Determine an appropriate size for the shared backend read-concurrency safeguard, informed by the actual storage characteristics of target deployments (local disk vs. network-attached storage behave differently under concurrent reads).
- [ ] **Client identification.** Confirm whether any deployments run a reverse proxy or load balancer in front of the Block Node that would affect client-network-address-based identification, and whether that needs to be accounted for in the design.
- [ ] **Existing infrastructure overlap.** Confirm whether any existing infrastructure (service mesh, API gateway, or load balancer) already provides connection-rate or bandwidth limiting for Block Node deployments, to avoid duplicating protection that already exists elsewhere.
- [x] **Not-yet-implemented services.** Identify any gRPC services defined in the protocol but not yet implemented by a plugin, and note that their throttling classification should be revisited once they're built — their real resource cost is unknown today.

## Acceptance criteria

- [ ] Each item above has a documented finding, captured on this issue or linked from it.
- [ ] Findings are incorporated into the design doc before implementation begins.

## Related

Part of #3526.

Nana-EC:
Partial progress on the "streaming call lifecycle" item, from implementing Phase 1 (#3530) as a spike.

[...]

Issue #3526, opened 2026-08-22
Epic: API throttling and admission control for gRPC services


Nana-EC:
## Goal

Add a throttling / admission-control layer in front of the Block Node's gRPC APIs so that per-client rate and concurrency limits can be enforced per API, tuned to how expensive that API's calls actually are, alongside a shared safeguard that protects the block-storage read path from being overwhelmed by concurrent reads regardless of which client or API triggered them.

## Intention

The Block Node exposes several gRPC APIs with very different resource profiles: `publishBlockStream` (block ingest from consensus nodes), `subscribeBlockStream` (long-lived streaming reads), `getBlock` (on-demand reads — cheap for recent blocks, comparatively expensive for archived/historical ones), and lightweight status APIs (`serverStatus`, `serverStatusDetail`). None of these APIs currently have any rate limiting, per-client concurrency limiting, or admission control.

In practice this means a burst of read traffic — many concurrent subscribers, or a spike of `getBlock` requests against historical blocks — has no mechanism preventing it from consuming a disproportionate share of node resources (CPU, memory, disk I/O, and connection capacity). In the worst case, that contention could compete with the capacity the block-ingest path from consensus nodes depends on, even though ingest and read traffic are logically unrelated.

This epic tracks adding admission control that:
- Applies per-client rate and concurrency limits, tuned per API.
- Distinguishes cost *within* a single API where it matters (e.g. `getBlock`/`subscribeBlockStream` requests for live/recent blocks are cheaper than requests for historical/archived blocks, and should be limited differently).
- Protects the shared backend block-storage read path from concurrent-read overload, independent of per-client limits.
- Fits naturally on top of the project's existing gRPC service layer without hard-coupling to a specific web server implementation.
- Favors a small number of simple, well-understood, proven mechanisms (rate limiting + concurrency limiting) over a complex adaptive system, so the team can implement and operate it with confidence.

## Considerations

- **`publishBlockStream` (consensus-node ingest) is out of scope for throttling itself.** Protecting it from resource contention caused by other APIs is the motivation for this work, not a target for restriction. Misbehaving-publisher protection is a separate concern, tracked independently.
- **Initial client identification will be by network address**, not authenticated identity — no authentication mechanism exists for these read APIs today, and building one is a separate, larger effort. The design should keep this swappable so authenticated client identity can replace it later without a rework.
[...]

PR #3529, opened 2026-08-22
docs: API throttling and admission control design doc

### LINKED ISSUE #3527: Design doc: API throttling and admission control for gRPC services
## Summary

Produce a design document describing the throttling / admission-control approach for the Block Node's gRPC APIs.

The document should cover:

- Per-client rate and concurrency limiting, configurable per API and, where it matters, per request-cost tier (e.g. live vs. historical block reads for `getBlock` and `subscribeBlockStream`).
- A shared mechanism that protects the block-storage read path from concurrent-read overload, independent of and complementary to the per-client limits above.
- How the mechanism integrates with the project's existing gRPC service layer, without coupling the design to a specific web server implementation.
- The configuration surface — how an operator tunes limits per API without a code change, and what a new plugin/API needs to do to opt in.
- Metrics and observability for the new admission-control layer.
- Alternatives considered, and why they were or weren't chosen.

## Acceptance criteria

- [ ] Design doc added at `docs/design/apis/api-throttling.md`, following the structure in `docs/design/design-doc-template.md`.
- [ ] Doc reviewed and approved.
- [ ] Any open design questions that need empirical validation are captured and handed off to the companion spike issue.

## Related

Part of #3526.

Nana-EC:
## Summary
- Adds a design doc proposing a per-client rate/concurrency admission gate at the gRPC service boundary (`getBlock`, `subscribeBlockStream`, `serverStatus`/`serverStatusDetail`), weighted by request cost (live vs. historical block reads), plus a shared backend bulkhead protecting block storage reads.
- `publishBlockStream` is explicitly out of scope for throttling — the design exists to protect the ingest path from read-side resource contention, not to further constrain it.
- No code changes; this is the design-doc deliverable for the throttling epic. Foundational assumptions (e.g. correct gRPC streaming-call lifecycle handling for concurrency-permit release, default rate/concurrency values, backend bulkhead sizing) are tracked separately in the companion spike, #3528.

Closes #3527.
Part of #3526.

## Test plan
- [ ] `markdownlint -c .markdownlint.yaml docs/design/apis/api-throttling.md` passes (verified locally)
- [ ] Design reviewed by the team per the acceptance criteria in #3527


docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,421 @@
...
+   protect. See [Performance](#performance) under Design.
+
+**Non-goals for this iteration:**
+
+- **Throttling `publishBlockStream`.** This work exists to protect the ingest path from resource contention caused
+  by other APIs, not to further constrain it. Misbehaving-publisher protection (e.g. penalizing a publisher that
+  repeatedly sends invalid data) is a separate concern.
+- **Authenticated client identity.** No authentication mechanism exists for the read APIs today. Clients are
+  identified by network address for now; the identification mechanism is designed to be replaceable without
+  reworking the rest of the system.
```
jsync-swirlds: Perhaps worth noting that this is very unreliable.  It's entirely possible that an ingest element in Kubernetes or load balancer could cause all connections to _appear_ to come from a single IP, or localhost.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,421 @@
...
+<dt>Client key</dt><dd>An identifier used to group a caller's requests for the purpose of per-client limits. Derived
+from the caller's network address in this design.</dd>
+<dt>Weight class</dt><dd>A cost tier assigned to a request (e.g. LIGHT, MODERATE, HEAVY) that determines which
+rate/concurrency policy applies to it.</dd>
+<dt>Content-aware weigher</dt><dd>A per-API function that inspects a request's content (e.g. the requested block
+number) to classify it into a weight class before admission, rather than relying on a single static weight for the
+whole API.</dd>
+<dt>GCRA (Generic Cell Rate Algorithm)</dt><dd>A leaky-bucket-equivalent rate-limiting algorithm that tracks a single
+"theoretical arrival time" value per limited key, admitting a request only if it isn't arriving too far ahead of the
+evenly-paced schedule that value implies.</dd>
```
jsync-swirlds: Rather than describing in terms of GCRA, which incorporates a lot of ATM specific terminology and concepts, perhaps remove that term entirely and describe in terms of "Leaky Bucket As Meter", which is a closer match to the actual intent here.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,421 @@
...
+
+### `ThrottlePolicy`
+
+The runtime policy applied to one gRPC method for one client: a rate (requests per second), a burst tolerance, a
+per-client concurrency ceiling, and a node-wide concurrency ceiling.
+
+### `GcraLimiter`
+
+A lock-free rate limiter implementing GCRA, keyed per client. Holds one monotonic-clock timestamp (the theoretical
+arrival time) per key, advanced via compare-and-swap on each admitted request.
```
jsync-swirlds: timestamps are a specific element required by ATM (because it's measuring network arrival times).
We should have a simple counter, not a timestamp per key.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,421 @@
...
+
+### `ClientKeyExtractor`
+
+Derives a client key from an incoming call's connection metadata. The default implementation uses the caller's
+remote network address. The interface is designed so an authenticated-identity-based implementation (e.g. from a
+client certificate) can replace it later without changing anything else in the system.
```
jsync-swirlds: Perhaps note both mTLS and OAuth concepts, so we don't accidentally tie too tightly to only one possible future authentication approach.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,421 @@
...
+### `ContentAwareWeigher`
+
+An optional, per-API function that classifies a request into a weight class based on its content, evaluated once at
+admission time before any rate/concurrency check runs. Used by `getBlock` and `subscribeBlockStream` to distinguish
+live/recent requests from historical ones.
+
+### The admission decorator
+
+A wrapper around a plugin's `ServiceInterface` implementation that performs the admission decision on every call
+and, if admitted, tracks the call's concurrency permit for its full lifetime.
```
jsync-swirlds, line 125: Should we implement this more as a [ServerInterceptor](https://grpc.github.io/grpc-java/javadoc/io/grpc/ServerInterceptor.html), supported by Helidon's [GrpcRouting Builder](https://helidon.io/docs/v4/apidocs/io.helidon.webserver.grpc/io/helidon/webserver/grpc/GrpcRouting.Builder.html#intercept(int,io.grpc.ServerInterceptor...))?

We would need to expose the interceptor element in the PBJ builder, but it might be a more efficient alternative, and moves the configuration clearly to plugin API registration during `start`.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,421 @@
...
+A method's node-wide concurrency ceiling is different: it represents an allocation of one shared, node-wide capacity
+budget (connections, heap, disk I/O) across every API on the node, which is inherently a node-level view rather than
+something any single plugin can reason about on its own. Node-wide ceilings therefore live in one small, shared,
+node-level configuration record.
+
+Enforcement stays centralized in the one place that already sees every plugin's service registration: a plugin
+hands its own per-client policy in at registration time, and the registration point merges it with the
+corresponding node-wide ceiling before applying the decorator described above. This keeps per-client tuning fully
+owned by the plugin that the limit governs, while keeping the actual enforcement logic — and the ability to validate
+that per-plugin and node-wide numbers agree with each other — in one place.
```
jsync-swirlds, line 256: This is good, but I think we can move the entire implementation to the registration step by making it an interceptor, and avoid needing to modify how plugins interact with the pipelines.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,421 @@
...
+|---------------------------|--------------------------------------------------------|-----------------------------------------------------------------------------|
+| Admitted calls            | Counter, labeled by service/method/weight class        | Calls that passed all admission checks                                      |
+| Rejected calls            | Counter, labeled by service/method/weight class/reason | Calls rejected, broken down by which check rejected them                    |
+| Per-client in-flight      | Gauge, labeled by service/method                       | Distinct clients currently holding at least one in-flight call for a method |
+| Node-wide in-flight       | Gauge, labeled by service/method                       | Current node-wide concurrent call count per method                          |
+| Client-state table size   | Gauge                                                  | Size of the per-client bookkeeping table, to catch unexpected growth        |
+| Block-read bulkhead usage | Gauge (in-use / available)                             | Current utilization of the shared backend read permit pool                  |
```
jsync-swirlds, line 416: A lot of this is already available in Helidon metrics if we enable those and add them to the OTEL collection.

jsync-swirlds (commented):
One major suggestion, to implement as an interceptor instead of wrapping the service interface.
Also a suggestion to describe as straight "Leaky Bucket As Meter" instead of bringing in GCRA baggage from ATM (Asynchronous Transfer Mode) networking.

docs/design/apis/api-throttling.md
Nana-EC: What would cause the scenario you noted to occur. Is this the case with the current chart and other solo provisioner deployment flow?
Granted a node operator can replace the ingest flow/loadBalancer as desired but as we've recommended is the scenario you highlight applicable.

Also do you have a suggestion?
As noted it's designed to be replaced if there's a reliable alternative.

docs/design/apis/api-throttling.md
Nana-EC: What's the logic for a simple counter. Are you suggesting a simplification of just a count in time as to the entries in the bucket vs the estimated arrival time?

docs/design/apis/api-throttling.md
Nana-EC: Let me research. 
Helidon didn't seem to play well with a few other options.
All in all once this settles i'm hoping we can contribute some of this back to PBJ itself.
Let me look and circle back

Nana-EC: Also is the efficiency in code complexity or you mean in resource management ?

docs/design/apis/api-throttling.md
Nana-EC: I'm interested, I'll look into it. Thanks

docs/design/apis/api-throttling.md
Nana-EC: Will take a look, Helidon didn't play well with other options so I may have missed this.

docs/design/apis/api-throttling.md
jsync-swirlds: It has nothing to do with our software, any L4+ router or load balancer may appear as the source of the connection depending on its configuration.

I'm not suggesting a change, but perhaps we note in the documentation that IP can be proxied or replaced depending on routing and load balancing, and while there are HTTP headers to report the original IP, not all systems fill that in.

docs/design/apis/api-throttling.md
jsync-swirlds: We have no need to "predict arrival time" as there _is no arrival time_.  This is where ATM is sneaking into the design. The networking GCRA has to track not just cells, but predict cell arrival time so it's keeping the VC full but not over-full.

As far as I can tell there's no need for timestamps in our system at all, just increment for connections and decrement at a steady rate (the "leaky" part of the bucket).
That's the leaky bucket as meter; run a meter (counter) that represents the amount in the bucket, add to it for new connections and remove from it at a steady rate.

docs/design/apis/api-throttling.md
jsync-swirlds: Both, using an interceptor reduces code complexity and consolidates the changes around the API registration. Interceptors are also, generally, fast and efficient.

docs/design/apis/api-throttling.md
Nana-EC: Got it, I've made the limitation more explicit.

docs/design/apis/api-throttling.md
Nana-EC: Agreed, I think it's a labelling issue so I've resolved the description to be more accurate. The spike is consistent with the simpler approach

docs/design/apis/api-throttling.md
Nana-EC: I think there was a misunderstanding of the manner of the application of the concept.
You'll see from the spike it's not as constrained as you thought.

I've updated the Terms section to lead with "leaky bucket" (the actual model) and notes GCRA as the specific state representation, not the framing. On the mechanism itself, I don't think a counter is actually simpler here
- A literal fixed-window counter needs a timestamp anyway (to know when the window resets) and has the boundary-burst problem.
- A sliding-window log avoids that bug but needs more state per client than GCRA does (a list of recent timestamps, not one).
- A token bucket with lazy refill is mathematically the same thing GCRA is. It's the same two numbers (rate, burst), opposite sign convention (remaining tokens vs. schedule position), so I don't think it removes the "timestamp", it just renames it.

`GcraLimiter` as built is one `AtomicLong` (the theoretical arrival time) and one CAS loop with no thread, no queue, no per-bucket bookkeeping beyond that single value. I think what reads as ATM-specific machinery is the label, not anything actually in the intended implementation (take a look at the spike).

I think the state model as-is works but I'm open to hear concrete cases where it's still of concern, especially after looking at the spike.

docs/design/apis/api-throttling.md
Nana-EC: Agreed, updated

docs/design/apis/api-throttling.md
Nana-EC: From my initial research I'm not seeing `ServerInterceptor` support in PBJ.
My understanding is PBJ doesn't depend on `io.helidon.webserver.grpc` which exposes `GrpcRouting.Builder.intercept(io.grpc.ServerInterceptor)`.
Instead it uses `PbjRouting`/`PbjProtocolHandler` which implements Helidon's  `Http2SubProtocolSelector.SubProtocolHandler` which talks through `Pipeline<Bytes>`.
I can take a second look but I don't think the piping lines up, can you share how it would line up if you've confirmed it?

That said, I think I'm already covering the intent of your "intercepting" ask.
The admission wrapper attaches once, at service registration time, in ServiceBuilderImpl.registerGrpcService(...)`, no plugin business logic changes in supporting this. Plugins never see or handle throttling in their own logic code. So functionally it's "the interceptor equivalent" for our stack, just built on `ServiceInterface`. Goal 4 calls this out specifically so the mechanism doesn't couple us to one web server.
Let me know if this lines up with your initial concern or if you think the  ServerInterceptor is a manageable lift.

docs/design/apis/api-throttling.md
Nana-EC: See comment above but I think it's the same diff.

docs/design/apis/api-throttling.md
jsync-swirlds: Leaky bucket as meter works with a single counter, but it doesn't work as you described.

1. The counter is only incremented by new requests.
1. The counter is decremented by a separate task (often a different thread) at a consistent rate; there is no timestamp check, just a regular monotonic "drip".
1. There is also no window, just increment as needed, and decrement monotonically.

We're not trying to measure requests in a time window; we're just filling and emptying a virtual bucket represented by the counter.

I don't think we need to overly specify the mechanism. We can describe a "leaky bucket as meter" instead of "leaky bucket of tokens" and let the implementation arise as needed.

docs/design/apis/api-throttling.md
jsync-swirlds: PBJ doesn't support interceptor, not because it cannot, but because it was never requested.

We're picking a highly specific implementation to work around a gap in the PBJ plugin (in my opinion) rather than using mechanisms that are found in nearly all environments and don't require wrapping an object when there is _already support_ (it's in Helidon, and we could access directly when creating the webserver, we don't actually need PBJ support) for interceptors.

PBJ _does_ use the Helidon webserver (we create it in App when plugins register new APIs), and it isn't just the gRPC support, we can add service interceptors on any webserver; and this throttling isn't gRPC specific, so we can filter for resources much earlier (and below the gRPC layer, avoiding parsing overhead) using an interceptor.

Note, as I said above, every webserver I know of supports an interceptor or filter mechanism, it's about as universal as anything in the space.

Service Interceptor is _simpler_ than the current spike, and changes much fewer files because it doesn't _require_ a change to every plugin (we can add the per-plugin config and throttles separately from enabling throttling) and has much less impact on test fixtures.

docs/design/apis/api-throttling.md
jsync-swirlds: I disagree that it is the same.
1. we are _currently_ wrapping each individual service interface
1. The proposalk is to assign an instance of a common class (which is not specific to any service) to the webserver interceptor list when the webserver routing is created.

I believe that using the mechanisms in the webserver that are specifically designed for throttling, authentication, etc... is naturally cleaner and more effective than building a new custom approach.

docs/design/apis/api-throttling.md
Nana-EC: Agreed on docs suggestions here.
I'll leave the Terms entry describing the model ("leaky bucket") without prescribing the state mechanism underneath, so it doesn't read as committing to either approach.

On implementation, I'd like to keep GcraLimiter as built in the spike rather than move to a periodic-decrement task.
My concern is a shared decrement thread has to visit every active client's counter on each tick, which is O(clients) per tick; GcraLimiter's CAS-per-call is O(1) and needs no thread at all.
Happy to revisit if you've got a concrete case the current model handles wrong, I'd appreciate you exposing that in the spike specifically so we're all on the same page..
Otherwise I think this one's just a documentation item, not a mechanism disagreement.

docs/design/apis/api-throttling.md
Nana-EC: Agreed, we haven't requested PBJ to implement this to date.
I would also have concerns around hand-building our own approach here vs what i've proposed here and illustrated in the spike or waiting on PBJ. What's the complexity and footprint of your approach as I'm open minded?
As you see in the spike the suggested approaches impact is minimal, each plugin only adds configs, there's no additional plugin logic to achieve the basic throttling. 
I'd prefer an approach that is validated and get us moving in this sprint, if we need to modify with a light lift later we can, what I've added seems to match that.
I'm open to a simpler and proven route but I'd ask for a specific suggestion from you with code snippets to illustrate on the spike. That would move things along faster than debating how the piping might/should work. My spike has validated certain pathways, so illustrating your validated suggestions would be helpful if you propose adopting them immediately as a replacement. 

Regarding the PBJ ticket, I'd say go ahead and open a ticket for PBJ to request such support as I agree on it's usefulness. You have a clearer picture on how that may work so you can make the appropriate request. I can reference the ticket and note it in the design doc that we incorporate it once delivered.

Let me know if you have an alternative speedy and functional suggestion.

docs/design/apis/api-throttling.md
Nana-EC: Fair, however, Helidon's metrics are all transport-level (requests, latencies, status codes) and has no concept of why a request may have been rejected (rate vs. per-client concurrency vs. global ceiling).

docs/design/apis/api-throttling.md
Nana-EC: Fair distinction.
I took another look to try to improve the registration logic.
`PbjRouting.Builder.service(ServiceInterface, PbjGrpcServiceConfig)` already takes a per-service config alongside the service, so `ServiceBuilderImpl` can wrap each registered service in one common decorator. This way each plugin no long has to pick a different overload. 
I'll update it in the spike

docs/design/apis/api-throttling.md
jsync-swirlds: Wrapping an interface is architecturally poor, in my opinion.
I think we're typing past one another here.

I'm recommending _using a standard approach_ while you keep describing the standard approach as something custom that has to be hand-built.
- We do not have to hand build this, I've noted repeatedly that the calls are already available.

There is a complete disconnect as I read it.

Separately:
It almost feels like there is a strong desire to take a spike (which is, definitionally, code to be discarded) and merge that to main as a completed task. If that perception is accurate, I think we're already completely off track.

docs/design/apis/api-throttling.md
jsync-swirlds: I'm recommending using calls on the webserver, and you're turning to the PBJ interfaces.

Those are not in the same realm.

docs/design/apis/api-throttling.md
jsync-swirlds: Helidon offers a lot of other metrics tied into the web service as well; many of which remove the need to add _some_ of these.

The Helidon libraries we're using operate well above the transport layer and are aware of (and expose metrics for) HTTP/2, among other application layer characteristics.

docs/design/apis/api-throttling.md
Nana-EC: Another try here to get alignment.
Unfortunately from my attemps it still doesn't work as cleanly as you're expecting. The Helidon <> PBJ logic does not support the mode you're thinking of. It would likely work if we were using the native Helidon gRPC module.

I've spent time to provide a [service-interceptor-gist](https://gist.github.com/Nana-EC/4a357dd690ef6037755c71461133cf3b) to validate the claims and considerations.


With this info the options include
1. Stick with the Service Interface approach I've suggested and validated
2. Open a Helidon bug and wait for an update to their `Http2ServerStream.handle()` dispatch logic
3. Open a PBJ issue and wait for it to add it's own interceptor hook
4. BN authors its own Http2SubProtocolProvider (defeats the purpose of using PBJ)

I think we can go with 1 and leave the door open for adopting 2 or 3 in the future.

jsync-swirlds: I would recommend (1) _and_ (3). Option (1) for a "want it today" option and (3) for a more general option that applies to other substrates (e.g. Netty)

There is no bug in Helidon, it's doing exactly what the PBJ plugin is asking it to do here, PBJ is informing Helidon that it will handle everything so Helidon should get out of the way; PBJ simply isn't implementing the filters and filter chains that Helidon (and every other webserver) use for this type of functionality.

I think (4) is a non-starter.

docs/design/apis/api-throttling.md
jsync-swirlds: Addressed elsewhere.

jsync-swirlds (commented):
I have no further objections that should block proceeding.

I still assert that this is not a great approach, but I will not block progress if others are willing to approve.

berryware:
I am in the camp of using well-defined gRPC constructs like Interceptors/MetadataCredentialsPlugin for filtering and auth. Having that in PBJ is a benefit to all. That said, if we do not want to be constrained or dependent on the PBJ development cycle, I am okay with proceeding with this design.

If PBJ ever adds in interceptors, will we be able to switch back?

berryware (approved):
Approved with non-blocking comment

Nana-EC:
> I am in the camp of using well-defined gRPC constructs like Interceptors/MetadataCredentialsPlugin for filtering and auth. Having that in PBJ is a benefit to all. That said, if we do not want to be constrained or dependent on the PBJ development cycle, I am okay with proceeding with this design.
> 
> If PBJ ever adds in interceptors, will we be able to switch back?

MetadataCredentialsPlugin is the C++ core API I believe, the closest server-side Java equivalent is a ServerInterceptor reading io.grpc.Metadata, which runs into the same issue as the Filter/interceptor thread above - it needs a io.grpc.Server dispatch, which we don't have in our PBJ dependent stack.

That said, I don't think auth is blocked today, PBJ RequestOptions.metadata() should already return every HTTP/2 header the client sends and would be available at the same attachment point we're using for throttling.

To your question, yes. ThrottledServiceInterface's admission logic (GcraLimiter, SingleWeightThrottle) has zero PBJ/gRPC dependency by design. So only the thin wrapper and registration logic would need to change if PBJ ever ships an interceptor hook and we want to adopt it. This would also be true for a future auth decorator.

ata-nas (approved):
Unblocking so we can move forward, but I also agree with @jsync-swirlds and @berryware about possible PBJ extensions and the different approach (to wrapping) that @jsync-swirlds notes on.

I guess discussion is not entirely over here and will continue in follow ups.

Also, do we ever envision possible "priority" awareness for throttling? Possibly giving more priority to some connections/clients/requests than others when we see high concurrency?

Finally, note that some of your MD renders poorly, can be addressed in a follow up.

Image

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+from the caller's network address in this design.</dd>
+<dt>Weight class</dt><dd>A cost tier assigned to a request (e.g. LIGHT, MODERATE, HEAVY) that determines which
+rate/concurrency policy applies to it.</dd>
+<dt>Content-aware weigher</dt><dd>A per-API function that inspects a request's content (e.g. the requested block
+number) to classify it into a weight class before admission, rather than relying on a single static weight for the
+whole API.</dd>
+<dt>Leaky bucket (rate limiter)</dt><dd>The rate-limiting model used per client per method: a request adds to a
+conceptual bucket, the bucket drains at a fixed rate, and a request is admitted only if the bucket isn't already
+full. Used strictly as a policer here (reject on arrival), not a shaper (delay and re-admit later); see
+[Alternatives considered](#alternatives-considered). The state representation used to implement this model — GCRA,
```
AlfredoG87, line 87: Broken link.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+see [`GcraLimiter`](#gcralimiter) — is an implementation choice, not part of the model itself.</dd>
```
AlfredoG87, line 88: also broken link.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+## Entities
+
+### `ThrottlePolicy`
+
+The runtime policy applied to one gRPC method for one client: a rate (requests per second), a burst tolerance, a
+per-client concurrency ceiling, and a node-wide concurrency ceiling.
+
+### `GcraLimiter`
+
+A lock-free leaky-bucket rate limiter, keyed per client, implemented via GCRA. Holds one monotonic-clock timestamp
```
AlfredoG87, line 104: maybe a reference to GCRA description in-depth link such as:
https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+(the theoretical arrival time) per key, advanced via compare-and-swap on each admitted request — deliberately not a
+token counter, since that would need the same timestamp anyway (to know when to refill) plus a second field.
+
+### `ClientKeyExtractor`
+
+Derives a client key from an incoming call's connection metadata. The default implementation uses the caller's
+remote network address. The interface is designed so an authenticated-identity-based implementation (e.g. an mTLS
+client certificate, or an OAuth/bearer-token subject once a token-based auth scheme exists) can replace it later
+without changing anything else in the system — the design deliberately doesn't commit to either as the eventual
+mechanism.
```
AlfredoG87, line 114: are we going to use the different interfaces at the time, have we considered the performance penalty of using interfaces instead of a concrete class?

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+
+### `ContentAwareWeigher`
+
+An optional, per-API function that classifies a request into a weight class based on its content, evaluated once at
+admission time before any rate/concurrency check runs. Used by `getBlock` and `subscribeBlockStream` to distinguish
+live/recent requests from historical ones.
```
AlfredoG87, line 120: how is the _cutover_ block number for defining what is on `recent` vs `historical` storage, do we cache that number, if so, for how long? is this an implementation detail worth defining on this design doc?

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+
+### The admission decorator
+
+A wrapper around a plugin's `ServiceInterface` implementation that performs the admission decision on every call
+and, if admitted, tracks the call's concurrency permit for its full lifetime.
```
AlfredoG87, line 125: how is this implemented? does it keep the goal of being decoupled from an specific webServer framework such as Helidon or Netty?

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+
+### `BlockReadBulkhead`
+
+A single, shared, bounded permit pool protecting the block-storage read path. Not keyed by client — it protects the
+resource itself, independent of who is reading from it.
```
AlfredoG87, line 130: How is this number defined, updated and why would it be changed?

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Per-API and global throttle configuration
+
+Each plugin owns a small configuration record describing its own per-client limits (rate, burst, per-client
+concurrency). Node-wide concurrency ceilings, which represent an allocation of shared node capacity across APIs,
+live in one shared, node-level configuration record.
```
AlfredoG87, line 136: where is this configuration record defined? does it expose the same config fields? since some APIs have different rate limiting considerations, examples `getBlock` and `subscribe` are context aware, does a different rate consideration could need a different rate config field?

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+almost immediately — but that does not mean the call is finished, since the server may continue streaming responses
+for a long time afterward. The `responses` pipeline's completion callbacks, by contrast, reliably fire exactly once
+when the call actually ends — on normal completion, on a business-logic error, on client cancellation, and on a
+deadline being exceeded — for both unary and streaming calls alike. The release logic needs its own single-fire
+guard, since more than one of those signals can arrive for the same call (for example, a cancellation arriving
+immediately after a business-logic error).
+
+**Content-aware weighting for `getBlock` and `subscribeBlockStream`.** A single static weight per API cannot express
+that a historical block read is more expensive than a live one. Both APIs register a `ContentAwareWeigher` that
+inspects the requested block number (for `getBlock`) or the requested start block (for `subscribeBlockStream`)
```
AlfredoG87, line 207: Does the ContentAwareWeigher consider the amount of blocks as opposed to requests, in `getBlock` a single requests = 1 block. on `subscribe` a single requests can have many historical blocks and many recent blocks. And the ReadStorage Resources should be limited based on amount of blocks (historical/recent), not requests.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+  session over one momentary saturation instant is a worse outcome than a short internal delay before retrying.
+  This wait is purely internal scheduling for already-admitted work — it never affects the admission decision in
+  Component A.
+
+The node-wide concurrency ceilings Component A applies to historical `getBlock` requests should be sized with this
+bulkhead's capacity in mind, since both draw from the same underlying resource — but because the bulkhead is also
+shared with subscriber catch-up traffic, which Component A's `getBlock` ceiling has no visibility into, sizing the
+two independently is a reasonable approximation rather than a guarantee that the bulkhead can never be contended by
+both sources at once. The bulkhead's own bounded behavior (reject or brief wait, never unbounded growth) is what
+keeps that scenario safe even so.
```
AlfredoG87, line 238: Not sure the sizing the two independently means here, does it means that there are 2 bulkhead limits for each source getBlock and subscribe?

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+- **Two implementation details are worth calling out explicitly, since they are the parts most likely to erode this
+  if done carelessly:**
+  - **Client-key derivation should happen once per connection, not once per call.** Deriving a key from
+    `RequestOptions.remoteAddress()` on every single unary call (e.g. every `getBlock` request on a connection that
+    is reused for many calls) would allocate a string repeatedly for no new information; the key should be computed
+    once and reused for the life of the underlying connection where possible.
+  - **Content-aware weighing must not require a full protobuf deserialization.** A weigher only needs to read one or
+    two fields (e.g. a block number) to classify a request. It should do a targeted read of that field directly from
+    the wire format, not fully deserialize the request message before the real handler does its own full parse —
+    otherwise every classified call would pay for parsing the request twice.
```
AlfredoG87, line 289: for `getBlock` and `subscribe` we are indeed parsing twice. which are the only current context-aware methods.

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+- **Acceptance criterion.** This goal should be validated, not assumed: a benchmark comparing per-call latency and
```
AlfredoG87, line 290: an specific way to do the benchmark? jmh microbenchmark?

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+  allocation with admission control enabled versus disabled on the same hardware belongs in the acceptance tests for
```
AlfredoG87, line 291: how are we going to assert on the same hardware?

docs/design/apis/api-throttling.md
```diff
@@ -0,0 +1,444 @@
...
+### Extensibility
+
+Two extension points are designed in from the start, since both are anticipated future work:
+
+- **Client identification.** `ClientKeyExtractor` is a small, isolated interface specifically so that authenticated
+  client identity — an API key, or an mTLS client certificate — can replace network-address-based identification
+  later by adding one new implementation, with no change needed to the admission decorator, `ThrottlePolicy`, or any
+  per-plugin configuration. `RequestOptions.remoteCertificateChain()` already exists today, unused, as exactly what
+  a future certificate-based extractor would read.
+- **Additional bulkheads for other shared resources.** `BlockReadBulkhead` is one instance of a general pattern: a
```
AlfredoG87, line 303: why not have 2 BlockReadBulkhead, one for Recent and another for Historical blocks?

AlfredoG87 (commented):
Posting comments post-merge. Nothing is blocking.

PR #3544, opened 2026-08-25
fix: reclaim stuck resend slots that never complete

### LINKED ISSUE #2934: fix(stream-publisher): reclaim resend slots that never complete
### Story Form
As a Block Node operator running the publisher (Tier-1) plugin
I want an incomplete block-resend slot to be reclaimed automatically after a timeout
So that a publisher that wins a resend slot and then stops sending cannot indefinitely block the acknowledgement and persistence of all subsequent blocks

### Technical Notes
There are three items here.
1. When a block is being resent, the plugin must still detect if the publisher stalls. This detection must be based on the latest block when the resend begins and the latest block when the resend ends.
2. We need a notification to signal backfill to immediately request a block that is to be resent, so we can get the block from another block node if we have a buggy publisher making it difficult to get that block directly.
3. We need to prevent the publisher that failed to send a block (or sent a bad block) from _also_ resending that block.  This is already partly implemented, but we need to extend that implementation to make sure we get resent blocks from _other_ publishers if at all possible.
   - Note, there will alway be other publishers in a stall detection case, but if the resend is because of a bad block proof or failed persistence, then we may not have an alternative publisher source.

#### Problem
When the publisher manager asks a publisher to resend a block (`RESEND(N)`), the first publisher to send block `N`'s header wins the slot: `getActionForHeader` removes `N` from `blocksToResend`, adds it to `activeResendBlocks`, and returns `ACCEPT`. From that point on, block `N` is exempt from stall detection — `checkForStalledHandlers` skips any block where `isResendingLive(N)` is true — because during catch-up the live edge is far ahead of `N` and the normal block-distance stall rule (`completedBlockNumber > candidateBlock + maxBlocksBeforeStalled`) would otherwise fire a false positive.

The gap: If the publisher that won the slot sends the header and then makes no further progress, block `N` stays in `activeResendBlocks` indefinitely. It is only cleared by `endOfBlock`, `blockIsEnding`, or a failed `handleVerification` — none of which occur for a publisher that simply stops streaming. Because the slot is never reclaimed:

- `RESEND(N)` is never re-issued (it is gone from `blocksToResend`).
- A different publisher offering block `N` receives `SKIP` and cannot take over (`N` is below `nextUnstreamedBlockNumber`).
- `correctForResendAndStreaming(...)` clamps every acknowledgement to `earliestKnownGap - 1`, where `earliestKnownGap` includes `activeResendBlocks.first()`. Acknowledgement and persistence therefore stop advancing for every block at or after `N` — even blocks that streamed and completed fine. This is a necessary limit, because we cannot acknowledge across gaps without risking a lost block.

Net effect: one incomplete resend wedges the acknowledgement/persistence pipeline for all later blocks until something external frees that block.  We need to both detect stall in resent blocks, and also make some effort to avoid receiving a resent block from the same source as the original, failed, block.

#### Proposed fix
Keep the `isResendingLive` exemption in `checkForStalledHandlers`, but add a second stall check that only applies to resending handlers.
- Add a `resendStartBlock` counter to the Handler, set this value when the block number is removed from `blocksToResend` and added to `activeResendBlocks`. Clear this value when an `endOfBlock` or new `Header` is received for that handler.
- When checking for stalled handlers, if isResendingLive(N) is true for the block a handler is receiving, then check that handler against it's `resendStartBlock` instead of the block number that handler is receiving.
   - The stall detection for a handler that is receiving a resent block is otherwise the same as any other Handler.

The following item is not required, and may be deferred to a later enhancement.
- Add a "handler source" to the `blocksToResend` data. If a Header for the block to resend is received by the same handler, and there is more than one active Handler, reply with SKIP_BLOCK instead of accepting that block.

#### Acceptance criteria
- [ ] A block holding a resend `ACCEPT` slot that makes no progress within the configured block limit is reclaimed (removed from `activeResendBlocks`, re-added to `blocksToResend`, owning handler ended with `TIMEOUT`).
- [ ] A legitimately-progressing resend is **not** reclaimed (no regression of the `isResendingLive` false-positive protection).
- [ ] After reclaim, a subsequent publisher offering that block is granted the slot and persistence/acknowledgement _eventually_ advances past the previously stuck block.
- [ ] The general stall handling process and metrics are unchanged.

rockysingh:
Fixes #2934.

## Summary

When a publisher wins `RESEND(N)` via the header path and then goes silent, block `N` stays in `activeResendBlocks` indefinitely — it's only cleared by `endOfBlock` / `blockIsEnding` / `handleVerification`, none of which fire for a publisher that just stops streaming. Because `correctForResendAndStreaming` clamps every ack at `activeResendBlocks.first() - 1`, one stuck resend wedges acknowledgement and persistence of every block at or after `N`. The existing `isResendingLive` exemption keeps the normal stall detector from firing (to avoid false positives during legitimate concurrent resends), so nothing else reclaims the slot.

## Fix

Add a second stall branch in `checkForStalledHandlers` that only applies to resending handlers. It reclaims the slot when the live tip advances past `resendStartReference + K` blocks after the resend was accepted.

- Record `lastPersistedBlockNumber` at the moment the resend slot is won (in `getActionForHeader`, alongside `activeResendBlocks.add(N)`).
- In `checkForStalledHandlers`, if `isResendingLive(candidateBlock)` is true and `completedBlockNumber > resendStartReference + K`, remove from `activeResendBlocks`, clear the reference, end the handler with `EndStream(TIMEOUT)`, and re-add the block to `blocksToResend` so another publisher can supply it.

Design choices worth flagging for review:

- **Reuses the existing `MaxFutureBlocksBeforeStalled` config knob** (same concept, same unit) rather than introducing a parallel `MaxBlocksBeforeResendStalled` key — one dial for operators to tune.
- **Stored on the manager** (parallel `ConcurrentMap<Long, Long>` keyed by block number) instead of on `PublisherHandler` — matches the existing pattern for `blocksToResend`, `activeResendBlocks`, `activeStreamHandlerByBlock`, and avoids adding a getter/coupling to Handler state. Semantically identical to a per-Handler field.
- **Reference map cleared at every site that removes from `activeResendBlocks`** (`endOfBlock`, `blockIsEnding`, `endAllHandlersAndClearTracking`). One-liner per site.
- **`Long.MAX_VALUE` default** in the getOrDefault so a race between reference-recording and stall-check can never fire a spurious reclaim.

## Tests

Three new tests under `StallDetectionTests` in `LiveStreamPublisherManagerTest`, one per acceptance criterion from the ticket:

- `testStuckResendHolderReceivesTimeout` — win RESEND(1), go silent, complete blocks 2 + 3 elsewhere → asserts stuck holder receives `EndStream(TIMEOUT)` and is shut down.
- `testReclaimedBlockOfferedToNextPublisher` — after the reclaim fires, the completing handler receives `ResendBlock(1)` on its next `endOfBlock` (so another publisher can supply the block).
- `testLegitResendNotReclaimed` — resend holder completes its block in time; live tip subsequently advances well past `K` → no `EndStream(TIMEOUT)` fires. Guards against regression of the `isResendingLive` false-positive protection.

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -35,7 +35,6 @@ import java.util.Objects;
 import java.util.TreeSet;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CancellationException;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
@@ -142,8 +141,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
     private final ConcurrentSkipListMap<Long, BlockItemUnparsed> blockProofs;
     private final ConcurrentSkipListSet<Long> endBlocksReceived;
     private final ConcurrentSkipListSet<Long> blocksToResend;
-    private final ConcurrentSkipListSet<Long> activeResendBlocks;
-    private final ConcurrentMap<Long, Long> resendStartReferenceByBlock;
+    private final ConcurrentSkipListMap<Long, Long> activeResendBlocks;
     /// Flow state per handler ID; created in addHandler, removed in removeHandler.
     private final ConcurrentSkipListMap<Long, HandlerFlowState> handlerFlowState;
     /// Maps block number to the ID of the handler that currently holds ACCEPT for that block.
@@ -181,8 +179,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         blockProofs = new ConcurrentSkipListMap<>();
         endBlocksReceived = new ConcurrentSkipListSet<>();
         blocksToResend = new ConcurrentSkipListSet<>();
-        activeResendBlocks = new ConcurrentSkipListSet<>();
-        resendStartReferenceByBlock = new ConcurrentHashMap<>();
+        activeResendBlocks = new ConcurrentSkipListMap<>();
         activeStreamHandlerByBlock = new ConcurrentSkipListMap<>();
         handlerFlowState = new ConcurrentSkipListMap<>();
         initializeBlockNumbers(serverContext);
@@ -305,7 +302,6 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         }
         // Remove from the active resend set, if it's present
         activeResendBlocks.remove(blockNumber);
-        resendStartReferenceByBlock.remove(blockNumber);
         // The completing block is no longer an ACCEPT candidate for stall detection.
         activeStreamHandlerByBlock.remove(blockNumber);
         // After removing the completing block, check whether any remaining
@@ -328,7 +324,6 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         final Deque<BlockItemSetUnparsed> deque = queueByBlockMap.remove(blockNumber);
         // Remove from the active resend set, if it's present (it should not be)
         activeResendBlocks.remove(blockNumber);
-        resendStartReferenceByBlock.remove(blockNumber);
         // Remove this block from the ACCEPT winner tracking in all cases;
         // if the handler dropped mid-block there is no stall to detect here.
         activeStreamHandlerByBlock.remove(blockNumber);
@@ -623,7 +618,6 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
             // resent that was never sent.
             blocksToResend.clear();
             activeResendBlocks.clear();
-            resendStartReferenceByBlock.clear();
             blockProofs.clear();
             queueByBlockMap.clear();
             endBlocksReceived.clear();
@@ -655,7 +649,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager 
         // Use block number + 1 as a default if something isn't set, because we'll reduce by one later.
         final long defaultValue = blockNumber + 1;
         final long earliestInFlight = queueByBlockMap.isEmpty() ? defaultValue : queueByBlockMap.firstKey();
-        final long activeResend = activeResendBlocks.isEmpty() ? defaultValue : activeResendBlocks.first();
+        final long activeResend = activeResendBlocks.isEmpty() ? defaultValue : activeResendBlocks.firstKey();
         final long scheduledResend = blocksToResend.isEmpty() ? defaultValue : blocksToResend.first();
         final long earliestResend = Math.min(activeResend, scheduledResend);
         final long earliestKnownGap = Math.min(earliestInFlight, earliestResend);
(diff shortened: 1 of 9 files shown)


block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -137,6 +140,7 @@ public final class LiveStreamPublisherManager implements StreamPublisherManager
     private final ConcurrentSkipListSet<Long> endBlocksReceived;
     private final ConcurrentSkipListSet<Long> blocksToResend;
     private final ConcurrentSkipListSet<Long> activeResendBlocks;
+    private final ConcurrentMap<Long, Long> resendStartReferenceByBlock;
```
jsync-swirlds: So, instead of adding a map (and an interface reference at that, please don't use interfaces in hot paths), we could use a `ConcurrentSkipListMap` for the `activeResendBlocks` and use that alone (no new map) for both "is active resend" and "is stalled".

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
```diff
@@ -721,6 +729,21 @@ private void checkForStalledHandlers(final long completedBlockNumber) {
                 // This thread won the CAS — execute the stall action.
                 endStalledBlock(completedBlockNumber, handlerId, candidateBlock, STALL_DETECTED_LOG_MESSAGE);
                 blocksToResend.add(candidateBlock);
+            } else if (isResendingLive(candidateBlock)
+                    && completedBlockNumber
+                            > resendStartReferenceByBlock.getOrDefault(candidateBlock, Long.MAX_VALUE)
```
jsync-swirlds: Why would you use a getOrDefault here?
It becomes academic if we stick with a single map, but it would make more sense to just check that the reference exists (Map.contains(key) before this clause.

jsync-swirlds (changes requested):
We could create this more efficiently (particularly as to memory and allocations) if we upgraded the current Set to a Map (and use skip list map instead of adding a hash map; there are good reasons we don't use hash maps for the other maps in this class).

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
rockysingh: Done. Merged into a single ConcurrentSkipListMap keyed by block number with the reference block as the value.

block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
rockysingh: Switched to containsKey plus get. Cleaner now that the reference lives in the same map.

jsync-swirlds (dismissed):

Nana-EC (dismissed):
LG, might need to fix failing CI

AlfredoG87 (dismissed):

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublishApiUtility.java
```diff
@@ -17,28 +16,38 @@ public final class PublishApiUtility {
...
-    public static void endThisBlock(final PublisherHandler publisherHandler, final long blockNumber) {
+    public static void endThisBlock(final PublisherHandler handler, final long blockNumber) {
         final BlockEnd endOfBlock =
                 BlockEnd.newBuilder().blockNumber(blockNumber).build();
         final PublishStreamRequestUnparsed request =
                 PublishStreamRequestUnparsed.newBuilder().endOfBlock(endOfBlock).build();
-        publisherHandler.onNext(request);
+        // Avoid deadlock by ensuring there is a messaging budget to spend first.
+        final long currentBudget = handler.getMessageBudget();
+        if (currentBudget < 10) {
```
Nana-EC, line 28: Q:Is 10 an arbitrary number consistently used here or what's the logic behind that value for the budget that ensure we don't hit this again?

Nana-EC (approved):
Unblocking due to pass but unclear why 10 is the budget set.
If clarification or improvement to make it future proof is needed it can be done in follow up

block-node/stream-publisher/src/test/java/org/hiero/block/node/stream/publisher/PublishApiUtility.java
jsync-swirlds: This is just a backstop to prevent deadlock; it's just enough to ensure that tests that do not pay attention to the budget don't get shut down.
10 is 10 calls to the `onNext` method, which is plenty for a single test block.

Just 1 wasn't enough, so I chose a larger, but still small, value.

ata-nas (approved):

PR #3547, opened 2026-08-25
fix: tolerate null MD5 in updateDayListings for GCS composite objects 

### LINKED ISSUE #3546: fix(tools): days updateDayListings NPE on GCS composite objects (null MD5)
## Summary

`days updateDayListings` throws `NullPointerException` in `UpdateDayListingsCommand.convertToListingRecordFileStatic` when the bucket lister returns a `ChainFile` whose `md5()` is `null`. GCS composite objects (created via multipart upload or `gcloud storage compose`) do not expose an MD5 in their object metadata, and `BuckyBucketLister` (S3-compatible → GCS via HMAC) correctly surfaces that as `null`. The tool doesn't null-check before calling `Base64.getDecoder().decode(...)` and crashes on the first such object.

## Repro

Custom-network config (sphere) with a GCS bucket that contains at least one composite object for the target day.

```
java -jar tools-*-all.jar days updateDayListings \
  --cache --cache-dir <cache> \
  --network=other \
  --listing-dir <listing-dir> \
  --day 2026-08-14
```

Observed:

```
[BuckyBucketLister] Found 139424 files for 2026-08-14

Error updating listings for 2026-08-14: Cannot invoke "String.getBytes(java.nio.charset.Charset)" because "src" is null
java.lang.NullPointerException: Cannot invoke "String.getBytes(java.nio.charset.Charset)" because "src" is null
    at java.base/java.util.Base64$Decoder.decode(Base64.java:593)
    at org.hiero.block.tools.days.subcommands.UpdateDayListingsCommand.convertToListingRecordFileStatic(UpdateDayListingsCommand.java:413)
    at org.hiero.block.tools.days.subcommands.UpdateDayListingsCommand.processDayStatic(UpdateDayListingsCommand.java:389)
    at org.hiero.block.tools.days.subcommands.UpdateDayListingsCommand.updateListingsForSingleDay(UpdateDayListingsCommand.java:491)
    at org.hiero.block.tools.days.subcommands.UpdateDayListingsCommand.run(UpdateDayListingsCommand.java:106)
    …
```

## Root cause

`tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/UpdateDayListingsCommand.java:411-414`:

```java
// Convert MD5 from Base64 to hex
final String md5Base64 = chainFile.md5();
final byte[] md5Bytes = Base64.getDecoder().decode(md5Base64);   // NPE if md5Base64 == null
final String md5Hex = HexFormat.of().formatHex(md5Bytes);
```

GCS composite objects have no MD5 exposed in metadata — this is documented behavior on the GCS side and not fixable in the bucket lister. The tool must tolerate a null MD5.

## Impact

Blocks every `updateDayListings` run against a bucket that contains any composite object — a single bad file terminates the whole day's processing before any listing file is written.

## Proposed fix

Null-check `md5Base64` before decoding; use an empty string for the hex when MD5 is unavailable. Any downstream code that relies on the MD5 already needs to tolerate missing values against GCS composite objects, so surfacing `""` here is consistent.

```java
final String md5Base64 = chainFile.md5();
final String md5Hex;
if (md5Base64 == null || md5Base64.isEmpty()) {
    md5Hex = "";
} else {
    final byte[] md5Bytes = Base64.getDecoder().decode(md5Base64);
    md5Hex = HexFormat.of().formatHex(md5Bytes);
}
```

## Acceptance criteria

- [ ] `updateDayListings` completes against a bucket containing at least one composite (null-MD5) object.
- [ ] The written listing includes the file with an empty MD5 field (or whatever downstream sentinel is agreed).
- [ ] Existing tests still pass.
- [ ] A regression test exercises the null-MD5 path.

rockysingh:
Fixes #3546.

## Summary

`days updateDayListings` throws `NullPointerException` and drops the whole day's listing when `BuckyBucketLister` (S3-compatible → GCS via HMAC) surfaces a `ChainFile` with a null `md5()`. GCS does not compute an MD5 for **composite objects** (created via multipart upload or `gcloud storage compose`) — this is documented GCS behavior, not a lister bug — so the tool must tolerate a null MD5.

## Fix

`UpdateDayListingsCommand.convertToListingRecordFile{,Static}` — both variants had bit-for-bit identical code — now null-check `md5Base64` before decoding and emit an empty hex string when MD5 is unavailable:

```java
final String md5Base64 = chainFile.md5();
final String md5Hex;
if (md5Base64 == null || md5Base64.isEmpty()) {
    md5Hex = "";
} else {
    final byte[] md5Bytes = Base64.getDecoder().decode(md5Base64);
    md5Hex = HexFormat.of().formatHex(md5Bytes);
}
```

Downstream consumers of the listing that inspect MD5 will need to treat an empty string as "no MD5 available." This is consistent with the underlying GCS reality: composite objects genuinely have no MD5 to compare against.

## Repro

Any bucket containing at least one GCS composite object:

```
java -jar tools-*-all.jar days updateDayListings \
  --cache --cache-dir <cache> --network=other \
  --listing-dir <listing-dir> --day <day>
```

Pre-fix: `NullPointerException` from `Base64.getDecoder().decode(null)` at `UpdateDayListingsCommand.java:413` — no listing file written.

Post-fix: listing completes; composite objects appear with `md5=""` (or whatever downstream sentinel is agreed).

## Workaround (for anyone on this while the fix rolls)

Drop `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` (or equivalent HMAC creds) from the environment before running. The tool falls back to `GCPBucketLister` via ADC, which handles the metadata differently and does not fail on the same object.

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/UpdateDayListingsCommand.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/subcommands/UpdateDayListingsCommand.java
@@ -408,10 +408,18 @@ public class UpdateDayListingsCommand implements Runnable {
             relativePath = path;
         }
 
-        // Convert MD5 from Base64 to hex
+        // Convert MD5 from Base64 to hex. GCS composite objects (created via multipart
+        // upload or gcloud storage compose) do not expose an MD5 in object metadata,
+        // so BuckyBucketLister surfaces it as null; write an empty MD5 for those and
+        // let downstream consumers treat empty as "no MD5 available" (see #3546).
         final String md5Base64 = chainFile.md5();
-        final byte[] md5Bytes = Base64.getDecoder().decode(md5Base64);
-        final String md5Hex = HexFormat.of().formatHex(md5Bytes);
+        final String md5Hex;
+        if (md5Base64 == null || md5Base64.isEmpty()) {
+            md5Hex = "";
+        } else {
+            final byte[] md5Bytes = Base64.getDecoder().decode(md5Base64);
+            md5Hex = HexFormat.of().formatHex(md5Bytes);
+        }
 
         // Extract timestamp from path
         final java.time.LocalDateTime timestamp =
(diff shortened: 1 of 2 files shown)


jsync-swirlds (approved):

berryware (approved):
Looks good

PR #3557, opened 2026-08-26
fix: Increase the reliability and speed of WRB CLI

### LINKED ISSUE #3543: Update the WRB CLI to be compatible with solo E2E Tests
### Story Form

As a BlockNode engineer, I want to be able to test the WRB CLI using a solo E2E Test environment so that I can prove the the functionality works correctly. 

### Technical Notes

The solo E2E test environment is not 100% identical to mainnet and previewnet, so some modifications need to be made to make the WRB CLI work in that environment. 

1. Solo's MinIO only Solo's MinIO only stores .rcd.gz, never .rcd_sig. The code throws for any non-trailing record group with no sig files, killing the wrap before a single block was produced.
2. live-wrap poll loop unconditionally deletes and re-wraps the current zip range even when block_times.bin hasn't advanced, causing wasted work and spurious non-zero exits that eventually triggers the poll loop's "give up" counter.
3. block-0 genesis roster does not have the right timestamp. Block 0's own address book update is the network's founding roster and must take effect starting at block 0, not block0Time + 1ns. The +1ns offset (correct for all later blocks, which are signed with the old roster) was being applied unconditionally, causing block-0 signature verification to check against a stale bootstrap entry.

berryware:
- wrap no longer aborts on missing .rcd_sig files
- live-wrap poll loop exits 0 immediately when nothing is new.
- block-0 genesis roster applied at the right timestamp

## Reviewer Notes
<analysis>
The Java tools changes in this branch fix three independent bugs, all surfaced by the Solo e2e test environment:

1. **`TarZstdDayReaderUsingExec` — wrap no longer aborts on missing `.rcd_sig` files.** Solo's MinIO only stores `.rcd.gz`, never `.rcd_sig`. The old code threw for any non-trailing record group with no sig files, killing the wrap before a single block was produced. The fix logs a warning and continues wrapping with 0 signature files; 

2. **`ToWrappedBlocksCommand` — live-wrap poll loop exits 0 immediately when nothing is new.** Without this, every poll unconditionally deleted and re-wrapped the current zip range even when block_times.bin hadn't advanced, causing wasted work and spurious non-zero exits that would eventually trigger the poll loop's "give up" counter.

3. **`ToWrappedBlocksCommand` + `AddressBookRegistry` — block-0 genesis roster applied at the right timestamp.** Block 0's own address book update is the network's founding roster and must take effect starting at block 0, not `block0Time + 1ns`. The `+1ns` offset (correct for all later blocks, which are signed with the *old* roster) was being applied unconditionally, causing block-0 signature verification to check against a stale bootstrap entry. The fix uses `blockInstant` exactly for `blockNum == 0`, and `AddressBookRegistry.updateAddressBook()` now replaces the existing entry in place when the caller supplies a timestamp that exactly matches the last entry (the genesis case), preventing duplicate-timestamp entries that confuse binary search.

## Related Issue(s)
Closes #3543

### DIFF
--- a/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/model/TarZstdDayReaderUsingExec.java
+++ b/tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/model/TarZstdDayReaderUsingExec.java
@@ -349,21 +349,19 @@ public class TarZstdDayReaderUsingExec {
                         "Primary record file not found for baseKey='" + baseKey + "' in dir='" + currentDir + "'");
             }
 
-            // Signature files are optional: the chronologically last group may legitimately be
-            // missing its .rcd_sig (still uploading), and any group may be missing sigs when the
-            // record source (e.g. MinIO) does not store .rcd_sig files at all (e.g. Solo's MinIO
-            // only stores .rcd.gz). Skip the trailing group so the caller can retry once the sig
-            // lands; for non-trailing groups log a warning and wrap without RSA signatures —
-            // TSS-signed blocks don't need RSA sigs, and RSA-signed blocks produce 0 valid sigs
-            // (cosmetically wrong but non-fatal: validation simply reports 0 verified nodes).
+            // Signature files are required: a record file with no RSA signatures cannot be
+            // verified and must not be wrapped. The chronologically last group may legitimately
+            // be missing its .rcd_sig (still uploading) — skip it so the caller can retry once
+            // signatures land. Any other sig-less group is also skipped with a warning.
             if (signatureFiles.isEmpty()) {
                 if (groupIndex == lastGroupIndex) {
                     System.err.println("Skipping incomplete trailing record for baseKey='" + baseKey + "' in dir='"
                             + currentDir + "' (no signature file yet; likely still uploading)");
-                    continue;
+                } else {
+                    System.err.println("WARNING: Skipping record baseKey='" + baseKey + "' in dir='" + currentDir
+                            + "': no RSA signature files found; record cannot be verified");
                 }
-                System.err.println("WARNING: No signature files for baseKey='" + baseKey + "' in dir='" + currentDir
-                        + "'; wrapping without RSA signatures");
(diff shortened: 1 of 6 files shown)


tools-and-tests/tools/src/main/java/org/hiero/block/tools/days/model/TarZstdDayReaderUsingExec.java
```diff
@@ -340,17 +349,21 @@ private static void processDirectoryFiles(
...
             }
 
-            // There must be at least one signature file for the group; enforce invariant
+            // Signature files are optional: the chronologically last group may legitimately be
+            // missing its .rcd_sig (still uploading), and any group may be missing sigs when the
+            // record source (e.g. MinIO) does not store .rcd_sig files at all (e.g. Solo's MinIO
+            // only stores .rcd.gz). Skip the trailing group so the caller can retry once the sig
+            // lands; for non-trailing groups log a warning and wrap without RSA signatures —
+            // TSS-signed blocks don't need RSA sigs, and RSA-signed blocks produce 0 valid sigs
+            // (cosmetically wrong but non-fatal: validation simply reports 0 verified nodes).
```
jsync-swirlds: This isn't entirely accurate.
- There are no TSS signed record files, and never will be (as far as we know). Even if there was a TSS signature, it would be in a `sig` file. There is nowhere in a record file to hold a signature.
- A Record file with no RSA signatures is not valid, and _must not_ be wrapped, because it cannot be verified. From a CLI perspective we should always have at least 15 signatures (except for very old record files).

We shouldn't crash/exit the tool when signature files are missing, but we shouldn't wrap that record file either.

berryware: Updated comment, removed the wrapping of record files with no RSA signatures

jsync-swirlds (approved):

ata-nas (approved):

PR #3568, opened 2026-08-28
fix: Backfill counts a verification-failed block as persisted, so it is never re-fetched

### LINKED ISSUE #3541: Backfill counts a verification-failed block as persisted, so it is never re-fetched
### Story Form

As a Block Node operator,
I want a backfilled block whose verification fails to be fetched again,
So that a transient failure costs one retry instead of leaving a permanent hole in the node's history.

### Technical Notes

Observed in run 32809922059, job run-tests (fan-out-3cn-2bn, ...), on Block Node 0.41.0-rc1. block-node-2 has no publisher and backfills from block-node-1. It ends the run holding exactly one block (0) and never recovers.

Sequence from block-node-2-stdout.log:
  
1. 05:21:42.556 — blocks 1, 2, 3, 4 fail verification with MISSING_VERIFICATION_DATA.
2. 05:21:42.615 — VerificationDataProvider.updateTssData runs, i.e. after those four blocks were verified. TSS data is loaded while processing block 0; the backfill chunk [0->4] is verified concurrently, so blocks 1-4 race ahead of the data they need.
3. 05:21:42.763 — block 0 verifies and persists. It is the only block that ever lands.
4. 05:22:01.886 — Partially backfilled gap [0]->[5], completed up to block [4], then Reset liveTailHighWaterMark to [4].
5. Every later scan: Determined available peer ranges beyond block [0]: [[1->91]] → Adjusted live-tail gap from [1->91] to [5->91]. Blocks 1-4 are outside every gap the plugin will ever schedule again.

The reason step 4 reports "completed up to block [4]" for four blocks that never persisted:

- `BackfillPersistenceAwaiter.handleVerification` counts the latch down on verification failure, to fail fast rather than wait for a persistence notification that will never arrive.
- `BackfillPersistenceAwaiter.awaitPersistence` returns `latch.await(...)`, which is true for any countdown. It cannot distinguish "persisted" from "failed fast".
- `BackfillRunner.awaitBlocksPersistence` treats that true as success and leaves the block out of stillPending, so run(gap) returns 4 as the last successful block.

### LINKED ISSUE #3542: configure-bn-live-tail-backfill.sh aborts the test on a single metrics read
### Story Form

As an engineer reading a Solo E2E result,
I want the backfill test to survive a momentary port-forward blip,
So that a red `2cn-2bn-backfill` job means the Block Node failed to backfill, not that one curl timed out.

### Technical Notes

Run 32809922059, job run-tests (2cn-2bn-backfill, ...):

[25:33] EVENT configure-bn1-live-tail-backfill: Raise BN1's earliestManagedBlock above the chain head and throttle its backfill
[backfill-live-tail-config] ERROR: could not read the network height from block-node-2 (is its port-forward up?)
[25:38] FAIL configure-bn1-live-tail-backfill failed

The same job passed in the scheduled run 32815994520. The only material difference is elapsed time: wait-for-network-height took 24m31s on the red run and 18m31s on the green one, and port forwards are established once in the Solo Port forwards workflow step, long before that. The test's only refresh-ports event runs after scale-up-bn1, i.e. two events too late.

read_bn_metric_int (scripts/backfill/common.sh) issues a single 

```
curl -s --max-time 5 http://localhost:$(bn_metrics_port ...)/metrics 
```
and returns 1 if it produces nothing.

configure-bn-live-tail-backfill.sh turns that single miss straight into fail. There is no retry and no re-establishment of the forward.

Two consequences beyond the first failure:

- The run then reports a second, misleading failure — assert-backfill-during-live-stream fails with state file /tmp/backfill-live-tail.state not found, which is purely a cascade of the aborted configure step.
- The tss-signature-transition test that runs afterwards in the same job fails with WRAPS not detected within 1000 blocks, because block-node-1 was left wiped and empty. Also a cascade.

ivannov:
## Reviewer Notes

* Backfill no longer counts a verification-failed block as persisted
* Fetches the genesis block in a chunk of its own, so later blocks cannot race the TSS bootstrap
* Makes the Solo E2E harness survive a dropped port forward
* Prints captured CLI output in the WRB test scripts so a failed write on a slow CI pipe cannot fail the step

## Related Issue(s)

Closes #3541
Closes #3542

### DIFF
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
@@ -9,7 +9,6 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import org.hiero.block.node.spi.blockmessaging.BlockNotificationHandler;
 import org.hiero.block.node.spi.blockmessaging.BlockSource;
 import org.hiero.block.node.spi.blockmessaging.PersistedNotification;
@@ -41,11 +40,36 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
      * The flag is last-write-wins - several persistence plugins can report on the same block, so a
      * late failure can overwrite an earlier success. The worst case is one extra fetch of a block
      * that is idempotent to re-persist, which is cheaper than sticky-success bookkeeping.
-     *
-     * @param latch released once the block's fate is known
-     * @param succeeded whether the block was reported as successfully persisted
      */
-    private record Pending(CountDownLatch latch, AtomicBoolean succeeded) {}
+    private static final class Pending {
+        private final CountDownLatch latch = new CountDownLatch(1);
+        private volatile boolean succeeded = false;
+
+        /**
+         * Records the outcome and releases the latch. The outcome is written first, so a waiter woken
+         * by the latch always sees it.
+         *
+         * @param successful whether the block was reported as successfully persisted
+         */
+        void complete(final boolean successful) {
+            succeeded = successful;
+            latch.countDown();
+        }
+
+        /**
+         * @return whether the block was reported as successfully persisted
+         */
+        boolean succeeded() {
+            return succeeded;
+        }
+
+        /**
+         * @return the latch released once the block's fate is known
+         */
+        CountDownLatch latch() {
+            return latch;
+        }
+    }
 
     /**
      * Map of block numbers to the latch/outcome pair released when the block's fate is known.
@@ -65,7 +89,7 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
         pendingBlocks.computeIfAbsent(blockNumber, k -> {
             final String trackingBlockMsg = "Tracking block [{0}] for persistence";
             LOGGER.log(TRACE, trackingBlockMsg, blockNumber);
-            return new Pending(new CountDownLatch(1), new AtomicBoolean(false));
+            return new Pending();
         });
     }
 
@@ -95,16 +119,12 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
             boolean completed = pending.latch().await(timeoutMs, TimeUnit.MILLISECONDS);
             if (completed) {
                 final String persistenceConfirmedMsg = "Block [{0}] persistence resolved, succeeded=[{1}]";
-                LOGGER.log(
-                        TRACE,
-                        persistenceConfirmedMsg,
-                        blockNumber,
-                        pending.succeeded().get());
+                LOGGER.log(TRACE, persistenceConfirmedMsg, blockNumber, pending.succeeded());
             } else {
                 final String persistenceTimedOutMsg = "Block [{0}] persistence timed out after [{1}]ms";
                 LOGGER.log(DEBUG, persistenceTimedOutMsg, blockNumber, timeoutMs);
             }
-            return completed && pending.succeeded().get();
+            return completed && pending.succeeded();
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             final String waitInterruptedMsg = "Block [%s] persistence wait interrupted".formatted(blockNumber);
@@ -139,9 +159,7 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
                 final String persistenceFailedMsg = "Block [{0}] persistence failed";
                 LOGGER.log(INFO, persistenceFailedMsg, blockNumber);
             }
-            // Record the outcome before releasing, so a waiter woken by the latch always sees it.
-            pending.succeeded().set(notification.succeeded());
-            pending.latch().countDown();
+            pending.complete(notification.succeeded());
         }
     }
 
@@ -167,9 +185,7 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
             if (pending != null) {
                 final String verificationFailedMsg = "Block [{0}] verification failed, marking block as not persisted";
                 LOGGER.log(INFO, verificationFailedMsg, blockNumber);
-                // Record the outcome before releasing, so a waiter woken by the latch always sees it.
-                pending.succeeded().set(false);
-                pending.latch().countDown();
+                pending.complete(false);
             }
         }
     }
@@ -179,10 +195,10 @@ public class BackfillPersistenceAwaiter implements BlockNotificationHandler {
      * resetting state.
      */
     public void clear() {
-        // Release all waiting threads before clearing. The outcome flag is left as-is (false unless
-        // something already reported success), so a block abandoned by a shutdown reads as not persisted.
+        // Release all waiting threads before clearing. A block abandoned by a shutdown is reported as
+        // not persisted, so it is re-detected and re-fetched on the next run.
         for (Pending pending : pendingBlocks.values()) {
-            pending.latch().countDown();
+            pending.complete(false);
         }
         pendingBlocks.clear();
     }
--- a/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
+++ b/block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
@@ -9,6 +9,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.hiero.block.node.spi.blockmessaging.BlockNotificationHandler;
 import org.hiero.block.node.spi.blockmessaging.BlockSource;
 import org.hiero.block.node.spi.blockmessaging.PersistedNotification;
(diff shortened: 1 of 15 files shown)


block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
```diff
@@ -26,16 +27,32 @@
...
-     * notification is received.
+     * A tracked block: the latch released once its fate is known, plus whether that fate was success.
+     * The flag is last-write-wins - several persistence plugins can report on the same block, so a
+     * late failure can overwrite an earlier success. The worst case is one extra fetch of a block
+     * that is idempotent to re-persist, which is cheaper than sticky-success bookkeeping.
+     *
+     * @param latch released once the block's fate is known
+     * @param succeeded whether the block was reported as successfully persisted
+     */
+    private record Pending(CountDownLatch latch, AtomicBoolean succeeded) {}
```
jsync-swirlds: Instead of using an AtomicBoolean, we could use a private internal `boolean` and implement succeeded here:

```suggestion
    private record Pending(CountDownLatch latch) {
        volatile private boolean succeeded = false;

        public void complete(final boolean successful) {
            succeeded = successful;
            latch.countDown();
        }
    }
```

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
```diff
@@ -70,15 +92,19 @@ public boolean awaitPersistence(long blockNumber, long timeoutMs) {
...
+                        TRACE,
+                        persistenceConfirmedMsg,
+                        blockNumber,
+                        pending.succeeded().get());
             } else {
                 final String persistenceTimedOutMsg = "Block [{0}] persistence timed out after [{1}]ms";
                 LOGGER.log(DEBUG, persistenceTimedOutMsg, blockNumber, timeoutMs);
             }
-            return completed;
+            return completed && pending.succeeded().get();
```
jsync-swirlds: If the change in the record is taken:

```suggestion
            return completed && pending.succeeded();
```

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
```diff
@@ -103,23 +130,26 @@ public void handlePersisted(@NonNull PersistedNotification notification) {
...
                 final String receivedConfirmationMsg = "Received persistence confirmation for block [{0}]";
                 LOGGER.log(TRACE, receivedConfirmationMsg, blockNumber);
             } else {
                 final String persistenceFailedMsg = "Block [{0}] persistence failed";
                 LOGGER.log(INFO, persistenceFailedMsg, blockNumber);
             }
-            latch.countDown();
+            // Record the outcome before releasing, so a waiter woken by the latch always sees it.
+            pending.succeeded().set(notification.succeeded());
+            pending.latch().countDown();
```
jsync-swirlds: If the change in the record is taken:

```suggestion
            pending.complete(notification.succeeded());
```

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
```diff
@@ -133,11 +163,13 @@ public void handleVerification(@NonNull VerificationNotification notification) {
...
-            if (latch != null) {
-                final String verificationFailedMsg = "Block [{0}] verification failed, releasing latch";
+            Pending pending = pendingBlocks.get(blockNumber);
+            if (pending != null) {
+                final String verificationFailedMsg = "Block [{0}] verification failed, marking block as not persisted";
                 LOGGER.log(INFO, verificationFailedMsg, blockNumber);
-                latch.countDown();
+                // Record the outcome before releasing, so a waiter woken by the latch always sees it.
+                pending.succeeded().set(false);
+                pending.latch().countDown();
```
jsync-swirlds: If the change in the record is taken:

```suggestion
                pending.complete(false);
```

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
```diff
@@ -147,9 +179,10 @@ public void handleVerification(@NonNull VerificationNotification notification) {
      * resetting state.
      */
     public void clear() {
-        // Release all waiting threads before clearing
-        for (CountDownLatch latch : pendingBlocks.values()) {
-            latch.countDown();
+        // Release all waiting threads before clearing. The outcome flag is left as-is (false unless
+        // something already reported success), so a block abandoned by a shutdown reads as not persisted.
+        for (Pending pending : pendingBlocks.values()) {
+            pending.latch().countDown();
```
jsync-swirlds: If the change in the record is taken:

```suggestion
                pending.complete(false);
```

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillRunner.java
jsync-swirlds: This shouldn't declare InterruptedException; I don't see anywhere it's thrown.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillRunner.java
```diff
@@ -282,22 +282,21 @@ private boolean sendChunkAndAwaitPersistenceWithRetries(
...
-                // actually be in flight.
+                // failure, timeout, or interrupt alike), so without re-tracking here the next await call
+                // would immediately report "already persisted or not tracked" for a block that may still
+                // actually be in flight. A block that definitively failed is re-tracked too and burns
+                // another perBlockProcessingTimeout on a latch nothing will release; that is accepted,
+                // since giving it its own outcome state costs more than the rare extra wait.
                 for (long blockNumber : stillPending) {
                     persistenceAwaiter.trackBlock(blockNumber);
                 }
+                Thread.sleep(Math.max(0, (long) config.initialRetryDelay() * attempt));
```
jsync-swirlds: If you use `LockSupport.parkNanos`, you can avoid the `InterruptedException` here and not declare it thrown by this method (which would remove the need to declare it in other methods up the chain).

About the only major difference between `sleep` and `parkNanos` is that sleep _uses_ `parkNanos` but throws `InterruptedException` if the `parkNanos` call completes early.

In almost all of our usage we don't need (or want) that exception.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-port-forward.sh
```diff
@@ -69,10 +78,73 @@ done
...
+  pkill -f "${PF_PATTERN}" 2>/dev/null || true
+}
+
+if [[ "${STOP_ONLY}" == "true" ]]; then
+  echo "Stopping port forwards for namespace: ${NAMESPACE}"
+  stop_forwards
+  echo "Port forwards stopped."
+  exit 0
+fi
+
```
jsync-swirlds, line 109: Instead of manual pkill, can we use native Kubernetes commands to stop the port forwarding?

jsync-swirlds (commented):
A few non-blocking suggestions/questions.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
ivannov: Changed to a small private class (records cannot declare mutable instance fields). The countdown latch is also internal detail

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
ivannov: Done

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
ivannov: Done

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
ivannov: Done

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
ivannov: Done, also with a change in he comment

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillRunner.java
ivannov: Correct, removed

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillRunner.java
ivannov: Changed to `parkNanos`

tools-and-tests/scripts/solo-e2e-test/scripts/solo-port-forward.sh
ivannov: I am not sure whether this can be done with `kubectl`

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
```diff
@@ -70,15 +116,15 @@ public boolean awaitPersistence(long blockNumber, long timeoutMs) {
...
-                final String persistenceConfirmedMsg = "Block [{0}] persistence confirmed";
-                LOGGER.log(TRACE, persistenceConfirmedMsg, blockNumber);
+                final String persistenceConfirmedMsg = "Block [{0}] persistence resolved, succeeded=[{1}]";
+                LOGGER.log(TRACE, persistenceConfirmedMsg, blockNumber, pending.succeeded());
             } else {
                 final String persistenceTimedOutMsg = "Block [{0}] persistence timed out after [{1}]ms";
                 LOGGER.log(DEBUG, persistenceTimedOutMsg, blockNumber, timeoutMs);
             }
-            return completed;
+            return completed && pending.succeeded();
```
ata-nas, line 131: What if persistence takes longer than the latch? It still could have succeeded, but not in time for our latch's time limit. I saw the javadoc, but was wondering how the caller of this method interprets this.

ivannov: The caller doesn't distinguish "failed" from "timed out" - `awaitBlocksPersistence` puts every unconfirmed block in stillPending, and `sendChunkAndAwaitPersistenceWithRetries` retries it and waits again without resending. So a slow but successful persistence usually just gets confirmed on a later attempt.

If it is so slow, that `config.maxRetries()` retries are still not enough for reporting success, then the chunk is reported incomplete, and the block is picked up by the next gap scan. It will read the actual stored ranges and if the block was meanwhile stored, it will not re-fetch it.

In the worst case is one wasted fetch.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-port-forward.sh
jsync-swirlds: We should investigate that (can be a follow-up).
Process kill is a somewhat harsh action and requires root; while kubectl is much milder (it allows the Kubelet to clean up internal tracking, for instance) and may not require root. Generally, if we can do so, we should avoid killing processes that might be managed by daemons that continue after the process is killed.

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
```diff
@@ -26,16 +26,57 @@
...
+        boolean succeeded() {
+            return succeeded;
+        }
+
+        /**
+         * @return the latch released once the block's fate is known
+         */
+        CountDownLatch latch() {
+            return latch;
+        }
```
jsync-swirlds, line 75: Non-blocking
We _could_ hide the latch entirely by offering an `await` method here instead of returning the latch.

jsync-swirlds (dismissed):

jsync-swirlds (approved):

block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPersistenceAwaiter.java
ata-nas: Okay, so long as we don't keep retrying, we are fine, thanks for confirming.

ata-nas (approved):

PR #3572, opened 2026-08-29
chore: Ensure inbound statusz request returns empty if known publishers is empty.

### LINKED ISSUE #3540: chore: fix handling when backfill configured but known-publishers is not.
### Story Form

As a Block Node Operator
I want to upgrade block node without updating configuration
So that my block node upgrades are as simple and reliable as possible

### Technical Notes

When we have some data configured for inbound and outbound connection status but not all data, we need to check plugins and possibly report nothing rather than reporting partially configured data.

jsync-swirlds:
## Reviewer Notes
* Adjusted inbound status handler to return empty `NetworkData` if known publishers is empty.
* Removed one high-frequency, and unnecessary, trace log.

## Related Issue(s)
 Resolves #3540

### DIFF
--- a/block-node/health/src/main/java/org/hiero/block/node/health/InboundStatusHandler.java
+++ b/block-node/health/src/main/java/org/hiero/block/node/health/InboundStatusHandler.java
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.hiero.block.node.health;
 
-import static org.hiero.block.node.health.HttpConnectionSupport.closeAfterHttp1;
-
 import edu.umd.cs.findbugs.annotations.NonNull;
 import io.helidon.http.HeaderNames;
 import io.helidon.webserver.http.ServerRequest;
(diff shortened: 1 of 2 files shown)


block-node/health/src/main/java/org/hiero/block/node/health/InboundStatusHandler.java
```diff
@@ -44,9 +42,11 @@ void createAndSendResponse() {
         endpoints.addAll(appState.knownPublishers().activeEndpoints());
         endpoints.addAll(appState.inboundPartners().activeEndpoints());
         endpoints.addAll(appState.backfillSources().activeEndpoints());
-        final NetworkData data =
-                NetworkData.newBuilder().activeEndpoints(endpoints).build();
-        closeAfterHttp1(request, response.status(200).header(HeaderNames.CONTENT_TYPE, "application/json"))
+        final NetworkData data = appState.knownPublishers().activeEndpoints().isEmpty()
```
ata-nas, line 45: If I understand this correctly, we want to have either all data present, or send nothing if partially configured. Does that mean we have to comprehensively include a check for `knownPublishers`, `inboundPartners` and `backfillSources` to be present instead of only `knownPublishers` to be present? If inbound partners and backfill sources are optional in general, that is understandable, but publishers are optional too? Could you clarify? Maybe I misunderstand something.

jsync-swirlds: Not all-or-nothing.
Just known-publishers, because backfill can be configured without configuring publishers and that would create issues for existing deployments.

A future release should remove this, and add a more complex plugin query process (so different plugins can inform for their own APIs) but that's not currently needed.

ata-nas: We can leave a note for future removal/change in favor of a more robust handling, possibly a link to GH issue?

ata-nas (approved):

block-node/health/src/main/java/org/hiero/block/node/health/InboundStatusHandler.java
jsync-swirlds: I'll put in an issue; probably won't link here (might link from other direction).

berryware (approved):
Looks good

PR #3582, opened 2026-09-01
chore(3552): send acknowledgement after skip when skip sent instead of duplicate

### LINKED ISSUE #3554: Change response to "close" duplicate blocks to "SKIP_AND_ACK" instead of "SKIP"
### Story Form

As a Publisher
I want to receive an Acknowledgement for skipped duplicate blocks
So that I can recover from back-pressure efficiently

### Situation
* When a consensus network pauses due to backpressure, there are race conditions when the network begins to recover.
* Those race conditions can result in "nearby" duplicate blocks that need to be cleared in order for the network to progress the later blocks.
* This combination means publishers might get a series of SKIP responses, but no acknowledgement.

### Desired Change
* Add a new "action" in `LiveStreamPublisherManager` for `SKIP_AND_ACK`.
   * This change is localized [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java#L927-L929).
* This action will send _two_ responses in order, `skip_block`, then `acknowledgement`
* The skip informs the publisher to proceed to the next block, the acknowledgement informs the publisher that the skipped block is acknowledged.  Further, Acknowledgement is comprehensive, so the acknowledgement can enable the publisher to avoid sending further duplicate blocks.
* As part of this task, we may change the default value for `duplicateBlockSkipWindow` to a larger value, to better support rapid catch-up in these situations while maintaining safety for larger gaps.

jsync-swirlds:
## Reviewer Notes
* Added an action for skip-and-ack
* Added a handler block result for skip-and-ack
* Extracted base logic for sending skip and acknowledgement
   * Made both skip and skip-and-ack call the skip method
   * The base acknowledgement doesn't clear queues, so we can minimize impact when acknowledging a duplicate.
* Added a check for trace is-loggable around a costly trace logging block.

## Related Issue(s)
 Resolves #3554.

### DIFF
--- a/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
+++ b/block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/LiveStreamPublisherManager.java
@@ -67,7 +67,6 @@ import org.hiero.block.node.spi.blockmessaging.PublisherStatusUpdateNotification
 import org.hiero.block.node.spi.blockmessaging.PublisherStatusUpdateNotification.UpdateType;
 import org.hiero.block.node.spi.blockmessaging.VerificationNotification;
 import org.hiero.block.node.spi.blockmessaging.VerificationNotification.FailureInfo;
-import org.hiero.block.node.spi.blockmessaging.VerificationNotification.FailureType;
 import org.hiero.block.node.spi.threading.ThreadPoolManager;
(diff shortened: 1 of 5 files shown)


ata-nas (approved):

AlfredoG87 (approved):
LGTM

PR #3585, opened 2026-09-01
docs: fix incorrect port numbers in blockNodeSources examples

deshmukhpranali:
## Reviewer Notes

The peer Block Node query examples in the RSA and TSS bootstrap sections used wrong ports: 40980 (the LFH subscriber port) and 40902 (a unit-test fixture port). Both plugins call serverStatusDetail on peer BNs, so the correct default is the standard gRPC port 40840. Add notes for LFH deployments that expose per-service ports, listing the correct status_port (40982) and subscribe_port (40980) fields from BlockNodeSourceConfig.

### DIFF
--- a/docs/block-node/operations/preparing-your-block-node-for-wrb-cutover.md
+++ b/docs/block-node/operations/preparing-your-block-node-for-wrb-cutover.md
@@ -14,7 +14,7 @@ Complete all steps in this section before every upgrade, regardless of release.
 
 ### Confirm Block Node health
 
-Before upgrading, confirm the Block Node is healthy. Do not upgrade an unhealthy node — an in-progress failure becomes a stuck rollout.
+Before upgrading, confirm the Block Node is healthy. Do not upgrade an unhealthy node - an in-progress failure becomes a stuck rollout.
 
 1. List all pods and confirm the Block Node pod is `Running` with all containers ready:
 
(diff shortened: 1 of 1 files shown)


docs/block-node/operations/preparing-your-block-node-for-wrb-cutover.md
```diff
@@ -251,12 +251,17 @@ The file format is a JSON object with a `nodes` array:
 ```json
 {
   "nodes": [
-    { "address": "peer1.example.com", "port": 40980, "priority": 1, "name": "peer-bn-1" },
-    { "address": "peer2.example.com", "port": 40980, "priority": 2, "name": "peer-bn-2" }
+    { "address": "peer1.example.com", "port": 40840, "priority": 1, "name": "peer-bn-1" },
+    { "address": "peer2.example.com", "port": 40840, "priority": 2, "name": "peer-bn-2" }
```
jsync-swirlds: It makes sense to set `"subscribe_port": 40980`. `port` isn't really used and hopefully we'll be able to deprecate and remove it in the future.
It also helps to include `"status_port": 40982`.

docs/block-node/operations/preparing-your-block-node-for-wrb-cutover.md
```diff
@@ -385,10 +390,10 @@ Set `roster.bootstrap.tss.blockNodeSourcesPath` (default: `""`) to the path of a
...
   "nodes": [
-    { "address": "peer1.example.com", "port": 40902, "priority": 1, "node_id": 1, "name": "peer-bn-1" },
+    { "address": "peer1.example.com", "port": 40840, "priority": 1, "node_id": 1, "name": "peer-bn-1" },
     {
       "address": "peer2.example.com",
-      "port": 40902,
+      "port": 40840,
       "priority": 2,
       "node_id": 2,
       "name": "peer-bn-2",
```
jsync-swirlds, line 400: Here too, I think we should keep the new ports.

Peer BNs don't use different ports from any other client, so the ports need to be the same.

jsync-swirlds (approved):

PR #3589, opened 2026-09-02
chore: Remote Block Node Flavors in E2E

### LINKED ISSUE #2127: Remote Block Node Flavors in E2E
### Story Form

As a block node engineer
I want to test remote BN flavors in E2E deployments
So that I can catch integration bugs with different BN deployment types

### Technical Notes

Extend E2E test coverage to include remote Block Node flavors and deployment configurations. Different BN flavors may have subtle integration differences that only surface in E2E testing.

## Goals

- Validate remote BN flavors work correctly in E2E environment
- Catch integration bugs specific to BN deployment types
- Ensure consistency across BN flavors
- MN receives blocks from all BN flavors correctly

## Implementation Notes

Can be implemented using existing test framework:
- Extend topology schema to support BN flavor specification
- Deploy different BN flavors in test topologies
- Use existing assertions to validate behavior
- May need configuration for remote BN endpoints

**Dependency:** Requires bucket configuration (S3-compatible storage)
- Can use MinIO or reuse the bucket Solo already creates for CN
- Bucket dependency needs to be sorted before remote BN flavors can be tested
- Configuration for bucket endpoint, credentials, and bucket name

## Acceptance Criteria

- [ ] Topology supports BN flavor configuration
- [ ] Remote BN flavors deploy successfully in E2E
- [ ] Blocks flowing verified for each flavor
- [ ] No flavor-specific integration bugs

ivannov:
## Reviewer Notes

* Adds a `flavor:` field to the topology schema so each Block Node can be deployed with a named plugin profile
* Renames the cloud-only plugin profile to `rfh`
* Emits the roster-bootstrap TSS sources path alongside the backfill sources path for any Block Node with peers, so backfilled blocks can be verified
* Adds a `max_verify_failed` budget to the `no-errors` assertion to absorb TSS warm-up failures
* Adds the `2cn-3bn-flavors` topology and `bn-flavors` test covering the lfh, minimal and all profiles
* Wires the new topology and test into the scheduled E2E matrix

## Related Issue(s)

Closes #2127

### DIFF
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -82,7 +82,7 @@ on:
         type: string
         default: "5"
       test-definition:
-        description: "Test definitions (comma-separated): smoke-test, basic-load, node-restart-resilience, full-history-backfill, tss-signature-transition, wrb-cli-hash-validation, wrb-cli-wrapping-validation, wrb-differential-test, wrb-core-validation, rsa-roster-verification, wrb-distribution-step12, wrb-distribution-step345, wrb-distribution-step89, wrb-distribution-step10, chaos-foundation-smoke, latency-cn-to-cn, latency-cn-to-bn, latency-bn-to-bn, latency-all-three, latency-stress, latency-severe"
+        description: "Test definitions (comma-separated): smoke-test, basic-load, high-load, node-restart-resilience, full-history-backfill, archive-backfill, bn-flavors, tss-signature-transition, wrb-cli-hash-validation, wrb-cli-wrapping-validation, wrb-cli-jumpstart-validation, wrb-differential-test, wrb-core-validation, rsa-roster-verification, wrb-distribution-step12, wrb-distribution-step345, wrb-distribution-step67, wrb-distribution-step89, wrb-distribution-step10, chaos-foundation-smoke, latency-cn-to-cn, latency-cn-to-bn, latency-bn-to-bn, latency-all-three, latency-stress, latency-severe"
         type: string
         default: "none"
     outputs:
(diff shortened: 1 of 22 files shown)


.github/workflows/solo-e2e-test.yml
```diff
@@ -104,6 +104,8 @@ on:
         options:
           - single # 1 CN, 1 BN
           - 2cn-2bn-backfill # 2 CN, 2 BN with backfill
+          - 2cn-2bn-archive # 2 CN streaming to BN2; BN1 backfills from BN2 and archives to RustFS
```
Nana-EC, line 107: I'm not sure about these 2 new names. They don't give insight into what the actual topology. Of course descriptions serve a purpose but names with insight are also useful.
I'll think on it also

tools-and-tests/scripts/solo-e2e-test/tests/bn-flavors.yaml
```diff
@@ -0,0 +1,162 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# BN Plugin Flavors E2E Test
```
Nana-EC: Should be profiles to be consistent across repo.
Should also be plugin-profiles.yaml
```suggestion
# BN Plugin Profiles E2E Test
```

.github/workflows/solo-e2e-test.yml
```diff
@@ -104,6 +104,8 @@ on:
...
+          - 2cn-3bn-flavors # 2 CN, 3 BN exercising the lfh, minimal and all plugin flavors
```
Nana-EC:
```suggestion
          - 2cn-3bn-plugin-profiles # 2 CN, 3 BN exercising the lfh, minimal and all plugin profiles
```

Nana-EC (commented):
Coverage is good.
Naming is hard though so the `flavors` reference was a little hard. I'd suggest using `profiles` to be consistent across repo

.github/workflows/solo-e2e-test.yml
ivannov: I renamed `2cn-3bn-flavors` to `2cn-3bn-plugin-profiles` per your suggestion below. Leaving `2cn-2bn-archive` as is for now since it follows the `2cn-2bn-backfill` pattern

.github/workflows/solo-e2e-test.yml
ivannov: Applied

tools-and-tests/scripts/solo-e2e-test/tests/bn-flavors.yaml
ivannov: Applied. Also renamed the file

Nana-EC (approved):
LG from a high level. If it runs I'm good.
Another place for @AlfredoG87 to review 😅

PR #3591, opened 2026-09-02
chore: Live Active Session Buffer Size Metric and Evicted Sessions Log in Verification

### LINKED ISSUE #3590: Live Active Session Buffer Size Metric and Evicted Sessions Log in Verification
### Story Form

As a Block Node operator
I want to be aware of the live state of active sessions buffer and also be aware of evicted sessions
So that I can track those

### Technical Notes

- We want to add a metric that will show us live the state of the active sessions buffer in the session handler
- We want to add an INFO log for cancelled notifications in the result handler

ata-nas:
* Add verification_active_sessions gauge tracking the live size of the active sessions buffer
* Add INFO log for cancelled verification notifications in the result handler
  * Demote the stage-reported cancellation log from WARNING to INFO, expected evictions must not page on-call
* Add tests
  * Gauge follows session activation, eviction, and overfull buffer
* Document the new gauge in metrics.md

Resolves #3590

### DIFF
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
@@ -123,6 +123,7 @@ public final class BlockSessionHandler {
                 }
             }
         }
+        updateActiveSessionsMetric();
     }
 
     /// Process the reception of live blocks from the publisher. Publisher supplied [BlockItems] can only
@@ -231,5 +232,11 @@ public final class BlockSessionHandler {
                 activePublisherSession.compareAndSet(removed, null);
             }
         }
+        updateActiveSessionsMetric();
+    }
+
+    /// Update the active sessions gauge with the current size of the active sessions buffer.
+    private void updateActiveSessionsMetric() {
+        sessionHandlerMetrics.verificationActiveSessions().set(activeSessions.size());
     }
 }
--- a/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
+++ b/block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
@@ -123,7 +123,7 @@ public final class BlockSessionHandler {
                 }
             }
         }
-        updateActiveSessionsMetric();
+        sessionHandlerMetrics.verificationActiveSessions().set(activeSessions.size());
     }
 
     /// Process the reception of live blocks from the publisher. Publisher supplied [BlockItems] can only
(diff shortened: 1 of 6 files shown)


block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
```diff
@@ -231,5 +232,11 @@ private void activateSession(final BlockVerificationSession session) {
                 activePublisherSession.compareAndSet(removed, null);
             }
         }
+        updateActiveSessionsMetric();
+    }
+
+    /// Update the active sessions gauge with the current size of the active sessions buffer.
+    private void updateActiveSessionsMetric() {
+        sessionHandlerMetrics.verificationActiveSessions().set(activeSessions.size());
     }
```
jsync-swirlds: We shouldn't wrap metrics in a method; it's extra overhead for what should be extremely low impact.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
```diff
@@ -167,47 +174,30 @@ private boolean handle(final BlockVerificationResult verificationResult, final T
...
-        if (throwable instanceof CompletionException ce
-                && ce.getCause() instanceof VerificationSessionFailedException vfe) {
-            hapiVersion = vfe.getHapiVersion();
-            blockItems = vfe.getBlockItems();
-        } else {
-            hapiVersion = null;
-            blockItems = null;
-        }
+        SemanticVersion hapiVersion = null;
+        List<BlockItemUnparsed> blockItems = null;
```
jsync-swirlds, line 186: These appear to be assigned, but never used.
Can they be deleted entirely?

jsync-swirlds (commented):
Looks good, just two small items.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
ata-nas: Look at the `finally` clause, they are passed, nullable, to the bad block dumper.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/BlockSessionHandler.java
ata-nas: Sure, let me update.

ata-nas: ✅ Done, take a look.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/session/SessionResultHandler.java
jsync-swirlds: Thanks, did not see them there.

jsync-swirlds (approved):

Nana-EC (approved):

PR #3611, opened 2026-09-04
fix: tss-signature-transition runs after tests that deliberately kill block-node-1

### LINKED ISSUE #3603: `tss-signature-transition` runs after tests that deliberately kill `block-node-1`
### Story Form

As a BlockNode engineer,
I want `tss-signature-transition` to run only against a topology that is still intact,
So that a red job does not mean that the previous test killed the node it targets.

### Technical Notes

- `solo-e2e-scheduler.yml:175`: `{"topology":"2cn-2bn-archive","tests":"archive-backfill,tss-signature-transition"}`
- `archive-backfill` test kills BN1 by design and never restores it. Its own assertions target only BN2/BN3
- `tss-signature-transition` targets `all` and `block-node-1`
- Line 174 is the same shape: `full-history-backfill` wipes BN1 first. Noted as a cascade in #3542
- The dead target then hung the job 5h32m: `grpcurl` (`monitor-block-proofs.sh:62`) has no `-max-time` and stalls on a Service with no endpoints

ivannov:
## Reviewer Notes

* Runs the `tss-signature-transition` test before the backfill tests in topologies where a Block Node is killed, so it no longer waits on a node that is already gone
* Bounds the block proof monitor with a per-call gRPC deadline
* Adds step timeouts to the deploy and test steps of the Solo E2E workflow

## Related Issue(s)

Closes #3603

### DIFF
--- a/.github/workflows/solo-e2e-scheduler.yml
+++ b/.github/workflows/solo-e2e-scheduler.yml
@@ -159,7 +159,7 @@ jobs:
               # Quick validation - single topology smoke test + backfill test
               MATRIX='[
                 {"topology":"single","tests":"smoke-test,tss-signature-transition,wrb-cli-hash-validation,wrb-cli-wrapping-validation"},
-                {"topology":"2cn-2bn-backfill","tests":"full-history-backfill,tss-signature-transition"},
+                {"topology":"2cn-2bn-backfill","tests":"tss-signature-transition,full-history-backfill"},
                 {"topology":"single-wrb-rsa","tests":"smoke-test,rsa-roster-verification"},
                 {"topology":"wrb-differential-test","tests":"wrb-differential-test"}
               ]'
@@ -171,8 +171,8 @@ jobs:
                 {"topology":"paired-3","tests":"smoke-test,basic-load,tss-signature-transition"},
                 {"topology":"3cn-1bn","tests":"smoke-test,tss-signature-transition"},
                 {"topology":"fan-out-3cn-2bn","tests":"smoke-test,tss-signature-transition"},
-                {"topology":"2cn-2bn-backfill","tests":"full-history-backfill,tss-signature-transition"},
-                {"topology":"2cn-2bn-archive","tests":"archive-backfill,tss-signature-transition"},
+                {"topology":"2cn-2bn-backfill","tests":"tss-signature-transition,full-history-backfill"},
+                {"topology":"2cn-2bn-archive","tests":"tss-signature-transition,archive-backfill"},
                 {"topology":"7cn-3bn-distributed","tests":"smoke-test,tss-signature-transition"},
                 {"topology":"single-wrb-rsa","tests":"smoke-test,rsa-roster-verification"},
                 {"topology":"3cn-2bn-wrb-rsa","tests":"smoke-test,rsa-roster-verification"},
--- a/.github/workflows/solo-e2e-scheduler.yml
+++ b/.github/workflows/solo-e2e-scheduler.yml
@@ -172,7 +172,7 @@ jobs:
                 {"topology":"3cn-1bn","tests":"smoke-test,tss-signature-transition"},
                 {"topology":"fan-out-3cn-2bn","tests":"smoke-test,tss-signature-transition"},
                 {"topology":"2cn-2bn-backfill","tests":"tss-signature-transition,full-history-backfill"},
-                {"topology":"2cn-2bn-archive","tests":"tss-signature-transition,archive-backfill"},
+                {"topology":"2cn-2bn-archive","tests":"archive-backfill"},
                 {"topology":"7cn-3bn-distributed","tests":"smoke-test,tss-signature-transition"},
                 {"topology":"single-wrb-rsa","tests":"smoke-test,rsa-roster-verification"},
                 {"topology":"3cn-2bn-wrb-rsa","tests":"smoke-test,rsa-roster-verification"},
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -440,6 +440,7 @@ jobs:
 
       # Deploy network using shared script
       - name: Deploy network
+        timeout-minutes: 20
         env:
           # Empty unless the Consensus Node was built from source above, in which case Solo
           # is pointed at those jars instead of downloading a (nonexistent) SNAPSHOT zip.
@@ -596,6 +597,7 @@ jobs:
       # Run each test definition sequentially on the same deployment
       - name: Run Test Definitions
         id: tests_end
+        timeout-minutes: 75
         if: ${{ inputs.test-definition != 'none' && inputs.test-definition != '' }}
         run: |
           TESTS_DIR="${{ github.workspace }}/tools-and-tests/scripts/solo-e2e-test/tests"
--- a/.github/workflows/solo-e2e-test.yml
+++ b/.github/workflows/solo-e2e-test.yml
@@ -597,7 +597,7 @@ jobs:
       # Run each test definition sequentially on the same deployment
       - name: Run Test Definitions
         id: tests_end
-        timeout-minutes: 75
+        timeout-minutes: 90
         if: ${{ inputs.test-definition != 'none' && inputs.test-definition != '' }}
         run: |
           TESTS_DIR="${{ github.workspace }}/tools-and-tests/scripts/solo-e2e-test/tests"
--- a/tools-and-tests/scripts/solo-e2e-test/scripts/monitor-block-proofs.sh
+++ b/tools-and-tests/scripts/solo-e2e-test/scripts/monitor-block-proofs.sh
@@ -43,6 +43,40 @@ STEP=50
 # before the data flood kills the port-forward. Rejected blocks are treated as
 # NOT_AVAILABLE, which the binary search handles correctly.
 MAX_MSG_SIZE=10000000
+# Per-call wall-clock ceiling for grpcurl. A Service whose pods are gone still
+# accepts a kubectl port-forward connection and then stalls forever, so without
+# this a single call can block the whole script past the max_block*2
+# deadline before check_hard_timeout gets a chance to run. 60s is well clear of a legitimate
+# fetch: observed blocks are ~40KB, the largest seen 983.9KB, and MAX_MSG_SIZE
+# caps a download at 10MB. On expiry grpcurl reports "context deadline
+# exceeded", which the CONNECTION_ERROR pattern below already handles.
+MAX_CALL_SECONDS=60
+
+# Hard timeout: max_block * 2s (block production interval).
+# If max_block=1000, timeout=2000s (~33 min). 0 means no timeout.
+if [[ "$MAX_BLOCK" -gt 0 ]]; then
+  HARD_TIMEOUT=$(( MAX_BLOCK * 2 ))
+else
+  HARD_TIMEOUT=0
+fi
+START_TIME=$(date +%s)
+
+# Checked from get_sig_type_safe (every gRPC call site: scan loop, binary
+# search, lower-bound search, print_result) rather than only between scan
+# iterations, so post-detection work can't run unbounded past the deadline.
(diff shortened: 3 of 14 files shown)


AlfredoG87 (changes requested):
The ordering fix is right and the three layer approach makes sense. The issue is that two
of the new safety nets misreport the failures they catch. I reproduced both locally with
a fake `grpcurl` and no cluster, details below.

Worth noting none of this has run yet. `solo-e2e-test.yml` is `workflow_call` only and the
scheduler is schedule/tag/dispatch, so nothing here is PR triggered, including
`run-all.sh` at `solo-e2e-test.yml:406` which is where the new suite would run. Could you
kick off a `workflow_dispatch` of the scheduler against this branch before merge? The
reorder is the part I most want to see green on a real topology.

Actions tab > Solo E2E Scheduler > Run workflow, then set "Use workflow from" to
`3603-remove-tss-test`, `run-type` to `weekend`, and `notify-slack` off. Or:

```
gh workflow run solo-e2e-scheduler.yml --ref 3603-remove-tss-test \
  -f run-type=weekend -f notify-slack=false
```

`daily` is cheaper but only covers `2cn-2bn-backfill`. `weekend` is the full matrix and
also picks up `2cn-2bn-archive`, which is the one that actually hung for 5h32m, so that
is the entry worth seeing green.

**Update: I have already run it so you can take a look when you see this.**
https://github.com/hiero-ledger/hiero-block-node/actions/runs/33927189342

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
```diff
@@ -1380,17 +1389,26 @@ function assert_signature_transition {
...
 
     local pf_cmd="${SCRIPT_DIR}/solo-port-forward.sh --namespace ${NAMESPACE}"
-    local output
-    if output=$("$script" "${PROTO_PATH}" "localhost:${port}" "$max_block" "${pf_cmd}" 2>&1); then
+    local output status
+    local hard_timeout=0
+    if [[ "$max_block" -gt 0 ]]; then
+        hard_timeout=$(( max_block * 2 + SIGNATURE_TRANSITION_GRACE_SECONDS ))
+    fi
+    if output=$(timeout --foreground "${hard_timeout}" "$script" "${PROTO_PATH}" "localhost:${port}" "$max_block" "${pf_cmd}" 2>&1); then
```
AlfredoG87, line 1397: [blocking] `timeout` is GNU coreutils, macOS has neither `timeout` nor `gtimeout`. So this returns 127 and `assert_signature_transition` fails on every local run, and `task test:unit` goes red. On my machine:

```
$ scripts/test/test-signature-transition.sh
  FAIL  names the timeout rather than 'WRAPS not detected'
  FAIL  reports 'WRAPS not detected' with the exit status, not a timeout
  Passed: 3, Failed: 2
```

Same suite with a GNU compatible `timeout` shim on PATH passes 5/5, so the logic is fine, it is only the missing binary. CI is Linux so this will not show up in the checks, same as the `declare -A` case.

Suggest resolving `timeout`/`gtimeout` once and running the script unwrapped when neither exists, the monitor self bounds anyway. Plus a line in `task check` and the README prerequisites table.

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
```diff
@@ -1380,17 +1389,26 @@ function assert_signature_transition {
...
         transition_summary=$(echo "${output}" | grep -A 10 "=== Signature Transition ===")
         transition_block=$(echo "${transition_summary}" | grep "First WRAPS block:" | awk '{print $NF}')
         echo "${target}: Schnorr -> WRAPS at block ${transition_block:-unknown}"
         echo "${transition_summary}"
         return 0
     else
+        status=$?
         echo "${output}" >&2
-        echo "${target}: WRAPS not detected within ${max_block} blocks"
+        if [[ "${status}" -eq 124 ]]; then
```
AlfredoG87, line 1407: [blocking] This branch never fires for the deadline it was written for. `check_hard_timeout` ends with `kill -TERM "$$"`, so the script exits 143, not 124. Reproduced with `max_block=50` (100s script deadline) and a 160s wrapper, which is the ordering your comment at monitor-block-proofs.sh:42 describes as normal:

```
monitor-block-proofs.sh alone:   exit 143, elapsed 100s
  stderr: Status: WRAPS NOT DETECTED / Hard timeout: 100s elapsed

via assert_signature_transition:
  block-node-1: WRAPS not detected within 50 blocks (exit 143)
```

So the expected path gets the misleading message and the 124 branch only fires in the case you say should be rare. Either add 143 here, or trap TERM in the monitor and exit a code the runner can name.

tools-and-tests/scripts/solo-e2e-test/scripts/monitor-block-proofs.sh
```diff
@@ -138,6 +173,7 @@ print(f'{sig_bytes} {binary_size}')
 #   3. If retry also fails (oversized block crashing port-forward), returns NOT_AVAILABLE
 # Without PORT_FORWARD_CMD, CONNECTION_ERROR is returned as-is for the caller to handle.
 function get_sig_type_safe {
+  check_hard_timeout
```
AlfredoG87, line 176: [blocking] Moving `check_hard_timeout` into `get_sig_type_safe` means a successful detection can report as a failure. `binary_search_transition`, `find_schnorr_lower_bound` and `print_result` all go through here, so the deadline can now lapse after WRAPS was already found and printed. Same inputs, your branch vs main:

```
this PR (max_block=50, 20s/call):    main:
  Block 50: WRAPS (3432 bytes)         Block 50: WRAPS (3432 bytes)
  WRAPS detected, binary searching...  WRAPS detected, binary searching...
  Status: WRAPS NOT DETECTED           First WRAPS block: 1
  exit 143 (102s)                      exit 0 (185s)
```

Bounding the post detection phase is the right call, leaving it unbounded is the bug you are fixing. But once the transition is observed the abort should report the block it found instead of denying it, otherwise a run that proved WRAPS arrived gets filed as a run that proved it did not.

tools-and-tests/scripts/solo-e2e-test/scripts/test/test-signature-transition.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#
+# Fixture-based unit test for assert_signature_transition in solo-test-runner.sh:
```
AlfredoG87, line 4: [suggestion] Nothing covers `check_hard_timeout` itself, which is the trickiest new piece, `kill -TERM "$$"` escaping nested `$(...)`. The suite only tests the outer wrapper. A fixture test there would have caught the 143 vs 124 thing above. I did it with a fake `grpcurl` on PATH and no cluster if that helps, happy to hand it over.

tools-and-tests/scripts/solo-e2e-test/scripts/test/test-signature-transition.sh
```diff
@@ -0,0 +1,78 @@
...
+# Any existing directory satisfies validate_proto_path; the fake monitor ignores it.
+# shellcheck disable=SC2034 # consumed by the sourced runner
+PROTO_PATH="${SCRIPT_DIR}"
+# max_block 1 plus 1s of grace gives a 3s ceiling.
+# shellcheck disable=SC2034 # consumed by the sourced runner
+SIGNATURE_TRANSITION_GRACE_SECONDS=1
+
+passed=0
+failed=0
+function record {
```
AlfredoG87, line 32: [suggestion] `record` here takes a precomputed boolean, `test-archive-assertions.sh` takes `name expected actual` and prints both. On failure this one prints nothing useful, when it failed on my machine I got `FAIL  names the timeout rather than 'WRAPS not detected'` and had to reconstruct the actual output by hand.

.github/workflows/solo-e2e-test.yml
```diff
@@ -596,6 +597,7 @@ jobs:
       # Run each test definition sequentially on the same deployment
       - name: Run Test Definitions
         id: tests_end
+        timeout-minutes: 75
```
AlfredoG87, line 600: [suggestion] Minor, and only about diagnostics. Numbers below are from main, the branch had no Solo E2E run when I looked, but the two internal deadlines are unchanged by this PR so it should carry over. Healthy `2cn-2bn-backfill` runs take 23 to 24 min so 75 is plenty of headroom. But the failing runs land at 75, 76 and 77 min because both tests burn their full internal deadlines (`HEIGHT_TIMEOUT` 2400s, then `max_block*2` 2000s). At 75 we cut those off right at the end and lose the tss failure line. 80 to 85 would let the script's own message through, which is the more useful output.

.github/workflows/solo-e2e-scheduler.yml
```diff
@@ -171,8 +171,8 @@ jobs:
                 {"topology":"paired-3","tests":"smoke-test,basic-load,tss-signature-transition"},
                 {"topology":"3cn-1bn","tests":"smoke-test,tss-signature-transition"},
                 {"topology":"fan-out-3cn-2bn","tests":"smoke-test,tss-signature-transition"},
-                {"topology":"2cn-2bn-backfill","tests":"full-history-backfill,tss-signature-transition"},
-                {"topology":"2cn-2bn-archive","tests":"archive-backfill,tss-signature-transition"},
+                {"topology":"2cn-2bn-backfill","tests":"tss-signature-transition,full-history-backfill"},
```
AlfredoG87, line 174: [suggestion] Not for this PR, and about `tests/tss-signature-transition.yaml` rather than this line, but this is the line it protects. The ordering fix is a convention and nothing stops the next matrix edit from reintroducing it. `all-healthy` already targets `all`, and `assert_node_healthy_single` returns "Pod not found" for a scaled down BN, so the dead BN1 is detectable. If assertions short circuited on that, or `wraps-transition` depended on it, the hang could not come back regardless of ordering. Maybe a follow up to #3542.

AlfredoG87 (commented):
Per-line detail for the review above.

.github/workflows/solo-e2e-test.yml
ivannov: 2000s+180s then 2400s is around 73 - 74 minutes before setup, so going to 90 rather than 85

tools-and-tests/scripts/solo-e2e-test/scripts/test/test-signature-transition.sh
ivannov: Added a test fixture for that

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
ivannov: Correct — all nine tss jobs printed (exit 143)

Took your second option: `trap 'exit 3' TERM` in the monitor. And then 3 handled next to 124 in the runner

tools-and-tests/scripts/solo-e2e-test/scripts/solo-test-runner.sh
ivannov: Changed the tests to skip if timeout or gtimeout are not present

tools-and-tests/scripts/solo-e2e-test/scripts/test/test-signature-transition.sh
ivannov: Agreed. Matching `test-archive-assertions.sh` and passing the real values.

.github/workflows/solo-e2e-scheduler.yml
ivannov: `assert_signature_transition` now pre-flights its target and fails in seconds instead of waiting out max_block * 2

Nana-EC (approved):
Unblocking

AlfredoG87 (approved):
Unblocked

PR #3626, opened 2026-09-08
docs(block-node): add Kubernetes resource sizing guidance for Block Node

### LINKED ISSUE #3620: docs: add Kubernetes resource sizing guidance for Block Node deployments
The hardware specifications document covers VM sizing but provides no Kubernetes-specific guidance (container resource requests, limits, or JVM heap configuration). Operators deploying in Kubernetes have no reference for what to configure and may experience OOM pod restarts.

**Required coverage:**
- Minimum container resource requests and limits (CPU, memory) for dev/testnet/previewnet
- JVM heap configuration and its relationship to container memory limit
- How to detect and fix OOM restarts in a BN pod

**Needs from team:** Engineering to confirm recommended Kubernetes resource values.

**Source:** Operator feedback (September 2026).

deshmukhpranali:
## Reviewer Notes

Documents the default Helm chart resource/JVM mismatch that OOM-kills pods at startup (15Gi container limit vs -Xms16G/-Xmx16G heap) and provides correct override values with the blockNode.config.JAVA_OPTS key path, a 75-90% heap-to-limit sizing rule, and kubectl commands for diagnosing OOM-killed pods.

Verified against values.yaml, configmap.yaml template, helm template dry-run, and all values-overrides profiles (mini, mid-size, LFH).

## Related Issue(s)
Closes #3620

### DIFF
--- a/docs/block-node/operations/block-node-hardware-specifications.md
+++ b/docs/block-node/operations/block-node-hardware-specifications.md
@@ -11,8 +11,8 @@ deployments.
 
 |                       Node type                        |        Network        |                                  See section                                  |
 |--------------------------------------------------------|-----------------------|-------------------------------------------------------------------------------|
-| Tier 1 — receives stream directly from Consensus Nodes | Mainnet               | [Tier 1 Mainnet Server Specifications](#tier-1-mainnet-server-specifications) |
-| Tier 2 — receives stream from another Block Node       | Mainnet               | [Tier 2 Server Specifications](#tier-2-server-specifications)                 |
+| Tier 1 - receives stream directly from Consensus Nodes | Mainnet               | [Tier 1 Mainnet Server Specifications](#tier-1-mainnet-server-specifications) |
+| Tier 2 - receives stream from another Block Node       | Mainnet               | [Tier 2 Server Specifications](#tier-2-server-specifications)                 |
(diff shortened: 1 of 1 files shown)


docs/block-node/operations/block-node-hardware-specifications.md
```diff
@@ -169,6 +169,49 @@ Applying the block size model at lower TPS:
...
+|------------|---------|
+| `resources.requests.cpu` | `4` |
+| `resources.requests.memory` | `12Gi` |
+| `resources.limits.cpu` | `4` |
+| `resources.limits.memory` | `15Gi` |
+| `blockNode.config.JAVA_OPTS` (heap flags) | `-Xms16G -Xmx16G` |
+
+> **Warning:** The default `JAVA_OPTS` sets a 16 GiB heap (`-Xms16G -Xmx16G`), which exceeds
+> the default 15 GiB memory limit. A pod using these defaults will be OOM-killed at startup.
+> Always override both `resources.limits.memory` and `JAVA_OPTS` together before deploying.
```
jsync-swirlds, line 190: So rather than documenting a bug, what we need to do is fix this bug.
The helm chart is incorrect.
12/15 is a required setting to fit within a defined hardware requirement.
the -Xms/-Xmx values should be modified to 10G for this profile, and a metaspace flag should be added if not already present.

jsync-swirlds (commented):
This looks like documenting a bug; I suspect we should have filed the bug issue instead.

PR #3629, opened 2026-09-09
fix: raise grpcurl receive limit on status calls in bn-endpoint-checker

rockysingh:
## Summary

- `grpc_call()` (used for `serverStatus` and `serverStatusDetail`) inherited grpcurl's 4 MiB default receive limit.
- On nodes with a large number of archived block ranges (observed on mainnet at 99M+ blocks), `serverStatusDetail` responses exceed 4 MiB and fail with `ResourceExhausted`.

## Observed failure

Mainnet Tier-0 Block Node (chart 0.41.0, ~99.6M blocks archived):

```
serverStatusDetail WARN (method unavailable or proxy limitation)  (603 ms)
     ERROR:
       Code: ResourceExhausted
       Message: grpc: received message larger than max (4338533 vs. 4194304)
```

After the fix, `serverStatusDetail` returns normally on mainnet and the WARN clears.

### DIFF
--- a/tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
+++ b/tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
@@ -70,6 +70,14 @@
 #                           blocks (17+ MiB observed), causing a
 #                           ResourceExhausted error. Must be a positive integer
 #                           between 1 and 100 (inclusive). Defaults to 20.
+#       --max-response-sz MIB
+#                           Maximum gRPC response size in MiB accepted from
+#                           BlockNodeService calls (serverStatus,
+#                           serverStatusDetail). grpcurl's built-in default is
+#                           4 MiB, which is too small for serverStatusDetail on
+#                           nodes with many archived block ranges (99M+ blocks
+#                           observed on mainnet). Must be a positive integer
+#                           between 1 and 512 (inclusive). Defaults to 64.
 #   -h, --help              Print this help text and exit.
 #
 # ENVIRONMENT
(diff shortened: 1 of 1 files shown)


tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
```diff
@@ -519,6 +519,13 @@ grpc_call() {
     -connect-timeout 5
     -import-path "${RESOLVED_PROTO_DIR}"
     -proto        "${NODE_SERVICE_PROTO}"
+    # grpcurl defaults to a 4 MiB receive limit. serverStatusDetail on nodes
+    # with a large number of archived block ranges (observed on mainnet with
+    # 99M+ blocks) can exceed that, causing the call to fail with
+    # ResourceExhausted, which the caller surfaces as a misleading
+    # "method unavailable or proxy limitation" warning. 64 MiB is a safe
+    # upper bound for the status responses seen in the wild.
+    -max-msg-sz   67108864
```
Nana-EC: Nice catch.
You'll want to expose this as a flag on the script.
I added a specific flag for blocks at https://github.com/hiero-ledger/hiero-block-node/blob/main/tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh#L546
For this maybe it's a general flag for grpcCall responses. I'd suggest `--max-response-sz`

Nana-EC (commented):
I would suggest exposing a `--max-response-sz` flag that configures the value so that a caller can manage future response sizes as needed

tools-and-tests/scripts/node-operations/bn-endpoint-checker.sh
rockysingh: Done in c9e46662. Exposed as `--max-response-sz` (default 64 MiB, range 1 to 512).

AlfredoG87 (approved):
LGTM

ata-nas (approved):

PR #3609, opened 2026-09-03
feat: Add Application State notifications for the Application State Facility

berryware:
## Reviewer Notes

Add the ApplicationStateNotifications in to BlockMessaging.

**Step 1 — 4 new notification records** (all in `spi-plugins`, `blockmessaging` package):
- [`TssDataNotification.java`](block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/TssDataNotification.java)
- [`AddressBookHistoryNotification.java`](block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/AddressBookHistoryNotification.java)
- [`StoredBlocksNotification.java`](block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/StoredBlocksNotification.java)
- [`AvailableBlocksNotification.java`](block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/AvailableBlocksNotification.java)

**Step 2 — `ApplicationStatetNotificationHandler` interface** ([`ApplicationStateNotificationHandler.java`](block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/ApplicationStateNotificationHandler.java)): extends `GatingHandler`, four default no-op methods.

**Step 3 — Disruptor wiring**:
- [`BlockNotificationRingEvent`](block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockNotificationRingEvent.java) — 4 new nullable fields, 4 new `set()`/`get()` pairs; every `set` nulls all 9 fields.
- [`BlockMessagingFacility`](block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/blockmessaging/BlockMessagingFacility.java) — 6 new methods: 4 sends + register/unregister.
- [`BlockMessagingFacilityImpl`](block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java) — full implementation: 4 metrics, 2 handler maps, pre-registration list, 4 send methods (via `messageForwarder` → existing ring), `register`/`unregister`, `start`/`stop` updates, fixed spurious "no notification set" log for app state events.
- [`TestBlockMessagingFacility`](block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/TestBlockMessagingFacility.java) — synchronous stub implementation of all 6 new methods.
[...]

PR #3631, opened 2026-09-09
feat: Wire Up the Application State Notifications and the ApplicationStateNotificationHandlers

### LINKED ISSUE #3607: F2 - Wire Up the Application State Notifications and the ApplicationStateNotificationHandlers
### Story Form

As a BlockNode Engineer
I want to performant state change notifications.

### Step 4 — Move change detection into the 3 interface update methods on `BlockNodeApp`

**`updateTssData(TssData newTssData)`** — currently enqueues into `tssDataUpdates` queue for the scan loop to drain. Replace with direct compare-and-dispatch:

```java
@Override
public void updateTssData(TssData newTssData) {
    if (newTssData == null) return;
    TssData current = blockNodeContext.tssData();
    if (Objects.equals(current, newTssData)) return;
    // Update context snapshot
    blockNodeContext = new Builder(blockNodeContext).tssData(newTssData).build();
    blockMessaging.sendTssDataUpdate(new TssDataNotification(newTssData));
}
```

Remove `tssDataUpdates` queue and its drain in `checkForApplicationStateUpdates`.

**`updateAddressBookHistory(RangedAddressBookHistory history)`** — already has a comparison (`isNewerHistory`). Keep that guard, then after updating the context snapshot, dispatch:

```java
blockNodeContext = new Builder(blockNodeContext).rangedAddressBookHistory(history).build();
blockMessaging.sendAddressBookHistoryUpdate(new AddressBookHistoryNotification(history));
```

Remove `pendingAddressBookHistory` AtomicReference and its drain in the scan loop.

**`addStoredBlockRange(LongRange blockRange)`** — currently just calls `storedBlocks.add(blockRange)`. After adding, compute the merged snapshot and dispatch:

```java
@Override
public void addStoredBlockRange(LongRange blockRange) {
    storedBlocks.add(blockRange);
    List<BlockRange> merged = mergeRanges(storedBlocks, historicalBlockFacility.availableBlocks());
    blockNodeContext = new Builder(blockNodeContext).storedBlocks(merged).build();
    blockMessaging.sendStoredBlocksUpdate(new StoredBlocksNotification(merged));
}
```

---

Clean direction. Here's the revised Step 5, replacing the signal/callback approach:

---

### Step 5 (revised) — Add `updateAvailableBlocks` to `ApplicationStateFacility`

Add one update method to the `ApplicationStateFacility` interface, matching the pattern of the other three fields:

```java
void updateAvailableBlocks(BlockRangeSet availableBlocks);
```

**Provider plugins** (`BlockFileRecentPlugin`, `BlockFileHistoricPlugin`) call `context.applicationStateFacility().updateAvailableBlocks(...)` wherever they currently mutate their own `ConcurrentLongRangeSet` directly — on block receipt, startup scan, and eviction.

**`BlockNodeApp`** implements this method. Since multiple providers each report their own set, `BlockNodeApp` maintains a per-provider map keyed by `BlockProviderPlugin` identity to aggregate the union. On each call it recomputes the union, compares with the current context value, and if changed:

```java
blockNodeContext = new Builder(blockNodeContext).availableBlocks(updated).storedBlocks(merged).build();
blockMessaging.sendAvailableBlocksUpdate(new AvailableBlocksNotification(updated));
blockMessaging.sendStoredBlocksUpdate(new StoredBlocksNotification(merged));  // merged view also changes
```

**`HistoricalBlockFacilityImpl`** — if it needs to update its internal routing table or `CombinedBlockRangeSet` view in response to available-block changes — implements `ContextNotificationHandler`, registers for `AvailableBlocksNotification` during `init()`, and updates its internal state from the notification. It does not own the source-of-truth anymore; it is a consumer like any other plugin.

---

### Step 6 — Handle the bootstrap initial dispatch

`VerificationServicePlugin` relies on receiving an initial context before `start()`. After all plugins have completed `init()` (but before `start()` is called on any), dispatch one initial round of all 4 notifications with startup values. This replaces the current pattern where `init()` receives a populated `BlockNodeContext`.

In `BlockNodeApp.start()`, after the plugin init loop, add:

```java
blockMessaging.sendTssDataUpdate(new TssDataNotification(blockNodeContext.tssData()));
blockMessaging.sendAddressBookHistoryUpdate(new AddressBookHistoryNotification(blockNodeContext.rangedAddressBookHistory()));
blockMessaging.sendStoredBlocksUpdate(new StoredBlocksNotification(blockNodeContext.storedBlocks()));
blockMessaging.sendAvailableBlocksUpdate(new AvailableBlocksNotification(blockNodeContext.availableBlocks()));
```

---

### Step 7 — Migrate each plugin

Each plugin implements `ApplicationStateNotificationHandler` and registers during `init()`:

```java
context.blockMessaging().registerApplicationStateNotificationHandler(this, false, name());
```

| Plugin | Implements | Replaces `onContextUpdate` with |
|---|---|---|
| [`BackfillPlugin`](block-node/backfill/src/main/java/org/hiero/block/node/backfill/BackfillPlugin.java) | `handleStoredBlocksUpdate` | Rebuild `knownBlockRanges` from `notification.storedBlocks()` |
| [`VerificationServicePlugin`](block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/VerificationServicePlugin.java) | `handleTssDataUpdate`, `handleStoredBlocksUpdate` | `safeUpdateTssData` from tss notification; `updateLastVerifiedBlock` from stored notification |
| [`RosterBootstrapTssPlugin`](block-node/roster-bootstrap-tss/src/main/java/org/hiero/block/node/roster/bootstrap/tss/RosterBootstrapTssPlugin.java) | `handleTssDataUpdate`, `handleAddressBookHistoryUpdate` | Update `volatile` fields for each |
| [`RsaRosterBootstrapPlugin`](block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java) | `handleAddressBookHistoryUpdate` | Update `volatile` context reference |
| [`ServerStatusServicePlugin`](block-node/server-status/src/main/java/org/hiero/block/node/server/status/ServerStatusServicePlugin.java) | all 4 | Store individual `volatile` fields instead of whole context |
| [`StreamPublisherPlugin`](block-node/stream-publisher/src/main/java/org/hiero/block/node/stream/publisher/StreamPublisherPlugin.java) | `handleStoredBlocksUpdate`, `handleAvailableBlocksUpdate` | Update the `AtomicReference` fields used by publisher threads |

### LINKED ISSUE #3608: F3 - Remove all onContextUpdate related code
### Story Form

As a BlockNode engineer
I want state change notifications to be as performant as possible

### Step 8 — Simplify `checkForApplicationStateUpdates` and `updateBlockNodeContext`

With all 4 update paths now dispatching eagerly, the scan loop in `checkForApplicationStateUpdates` no longer needs to poll for changes to these 4 fields. Remove that logic from `updateBlockNodeContext`. What remains in the scan loop (if anything) depends on whether there are other responsibilities — audit and trim to only what's still needed.

---

### Step 9 — Update `PluginTestBase`

In [`PluginTestBase.java`](block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java), replace the 6 `plugin.onContextUpdate(blockNodeContext)` call sites with typed notification dispatch calls:

- `updateTssData()` → `blockMessaging.sendTssDataUpdate(...)`
- `updateAddressBook()` / `updateAddressBookHistory()` → `blockMessaging.sendAddressBookHistoryUpdate(...)`
- `replaceStoredBlocks()` → `blockMessaging.sendStoredBlocksUpdate(...)`
- `replaceAvailableBlocks()` → `blockMessaging.sendAvailableBlocksUpdate(...)`
- `refreshContext()` → dispatch all 4

---

### Step 10 — Remove `onContextUpdate` from `BlockNodePlugin`

Delete the method from [`BlockNodePlugin.java`](block-node/spi-plugins/src/main/java/org/hiero/block/node/spi/BlockNodePlugin.java). Update the interface Javadoc to describe the new `ContextNotificationHandler` registration pattern.

---

### Step 11 — Shrink `BlockNodeContext`

After all plugins are migrated and no plugin code reads `tssData`, `rangedAddressBookHistory`, `storedBlocks`, or `availableBlocks` from the context record, remove those 4 components from the record and its `Builder`. The context then carries only the 9 stable fields set at startup.

berryware:
## Reviewer Notes

This PR completes the migration of the Application State Facility to the direct-notification pattern introduced in #3609. Prior to this change, `BlockNodeApp` used a periodic scanner (`checkForApplicationStateUpdates`) and intermediate queues/`AtomicReference`s to accumulate state changes and then broadcast them to plugins via the now-removed `BlockNodePlugin.onContextUpdate`. This PR replaces that polling approach with immediate dispatch: each update method fires its notification the moment the data changes.

### What changed

**`BlockNodeApp` - direct dispatch in all update methods**

- The current TSS data, address book history (together with its lookup index), stored blocks and available blocks are held in `AtomicReference`s and updated with compare-and-set retry loops, since several plugin threads update them concurrently.
- Persisting and dispatching run on a single `ApplicationStateDispatcher` thread (renamed from `ApplicationStateScanner`), so writes are serialized and a notification never carries older data than one already sent. Before that thread exists (startup load) or after it is gone (shutdown), they run inline.
- `updateTssData` - installs the data immediately, then persists and dispatches `TssDataNotification` on the dispatcher; drops updates that are not strictly newer than the current value.
- `updateAddressBookHistory` - installs the history and its index immediately, then persists and dispatches `AddressBookHistoryNotification` on the dispatcher; drops non-newer histories.
- `addStoredBlockRange` - adds the range and dispatches `StoredBlocksNotification` if the merged set changed.
- `updateAvailableBlocks()` - reads the live `HistoricalBlockFacility` union and dispatches `AvailableBlocksNotification` and/or `StoredBlocksNotification` if either changed.
- `checkForApplicationStateUpdates` - renamed to `persistBlockRangesIfDue` and reduced to only the `BLOCK_RANGE_PERSIST_INTERVAL` threshold check; all live-update dispatching removed.
- `startApplicationStateFacility` - starts the messaging facility (index-0 plugin) first so notifications can be dispatched during `loadApplicationState`, then dispatches the available/stored blocks the providers loaded during `init()`; symmetrically stops it in `stopApplicationStateFacility`.
- `shutdown` - `stopApplicationStateFacility` now runs after `serviceBuilder.stopAll()`, so the messaging facility is not stopped while gRPC is still accepting blocks.
- `start` / `shutdown` - skip the messaging facility in the plugin start/stop loop (it is managed by the state facility instead).
- TSS data, address book history and block ranges are persisted through one `replaceFile` helper (unique tmp file + atomic move, non-atomic fallback; the target is never deleted first).
- Removed: `ConcurrentLinkedQueue<TssData> tssDataUpdates`, `AtomicReference<RangedAddressBookHistory> pendingAddressBookHistory`, `getPendingTssData()`, `updateBlockNodeContext()`.

**`BlockNodeContext` (SPI) - static context only**

- Removed dynamic fields: `tssData`, `rangedAddressBookHistory`, `storedBlocks`, `availableBlocks`, and the `nodeAddressBook()` helper.
- Removed the `Builder` inner class (no longer needed).
- `BlockNodeContext` is now a 9-field record carrying only structural/static node configuration.

**`BlockNodePlugin` (SPI)**

- Removed `onContextUpdate(BlockNodeContext)` default method. Plugins that need dynamic state should implement `ApplicationStateNotificationHandler` and register during `init`, or read the current value from `ApplicationStateFacility` in `start()`.

**`ApplicationStateFacility` (SPI)**

- Added `updateAvailableBlocks()` - called by block provider plugins after their available-blocks set changes; the union is re-read live from `HistoricalBlockFacility`.
- Added `tssData()`, `rangedAddressBookHistory()` and `storedBlocks()` - let plugins read the current state synchronously in `start()`.

**Plugin migrations - `onContextUpdate` -> `ApplicationStateNotificationHandler` / `ApplicationStateFacility` getters**

Every plugin that previously overrode `onContextUpdate` now either implements `ApplicationStateNotificationHandler` and registers with `blockMessaging` during `init` (unregistering in `stop`), or reads the current state from `ApplicationStateFacility` in `start()`:

| Plugin | Handlers implemented / getters used |
|--------|----------------------|
| `VerificationServicePlugin` | `handleTssDataUpdate`, `handleStoredBlocksUpdate`; reads `tssData()` and `storedBlocks()` in `start()` |
| `ServerStatusServicePlugin` | `handleTssDataUpdate`, `handleAddressBookHistoryUpdate`, `handleStoredBlocksUpdate`, `handleAvailableBlocksUpdate` (live-max `fixAvailable` workaround removed) |
| `StreamPublisherPlugin` | reads `storedBlocks()` in `start()` (no handler) |
| `RsaRosterBootstrapPlugin` | reads `rangedAddressBookHistory()` (no handler) |
| `RosterBootstrapTssPlugin` | *(no handler - TSS data updates flow to `ApplicationStateFacility`, not back to this plugin)* |

**Block provider plugins - notify on available-blocks changes**

- `BlockFileHistoricPlugin` - calls `updateAvailableBlocks()` on zip archival and on zip deletion.
- `BlockFileRecentPlugin` - calls `updateAvailableBlocks()` on block write and once per retention deletion batch.
- Blocks loaded during `init()` need no call; the startup union is dispatched when the application state facility starts.

**`BackfillPlugin`**

- Migrated from `onContextUpdate` to `ApplicationStateNotificationHandler` (`handleStoredBlocksUpdate`).

**`BlockMessagingFacilityImpl`**

- Notification sends after `stop()` are dropped with a DEBUG log instead of throwing `RejectedExecutionException` into the caller.

**Test infrastructure**

- `PluginTestBase` - update methods (`updateTssData`, `updateAddressBookHistory`, `replaceStoredBlocks`, `replaceAvailableBlocks`) now dispatch notifications through the test messaging facility instead of rebuilding `BlockNodeContext` and calling `onContextUpdate`. `BlockNodeContext` constructor call reduced to 9 args. Implements the new getters.
- `TestApplicationStateFacility` - added `updateAvailableBlocks()` no-op and the new getters.
- `BlockNodeAppTest` - removed 16 manual `loadedPlugins.get(0).start()/.stop()` calls (now handled by `startApplicationStateFacility`/`stopApplicationStateFacility`). Updated `testPluginStartupInOrder` override to iterate the `plugins` parameter rather than `loadedPlugins` to avoid double-starting the messaging facility. `testAvailableBlocksInContextComesFromHistoricalFacility` now relies on the startup dispatch in `startApplicationStateFacility`. Added tests for concurrent updates, inline sync when the dispatcher rejects, runtime provider changes, startup/init notifications reaching handlers, and the non-atomic move fallback.

## Related Issue(s)
 Closes #3607.
 Closes #3608.


block-node/app-config/src/main/java/org/hiero/block/node/app/config/state/ApplicationStateConfig.java
```diff
@@ -35,8 +35,8 @@
  *     `Node` entries to fill in the current Node Store (a.k.a. address book data) for determining
  *     valid publishers and publisher endpoints.
  * @param updateScanInterval The amount of milliseconds that the {@code ApplicationStateFacility} waits between
- *     checking to see if there are any {@code TssData} updates to process.
- * @param updateInitialDelay The amount of milliseconds before the first scheduled scan. Defaults to {@code 0}
+ *     checking whether the stored block ranges need to be persisted.
+ * @param updateInitialDelay The amount of milliseconds before the first scheduled persist check. Defaults to {@code 0}
```
ata-nas: This affected tss data updates before, now we say it affects stored block ranges? Is that correct? Does this scan interval not affect multiple things?

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
```diff
@@ -320,31 +324,31 @@ private BlockNodeVersions buildBlockNodeVersions() {
      */
     @Override
     public void updateTssData(TssData tssData) {
-        blockNodeContext =
-                new BlockNodeContext.Builder(blockNodeContext).tssData(tssData).build();
-        plugin.onContextUpdate(blockNodeContext);
+        this.currentTssData = tssData;
+        blockMessaging.sendTssDataUpdate(new TssDataNotification(tssData));
```
ata-nas: Should we simply skip a tss update if `tssData` argument is `null`?

block-node/app/src/testFixtures/java/org/hiero/block/node/app/fixtures/plugintest/PluginTestBase.java
```diff
@@ -320,31 +324,31 @@ private BlockNodeVersions buildBlockNodeVersions() {
...
     }
 
     public boolean updateAddressBook(NodeAddressBook nodeAddressBook) {
-        blockNodeContext = new BlockNodeContext.Builder(blockNodeContext)
-                .nodeAddressBook(nodeAddressBook)
+        final RangedAddressBookHistory history = RangedAddressBookHistory.newBuilder()
+                .addressBooks(List.of(org.hiero.block.api.RangedNodeAddressBook.newBuilder()
+                        .addressBook(nodeAddressBook)
+                        .startBlock(0)
+                        .endBlock(-1)
```
ata-nas: Are these hard coded values always what we want in test base?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -418,21 +437,59 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
     ///
-    /// @param tssData The TssData to be updated on the \`BlockNodeContext\`
+    /// Installs the data immediately if it is newer than the currently stored value; persisting it
+    /// and dispatching the notification happen on the ApplicationStateDispatcher thread.
     @Override
     public void updateTssData(TssData tssData) {
-        if (tssData != null) tssDataUpdates.add(tssData);
+        while (true) {
+            final TssData current = currentTssData.get();
+            if (tssData == null || (current != null && tssData.validFromBlock() <= current.validFromBlock())) {
```
ata-nas: I think we can compare `tssData.equals(current)` possibly. I think that is what we did in verification. The complier that will compile the TssData class has an implementation of equals and hashcode. Something to consider, otherwise tss data that is different, but with the same valid from block, it will not be replaced.

We are more restrictive now in verification, here is now we do it for reference:

Image

Not saying we should be doing the same. My question is: I am not sure if we should allow all updates so long as `!tssData.equals(current)` regardless of valid from block? And then, for verification, should we also accept updates w/o asserting `updatedTssData.validFromBlock() > localCurrentTss.validFromBlock()`?

@jsync-swirlds any opinion here?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -418,21 +437,59 @@ protected void startPlugins(List<BlockNodePlugin> plugins) {
...
-    /// the current TssData.
+    /// {@inheritDoc}
     ///
-    /// @param tssData The TssData to be updated on the \`BlockNodeContext\`
+    /// Installs the data immediately if it is newer than the currently stored value; persisting it
+    /// and dispatching the notification happen on the ApplicationStateDispatcher thread.
     @Override
     public void updateTssData(TssData tssData) {
-        if (tssData != null) tssDataUpdates.add(tssData);
+        while (true) {
```
ata-nas: This `while (true)` is quite dangerous, I don't think we should rely on a successful cas to exit.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -583,94 +731,27 @@ private void checkForApplicationStateUpdates() {
...
-                || !storedBlockRange.equals(context.storedBlocks())) {
-            builder.storedBlocks(storedBlockRange);
-        }
-
-        LOGGER.log(TRACE, "BlockNodeContext updated");
-        // update the BlockNodeContext
-        blockNodeContext = builder.build();
-        return true;
+        // Stop the messaging facility that was started in startApplicationStateFacility.
+        loadedPlugins.getFirst().stop();
```
ata-nas: Is this the correct place to stop the messaging facility?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -375,14 +396,15 @@ public void shutdown(String className, String reason) {
...
-
             // wait for the shutdown delay
             LockSupport.parkNanos(serverConfig.shutdownDelayMillis() * 1_000_000L);
             serviceBuilder.stopAll();
-            // Stop all the facilities &  plugins
-            for (BlockNodePlugin plugin : loadedPlugins) {
+            // Stop the application state facility only once the servers are closed. It stops the
+            // messaging facility, and stopping that while gRPC is still accepting would drop inbound
+            // blocks into a halted ring buffer and reject every notification send.
+            stopApplicationStateFacility();
```
jsync-swirlds: Why are we treating this different from other plugins?

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
```diff
@@ -341,10 +361,11 @@ protected final BlockNodeVersions versionInfo(final List<BlockNodePlugin> plugin
...
     /// and starts the metrics.
     public void start() {
-        // start the ApplicationStateFacility
+        // startApplicationStateFacility starts the messaging facility (loadedPlugins.get(0)) and
+        // loads persisted state, dispatching initial notifications directly.
         startApplicationStateFacility();
-        // start the plugins
-        startPlugins(loadedPlugins);
+        // Start the remaining plugins; the messaging facility (index 0) is already running.
+        startPlugins(loadedPlugins.subList(1, loadedPlugins.size()));
```
jsync-swirlds: How are we ensuring that index `0` is the messaging facility?

ata-nas (commented):
I had an unreleased draft review...

Posting now, hope this is still relevant, but I will go over this again.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
jsync-swirlds: We should definitely check valid from block (at minimum) for every TSS data update.

block-node/facility-messaging/src/main/java/org/hiero/block/node/messaging/BlockMessagingFacilityImpl.java
```diff
@@ -497,12 +498,27 @@ public synchronized void unregisterBlockItemHandler(final BlockItemHandler handl
...
+     *
+     * @param send the send to run on the message forwarder
+     */
+    private void forward(final Runnable send) {
+        try {
+            messageForwarder.submit(send);
+        } catch (final RejectedExecutionException e) {
+            LOGGER.log(DEBUG, "Messaging facility is stopped, dropping notification");
+        }
+    }
```
jsync-swirlds: This is a little concerning as rejection can happen for a lot of other reasons than a shut down.
We might consider alternative rejected execution handlers (such as running on the calling thread), which can avoid throwing and also handle overload conditions more effectively.

jsync-swirlds (commented):
Only part way through, but publishing comments for now.

block-node/app/src/main/java/org/hiero/block/node/app/BlockNodeApp.java
ata-nas: I also saw that but decided not to comment. It is added first in the list, rather than saved as a field and used. I don't like this, but if we are soon migrating to another approach, we might have this for a while.

Issue #3598, opened 2026-09-03
Design a real-time block-streaming client capability for Block Node to Block Node replication


Nana-EC:
Part of #3597

### Story Form

As a Block Node maintainer
I want a reviewed design document for how a Block Node can continuously and reliably receive blocks in real time from another Block Node
So that implementation work can proceed from a shared, well-understood approach rather than ad hoc decisions made during coding

### Technical Notes

This ticket's deliverable is a written design document, not code.

The Block Node currently pulls blocks from peer Block Nodes only through the backfill capability, which detects gaps on a timer and fetches them in bounded batches. The goal described in the parent epic is a complementary capability: a Block Node acting as a client that maintains a live streaming connection to a peer Block Node's block-stream subscription endpoint, continuously receiving blocks in real time rather than on a polling interval.

The design document should propose an approach and address at least the following:

- **Peer configuration**: how an operator configures one or more candidate peer Block Nodes to receive from, including how preference/priority among them is expressed.
- **Source selection and failover**: how the system chooses which configured peer to actively receive from, and how/when it switches to a different one. This should distinguish a peer that has become unavailable (e.g., disconnected or erroring) from one that is still connected but performing poorly (e.g., meaningfully delayed relative to real time), and should describe whether/how a previously-abandoned peer can be reconsidered later.
- **Pipeline integration**: how blocks received this way are integrated into the Block Node's existing ingestion, verification, and storage pipeline so they are treated consistently with blocks obtained through any other path.
- **Relationship to backfill**: how this capability's responsibilities are scoped relative to the existing backfill capability so the two do not duplicate effort or conflict, including how a large existing backlog (a node starting out far behind a peer) is expected to be handled.
- **Configuration, observability, and testing**: the relevant configuration surface, metrics/logging needed for operators to understand current source and switch history, and the testing strategy (unit and end-to-end) that will be used to validate the approach.

### Acceptance Criteria

- [ ] A design document addressing the areas above is written and merged into the repository's design documentation.
- [ ] The document is detailed enough that implementation work can be scoped into concrete follow-up issues without further design decisions being made ad hoc.
- [ ] The document explicitly calls out any open questions, risks, or assumptions that still need validation during implementation.

Issue #3597, opened 2026-09-03
Epic: Real-time block streaming client plugin


Nana-EC:
### Goal

Enable a Block Node to receive blocks from another Block Node in real time over a live streaming connection, so that deployments which need to closely mirror another Block Node's data are not limited by the delay inherent in periodic gap-fill polling.

### Intention

Today, a Block Node can only pull blocks from peer Block Nodes through the backfill capability, which periodically scans for gaps in its stored block sequence and fetches them in bounded batches. This works well for closing historical gaps, but it is fundamentally a polling model with scan-interval and per-batch latency built in. External consumers of a Block Node's block stream, such as Mirror Nodes and other subscribers, already benefit from lower-latency access by opening a single subscription and consuming blocks as they arrive. This epic covers giving a Block Node the same capability with respect to another Block Node: the ability to act as a client that opens and maintains a live streaming connection to a peer Block Node's block-stream subscription endpoint, continuously receiving and ingesting blocks as they become available, and doing so reliably across a configured set of candidate peers.

### Considerations

- This capability should complement, not replace, the existing backfill capability. Backfill's periodic, multi-peer gap-filling role should remain intact; the two should be able to run together without duplicating effort or conflicting.
- A Block Node adopting this capability may start out significantly behind a peer's current tip and need to catch up a large backlog in addition to staying live going forward. The design should account for both cases.
- Peer Block Nodes should be configurable as a set of candidate targets (not just a single fixed endpoint), so an operator can express preferred and fallback sources. The existing configuration approach used for peer targets elsewhere in the Block Node is a reasonable starting point for consistency, but the final configuration model is left to the design phase.
- The system should optimize for availability and low latency when choosing which peer to receive from, and should be able to detect and react when its current source becomes unavailable or is performing poorly (e.g., meaningfully delayed relative to real time), switching to a healthier alternative.
- Blocks received this way must be integrated into the Block Node's existing ingestion, verification, and storage pipeline like blocks obtained through any other path, and operators should have enough observability (logging/metrics) to understand which peer is currently the active source and why a switch occurred.

### Technical Details

The Block Node already exposes a block-stream subscription service used by external consumers (e.g., Mirror Nodes) to receive historical and live blocks over a single streaming connection. This epic is about consuming that same kind of capability as a client, against another Block Node, rather than about changing what is exposed to external consumers.

### Tasks and Features

[...]

PR #3635, opened 2026-09-09
docs: design for real-time block streaming client plugin

rockysingh:
## Summary

Design document for a new BN-to-BN real-time streaming client plugin (Story #3598, Epic #3597).

The plugin consumes a peer Block Node's `BlockStreamSubscribeService.subscribeBlockStream` RPC as an open-ended live tail (`end_block_number = uint64_max`), enabling cross-BN replication for deployments where the receiving BN is not directly connected to consensus node publishers. It complements the existing `BackfillPlugin` by owning the live edge while backfill continues to handle historical gaps.

Document covers:

- **Peer configuration** via the shared `BlockNodeSource` PBJ message (which already carries a purpose-built `subscribe_port` field).
- **Source selection and failover** reusing `PriorityHealthBasedStrategy` with a client-side stale-watchdog for "meaningfully delayed" peers.
- **Pipeline integration** via `BlockMessagingFacility.sendBlockItems`, making replicated blocks look identical to directly-published blocks so downstream verification, persistence, archive and subscriber fan-out all pick them up automatically.
- **Coexistence with `StreamPublisherPlugin`** enforced as a hard mutex at `init()` (the item ring's single-producer contract requires this).
- **Startup and reconnection** semantics.
- **Configuration, metrics, exceptions, security, acceptance tests**, and open design questions.

`docs/design/streaming/subscribe-client-plugin.md`.

## Test plan

- [x] Design claims cross-checked against `BlockMessagingFacility.java`, `block_stream_subscribe_service.proto`, `block_node_source.proto`, and the existing Backfill + Bootstrap-RSA plugin implementations.
- [ ] Reviewer sanity check on the injection-point decision (item ring vs new notification type) and the mutex-with-StreamPublisher constraint.

### DIFF
--- a/docs/design/streaming/subscribe-client-plugin.md
+++ b/docs/design/streaming/subscribe-client-plugin.md
@@ -1,27 +1,5 @@
 # Subscribe Client Plugin Design Document
 
-## Table of Contents
-
-1. [Purpose](#purpose)
-2. [Goals](#goals)
-3. [Non-Goals](#non-goals)
-4. [Terms](#terms)
-5. [Entities](#entities)
-6. [Design](#design)
-   - [Peer Configuration](#peer-configuration)
-   - [Source Selection and Failover](#source-selection-and-failover)
-   - [Pipeline Integration](#pipeline-integration)
-   - [Relationship to Backfill](#relationship-to-backfill)
-   - [Coexistence with the Publisher Plugin](#coexistence-with-the-publisher-plugin)
-   - [Startup and Reconnection](#startup-and-reconnection)
-7. [Diagram](#diagram)
-8. [Configuration](#configuration)
-9. [Metrics](#metrics)
-10. [Exceptions](#exceptions)
-11. [Security](#security)
-12. [Acceptance Tests](#acceptance-tests)
-13. [Open Questions](#open-questions)
-
 ## Purpose
 
 Enable a Block Node (BN) to receive blocks from another BN in real time by
@@ -31,8 +9,8 @@ deployments where the receiving BN is not directly connected to consensus
 node publishers.
 
 Real-time replication complements the existing `BackfillPlugin` (which
-periodically fetches historical gaps). The Backfill plugin remains
-responsible for catching up older ranges; this plugin owns the live edge.
+periodically fetches historical gaps). Backfill continues to own catching
+up older ranges; this plugin owns the live edge.
 
 ## Goals
 
@@ -40,26 +18,30 @@ responsible for catching up older ranges; this plugin owns the live edge.
    stays within a small, bounded number of blocks of the peer's tip.
 2. Prefer higher-priority peers and fail over to lower-priority peers when
    the primary is unavailable or meaningfully delayed.
-3. Deliver received blocks into the local ingestion pipeline in a form
-   indistinguishable from directly-published blocks, so all downstream
-   plugins (verification, persistence, archive, subscriber fan-out) act on
-   them automatically.
-4. Reuse the shared BN-to-BN client stack (`BlockStreamSubscribeUnparsedClient`,
-   `BlockNodeSource` peer config, `PriorityHealthBasedStrategy` selection).
-5. Provide operator instrumentation (metrics, logs) that make replication
+3. Deliver received blocks into the local ingestion pipeline through the
+   new **Unvalidated Blocks ring buffer** introduced by Epic #3612 (story
+   #3614), keeping this path fully isolated from the live-publisher ring.
+4. Support two delivery modes selectable by global config: **immediate**
+   (forward each `BlockItemSet` as it arrives, lowest latency) and
+   **full-block** (accumulate items and deliver a whole block once
+   complete, simplest downstream contract).
+5. Reuse the shared BN-to-BN client stack
+   (`BlockStreamSubscribeUnparsedClient`, `BlockNodeSource` peer config,
+   `PriorityHealthBasedStrategy` selection).
+6. Provide operator instrumentation (metrics, logs) that make replication
    lag, reconnects, and failover visible.
 
 ## Non-Goals
 
 1. Historical gap detection or backfill. That remains `BackfillPlugin`'s
    responsibility. This plugin only tails live.
-2. Publisher-side responsibilities (accepting inbound consensus streams,
-   arbitrating between concurrent publishers). This plugin is a **consumer**
-   of a peer BN's subscribe stream, not a producer.
+2. Reworking the `StreamPublisherPlugin`, `VerificationServicePlugin`, or
+   the existing item ring. The design deliberately uses a separate ring
+   buffer so the well-tested publisher path is untouched.
 3. Multi-source aggregation. Only one peer streams to us at a time; the
    others are standby.
-4. Block validation beyond what `VerificationServicePlugin` already applies
-   on the item ring.
+4. Making the two delivery modes selectable per-peer. Mode is a **global
+   on/off**; every configured peer uses the same mode.
 
 ## Terms
 
@@ -79,9 +61,29 @@ responsible for catching up older ranges; this plugin owns the live edge.
       <code>end_block_number = uint64_max</code>, causing the server to stream
       indefinitely until the client half-closes or the connection breaks.</dd>
 
+  <dt>Immediate Mode</dt>
+  <dd>Delivery mode where each <code>BlockItemSet</code> received from the peer
+      is forwarded downstream as soon as it arrives. Lowest end-to-end latency;
+      downstream consumers must handle partial-block streams. Considered a
+      "nice to have" — expected to be used sparingly (e.g. Jasper or other
+      latency-sensitive downstream consumers).</dd>
+
+  <dt>Full-Block Mode</dt>
+  <dd>Delivery mode where item sets are accumulated until a complete block is
+      assembled, then delivered as a single whole-block message. Simpler
+      downstream contract; higher latency by one block-assembly interval.
+      Default mode.</dd>
+
+  <dt>Unvalidated Blocks Ring</dt>
+  <dd>The new ring buffer added by Epic #3612 (story #3614) for full blocks
+      that have not yet been verified locally. Sources include backfill,
+      future tier-two-subscriber plugins (this plugin), and gossip. Isolated
+      from the live-publisher item ring so this plugin cannot back-pressure
+      the live path.</dd>
+
   <dt>Meaningfully Delayed</dt>
-  <dd>A peer that is still reachable but is not delivering blocks to us at the
-      expected cadence. Detected client-side by
+  <dd>A peer that is still reachable but not delivering blocks at the expected
+      cadence. Detected client-side by
       <code>time_since_last_BlockEnd &gt; staleThresholdMs</code>. Triggers
       failover to a lower-priority peer.</dd>
 
@@ -94,14 +96,11 @@ responsible for catching up older ranges; this plugin owns the live edge.
       <code>SourceHealth</code>) and opening a new subscribe stream against
       the next selectable peer.</dd>
 
-  <dt>BlockItems</dt>
-  <dd>The record <code>(items, blockNumber, isStartOfNewBlock, isEndOfBlock)</code>
-      that <code>BlockMessagingFacility.sendBlockItems</code> accepts. Item
-      ring payload.</dd>
-
-  <dt>Item Ring</dt>
-  <dd>The LMAX disruptor inside <code>BlockMessagingFacility</code> that
-      carries <code>BlockItems</code>. Ordered by contract; single-producer.</dd>
+  <dt>SubscribedBlockNotification</dt>
+  <dd>New notification message (defined by this plugin) carrying either a
+      full <code>BlockUnparsed</code> or a <code>BlockItemSetUnparsed</code>
+      via a <code>oneof</code>. Exactly one of the two variants is set per
+      notification, driven by the configured delivery mode.</dd>
 </dl>
 
 ## Entities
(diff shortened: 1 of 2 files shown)


docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+
+Enable a Block Node (BN) to receive blocks from another BN in real time by
+consuming a peer's `BlockStreamSubscribeService.subscribeBlockStream` RPC as
+a long-lived open-ended live tail. This provides cross-BN replication for
+deployments where the receiving BN is not directly connected to consensus
+node publishers.
+
+Real-time replication complements the existing `BackfillPlugin` (which
+periodically fetches historical gaps). The Backfill plugin remains
+responsible for catching up older ranges; this plugin owns the live edge.
```
Nana-EC: I wouldn't say this is accurate. This plugin should be able to stream blocks anywhere from history.
Initially it was considered a replacement for backfill in the RFH case, but it's worth exploring if we'd want an RFH to utilize both backfill plugin for historical and this plugin for recent.
In that case the greedy logic of backfill might be replaceable.

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+
+## Goals
+
+1. Continuously stream the peer's live block tail into the local BN so it
+   stays within a small, bounded number of blocks of the peer's tip.
+2. Prefer higher-priority peers and fail over to lower-priority peers when
+   the primary is unavailable or meaningfully delayed.
+3. Deliver received blocks into the local ingestion pipeline in a form
```
Nana-EC: Aggressive posture means high latency means there needs to be support for just passing along BlockItemSets.
Question is does the plugin need additional logic to ensure BlockItemSets from different blocks aren't mixed up or if BN needs to take postures to validate a full block.
Verification plugin should already handle some of this logic so we should confirm if that's the case so we don't repeat logic in the new plugin.

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+   indistinguishable from directly-published blocks, so all downstream
```
Nana-EC: Team agreed that rather we want to be able to distinguish blocks from publisher vs blocks from this plugin so that we can properly protect and prioritize the flows

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+      an open-ended live tail (<code>end_block_number = uint64_max</code>).</dd>
+
+  <dt>Peer BN</dt>
+  <dd>Another Block Node instance that exposes
+      <code>BlockStreamSubscribeService</code>. Described by a
+      <code>BlockNodeSourceConfig</code> entry in the peer-sources JSON file.</dd>
+
+  <dt>Live Tail</dt>
+  <dd>The RPC mode where the client requests
+      <code>start_block_number = local_tip + 1</code> and
```
Nana-EC, line 60: Doesn't necessarily need to be local_tip+1, more so just needs to be close to the tip. A block node could start historical and catch up and end up streaming at the live tip.
The more important part is the end_block_number signifying continued streaming.

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+      carries <code>BlockItems</code>. Ordered by contract; single-producer.</dd>
+</dl>
+
+## Entities
+
+- **`SubscribeClientPlugin`** implements `BlockNodePlugin`. Registers itself,
+  starts the streaming loop on `start()`, tears it down on `stop()`.
+- **`SubscribeClientConfiguration`** — `@ConfigData("subscribe.client")`
+  record with peer-sources file path, thresholds, and tuning knobs.
+- **`SubscribeClientConfigExtension`** — registers the config record.
```
Nana-EC, line 112: Can you elaborate. Not sure what the value of this is and which config is being registered

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+- **`SubscribeSessionRunner`** — long-lived loop that owns the active
+  peer selection and drives one `BlockStreamSubscribeUnparsedClient` call at
+  a time. Runs on a dedicated platform thread.
```
Nana-EC: Is the choice of platform thread here due to the previous Helidon limitation i.e. [HTTP/2 Connection Fails on Virtual Threads in Container Environments using K8S](https://github.com/hiero-ledger/hiero-block-node/issues/2091) or for some other reason?
We should check if that limitation still exists and if not utilize virtual threads.

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+Rationale for reusing `BlockNodeSource`: the proto already carries a
+dedicated `subscribe_port` field designed for this use case, and the JSON
+format is identical to what operators already maintain for Backfill.
+
+### Source Selection and Failover
+
+Selection reuses `PriorityHealthBasedStrategy`:
+1. Filter out peers currently in exponential backoff.
+2. Sort by `(priority ASC, healthScore ASC)`.
+3. Pick the first candidate whose pre-flight `serverStatus` shows a tip at
```
Nana-EC, line 157: I think this should be pick the "candidates", and then there should be a 4th that is picks the candidate with lowest latency.

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+- If `SubscribeClientPlugin` is enabled and `StreamPublisherPlugin` is also
+  present on the plugin classpath and enabled, `init()` throws with a clear
+  error message.
+- The reverse check lives symmetrically in `StreamPublisherPlugin` (or in a
+  shared validator called by both) to avoid coupling to plugin load order.
+
+A "replica BN" deployment profile disables `StreamPublisherPlugin` and
+enables `SubscribeClientPlugin`; a "primary BN" deployment does the opposite.
+
+### Startup and Reconnection
```
Nana-EC, line 296: not sure if this is mixing up plugin init() and start() but we should break it out. It's fine for init to parse and validate the peers, the actual subscribe though should go out in start()

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+    SC->>BMF: sendBlockItems(N+2)
+```
+
+## Configuration
+
+`@ConfigData("subscribe.client")` record:
+
+|          Field          |   Type   |  Default  |                                 Purpose                                 |
+|-------------------------|----------|-----------|-------------------------------------------------------------------------|
+| `enabled`               | boolean  | `false`   | Master switch. Must be `false` when `StreamPublisherPlugin` is enabled. |
```
Nana-EC: Plugins don't have `enabled` by design. Their presence enables logic so you should remove this

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,465 @@
...
+| `blockNodeSourcesPath`  | String   | `""`      | Path to peer-sources JSON (parsed as PBJ `BlockNodeSource`).            |
+| `staleThresholdMs`      | long     | `3000`    | Time since last `BlockEnd` before failing over.                         |
+| `initialRetryDelayMs`   | long     | `500`     | Base for exponential per-peer backoff.                                  |
+| `maxBackoffMs`          | long     | `60000`   | Cap on per-peer backoff.                                                |
+| `reconnectMinDelayMs`   | long     | `250`     | Minimum sleep between session iterations.                               |
+| `preflightServerStatus` | boolean  | `true`    | Whether to call `serverStatus` before opening a subscribe stream.       |
```
Nana-EC: Q: Why is this optional? We should always to check the peer BN is up and have the right set of blocks before subscribing .

Nana-EC (changes requested):
Initial thoughts on 1st pass, will circle back

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Updated in 7410190e. Live Tail term now says the important part is `end_block_number = uint64_max` and start can be anywhere from near tip to deep history.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Elaborated in 7410190e. Now spells out the JPMS `provides` mechanism, what it registers, and the pattern parallel with `BackfillConfigExtension`.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Added in 7410190e. Steps are now: (1) filter backoff, (2) sort by priority + health, (3) reduce to candidates by pre-flight `serverStatus`, (4) pick the candidate with lowest observed `serverStatus` RTT (priority as tiebreaker).

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Split in 7410190e. `init()` parses + validates peers, `start()` launches the SubscribeSessionRunner, `stop()` cancels and releases.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Removed in 7410190e. Config table has no `enabled` row; added a note that presence of the plugin on the classpath is what enables it.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Made unconditional in 7410190e. `preflightServerStatus` field is gone; every reconnect calls `serverStatus` before opening the subscribe stream.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Kept the plugin live-tail-only for v1 but the Live Tail term now allows cold-start requests to begin deep in history and stream through to live via `end_block_number = uint64_max`. Backfill still owns bounded gap-fills. Happy to broaden further if the team wants this plugin to subsume backfill greedy mode.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Adding an explicit ordering section. Plan is to document what the verification plugin enforces today on the item ring vs what a plugin-owned per-block queue needs to add. Landing with the next commit.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Adding `BlockSource.SUBSCRIBED` and threading it through the pipeline-integration + metrics sections. Landing with the next commit.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Investigating #2091 now. If it is still open, the doc keeps platform threads with an explicit reference; if fixed, switching to virtual threads. Follow-up commit.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Non-Goal #1 reframed in b288ff78. Backfill still owns bounded historical gap-fills; this plugin can request a `start_block_number` deep in history and stream through to live via `end_block_number = uint64_max`. Subsuming Backfill greedy mode is called out as a later-iteration option, not v1.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Ordering subsection expanded in b288ff78. Covers per-block/per-stream ordering from `SubscribeSessionRunner` being single-threaded, the difference between item-ring ordering (via `BlockItems` flags) and Unvalidated Blocks ring ordering (via explicit `block_number` + `oneof` shape), and the failover ordering behaviour. No new cross-plugin ordering primitives required.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Added in b288ff78. `BlockSource.SUBSCRIBED` is a one-line addition to `BlockSource.java` (currently `UNKNOWN|PUBLISHER|BACKFILL|HISTORY`). Documented in Pipeline Integration and called out as a distinct goal so downstream plugins can protect + prioritize publisher-originated blocks.

docs/design/streaming/subscribe-client-plugin.md
rockysingh: Investigated: [#2091](https://github.com/hiero-ledger/hiero-block-node/issues/2091) was closed 2026-02-04. `SubscribeSessionRunner` now runs on a virtual thread (b288ff78, plus diagram update). Left a rationale note explaining the Backfill-era platform-thread pinning is no longer required.

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,665 @@
...
+The design **strongly discourages** enabling both in the same deployment
+via documented operator guidance rather than a hard `init()`-time mutex.
+Startup logs a WARN when both are enabled to make the situation visible.
+
+Typical deployments:
+
+- **Primary BN** — `StreamPublisherPlugin` deployed,
+  `SubscribeClientPlugin` not deployed. Receives directly from consensus.
+- **Replica BN** — `SubscribeClientPlugin` deployed,
+  `StreamPublisherPlugin` not deployed. Mirrors a primary BN.
```
ivannov: Question: with respect to persistence plugins we have two profiles: LFH and RFH. They make sure that two alternative plugin groups (history and cloud) are present in their respective profile

What will be the deployment model for primary vs replica block node? Do we add new profiles (and thus turn them into a matrix)?

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,665 @@
...
+|----------------|--------------------------------------------------------|------------------------------------------------------------------------------|
+| Time horizon   | Near `local_tip` → ∞ (`end_block_number = uint64_max`) | Historical gaps                                                              |
+| Trigger        | Continuous open-ended stream                           | Gap detection + periodic sweep                                               |
+| RPC            | `subscribeBlockStream` (open-ended)                    | `subscribeBlockStream` (bounded ranges)                                      |
+| Injection ring | Unvalidated Blocks (#3614)                             | Unvalidated Blocks (#3614) — migrated from `sendBackfilledBlockNotification` |
+| Coexist?       | Yes — non-overlapping block ranges                     | Yes — non-overlapping block ranges                                           |
+
+On cold start, if the local BN is far behind, Backfill's historical
+scheduler brings the archive current. Subscribe Client sits idle until
+`local_tip` is close enough to the peer's tip for a live subscribe to
```
ivannov: Question: can you elaborate what _close enough_ is?

docs/design/streaming/subscribe-client-plugin.md
```diff
@@ -0,0 +1,665 @@
...
+3. `SubscribedBlockPublisher` in **full-block mode** buffers item sets and
+   emits exactly one notification per complete block with the `full_block`
+   variant set, keyed on `BlockEnd`.
+4. Stale watchdog fires: given a stream that stops sending `BlockEnd` for
+   longer than `staleThresholdMs`, the current stream is cancelled and a
+   failover happens.
+5. Terminal `Code != SUCCESS` triggers peer failure marking + backoff.
+6. Backoff is exponential per peer and resets on successful reconnect.
+7. Config validation: missing peer file fails `init()`; zero-entry peer
+   file fails `init()`; both-plugins-enabled logs WARN but does NOT fail.
```
ivannov: Maybe also some failover tests on unit level?

PR #3637, opened 2026-09-10
docs: address API throttling design doc review follow-ups (#3630)

### LINKED ISSUE #3630: Follow-ups from the API throttling design doc review (PR #3529)
## Context

PR #3529 (`docs: API throttling and admission control design doc`) merged on 2026-09-09 with
approvals from @berryware and @ata-nas, and a non-blocking stand-down from @jsync-swirlds. A
post-merge review round raised 14 comments, none blocking. This issue collects the items that need a
decision or a doc change, plus the two threads other reviewers explicitly deferred to follow-ups.

- Design doc: `docs/design/apis/api-throttling.md`
- Epic: #3526 · Design issue: #3527 · Companion spike: #3528
- Review round: <https://github.com/hiero-ledger/hiero-block-node/pull/3529#pullrequestreview-5149452314>

Items are grouped by what they cost to resolve. Group A is doc-only. Group B needs design decisions
that affect the implementation PR. Group C affects the acceptance criteria. Group D is other
reviewers' deferred threads, recorded here so they are not lost.

---

### A1. Markdown inside the `<dl>` Terms block is not rendered

The `Terms` section is a raw HTML `<dl>` block with no blank lines between `<dl>` and `</dl>`. Per
CommonMark, that makes the whole span a single HTML block, so markdown inside it is passed through
literally. Two links render as raw source text rather than links:

- `[Alternatives considered](#alternatives-considered)` (in the *Leaky bucket* definition)
- `[`GcraLimiter`](#gcralimiter)` (same definition)

Both anchors are themselves correct — the breakage is the enclosing HTML block, not the targets.
Inline code spans in the same block are also rendered as literal backticks. This is the same defect
@ata-nas screenshotted in <https://github.com/hiero-ledger/hiero-block-node/pull/3529#pullrequestreview-5143378232>.

Fix: convert the `<dl>` to a markdown definition list (bold term + description, matching the rest of
the doc), or insert blank lines around each `<dd>` body so the inner content is parsed as markdown.
Converting to plain markdown is preferable — it keeps the doc consistent and avoids reintroducing the
same trap on the next edit.

### A2. Decide the GCRA framing, then link it once

Two review threads pull in opposite directions and should be settled together rather than patched
independently:

- @jsync-swirlds asked to describe the mechanism as a plain "leaky bucket as meter" and drop the
  GCRA framing, which imports ATM-networking vocabulary that does not help a reader here.
- The post-merge round asked for an in-depth GCRA reference
  (<https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm>) at the `GcraLimiter` entity.

These are compatible if GCRA is scoped to exactly one place: keep the *model* described as a leaky
bucket used as a policer (no GCRA in `Terms` or `Goals`), and mention GCRA only in the `GcraLimiter`
entity as the state representation, with the reference link there. That satisfies both comments and
removes the GCRA mention from the definition where it currently reads as part of the model.

### A3. Clarify "sizing the two independently"

`Component B — shared backend block-read bulkhead` ends with: *"sizing the two independently is a
reasonable approximation rather than a guarantee."* "The two" is ambiguous — it reads as if there
are two bulkheads. It means Component A's node-wide ceiling for historical `getBlock` and Component
B's bulkhead permit count. Name both explicitly.

---

### B1. Content-aware weighing cannot run at `open()` time

The doc states the weigher is *"evaluated once at admission time before any rate/concurrency check
runs"* and the admission-decision diagram places `classify` first. At the chosen attachment point
this is not implementable: `ServiceInterface.open()` receives only `(Method, RequestOptions,
Pipeline)`. The request bytes are not available yet — they arrive later, as `onNext(Bytes)` on the
pipeline `open()` returns.

See `block-node/block-access-service/src/main/java/org/hiero/block/node/access/service/BlockAccessServicePlugin.java:66-82`:
`open()` returns a `Pipelines.unary()` pipeline whose `mapRequest` parses the bytes on a subsequent
callback. The same shape applies to `SubscriberServicePlugin.open()`.

So the real ordering has to be one of:

1. **Two-phase admission** — run the identity-only checks (global concurrency, per-client
   concurrency, rate) at `open()` using a default weight, then re-check against the weight-class
   policy on first `onNext`, releasing and rejecting if the heavier policy denies it. This means a
   rejected historical call has already consumed a rate slot, which is exactly what the documented
   check ordering was designed to avoid.
2. **Weigh on first `onNext` only** — no checks at `open()`, all admission deferred to the first
   request message. Simpler and preserves "the rate check mutates state last," but the permit
   lifecycle then has to cover the window between `open()` and the first message.
3. **A different attachment point** — an interceptor that sees the request payload, which is
   @jsync-swirlds' original suggestion (see D1) and may require a PBJ extension.

Whichever is chosen, the doc's check ordering, the admission-decision diagram, and the
`Concurrency-permit lifecycle` paragraph all need updating to match. This is the highest-value item
in this issue: the current text describes an ordering that the stated attachment point cannot
deliver.

### B2. There is no single recent/historical boundary to compare against

The weigher is specified as classifying *"against the current recent/historical boundary,"* but no
such number exists in the system today. `HistoricalBlockFacilityImpl` resolves a block by iterating
providers in `defaultPriority()` order and asking each for `availableBlocks()`
(`block-node/app/src/main/java/org/hiero/block/node/app/HistoricalBlockFacilityImpl.java:54`); the
tier is a property of *which provider answers*, not of a threshold. There are several providers with
distinct cost profiles (`blocks-file-recent`, `blocks-file-historic`, `cloud-storage-archive`,
`s3-archive`), so "recent vs historical" is not even binary.

The design needs to specify:

- What signal the weigher reads, and how it is cheap enough for the hot path (querying provider
  `availableBlocks()` per call is not).
- Whether that value is cached, and if so its refresh policy and staleness tolerance — a boundary
  that lags behind ingest will misclassify newly-written blocks as historical.
- How `getBlock` with `retrieveLatest` is classified. There is no block number in the request at
  all; resolving it requires `blockProvider.availableBlocks().max()`
  (`BlockAccessServicePlugin.java:102`), which is precisely the kind of backend lookup the weigher
  is supposed to avoid. `blockNumber == -1` has the same problem.

### B3. Per-weight-class limits are promised but not expressible in the configuration

`Goals` item 2 and the Component A weighting paragraph both state that historical requests are
checked against a *stricter policy* than live ones. The `Configuration` table exposes one triple per
method — `ratePerSecond`, `burstTolerance`, `maxConcurrentPerClient` — with no dimension for weight
class, and the node-level record holds one `maxConcurrentGlobal`-equivalent per method, likewise
unqualified.

Either the config needs to be per (method, weight class), or the doc needs to say how a single
triple yields differentiated treatment (e.g. weight as a multiplier on bucket cost rather than as a
policy selector). The second option is cheaper and probably right, but it is a different mechanism
from what `ThrottlePolicy` and `Weight class` currently describe, and acceptance test 4 ("limited
independently... using the same client's traffic on both tiers") reads as requiring the first.

### B4. One bulkhead, or one per storage tier?

`BlockReadBulkhead` is a single pool guarding *"every read against block storage."* Recent and
historical reads have materially different cost and different backing stores, and the `Extensibility`
section already establishes that additional independently-sized bulkheads are a supported shape.
Worth deciding explicitly:

- Should there be a separate pool per tier, so a burst of archive reads cannot starve recent reads?
  With one shared pool, saturation from slow archive reads blocks cheap recent reads that would have
  completed immediately.
- Should recent reads take a permit at all? If a recent-block read is served from memory or a warm
  local file, the permit is pure overhead on the cheap path — and it is the path most sensitive to
  added latency.

### B5. Weight is per request, not per block

`getBlock` is one request for one block. A `subscribeBlockStream` request can be one request for an
arbitrarily large range, and it is classified once, at admission, from its start block. So a
subscription starting 500,000 blocks back and one starting 2 blocks back are charged the same
admission cost, and a subscription that starts live and later falls back into history is never
reclassified.

The doc acknowledges this and defers to Component B, which is a defensible answer for *concurrency*
— the bulkhead does bound in-flight reads regardless of range size. It is not an answer for *rate*:
per-client rate limiting on subscribe counts session opens, so total read work a client can induce
is unbounded by Component A. Either state that as an accepted limitation with the reasoning, or
introduce a range-aware cost for subscribe.

### B6. Reconcile node-wide ceilings with the existing `maxConcurrentStreams`

`WebServerHttp2Config.maxConcurrentStreams` already exists and defaults to `8192`
(`block-node/app-config/src/main/java/org/hiero/block/node/app/config/WebServerHttp2Config.java:16`).
That is an existing node-wide concurrency bound sitting below the proposed one. The new node-level
ceilings should be documented in relation to it: which one is expected to bind first, whether
validation should reject a configuration where the sum of per-method ceilings exceeds the HTTP/2
limit, and whether the HTTP/2 default needs revisiting once real ceilings exist.

### B7. A targeted wire-format field read has no precedent in this codebase

The `Performance` section requires that the weigher *"do a targeted read of that field directly from
the wire format, not fully deserialize the request message."* No hand-written partial protobuf
parsing exists in the project today — `ProtoParserTools` appears only in PBJ-generated code. So this
requirement implies either a new shared utility for single-field wire reads, or accepting the double
parse the doc warns against.

Decide which, and if the utility is in scope, say where it lives and who owns it. Note that the cost
being avoided is small (a partial scan of a tiny request message), so "accept the second parse,
measure it, optimize only if the benchmark in C1 shows it matters" is a legitimate outcome — but it
should be a stated decision, not left as a requirement nobody can satisfy.

### B8. Scope of `ClientKeyExtractor` pluggability

The doc establishes *why* `ClientKeyExtractor` is a separate interface but not *how* an
implementation is selected. Open questions: is there one extractor per node or one per plugin; is
the choice configuration-driven or wiring-driven; can it change without a restart; and if multiple
implementations are ever active at once, do their keys share a namespace (two extractors producing
the same string for different callers would silently merge their limits).

> The original comment framed this as a dispatch-cost concern. That part can be dropped — with a
> single registered implementation the call site is monomorphic and inlines away, and the doc already
> handles the actual per-call cost by requiring key derivation once per connection rather than once
> per call. The selection and namespacing questions above are the part worth answering.

---

### C1. Name the benchmark harness

Acceptance test 9 requires *"a benchmark measuring per-call latency and allocation rate, with
admission control enabled versus disabled."* There is no JMH dependency or benchmark source set in
the repo today, so this needs a decision: add a JMH module, or express the criterion against the
existing performance/e2e tooling. Allocation rate in particular is hard to measure meaningfully
outside a microbenchmark harness.

### C2. Define what "on the same hardware" means

Acceptance test 9 says the enabled-vs-disabled comparison runs *"on the same hardware."* On shared
CI runners that condition does not hold between jobs, and absolute latency numbers are not
comparable across runs. Specify the environment (a dedicated runner, or both arms in one process
within a single job) and express the pass condition as a relative threshold against the
same-run baseline rather than an absolute number.

---

### D1. Interceptor vs. service-interface decorator (@jsync-swirlds)

@jsync-swirlds' main suggestion was to implement admission control as an interceptor rather than by
wrapping `ServiceInterface`, and stood down without withdrawing the objection: *"I still assert that
this is not a great approach, but I will not block progress."* @ata-nas and @berryware both agreed
the question is still open and noted it may want a PBJ extension.

B1 is direct evidence for revisiting this: the decorator cannot see request content at `open()`,
which is what content-aware weighing needs. If an interceptor can, that changes the trade-off the
doc's `Alternatives considered` section should record. This should be resolved before the
implementation PR, not after.

### D2. Priority-aware throttling (@ata-nas)

> "Do we ever envision possible 'priority' awareness for throttling? Possibly giving more priority to
> some connections/clients/requests than others when we see high concurrency?"

Currently out of scope and not mentioned in the doc. It interacts with the unauthenticated-identity
limitation — priority tiers are only meaningful with a trustworthy client identity, so this is
plausibly gated on the `ClientKeyExtractor` successor rather than an independent piece of work. Worth
one line in `Non-goals` recording that and why.

---

## Suggested disposition

| Group | Items | Where |
|-------|-------|-------|
| A | A1, A2, A3 | One small docs PR, no design discussion needed |
| B | B1, B2, B3 | Resolve before the implementation PR — each changes documented behavior |
| B | B4, B5, B6, B7, B8 | Resolve during implementation; fold answers back into the doc |
| C | C1, C2 | Fold into the implementation issue's acceptance criteria |
| D | D1 | Resolve before the implementation PR, with @jsync-swirlds |
| D | D2 | Record as a non-goal now; separate issue under the epic if pursued |

Nana-EC:
## Summary

Addresses the doc-fixable items from #3630 (post-merge review follow-ups on #3529, `docs/design/apis/api-throttling.md`). Doc-only change, no code touched.

**Group A (doc fixes, no design discussion needed):**
- A1: converted the `Terms` section from a raw HTML `<dl>` (broke link/code-span rendering) to a markdown list
- A2: scoped GCRA framing to the `GcraLimiter` entity only, added an in-depth reference link there
- A3: named both quantities in the Component B sizing note instead of "the two"

**Group D2 (non-goal, no design discussion needed):**
- Recorded priority-aware throttling as an explicit non-goal, gated on the `ClientKeyExtractor` successor

**Group B (design decisions), cross-checked against actual implementation code — not guessed:**
- B4, B8: recorded decisions already made in open PRs #3627 (shared block-read bulkhead) and #3628 (per-client admission mechanism for `serverStatus`)
- B1, B2, B3, B7: validated directly against the exploratory spike on `spike/3530-throttle-phase1-mechanism-serverstatus` (#3535) — read the actual `WeightedThrottledServiceInterface`, `GetBlockWeigher`, `SubscribeStreamWeigher` code and their unit/e2e tests, not just the PR description. B3 corrects an earlier commit on this branch that guessed wrong before that spike code was checked.

**Follow-up refinements after initial resolution:**
- Added an `Alternatives considered` entry explaining why the weighted admission decision (B1) isn't split between `open()` (global/per-client concurrency) and `onNext` (rate check only): there's no weight-agnostic version of the concurrency checks to split off, since those ceilings are fields on each weight class's own `ThrottlePolicy`; a provisional-admit-then-reconcile approach would also reopen a release/reacquire race window on top of the permit-release subtlety already documented elsewhere in the doc.
- Clarified "Where admission control attaches," which previously read as if the admission check always runs synchronously inside `open()` — true for most methods, but not for the two weighted ones (B1) — ahead of the fuller explanation later in the doc.
- A clarity/conciseness pass over the whole document: consistent terminology, a fixed comma splice, tightened sentences, aligned cross-reference wording between the `ContentAwareWeigher` entity and where it's explained in full.

## Caveat

B1/B2/B3/B7's resolutions are validated against an **exploratory spike** (#3535, not a merged/clean PR) — described in the doc as the current implementation's behavior, with the underlying code cited. Since this PR was opened, the actual production implementation has started landing as real (draft) PRs forward-ported from that spike onto current `main`: #3628 (Phase 1 — core mechanism + `serverStatus`) and #3627 (Phase 4 — shared block-read bulkhead). If either changes materially from what's described here before merging, these sections may need a follow-up.

Resolves #3630

### DIFF
--- a/docs/design/apis/api-throttling.md
+++ b/docs/design/apis/api-throttling.md
@@ -83,7 +83,8 @@ client or API triggered it.
 - **Leaky bucket (rate limiter)** — The rate-limiting model used per client per method: a request adds to a
   conceptual bucket, the bucket drains at a fixed rate, and a request is admitted only if the bucket isn't already
   full. Used strictly as a policer here (reject on arrival), not a shaper (delay and re-admit later); see
-  [Alternatives considered](#alternatives-considered).
+  [Alternatives considered](#alternatives-considered). The state representation used to implement this model is an
+  implementation choice, not part of the model itself — see [`GcraLimiter`](#gcralimiter).
 - **Bulkhead** — A bounded pool of permits that caps how many callers can concurrently use a shared resource,
   independent of who those callers are.
 - **Concurrency permit** — A slot representing one in-flight call or session against a limit; acquired on admission
@@ -98,9 +99,11 @@ per-client concurrency ceiling, and a node-wide concurrency ceiling.
 
 ### `GcraLimiter`
 
-A lock-free leaky-bucket rate limiter, keyed per client, implemented via GCRA. Holds one monotonic-clock timestamp
-(the theoretical arrival time) per key, advanced via compare-and-swap on each admitted request — deliberately not a
-token counter, since that would need the same timestamp anyway (to know when to refill) plus a second field.
+A lock-free leaky-bucket rate limiter, keyed per client, implemented via the
+[Generic Cell Rate Algorithm](https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm) (GCRA). Holds one
+monotonic-clock timestamp (the theoretical arrival time) per key, advanced via compare-and-swap on each admitted
+request — deliberately not a token counter, since that would need the same timestamp anyway (to know when to
+refill) plus a second field.
 
 ### `ClientKeyExtractor`
 
--- a/docs/design/apis/api-throttling.md
+++ b/docs/design/apis/api-throttling.md
@@ -224,10 +224,23 @@ immediately after a business-logic error).
 **Content-aware weighting for `getBlock` and `subscribeBlockStream`.** A single static weight per API cannot express
 that a historical block read is more expensive than a live one. Both APIs register a `ContentAwareWeigher` that
 inspects the requested block number (for `getBlock`) or the requested start block (for `subscribeBlockStream`)
-against the current recent/historical boundary, so a historical request is checked against a stricter policy than
-a live one. For `subscribeBlockStream`, this classification happens once, at admission time — a session that starts
-as a live subscription and later needs to catch up on history is protected by Component B below, not by
-re-evaluating its weight mid-session.
+against the recent/historical boundary described below, so a historical request is checked against a stricter
+policy than a live one. For `subscribeBlockStream`, this classification happens once, at admission time — a
+session that starts as a live subscription and later needs to catch up on history is protected by Component B
+below, not by re-evaluating its weight mid-session.
+
+**The recent/historical boundary** is distance from the tip: a request is historical if it targets a block more
+than a configurable `historicalThresholdBlocks` behind the current maximum available block, queried fresh on each
+call rather than cached — both `getBlock` and `subscribeBlockStream` already read that same value unconditionally
+elsewhere on their normal request path, so this doesn't add a new kind of cost, just one more read of a value
+already on the hot path. `getBlock`'s `retrieveLatest` (and a missing or negative block number) is always
+classified live, rather than resolving the actual latest block number just to weigh the call — the classification
+itself does not need to know what "latest" resolves to, only that the request isn't asking for something behind
+it. This threshold is a self-contained approximation, deliberately independent of the recent-storage-tier plugin's
+own retention boundary so the weigher doesn't require a specific storage-tier plugin to be present — the two are
+expected to be configured to match, but nothing enforces that if either is changed independently. Whether this
+approximation tracks the real recent/historical boundary closely enough in practice belongs in the acceptance
+tests for the implementation (see [Acceptance Tests](#acceptance-tests)).
 
 This changes where in the call the ordered checks above actually run, for these two APIs specifically: since
 classification needs the request bytes, and those aren't available until `onNext` (see `ContentAwareWeigher`), a
@@ -517,3 +530,6 @@ produced. `publishBlockStream` is not subject to any admission check and is unaf
 9. A benchmark measuring per-call latency and allocation rate, with admission control enabled versus disabled on the
    same hardware, shows no meaningful regression on throttled APIs and zero measurable overhead on
    `publishBlockStream`.
+10. The recent/historical boundary's default (`historicalThresholdBlocks`) tracks the recent-storage-tier plugin's
+    own retention boundary closely enough in practice that newly-ingested blocks are not misclassified as
+    historical under normal operation, despite the two values not being enforced to stay in sync.
--- a/docs/design/apis/api-throttling.md
+++ b/docs/design/apis/api-throttling.md
@@ -241,6 +241,13 @@ historical-`getBlock` ceiling and the bulkhead's permit count independently is a
 a guarantee that the bulkhead can never be contended by both sources at once. The bulkhead's own bounded behavior
 (reject or brief wait, never unbounded growth) is what keeps that scenario safe even so.
 
+The initial implementation uses one shared pool for all tiers, including recent/live reads — there is no per-tier
+bulkhead, and a `getBlock` call for a live block acquires and releases a permit the same as a historical one. This
+was a deliberate simplicity choice for the first delivery (see [Extensibility](#extensibility)) rather than a claim
+that a recent-block read is free of overhead: if warm-path latency sensitivity proves material, a separate,
+generously-sized bulkhead for recent reads — or skipping the bulkhead entirely for reads served from an
+in-memory/warm-file path — is the documented extension point to use.
+
 ### Configuration ownership
 
 Each gRPC method has exactly one plugin that implements it, so a client's rate and per-client concurrency limits for
--- a/docs/design/apis/api-throttling.md
+++ b/docs/design/apis/api-throttling.md
@@ -119,9 +119,16 @@ mechanism.
 
 ### `ContentAwareWeigher`
 
-An optional, per-API function that classifies a request into a weight class based on its content, evaluated once at
-admission time before any rate/concurrency check runs. Used by `getBlock` and `subscribeBlockStream` to distinguish
-live/recent requests from historical ones.
+An optional, per-API function that classifies a request into a weight class based on its content. Used by
+`getBlock` and `subscribeBlockStream` to distinguish live/recent requests from historical ones.
+
+A weigher cannot classify a call inside `open()` — for both unary and server-streaming calls, the request bytes
+are not available there; they arrive later, via `onNext` on the pipeline `open()` returns. Where a weigher is
+registered, admission is therefore deferred in full: `open()` only extracts the client key and constructs the
+delegate's pipeline (cheap — no business logic runs yet), and classification plus every Component A check (global
+concurrency, per-client concurrency, rate) run together, once, in a wrapping pipeline's own `onNext`, before the
+delegate's `onNext` — and therefore its business logic — ever runs. A rejected call never reaches the delegate at
+all, so nothing is admitted provisionally and no check is ever paid for twice.
 
 ### The admission decorator
 
@@ -193,7 +200,8 @@ For every call, in order — the first check that rejects wins, and no later che
 
 If every check passes, the call is admitted: both concurrency counters are incremented, and the real service's
 `open()` is invoked. If any check fails, the call is rejected immediately (see [Exceptions](#exceptions)) and the
-real service method is never invoked.
+real service method is never invoked. (For `getBlock` and `subscribeBlockStream` specifically, these checks run at
+a different point in the call than described above — see "Content-aware weighting" below for why.)
 
 **Overload prevention vs. single-consumer abuse are different guarantees.** The global concurrency check above,
 and `BlockReadBulkhead` (Component B), are identity-agnostic: they cap total in-flight work regardless of who's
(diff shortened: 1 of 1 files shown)


docs/design/apis/api-throttling.md
```diff
@@ -176,13 +191,14 @@ For every call, in order — the first check that rejects wins, and no later che
...
    (a pure read of per-client state)
-3. **Rate check (leaky bucket, via GCRA)** — is this client calling faster than its allowed rate? This check is the only one that
-   mutates state (it advances the client's theoretical-arrival-time marker), so it deliberately runs last: a call
-   that's going to be rejected by a cheaper check must not be allowed to consume a rate-limiting slot first.
+3. **Rate check (leaky bucket, via GCRA)** — is this client calling faster than its allowed rate? This check is the only
+   one that mutates state (it advances the client's theoretical-arrival-time marker), so it deliberately runs last.
+   A call that's going to be rejected by a cheaper check must not be allowed to consume a rate-limiting slot first.
 
 If every check passes, the call is admitted: both concurrency counters are incremented, and the real service's
 `open()` is invoked. If any check fails, the call is rejected immediately (see [Exceptions](#exceptions)) and the
```
rockysingh, line 199: Weighted paths (`getBlock`, `subscribeBlockStream`) call `delegate.open()` before admission, so `open()` does run for rejected calls. Worth softening this line so implementers don't put side effects in `open()` assuming otherwise.

docs/design/apis/api-throttling.md
```diff
@@ -145,23 +150,33 @@ can be introduced later as a new implementation of this one interface, without c
 `ThrottlePolicy`, or any configuration record. `RequestOptions.remoteCertificateChain()` already exists on the
 underlying request options today, unused — it is exactly what a future `TlsCertificateKeyExtractor` would read from.
 
+Selection is wiring-driven, not configuration-driven: the registration point constructs the extractor directly,
+one instance per throttled service, so switching implementations needs a code change and a restart, not a config
+change. Each throttled service also owns its own isolated per-client state table, so two extractors producing the
```
rockysingh, line 155: This says per-service isolated tables, but L295 describes one map keyed by (client, method). Which is it? If per-service, the `Client-state table size` gauge on L503 needs a service label.

docs/design/apis/api-throttling.md
```diff
@@ -192,23 +208,36 @@ contrast, are a best-effort deterrent for the common case (a buggy or greedy sin
 authenticated identity, they cannot be a guarantee against a motivated, distributed adversary.
 
 **Concurrency-permit lifecycle.** The decorator must release a call's concurrency permit exactly once, whenever the
```
rockysingh, line 210: For a weighted call rejected in `onNext`, `delegate.open()` already ran and built a pipeline. Who sends the terminal on that responses pipeline? Leak risk if unspecified.

rockysingh (commented):
Three consistency questions on the throttling model. All non-blocking, but flagging anyway.

PR #3641, opened 2026-09-11
chore(verification): RSAProofVerifier Support for V2/V5

### LINKED ISSUE #3640: RSAProofVerifier rejects historical V2/V5 WRB blocks — blocks Tier 1 backfill from full-history Tier 0
## Summary

`RSAProofVerifier` in the reworked verification plugin (PR #2869) is scoped Phase-2a to `SignedRecordFileProof.version == 6` and hard-rejects V2 and V5. This blocks any Tier 1 Block Node backfilling from a full-history Tier 0 (e.g. the mainnet SP-WRB BN on s03) because the mainnet WRB archive begins with V2 and V5 record files.

## Observed failure

Reported by @Nana-EC while testing a Tier 1 BN backfilling from Tier 0:

```
WARNING [org.hiero.block.node.block.verification.verifier.RSAProofVerifier verify]
  Unsupported SignedRecordFileProof version 2 in block 1,502 - only V6 is supported

WARNING [org.hiero.block.node.block.verification.session.SessionResultHandler handleThrowable]
  Session for block 1502 with source BACKFILL completed exceptionally

VerificationSessionFailedException: Verification session for block 1501 with source BACKFILL
  failed with MISSING_MANDATORY_FIELD
```

Failures start from block 0 and continue through every V2/V5-era block. Nana had to scale their BN down; this is a blocker for Tier 1 rollout.

## Root cause

**Wrap side is behaving correctly.** `RecordBlockConverter.java:68` and `RecordBlockConverter.java:175` preserve the source record-file format version in the `SignedRecordFileProof`:

```java
new SignedRecordFileProof(universalRecordFile.recordFormatVersion(), signatures)
```

This is intentional: the signatures on V2/V5 blocks were computed over V2/V5 serialization, not over the normalized-to-V6 protobuf payload. Relabeling them as version=6 would break signature verification because the hash algorithm differs by version.

**Verifier side is scoped too narrowly.** `RSAProofVerifier.java:107`:

```java
} else if (version != 6) {
    // Phase 2a scope: only record file format version 6 is supported
    LOGGER.log(WARNING,
        \"Unsupported SignedRecordFileProof version {0} in block {1} - only V6 is supported\",
        version, blockNumber);
    result = SessionFailureType.MISSING_MANDATORY_FIELD;
}
```

Javadoc line 76 also explicitly states: *\"Algorithm (V6 only for Phase 2a)\"*. There is a passing unit test at `RSAProofVerifierTest.java:460` that codifies the V5 rejection: *\"Version 5 - only V6 is supported in Phase 2a\"*.

## Why the Tier 0 BNs never hit this

The Tier 0 mainnet SP-WRB BN's archive (blocks 0 through ~99.6M) was populated by file-level cp (`05-backfill-wrb-to-bn.sh`), so those blocks never went through `RSAProofVerifier`. Only live-pushed blocks (99.6M+) get verified, and those are all firmly V6. A Tier 1 that backfills from Tier 0 via `subscribeBlockStream` starts from block 0 and hits every V2/V5 block.

## Not a regression per se

The pre-rework verification plugin had zero real WRB verification either — `DummyVerificationSession` rubber-stamped pre-0.72 blocks. The rework strictly improved verification (real algorithm vs stub) but under-scoped the version support. From an actual-security standpoint the new behavior is better; from a Tier 1 rollout standpoint the Phase-2a scope needs to lift.

## Proposed fix

Extend `RSAProofVerifier` to support V2 and V5 in addition to V6. Concretely:

1. **Port V2/V5 signed-payload construction** from the tools' `SignatureDataExtractor` (`tools-and-tests/tools/src/main/java/org/hiero/block/tools/blocks/validation/SignatureDataExtractor.java`) into the block-verification module. That extractor already documents both algorithms:
   > For V6, only the raw RecordStreamFile bytes are captured. For V5/V2, the start/end running hashes and per-item raw transaction/record bytes are also extracted.
2. **Update the hashing stage upstream of the verifier** to compute the version-appropriate signed payload (V6 today produces `SHA-384(int32(6) || rawRecordStreamFileBytes)`; V2/V5 need the per-item + running-hash construction).
3. **Extend `RSAProofVerifier.verify()`** to switch on `version` and use the right signed payload.
4. **Update `RSAProofVerifierTest.java:460`** — the test currently pins the rejection behavior; it should be flipped to assert successful verification of V2/V5 blocks (with real V2/V5 fixture data).
5. **Refresh javadoc / comments** — remove Phase-2a scope notes on `RSAProofVerifier.java:76,107`.

## Related

- Reworked verification plugin: PR #2869 (Atanas Atanasov)
- Historical-roster support: PR #3122 (David Berry)
- Wrap-side version preservation: original in commit `ce40ea6987` (Jasper Potts, PR #1769); parallel-pipeline copy at commit `5578cce6d1` (Jasper Potts, PR #2258); mirror call site added in `d5ab8beb47` (Rocky Singh).

## Reviewer

@ata-nas — you built the reworked verifier and know the Phase-2a scope decisions; would appreciate your review on both the fix and any test-fixture concerns for V2/V5 sample blocks.

rockysingh:
* Add RecordFileSignedPayload, internal to block-verification and self-contained, porting the V2/V5 signed-payload reconstruction from the tools' SignatureDataExtractor
* Compute the signed payload in RSAProofVerifier from the block's RECORD_FILE item, per the record file format version the proof declares
* Accept proof versions 2, 5 and 6; classify every payload failure (missing item, unsupported version, missing components, malformed contents)
* Supply the block and HAPI proto version to RSAProofVerifier and remove the V6-only payload computation from BlockHasher
* Update the WRB verification scope notes in bootstrap-roster-plugin.md
* Add tests
  * RecordFileSignedPayloadTest against an independent legacy-format reconstruction oracle, covering the payload constructions and the block-level entry point
  * V2/V5 acceptance, wrong-version-payload, missing-component and mixed-version proof tests
  * Real-data hasher, verifier and full-plugin verification of V2/V5 fixtures

Resolves #3640


block-node/app/src/testFixtures/resources/test-blocks/WRB/V2/address-book.json
```diff
@@ -0,0 +1,41 @@
+{
+  "nodeAddress": [{
```
Nana-EC, line 2: Q: is this the genesis address book? File name should probably highlight genesis of the file contents should note the era if we're using the appropriate schema

block-node/app/src/testFixtures/resources/test-blocks/WRB/V5/address-book.json
```diff
@@ -0,0 +1,74 @@
+{
+  "nodeAddress": [{
```
Nana-EC, line 2: Similar needs to highlight starting block

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
```diff
@@ -108,8 +107,9 @@ public final class BlockHasher implements Supplier<HashingResult> {
     /// The HAPI proto version from the block header.
     private SemanticVersion hapiProtoVersion;
     /// Raw serialized bytes of the outer `RecordFileItem` proto message, captured when a
-    /// `RECORD_FILE` item is seen. Field 2 of this proto holds the `record_file_contents`
-    /// bytes required to compute the V6 signed payload. Null until such an item is encountered.
+    /// `RECORD_FILE` item is seen and used to detect a second such item (a WRB block carries
```
Nana-EC, line 110: Q: did you want to be more specific here? Your suggesting the existence of a second block item but it's not clear what the 2nd item is or should be given it's always 2

ivannov (approved):

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RecordFileSignedPayload.java
jsync-swirlds, line 1: Why did we copy a lot of code from the hashing utility and rsa proof verifier?
I might be missing something, but it seems like these methods should remain where they were, or be moved to a local utility class instead of here.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
```diff
@@ -62,7 +62,7 @@ The BN automatically determines which proof type to verify based on the proof pr
   `BlockNodeContext` and notifies all plugins via `onContextUpdate`.
 - Periodically refresh the address book from both the peer BN and Mirror Node while running.
 - Define the RSA signature verification algorithm precisely enough to be implemented from this document. All record
-  file format versions that exist on mainnet (v2, v5 and v6) are supported; v2/v5 support was added by #3640.
+  file format versions that ever existed (v2, v5 and v6) are supported; v2/v5 support was added by #3640.
```
jsync-swirlds, line 65: This isn't quite accurate, 1, 2, and 3 did exist, they just were never used in a current production network.

jsync-swirlds (approved):
A couple non-blocking items.

Nana-EC (approved):
LG.
Confirmed these changes in [0.41.1-wrbv2v5.2](https://github.com/hiero-ledger/hiero-block-node/tree/refs/tags/v0.41.1-wrbv2v5.2) and they are operational.
Let's merge this, cherry pick to 0.42 and 0.41 and apply improvements based on feedback in 0.44

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/verifier/RecordFileSignedPayload.java
ata-nas: I wanted to keep everything in one place, that was my reason. We can revert if preferred in a follow up. The local utility class is the one you comment on.

docs/design/wrb-streaming/bootstrap-roster-plugin.md
ata-nas: Thanks, will update in a follow up.

block-node/app/src/testFixtures/resources/test-blocks/WRB/V2/address-book.json
ata-nas: These test fixtures are taken from tools.

block-node/app/src/testFixtures/resources/test-blocks/WRB/V5/address-book.json
ata-nas: These test fixtures are taken from tools.

block-node/block-verification/src/main/java/org/hiero/block/node/block/verification/hasher/BlockHasher.java
ata-nas: Here we mean that a WRB is expected to have only one record file item. We still need to parse it in order to add to root hash produced by hasher, but we also assert that the field is not empty, failing fast if it is, and if we detect a second such item, we fail also.

PR #3645, opened 2026-09-14
fix: MirrorNodeUtils.readUrl does not retry UnknownHostException / SocketTimeoutException

### LINKED ISSUE #3644: MirrorNodeUtils.readUrl does not retry UnknownHostException / SocketTimeoutException — kills the CLI JVM on transient DNS blips
## Summary

`MirrorNodeUtils.readUrl` has retry logic, but its `isRetryableException()` gate only matches on **exception message strings**. Type-based transient failures like `UnknownHostException` (DNS failure) and `SocketTimeoutException` don't have those strings in their messages, so the retry is skipped, the wrapper throws `UncheckedIOException`, and the caller in `LiveSequential.processBlocksSequentially` (line 653) lets it bubble to `main`, terminating the JVM.

Consequence: a **transient GCP DNS blip on the mirror-node hostname** (`{mainnet,testnet,previewnet}-public.mirrornode.hedera.com`) kills the live-sequential push process outright. Observed multiple times on our SP-WRB Tier-0 CLIs (mainnet, testnet, previewnet) — sometimes taking down all three simultaneously when the DNS event affects the shared GCP resolver.

## Repro / observation

Stack tail seen on every SP-WRB CLI crash (mainnet example):

```
Caused by: java.io.UncheckedIOException: java.net.UnknownHostException: mainnet-public.mirrornode.hedera.com
        at org.hiero.block.tools.mirrornode.MirrorNodeUtils.readUrl(MirrorNodeUtils.java:82)
        at org.hiero.block.tools.mirrornode.MirrorNodeUtils.readUrl(MirrorNodeUtils.java:33)
        ...
Caused by: java.net.UnknownHostException: mainnet-public.mirrornode.hedera.com
        at java.base/sun.nio.ch.NioSocketImpl.connect(...)
        at java.base/java.net.HttpURLConnection.getResponseCode(HttpURLConnection.java:493)
        at org.hiero.block.tools.mirrornode.MirrorNodeUtils.readUrl(MirrorNodeUtils.java:52)
```

The CLI then runs its shutdown hook cleanly (`Shutdown requested... Draining live-push client (queue=0 acked=N submitted=N)`) and exits — but the operator has to notice and restart.

Same failure class also observed as `java.net.SocketTimeoutException: Connect timed out` on GCS list operations (the CLI has other network calls that share this fragility).

## Root cause

`tools-and-tests/tools/src/main/java/org/hiero/block/tools/mirrornode/MirrorNodeUtils.java:93-103`:

```java
private static boolean isRetryableException(IOException e) {
    String msg = e.getMessage();
    if (msg == null) return false;
    return msg.contains("503")
            || msg.contains("502")
            || msg.contains("500")
            || msg.contains("504")
            || msg.contains("429")
            || msg.contains("Connection reset")
            || msg.contains("Connection timed out");
}
```

- `UnknownHostException.getMessage()` returns just the hostname (`"mainnet-public.mirrornode.hedera.com"`) → no match → not retried.
- `SocketTimeoutException.getMessage()` is usually `"Connect timed out"` or `"Read timed out"` — the first matches by luck of the `Connection timed out` substring but the "Read timed out" variant does not.
- `ConnectException` messages vary by JDK; "Connection refused" isn't matched.

## Proposed fix

Switch to a **type-based check** as the primary gate, and keep the message-substring list as a fallback for `IOException` with an unclear subtype. Concretely, add `instanceof` cases for the classic transient network types before the string match:

```java
private static boolean isRetryableException(IOException e) {
    // Type-based check for classic transient network failures.
    if (e instanceof UnknownHostException
            || e instanceof SocketTimeoutException
            || e instanceof java.net.ConnectException
            || e instanceof java.net.NoRouteToHostException) {
        return true;
    }
    // Fallback: message-substring match for cases where the type is a bare IOException
    // (e.g. the HTTP-status shim above that packages 5xx/429 as IOException).
    String msg = e.getMessage();
    if (msg == null) return false;
    return msg.contains("503")
            || msg.contains("502")
            || msg.contains("500")
            || msg.contains("504")
            || msg.contains("429")
            || msg.contains("Connection reset")
            || msg.contains("Connection timed out")
            || msg.contains("Broken pipe");
}
```

## Operational impact

- Mainnet SP-WRB CLI: killed by this ~5 times since 2026-09-04.
- Testnet + previewnet SP-WRB CLIs: killed simultaneously with mainnet twice (looks like a shared GCP DNS event).
- Every occurrence requires an operator to notice and re-run `06-run-live-with-bn.sh` (or the env's equivalent). The BN is fine — only the CLI push side dies — but the BN stops advancing until the CLI comes back.

## Related follow-ups

- The 30s `connectTimeout` / `readTimeout` in `readUrl` (lines 50-51) is quite short for public mirror-node endpoints under load; leaving as-is is fine given the retry, but worth revisiting once the retry is fixed.
- The same fragility likely exists in the GCS listing path (`SocketTimeoutException: Connect timed out` on GCS list, seen killing the CLI similarly). Out of scope for this issue but worth a separate one if it repeats.

ivannov:
## Reviewer Notes

* The retry check in mirror node reads looks at the exception type first and keeps the existing message matching as a fallback for HTTP 5xx/429 responses
* Adds unit tests for the retry decision
 
## Related Issue(s)

Closes #3644


tools-and-tests/tools/src/main/java/org/hiero/block/tools/mirrornode/MirrorNodeUtils.java
```diff
@@ -90,16 +94,22 @@ public static JsonObject readUrl(URL url) {
...
+                || e instanceof NoRouteToHostException
+                || (msg != null
+                        && (msg.contains("503")
+                                || msg.contains("502")
+                                || msg.contains("500")
+                                || msg.contains("504")
+                                || msg.contains("429")
+                                || msg.contains("Connection reset")
+                                || msg.contains("Connection timed out")
+                                || msg.contains("Broken pipe")));
```
ata-nas, line 112: Should we also consider to compare the strings as not case sensitive? Not sure about this, just something that I thought of.

AlfredoG87 (approved):

tools-and-tests/tools/src/test/java/org/hiero/block/tools/mirrornode/MirrorNodeUtilsTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.io.IOException;
+import java.net.ConnectException;
+import java.net.NoRouteToHostException;
+import java.net.SocketTimeoutException;
+import java.net.UnknownHostException;
+import org.junit.jupiter.api.Test;
+
+/** Tests for {@link MirrorNodeUtils}. */
+class MirrorNodeUtilsTest {
```
Nana-EC, line 16: Nice addition.
In follow up the current tests for retryableMessagesAreRetryable() only exercises the "503", "429", "Connection reset", and "Broken pipe" paths of the message fallback. The "500", "502", "504", and "Connection timed out" don't have coverage.

Nana-EC (approved):
LG
UT coverage expansion suggestion

tools-and-tests/tools/src/main/java/org/hiero/block/tools/mirrornode/MirrorNodeUtils.java
rockysingh: We could use: `HttpStatusException && RETRYABLE_HTTP_CODES.contains(e.status())`

I am not going to block it on that. We can add this later on.

rockysingh (approved):
lgtm - does primary what the issue asked for - the rest is nice to have.

tools-and-tests/tools/src/main/java/org/hiero/block/tools/mirrornode/MirrorNodeUtils.java
ivannov: I created a follow-up issue: https://github.com/hiero-ledger/hiero-block-node/issues/3648

tools-and-tests/tools/src/test/java/org/hiero/block/tools/mirrornode/MirrorNodeUtilsTest.java
ivannov: I created a follow-up issue: https://github.com/hiero-ledger/hiero-block-node/issues/3648

PR #3651, opened 2026-09-15
chore: Follow-ups from #3645 review on MirrorNodeUtils.isRetryableException

### LINKED ISSUE #3648: Follow-ups from #3645 review on MirrorNodeUtils.isRetryableException
### Story Form

N / A

### Technical Notes

- The numeric substrings ("500", "502", "503", "504", "429") never match a retryable response and should be removed
  - Retryable codes are handled by `RETRYABLE_HTTP_CODES` and never reach the message check
  - They do match 404s whose URL contains a number like that (e.g. blocks/5000), which get retried 5 times instead of failing fast
- Check HTTP status by type instead of by message: `HttpStatusException` && `RETRYABLE_HTTP_CODES.contains(e.status())`
- Test coverage:
  - Add Connection timed out to `retryableMessagesAreRetryable()`
  - Drop the HTTP 503 / HTTP 429 message assertions, since `readUrl` never passes those messages to the check
  - Add a regression test for a 404 with 500 in the URL

ivannov:
## Reviewer Notes

* Picks retryable HTTP failures by the actual response status, not by matching status numbers in the exception message
* Retryable statuses and network exceptions now share one retry path, which removes the duplicate backoff branch
* Updates the unit tests for the new signature and adds a local HTTP server test showing a 404 fails after one request
  
## Related Issue(s)

Closes #3648


tools-and-tests/tools/src/main/java/org/hiero/block/tools/mirrornode/MirrorNodeUtils.java
```diff
@@ -49,30 +51,23 @@ public static JsonObject readUrl(String url) {
...
-                                + "s (attempt " + attempt + "/" + MAX_RETRIES + ")...");
-                        Thread.sleep(delay);
-                    }
                 } else {
                     throw new IOException("HTTP " + responseCode + " for URL: " + url);
                 }
             } catch (IOException e) {
                 lastException = e;
-                if (attempt < MAX_RETRIES && isRetryableException(e)) {
+                if (attempt < MAX_RETRIES && isRetryableException(responseCode, e)) {
```
berryware: Would it be better to call isRetryableException just once?
`boolean retryable = isRetryableException(responseCode, e);
if (attempt < MAX_RETRIES && retryable) {
    // retry
} else if (!retryable) {
    throw new UncheckedIOException(e);
}`

ivannov: Yes, you are right. Good catch!

berryware (approved):
looks good

Nana-EC (approved):

PR #3653, opened 2026-09-16
chore: Tag-triggered E2E runs start before the release artifacts are published

### LINKED ISSUE #3516: Tag-triggered E2E runs start before the release artifacts are published

ivannov:
## Reviewer Notes

* Blocks the Solo E2E scheduler until the Block Node chart and every plugin jar the chart resolves are retrievable
* Adds the polling script, which fails after a bounded wait with a message naming the missing artifacts
* Wires the wait into the configure job with Helm installed and a job timeout that matches the bounded wait

## Related Issue(s)

Closes #3516


tools-and-tests/scripts/solo-e2e-test/scripts/wait-for-bn-artifacts.sh
```diff
@@ -0,0 +1,79 @@
...
+# and even after the release workflow finishes, Maven Central keeps serving 404 for a while.
+# Without this the run fails with a Helm or plugin-resolver error that looks like a deploy bug.
+#
+# The chart is pushed only after the container images and jars, so its presence implies the
+# images are published too.
+#
+# Usage: ./wait-for-bn-artifacts.sh <bn_version>     (e.g. v0.41.0-rc1 or 0.41.0-rc1)
+# SNAPSHOT versions are skipped: they are republished on every main push.
+
+set -o pipefail
```
berryware: Should the -u option be used here as well?

berryware (dismissed):

tools-and-tests/scripts/solo-e2e-test/scripts/wait-for-bn-artifacts.sh
ivannov: Yes, you are right. Added -u as well

berryware (approved):
Looks good

ata-nas (approved):

AlfredoG87 (approved):
LGTM 👍

PR #3657, opened 2026-09-16
chore: Tolerate Solo 0.88.1 pod-readiness fast-fail during the Solo E2E deploy

### LINKED ISSUE #3649: Tolerate Solo 0.88.1 pod-readiness fast-fail during the Solo E2E deploy
### Story Form

As a Block Node developer reading the Solo E2E results,
I want a Mirror Node pinger that restarts once during startup to be allowed to recover,
So that a topology's Block Node tests run instead of the whole job dying in "Deploy network".

### Technical Notes

Since the Solo bump to 0.88.1 (#3633), 16 of the 38 failed jobs between 09-11 and 09-15 died in "Deploy network" with the same error. No Block Node test ran in any of them:

```
Detected fatal pod state for "mirror-1-pinger-6fc796587d-2rtm2" (1/3); retrying
Detected fatal pod state for "mirror-1-pinger-6fc796587d-2rtm2" (2/3); retrying
[SOLO-3031] Error deploying mirror node: Pod readiness check failed ... app.kubernetes.io/component=pinger ...: Failed to create Kubernetes pod
```

Solo's diagnostics show the pinger container `Running` / `Ready: False`, with `exitCode: 1`, `reason: Error`. So the container exits during startup, Kubernetes restarts it, and it goes into `CrashLoopBackOff`.

Why this started with 0.88.1:

- hiero-ledger/solo#5890 (merged 2026-08-28, shipped in 0.88.1) added `CrashLoopBackOff` to `K8ClientPods.FATAL_WAITING_REASONS`. After `FATAL_ERROR_RETRY_THRESHOLD = 3` checks at 2s intervals, every pod-readiness wait now gives up. The PR was aimed at crashed consensus node containers, but the pinger's readiness wait uses the same code.
- Solo 0.84.1 had no `CrashLoopBackOff` entry and waited up to 900 x 2s for the pinger. On the 09-10 run (Solo 0.84.1) the pinger became Ready 4s into the wait.
- Solo knows the pinger crashes at startup. Its own `mirror-node.ts` defers the pinger in one-shot parallel deploys because starting it too early "crashes it once". Our `solo-deploy-network.sh` calls `solo mirror node add --pinger` directly, so that deferral never applies to us.

The failure is intermittent. On 09-13, `paired-3` and `7cn-3bn-distributed` deployed and passed. `single-wrb-rsa` has not hit it so far.

We can't see why the pinger exits 1. Solo writes the pinger log to `~/.solo/logs/hiero-components-logs/`, but the workflow only uploads `cn-logs`, `BN-logs` and `MN-logs` (importer only).

ivannov:
## Reviewer Notes

* `solo consensus node start` / `solo mirror node add` exiting non-zero no longer fails the deploy outright: wait up to 300s for the pods to become ready, continue if they do
* Still fatal if the pods do not recover. Only the Mirror Node pinger may stay down; it only generates load.
* Same handling in `wrb-sequential-comparison.sh`.
* Artifacts: pinger logs/describe into MN-logs, Solo's `hiero-components-logs` into CN-logs (BN logs stripped as duplicates).

## Related Issue(s)

Closes #3649


tools-and-tests/scripts/solo-e2e-test/scripts/wrb-sequential-comparison.sh
```diff
@@ -59,6 +59,31 @@ function kctl {
     kubectl --context "${CONTEXT}" "$@"
 }
 
+POD_RECOVERY_TIMEOUT_SECONDS=300
```
berryware: Should this be readonly. as well?

tools-and-tests/scripts/solo-e2e-test/scripts/solo-deploy-network.sh
```diff
@@ -791,10 +816,21 @@ function deploy_consensus_nodes {
...
   start_task "Starting consensus nodes"
+  local start_status=0
   solo consensus node start \
     --deployment "${DEPLOYMENT}" \
-    --node-aliases "${NODE_ALIASES}" || fail "ERROR: Failed to start consensus nodes" 1
-  end_task
+    --node-aliases "${NODE_ALIASES}" || start_status=$?
+  if [[ "${start_status}" -ne 0 ]]; then
+    end_task "FAILED (checking pods)"
+    log_line "  WARNING: 'solo consensus node start' failed; checking whether the pods recovered."
```
berryware, line 825: The issue and PR refer to only checking Mirror Node. Should the descriptions be updated?

ivannov: Yeah, initially it was just the mirror node. But after yesterday's run, I realized that Solo has the same behavior towards other pods as well.

I changed the title and the reviewer notes accordingly

berryware (approved):
Looks good

ata-nas (approved):

AlfredoG87 (approved):
LGTM

Issue #3335, opened 2026-07-31
Correct all of the chart config and values files in the charts/block-node-server/values-overrides/ folder


jsync-swirlds:
### Story Form

N/A

### Technical Notes

There are several issues with the current values overrides.  The list below is a starting point, and is not necessarily comprehensive.

- LFH values currently place the `archiveStorage` on the `fast-storage` PVC mount.  The archive storage must be on the `archive-storage` PVC in order to use the bulk drives.
- Production configs are not present or are tuned for workstation-class environments rather than servers.
- Production-intent values files do not set queue sizes and other parameters to make use of production RAM available, reducing performance.
- Some of the smaller environments do not override messaging values to match the environment
- The production environments do not have production-level messaging values, and default values need optimization for production level resources.

We propose to start clean and create a thoroughly considered and well tested set of both configuration and chart values files outside the "chart" subfolders to provide clear and unambiguous recommended settings for different environments.

Additional configuration and values overrides for specific testing environments will be developed, and should stored in the configuration repositories for those environments, rather than in the Block Node repository, to avoid creating unnecessary clutter.

PR #3658, opened 2026-09-17
fix: correct lfh-values.yaml indentation and update provisioner archive storage

Nana-EC:
## Summary
- Fixes indentation in `lfh-values.yaml`: `logs`, `initContainers`, and `persistence` were nested one level too deep under `blockNode.env`, so they weren't being applied at the intended `blockNode` config scope.
- Bumps the lfh provisioner chart version to `0.41.1`.
- Moves `archivePath` from `/mnt/fast-storage/block-node/archive` to a dedicated `/mnt/archive-storage/block-node/archive` in both `lfh-provisioner-config.yaml` and `provisioner-test-config.yaml`.
- Reduces the lfh log volume size from `100Gi` to `50Gi`.

Partially addresses #3335


berryware (approved):
Looks good

rockysingh (approved):

PR #3659, opened 2026-09-18
fix: roster-bootstrap-rsa cannot load the genesis roster once the Mirror Node has ingested the genesis address book, so WRB-RSA topologies store no blocks

### LINKED ISSUE #3650: `roster-bootstrap-rsa` cannot load the genesis roster once the Mirror Node has ingested the genesis address book, so WRB-RSA topologies store no blocks
### Story Form

As a Block Node operator running RSA/WRB verification on a fresh network,
I want the Block Node to resolve the genesis address book to block 0 whatever the Mirror Node has ingested,
So that the Block Node can verify and store blocks from block 0 instead of rejecting every block with `MISSING_VERIFICATION_DATA`.

### Technical Notes

Since the Solo 0.88.1 bump, every `single-wrb-rsa` run (09-11 through 09-15) and every `wrb-differential-test` run (09-12 through 09-15) fails the same way:

```
FAIL all-have-blocks: block-node-1: No blocks available (firstAvailableBlock=18446744073709551615 ...)
FAIL rsa-verification-active: block-node-1: RSA roster not loaded (roster_entries_loaded=0)
```

On the 09-10 run (Solo 0.84.1) and on the 09-15 tag run (Solo 0.84.1), both topologies pass.

The BN log shows:

```
06:33:12 SEVERE  RSA address book could not be created - Mirror Node API unavailable   (until MN is up)
06:36:42 WARNING HTTP 404 from Mirror Node blocks API at http://mirror-1-restjava:80/api/v1/blocks?timestamp=gte:1789454107.246481181&order=asc&limit=1
06:36:42 WARNING Mirror Node returned zero nodes with a valid public_key from http://mirror-1-restjava:80.
... repeats every 5s until the log is collected; roster never loads
VerificationSessionFailedException: Verification session for block 0 with source PUBLISHER failed with MISSING_VERIFICATION_DATA
```

The CN side shows `EndOfStream ... responseCode: ERROR`, then `BLOCK_NODE_BEHIND` and `No block nodes available for streaming`.

How it breaks (`RsaRosterBootstrapPlugin.collectEras` / `fetchBlockRange`):

1. The plugin maps each Mirror Node address book era to a block range. When an era's `from` is blank or below 1s (`isGenesisFrom`), it maps straight to block 0. Any other era needs `GET /api/v1/blocks?timestamp=gte:<from>`.
2. At startup the Mirror Node serves only its bootstrap address book, whose `from` is a genesis sentinel. Once the importer processes the genesis `0.0.101`/`0.0.102` file update from the stream, the era's `from` becomes that transaction's real consensus timestamp. In the failing run that is `1789454107.246481179`, and the plugin's query asks for `gte:1789454107.246481181`.
3. `mirrorNodeBaseUrl` points at the `rest-java` service, and `rest-java` returns 404 for `/api/v1/blocks`. In the 09-10 run the Mirror Node had ingested 122 blocks and the endpoint still returned 404 every minute. `fetchBlockRange` returns null, the entry is skipped, and the plugin logs a misleading "zero nodes with a valid public_key".
4. So the roster only loads if the plugin polls in the window between the Mirror Node loading its bootstrap book and ingesting the genesis file update.
   - Solo 0.84.1: the Mirror Node read blocks from the Block Node (`blocknode` profile). The Block Node rejects everything until it has a roster, so the Mirror Node could not ingest the genesis update first. The window stayed open and the roster loaded (06:37:17 bootstrap book, 06:37:22 roster loaded, 06:37:40 genesis update).
   - Solo 0.88.1: the Mirror Node starts in RECORD mode and reads record files straight from the CN. It processed the genesis update at 06:36:40.977, 1.3s after the bootstrap book (06:36:39.637) and before the plugin's first poll (06:36:42). The window closes before the first poll, every time.

Solo 0.88.1 only exposed this. The plugin cannot load the genesis roster from any Mirror Node that has already ingested the genesis address book update, unless `/api/v1/blocks` is reachable at the same base URL.

ivannov:
## Reviewer Notes

* Loads the RSA roster from the Mirror Node even when the blocks API cannot place any address-book era, by treating the oldest era as the genesis book covering block 0 onwards
* Restricts the fallback to the initial history build and to the case where no era resolved at all, so a working blocks API stays authoritative and a newer unplaceable era is dropped with a warning instead of claiming block 0
* Keeps the history at one era when a rescued era is later placed by a recovered blocks API, rather than splitting it into two identical books
* Fixes the roster poll interval key in the Solo deploy overlay, which previously mapped to no property

## Related Issue(s)

Closes #3650


berryware (approved):
Looks good

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -371,19 +374,26 @@ private void fetchFromMirrorNode() {
...
     /// (possibly empty) map of era key to {@link Era} otherwise.
     private Map<String, Era> collectEras(
-            final HttpClient nodesClient, final HttpClient blocksClient, final long currentLastStartBlock) {
+            final HttpClient nodesClient,
+            final HttpClient blocksClient,
+            final long currentLastStartBlock,
+            final boolean fullBuild) {
         final Map<String, Era> eras = new LinkedHashMap<>();
         String nextUrl = config.mirrorNodeBaseUrl() + "/api/v1/network/nodes?limit=" + config.mirrorNodePageSize()
                 + "&order=desc";
```
ata-nas: I do see we call the API to order for us, but is it worth it to also have a tree map here? Not trusting the API per se?

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
```diff
@@ -477,13 +560,20 @@ private long[] fetchBlockRange(final HttpClient client, final String from, final
...
+    /// the nanosecond part, which only has to separate address-book updates seconds apart.
+    private static double parseFromSeconds(final String from) {
         if (from == null || from.isBlank()) {
-            return true;
+            return 0.0;
         }
         try {
-            return Double.parseDouble(from) < 1.0;
+            return Double.parseDouble(from);
         } catch (final NumberFormatException e) {
```
ata-nas: Is this worth a TRACE/DEBUG log?

ata-nas (approved):
LG, couple of questions, optional.

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
ivannov: Actually nowhere in the code we rely on ordering of the eras. Once they are collected and processed, we do an explicit sort and order them by starting block. So linked hash map is not required here either

block-node/roster-bootstrap-rsa/src/main/java/org/hiero/block/node/roster/bootstrap/rsa/RsaRosterBootstrapPlugin.java
ivannov: Looking at other locations in the Block Node, where we return a sentinel like value (-1, false, etc.), we don't log a message anywhere. So I believe this is the pattern for the project

